PR #782, opened 2022-10-24
Update React Query to version 4

### LINKED ISSUE #717: Update React Query to version 4
Since the creation of #542, React Query 4 has become stable. We should update as soon as possible. There are several breaking changes, but I went through the list and nothing will seem to affect any of our functionalities.

The migration should be fairly straightforward following their guide: https://tanstack.com/query/v4/docs/guides/migrating-to-react-query-4

sodic:
Fixes #717, #542.


waspc/data/Generator/templates/react-app/src/actions/index.js
```diff
@@ -1,7 +1,7 @@
 import {
   useMutation,
   useQueryClient,
-} from 'react-query'
+} from '@tanstack/react-query'
```
sodic, line 4: Reference: https://tanstack.com/query/v4/docs/guides/migrating-to-react-query-4#react-query-is-now-tanstackreact-query

waspc/data/Generator/templates/react-app/src/auth/useAuth.js
```diff
@@ -18,4 +18,4 @@ async function getMe() {
   }
 }
 
-getMe.queryCacheKey = 'auth/me'
+getMe.queryCacheKey = ['auth/me']
```
sodic, line 21: Reference for this and all other array-related changes: https://tanstack.com/query/v4/docs/guides/migrating-to-react-query-4#query-keys-and-mutation-keys-need-to-be-an-array

waspc/data/Generator/templates/react-app/src/operations/resources.js
```diff
@@ -47,10 +47,10 @@ export async function removeQueries() {
 
 export async function invalidateAndRemoveQueries() {
   const queryClient = await queryClientInitialized
-  // If we don't invalidate the queries before removing them, Wasp will stay on
+  // If we don't reset the queries before removing them, Wasp will stay on
   // the same page. The user would have to manually refresh the page to "finish"
   // logging out.
-  queryClient.invalidateQueries()
+  queryClient.resetQueries()
```
sodic, line 53: This was a tricky one to get working, still not entirely sure about the difference between `invalidateQueries` and `resetQueries`.

From what I understood from the documentation, we should have always been using `resetQueries`, but `invalidateQueries` was working by accident.

It seems like `invalidateQueries` has been updating queries to their pre-loaded state until version 4 ([but wasn't supposed to](https://tanstack.com/query/v4/docs/guides/migrating-to-react-query-4#query-keys-and-mutation-keys-need-to-be-an-array)).

Anyway, I'll create a different issue for getting to the bottom of what's happening here, but this change matches our current behavior.

sodic:
There were no user-facing changes (i.e., no need to update the docs).

We weren't using most of the features that had breaking changes in this version. I wasn't sure what to do about [this one](https://tanstack.com/query/v4/docs/guides/migrating-to-react-query-4#consistent-behavior-for-cancelrefetch). I tested our apps out with both options, we weren't counting on any kind of default behavior. The new default seemed reasonable, so I left it in.

Martinsos (approved):
Sounds good to me!
If they think it is reasonable defaults, and you also thought it through a bit and tested and it makes sense, then I don't think I have anything to add. We can get back to it in the future if really needed.
LGTM!

waspc/data/Generator/templates/react-app/src/operations/resources.js
Martinsos: Interesting, I remember we discussed this but couldn't quiet remember all the details. What you wrote sounds reasonable!

shayneczyzewski:
Nice job here @sodic! Glad we have upgraded another core dep! 🥳

PR #786, opened 2022-10-27
Feat: Validate wasp version from spec

### LINKED ISSUE #578: Add tracking of wasp version to the Wasp project
Idea: what if we had a field in .wasp that specifies which Wasp version is used for building specific project? I think at some point we will want to have that for sure, right?
So when you want to run a Wasp project, you are sure you are not using wrong Wasp compiler version. It should be enough to just specify major.minor version -> then we know that it has to be that major version, and minor version has to be bigger then specific minor version. So basically if we have 3.2 in the file, we will allow 3.3, 3.4 and similar, but won't allow 4.0 or anything above that. This assumes you won't ever want single Wasp project to support being built with different major versions, which I think it is ok assumption for now.
We could say format has to be waspVersion: ^major.minor or waspVersion: ^0.major.minor to also support alpha/beta releases, and ^ is there to makeit more clear how wasp handles this version + to possibly make it easier to add more freedom in the future.

I would imagine this field would be required, so you can't have Wasp project for which it is not clear which Wasp version needs to be used. And if your `wasp` CLI doesn't match the version of the project, it would throw an error.

Field could be `app.waspVersion`, or maybe `wasp.version` (with `wasp` being new declaration for `wasp` specific config. In theory it could even be a separate file, but it seems weird to have it in separate file when we have .wasp.

NeoLight1010:
# Description

This PR adds validation of the specific Wasp version required to build an app. This was done by adding the field `wasp.version` to `app` in the app's specification, and validating it in the `validateAppSpec` function of `AppSpec.Valid`.

- The only valid format for `app.wasp.version` for now is `^0.major.minor.version` (check the linked issue for the discussion).
- The error message when the current version is invalid is similar to the Node version mismatch message, as suggested by @sodic .
- The `CreateNewApp` module was changed so that the `wasp new` command adds the `app.wasp.version` field to the generated spec file.
- Tests were added for this feature.
-  E2E tests were modified for the changes.
- The documentation was extended to include the new `app.wasp` dictionary field and the `app.wasp.version` string field.

#### New changes after review

- waspc's version was changed to use regular SemVer: `0.6.0.0` -> `0.6.0` 

Fixes #578 

> *Note*
> I'm not too good with Haskell yet, but I refactored my code multiple times to make it better.
> I'll be happy to refactor the code if you can guide me on where it can improve. Thank you and sorry for the inconvenience.

## Type of change

- [ ] Code cleanup

### DIFF
--- a/waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
+++ b/waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
@@ -6,6 +6,8 @@ where
 import Control.Monad.Except (throwError)
 import Control.Monad.IO.Class (liftIO)
 import Data.List (intercalate)
+import qualified Data.Version
+import qualified Paths_waspc
 import StrongPath (Abs, Dir, File', Path', Rel, reldir, relfile, (</>))
 import qualified StrongPath as SP
 import System.Directory (createDirectory, getCurrentDirectory)
@@ -20,8 +22,6 @@ import qualified Wasp.Cli.Common as Common
 import qualified Wasp.Data
 import Wasp.Util (indent, kebabToCamelCase)
 import qualified Wasp.Util.Terminal as Term
-import qualified Data.Version
-import qualified Paths_waspc
 
 data ProjectInfo = ProjectInfo
   { _projectName :: String,
--- a/waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
+++ b/waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
@@ -20,6 +20,8 @@ import qualified Wasp.Cli.Common as Common
 import qualified Wasp.Data
 import Wasp.Util (indent, kebabToCamelCase)
 import qualified Wasp.Util.Terminal as Term
+import qualified Data.Version
+import qualified Paths_waspc
 
 data ProjectInfo = ProjectInfo
   { _projectName :: String,
@@ -122,6 +124,9 @@ createNewProject' (ProjectInfo projectName appName) = do
     mainWaspFileContent =
       unlines
         [ "app %s {" `printf` appName,
+          "  wasp: {",
+          "    version: \"^%s\"" `printf` Data.Version.showVersion Paths_waspc.version,
+          "  },",
           "  title: \"%s\"" `printf` projectName,
           "}",
           "",
--- a/waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
+++ b/waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
@@ -6,8 +6,6 @@ where
 import Control.Monad.Except (throwError)
 import Control.Monad.IO.Class (liftIO)
 import Data.List (intercalate)
-import qualified Data.Version
-import qualified Paths_waspc
 import StrongPath (Abs, Dir, File', Path', Rel, reldir, relfile, (</>))
 import qualified StrongPath as SP
 import System.Directory (createDirectory, getCurrentDirectory)
@@ -22,6 +20,7 @@ import qualified Wasp.Cli.Common as Common
 import qualified Wasp.Data
 import Wasp.Util (indent, kebabToCamelCase)
 import qualified Wasp.Util.Terminal as Term
+import qualified Wasp.Version as WV
 
 data ProjectInfo = ProjectInfo
   { _projectName :: String,
@@ -125,7 +124,7 @@ createNewProject' (ProjectInfo projectName appName) = do
       unlines
         [ "app %s {" `printf` appName,
           "  wasp: {",
-          "    version: \"^%s\"" `printf` Data.Version.showVersion Paths_waspc.version,
+          "    version: \"^%s\"" `printf` show WV.waspVersion,
           "  },",
           "  title: \"%s\"" `printf` projectName,
           "}",
--- a/waspc/src/Wasp/AppSpec/Valid.hs
+++ b/waspc/src/Wasp/AppSpec/Valid.hs
@@ -8,8 +8,14 @@ module Wasp.AppSpec.Valid
   )
 where
 
+import Control.Monad (unless)
 import Data.List (find)
 import Data.Maybe (isJust)
+import qualified Data.Version as DV
+import GHC.Natural (naturalFromInteger)
+import qualified Paths_waspc
+import Text.Read (readMaybe)
+import Text.Regex.TDFA ((=~))
 import Wasp.AppSpec (AppSpec)
 import qualified Wasp.AppSpec as AS
 import Wasp.AppSpec.App (App)
@@ -23,13 +29,7 @@ import qualified Wasp.AppSpec.Entity as Entity
 import qualified Wasp.AppSpec.Entity.Field as Entity.Field
 import qualified Wasp.AppSpec.Page as Page
 import Wasp.AppSpec.Util (isPgBossJobExecutorUsed)
-import Text.Regex.TDFA ((=~))
 import qualified Wasp.SemanticVersion as SV
-import GHC.Natural (naturalFromInteger)
-import qualified Paths_waspc
-import qualified Data.Version as DV
-import Control.Monad (unless)
-import Text.Read (readMaybe)
 
 data ValidationError = GenericValidationError String
   deriving (Show, Eq)
--- a/waspc/src/Wasp/AppSpec/Valid.hs
+++ b/waspc/src/Wasp/AppSpec/Valid.hs
@@ -59,9 +59,7 @@ validateExactlyOneAppExists spec =
           "You have more than one 'app' declaration in your Wasp app. You have " ++ show (length apps) ++ "."
 
 validateWasp :: AppSpec -> [ValidationError]
-validateWasp spec = validateWaspVersion specWaspVersionStr
-  where
-    specWaspVersionStr = Wasp.version $ App.wasp (snd $ getApp spec)
+validateWasp = validateWaspVersion . Wasp.version . App.wasp . snd . getApp
 
 validateWaspVersion :: String -> [ValidationError]
 validateWaspVersion specWaspVersionStr = eitherUnitToErrorList $ do
--- a/waspc/src/Wasp/AppSpec/Valid.hs
+++ b/waspc/src/Wasp/AppSpec/Valid.hs
@@ -69,6 +69,8 @@ validateWaspVersion specWaspVersionStr = eitherUnitToErrorList $ do
     )
     $ Left $ incompatibleVersionError currentWaspVersion specWaspVersionRange
   where
+    -- TODO: Use version range parser from SemanticVersion when it is fully implemented.
+
     parseWaspVersionRange :: String -> Either ValidationError SV.Range
     parseWaspVersionRange waspVersionRangeStr = do
       let (_ :: String, _ :: String, _ :: String, waspVersionRangeDigits :: [String]) =
--- a/waspc/src/Wasp/AppSpec/Valid.hs
+++ b/waspc/src/Wasp/AppSpec/Valid.hs
@@ -92,10 +92,9 @@ validateWaspVersion specWaspVersionStr = eitherUnitToErrorList $ do
           ]
 
     currentWaspVersion :: SV.Version
-    currentWaspVersion = SV.Version (intToNat major) (intToNat minor) (intToNat patch)
+    currentWaspVersion = SV.Version (toEnum major) (toEnum minor) (toEnum patch)
       where
         DV.Version [major, minor, patch] _ = Paths_waspc.version
-        intToNat = naturalFromInteger . toInteger
 
     eitherUnitToErrorList :: Either e () -> [e]
     eitherUnitToErrorList (Left e) = [e]
--- a/waspc/src/Wasp/AppSpec/Valid.hs
+++ b/waspc/src/Wasp/AppSpec/Valid.hs
@@ -72,7 +72,7 @@ validateWaspVersion specWaspVersionStr = eitherUnitToErrorList $ do
     parseWaspVersionRange :: String -> Either ValidationError SV.Range
     parseWaspVersionRange waspVersionRangeStr = do
       let (_ :: String, _ :: String, _ :: String, waspVersionRangeDigits :: [String]) =
-            waspVersionRangeStr =~ ("\\^([0-9]+).([0-9]+).([0-9]+)$" :: String)
+            waspVersionRangeStr =~ ("\\`\\^([0-9]+)\\.([0-9]+)\\.([0-9]+)\\'" :: String)
 
       waspSpecVersion <- case mapM readMaybe waspVersionRangeDigits of
         Just [major, minor, patch] -> Right $ SV.Version major minor patch
--- a/waspc/src/Wasp/AppSpec/Valid.hs
+++ b/waspc/src/Wasp/AppSpec/Valid.hs
@@ -71,6 +71,9 @@ validateWaspVersion specWaspVersionStr = eitherUnitToErrorList $ do
 
     parseWaspVersionRange :: String -> Either ValidationError SV.Range
     parseWaspVersionRange waspVersionRangeStr = do
+      -- Only ^x.y.z is allowed here because it was the easiest solution to start
+      -- with at the moment. In the future, we plan to allow any SemVer
+      -- definition.
       let (_ :: String, _ :: String, _ :: String, waspVersionRangeDigits :: [String]) =
             waspVersionRangeStr =~ ("\\`\\^([0-9]+)\\.([0-9]+)\\.([0-9]+)\\'" :: String)
 
--- a/waspc/src/Wasp/AppSpec/Valid.hs
+++ b/waspc/src/Wasp/AppSpec/Valid.hs
@@ -65,28 +65,30 @@ validateWasp spec = validateWaspVersion specWaspVersionStr
 
 validateWaspVersion :: String -> [ValidationError]
 validateWaspVersion specWaspVersionStr = eitherUnitToErrorList $ do
-  specWaspVersion <- parseWaspVersion specWaspVersionStr
+  specWaspVersionRange <- parseWaspVersionRange specWaspVersionStr
   unless
-    ( SV.isVersionInRange currentWaspVersion $
-        SV.Range [SV.backwardsCompatibleWith specWaspVersion]
+    ( SV.isVersionInRange currentWaspVersion specWaspVersionRange
     )
-    $ Left $ incompatibleVersionError currentWaspVersion specWaspVersion
+    $ Left $ incompatibleVersionError currentWaspVersion specWaspVersionRange
   where
-    parseWaspVersion :: String -> Either ValidationError SV.Version
-    parseWaspVersion waspVersionStr = do
-      let (_ :: String, _ :: String, _ :: String, waspVersionDigits :: [String]) =
-            waspVersionStr =~ ("\\^([0-9]+).([0-9]+).([0-9]+)$" :: String)
-      case mapM readMaybe waspVersionDigits of
+    parseWaspVersionRange :: String -> Either ValidationError SV.Range
+    parseWaspVersionRange waspVersionRangeStr = do
+      let (_ :: String, _ :: String, _ :: String, waspVersionRangeDigits :: [String]) =
+            waspVersionRangeStr =~ ("\\^([0-9]+).([0-9]+).([0-9]+)$" :: String)
+
+      waspSpecVersion <- case mapM readMaybe waspVersionRangeDigits of
         Just [major, minor, patch] -> Right $ SV.Version major minor patch
         __ -> Left $ GenericValidationError "Wasp version should be in the format ^major.minor.patch"
 
-    incompatibleVersionError :: SV.Version -> SV.Version -> ValidationError
-    incompatibleVersionError actualVersion expectedVersion =
+      Right $ SV.Range [SV.backwardsCompatibleWith waspSpecVersion]
+
+    incompatibleVersionError :: SV.Version -> SV.Range -> ValidationError
+    incompatibleVersionError actualVersion expectedVersionRange =
       GenericValidationError $
         unwords
           [ "Your Wasp version does not match the app's requirements.",
             "You are running Wasp " ++ show actualVersion ++ ".",
-            "This app requires Wasp ^" ++ show expectedVersion ++ "."
+            "This app requires Wasp " ++ show expectedVersionRange ++ "."
           ]
 
     currentWaspVersion :: SV.Version
--- a/waspc/src/Wasp/AppSpec/Valid.hs
+++ b/waspc/src/Wasp/AppSpec/Valid.hs
@@ -25,9 +25,11 @@ import qualified Wasp.AppSpec.Page as Page
 import Wasp.AppSpec.Util (isPgBossJobExecutorUsed)
 import Text.Regex.TDFA ((=~))
 import qualified Wasp.SemanticVersion as SV
-import GHC.Natural (wordToNatural)
+import GHC.Natural (naturalFromInteger)
 import qualified Paths_waspc
 import qualified Data.Version as DV
+import Control.Monad (unless)
+import Text.Read (readMaybe)
 
 data ValidationError = GenericValidationError String
   deriving (Show, Eq)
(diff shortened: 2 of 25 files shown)


waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -49,6 +56,55 @@ validateExactlyOneAppExists spec =
...
+        incompatibleVersionMessage actualVersion expectedVersion =
+          unwords
+            [ "Your Wasp version does not match the app's requirements.",
+              "You are running Wasp 0." ++ show actualVersion ++ ".",
+              "This app requires Wasp ^0." ++ show expectedVersion ++ "."
+            ]
+
+        currentWaspSemVer = SV.Version (toEnum currentMajor) (toEnum currentMinor) (toEnum currentPatch)
+          where
+            DV.Version (_ : currentMajor : currentMinor : currentPatch : _) _ = Paths_waspc.version
```
Martinsos, line 106: Great work here, especially taking into account you are relatively new to Haskell!!!

You did a pretty good job here, but I had a feeling couple things might be simplified. One thing that I notice first is three levels of `where` -> that is valid, but is a bit unusual and it does make reading the function a bit harder. Yeah, I had a bit hard time reading the function, although it wasn't badly written, it just felt a bit too verbose actually.

I also noticed you used `Word` -> You can instead directly read those version digits into Natural, `read` adapts to read whatever you need.

Finally, I wasn't sure what happens when you call `toEnum` to convert Int to Natural, so I implemented function for it instead.

Then I reorganized it all a bit also, and I made it clear we are dealing with pre-release versions.

Here is the version I ended up with:
```hs
validateWaspVersion :: String -> [ValidationError]
validateWaspVersion specWaspVersionStr = eitherUnitToErrorList $ do
  specWaspPreReleaseVersion <- parseWaspPreReleaseVersion specWaspVersionStr
  unless
    ( SV.isVersionInRange currentWaspPreReleaseVersion $
        SV.Range [SV.backwardsCompatibleWith specWaspPreReleaseVersion]
    )
    $ Left $ incompatiblePreReleaseVersionError currentWaspPreReleaseVersion specWaspPreReleaseVersion
  where
    -- Given a wasp pre-release version (meaning it is not production ready yet,
    -- so it starts with 0) of format '^0.major.minor.patch`, it returns semver
    -- version major.minor.patch .
    parseWaspPreReleaseVersion :: String -> Either ValidationError SV.Version
    parseWaspPreReleaseVersion waspVersionStr = do
      let (_ :: String, _ :: String, _ :: String, waspVersionDigits :: [String]) =
            waspVersionStr =~ ("\\^0\\.([0-9]+).([0-9]+).([0-9]+)$" :: String)
      case mapM readMaybe waspVersionDigits of
        Just [major, minor, patch] -> Right $ SV.Version major minor patch
        __ -> Left $ GenericValidationError "Wasp version should be in the format ^0.major.minor.patch"

    incompatiblePreReleaseVersionError :: SV.Version -> SV.Version -> ValidationError
    incompatiblePreReleaseVersionError actualVersion expectedVersion =
      GenericValidationError $
        unwords
            "You are running Wasp 0." ++ show actualVersion ++ ".",
            "This app requires Wasp ^0." ++ show expectedVersion ++ "."
          ]

    currentWaspPreReleaseVersion :: SV.Version
    currentWaspPreReleaseVersion = SV.Version (intToNat major) (intToNat minor) (intToNat patch)
      where
        DV.Version [0, major, minor, patch] _ = Paths_waspc.version
        intToNat = naturalFromInteger . toInteger

    eitherUnitToErrorList :: Either e () -> [e]
    eitherUnitToErrorList (Left e) = [e]
    eitherUnitToErrorList (Right ()) = []

```

I thought this reads a bit easier, what do you think?

Regardless of all of this, I had another thought -> a lot of complexity here is coming from the fact that we have this special case of `0.x.y.z` version. This is not standard Semver, standard Semver is `0.x.y` where they treat `x` as major and `y` as minor. This is my fault -> I for some reason introduced this non-standard way for versioning some time ago.
Now that I think about it, I don't think we gained much by doing `0.x.y.z`, and we could probably benefit from switching to `0.x.y` which is standard! What do you think?
So what I would propose, is that we actually refactor now this PR assuming that we use `0.x.y` for versioning, and that the version under which this code will be next released is `0.7.0`. While doing this reactoring you can look at your old code, my code that I proposed above, and try to get the best out of that all.
That should simplify this piece of code, and also make our lives easier in the future!

Btw we are actually handling correctly case in SemVer when first digit is 0, in our SemanticVersion.hs, so that is already taken care of.

Martinsos:
This is awesome @NeoLight1010 , really nice and thorough work!!! I enjoyed reviewing it.
I left some suggestions on how to improve the code (which is already pretty good though!), and also a suggestion that we switch from 0.a.b.c to 0.a.b which should make our lives easier, check it out.

waspc/src/Wasp/AppSpec/Valid.hs
NeoLight1010: Thank you so much! Your refactor is so much better than my implementation. I learned a lot just by reading it :smile: 

I think changing the versioning scheme to regular SemVer would be really nice! It would clear up any confusion that may arise about Wasp's version (actually, I was confused about this when I started working on the issue).

What should we do about the tests, though? They will fail until the project's version is changed to `0.7.0`. Unless we change the version in the `cabal` file in this same PR.

> standard Semver is 0.x.y where they treat x as major and y as minor

I think SemVer specifies `0.y.z` to be `major: 0`, `minor: y`, `patch: z`. https://semver.org/#spec-item-4

Let me know what you think :smile:

Martinsos: I am glad you found it useful! I think your implementation was already pretty good though, I mostly did some reorganizing + a bit of monads (Either).

Regarding the version: let's do the following then, let's set project's version from `0.6.0.0` to `0.6.0`, in the project (so in .cabal file, and maybe some other files).
Then once we do new release, we will increase it to 0.7.0 -> but that is not your concern right now.

SemVer and 0.x.y -> you are right, those are still called minor and patch, which is why I said "they treat x as major and y as minor", not that they are actually major and minor. You can check this out here: https://github.com/npm/node-semver#caret-ranges-123-025-004  -> x is considered to be introducing breaking changes, while y is considered to be doing additions.

All together, sounds good to me and let's go with `0.6.0`!

NeoLight1010: I just applied the corresponding changes. Thank you for your help :D

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -49,6 +58,47 @@ validateExactlyOneAppExists spec =
...
+    )
+    $ Left $ incompatibleVersionError currentWaspVersion specWaspVersion
+  where
+    parseWaspVersion :: String -> Either ValidationError SV.Version
+    parseWaspVersion waspVersionStr = do
+      let (_ :: String, _ :: String, _ :: String, waspVersionDigits :: [String]) =
+            waspVersionStr =~ ("\\^([0-9]+).([0-9]+).([0-9]+)$" :: String)
+      case mapM readMaybe waspVersionDigits of
+        Just [major, minor, patch] -> Right $ SV.Version major minor patch
+        __ -> Left $ GenericValidationError "Wasp version should be in the format ^major.minor.patch"
```
Martinsos, line 81: Now that we have normal semver, we could actually produce a range here -> because it what we are really dealing with here, with range.

So we could change the signature to
```hs
parseSpecWaspVersionRange :: String -> Either ValidationError -> SV.Range
```

and then also modify `incompatibleVersionError` to be `SV.version -> SV.Range -> ValidationError`.

I think that would make it a bit nicer, more to the point.

waspc/test/AppSpec/ValidTest.hs
```diff
@@ -41,6 +44,50 @@ spec_AppSpecValid = do
...
 
+    describe "'waspVersion' validation" $ do
+      describe "should validate 'waspVersion' format" $ do
+        it "returns no error if waspVersion is compatible" $ do
+          ASV.validateAppSpec basicAppSpec `shouldBe` []
+
+        it "returns an error if 'waspVersion' has an incorrect format" $ do
+          let basicAppWithInvalidWaspVersionFormat = basicApp {AS.App.wasp = AS.Wasp.Wasp {AS.Wasp.version = "0.5.2"}}
+          let basicAppDeclWithInvalidWaspVersionFormat =
+                AS.Decl.makeDecl "TestApp" basicAppWithInvalidWaspVersionFormat
```
Martinsos, line 55: I would just inline this into `AS.decls = [<inline here>]`, I don't think you get much with this extra name.

waspc/test/AppSpec/ValidTest.hs
```diff
@@ -41,6 +44,50 @@ spec_AppSpecValid = do
...
+                basicApp {AS.App.wasp = AS.Wasp.Wasp {AS.Wasp.version = "^" ++ DV.showVersion incompatibleWaspVersion}}
+
+          let basicAppDeclWithIncompatibleWaspVersion =
+                AS.Decl.makeDecl "TestApp" basicAppWithIncompatibleWaspVersion
+
+          ASV.validateAppSpec
+            ( basicAppSpec
+                { AS.decls = [basicAppDeclWithIncompatibleWaspVersion]
+                }
+            )
```
Martinsos, line 82: This part is very similar to the same piece of code you have in the test above -> basically taking basicAppSpec and modifying setting its wasp.version field.

I would advise you either create a function for it and use it in these two places, or you modify the process of creating basicAppSpec so it takes version as an argument (but that might be a bit bigger change).

web/docs/language/features.md
```diff
@@ -1173,6 +1181,30 @@ entity SocialLogin {=psl
...
+```
+
+`app.wasp` is a dictionary with the following fields:
+
+#### `version: string` (required)
+
+`version` declares the compatible Wasp versions for the app. It is specified
+in a SemVer format.
+
+**Note**: For now, `version` only supports the format `^0.major.minor.patch`. This
```
Martinsos, line 1205: Let's update this one to be `^0.y.z`.

Martinsos (commented):
Almost there @NeoLight1010 !
I left a couple more comments, check them out, and after this I think we can merge.

web/docs/language/features.md
NeoLight1010: As the `SemanticVersion` module handles the `0.y.z` case correctly, I think it's okay to leave it as `^x.y.z` instead. What do you think?

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -49,6 +58,49 @@ validateExactlyOneAppExists spec =
...
+  where
+    specWaspVersionStr = Wasp.version $ App.wasp (snd $ getApp spec)
+
+validateWaspVersion :: String -> [ValidationError]
+validateWaspVersion specWaspVersionStr = eitherUnitToErrorList $ do
+  specWaspVersionRange <- parseWaspVersionRange specWaspVersionStr
+  unless
+    ( SV.isVersionInRange currentWaspVersion specWaspVersionRange
+    )
+    $ Left $ incompatibleVersionError currentWaspVersion specWaspVersionRange
```
sodic, line 72: You can use list comprehensions to avoid wrapping in and out of Either and return a list of errors directly:
```haskell
validateWaspVersion :: String -> [ValidationError]
validateWaspVersion specWaspVersionStr = case parseWaspVersionRange specWaspVersionStr of
  Left err -> [err]
  Right versionRange ->
    [ incompatibleVersionError currentWaspVersion versionRange
      | not $ SV.isVersionInRange currentWaspVersion versionRange
    ]
  where
    ...
```
Doing so will allow you to remove the `eitherUnitToErrorList` function.

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -49,6 +58,49 @@ validateExactlyOneAppExists spec =
...
+          [ "Your Wasp version does not match the app's requirements.",
+            "You are running Wasp " ++ show actualVersion ++ ".",
+            "This app requires Wasp " ++ show expectedVersionRange ++ "."
+          ]
+
+    currentWaspVersion :: SV.Version
+    currentWaspVersion = SV.Version (intToNat major) (intToNat minor) (intToNat patch)
+      where
+        DV.Version [major, minor, patch] _ = Paths_waspc.version
+        intToNat = naturalFromInteger . toInteger
```
sodic, line 98: You can use the already available [`toEnum` function](https://hoogle.haskell.org/?hoogle=Int+-%3E+Natural&scope=set%3Ahaskell-platform) instead of defining your own.

sodic (commented):
Great work on this feature, can't wait to ship it!

I've left a couple of code refactoring suggestions, nothing major. Plus, one bonus thing to do if you're up for it (no worries if not).

I'll leave it to @Martinsos to finalize and merge when ready.

waspc/src/Wasp/AppSpec/Valid.hs
NeoLight1010: Thank you for reviewing my PR :smile: 

The use of `eitherUnitToErrorList` comes from [the review by @Martinsos ](https://github.com/wasp-lang/wasp/pull/786#discussion_r1006936789) . Personally, I like more wrapping it in an `Either` as it seems more appropriate for a validation function, but I'm not sure which one is more readable. What do you think?

waspc/src/Wasp/AppSpec/Valid.hs
NeoLight1010: This one also comes from @Martinsos 's review. In fact, I was using `toEnum` in the first iteration of this PR:

> Finally, I wasn't sure what happens when you call toEnum to convert Int to Natural, so I implemented function for it instead.

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -49,6 +58,49 @@ validateExactlyOneAppExists spec =
...
+  specWaspVersionRange <- parseWaspVersionRange specWaspVersionStr
+  unless
+    ( SV.isVersionInRange currentWaspVersion specWaspVersionRange
+    )
+    $ Left $ incompatibleVersionError currentWaspVersion specWaspVersionRange
+  where
+    parseWaspVersionRange :: String -> Either ValidationError SV.Range
+    parseWaspVersionRange waspVersionRangeStr = do
+      let (_ :: String, _ :: String, _ :: String, waspVersionRangeDigits :: [String]) =
+            waspVersionRangeStr =~ ("\\^([0-9]+).([0-9]+).([0-9]+)$" :: String)
```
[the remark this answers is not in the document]
sodic: Oh, sure, but just a note: no need to implement all of them, just the ones we currently support (which is still many, but fewer than all :).

waspc/src/Wasp/AppSpec/Valid.hs
sodic: > Personally, I like more wrapping it in an Either as it seems more appropriate for a validation function

I'd normally agree, but in this case we only wrap the `Either` value to unwrap it (i.e., we get no benefits from using the Either type, only overhead). In other words, this function is the validation function and is set to return a list of errors. I don't think we're gaining much by implementing its body in terms of Either if we aren't using it for anything besides immediately converting it into a list. Plus, it adds extra complexity for the reader (i.e., an additional local function, a `do` block, a call to `unless`).
Moreover, the part of the code that uses `Either` isn't even a local function (i.e., inside a where clause), it's completely inline.

All in all, unless I'm missing something, I'd just directly construct a list. @Martinsos can correct me if I am.

waspc/src/Wasp/AppSpec/Valid.hs
sodic: Whoop, :sweat_smile: 

My bad, missed that comment. I thought there was a chance it might have been done for readability purposes. It's fine either way.

@Martinsos, can't say I fully understand the reasoning though, did you want to say "_it's easier for someone unfamiliar with the API to anticipate what to expect when calling `toInteger` and `naturalFromInteger` than `toEnum` (i.e., this last one might return something weird)?_"

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
```diff
@@ -122,6 +124,9 @@ createNewProject' (ProjectInfo projectName appName) = do
     mainWaspFileContent =
       unlines
         [ "app %s {" `printf` appName,
+          "  wasp: {",
+          "    version: \"^%s\"" `printf` Data.Version.showVersion Paths_waspc.version,
```
sodic, line 128: @Martinsos and @NeoLight1010, what do you guys think, should we abstract this (Wasp's current version) under a module? It's pretty low level and we use it in several places.

If we extract it into a module, the usages becomes a bit cleaner:
```haskell
import Environment (waspVersion) -- I made up a random name

-- And then we just refer to 'waspVersion'

File: waspc/src/Wasp/AppSpec/Valid.hs (hunk 6 above)
Comment (NeoLight1010) [reply waspc/src/Wasp/AppSpec/Valid.hs:70]: You're right. It does add to the complexity of the function. I'm going to refactor it :D

File: waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs (hunk 9 above)
Comment (NeoLight1010) [reply waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs:128]: I think it's a good idea :D using `Paths_waspc` is kind of unintuitive for a Haskell beginner like me. I think `Environment` is a good name for the module.

File: waspc/src/Wasp/AppSpec/Valid.hs (hunk 8 above)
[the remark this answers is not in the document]
Comment (NeoLight1010) [reply waspc/src/Wasp/AppSpec/Valid.hs:75]: You're right. What about implementing `parseVersionRange` in `SemanticVersion`, but supporting only the `^x.y.z` and `x.y.z` cases for now? That way we can expand the functionality later :D

File: waspc/src/Wasp/AppSpec/Valid.hs [hunk 10]
```diff
@@ -49,6 +58,49 @@ validateExactlyOneAppExists spec =
...
+
+      waspSpecVersion <- case mapM readMaybe waspVersionRangeDigits of
+        Just [major, minor, patch] -> Right $ SV.Version major minor patch
+        __ -> Left $ GenericValidationError "Wasp version should be in the format ^major.minor.patch"
```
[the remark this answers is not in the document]
NeoLight1010: That would be great! In that case, the `Wasp version should be in the format ^major.minor.patch` would be inaccurate. I was thinking of something like `Invalid Wasp version range format. Check out the documentation for information on supported version range specifiers: <link to docs>`.

web/docs/language/features.md
Martinsos: Yup, good point, forgot about that! Let's do it!

waspc/src/Wasp/AppSpec/Valid.hs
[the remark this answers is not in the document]
Martinsos: We could! I wanted to make it simple for the start though -> otherwise we can go and support all of the formats? Not sure what is the best hm. I thought this would be easier to parse. For now it doesn't really matter that they can pin the specific version, we are in beta and the point was just to give them something that works, and I thought `^` is a bit nicer than just fixed version. But we could do both.
TLDR; good with me, but we can also leave that for later and then support everything.

waspc/src/Wasp/AppSpec/Valid.hs
Martinsos: @sodic the approach with Either was suggested by me, since it seemed to me that we could benefit from short-circuiting that it offers as a Monad. Validation happens in multiple steps, and can fail in any of those, so it is a good fit for short-circuiting. In this case, it is only two steps, but even with that we get to avoid nesting the second check (isVersionInRange) into the resolution of first step (parsing), so I would say that is a benefit of this approach.

I didn't understand the part with `where` and local function.

I would argue that extra local function is very descriptive and doesn't really affect the reader.

I agree that using monads and therefore do and unless are a bit more complex for the beginner, but we are writing our codebase with assumption that readers do have at least basic understanding of monads, so I wouldn't worry about that.

At the end I think both approaches are ok, the Either one and one you suggested -> I like Either a bit more due to no nesting and ability to expand by adding more steps easily if needed, but I do like that with your approach there is no need for that local function (although that doesn't break the deal for me).

waspc/src/Wasp/AppSpec/Valid.hs
Martinsos: Yup @NeoLight1010 actually used `toEnum` but I suggested this!
To be honest I was going pretty quickly through this and haven't properly investigated what `toEnum` is but concluded that it sounds suspicious and went with this hand rolled solution. Maybe I should have investigated it more though. The thing is, it converts Int to Enum. But Enum is just something enumerated, it doesn't know it is a number. So I felt like there are no guarantees here that it will do reasonable number to number conversion -> Enum cares just about order, not about absolute values.

Looking at it now, I still don't know if toEnum guarantees me what I would expect in this case hm!

On the other hand:
 - `toInteger` -> yup, I do expect this one to convert from Int to corresponding Integer.
 - `naturalFromInteger` -> also, this one is very targeted, I am sure it will give me corresponding natural, and throw an error or something if Integer is not natural.

So I am not sure how to go about this, it does seem a bit silly to write this when there is `toEnum`, and it works as we need it I believe, but in theory I am not sure what is  `toEnum` promising is enough, and I do think reading it in code is a bit confusing (why I am converting number to enum?).

Summary: I am good with using `toEnum` if you like, but it is a bit weird function I think, to be used like this! But maybe I am overthinking it.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
Martinsos: Sounds interesting! It would be nice because we can endow it with extra information -> we know it is semantic version type of version. So we can have it have that type.
We can put it under smth like `Wasp.Version`, instead of using generic name like `Environment`.

waspc/src/Wasp/AppSpec/Valid.hs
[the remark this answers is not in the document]
NeoLight1010: Cool. I think I can leave it as it is now, then

waspc/src/Wasp/AppSpec/Valid.hs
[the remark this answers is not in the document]
NeoLight1010: I agree with you.  Maybe we're starting to fall out of the scope of the PR. I will add the TODO and leave `parseVersionRange` for another PR :D

waspc/src/Wasp/AppSpec/Valid.hs
sodic: Well, @NeoLight1010, I guess you can choose whichever version you like better then :smile:

NeoLight1010: Yup, I guess it's a matter of personal preference at this point. I'll leave the `Either` one for now. Thank you for your feedback :D

waspc/src/Wasp/AppSpec/Valid.hs
sodic: I'd say it's just a standard library function with a bad name. Every language has its fair share of those. I don't think we should be implementing our own version of the same function because we don't like its name - it is what it is :)

Think about PHP's `implode` (a string join operation). It's a horrible name (IMO). Still, if you're a PHP developer, you're stuck with it. Writing your own function that does the same thing and naming it `join` or `intercalate` (also a weird name btw) will only raise questions with readers familiar with the language: "Why did they write their own `implode`? is this an edge case in which they differ? Is the original function not performant enough? Does this one even work correctly?..." 
Finally, they'll maybe have to read (or at least glance at) the source code of your function to figure out what it does. This is not the case when dealing with a standard function (regardless of how bad its name is).

With that said, in this case it's a pretty small and irrelevant change, so we can go either way as far as I'm concerned. But if I'm following my gut, I'd personally just choose `toEnum`.

NeoLight1010: I agree with @sodic here. Also, wow, `implode` is such a bad name o.0 .

I can go with `toEnum`, if that's ok :smile:

NeoLight1010:
I implemented the changes suggested in your reviews :D 

One last thing. I added a commit with changes to the regex used to validate the version range string. It addressed the following:
- The regex wasn't checking that the match is at the **start** of the string. `backslash backtick` is used now for checking this, as specified in the `regex-tdfa` docs.
- The `.` symbols were not escaped, so they were matching any character.
- The `$` token is not specified as supported in the regex-tdfa docs. `backslash '` is now used instead.

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -49,6 +56,43 @@ validateExactlyOneAppExists spec =
...
+  where
+    -- TODO: Use version range parser from SemanticVersion when it is fully implemented.
+
+    parseWaspVersionRange :: String -> Either ValidationError SV.Range
+    parseWaspVersionRange waspVersionRangeStr = do
+      let (_ :: String, _ :: String, _ :: String, waspVersionRangeDigits :: [String]) =
+            waspVersionRangeStr =~ ("\\`\\^([0-9]+)\\.([0-9]+)\\.([0-9]+)\\'" :: String)
+
+      waspSpecVersion <- case mapM readMaybe waspVersionRangeDigits of
+        Just [major, minor, patch] -> Right $ SV.Version major minor patch
```
Martinsos, line 78: Please put a comment here explaining that we use only `^` for now only because it was the easiest solution to start with and that we plan to allow any semver definition in the future.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
```diff
@@ -6,6 +6,8 @@ where
 import Control.Monad.Except (throwError)
 import Control.Monad.IO.Class (liftIO)
 import Data.List (intercalate)
+import qualified Data.Version
+import qualified Paths_waspc
```
Martinsos, line 10: Let's use Wasp.Version here now that we have it!

waspc/test/AnalyzerTest.hs
```diff
@@ -7,6 +7,8 @@ import qualified Data.Aeson as Aeson
 import Data.Either (isRight)
 import Data.List (intercalate)
 import Data.Maybe (fromJust)
+import Data.Version (showVersion)
+import qualified Paths_waspc
```
Martinsos, line 11: Let's use Wasp.Version!

waspc/test/AppSpec/ValidTest.hs
```diff
@@ -41,6 +44,39 @@ spec_AppSpecValid = do
                          "You have more than one 'app' declaration in your Wasp app. You have 2."
                      ]
 
+    describe "'waspVersion' validation" $ do
```
Martinsos, line 47: Nice, these tests now became concise!

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -49,6 +56,43 @@ validateExactlyOneAppExists spec =
...
+validateWasp :: AppSpec -> [ValidationError]
+validateWasp = validateWaspVersion . Wasp.version . App.wasp . snd . getApp
+
+validateWaspVersion :: String -> [ValidationError]
+validateWaspVersion specWaspVersionStr = eitherUnitToErrorList $ do
+  specWaspVersionRange <- parseWaspVersionRange specWaspVersionStr
+  unless
+    ( SV.isVersionInRange WV.waspVersion specWaspVersionRange
+    )
+    $ Left $ incompatibleVersionError WV.waspVersion specWaspVersionRange
```
Martinsos, line 68:
```suggestion
  unless (SV.isVersionInRange WV.waspVersion specWaspVersionRange) $
    Left $ incompatibleVersionError WV.waspVersion specWaspVersionRange
```

Martinsos (approved):
@NeoLight1010 that should be it, this really shaped up nicely!
I left a couple tiny comments but already approved -> pls take a quick stab at those and then I will merge!

Martinsos:
Awesome @NeoLight1010 , merged it! Thanks so much for this work!

I immediately created a follow up task: https://github.com/wasp-lang/wasp/issues/796 . Alas, right now users can't install specific Wasp version! So far it wasn't that important, but now with your PR it becomes important, so therefore issue for that.

NeoLight1010:
Thanks for all your help during this PR @Martinsos @sodic ! I would like to work on that issue. Can you assign it to me, please? :)

Martinsos:
Whoops I didn't see your message and I went to gave it a little try, just to just how difficult it is, and I ended up implementing it! Took me a couple of hours though. @NeoLight1010 it is missing updating of docs but I think it works well, check it out here: https://github.com/wasp-lang/get-wasp/pull/3 and feel free to leave any comments.

But we have other possibly more interesting features: one fun thing is expanding SemanticVersion.hs to have parsing, as @sodic suggested! If we make SemanticVersion.hs proper enough, we could even make it a standalone library on Hackage at some point.

Martinsos:
Whoops I didn't see your message and I went to gave it a little try, just to just how difficult it is, and I ended up implementing it! Took me a couple of hours though. @NeoLight1010 it is missing updating of docs but I think it works well, check it out here: https://github.com/wasp-lang/get-wasp/pull/3 and feel free to leave any comments.

But we have other possibly more interesting features: one fun thing is expanding SemanticVersion.hs to have parsing, as @sodic suggested! If we make SemanticVersion.hs proper enough, we could even make it a standalone library on Hackage at some point.

Martinsos:
Whoops I didn't see your message and I went to gave it a little try, just to just how difficult it is, and I ended up implementing it! Took me a couple of hours though. @NeoLight1010 it is missing updating of docs but I think it works well, check it out here: https://github.com/wasp-lang/get-wasp/pull/3 and feel free to leave any comments.

But we have other possibly more interesting features: one fun thing is expanding SemanticVersion.hs to have parsing, as @sodic suggested! If we make SemanticVersion.hs proper enough, we could even make it a standalone library on Hackage at some point.

PR #788, opened 2022-10-27
Updates node/npm to current LTS

### LINKED ISSUE #768: Upgrade Wasp to Node 18 LTS once it comes out (in Oct '22)

shayneczyzewski:
# Description

Updates node from v16 to the current LTS v18. Updates npm as well. Relevant links:
- https://nodejs.org/en/
- https://nodejs.org/en/download/releases/
- https://github.com/nodejs/Release
- https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V18.md#nodejs-18-changelog

Fixes #768

## Type of change

- [ ] Code cleanup

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -102,7 +102,7 @@ jobs:
       - name: Set up Node
         uses: actions/setup-node@v2
         with:
-          node-version: '16'
+          node-version: '18'
 
       - name: Run tests
         run: cabal test
--- a/waspc/e2e-test/test-outputs/waspMigrate-golden/waspMigrate/.wasp/out/Dockerfile
+++ b/waspc/e2e-test/test-outputs/waspMigrate-golden/waspMigrate/.wasp/out/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:16-alpine AS node
+FROM node:18-alpine AS node
 
 
 FROM node AS base
--- a/waspc/src/Wasp/Generator/Common.hs
+++ b/waspc/src/Wasp/Generator/Common.hs
@@ -24,13 +24,13 @@ nodeVersionRange :: SV.Range
 nodeVersionRange = SV.Range [SV.backwardsCompatibleWith latestNodeLTSVersion]
 
 latestNodeLTSVersion :: SV.Version
-latestNodeLTSVersion = SV.Version 16 0 0
+latestNodeLTSVersion = SV.Version 18 12 0
 
 -- | Range of npm versions that Wasp and generated projects work correctly with.
 npmVersionRange :: SV.Range
 npmVersionRange = SV.Range [SV.backwardsCompatibleWith latestLTSVersion]
   where
-    latestLTSVersion = SV.Version 8 0 0 -- Goes with node 16 (but also higher versions too)
+    latestLTSVersion = SV.Version 8 19 2 -- Goes with node 18 (but also higher versions too).
 
 prismaVersion :: SV.Version
 prismaVersion = SV.Version 3 15 2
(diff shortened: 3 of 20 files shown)


waspc/src/Wasp/Generator/Common.hs
```diff
@@ -24,13 +24,13 @@ nodeVersionRange :: SV.Range
 nodeVersionRange = SV.Range [SV.backwardsCompatibleWith latestNodeLTSVersion]
 
 latestNodeLTSVersion :: SV.Version
-latestNodeLTSVersion = SV.Version 16 0 0
+latestNodeLTSVersion = SV.Version 18 12 0
```
shayneczyzewski, line 27: Instead of just going with `18 0 0` I used the actual minor, since prior minors were not LTS. Same with npm.

.github/workflows/ci.yaml
```diff
@@ -102,7 +102,7 @@ jobs:
       - name: Set up Node
         uses: actions/setup-node@v2
         with:
-          node-version: '16'
+          node-version: '18'
```
sodic, line 105: Should we specify the minor version here as well?

waspc/e2e-test/test-outputs/waspMigrate-golden/waspMigrate/.wasp/out/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:16-alpine AS node
+FROM node:18-alpine AS node
```
sodic, line 1: Same question, no minor here?

web/docs/getting-started.md
```diff
@@ -12,8 +12,8 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
 ## 1. Requirements
 
 You need to have `node` (and `npm`) installed on your machine and available in `PATH`.
-- `node`: 16.x.x
-- `npm`: 8.x.x
+- `node`: ^18.12.0
+- `npm`: ^8.19.2
```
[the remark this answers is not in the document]
shayneczyzewski: Yeah, I debated how to handle this one. The trick is LTS can also increment minor versions, so it won't always be `18.12.x` and could instead be `18.13.x` and above, for example. I think semver may be the best way to capture this accurately.

Here is a screenshot of the prior LTS releases:
Screenshot 2022-10-28 123834

.github/workflows/ci.yaml
shayneczyzewski: Yeah, we could. However, we would also need to remember to update it over time as new LTS minor releases came out. The good thing as of now is since 18 is currently in LTS, each minor beyond that should also be LTS. I don't think if anyone tries to get an 18 from today onwards we should ever go backwards to anything before LTS, either.

waspc/e2e-test/test-outputs/waspMigrate-golden/waspMigrate/.wasp/out/Dockerfile
shayneczyzewski: Ditto to CI answer. I believe it should always be `18.12.0` or higher from now on, and always stay as LTS in here: https://github.com/nodejs/docker-node/blob/f05fbf88068ba29ac9b544a72c9471ba60243e4b/18/alpine3.16/Dockerfile

web/docs/getting-started.md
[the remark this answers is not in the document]
sodic: Right, I completely forgot that users could (and we'd want them to) have larger minor versions too :sweat_smile:.

I guess I was left in the mindset that an LTS version starts with a `0` as its minor, which makes it a bit simpler. `x` on the minor spot basically means "anything above zero", while we can't convey "anything above 12" without a caret.

In fact, our previous specifier (`16.x.x`) was probably wrong (since the previous LTS most likely wasn't `16.0.0`), nice catch!

Anyway, all good, I'm resolving.

sodic (approved):

PR #794, opened 2022-10-31
Improved how Wasp prints output of underlying processes.

### LINKED ISSUE #793: Missing prefix (server/web app/db) for underlying process in `wasp start` when interrupted by other output
If we have following job output coming from server:
```
"Hello\n"
"hi\n"
"wohoo\n"
```

and there is a bit of pause between "hi\n" and "wohoo\n" and some other non-job output gets printed there, for example a warning by Wasp: "[wasp] watch out\n", then "wohoo" won't have "Server:" prefix in front of it, although we want it to have it.

So we will get:
```
Server: Hello
Server: hi
Server:
[wasp] watch out
wohoo
Server:
```

instead of 
```
Server: Hello
Server: hi
Server:
[wasp] watch out
Server: wohoo
Server:
```

This is due to how our logic in Wasp.Generator.Job.IO works, in `printPrefixedJobMessage` -> while we do account for different jobs interrupting each other, so for example if WebApp interrupts Server that will be ok, we don't account for some other input interrupting us.

How can we fix this? Some ideas:
1. When a message ends with "\n", don't print it immediately like we do now, instead "save" it for later, by remembering we have it waiting, and when the next messages comes from that same job, do smth like `print $ "\n" ++ prefix ++ message`. This idea is actually pretty neat, I think I like it.
2. Different idea is to recognize that Wasp has written warnings or some output and based on that add a prefix, same like we do output from other jobs. I like the first idea better though, it is simpler.

To elaborate first idea a bit, I imagine it working smth like this:
```hs
printPrefixedJobMessage :: RemainingNewlines -> Maybe J.JobMessage -> J.JobMessage -> IO RemainingNewlines
printPrefixedJobMessage remainingNewlines maybePrevJobMessage jobMessage = do
  let outHandle = getJobMessageOutHandle jobMessage
      prefix = makeJobMessagePrefix jobMessage
      content = getJobMessageContent jobMessage
      contentEndsWithNewline = "\n" `isSuffixOf` content
      content' = if contentEndsWithNewline then init content else content
      remainingNewlines' = if contentEndsWithNewline
                             then remainingNewlines `set` (J._jobType jobMessage)
                             else remainingNewlines `unset` (J._jobType jobMessage)

  let maybeAddPrefixAtStart =
        if (remainingNewlines `contains` J._jobType jobMessage)
          then "\n" <> prefix
          else ""
      addPrefixAfterSubstr substr = T.intercalate (substr <> prefix) . T.splitOn substr
      addPrefix = maybeAddPrefixAtStart . addPrefixAfterSubstr "\n" . addPrefixAfterSubstr "\r"

  T.IO.hPutStr outHandle $ addPrefix content'
  hFlush outHandle
  return remainingNewlines
```

Yup, I think this should mostly work!
Keep in mind this is a bit pseudocodish and I made up some functions and stuff here.
I imagine RemainingNewlines would be some kind of Set that contains job types -> if it has that job type in it, then there is a remaining newline from that type of job, otherwise there isn't.

Martinsos:
Fixes #793

### DIFF
--- a/waspc/cli/src/Wasp/Cli/Command.hs
+++ b/waspc/cli/src/Wasp/Cli/Command.hs
@@ -17,8 +17,7 @@ newtype Command a = Command {_runCommand :: ExceptT CommandError IO a}
 
 runCommand :: Command a -> IO ()
 runCommand cmd = do
-  errorOrResult <- runExceptT $ _runCommand cmd
-  case errorOrResult of
+  runExceptT (_runCommand cmd) >>= \case
     Left cmdError -> cliSendMessage $ Msg.Failure (_errorTitle cmdError) (_errorMsg cmdError)
     Right _ -> return ()
 
--- a/waspc/src/Wasp/Generator/Job/Common.hs
+++ b/waspc/src/Wasp/Generator/Job/Common.hs
@@ -1,0 +1,24 @@
+module Wasp.Generator.Job.Common
+  ( getJobMessageOutHandle,
+    getJobMessageContent,
+  )
+where
+
+import qualified Data.Text as T
+import System.Exit (ExitCode (..))
+import System.IO (Handle, stderr, stdout)
+import qualified Wasp.Generator.Job as J
+
+getJobMessageOutHandle :: J.JobMessage -> Handle
+getJobMessageOutHandle jobMsg = case J._data jobMsg of
+  J.JobOutput _ outputType ->
+    case outputType of
+      J.Stdout -> stdout
+      J.Stderr -> stderr
+  J.JobExit _ -> stdout
+
+getJobMessageContent :: J.JobMessage -> T.Text
+getJobMessageContent jobMsg = case J._data jobMsg of
+  J.JobOutput output _ -> output
+  J.JobExit ExitSuccess -> "Job exited successfully."
+  J.JobExit (ExitFailure exitCode) -> T.pack $ "Job failed with exit code " <> show exitCode
--- a/waspc/src/Wasp/Generator/Job/IO/PrefixedWriter.hs
+++ b/waspc/src/Wasp/Generator/Job/IO/PrefixedWriter.hs
@@ -1,0 +1,180 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+module Wasp.Generator.Job.IO.PrefixedWriter
+  ( printJobMessagePrefixed,
+    runPrefixedWriter,
+    PrefixedWriter,
+  )
+where
+
+import Control.Monad.IO.Class (MonadIO, liftIO)
+import Control.Monad.State (get, put)
+import Control.Monad.State.Strict (MonadState, StateT, runStateT)
+import qualified Data.Set as S
+import qualified Data.Text as T
+import qualified Data.Text.IO as T.IO
+import System.IO (hFlush, stderr)
+import qualified Wasp.Generator.Job as J
+import Wasp.Generator.Job.Common (getJobMessageContent, getJobMessageOutHandle)
+import qualified Wasp.Util.Terminal as Term
+
+-- |
+-- Imagine you have a job sending following two messages:
+--  1. "First"
+--  2. " line\n"
+--  3. "Second line"
+--
+-- What we want is to prefix new lines with a corresponding job prefix before we print them,
+-- e.g. "Server: ". So we want to output this as:
+--   Server: First line
+--   Server: Second line
+--
+-- This is what this function does, it properly prefixes the given message and then prints it.
+-- Prefixes include job type name, and optional indication that output is stderr, e.g.:
+-- "Server:", "Web app:", "Db(stderr):"
+--
+-- * Implementation details:
+--
+-- Simplest (and naive) way to go about this is to add prefix after any newline.
+-- However, what can happen is that third message from above comes quite later than the second
+-- message, and in the meantime some other output (e.g. from another job) is printed.
+-- In such case, we get following:
+--   Server: First line
+--   Server:
+--   Some other output!
+--   Second line
+--
+-- This is not an easy problem to solve, as we can't know what kind of message is coming next,
+-- and there are always situations where some other output might interrupt us.
+-- But, what we can at least do is avoid having this situation described above, where newline is
+-- "kidnapped", and we do that by postponing the newline (make it pending) for later,
+-- until the next message from the same output (job + output stream) arrives.
+--
+-- Specifically, what we do is postpone printing of a newline if it is the last character in a message.
+-- We make it pending instead, and once the new message comes from the same output, we apply it at the
+-- start of that message.
+--
+-- This way we get proper output in the situation as described above:
+--   Server: First line
+--   Some other output!
+--   Server: Second line
+--
+-- We additionaly check if the last message (before the current message) was from the same output.
+-- If not, or there was no previous message, then we ensure there is prefix at the start of
+-- the message. This helps with situations where output from one job was interrupted by the
+-- output from another job, or when message is the very first message.
+printJobMessagePrefixed :: J.JobMessage -> PrefixedWriter ()
+printJobMessagePrefixed jobMessage = do
+  (PrefixedWriterState outputsWithPendingNewline lastJobMessage) <- get
+
+  let (outputsWithPendingNewline', messageContent) =
+        applyPendingNewline outputsWithPendingNewline jobMessage
+  let prefixedMessageContent = addPrefixWhereNeeded lastJobMessage messageContent
+
+  put $ PrefixedWriterState outputsWithPendingNewline' (Just jobMessage)
+
+  liftIO $ printPrefixedMessageContent prefixedMessageContent
+  where
+    printPrefixedMessageContent :: T.Text -> IO ()
+    printPrefixedMessageContent content = T.IO.hPutStr outHandle content >> hFlush outHandle
+      where
+        outHandle = getJobMessageOutHandle jobMessage
+
+    -- TODO: We haven't considered Windows much here, so in the future we might
+    --   want to check that this works ok on Windows and tweak it a bit if not.
+    addPrefixWhereNeeded :: Maybe J.JobMessage -> T.Text -> T.Text
+    addPrefixWhereNeeded lastJobMessage =
+      ensureNewlineAtStartIfInterruptingAnotherOutput
+        . ensurePrefixAtStartIfNotContinuingOnSameOutput
+        . addPrefixAfterSubstr "\r"
+        . addPrefixAfterSubstr "\n"
+      where
+        addPrefixAfterSubstr :: T.Text -> T.Text -> T.Text
+        addPrefixAfterSubstr substr = T.intercalate (substr <> prefix) . T.splitOn substr
+
+        ensurePrefixAtStartIfNotContinuingOnSameOutput :: T.Text -> T.Text
+        ensurePrefixAtStartIfNotContinuingOnSameOutput text =
+          let continuingOnSameOutput =
+                (getJobMessageOutput <$> lastJobMessage) == Just (getJobMessageOutput jobMessage)
+              prefixAtStart =
+                or [(delimiter <> prefix) `T.isPrefixOf` text | delimiter <- ["\r", "\n", ""]]
+           in if not continuingOnSameOutput && not prefixAtStart then prefix <> text else text
+
+        ensureNewlineAtStartIfInterruptingAnotherOutput :: T.Text -> T.Text
+        ensureNewlineAtStartIfInterruptingAnotherOutput text =
+          let interruptingAnotherOutput =
+                (getJobMessageOutput <$> lastJobMessage) /= Just (getJobMessageOutput jobMessage)
+              newlineAtStart = "\n" `T.isPrefixOf` text
+           in if interruptingAnotherOutput && not newlineAtStart then "\n" <> text else text
+
+        prefix :: T.Text
+        prefix = makeJobMessagePrefix jobMessage
+
+newtype PrefixedWriter a = PrefixedWriter {_runPrefixedWriter :: StateT PrefixedWriterState IO a}
+  deriving (Functor, Applicative, Monad, MonadIO, MonadState PrefixedWriterState)
+
+data PrefixedWriterState = PrefixedWriterState
+  { _outputsWithPendingNewline :: !OutputsWithPendingNewline,
+    _lastJobMessage :: !(Maybe J.JobMessage)
+  }
+
+runPrefixedWriter :: PrefixedWriter a -> IO a
+runPrefixedWriter pw = fst <$> runStateT (_runPrefixedWriter pw) initState
+  where
+    initState =
+      PrefixedWriterState
+        { _outputsWithPendingNewline = S.empty,
+          _lastJobMessage = Nothing
+        }
+
+-- Job message output type.
+data Output = Output
+  { _outputJobType :: !J.JobType,
+    _outputIsStderr :: !Bool
+  }
+  deriving (Eq, Ord)
+
+type OutputsWithPendingNewline = S.Set Output
+
+-- | Given a set of job message outputs with pending newline and a job message,
+-- it applies any pending newline (newline from the previous messages from the same output)
+-- to the job message content while also detecting if content ends with a newline
+-- and in that case adds it to the set of pending newlines (while removing used pending newline).
+-- It returns this updated content and updated set of pending newlines.
+applyPendingNewline ::
+  OutputsWithPendingNewline -> J.JobMessage -> (OutputsWithPendingNewline, T.Text)
+applyPendingNewline outputsWithPendingNewline jobMessage = (outputsWithPendingNewline', content')
+  where
+    content' = addPendingNewlineToStartIfAny $ removeTrailingNewlineIfAny content
+      where
+        removeTrailingNewlineIfAny = if contentEndsWithNewline then T.init else id
+        addPendingNewlineToStartIfAny =
+          if getJobMessageOutput jobMessage `S.member` outputsWithPendingNewline then ("\n" <>) else id
+
+    outputsWithPendingNewline' = updateOp output outputsWithPendingNewline
+      where
+        updateOp = if contentEndsWithNewline then S.insert else S.delete
+
+    contentEndsWithNewline = "\n" `T.isSuffixOf` content
+
+    output = getJobMessageOutput jobMessage
+    content = getJobMessageContent jobMessage
+
+getJobMessageOutput :: J.JobMessage -> Output
+getJobMessageOutput jm =
+  Output
+    { _outputJobType = J._jobType jm,
+      _outputIsStderr = getJobMessageOutHandle jm == stderr
+    }
+
+makeJobMessagePrefix :: J.JobMessage -> T.Text
+makeJobMessagePrefix jobMsg =
+  case J._jobType jobMsg of
+    J.Server -> T.pack $ Term.applyStyles [Term.Magenta] "Server"
+    J.WebApp -> T.pack $ Term.applyStyles [Term.Cyan] "Web app"
+    J.Db -> T.pack $ Term.applyStyles [Term.Blue] "Db"
+    <> ( if getJobMessageOutHandle jobMsg == stderr
+           then T.pack $ Term.applyStyles [Term.Yellow] "(stderr)"
+           else ""
+       )
+    <> ": "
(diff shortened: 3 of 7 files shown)


waspc/cli/src/Wasp/Cli/Command.hs
```diff
@@ -17,8 +17,7 @@ newtype Command a = Command {_runCommand :: ExceptT CommandError IO a}
 
 runCommand :: Command a -> IO ()
 runCommand cmd = do
-  errorOrResult <- runExceptT $ _runCommand cmd
-  case errorOrResult of
+  runExceptT (_runCommand cmd) >>= \case
```
Martinsos, line 20: Unrelated refactoring, but it is nice so I did it!

It is cool, with `\case` (which comes from LambdaCase extension) we can directly feed the result of monadic action into the case instead of having to extract it into a variable first, nice!
@matijaSos this is what I mentioned today to you.

waspc/src/Wasp/Generator/Job/Common.hs
```diff
@@ -0,0 +1,24 @@
+module Wasp.Generator.Job.Common
```
Martinsos, line 1: I just moved this stuff from another file here, this is not new.

waspc/src/Wasp/Generator/NpmInstall.hs
```diff
@@ -101,25 +102,27 @@ loadInstalledFullStackNpmDependencies dstDir = do
...
             J.JobOutput {} ->
-              printPrefixedJobMessage prevJobMsg jobMsg
-                >> go (Just jobMsg) (isWebAppDone, isServerDone) chan
+              printPrefixedJobMessage jobMsg
+                >> go (isWebAppDone, isServerDone) chan
             J.JobExit {} -> case J._jobType jobMsg of
-              J.WebApp -> go (Just jobMsg) (True, isServerDone) chan
-              J.Server -> go (Just jobMsg) (isWebAppDone, True) chan
+              J.WebApp -> go (True, isServerDone) chan
+              J.Server -> go (isWebAppDone, True) chan
```
Martinsos, line 125: Introducing state monad allowed us to hide explicit state from here!
So now here we can call prefixed writing without having to know about how it keeps its state! It just works. Before state was "previous message", now it is not visible. Super cool!

waspc/src/Wasp/Generator/Job/Common.hs
```diff
@@ -0,0 +1,24 @@
...
+    getJobMessageContent,
+  )
+where
+
+import qualified Data.Text as T
+import System.Exit (ExitCode (..))
+import System.IO (Handle, stderr, stdout)
+import qualified Wasp.Generator.Job as J
+
+getJobMessageOutHandle :: J.JobMessage -> Handle
```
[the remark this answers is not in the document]
shayneczyzewski: Tricky call. I think both are valid. I would say here slightly since they are more specific utility than just operating on the types themselves as generic helpers.

waspc/src/Wasp/Generator/Job/IO/PrefixedWriter.hs
```diff
@@ -0,0 +1,125 @@
...
+
+  liftIO $ do
+    T.IO.hPutStr outHandle $ addPrefixWhereNeeded content'
+    hFlush outHandle
+  where
+    addPrefixWhereNeeded :: T.Text -> T.Text
+    addPrefixWhereNeeded =
+      let prefix = makeJobMessagePrefix jobMessage
+          addPrefixAfterSubstr substr = T.intercalate (substr <> prefix) . T.splitOn substr
+       in addPrefixAfterSubstr "\r" . addPrefixAfterSubstr "\n"
```
shayneczyzewski, line 74: Why again are we splitting the `\r` and `\n` cases like this again, vs treating them as `\r\n`?

waspc/src/Wasp/Generator/Job/IO/PrefixedWriter.hs
```diff
@@ -0,0 +1,125 @@
...
+      let addPendingNewlineToStartIfAny =
+            if jobType `S.member` jobsWithPendingNewline then ("\n" <>) else id
+          removeTrailingNewlineIfAny =
+            if contentEndsWithNewline then T.init else id
+       in addPendingNewlineToStartIfAny $ removeTrailingNewlineIfAny content
+    jobsWithPendingNewlines' =
+      if contentEndsWithNewline
+        then S.insert jobType jobsWithPendingNewline
+        else S.delete jobType jobsWithPendingNewline
+    contentEndsWithNewline = "\n" `T.isSuffixOf` content
```
shayneczyzewski, line 116: I found some of the inner "if-ing" a bit hard to grok... is this any easier perhaps?

```haskell
applyPendingNewline ::
  JobsWithPendingNewline -> (J.JobType, T.Text) -> (JobsWithPendingNewline, T.Text)
applyPendingNewline jobsWithPendingNewline (jobType, content) =
  let content' = addPendingNewlineToStartIfAny $ removeTrailingNewlineIfAny content
      jobsWithPendingNewlines' = updateOp jobType jobsWithPendingNewline
   in (jobsWithPendingNewlines', content')
  where
    updateOp :: JobType -> Set JobType -> Set JobType
    updateOp = if contentEndsWithNewline then S.insert else S.delete

    addPendingNewlineToStartIfAny = if jobType `S.member` jobsWithPendingNewline then ("\n" <>) else id
    removeTrailingNewlineIfAny = if contentEndsWithNewline then T.init else id
    contentEndsWithNewline = "\n" `T.isSuffixOf` content
```

waspc/src/Wasp/Generator/NpmInstall.hs
shayneczyzewski: That is very cool you abstracted away the previous message - nice! 👍🏻

shayneczyzewski (commented):
Hey @Martinsos - this is super cool! It made my head hurt, so I am taking a pause for now and will re-review tomorrow with fresh eyes/brain :D Dang Monad Transformers and all that stuff haha

One quick question, is there a clean way to write tests for this? I just did some playing in the repl, and may have found an undesired case, but I wonder if we can do some sort of fuzzing/prop testing (or even explicit unit tests) around this? Maybe not worth it?

Anyways, when I try the following:
```haskell
:set +m

import Wasp.Generator.Job.IO.PrefixedWriter (PrefixedWriter, printPrefixedJobMessage, runPrefixedWriter)
import qualified Wasp.Generator.Job as J

runPrefixedWriter $ do
  printPrefixedJobMessage $ J.JobMessage { J._data = J.JobOutput "line 1a" J.Stdout, J._jobType = J.Server }
  printPrefixedJobMessage $ J.JobMessage { J._data = J.JobOutput "one\n" J.Stdout, J._jobType = J.Db }
  printPrefixedJobMessage $ J.JobMessage { J._data = J.JobOutput "line 1b\n" J.Stdout, J._jobType = J.Server }
  printPrefixedJobMessage $ J.JobMessage { J._data = J.JobOutput "line 2\n" J.Stdout, J._jobType = J.Server }
```

I get this interleaving:
```bash
Server: line 1a
Db: oneline 1b
Server: line 2
```

I am guessing this second output line is not a case we want, or are we explicitly not protecting against that? Anyways, great work, I will resume tomorrow after the brain break!

waspc/src/Wasp/Generator/Job/IO/PrefixedWriter.hs
Martinsos: Pfuf, those are good questions!
`\r` is valid because it can be used to "reset the cartridge" -> so it moves cursor to the start of the line removing previous output and writes what comes after it. It can be used for stuff like showing a progress bar, or percentages growing. Prisma uses it in some of its output, I think npm also possibly (e.g. when installing stuff).

As for windows -> I know they use \r\n in the files, but maybe they don't in the shell?
I would maybe just ignore this one for now, and we can handle that when the windows come :D.

Martinsos:
> Hey @Martinsos - this is super cool! It made my head hurt, so I am taking a pause for now and will re-review tomorrow with fresh eyes/brain :D Dang Monad Transformers and all that stuff haha
> 
> One quick question, is there a clean way to write tests for this? I just did some playing in the repl, and may have found an undesired case, but I wonder if we can do some sort of fuzzing/prop testing (or even explicit unit tests) around this? Maybe not worth it?
> 
> Anyways, when I try the following:
> 
> ```haskell
> :set +m
> 
> import Wasp.Generator.Job.IO.PrefixedWriter (PrefixedWriter, printPrefixedJobMessage, runPrefixedWriter)
> import qualified Wasp.Generator.Job as J
> 
> runPrefixedWriter $ do
>   printPrefixedJobMessage $ J.JobMessage { J._data = J.JobOutput "line 1a" J.Stdout, J._jobType = J.Server }
>   printPrefixedJobMessage $ J.JobMessage { J._data = J.JobOutput "one\n" J.Stdout, J._jobType = J.Db }
>   printPrefixedJobMessage $ J.JobMessage { J._data = J.JobOutput "line 1b\n" J.Stdout, J._jobType = J.Server }
>   printPrefixedJobMessage $ J.JobMessage { J._data = J.JobOutput "line 2\n" J.Stdout, J._jobType = J.Server }
> ```
> 
> I get this interleaving:
> 
> ```shell
> Server: line 1a
> Db: oneline 1b
> Server: line 2
> ```
> 
> I am guessing this second output line is not a case we want, or are we explicitly not protecting against that? Anyways, great work, I will resume tomorrow after the brain break!

You are right, this is problematic case! I on purpose gave up on handling it for now, but now that you write about it, I realized we did have that solved before, since we had information about the previous message. We could add that information back to the state of the PrefixedWriter and again handle it.
The thing is, we can't really defend ourselves completely -> there can be output from non-job, and we will have no idea about it, and we will get this issue, we can't avoid it.

Good thing though is, that in practice, messages usually don't came broken into pieces like that, mostly they come with newline at the end. Even if they do come broken in the middle, which is pretty / super rare from what I saw, they will then very likely come very quickly after each other, again making the chance of this happening very low. So I would say in practice, this is not likely to happen at all.

I think it still might be worth tracking the last message and defending against this use case, I will add that.

waspc/src/Wasp/Generator/Job/IO/PrefixedWriter.hs
Martinsos: I agree this part is a bit complex, I tried simplifying it couple of times but never got to perfect solution. This looks interesting I will try using it, maybe doing some hybrid!

waspc/src/Wasp/Generator/Job/IO/PrefixedWriter.hs
shayneczyzewski: Ah, yes, silly Prisma-like executables. Forgot they do carriage return stuff and some other funky ANSI symbols. Ok, makes sense! Yes, let's punt on Windows for now! :D

shayneczyzewski:
> > Hey @Martinsos - this is super cool! It made my head hurt, so I am taking a pause for now and will re-review tomorrow with fresh eyes/brain :D Dang Monad Transformers and all that stuff haha
> > One quick question, is there a clean way to write tests for this? I just did some playing in the repl, and may have found an undesired case, but I wonder if we can do some sort of fuzzing/prop testing (or even explicit unit tests) around this? Maybe not worth it?
> > Anyways, when I try the following:
> > ```haskell
> > :set +m
> > 
> > import Wasp.Generator.Job.IO.PrefixedWriter (PrefixedWriter, printPrefixedJobMessage, runPrefixedWriter)
> > import qualified Wasp.Generator.Job as J
> > 
> > runPrefixedWriter $ do
> >   printPrefixedJobMessage $ J.JobMessage { J._data = J.JobOutput "line 1a" J.Stdout, J._jobType = J.Server }
> >   printPrefixedJobMessage $ J.JobMessage { J._data = J.JobOutput "one\n" J.Stdout, J._jobType = J.Db }
> >   printPrefixedJobMessage $ J.JobMessage { J._data = J.JobOutput "line 1b\n" J.Stdout, J._jobType = J.Server }
> >   printPrefixedJobMessage $ J.JobMessage { J._data = J.JobOutput "line 2\n" J.Stdout, J._jobType = J.Server }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I get this interleaving:
> > ```shell
> > Server: line 1a
> > Db: oneline 1b
> > Server: line 2
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I am guessing this second output line is not a case we want, or are we explicitly not protecting against that? Anyways, great work, I will resume tomorrow after the brain break!
> 
> You are right, this is problematic case! I on purpose gave up on handling it for now, but now that you write about it, I realized we did have that solved before, since we had information about the previous message. We could add that information back to the state of the PrefixedWriter and again handle it. The thing is, we can't really defend ourselves completely -> there can be output from non-job, and we will have no idea about it, and we will get this issue, we can't avoid it.
> 
> Good thing though is, that in practice, messages usually don't came broken into pieces like that, mostly they come with newline at the end. Even if they do come broken in the middle, which is pretty / super rare from what I saw, they will then very likely come very quickly after each other, again making the chance of this happening very low. So I would say in practice, this is not likely to happen at all.
> 
> I think it still might be worth tracking the last message and defending against this use case, I will add that.

I think you are right, it feels like a very unlikely scenario. I don't think we should invest much time on it now (possibly ever), but whether we want to create an issue or not is your call. I'm good either way, just wanted to note it and confirm! 👍🏻 This is really looking nice!

waspc/src/Wasp/Generator/Job/IO/PrefixedWriter.hs
Martinsos: Hey `\r` is not that weird, we are also going to use it for `npm install` progress indicator probably :D!
Btw I am going to leave a TODO here to think more about Windows in the future.

waspc/src/Wasp/Generator/Job/IO/PrefixedWriter.hs
Martinsos: I went with something in between and ended up with this:
```hs
applyPendingNewline ::
  JobsWithPendingNewline -> (J.JobType, T.Text) -> (JobsWithPendingNewline, T.Text)
applyPendingNewline jobsWithPendingNewline (jobType, content) = (jobsWithPendingNewline', content')
  where
    content' = addPendingNewlineToStartIfAny $ removeTrailingNewlineIfAny content
      where
        removeTrailingNewlineIfAny =
          if contentEndsWithNewline then T.init else id
        addPendingNewlineToStartIfAny =
          if jobType `S.member` jobsWithPendingNewline then ("\n" <>) else id
    jobsWithPendingNewline' = updateOp jobType jobsWithPendingNewline
      where
        updateOp = if contentEndsWithNewline then S.insert else S.delete
    contentEndsWithNewline = "\n" `T.isSuffixOf` content
```

What does it look like?

Martinsos:
> > > Hey @Martinsos - this is super cool! It made my head hurt, so I am taking a pause for now and will re-review tomorrow with fresh eyes/brain :D Dang Monad Transformers and all that stuff haha
> > > One quick question, is there a clean way to write tests for this? I just did some playing in the repl, and may have found an undesired case, but I wonder if we can do some sort of fuzzing/prop testing (or even explicit unit tests) around this? Maybe not worth it?
> > > Anyways, when I try the following:
> > > ```haskell
> > > :set +m
> > > 
> > > import Wasp.Generator.Job.IO.PrefixedWriter (PrefixedWriter, printPrefixedJobMessage, runPrefixedWriter)
> > > import qualified Wasp.Generator.Job as J
> > > 
> > > runPrefixedWriter $ do
> > >   printPrefixedJobMessage $ J.JobMessage { J._data = J.JobOutput "line 1a" J.Stdout, J._jobType = J.Server }
> > >   printPrefixedJobMessage $ J.JobMessage { J._data = J.JobOutput "one\n" J.Stdout, J._jobType = J.Db }
> > >   printPrefixedJobMessage $ J.JobMessage { J._data = J.JobOutput "line 1b\n" J.Stdout, J._jobType = J.Server }
> > >   printPrefixedJobMessage $ J.JobMessage { J._data = J.JobOutput "line 2\n" J.Stdout, J._jobType = J.Server }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I get this interleaving:
> > > ```shell
> > > Server: line 1a
> > > Db: oneline 1b
> > > Server: line 2
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I am guessing this second output line is not a case we want, or are we explicitly not protecting against that? Anyways, great work, I will resume tomorrow after the brain break!
> > 
> > 
> > You are right, this is problematic case! I on purpose gave up on handling it for now, but now that you write about it, I realized we did have that solved before, since we had information about the previous message. We could add that information back to the state of the PrefixedWriter and again handle it. The thing is, we can't really defend ourselves completely -> there can be output from non-job, and we will have no idea about it, and we will get this issue, we can't avoid it.
> > Good thing though is, that in practice, messages usually don't came broken into pieces like that, mostly they come with newline at the end. Even if they do come broken in the middle, which is pretty / super rare from what I saw, they will then very likely come very quickly after each other, again making the chance of this happening very low. So I would say in practice, this is not likely to happen at all.
> > I think it still might be worth tracking the last message and defending against this use case, I will add that.
> 
> I think you are right, it feels like a very unlikely scenario. I don't think we should invest much time on it now (possibly ever), but whether we want to create an issue or not is your call. I'm good either way, just wanted to note it and confirm! 👍🏻 This is really looking nice!

@shayneczyzewski I also added this now to the prefixing logic, and while doing it, I also found another case with a missing newline sometimes -> anyway, this should be covered well now! I think it is good we added it, I feel the output is now more robust. It was also easy to add additional tracking of last message, which was cool.

While at it I also improved the looks of output a bit: gave nice color to "DB" and colored "stderr" in yellow to be more visible.

Here is a screenshot of what it look like now:
image

Martinsos:
Well I just realized from the screenshot above that I completely forgot that some messages can come from stderr while some can come from stdout! And those we also don't want to mix!

I now modified code to also take that into account. So now it doesn't care just about job type, but about something called "output".

Uff, I think this should be it! Maybe take a quick look at the `Output` type @shayneczyzewski and tell me if that seems sound.

waspc/src/Wasp/Generator/Job/IO/PrefixedWriter.hs
shayneczyzewski: Haha, yeah it's true `\r` is fine for interactive outputs, I was just joking since I spaced on it. :D TODO for Windows makes sense. 👍🏻

waspc/src/Wasp/Generator/Job/IO/PrefixedWriter.hs
shayneczyzewski: Looks good to me!

shayneczyzewski (approved):
Hey @Martinsos nice updates! Good catch on tracking `stdout` in the decision-making for prefixes. I think adding that bool to `Output` makes sense.

Glad to see tracking the last job message was made easier with `PrefixedWriterState` as well.

`printPrefixedJobMessage` does still feel a bit hard to follow to me sometimes, but I think you minimized as much complexity as you can and the detailed comment above helps readers as well. It is just a hard problem to solve, and the helpers do aid in understanding what is going on, so I think much of it is essential complexity.

I did a quick test and what is here seems to function nicely. 👍🏻 The only thing I would consider is possibly adding a space between the `JobType` and "(stderr)", so you would go from "Db(stderr):" to "Db (stderr):", but that may just be my visual preference.

Martinsos:
@shayneczyzewski you are right that printPrefixedJobMessage is still a bit too complex.
I now refactored it a bit more, take a look, I believe it is a bit more readable.
I also changed the name to printJobMessagePrefixed, since that is more correct -> we are not printing already prefixed job message, we prefix it ourselves and then print it.

As for space between Server and (stderr) -> there was actually space there before, but I removed it because it felt a bit like it was part of the input to the right. At least I subjectively thought so :D. So maybe let's leave it like this for now and we can change in the future if we figure out it is not the best?

shayneczyzewski:
> @shayneczyzewski you are right that printPrefixedJobMessage is still a bit too complex. I now refactored it a bit more, take a look, I believe it is a bit more readable. I also changed the name to printJobMessagePrefixed, since that is more correct -> we are not printing already prefixed job message, we prefix it ourselves and then print it.
> 
> I am sure it could be improved further, but it feels like I reached the threshold where refactoring pays off, I feel going further the ratio of invested time and what I got would not be the best.
> 
> As for space between Server and (stderr) -> there was actually space there before, but I removed it because it felt a bit like it was part of the input to the right. At least I subjectively thought so :D. So maybe let's leave it like this for now and we can change in the future if we figure out it is not the best?

Looking great! I like the refactor to simplify things a bit more. I think this will work just fine. 👍🏻 

No sweat on space, we can leave it as-is and see what others think. All good by me. 🚀

Issue #710, opened 2022-08-31
Enable JS/TS IDE support for js/ts files in ext-src in Wasp project


Martinsos:
Check https://github.com/wasp-lang/wasp/issues/604 for main discussion. We should implement (2) + (5) approach.

Martinsos:
Done

Issue #688, opened 2022-08-11
Wasp lang dev tooling


craigmc08:
This issue organizes all issues and features relevant to dev tooling for the wasp language.

### Wasp Analyzer

- [ ] #687
- [ ] #686
- [ ] Update typechecker so it can run on wasp source code containing syntax errors in it
- [ ] Update typechecker to be able to report multiple type errors

### waspls

- [ ] Merge waspls cabal project into the waspc cabal project
- [ ] Bundle waspls with waspc installation
- [ ] Implement semantic highlighting in the language server
- [ ] Improve autocompletion results
    - Add enum values to the results
    - Use expected type for the cursor location to filter results
- [ ] Implement more LSP features
    - Potential priority list: go-to-definition for wasp objects, go-to-definition for external imports, rename wasp objects

### JS/TS Intellisense

For the best effort/reward ratio here, we should focus on getting TS support for wasp projects. This is a highly requested feature and will also pave the way for getting TS intellisense to work since the TS compiler will understand how wasp projects work a little bit by then.

### VSCode extension

- [ ] Code formatting for wasp files
- [ ] Find a way to keep current syntax highlighting in sync with wasp language
    - With no semantic highlighting (yet), we have a textmate grammar defined in the vscode extension. This manually lists out all object types and enum values, which means we need to keep those lists in sync with wasp type definitions. Maybe we can find a way to keep this in sync automatically.

Martinsos:
Regarding JS/TS intelissense in user created files in ext-src/ -> we explored this a lot in https://github.com/wasp-lang/wasp/issues/604 . We kind of concluded that the best solution for now might be to "explain" to editor via jsconfig.json and by possibly adjusting position of some of our genreated files so that it all "looks good" to the editor and that JS/TS IDE can mostly work on stuff then.

Martinsos:
It would make sense to also include AppSpec validation errors into LSP. Right now those are not made a part of it. Only analysis is done, but app spec validation is not. We would want to do it and show those also.

Martinsos:
Also, it would be great if Wasp compiler would inform me about wrong ext/ import paths! It could check during compile time if those exist.

Martinsos:
Another thing: It would be great if we could get Wasp LS to the level where it offers autocompletion for dictionary fields! Right now it doesn't do that.

Martinsos:
Not any more relevant since we moved to eDSL in TypeScript!

Issue #604, opened 2022-05-24
RFC: Wasp IDE Integration


craigmc08:
# Motivation

One of the main point points in Wasp currently is the lack of editor integration while developing a Wasp project. This manifests in two separate ways:

1. No editor language features for `.wasp` files

2. Code generation breaks the existing JS IntelliSense in editors

While there does exist a Wasp extension for VSCode (https://github.com/wasp-lang/vscode-wasp), it is out of date with the current DSL version and only adds syntax highlighting. Thus, the existing solution only gives a partial solution to (1) and does nothing for (2).

Wasp is a tool that exists to make the process of developing full-stack web apps easier, and this is one place where Wasp makes it *harder* to write code, so these issues must be resolved.

In this RFC, we first outline what a Wasp extension would need to do generally. Then, we give more specifics on how an implementation of an extension for VSCode would look.

# Requirements

1. Wasp DSL language features
  1.1. Syntax highlighting for `.wasp` files
  1.2. In editor parse-/type-error reporting
  1.3. Automatic formatting
  1.4. Autocompletion for variable names
  1.5. Autocompletion for dictionary keys
  1.6. Autocompletion for JS import statements

2. JavaScript IntelliSense. Support all existing editor language features for JavaScript. Working in a Wasp project should not be a downgrade from developing outside of one in terms of JS language support.

Requirements (1) and (2) are completely disjoint and can be implemented independently of each other.

Additionally, both of these requirements can be developed incrementally: even just 1 feature from the list is an upgrade from the existing tooling. The sub-requirements of (1) are ordered in decreasing importance. For (2), the most important language features are:

- Autocompletion/auto-import for generated code
- `Go to definition` for queries and actions

# Implementation

The implementation section is split into two components. One for Wasp language features, and one for JavaScript IntelliSense.

## Wasp Language Features

To fulfill requirement (1), we would implement a Wasp language server conforming to the Language Server Protocol (https://microsoft.github.io/language-server-protocol/). Through this, we can support all sub-requirements.

The language server would be written in Haskell, using the [lsp](https://github.com/haskell/lsp) package and would rely on the existing Analyzer in [waspc](https://github.com/wasp-lang/wasp/tree/main/waspc) for parsing and type checking.

Along with the language server, a small extension would need to be written for each editor to integrate the language. For VSCode, there is [vscode-languageclient](https://www.npmjs.com/package/vscode-languageclient), which makes it extremely quick to connect a language server to VSCode.

### Alternatives

[...]

PR #795, opened 2022-10-31
[IDE support] Peek and go to definition functionalities via jsconfig

sodic:
The implementation for Peek definition, go to definition, and import resolution in external JavaScript source files.

Takes steps towards fixing #710, and (more generally) #604 and #688.


Martinsos:
Hm this PR @sodic has nothing but e2e updates! Is that a mistake?

sodic:
It's a mistake, but not on my part :smile: 

It has changes in the `waspc/data` directory, but there are no code changes, that's correct.

Martinsos:
Interesting, when I checked there were no changes under wasp/data. Maybe I followed the link that showed only the changes from the commit that updated e2e tests -> might be that.

waspc/data/Cli/templates/new/src/client/jsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
```
Martinsos, line 1: Is there some stuff here that we don't want the users to touch?
Should we document it somehow, if so? Maybe directly in this file? I know JSON by itself doesn't support comments but there are tricks to add them.

waspc/data/Cli/templates/new/src/shared/jsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
+    "allowJs": true,
+    "target": "es2018",
+    "baseUrl": ".",
+    "paths": {
+      "*": [
+        "*",
+        "../../.wasp/out/server/node_modules/*",
```
Martinsos, line 10: Ok, we decided to go with server.
Do we have any idea how to make this "right" in the future? Well, as "right" as possible?

Martinsos (approved):
LGTM! Approved but also asked couple of questions.

waspc/data/Cli/templates/new/src/shared/jsconfig.json
sodic: I can't think of a solution that doesn't involve separating server, client, and shared npm dependencies.

waspc/data/Cli/templates/new/src/client/jsconfig.json
sodic: Yes, definitely, I'll put it in.

PR #799, opened 2022-11-04
If wrong wasp version, message gives instructions how to install correct one.

Martinsos:
This follows https://github.com/wasp-lang/get-wasp/pull/3 .

### DIFF
--- a/waspc/src/Wasp/AppSpec/Valid.hs
+++ b/waspc/src/Wasp/AppSpec/Valid.hs
@@ -30,7 +30,10 @@ import qualified Wasp.SemanticVersion as SV
 import qualified Wasp.Version as WV
 
 data ValidationError = GenericValidationError String
-  deriving (Show, Eq)
+  deriving (Eq)
+
+instance Show ValidationError where
+  show (GenericValidationError e) = e
 
 validateAppSpec :: AppSpec -> [ValidationError]
 validateAppSpec spec =
@@ -84,10 +87,14 @@ validateWaspVersion specWaspVersionStr = eitherUnitToErrorList $ do
     incompatibleVersionError :: SV.Version -> SV.Range -> ValidationError
     incompatibleVersionError actualVersion expectedVersionRange =
       GenericValidationError $
-        unwords
+        unlines
           [ "Your Wasp version does not match the app's requirements.",
             "You are running Wasp " ++ show actualVersion ++ ".",
-            "This app requires Wasp " ++ show expectedVersionRange ++ "."
+            "This app requires Wasp " ++ show expectedVersionRange ++ ".",
+            "To install specific version of Wasp, do:",
+            "  curl -sSL https://get.wasp-lang.dev/installer.sh | sh -s -- -v x.y.z",
+            "where x.y.z is your desired version.",
+            "Check https://github.com/wasp-lang/wasp/releases for the list of valid versions."
           ]
 
     eitherUnitToErrorList :: Either e () -> [e]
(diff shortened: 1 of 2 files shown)


Martinsos:
I tested it @shayneczyzewski so no need to spend time on it yourself since it is a bit tricky to set it up.

shayneczyzewski:
> I tested it @shayneczyzewski so no need to spend time on it yourself since it is a bit tricky to set it up.

It's all good, I just bumped the wasp version in the example app higher than what our project is. 👍🏻 Here is a screenshot for reference:
Screenshot 2022-11-04 092154


Do we have other instances where errors produce two error lines? We may have, was just something I noticed.

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -84,10 +87,14 @@ validateWaspVersion specWaspVersionStr = eitherUnitToErrorList $ do
...
     incompatibleVersionError actualVersion expectedVersionRange =
       GenericValidationError $
-        unwords
+        unlines
           [ "Your Wasp version does not match the app's requirements.",
             "You are running Wasp " ++ show actualVersion ++ ".",
-            "This app requires Wasp " ++ show expectedVersionRange ++ "."
+            "This app requires Wasp " ++ show expectedVersionRange ++ ".",
+            "To install specific version of Wasp, do:",
+            "  curl -sSL https://get.wasp-lang.dev/installer.sh | sh -s -- -v x.y.z",
```
shayneczyzewski, line 95: Since our installer expects 4 parts, should we show them an extra zero or place somewhere?

Also, do we want to just construct the minimal version for them in the example? So in my case of `This app requires Wasp ^0.7.0.` we could say: `For example: curl -sSL https://get.wasp-lang.dev/installer.sh | sh -s -- -v 0.7.0.0` I think once our semantic version matches the installer (e.g. 3 parts) it will be easier, but now it may be confusing for them. Just a thought.

Martinsos: Ok so that is something that happened in another PR from a contributor couple of days ago -> we switched to x.y.z! That is already in main actually. So while the current version is 0.6.0.0, the next one will be 0.7.0, as per semver rules for alpha releases, where basically there is no "patch" field, it is just 0.major.minor . Ok actually it is 0.minor.patch, but minor is treated as major and patch is treated as minor.
So that should cover most of your question, but the part that remains is the idea of offering them immediately which version to install. That is a bit tricky because I have a semver range here, so I would need to pick a version from it, but I am not sure which version from that range exists or not. We could maybe assume it is the start of the range, but does not have to be true in theory.

We will anyway be upgrading the whole system of upgrading and picking versions in the future I think, possibly make it automatic based on the project and similar, so that is why I  wasn't super worried about optimizing this completely.
I will be writing more about that in another PR soon!

Martinsos:
> > I tested it @shayneczyzewski so no need to spend time on it yourself since it is a bit tricky to set it up.
> 
> It's all good, I just bumped the wasp version in the example app higher than what our project is. 👍🏻 Here is a screenshot for reference: Screenshot 2022-11-04 092154
> 
> Do we have other instances where errors produce two error lines? We may have, was just something I noticed.

Oh ok brainfart on my side, of course it is easy to test! Huh I made it complex for myself :D.

What do you mean with error messages in two lines?

waspc/src/Wasp/AppSpec/Valid.hs
shayneczyzewski: Ok thanks, the 3 fields make sense. Yeah, the picking a version for them is hard if we want to pick the latest, or some specific version in the range. But I think we could just pick the lowest semver version for them and assume it will work since it was in the wasp file, right? But this is not a big deal, let's just punt on all that for now. 👍🏻

shayneczyzewski (approved):
We chatted and agreed to look into the double-error printing later. Let's ship it! 🚀

PR #800, opened 2022-11-04
Don't require golden dirs to exist when running test:e2e:accept-all

sodic:
This change prevents the following:
image


Martinsos (approved):
Nice one :D! Just one letter but I actually also stumbled on this couple of days ago!

PR #827, opened 2022-11-24
Server-side TypeScript support

sodic:


waspc/data/Cli/templates/new/src/server/tsconfig.json
```diff
@@ -22,6 +24,7 @@
         // definitions.
         "node_modules/@types/*"
       ]
-    }
+    },
+    "typeRoots": ["../../.wasp/out/server/node_modules/@types"]
```
sodic, line 28: We need this for global stuff (such as `process`). The client might need it too, not sure. I'll add it to its `tsconfig` just in case.

waspc/data/Generator/templates/server/package.json
```diff
@@ -3,9 +3,11 @@
   "name": "server",
   "version": "0.0.0",
   "private": true,
-  "type": "module",
   "scripts": {
-    "start": "nodemon -r dotenv/config ./src/server.js",
+    "build": "npx tsc",
+    "start": "NODE_PATH=dist node dist/server.js",
+    "build-and-start": "npm run build && npm run start",
+    "start-nodemon": "nodemon -r dotenv/config --exec 'npm run build-and-start || exit 1'",
```
sodic, line 10: Watches JavaScript and TypeScript files, and recompiles and runs the code on each change.

The `|| exit 1` part tells nodemon to continue listening for changes even if the server dies with a fatal error.

waspc/data/Generator/templates/server/src/jobs/core/pgBoss/pgBossJob.js
```diff
@@ -55,7 +55,7 @@ class PgBossJob extends Job {
    * @param {object} jobArgs - The job arguments supplied by the user for their perform callback.
    * @param {object} jobOptions - pg-boss specific options for `boss.send()`, which can override their defaultJobOptions.
    */
-  async submit(jobArgs, jobOptions) {
+  async submit(jobArgs, jobOptions = {}) {
```
sodic, line 58: @shayneczyzewski We're using this as an optional parameter and TS complains if we don't specify a default value.

waspc/data/Generator/templates/server/package.json
Martinsos: Cool! Could you add a comment for that in the package.json itself? I think it is a good rule that if you have to explain code in the comment in the PR, then that comment better be in the code itself :D.

waspc/src/Wasp/Generator/ServerGenerator.hs
```diff
@@ -124,18 +126,19 @@ npmDepsForWasp spec =
...
           ++ depsRequiredByPassport spec
           ++ depsRequiredByJobs spec,
       N.waspDevDependencies =
         AS.Dependency.fromList
           [ ("nodemon", "^2.0.19"),
             ("standard", "^17.0.0"),
-            ("prisma", show prismaVersion)
+            ("prisma", show prismaVersion),
+            -- TODO: types for node
+            -- TODO: make optional
```
Martinsos, line 139: Maybe you could explain this just a bit more? Smth like `TODO: Make these conditional once we allow users to choose if they want to use TS or not.`.

waspc/data/Generator/templates/server/package.json
Martinsos, line 10: So `start` now runs server, without watching.
`start-nodemon` starts it while watching.

Usually people expect `start` to already do watching.

I would suggest following naming:
`start` as it is, `watch` instead of `start-nodemon`.

`run` instead of `start`, `start` instead of `start nodemon`.
Although I am not sure if `run` is allowed hm.

waspc/data/Cli/templates/new/src/server/tsconfig.json
Martinsos: Should we add comment for this in the code also, since you felt the need to comment here?
Imagine in the future one of us looking at this line and trying to figure out its purpose -> is it going to be hard to figure it out? If so, let' add a comment.

Martinsos (approved):
Nice job @sodic, you managed to get it working wohooo :D!
I left some minor comments but approved.
The main thing is commenting on stuff that later we will wonder why we did it this way -> better to comment now on it, so later we don't have to reverse engineer it and guess what is something for. I would advise rather having too much comment in this case than too little.
I think this is especially important for configuration. In the code itself, usually we can figure out what something does by looking at its environment and use cases. But with configs, usually effect is pretty far from the config itself and it can be really hard figuring out what was the intention and what are the effects it causes, so better to over-comment it.

waspc/data/Cli/templates/new/src/server/tsconfig.json
sodic: Sure, will do.

waspc/data/Generator/templates/server/package.json
sodic: I chose the first one.

PR #831, opened 2022-11-25
TypeScript fixes before Beta

sodic:
Implements the following
- Fixes the `Cannot write file because it would overwrite input file` error in _tsconfig.json_ files
- Top-level `await` support on the server
- Correct libraries and targets for both the client and the server
- _tsconfig.json_ files that extend the semi-official bases: https://github.com/tsconfig/bases
- `MainPage.tsx` became `MainPage.jsx`
- Correct types for `useAction` and `useQuery`
- Better explanations in _tsconfig.json_ files used for IDE support
- Improved type safety in user files


waspc/data/Generator/templates/react-app/tsconfig.json
```diff
@@ -1,23 +1,8 @@
...
-    "forceConsistentCasingInFileNames": true,
-    "noFallthroughCasesInSwitch": true,
-    "module": "esnext",
-    "moduleResolution": "node",
-    "resolveJsonModule": true,
-    "isolatedModules": true,
-    "noEmit": true,
-    "jsx": "react-jsx"
+    // Overriding this until we implement more complete TypeScript support.
+    "strict": false,
```
sodic, line 5: The other settings are implemented in `create-react-app/tsconfig`.

Martinsos (approved):
Awesome!!!

waspc/data/Generator/templates/server/package.json
```diff
@@ -3,11 +3,12 @@
   "name": "server",
   "version": "0.0.0",
   "private": true,
+  "type": "module",
   "scripts": {
     "build": "npx tsc",
-    "start": "NODE_PATH=dist node dist/server.js",
+    "start": "NODE_PATH=dist node -r dotenv/config dist/server.js",
```
sodic, line 9: This is necessary to ensure that the server reads the `.env` file.

shayneczyzewski (approved):

PR #862, opened 2022-11-30
Add support for `--create-only` migrations

### LINKED ISSUE #798: Add support for creating an empty migration via `--create-only`
**Is your feature request related to a problem? Please describe.**
There are times when no sensible migration can automatically be created. In those instances, users are told by Prisma they must run `prisma migrate dev --create-only`.

**Describe the solution you'd like**
We should have support for this as `wasp db migrate-dev --create-only`.

shayneczyzewski:
# Description

Adds support for `prisma db migrate --create-only` migrations. This is necessary when Prisma cannot automatically create a migration for you based on your schema changes.

This required a few additional changes:
- The CLI invocation for `db migrate-dev` was extended to use options for `--name foo` and `--create-only`
- The copy file draft clears out the contents of the destination directory (not strictly related, but discovered here that it should have behaved that way)
- Changed the way we handle the last concurrence checksum file indicating the user is in sync. Now we must also compare the migrations dir, not just the schema.

Fixes #798

## Type of change

- [ ] Code cleanup
 - change CLI for invoking with migration name from `wasp db migrate-dev foo` to `wasp db migrate-dev --name foo`


waspc/src/Wasp/Generator/DbGenerator/Common.hs
```diff
@@ -60,3 +66,42 @@ dbSchemaChecksumOnLastGenerateFileInDbRootDir = [relfile|schema.prisma.wasp-gene
...
+
+parseMigrateArgs :: Maybe [String] -> MigrateArgs
+parseMigrateArgs Nothing = emptyMigrateArgs
+parseMigrateArgs (Just migrateArgs) = do
+  go migrateArgs emptyMigrateArgs
+  where
+    go :: [String] -> MigrateArgs -> MigrateArgs
+    go ("--create-only" : rest) mArgs = go rest mArgs {_isCreateOnlyMigration = True}
+    go ("--name" : name : rest) mArgs = go rest mArgs {_migrationName = Just name}
+    go _ mArgs = mArgs
```
shayneczyzewski, line 94: Right now I just ignore and bail for any unexpected arg. Should this `error` instead?

waspc/src/Wasp/Generator/DbGenerator/Operations.hs
```diff
@@ -43,26 +49,35 @@ printJobMsgsUntilExitReceived chan = do
...
   return res
   where
     dbMigrationsDirInProjectRootDir = dbRootDirInProjectRootDir SP.</> dbMigrationsDirInDbRootDir
+    applyLastDbConcurrenceChecksumAction =
+      case lastDbConcurrenceChecksumAction of
+        WriteOnLastDbConcurrenceChecksum ->
+          writeDbSchemaChecksumToFile genProjectRootDirAbs (SP.castFile dbSchemaChecksumOnLastDbConcurrenceFileProjectRootDir)
+        RemoveOnLastDbConcurrenceChecksum ->
+          removeDbSchemaChecksumFile genProjectRootDirAbs (SP.castFile dbSchemaChecksumOnLastDbConcurrenceFileProjectRootDir)
+        IgnoreOnLastDbConcurrenceChecksum -> return ()
```
shayneczyzewski, line 81: This section got a little more complicated, but hopefully the new type helps. Basically we have some scenarios where we want to write the checksum indicating we are fully in sync with DB (like after a migrate), or remove it when we may have invalidated it (like when we create an empty migration).

When we create an empty migration, we cannot rely on the prisma schema alone to determine if we should migrate, as they can write anything in the new file. That requires our new migrate dir check.

waspc/src/Wasp/Generator/FileDraft/CopyDirFileDraft.hs
```diff
@@ -35,6 +42,8 @@ data CopyDirFileDraft = CopyDirFileDraft
 instance Writeable CopyDirFileDraft where
   write projectRootAbsPath draft = do
     srcDirExists <- doesDirectoryExist $ SP.fromAbsDir srcPathAbsDir
+    dstDirExists <- doesDirectoryExist $ SP.fromAbsDir dstPathAbsDir
+    when dstDirExists $ removeDirectoryRecursive dstPathAbsDir
```
shayneczyzewski, line 46: I added this for two reason:
1) It now matches the semantics of copy file, where if it exists it overwrites it and thus they are equal. Now, when we have a dir FD, we know they will be equal (i.e. the destination cannot have "extra" stuff).
2) When a user wants to remove a migration from their wasp project dir, we want that to be removed from the generated project. This may happen when they do a `--create-only` but then don't migrate and just remove the dir. Without this, it would stay in their generated project and the next migrate could apply it.

waspc/cli/exe/Main.hs
```diff
@@ -127,8 +127,8 @@ printVersion = do
 -- TODO(matija): maybe extract to a separate module, e.g. DbCli.hs?
 dbCli :: [String] -> IO ()
 dbCli args = case args of
-  ["migrate-dev", migrationName] -> runDbCommand $ Command.Db.Migrate.migrateDev (Just migrationName)
   ["migrate-dev"] -> runDbCommand $ Command.Db.Migrate.migrateDev Nothing
+  "migrate-dev" : migrateArgs -> runDbCommand $ Command.Db.Migrate.migrateDev $ Just migrateArgs
```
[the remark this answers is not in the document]
shayneczyzewski: In the example above, they would be passed in like: `["--name", "FOO", "--create-only"]`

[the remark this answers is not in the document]
Martinsos: Ha, this is getting a bit complex now!

Maybe we should just have it accept [String] then, and it can be empty if there are no arguments, instead of Maybe [String]? So [] would play role of Nothing.

Ideally, we would parse the strings and pass some nicely shaped params. That is a bit of work right now though, so I do agree with postponing it. We should probably switch the whole CLI to that applicative library and that would also help with this.

waspc/src/Wasp/Generator/DbGenerator/Common.hs
```diff
@@ -60,3 +66,42 @@ dbSchemaChecksumOnLastGenerateFileInDbRootDir = [relfile|schema.prisma.wasp-gene
...
 dbSchemaChecksumOnLastGenerateFileProjectRootDir = dbRootDirInProjectRootDir </> dbSchemaChecksumOnLastGenerateFileInDbRootDir
+
+data MigrateArgs = MigrateArgs
+  { _migrationName :: Maybe String,
+    _isCreateOnlyMigration :: Bool
+  }
+  deriving (Show, Eq)
+
+emptyMigrateArgs :: MigrateArgs
+emptyMigrateArgs = MigrateArgs {_migrationName = Nothing, _isCreateOnlyMigration = False}
```
Martinsos, line 77: rename to defaultMigrateArgs?

waspc/src/Wasp/Generator/DbGenerator/Common.hs
```diff
@@ -60,3 +66,42 @@ dbSchemaChecksumOnLastGenerateFileInDbRootDir = [relfile|schema.prisma.wasp-gene
...
+
+asArgs :: MigrateArgs -> [String]
```
Martinsos, line 79: AHa, so this transforms them into actual `prisma dev migrate` args?

waspc/src/Wasp/Generator/DbGenerator/Common.hs
```diff
@@ -60,3 +66,42 @@ dbSchemaChecksumOnLastGenerateFileInDbRootDir = [relfile|schema.prisma.wasp-gene
...
+asArgs migrateArgs = do
+  concat . concat $
+    [ [["--create-only"] | _isCreateOnlyMigration migrateArgs],
+      [["--name", name] | Just name <- [_migrationName migrateArgs]]
+    ]
+
+parseMigrateArgs :: Maybe [String] -> MigrateArgs
```
Martinsos, line 86: I would suggest moving this logic into CLI, where the command is parsed.
I would keep `asArgs` here, and I would keep `MigrateArgs` here, but I would move this logic that translates from CLI args to MigrateArgs.
asArgs could maybe go to Jobs/Db though, not 100% sure -> wherever is closer to Prisma.

waspc/src/Wasp/Generator/DbGenerator/Common.hs
```diff
@@ -60,3 +66,42 @@ dbSchemaChecksumOnLastGenerateFileInDbRootDir = [relfile|schema.prisma.wasp-gene
...
+parseMigrateArgs Nothing = emptyMigrateArgs
+parseMigrateArgs (Just migrateArgs) = do
+  go migrateArgs emptyMigrateArgs
+  where
+    go :: [String] -> MigrateArgs -> MigrateArgs
+    go ("--create-only" : rest) mArgs = go rest mArgs {_isCreateOnlyMigration = True}
+    go ("--name" : name : rest) mArgs = go rest mArgs {_migrationName = Just name}
```
Martinsos, line 93:
```suggestion
    go ("--create-only" : rest) mArgs = go rest $ mArgs {_isCreateOnlyMigration = True}
    go ("--name" : name : rest) mArgs = go rest $ mArgs {_migrationName = Just name}
```

I know it is redundant by linter, but for me it greatly increases readibilty, it took me some time to figure out that go takes 2 args here and not 3.

waspc/src/Wasp/Generator/DbGenerator/Common.hs
Martinsos: I think error would be better! That way, if they make minor typo, they will know they did a typo, instead of thinking args was applied.

waspc/src/Wasp/Generator/DbGenerator/Common.hs
```diff
@@ -60,3 +66,42 @@ dbSchemaChecksumOnLastGenerateFileInDbRootDir = [relfile|schema.prisma.wasp-gene
...
+      [["--name", name] | Just name <- [_migrationName migrateArgs]]
+    ]
+
+parseMigrateArgs :: Maybe [String] -> MigrateArgs
+parseMigrateArgs Nothing = emptyMigrateArgs
+parseMigrateArgs (Just migrateArgs) = do
+  go migrateArgs emptyMigrateArgs
+  where
+    go :: [String] -> MigrateArgs -> MigrateArgs
+    go ("--create-only" : rest) mArgs = go rest mArgs {_isCreateOnlyMigration = True}
```
Martinsos, line 92: Niiiceeee, proper fun recursion :D! Yeah this what we are about right :D!?

waspc/src/Wasp/Generator/DbGenerator/Common.hs
```diff
@@ -60,3 +66,42 @@ dbSchemaChecksumOnLastGenerateFileInDbRootDir = [relfile|schema.prisma.wasp-gene
...
+    go ("--name" : name : rest) mArgs = go rest mArgs {_migrationName = Just name}
+    go _ mArgs = mArgs
+
+data OnLastDbConcurrenceChecksumAction
```
Martinsos, line 96: What exactly does this stand for? So on it tells me what was done on the last DB concurrent checksum, or what should be done?

Martinsos: Ah got it know, it should be read as "Write OnLastDbConcurrenceChecksum".
Hmmmmmm.
I don't have much better idea. Maybe a short comment saying that this says what would be done with the checksum file.

Maybe `WriteOnLastDbConcurrentChecksumFile` would be better? So adding File at the end?

Or maybe `Write_OnLastDbConcurrentChecksumFile`?

This `On` is what confuses me, but I guess we can't get rid of it easily as it become name that we use on many places?

waspc/src/Wasp/Generator/DbGenerator/Common.hs
```diff
@@ -60,3 +66,42 @@ dbSchemaChecksumOnLastGenerateFileInDbRootDir = [relfile|schema.prisma.wasp-gene
...
+  = WriteOnLastDbConcurrenceChecksum
+  | RemoveOnLastDbConcurrenceChecksum
+  | IgnoreOnLastDbConcurrenceChecksum
+
+-- When we do a create-only migration, we need to remove the DbSchemaChecksumOnLastDbConcurrenceFile so they know to migrate.
+-- When we migrate, we need to write it to indicate the local code and DB are in sync.
```
Martinsos, line 102: In comments you are describing imlementation. Instead, would be better to define what we expect the function to do. I would go in the direction of explaining that this function, based on the args given, decides what we should do with the checksum file. Although, this should actually be visible from the type signature. So maybe we can improve the names there and avoid the comment all together!

Martinsos (commented):
@shayneczyzewski I didn't have time to check it all yet, but here is first half and I will do the second one tomorrow. But this is already somewhat standalone.

waspc/src/Wasp/Generator/DbGenerator/Common.hs
shayneczyzewski: Thanks, agree. I was leaning that way but this was the lazy "should I _realllyyy_ do it?" :D Will do!

waspc/cli/exe/Main.hs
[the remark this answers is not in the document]
shayneczyzewski: I like your idea of moving the parsing into CLI. I can then pass in the `MigrateArgs` 👍🏻 Will do thnx

waspc/src/Wasp/Generator/DbGenerator/Common.hs
shayneczyzewski: Sounds good, will do. That should simplify the interface of `Maybe [String]` as well 👍🏻

waspc/src/Wasp/Generator/DbGenerator/Common.hs
shayneczyzewski: I know, right! Haskell, baby!!! It has sadly been so long since I wrote a fun little function like this that it took me a second to get my mind into it though lol 😅

waspc/src/Wasp/Generator/DbGenerator/Common.hs
shayneczyzewski: Good call on maybe changing the name and adding comments. The trick here is it isn't only just write, it could be remove, or nothing (no-op). But I can think more on it to try and clear up the confusion. And it is complicated by the fact that what it refers to is a tricky notion (the On thing)... 🤔

waspc/src/Wasp/Generator/DbGenerator/Common.hs
shayneczyzewski: Good point, I can make it more about what we are doing at a higher level, and possibly remove it altogether if the sig can be cleaned up some to convey it! 👍🏻 It tends to be my crutch for, "this feels a bit complicated, just comment it to death", but that usually means the code itself can be improved to avoid it instead hehe

waspc/src/Wasp/Generator/DbGenerator/Common.hs
shayneczyzewski: Does this help improve things any?

```haskell
-- | This type tells us what we need to do with the DbSchemaChecksumOnLastDbConcurrenceFile.
data RefreshOnLastDbConcurrenceChecksumFile
  = WriteOnLastDbConcurrenceChecksumFile
  | RemoveOnLastDbConcurrenceChecksumFile
  | IgnoreOnLastDbConcurrenceChecksumFile
```

waspc/src/Wasp/Generator/DbGenerator/Common.hs
shayneczyzewski: The names are getting longer :D but more consistent now. I did feel the comment inside does help the reader some. Do you think this is any better?

```haskell
getOnLastDbConcurrenceChecksumFileRefreshAction :: MigrateArgs -> RefreshOnLastDbConcurrenceChecksumFile
getOnLastDbConcurrenceChecksumFileRefreshAction migrateArgs =
  -- Since a create-only migration allows users to write any SQL, we remove the file to force
  -- revalidation with the DB. If it is a regular migration, we write it since they will be in sync.
  if _isCreateOnlyMigration migrateArgs
    then RemoveOnLastDbConcurrenceChecksumFile
    else WriteOnLastDbConcurrenceChecksumFile
```

shayneczyzewski:
> @shayneczyzewski I didn't have time to check it all yet, but here is first half and I will do the second one tomorrow. But this is already somewhat standalone.

Thanks for the first round review @Martinsos, I appreciate it! I have addressed all comments (in some form). Have a great weekend and thanks again.

waspc/src/Wasp/Generator/DbGenerator/Common.hs
Martinsos: hehe sounds good :D!

waspc/src/Wasp/Generator/DbGenerator/Common.hs
Martinsos: I think it does! File makes you realize there is a noun there, I think that is enough, and comments clears it up.

waspc/src/Wasp/Generator/DbGenerator/Common.hs
Martinsos: I think this looks good!

Martinsos (approved):
@shayneczyzewski left some minor comments but all is looking good really, so approved! Nice job!!!

waspc/src/Wasp/Generator/FileDraft/CopyDirFileDraft.hs
shayneczyzewski: I don't think this will negatively affect the generation process. The only thing it should do is ensure there is nothing stale leftover, which is what we want. I don't think we have `createDirectory` alone sadly. Prob ok to just use what we have for now I think.

waspc/src/Wasp/Generator/DbGenerator/Common.hs
shayneczyzewski: Awesome, thanks!

waspc/test/Generator/MockWriteableMonad.hs
```diff
@@ -77,6 +77,9 @@ instance WriteableMonad MockWriteableMonad where
   copyDirectoryRecursive srcPath dstPath = MockWriteableMonad $ do
     modifyLogs (copyDirectoryRecursive_addCall (castDir srcPath) (castDir dstPath))
 
+  removeDirectoryRecursive dir = MockWriteableMonad $ do
```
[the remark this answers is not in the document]
shayneczyzewski: Oh yeah lol

web/docs/cli.md
```diff
@@ -16,6 +16,7 @@ COMMANDS
   GENERAL
     new <project-name>    Creates new Wasp project.
     version               Prints current version of CLI.
+    waspls                Run Wasp Language Server. Add --help to get more info.
```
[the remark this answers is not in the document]
shayneczyzewski: Thanks! Agreed, unsure without our special doc generation app :D

waspc/src/Wasp/Generator/DbGenerator/Operations.hs
```diff
@@ -116,3 +133,19 @@ doesSchemaMatchDb genProjectRootDirAbs = do
...
+areAllMigrationsAppliedToDb :: Path' Abs (Dir ProjectRootDir) -> IO (Maybe Bool)
+areAllMigrationsAppliedToDb genProjectRootDirAbs = do
+  chan <- newChan
+  (_, dbExitCode) <-
+    concurrently
+      (readJobMessagesAndPrintThemPrefixed chan)
+      (DbJobs.migrateStatus genProjectRootDirAbs chan)
+  case dbExitCode of
+    ExitSuccess -> return $ Just True
+    ExitFailure _ -> return Nothing
```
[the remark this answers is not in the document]
shayneczyzewski: Yeah this is tricky. I sorta liked it _not_ being just two states to explicitly capture there is some ambiguity in how the command returns the status. And it is similar to another command that does indicate all 3 states. My hope in the future is prisma takes my feedback and makes all their commands able to have these 3 states, so it was: a) documentation of the ambiguity, b) consistency with the one above, and c) hopefully planning ahead for the fix. For those reasons, I think it may make sense to leave as is, sound ok?

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
```diff
@@ -50,6 +50,13 @@ migrateDev projectDir maybeMigrationName = do
 
   runNodeCommandAsJob serverDir "script" scriptArgs J.Db
 
+asArgs :: MigrateArgs -> [String]
```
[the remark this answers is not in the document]
shayneczyzewski: Will update thanks!

waspc/src/Wasp/Generator/DbGenerator/Operations.hs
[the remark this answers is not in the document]
Martinsos: Sounds good!

Martinsos:
LGTM!

Issue #867, opened 2022-12-05
How to solve /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34` not found


EmmanuelTheCoder:
**Describe the bug**

There is something I observed with Wasp 0.7.1 on Windows (WSL). Before now, I used Wasp 0.6.0 and everything works fine but after installing Wasp 0.7.1 on Windows, trying to run any wasp command resulted in the following errors:

/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34` not found (required by ...wasp-lang/0.7.1/wasp-bin)

/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32` not found (required by ...wasp-lang/0.7.1/wasp-bin)

/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33` not found (required by ...wasp-lang/0.7.1/wasp-bin).

Below is photo evidence.

<img width="685" alt="wasp libc6 err" src="https://user-images.githubusercontent.com/55356373/205626173-18bf8fb8-c5d1-4c56-bc39-48bd7bc24667.png">

**To Reproduce**

1. Install Wasp 0.7.1 on Windows (WSL)
2. Try to run any wasp command and boom you have the terminal yelling at you some GLIBC error

**What does this mean and how to resolve it**

The error means that Wasp 0.7.1 was compiled against GLIBC 2.32 - 2.34 and therefore it requires any of those versions (preferably 2.34 or higher) to run. In my case, I was on glibc 2.31 which is the latest (and highest) version for Ubuntu 20.04 (Focal)

Since that was the case, trying to install a greater glibc of another Ubuntu version might kill my machine. This is to say that glibc for say version 22.4 will not work for version 20.04.  

The solution to this was that I needed to update my Ubuntu version to at least Impish (21.10) so that I can get access to glibc 2.34 to run Wasp on Windows (Impish uses glibc 2.34 - meets the glibc requirement of Wasp 0.7.1). I strongly recommend that you upgrade to Jammy Jellyfish (22.04LTS). This version uses glibc2.36 (this is what I am currently on and works well for Wasp 0.7.1). 

To check your current Ubuntu version, run:

```
cat /etc/os-release
```
To check your current glibc version, run:

```
ldd --version ldd
```
To upgrade glibc version, run:

```
sudo apt-get update -y
sudo apt-get install -y libc6

```
[...]

PR #868, opened 2022-12-06
Added a note on upgrading to the latest LTE for WSL and Ubuntu users

EmmanuelTheCoder:
Added a note inside the docs for WSL and Ubuntu users in order to avoid the `glibc_2.34` not found error when trying to run Wasp 0.7.1. This is in relation to issue #867.

### DIFF
--- a/web/docs/deploying.md
+++ b/web/docs/deploying.md
@@ -332,8 +332,7 @@ Now go back to your terminal and execute the following commands:
   Select `server` when prompted with `Select Service`. Press enter.
   Railway will now locate the Dockerfile and deploy your server 👍
 
-  When deployment is finished, you will see: `Deployment live at <url_to_wasp_backend>`
-  Copy this URL 📜. We need it for step 5!
+  When deployment is finished, you might see: `Deployment live at <url_to_wasp_backend>`. If not, go now to your [Railway dashboard](https://railway.app/dashboard) and in the server instance's `Settings` tab, click `Generate Domain`. Copy the new URL as we will need it for step 5! 📜
 
   4. Next, change into your app's frontend build directory `.wasp/build/web-app`:
   ```shell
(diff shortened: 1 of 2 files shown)


Martinsos:
Thanks @EmmanuelTheCoder ! This problem can also happen for Ubuntu users, so we should also put a NOTE for them?

Suggestion:
- let's put NOTE both for linux and for wsl
- let's make note shorter and a bit more generic, so we don't have to maintain it, smth like "NOTE: If you are getting glibc errors, likely solution is to upgrade your linux version. Check (link to issue) for more details."

EmmanuelTheCoder:
I am aware that this problem can also happen to Ubuntu users but putting "note for linux and wsl" in the docs will be a little confusing as Ubuntu is not the only Linux distro and we are not sure if this problem happens to users on other distros. 

I think a better way will be "Note for Ubuntu and WSL." Let me know your thoughts on this before amending the pull request @Martinsos .

I agree that we should make the note shorter and then point to the issue of GH in case anyone wants to read more about it.

Martinsos:
@EmmanuelTheCoder sorry for slow response on this one!
I think it can happen on any distro hm. Maybe we should just have a super short note like "getting glibc errors? Check here: <link to gh issue>".

I see you closed the PR, what happened?

EmmanuelTheCoder:
The main reason why I closed this was that I discovered I didn't create a branch for this, twas pushed to main. I will create another PR for this.

Which is better:
a. showing them how to solve the glibc error in docs in a few lines and then linking back here in case they want to read more.
b. making a super short note, "getting glibc errors? check here" that links here.

In my opinion, I think option a is better for good DevEx. 
@Martinsos

EmmanuelTheCoder:
Or what do you think?

cc @vincanger

Martinsos:
I am voting for b) due to it being easier to maintain for us + it will draw less of their attention from the docs, where there is already a lot of info.

EmmanuelTheCoder:
@Martinsos, made a pull request for this #895. Kept it super short as suggested.

PR #875, opened 2022-12-12
Update features.md

vincanger:
small changes to jobs docs

### DIFF
--- a/web/docs/language/features.md
+++ b/web/docs/language/features.md
@@ -705,7 +705,7 @@ job mySpecialJob {
 ####  `perform: dict` (required)
 
   - ##### `fn: ServerImport` (required)
-  An `async` JavaScript function of work to be performed. Since Wasp executes jobs on the server, you must import it from `@server`. The function receives a first argument which should be passed when the job is called, as well as the context containing any declared entities as the second (this is passed automatically by Wasp). Here is a sample signature:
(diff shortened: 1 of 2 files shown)


Martinsos:
LGTM but maybe best to have @shayneczyzewski have a look at it since he is Jobs master!

web/docs/language/features.md
```diff
@@ -695,7 +705,7 @@ job mySpecialJob {
 ####  `perform: dict` (required)
 
   - ##### `fn: ServerImport` (required)
-  An `async` JavaScript function of work to be performed. Since Wasp executes jobs on the server, you must import it from `@server`. The function receives a JSON value as the first argument and context containing any declared entities as the second. Here is a sample signature:
+  An `async` JavaScript function of work to be performed. Since Wasp executes jobs on the server, you must import it from `@server`. The function receives a first argument which should be passed when the job is called, as well as the context containing any declared entities as the second (this is passed automatically by Wasp). Here is a sample signature:
```
shayneczyzewski, line 708:
```suggestion
  An `async` JavaScript function of work to be performed. Since Wasp executes jobs on the server, you must import it from `@server`. The function receives a first argument which may be passed when the job is called, as well as the context containing any declared entities as the second (this is passed automatically by Wasp). Here is a sample signature:
```

shayneczyzewski (approved):
Looks good to me! The only caution I will note by changing `"@server/jobs/bar.js"` into `"@server/workers/bar.js"` is that in the actual Wasp exposed stuff, it all lives under `'@wasp/jobs/mySpecialJob.js'`. So users can call the dir workers but when we generate it and they want to reference a job, it will still be under `@wasp/jobs` namespace.

vincanger:
Should I merge this even though the macOS build failed?

Issue #546, opened 2022-04-01
[Generator] Default vs named imports


sodic:
JavaScript generally offers two approaches to importing and exporting names across modules.

The first option is using _named exports_:
```javascript
// foo.js
export const foo = // ...

// main.js
import { foo } from 'foo';
```
And the second option is using _default exports_:
```javascript
// foo.js
const foo = // ...
export default foo;

// imain.js
import foo from 'foo';

Martinsos:
Ha very interesting @sodic ! Thanks for this, I have never considered this deeply, but it all sounds very reasonable, and I am in for preferring named exports in Wasp over default ones.
I believe the main motivation was that when a certain file exports one single thing, it feels a bit weird to import it as named export. But we can do it , no problem!

How should we go about this? Should we not touch anything for now and then try to make this changes as we encounter those parts of the code? Or would you like to give it a try now, update how Wasp exports stuff? It would be a breaking change but that is ok at this stage.

sodic:
@Martinsos I probably wouldn't touch anything for now since our code is still changing rapidly. We can make the changes and update the docs as we encounter them.

Issue #487, opened 2022-03-03
Syntax considerations: How we define functions


sodic:
Wasp-generated code prefers defining functions as **arrow function expressions**. The docs suggest the same approach for user-defined Queries and Actions. This issue makes a case on why we should perhaps reconsider this preference.

## The dilemma
There are two "dimensions" when it comes to differentiating between function definitions in JavaScript:
1. Function statements vs. function expressions
2. Arrow syntax vs. regular syntax

Arrow functions statements do not exist, meaning that there are a total of three options: _regular-syntax statements_, _regular-syntax expressions_, and _arrow-syntax expressions_. Out of the three options, **the most appropriate way to define a named function seems to be a _regular-syntax statement_**:
```typescript
function f(foo) {
   // ....
} 
```
While the other two options certainly do have their use cases (e.g., forcing a lexical binding of `this`, shorter inline callbacks, etc.), they seem to be lacking when it comes to defining named functions. I'll do my best to list the advantages and disadvantages of all approaches. Most of the concerns are general, meaning some may not be particularly important for Wasp.

### Advantages statements have over expressions
1.  Despite assigning a function expression to a named variable, it remains anonymous in the call stack. This makes debugging more difficult, and stack traces less readable. While we could eliminate this issue by explicitly naming the function expression, it only works with the regular syntax: `const f = function f() { ... }`. This same can't be done with arrow functions, as they are always anonymous.
2.  We will run into problems if we decide to put exports (i.e., public members) at the top of the file (this is a different discussion :)). It isn't possible to reference functions defined as expressions before their assignment, which forces us to put non-exported functions at the top and exported functions (that use the non-exported ones) at the bottom. Function statements are [hoisted](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting), allowing us to list the public API before its implementation details.
3.  Statements are (arguably) easier to read. It's immediately clear that `function f...` represents a function. With `const f = function () ...`, not so much.
6.  Since function expressions can't be referenced before their assignment, mutual recursion isn't possible (though we'll very likely never need it)

### Advantages expressions have over statements
They become pretty handy in TypeScript. Let's suppose we want to define several functions that represent binary operations on numbers. All functions must follow the same interface (in Haskell, this would be `Double -> Double -> Double`).
<details>

<summary>Show example</summary>

```typescript
function add(x: number, y: number): number {
    return x + y;
}

function multiply(x: number, y: number): number {
    return x + y;
}

[...]

PR #886, opened 2022-12-20
add todo app w/ typescript

vincanger:
added todo app but refactored with typescript on server and client


examples/todo-typescript/src/client/MainPage.tsx
```diff
@@ -0,0 +1,94 @@
...
+const TasksList = ({tasks}: { tasks: Task[] }) => {
+  if (tasks.length === 0) return <p>No tasks yet.</p>;
+  return (
+    <div className='tasklist'>
+      {tasks.map((tsk, idx) => (
+        <Todo task={tsk} number={idx} key={idx} />
+      ))}
+    </div>
+  );
+};
```
sodic, line 70: I can recommend several things here.

### Numbering the list
You can skip sending the `number` prop if you turn this into an ordered list:
  - Change `<div className='tasklist'>` into `<ol className='tasklist'>`
  - Change the `div` inside `Todo` into an `li` (you can drop the `span` element that displays the number)
  
If you knew this but decided against it for styling reasons, ignore my comment :)

### Spreading object props
Assuming you drop the `number`, the `Todo` becomes a display model for `Task`. You can change it to directly accept the props you need inside the component, saving you from having to access `task.something` multiple times. 

To elaborate, instead of:
```typescript
const Todo = ({ task }: { task: Task }) => {
  // And having to access task.id, task.description, task.isDone...
}
```
You can do:
```typescript
const Todo = ({ id, isDone, description }: Task) => {
  // And directly access id, isDone, description
}
```
Don't worry, it won't make sending props more verbose. Just turn this:
```tsx
<Task task={task} key={idx} />
```
Into this:
```tsx
<Task {...task} key={idx} />
```

examples/todo-typescript/src/client/SignupPage.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import SignupForm from '@wasp/auth/forms/Signup';
```
sodic, line 3: I recommend avoiding default exports, check https://github.com/wasp-lang/wasp/issues/546 for details.

examples/todo-typescript/src/client/SignupPage.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+const SignupPage = () => {
```
sodic, line 5: I also recommend avoiding arrow function expressions (in favor of regular function statements). Check https://github.com/wasp-lang/wasp/issues/487 for details.

**NB:**  When we're typing functions, arrow functions are the only option we have, but that's not what we're doing here (the linked issue describes what that means in detail :).

examples/todo-typescript/src/server/actions.ts
```diff
@@ -0,0 +1,33 @@
+import HttpError from '@wasp/core/HttpError.js';
+import { Context, Task } from './serverTypes'
+
+type createArgs = { description: Task['description'] };
+
+export const createTask = async ({ description }: createArgs, context: Context) => {
```
sodic, line 6: This will soon be easier :smile:

examples/todo-typescript/src/server/actions.ts
```diff
@@ -0,0 +1,33 @@
+import HttpError from '@wasp/core/HttpError.js';
+import { Context, Task } from './serverTypes'
+
+type createArgs = { description: Task['description'] };
```
sodic, line 4: Nice work typing these!

Heads up, TypeScript uses capital letters for type names, which means this one should be called `CreateArgs`.

The same applies for the rest of the PR.

examples/todo-typescript/src/server/queries.ts
```diff
@@ -0,0 +1,9 @@
+import HttpError from '@wasp/core/HttpError.js';
+import { Context, Task } from './serverTypes'
+
+export const getTasks = async (args: null, context: Context): Promise<Task[]> => {
```
sodic, line 4: Are you sure the type of `args` is `null`? I think it's more likely to be an empty object `{}`.

If you wanted to say "I don't care what type this is" without resorting to `any`, it's best to use `unknown`.

examples/todo-typescript/src/server/serverTypes.ts
```diff
@@ -0,0 +1,11 @@
+import { Task as TaskType, User, Prisma } from '@prisma/client';
+
+export type Task = TaskType;
```
sodic, line 3: You don't need the type alias here, a direct export will do the trick:
```typescript
import { User, Prisma } from '@prisma/client';

export { Task } from '@prisma/client';

// ...
```

sodic (approved):
Nice work with this one, @vincanger!

I'm glad to see it's possible to leverage the type system even with such basic support for TypeScript.

Left a couple of comments you might want to take care of.

Martinsos:
> Nice work with this one, @vincanger! And sorry for taking so long to review it, feel free to ping me whenever I forget about a PR smile
> 
> I'm glad to see it's possible to leverage the type system even with such basic support for TypeScript.
> 
> Left a couple of comments you might want to take care of.

Hah @sodic don't underestimate your super cool TS support :D!

examples/todo-typescript/src/client/SignupPage.tsx
vincanger: isn't Wasp itself exporting the form as a default export?

examples/todo-typescript/src/client/MainPage.tsx
```diff
@@ -0,0 +1,88 @@
...
+    <ol className='tasklist'>
+      {tasks.map((tsk, idx) => (
+        <Todo {...tsk} key={idx} />
+      ))}
+    </ol>
+  );
+};
+
+function NewTaskForm() {
+  const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
```
sodic, line 69: Same thing applies: https://github.com/wasp-lang/wasp/pull/886#discussion_r1069416981.

You can ditch the `React` import.

examples/todo-typescript/src/client/SignupPage.tsx
sodic: Indeed it is :sweat_smile:
My bad!

examples/todo-typescript/src/server/actions.ts
```diff
@@ -0,0 +1,33 @@
+import HttpError from '@wasp/core/HttpError.js';
+import { Context, Task } from './serverTypes'
+
+type CreateArgs = { description: Task['description'] };
```
sodic, line 4: Oh, something I forgot to tell you. Check [this](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcAqBDAzgazgXjgG8AoOOASwBMAuOAOwFcBbAI2CjLiuCwGMoFMDAoR6dLDEH0A5lwpYAImOB1WECABtgGel0ZYOASVoMW7TgF8SJJCjgBhKDpjAAglBlYCcAAoU+HAAeTFwAGjgAch5+QWFRekiAPiA) out.

TypeScript's utility types are generally very useful and powerful, I recommend learning all of them: https://www.typescriptlang.org/docs/handbook/utility-types.html.

examples/todo-typescript/src/client/MainPage.tsx
```diff
@@ -0,0 +1,88 @@
...
+      });
+    } catch (err: any) {
+      window.alert('Error while updating task ' + err?.message);
+    }
+  };
+
+  return (
+    <li>
+      <input type='checkbox' id={id.toString()} checked={isDone} onChange={handleIsDoneChange} />
+      <span>{description}</span>{' '}
```
sodic, line 52: Hm, why do you need the space after the `span`?

examples/todo-typescript/src/server/actions.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  return context.entities.Task.create({
+    data: {
+      description,
+      user: { connect: { id: context.user.id } },
+    },
+  });
+};
+
+type UpdateArgs = { taskId: Task['id']; isDone: Task['isDone'] };
```
sodic, line 19: If you changed `taskId` to `id`, you could also simplify this with `Pick`.

sodic (approved):
Nice work! Left several more comments about things I didn't notice before.

examples/todo-typescript/src/client/MainPage.tsx
vincanger: 🚀

examples/todo-typescript/src/server/actions.ts
vincanger: oooh nice 
🚀

examples/todo-typescript/src/client/MainPage.tsx
vincanger: I don't :)
🚀

examples/todo-typescript/src/server/actions.ts
vincanger: 🚀

vincanger:
Thanks @sodic -- changes made 🚀

PR #888, opened 2022-12-20
WIP: Make Queries use `GET`

### LINKED ISSUE #65: On server error 500 internal details are also returned.
Use case: I implemented `getTask(id)` query and when I invoked it from client with an inexistent id (not in db), server sent 500 response along with Prisma error msg.

This should not have happened, we don't want to expose the internals. We had a generic message prepared to be used but it wasn't - we need to investigate why.

### LINKED ISSUE #445: Should we consider using `GET` instead of `POST` for Queries?
Ported over from Discord:

> shayne — Yesterday at 4:49 PM
Nothing urgent, but one general question about something I noticed a while ago when building Waspleau but was just reminded of: it seems we use POST for all operations. I think this makes sense for actions, since they can modify, but for queries should we be using GET instead?
Semantically it would make more sense to me given the read-only nature, but wasn't sure if there was a reason for current setup atm 

>martinsos — Yesterday at 5:53 PM
Good question! I remember that I wanted to make it "correct" by using GET for queries, and I remember there was a reason I didn't do it at the end, I believe it was due to not being sure how long is the query data. With GET, you are not guaranteed that server will read the "body" of request -> that depends on the implementation of that server, it is not specified by HTTP/REST -> but if you use only query string, then you are limited by its size also (again depending on implementations). So it was easiest to just go with POST for now (and I saw that BlitzJS is doing the same so I concluded probably the made same conclusions).
I am not saying we need to keep it like this, but this was the reasoning in the past.

>shayne — Yesterday at 8:32 PM
Thanks for the context! Ah, yeah I think most would probably not try to read any body for GET (which is prob correct 🤔 may even get stripped by some firewalls and LBs) and you would be constrained to max URL size. In most vanilla REST cases this should still be ok though, right? If they had a TON of info to send I guess we may want to consider what they are sending/requesting/why.
Also has some implications re: safe methods (ie, caching, re-requests, etc.), as well as possible dev expectations if they wanted to use Postman or something to hit it. We can keep an eye out and think more on it! 👍 My gut initially says if we have good reasons to default to POST when someone asks, lets do it. If we don't atm, we may consider using GET until we hit that use case just to be in line with vanilla expectations/standards... but it won't be a dealbreaker either way, so may not matter. haha 😄 Appreciate it! 

>martinsos — Today at 5:03 AM
I agree with you, and we should keep an eye on it. I believe the reasoning at the end was: let's go with POST so we don't have to worry about it any more, about hitting limits and similar. But maybe it was over-cautios decision. Would you open an issue for it, and you can copy paste our convo from here? So we have this in mind and we can see in the future if we can make it "more correct".

shayneczyzewski:
# Description

This PR is open mainly for the discussion around shifting Queries from `POST` to `GET`. The approach in this PR does move us in the "right direction" of using`GET` for their semantically correct operations. However, because Wasp lacks the input arg type info, if we simply made them URL params they would all come out on the server as strings. That would not be ideal since it would (a) be a breaking change, and (b) the user would have to do a ton of type conversions in their query functions. The approach taken here mirrors that of `POST` before- namely, serializing JSON as an `arg`. This is not ideal/traditional REST-style, however. :/

The main question is: do we wait on this until we allow users to specify the route and param types for Queries, or is this already a useful step in the right direction we want to take on, knowing it still may "look weird" to devs.

Before:
```
Request URL: http://localhost:3001/operations/get-task
Request Method: POST
Request Payload:
{"id":23}
```

After:
```
Request URL: http://localhost:3001/operations/get-task?args=%7B%22id%22:23%7D
Request Method: GET
```

Let's discuss! :D 

Fixes #445 

## Type of change

- [x] Code cleanup
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### DIFF
--- a/waspc/data/Generator/templates/server/src/app.js
+++ b/waspc/data/Generator/templates/server/src/app.js
@@ -35,8 +35,7 @@ app.use((err, req, res, next) => {
     return res.status(err.statusCode).json({ message: err.message, data: err.data })
   }
 
-  console.error(err);
-  return res.status(500).json({message: "An internal server error occurred."})
+  return next(err)
 })
 
 export default app
--- a/waspc/data/Generator/templates/server/src/app.js
+++ b/waspc/data/Generator/templates/server/src/app.js
@@ -35,7 +35,8 @@ app.use((err, req, res, next) => {
     return res.status(err.statusCode).json({ message: err.message, data: err.data })
   }
 
-  return next(err)
+  console.error(err);
+  return res.status(500).json({message: "An internal server error occurred."})
 })
 
 export default app
--- a/waspc/data/Generator/templates/server/src/routes/operations/_query.js
+++ b/waspc/data/Generator/templates/server/src/routes/operations/_query.js
@@ -4,14 +4,15 @@ import { handleRejection } from '../../utils.js'
 import {= operationName =} from "{= operationImportPath =}"
 
 export default handleRejection(async (req, res) => {
-  {=! TODO: When generating express route for query, generated code would be most human-like if we
-    generated GET route that uses query arguments.
-    However, for that, we need to know the types of the arguments so we can cast/parse them.
-    Also, there is limit on URI length, which could be problem if users want to send some bigger
-    JSON objects or smth.
-    So for now we are just going with POST that has JSON in the body -> generated code is not
-    as human-like as it should be though. =}
-  const args = req.body || {}
+  const args = req.query || {}
+
+  // TODO: Have a flag for if we should autoconvert numbers.
+  // Just doing some testing right now.
+  if (true) {
+    if (req.query.id) {
+      req.query.id = parseInt(req.query.id);
+    }
+  }
 
   const context = {
     {=# userEntityLower =}
--- a/waspc/data/Generator/templates/server/src/routes/operations/_query.js
+++ b/waspc/data/Generator/templates/server/src/routes/operations/_query.js
@@ -4,15 +4,7 @@ import { handleRejection } from '../../utils.js'
 import {= operationName =} from "{= operationImportPath =}"
 
 export default handleRejection(async (req, res) => {
-  const args = req.query || {}
-
-  // TODO: Have a flag for if we should autoconvert numbers.
-  // Just doing some testing right now.
-  if (true) {
-    if (req.query.id) {
-      req.query.id = parseInt(req.query.id);
-    }
-  }
+  const args = JSON.parse(req?.query?.args || '{}')
 
   const context = {
     {=# userEntityLower =}
--- a/waspc/data/Generator/templates/react-app/src/actions/core.js
+++ b/waspc/data/Generator/templates/react-app/src/actions/core.js
@@ -1,4 +1,4 @@
-import { callOperation } from '../operations'
+import { callAction } from '../operations'
 import {
   registerActionInProgress,
   registerActionDone,
@@ -11,7 +11,7 @@ export function createAction(actionRoute, entitiesUsed) {
       // The `return await` is not redundant here. If we removed the await, the
       // `finally` block would execute before the action finishes, prematurely
       // registering the action as done.
-      return await callOperation(actionRoute, args)
+      return await callAction(actionRoute, args)
     } finally {
       await registerActionDone(entitiesUsed, specificOptimisticUpdateDefinitions)
     }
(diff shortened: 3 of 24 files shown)


waspc/data/Generator/templates/server/src/app.js
```diff
@@ -35,7 +35,8 @@ app.use((err, req, res, next) => {
     return res.status(err.statusCode).json({ message: err.message, data: err.data })
   }
 
-  return next(err)
+  console.error(err);
+  return res.status(500).json({message: "An internal server error occurred."})
```
shayneczyzewski, line 39: This fixes leaking the internal error since we have no downstream handler for `next(err)` to transform it.

Martinsos: Dont we have a default error catcher that transforms any errors that are not shaped as HttpError into 500? I believe that is how Wasp works right now. Also, this is not connected with this PR right?

waspc/data/Generator/templates/server/src/routes/operations/_query.js
```diff
@@ -4,14 +4,7 @@ import { handleRejection } from '../../utils.js'
...
 
 export default handleRejection(async (req, res) => {
-  {=! TODO: When generating express route for query, generated code would be most human-like if we
-    generated GET route that uses query arguments.
-    However, for that, we need to know the types of the arguments so we can cast/parse them.
-    Also, there is limit on URI length, which could be problem if users want to send some bigger
-    JSON objects or smth.
-    So for now we are just going with POST that has JSON in the body -> generated code is not
-    as human-like as it should be though. =}
-  const args = req.body || {}
```
Martinsos, line 14: Ah yes well this captures it all!
Should we leave some of this comment around? Or maybe best just to create a GH issue to track this?

Martinsos (commented):
Looks good @shayneczyzewski !

This is a bit hard decision. What are pros and cons?
I guess this makes it a bit nicer / more semantically correct, right?
Yes it is still one big argument, but ok.
If there wasn't for limit, I would say it is certainly better. But if we say limit is not a problem, then that also stands.
So I think it is ok to do this change, why not!

And in the future we can type it better. But I don't see anything very bad with having one big argument, to be honest.

waspc/data/Generator/templates/server/src/app.js
shayneczyzewski: I kinda cheated since I knew it was small and this discussion PR would get some eyes to throw it in... so yes, not strictly related, but found during this one. :D

I don't see any default server error catcher/transformer. We do some reshaping on the client, but as of now, anything that is _not_ `instanceof HttpError` we just let it bubble up and leak details to the client. I can pull this out into a separate PR though.

shayneczyzewski: Moved here: https://github.com/wasp-lang/wasp/pull/894

shayneczyzewski:
Thanks for your thoughts @Martinsos. The more I think about this, the more I am inclined to leave it as a `POST` until we have enough information about args to make them proper `GET` requests. Right now, we are just trading one "strange" thing for another, which probably isn't worth it. And while the 2k char limit for some browsers for URIs is also there, I think that is a side concern- if you are adding that much info to your `GET` request I think "you are doing it wrong." :D

What do you think @sodic? Use this? Punt until we have proper types associated with route/query params in .wasp files? Something else? Thnx

Martinsos:
Makes sense! I am ok with both, although if I had to choose right now, I would probably go with this GET even though it has "weird" arguments. It still feels a bit more correct. But whatever you decide I am good with!

sodic:
I'm prefer keeping the POST for now.

I know some sites use long base64 URLs, but I always found it ugly. It's also a non-standard approach (i.e., browsers and the infrastructure aren't built to optimize on it). Plus, I think it makes for a worse developer experience (think about how debugging requests would look like in Chrome DevTools, or even by reading the logs).

Therefore, I agree with Shayne's last comment, we would be substituting one strange thing for another, and the one we have now is at least human readable and easy to debug.

Martinsos:
All right, sounds good to me!

shayneczyzewski:
Thanks for the input @Martinsos and @sodic. I will close this PR now. At least we reminded ourselves of the reasons to stick with `POST` until we have proper types for operations. :) One day... haha

Martinsos:
Hah yes :D. Maybe sooner than we think :D!

PR #895, opened 2022-12-22
Glibc note in docs

EmmanuelTheCoder:
I added a short note in the docs for Glibc errors as discussed in #868. The note points to #867


shayneczyzewski:
Note, we may not need a note anymore if this change works: https://github.com/wasp-lang/wasp/pull/907 We will see on the next release. Thanks for having this just in case though @EmmanuelTheCoder 👍🏻

Martinsos:
Ah it seems again that we arenot merging glibc note PR :D, sorry about that @EmmanuelTheCoder , but the reason is a happy one, which is the fact that we (well, @shayneczyzewski :D) fixed the issue!

PR #907, opened 2022-12-22
[CI]: Downgrade Ubuntu to 20.04 for build, upgrade deps

shayneczyzewski:
# Description

We have been seeing an uptick in user and service errors around GLIBC_2.34. This is because our CI file was using `ubuntu-latest`, and recently GitHub updated what version that points to: https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/

While a static binary using musl is the ideal long-term solution, it is fairly complex and a bit off the beaten path. At this stage, just reverting our build behavior back to pre-0.7.x Wasp version style, when things "just worked", should fix our current GLIBC issue (as they are almost always backward compatible). NOTE: This was also how we used to build in Oct, so it should "just work" as before.

Additionally, I noticed we were getting many warnings in our current runs related to deprecation warnings. I updated all our deps to latest and they are now gone. 🥳 

Example of existing warnings this PR removes by upgrading deps:
Screenshot 2022-12-22 141130

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -59,11 +59,11 @@ jobs:
           git config --global core.eol lf
 
       - name: Checkout the repo
-        uses: actions/checkout@v2.6.0
+        uses: actions/checkout@v2
 
       - name: Set up Haskell
         id: setup-haskell-cabal
-        uses: haskell/actions/setup@v1.1.4
+        uses: haskell/actions/setup@v1.1
         with:
           ghc-version: ${{ matrix.ghc }}
           cabal-version: ${{ matrix.cabal }}
@@ -74,7 +74,7 @@ jobs:
           cabal --version
 
       - name: Cache
-        uses: actions/cache@v2.1.7
+        uses: actions/cache@v2
         with:
           path: ${{ steps.setup-haskell-cabal.outputs.cabal-store }}
           # TODO: Right now, actions/cache updates cache only if cache was not fe
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -59,11 +59,11 @@ jobs:
           git config --global core.eol lf
 
       - name: Checkout the repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Set up Haskell
         id: setup-haskell-cabal
-        uses: haskell/actions/setup@v1
+        uses: haskell/actions/setup@v2
         with:
           ghc-version: ${{ matrix.ghc }}
           cabal-version: ${{ matrix.cabal }}
@@ -74,7 +74,7 @@ jobs:
           cabal --version
 
       - name: Cache
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         with:
           path: ${{ steps.setup-haskell-cabal.outputs.cabal-store }}
           # TODO: Right now, actions/cache updates cache only if cache was not fe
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -86,11 +90,11 @@ jobs:
           #   fails multiple times while others don't, its cache will likely get evicted,
           #   making it even slower to test and fix (uffff).
           #   When they fix this, we should remove ${{ github.run_id }} from the end of the key.
-          key: wasp-build-${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('waspc/waspc.cabal') }}-${{ hashFiles('waspc/cabal.project') }}-${{ github.run_id }}
-          restore-keys: wasp-build-${{ runner.os }}-${{ matrix.ghc }}-
+          key: wasp-build-${{ matrix.os }}-${{ matrix.ghc }}-${{ hashFiles('waspc/waspc.cabal') }}-${{ hashFiles('waspc/cabal.project') }}-${{ github.run_id }}
+          restore-keys: wasp-build-${{ matrix.os }}-${{ matrix.ghc }}-
```
shayneczyzewski, line 94: Changed from `runner.os` to `matrix.os` so it captures the actual Linux OS in cache name. This allowed me to test these changes, but I think overall it is good to keep this way too.

Martinsos: Makes sense, good catch!

.github/workflows/ci.yaml
```diff
@@ -40,7 +40,11 @@ jobs:
         cabal:
           - "3.6.2.0"
         os:
-          - ubuntu-latest
+          # Using Ubuntu 20.04 due to recent GLIBC_2.34 version mismatch errors for
```
Martinsos, line 43: Ok great! So if I compiled with older version of glibc, it is likely to work with the new version? If that is so that is great! We could then always ensure that we use a bit later ubuntu than the newest one, so that we don't get again into this situation where people have too old glibc.

Martinsos (approved):
This sounds good: effective and quick solution!

.github/workflows/ci.yaml
shayneczyzewski: Correct, everything I have found online seems to state the following:

> One of the GNU C Library's (glibc's) unwritten rules is that a program built against an old version of [glibc](https://www.gnu.org/software/libc/) will continue to work against newer versions of glibc.
- https://developers.redhat.com/blog/2019/08/01/how-the-gnu-c-library-handles-backward-compatibility

> Glibc has a versioning system that allows backward compatibility (older programs built to run on older versions of glibc will continue to run on new glibc); but it is of no help the other way around: programs that depend on newer glibc will usually not run on systems with older glibc.
- https://www.lightofdawn.org/wiki/wiki.cgi/NewAppsOnOldGlibc

And again, this is how it used to be working for us before GitHub made the change to `ubuntu-latest`. Up until Nov, it was already actually compiling everything against 20.04 :) and it worked for people on latest Linux versions to the best of our knowledge (i.e. you and Filip never had problems). So I think we are safe to be 1 behind latest for compiling.

Martinsos: Awesome! Let's explain in the comment that we are using a bit older Ubuntu version to ensure that glibc is not too knew, since stuff compiled on older glibc is compatible with new glibc but not other way around. So in the future, if we are tempted to update the ubuntu version, we will know to what to update it.

Issue #910, opened 2022-12-23
Full-stack type support for queries and actions


sodic:
We want our users to only have to define everything about a Query (or an Action) once - probably inside the Wasp file.

Wasp should then know how to type everything:
- The operation's implementation on the backend (both the arguments and the return type)
- The opearation's RPC function on the frontend

Subtasks:
- #902 
- #969 
- #975 
- #976

Related tasks:
- #901 
- #642 

> **Note**
> TRPC is also something worth looking into

sodic:
Closed by accident (GitHub misinterpreting my comment).

PR #945, opened 2023-01-10
TypeScript support for Queries and Actions (manual)

### LINKED ISSUE #901: Expose Prisma types as Wasp imports
Until we implement full-stack type support in Wasp , we should make entity types Prisma generates available to our users on both the server and the client.

**NOTE:** We need to make sure Prisma types cannot contain anything server-specific in terms of runtime.

This is how the usage might look like:
```typescript
import { Task } from '@wasp/entities/Task'

// ...

const { data: tasks } = useQuery<Task[]>(...)
```

### LINKED ISSUE #902: Add basic type support for Queries and Actions
Until we get proper full-stack type support, we should define and expose generic type constructors for typing Wasp Queries and Actions. 

This feature works together with the one described in #901. This is how its usage might look like:
```typescript
import { Query } from '@wasp/queries'
import { Task } from '@wasp/entities/task'

const getTasks: Query<[Task]> = (args, context) => {
  // args and context are now properly typed inside the function
}
```
**NOTE:** We wouldn't have a way to ensure that the entities passed as type arguments to the `Query` type match those listed by the Query declaration in the `.wasp` file. The users would have to worry about this themselves (for now at least).

sodic:
Fixes #902
Partially fixes #901 (server-side support)

https://user-images.githubusercontent.com/22752307/211559780-dcf52059-c2e0-4e4b-bd60-e6b2511351bc.mp4


waspc/src/Wasp/Generator/ServerGenerator/Common.hs
```diff
@@ -99,3 +102,14 @@ dotEnvServer = [relfile|.env.server|]
...
+buildEntityData :: String -> Aeson.Value
+buildEntityData name =
+  object
+    [ "name" .= name,
+      "prismaIdentifier" .= entityNameToPrismaIdentifier name
+    ]
+
+-- https://github.com/wasp-lang/wasp/issues/812#issuecomment-1335579353
+toESModulesImportPath :: FilePath -> FilePath
+toESModulesImportPath = (++ ".js") . reverse . drop 3 . reverse
```
infomiho, line 115: I'm not an expert by no means, but I know you can use `takeBaseName` from `System.FilePath.Posix` to get the base name from a path e.g. `takeBaseName "@ext/folder/name.js" == "name"`

waspc/data/Generator/templates/server/src/actions/_action.js
```diff
@@ -6,11 +6,13 @@ import prisma from '../dbClient.js'
...
-  return {= jsFnIdentifier =}(args, context)
+export default async function (args, context) {
+  return {= jsFnIdentifier =}(args, {
+    ...context,
+    entities: {
+      {=# entities =}
+      {= name =}: prisma.{= prismaIdentifier =},
+      {=/ entities =}
+    },
+  })
```
sodic, line 17: This function is exactly the same as the one in `src/queries/_query.ts`, but I think it's better to keep them separate  because they're semantically different despite having the same structure at the moment.

shayneczyzewski:
This is looking great, @sodic! 🥳 I will defer to others on much of the TS stuff in templates, all the other stuff looking good (but can do a more detailed final pass on the rest tomorrow). I just wanted to raise one point now- do we want to handle the situation where a user puts conflicting entities in their TS file vs. what they put into the .wasp file? For example:

```css
query getNumTasks {
  fn: import { getNumTasks } from "@server/queries.js",
  entities: [Task],
  auth: false
}
```

```js
// Here I am using Task and User, but only have Task in my .wasp file.
export const getNumTasks: AuthenticatedQuery<[Task, User], number> = async (args, context) => {
  console.log(context.entities.User.count())
  return context.entities.Task.count()
}
```

This will blow up right now with: `TypeError: Cannot read properties of undefined (reading 'count')`, which makes sense since we did not actually put that into the context. But I just worry if they didn't understand they were missing it on the wasp end it could be confusing for users, since in the TS types it _seems_ like it should be there.

I'm not sure of a great way to encode this in types (or even if we can/want to), but just wanted to raise it may be a potential area of confusion. But it could also just be me overthinking things, too. 😄

sodic:
@shayneczyzewski You're right. We don't currently enforce this, as mentioned by the PR's issue (#902):
> **NB:** We wouldn't have a way to ensure that the entities passed as type arguments to the Query type match those listed by the Query declaration in the .wasp file. The users would have to worry about this themselves (for now at least).

I've described what prevents us from having this functionality in #910.

But there's a twist!
It turns out we can have it after all (at least for the time being, full story in #910), and this is exactly what I did in the next PR:
 -  #952 

Have fun reviewing that one :)

shayneczyzewski:
> @shayneczyzewski You're right. We don't currently enforce this, as mentioned by the PR's issue (#902):
> 
> > **NB:** We wouldn't have a way to ensure that the entities passed as type arguments to the Query type match those listed by the Query declaration in the .wasp file. The users would have to worry about this themselves (for now at least).
> 
> I've described what prevents us from having this functionality in #910.
> 
> But there's a twist! It turns out we can have it after all (at least for the time being, full story in #910), and this is exactly what I did in the next PR:
> 
> * [TypeScript support for Queries and Actions (automatic) #952](https://github.com/wasp-lang/wasp/pull/952)
> 
> Have fun reviewing that one :)

Doh! Sorry for missing that note in the issue. And I love those plot twists- Haha can't wait to check out in the next one! Great job here again.

waspc/src/Wasp/Generator/ServerGenerator/JobGenerator.hs
```diff
@@ -81,13 +81,6 @@ genJob (jobName, job) =
         ]
     maybeJobSchedule = jobScheduleTmplData <$> J.schedule job
 
-    buildEntityData :: String -> Aeson.Value
-    buildEntityData entityName =
-      object
-        [ "name" .= entityName,
-          "prismaIdentifier" .= C.entityNameToPrismaIdentifier entityName
-        ]
```
shayneczyzewski, line 89: Nice job removing these two instances of duplication.

shayneczyzewski (approved):
Great job, @sodic! Based on our chat yesterday, I don't have anything meaningful to add here. Tried to review/research some of the TS in the process (i.e. it was fun to read online why one would chose `unknown` instead of `any` for the `Result`). I'll look at #952 next.

You still do have one convo open with @infomiho, and I'm sure @Martinsos will be interested to take a look before merging, but from my perspective it looks good and functions as intended. I think it will be a nice value add for our TS users 🚀

waspc/examples/todoApp/src/server/actions.ts
```diff
@@ -1,8 +1,9 @@
 import HttpError from '@wasp/core/HttpError.js'
+import { Task } from '@wasp/entities/'
```
infomiho, line 2: Maybe we could use `import type` for type-only imports? It could help with some tooling in the future, and it feels nice to make the distinction. https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#type-modifiers-on-import-names

```ts
import { type Task } from  '@wasp/entities/'
```

waspc/examples/todoApp/src/server/actions.ts
```diff
@@ -1,8 +1,9 @@
...
+import { AuthenticatedAction } from '@wasp/types'
 import { getSomeResource } from './serverSetup.js'
 
-
-export const createTask = async (task, context) => {
+export const createTask: AuthenticatedAction<[Task]> = async (task, context) => {
```
infomiho, line 6: If understand correctly, this list `[Task]` should the same list as you defined in the `.wasp` file?

infomiho (approved):
TS part looks good, some minor comments

waspc/examples/todoApp/src/server/actions.ts
sodic: Yes, it's a whole thing. Check this comment and follow links for the full story: https://github.com/wasp-lang/wasp/pull/945#issuecomment-1384501234

waspc/data/Generator/templates/server/src/actions/_action.js
```diff
@@ -6,11 +6,13 @@ import prisma from '../dbClient.js'
...
           consider in the future if it is worth removing this duplication. =}
 
-export default async (args, context) => {
-  context = { ...context, entities: {
-    {=# entities =}
-    {= name =}: prisma.{= prismaIdentifier =},
-    {=/ entities =}
-  }}
-  return {= jsFnIdentifier =}(args, context)
+export default async function (args, context) {
```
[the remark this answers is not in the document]
sodic: It's a mistake, nice catch. They should both be either `.js` or `.ts`. I think I'll stick with `js` until we start using TS features.

waspc/data/Generator/templates/server/src/types/index.ts
```diff
@@ -0,0 +1,53 @@
...
+{=# isAuthEnabled =}
+export type AuthenticatedQuery<Entities extends WaspEntity[], Result = unknown> = 
+  AuthenticatedOperation<Entities, Result>
+
+export type AuthenticatedAction<Entities extends WaspEntity[], Result = unknown> = 
+  AuthenticatedOperation<Entities, Result>
+
+type AuthenticatedOperation<Entities extends WaspEntity[], Result = unknown> = (
+  args: any,
+  context: {
```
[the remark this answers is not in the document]
sodic: I did have this at some point, but the problem is what the users sees when they hover over the type.

Ideally, I'd like the users to see as much of the structure as possible. When I inline it, they see this:

image

When I extract it into a type constructor, they see this:
image

I even wanted to inline `EntityMap`, but it just made things more confusing:
image
To make it useful, I'd have to inline `NameOf` and `DelegateFor`, which isn't really possible to do nicely.

With that said, extracting the common type constructor looks _much_ nicer. I just wanted to present this argument before we made the decision. You can now cast your vote :smile: 
- With `OperationContext`
- Without `OperationContext`

waspc/data/Generator/templates/server/src/actions/_action.js
Martinsos: Yup, I think that is correct approach! We have that happening with queries and actions on many places in the code and I think that is ok for now.

waspc/data/Generator/templates/server/src/entities/index.ts
```diff
@@ -0,0 +1,17 @@
...
+  {=/ entities =}
+} from '@prisma/client'
+  
+export {
+  {=# entities =}
+  {= name =},
+  {=/ entities =}
+} from '@prisma/client'
+
+export type WaspEntity =
```
Martinsos, line 14: Aha, this is just some kind of sum type that unifies all the entity types together?

waspc/src/Wasp/Generator/ServerGenerator/Common.hs
Martinsos: This function is not clear about what it does!
I agree that splitExtension is probably better choice, although I can't be for sure until I understand what we are trying to do here. Interesting you are always dropping 3 chars -> so you expect file that has extension of length 2?
Anyway, let's put some comments on it -> GH issue link is nice, but as a supportive thing, not as mandatory reading to understand what function does.
Maybe we could also improve the name.

waspc/src/Wasp/Generator/ServerGenerator/Common.hs
```diff
@@ -99,3 +102,14 @@ dotEnvServer = [relfile|.env.server|]
 -- client SDK identifiers. Useful when creating `context.entities` JS objects in Wasp templates.
 entityNameToPrismaIdentifier :: String -> String
 entityNameToPrismaIdentifier entityName = toLower (head entityName) : tail entityName
+
+buildEntityData :: String -> Aeson.Value
```
Martinsos, line 106: This is doing something quite specific while having a quite generic name.
So I would either give it more specific name, or inline it.

For example name could be: makeJsonWithEntityNameAndPrismaIdentifier .

Martinsos: I see now that it existed before already, but still, I would use this opportunity to improve this. Before it was localized so that is why it was ok -> now it is in Common and used in multiple places so we need a better name.

waspc/src/Wasp/Generator/ServerGenerator/OperationsRoutesG.hs
```diff
@@ -65,9 +65,10 @@ genOperationRoute spec operation tmplFile = return $ C.mkTmplFdWithDstAndData tm
           baseTmplData
 
     operationImportPath =
-      SP.fromRelFileP $
-        relPosixPathFromOperationsRoutesDirToSrcDir
-          </> fromJust (SP.relFileToPosix $ operationFileInSrcDir operation)
+      C.toESModulesImportPath $
```
Martinsos, line 68: Why do we need this here?

waspc/src/Wasp/Generator/ServerGenerator.hs
```diff
@@ -236,6 +237,30 @@ genRoutesDir spec =
...
+            object
+              [ "entities" .= allEntities,
+                "isAuthEnabled" .= isJust userEntityName,
+                "userEntityName" .= fromMaybe "" userEntityName,
+                "userViewName" .= fromMaybe "" userViewName
+              ]
+        )
+    allEntities = map (C.buildEntityData . fst) $ AS.getDecls @AS.Entity.Entity spec
+    userEntityName = AS.refName . AS.App.Auth.userEntity <$> AS.App.auth (snd $ getApp spec)
+    userViewName = (++ "View") <$> userEntityName
```
Martinsos, line 262: This one feels like a candidate for extracting at some point and going more global. But ok for now!

waspc/src/Wasp/Generator/ServerGenerator/Common.hs
Martinsos: Btw would it be candidate for using StrongPath? I believe we expect paths to be relative? It doesn't have to though, just wondering, depends if that brings some value or not.

Martinsos (commented):
@sodic looks great, this is such a nice improvement!

I left some comments, nothing major, biggest thing is toESModulesImportPath and that is also just improving it a bit.
But all together makes a lot of sense and will be awesome!

waspc/data/Generator/templates/server/src/entities/index.ts
sodic: Yes, that's exactly what it is.

waspc/src/Wasp/Generator/ServerGenerator/Common.hs
sodic: I'll change this in the my third PR (#962) to minimize conflicts.

waspc/src/Wasp/Generator/ServerGenerator/OperationsRoutesG.hs
sodic: Talked about it, I'll document it above the function.

waspc/src/Wasp/Generator/ServerGenerator.hs
sodic: Yes, same thing with `"password"`. I'll add a comment.

waspc/data/Generator/templates/server/src/types/index.ts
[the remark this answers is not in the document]
infomiho: I understand where you are coming from, and I agree that not seeing the fields is a horrible user experience. 

What if did something like wrapping the types in an `Expand` type that would force field reveal (something like this: https://stackoverflow.com/a/69288824)

My vote is to give user clarity:
1. with `OperationContext` + `Expand`
2. or without `OperationContext`

shayneczyzewski (approved):
LGTM 🚀 nice work!

infomiho (approved):

Martinsos (approved):
LGTM!

PR #951, opened 2023-01-13
Updates completion command info

### LINKED ISSUE #948: Update `wasp completion` help to mention `.zshrc` for MacOS
When running `wasp completion` the following is outputted:

```bash
..
After that, depending on your system, you will need to edit your bash profile:
- on MacOS (OSX): you will normally want to edit ~/.bashrc
...
```

It mentiones `.bashrc` even though `zsh` is the default shell on MacOS devices. It would be more fitting to mention `.zshrc` as most of MacOS users will be using that to configure their completions.

infomiho:
### Description
It updates the `wasp completion` command output to mention `.zshrc` instead of `.bashrc` for MacOS. 

Fixes #948


sodic (commented):
Hm, the source file, all functions, and the template file all have bash in their names. So, either:
- Our completion is shell-agnostic, which means we'll want to rename everything.
- Our completion truly only works with Bash (I know some tools ship different completion scripts for each shell. e.g. `fzf`), in which case we probably shouldn't rename anything here (and create an issue for supporting `zsh` completion).

I don't know which is it, @Martinsos might.

Martinsos:
Wohoo first PR @infomiho, nice :D!

@sodic good point there hm, I haven't thought about it! Maybe zsh works directly with bash-completion stuff, as in supports it? @infomiho could you investigate that and also test it on your Mac? Super quick search tells me it can but it seems like some extra command needs to be run? https://stackoverflow.com/questions/3249432/can-a-bash-tab-completion-script-be-used-in-zsh .

If not, then yes let's drop this and we can open separate issue to support zsh.
If it does support it, we can let this be called bash-completion throughout the code, since that is what it is, even if zsh supports it, and as for the message: I think it can remain as Miho changed it, since that is the default on Mac. Maaaybe we could mention in parentheses that they can also put it in .bash_profile if they use bash instead.

infomiho:
Couple of observations after testing out on a new user account on my Mac with vanilla `zsh` setup.

1. Using bash completion in plain zsh setup results in:
  ```
  /Users/testuser/.completions/wasp-completion:5: command not found: complete
  ```
2. A way around it is to init bash completion support like this: https://github.com/eddiezane/lunchy/pull/59/files#diff-7d4c95f8626f6ebbd1d427d80250686d6c1510dd87f47a16d6a941672bb8c0feR6

3. This setup worked:
```bash
autoload -U compinit && compinit -u
autoload bashcompinit && bashcompinit

export PATH=$PATH:/Users/testuser/.local/bin

source .completions/wasp-completion
```

This would mean that our completions _are_ bash specific, and if we wanted for user to do `wasp completion:generate > <your-chosen-directory>/wasp-completion` for ZSH we would need to change the output. Either we decide to write proper ZSH completions or opt for using the compatibility mode.

The command should then take a param e.g.
- `wasp completion:generate zsh > <your-chosen-directory>/wasp-completion`
- `wasp completion:generate bash > <your-chosen-directory>/wasp-completion`

I'm closing this PR since it's trying to solve the wrong problem IMHO, I'll create an issue for creating ZSH completions support.

PR #952, opened 2023-01-16
TypeScript support for Queries and Actions (automatic)

### LINKED ISSUE #910: Full-stack type support for queries and actions

### LINKED ISSUE #969: Generate specific types for Queries and Actions
## The idea
Defining this in a .wasp file:
```css
query getTasks {
  fn: import { getTasks } from "@server/queries.js",
  entities: [Task]
}
```
Tells us everything we need to know about the Query's arguments. Therefore, we want to be able to do:
```typescript
import { GetTasks} from '@wasp/queries'
    
const getTasks: GetTasks = (args, context) => {
  // The function knows the types of  'args' and 'context' 
}
```
And know the types of `args` and `context` inside the function's body. Wasp still doesn't support specifying operations' return types, but we can at least allow the users to specify them themselves.
```typescript
import { Task } from '@wasp/entities'
import { GetTasks} from '@wasp/queries'
    
// 'args', 'context' and the return value would now have the correct type
const getTasks: GetTasks<Task[]> = (args, context) => { 
  // The function knows the types of  'args' and 'context' 
  // It also knows it must return a value of type Task[]
}
```
This feature should rely on functionality introduced for #901.

## Main challenges
It isn't impossible to tell TypeScript how to type code "from the outside." In other words, we'd have to either:
- Generate empty (but typed) TS function definitions based on the declarations inside the .wasp file
- Generate type definitions users could import and use to type their Query function

The second approach could look something like this:
```typescript
import { GetTasks} from '@wasp/queries'
    
// 'args', 'context' and the return value would now have the correct type
const getTasks: GetTasks = (args, context) => { ... }
```

---
> **Warning**
> (Un)fortunately, as it currently stands, our compiler will happily parse the query declaration even if its function isn't defined. Therefore, the rest of this text doesn't currently apply, and I went with approach 2 in #952. If we ever fix (break) this functionality, the rest of the text will start to apply.

---

Both approaches suffer from a chicken-and-egg problem:
> To generate a type definition for a query, we must successfully compile the `.wasp` file containing the query's declaration. To successfully compile the `.wasp` file, every query declaration must include a valid import of its TS definition.

It's certainly possible to work around this limitation. For example, we could stop requiring external imports inside Query declarations. Instead, we could:
1. Generate a query type based on the import-less declaration.
2. Tell the users to type their desired Query implementation with it.
3. Find the appropriate implementation by looking for this type inside TypeScript's AST.

This approach does have its problems (e.g., it makes TypeScript mandatory), but is interesting to consider

Still, regardless of the approach we decide take, tackling this will surely entail substantial changes to the Wasp compiler.

sodic:
Fixes #969.

Partially fixes #910.

Requires #945 to work.

https://user-images.githubusercontent.com/22752307/212577606-736fc0ed-0229-4646-b4cd-a5f88a9e8549.mp4


waspc/examples/todoApp/src/server/queries.ts
```diff
@@ -1,8 +1,12 @@
...
 import { Task } from '@wasp/entities'
-import { AuthenticatedQuery } from '@wasp/types'
+import {
+  GetNumTasks,
+  GetTask,
+  GetTasks
+} from '@wasp/queries/types'
 
-export const getTasks: AuthenticatedQuery<[Task], Task[]> = async (args, context) => {
+export const getTasks: GetTasks<Task[]> = async (args, context) => {
```
sodic, line 9: The type parameter represents the return value, not entities.

waspc/src/Wasp/Generator/ServerGenerator/OperationsG.hs
```diff
@@ -10,45 +10,95 @@ where
...
+    operations = map (uncurry AS.Operation.ActionOp) $ AS.getActions spec
+    isAuthEnabledGlobally = isAuthEnabled spec
 
 -- | Here we generate JS file that basically imports JS query function provided by user,
 --   decorates it (mostly injects stuff into it) and exports. Idea is that the rest of the server,
 --   and user also, should use this new JS function, and not the old one directly.
-genQuery :: AppSpec -> (String, AS.Query.Query) -> Generator FileDraft
-genQuery _ (queryName, query) = return $ C.mkTmplFdWithDstAndData tmplFile dstFile (Just tmplData)
+genQuery :: (String, AS.Query.Query) -> Generator FileDraft
+genQuery (queryName, query) = return $ C.mkTmplFdWithDstAndData tmplFile dstFile (Just tmplData)
```
sodic, line 67: @shayneczyzewski I removed `AppSpec` from the list of arguments because the function doesn't use it.

I know you somtimes keep unused arguments for the sake of API consistency, but in this case I only see two functions that "belong" to the API (`genQuery` and `genAction`) and neither uses the `AppSpec` argument. Correct me if I missed something.

sodic (commented):
I am not a big fan of how I designed this:
- The two template files are pretty similar.
- The logic for generating file drafts is mostly the same. I've therefore extracted it into `genOperationTypes`, but that function feels like a weak abstraction (especially thanks of the `isAuthEnabledGlobally` argument, but all arguments are pretty unrelated).

Better ideas are welcome

waspc/src/Wasp/Generator/ServerGenerator/OperationsG.hs
shayneczyzewski: Sounds good to me! I think those signatures have had the param ignored for a few years now, so fully agree cleanup makes sense if unused. 👍🏻

shayneczyzewski:
> I am not a big fan of how I designed this:
> 
> * The two template files are pretty similar.
> * The logic for generating file drafts is mostly the same. I've therefore extracted it into `genOperationTypes`, but that function feels like a weak abstraction (especially thanks of the `isAuthEnabledGlobally` argument, but all its arguments are pretty unrelated).
> 
> Better ideas are welcome.

Nice job @sodic! Doing my typical fast first review to get a high-level idea, then circle back :D Also, my eyes/brain are a bit tired already hehe. 

I think keeping the two concepts separate, even though they share much of the same shape, is still probably fine. I'm not sure how we could get around it. If we tried to combine them in templates somehow, if at any point they start to diverge, we'd have to rip them back apart again. And it would just muddle the concepts.

One thing I totally forgot about was `isAuthEnabledGlobally` and its impact on the default nature of `auth` in operations. Makes sense, but that felt like a pretty big concept I implicitly understood but never paused to recognize how it was implemented until now. 😄

Lastly, while this snippet makes total semantic sense:
```
import {
  {=^ allUseAuth =}
  Action,
  {=/ allUseAuth =}
  {=# someUseAuth =}
  AuthenticatedAction,
  {=/ someUseAuth =}
} from '../types'
```

It did take me a bit of staring and wordplay in my head to realize it was "not all use auth" aka "some don't use auth". I know, I'm tired today lol But the all vs some right next to each other just threw me. But sadly, I don't have a better way to encode this myself, or a better way to say something is "non-auth". At this point, this comment is starting to feel like a rambling diary entry lol.

Ok, I'll wrap it up by saying the approach seems to make sense to me but will dive deeper in the next day or so in case I have any epiphanies on how to better handle the similarities.

Lastly, the availability of these new types I think is super helpful. We still of course can't guarantee they match the right name to the right type though (can we? maybe with declaration merging?). But that feels a bit much to tackle now with minimal gains.

sodic:
@shayneczyzewski Thanks for the comment. It was very interesting to read, more like a short story than a review comment :smile: 

> We still of course can't guarantee they match the right name to the right type though (can we? maybe with declaration merging?). But that feels a bit much to tackle now with minimal gains.

Unfortunately, no. I'm not sure what you mean by declaration merging. I described a theoretical approach we might take in #910 (the part that talks about working on TypeScript's AST), but I definitely wouldn't go there for now:
 - It would make TypeScript a necessary target for Wasp
 - It would take too long.

shayneczyzewski:
> @shayneczyzewski Thanks for the comment. It was very interesting to read, more like a short story than a review comment 😄
> 
> > We still of course can't guarantee they match the right name to the right type though (can we? maybe with declaration merging?). But that feels a bit much to tackle now with minimal gains.
> 
> Unfortunately, no. I'm not sure what you mean by declaration merging. I described a theoretical approach we might take in #910 (the part that talks about working on TypeScript's AST), but I definitely wouldn't go there for now:
> 
> * It would make TypeScript a necessary target for Wasp
> * It would take too long.

Haha thanks, could be a top and worst PR comment to date. :D

By declaration merging (showing my TS ignorance), I meant something like this: https://www.typescriptlang.org/docs/handbook/declaration-merging.html Basically wondering aloud if we were able to provide TS with a prototype definition of the expected functions that included the correct names and types, so when they then wrote the declaration in their own files it would automagically be correctly typed for them.

So something like this in the global scope, we say:
```ts
foo: Baz
```

and later, if they declare `foo`, it will implicitly be of type `Baz` via declaration merging. But I don't think that is a supported use case for that feature.

sodic:
@shayneczyzewski Oh, I get what you mean. That's actually an excellent idea. We'd be doing what C does when declaring functions.

I thought about doing it, and it would indeed be the best solution. This is how I imagined It could look like:
```typescript
// We generate a file that contains this, call it '@wasp/types'
declare function getTask(args: number, context: number) 

// ...
// Users import the declarations in their queries.ts file
import '@wasp/types'

function getTask(args, context) {
  // TS knows the types of args and context
}
```
Looks awesome, but yes, TS unfortunately [doesn't support it](https://stackoverflow.com/questions/57383099/typescript-function-declaration-separate-from-implementation).
Btw, the top commenter's suggestion doesn't help us because we already have a way to throw type errors when Query types aren't valid (i.e., just check their types when importing them from external code). What we want is IDE support.

shayneczyzewski:
> @shayneczyzewski Oh, I get what you mean. That's actually an excellent idea. We'd be doing what C does when declaring functions.
> 
> I thought about doing it, and it would indeed be the best solution. This is how I imagined It could look like:
> 
> ```ts
> // We generate a file that contains this, call it '@wasp/types'
> declare function getTask(args: number, context: number) 
> 
> // ...
> // Users import the declarations in their queries.ts file
> import '@wasp/types'
> 
> function getTask(args, context) {
>   // TS knows the types of args and context
> }
> ```
> 
> Looks awesome, but yes, TS unfortunately [doesn't support it](https://stackoverflow.com/questions/57383099/typescript-function-declaration-separate-from-implementation). Btw, the top commenter's suggestion doesn't help us because we already have a way to throw type errors when Query types aren't valid (i.e., just check their types when importing them from external code). What we want is IDE support.

Yes, exactly! I was trying to phrase my Google searches akin to "C-style function prototypes in TS" :D But that is a bummer it doesn't support it. This current approach then seems the most feasible then given current TS capabilities.

waspc/data/Generator/templates/server/src/actions/types.ts
```diff
@@ -0,0 +1,33 @@
...
+
+{=# operations =}
+export type {= typeName =}<Output = unknown> = 
+  {=# usesAuth =}
+  AuthenticatedAction<
+  {=/ usesAuth =}
+  {=^ usesAuth =}
+  Action<
+  {=/ usesAuth =}
+    [
```
infomiho, line 25: Nice simplification

waspc/src/Wasp/Generator/ServerGenerator/OperationsG.hs
```diff
@@ -10,45 +10,95 @@ where
...
+          "allEntities" .= nub (concatMap entityNames operations)
+        ]
+    operationTypeData operation =
+      object
+        [ "typeName" .= toUpperFirst (getName operation),
+          "entities" .= entityNames operation,
+          "usesAuth" .= usesAuth operation
+        ]
+    entityNames = maybe [] (map AS.refName) . AS.Operation.getEntities
+    usesAuth = fromMaybe isAuthEnabledGlobally . AS.Operation.getAuth
```
infomiho, line 97: Maybe make this produce value immediately 
```
fromMaybe isAuthEnabledGlobally . AS.Operation.getAuth $ operations
```
so you can use it like `any usesAuth` and `all usesAuth`.

infomiho (approved):
I know there some duplication in the template, but I think it's better to keep it simpler than overcomplicate with searching with some common abstraction. 

Looking forward to using this in the example apps.

Martinsos:
Oh wow this was some very cool brainstorming to read @shayneczyzewski and @sodic , especially the part with declaration merging! I didn't know TS can do that, pretty wild. Although keep in mind that next to all the other issues, we also don't know it's return type in advance, so even if we wanted to type those on our own, we wouldn't know how to type return value.

All of your thoughts on this sound good to me. One thought: what if we just parsed TS files and checked that they typed them with correct type? That doesn't sound that hard? When we have a query, we know in which files it resides, and under which name. So we can take that TS file, cram it into TS compiler, get the AST out, find the corresponding name, check if it has correct type, and if not, we tell them they messed up. Yes, this means doing manually what we would prefer TS to be doing, but still, it would work, and wouldn't be that terrible! I wouldn't be doing that now though, but I also don't think that is extremely hard to do, we could probably get that working in a couple of days if it was really important.
Although if we go into parsing TS files there are also other cool things we might start doing, but ok that is to be seen.

waspc/data/Generator/templates/server/src/actions/types.ts
```diff
@@ -0,0 +1,33 @@
+{{={= =}=}}
+import {
+  {=# allEntities =}
+  {= . =},
+  {=/ allEntities =}
+} from '../entities'
+
+import {
+  {=^ allUseAuth =}
```
Martinsos, line 9: As @shayneczyzewski commented, it is true that it is hard to see this `^` here, and even when we see it, it is not easy to remember waht it does.
One solution might be trying to avoid it -> in this case we could pass notAllUseAuth instead of allUseAuth and that would solve it. But I am also ok with leaving it as it is.

waspc/data/Generator/templates/server/src/queries/types.ts
```diff
@@ -0,0 +1,33 @@
+{{={= =}=}}
```
Martinsos, line 1: You can throw in a comment in both of these files thta they are very similar to each other. I did this in some other places. It means that we know about duplication, but we are ok with it. It also point whoever might be refactoring file to also look at the other one.

waspc/data/Generator/templates/server/src/queries/types.ts
```diff
@@ -0,0 +1,33 @@
...
+  {=^ allUseAuth =}
+  Query,
+  {=/ allUseAuth =}
+  {=# someUseAuth =}
+  AuthenticatedQuery,
+  {=/ someUseAuth =}
+} from '../types'
+
+{=# operations =}
+export type {= typeName =}<Output = unknown> =
```
Martinsos, line 18: Maybe call Output -> ReturnValue? Or Returns? I see that you called it "return value" below in other comments, that is why, and I also usually use that phrase.
Not a biggie, but maybe worth considering.

waspc/src/Wasp/Generator/ServerGenerator/OperationsG.hs
Martinsos: Nooooooo :D. Kidding, and I wonder if we should have at some point put AppSpec in Generator Monad. But maybe not. Anyway all good.

waspc/src/Wasp/Generator/ServerGenerator/OperationsG.hs
Martinsos: The thing is, any and all actually take a predicate function which they run on the given list(Foldable). So what you wrote couldn't be done. For that thought, you could use functions like `and` and `or`.

Martinsos (approved):
I think this is great!
Love the result, it is really cool, good trade off between getting some great value right now and not tackling too much at once.
Also, I tihnk keeping query and actions seperate is good choice, for the same reason Shayne described.
All together great!
Are we still exposing them with Query, AuthenticatedQuery, and other types? Are they needed in any way, I guess not at all, no reason for them to use them right?

waspc/src/Wasp/Generator/ServerGenerator/OperationsG.hs
sodic: I'd need two values (the `usesAuth` function, and the result of mapping it over the operations that I would then use `and` and `or`), which I think makes the entire a little less readable (you need three "jumps" to understand what's going on).
It is theoretically more efficient, but these arrays are almost always smaller than 10, and always in the two digit space.
Anyway, I'd like to leave it like this if that's fine with you.

waspc/data/Generator/templates/server/src/actions/types.ts
sodic: I think I got something better.

sodic: Fixed [here](https://github.com/wasp-lang/wasp/pull/952/files/27ded86925f799cb6805cfc2244e84f6a4e422c9..4b778fd8a06af328b5b32645e1548c66a3a278c8).

waspc/data/Generator/templates/server/src/queries/types.ts
sodic: Your suggestions might make more sense, but I think the current name is most inline with TS conventions:
- `ReturnType` and `SomethingType` are used for type constructors (extractors) and they already exist in the global scope
- `Returns` is a verb, and those are generally not used to name types in TypeScript
- `ReturnValue` could be ok, but `Value` is also a little weird to encounter in a type
- 
Therefore, unless you really like `ReturnValue`, I'd stick with `Output`.

waspc/data/Generator/templates/server/src/actions/types.ts
shayneczyzewski: I like the naming update 👍🏻

waspc/data/Generator/templates/server/src/actions/types.ts
```diff
@@ -0,0 +1,35 @@
+{{={= =}=}}
+{=! TODO: This template is exactly the same at the moment as one for action
+          types, consider whether it makes sense to address this in the future. =}
```
shayneczyzewski, line 3: Should this one say "same as query"? And similar for other.

shayneczyzewski (approved):
Changes LGTM @sodic, nice job! 🚀 Left one minor comment about a comment (very meta, I know :D).

waspc/data/Generator/templates/server/src/actions/types.ts
sodic: I can't believe I remembered to change them this time, but switched them up. Thanks!

Martinsos (approved):

PR #500, opened 2022-03-09
[Analyzer] Reverse the order of Wasp error messages

sodic:
# Description

This PR changes the order in which the compiler displays error messages and their context (check the issue for details).

Fixes #414

## Type of change

Please select the option(s) that is more relevant.

- [x] Code cleanup
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

PR #953, opened 2023-01-16
Adding uninstall command

### LINKED ISSUE #896: Implement a straight-forward way to uninstall Wasp
Having a straight-forward uninstall options is (IMO) a must:
- It shows we believe the product is useful
- It shows we respect our users and don't want to coerce them into keeping Wasp on their systems if they don't like it
- People sometimes attempt to fix issues by uninstalling and installing them (it won't help with Wasp, but they'll try it anyway), we might as well make it as ergonomic as possible
- Having an "official way" to uninstall a program removes all worries that there is still something left over (this ties into the previous point)

infomiho:
### Description

Fixes #896

Adding an uninstall command. It's still a draft PR, but it's a first basic working version:
- removing the Wasp dir with current `version`
- removes the script in `.local/bin`

Things to rework:
- [x] error handling
- [x] discuss if we want to remove ALL Wasp version vs. only current
- [x] investigate if there is a way to better encode which files need to be deleted
    - they are living in two places: install script and uninstall command


waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
```diff
@@ -0,0 +1,70 @@
...
+-- | Removes Wasp CLI from the system.
+-- It removes the follwing:
+-- {home}/.local/share/wasp-lang
+-- {home}/.local/bin/wasp
+uninstall :: Command ()
+uninstall = do
+  cliSendMessageC $
+    Msg.Info "Uninstalling Wasp CLI ..."
+
+  uninstallResult <- liftIO (try removeWaspFiles :: IO (Either SomeException ()))
```
infomiho, line 35: I'm not sure if this is the best way to handle errors, let me know 😃

sodic (commented):
Great work so far!

The main question of my review is whether we should reconsider how we define which files to delete (check https://github.com/wasp-lang/wasp/pull/953#discussion_r1071544116 for details).

Still, I reviewed the rest of the code as I would if there was no such question (for learning purposes). Therefore, the final version of the PR probably won't include most of my suggestions.

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
```diff
@@ -0,0 +1,70 @@
...
+  let waspDir = getWaspBinariesDir homeDir
+      waspBin = getWaspScript homeDir
+
+  -- Deleting dir with Wasp binaries
+  doesWaspDirExist <- liftIO $ doesDirectoryExist $ SP.fromAbsDir waspDir
+  when doesWaspDirExist $ do
+    removeDirectoryRecursive $ SP.toFilePath waspDir
+
+  -- Deleting the script referencing Wasp binary
+  doesWaspScriptExist <- liftIO $ doesFileExist $ SP.fromAbsFile waspBin
```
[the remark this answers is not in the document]
infomiho: Nice catch, I had everything together in one big function at first and this code was called inside of the Command monad (?) I guess I said that correctly 😄

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
```diff
@@ -0,0 +1,70 @@
...
+  when doesWaspScriptExist $ do
+    removeFile $ SP.fromAbsFile waspBin
+
+getHomeDir :: IO (SP.Path' SP.Abs (SP.Dir HomeDir))
+getHomeDir = fromJust . SP.parseAbsDir <$> getHomeDirectory
+
+getWaspBinariesDir :: SP.Path' SP.Abs (SP.Dir HomeDir) -> SP.Path' SP.Abs (SP.Dir String)
```
Martinsos, line 62: You didn't want to bother with having special `data` for WaspBinariesDir, which is fine, but in such case `String` is still a bit unusual, best to use `Dir ()` or `Dir'` which is alias for it. This is recommended in StrongPath docs.

```suggestion
getWaspBinariesDir :: SP.Path' SP.Abs (SP.Dir HomeDir) -> SP.Path' SP.Abs (SP.Dir')
```

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
```diff
@@ -0,0 +1,70 @@
...
+getWaspBinariesDir homeDir =
+  homeDir
+    SP.</> [SP.reldir|.local/share/wasp-lang|]
+
+getWaspScript :: SP.Path' SP.Abs (SP.Dir HomeDir) -> SP.Path' SP.Abs (SP.File String)
+getWaspScript homeDir =
+  homeDir
+    SP.</> [SP.relfile|.local/bin/wasp|]
```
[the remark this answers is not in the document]
Martinsos: Nice one @sodic , glad to see that StrongPath Stockholm syndrome has started working :D!
@sodic covered it all very well here, nothing to add really!

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
```diff
@@ -0,0 +1,70 @@
...
+
+  -- Deleting dir with Wasp binaries
+  doesWaspDirExist <- liftIO $ doesDirectoryExist $ SP.fromAbsDir waspDir
+  when doesWaspDirExist $ do
+    removeDirectoryRecursive $ SP.toFilePath waspDir
+
+  -- Deleting the script referencing Wasp binary
+  doesWaspScriptExist <- liftIO $ doesFileExist $ SP.fromAbsFile waspBin
+  when doesWaspScriptExist $ do
+    removeFile $ SP.fromAbsFile waspBin
```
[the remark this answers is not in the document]
Martinsos: Although it can also be fine calling them twice -> it shows intention at the place of the call, that you wanted to have abs file here, or abs dir, ... .

Martinsos (changes requested):
Nice job :)!

I have to admit I am always a bit scared when we have to delete some files for the user -> what if we delete wrong files somehow?

Left some comments, although @sodic already covered it pretty well.

I guess we could try to do some advanced mechanisms to prevent this, but one simple thing we could do is ask user, before we initiate deletion, if they are ok with us deleting specified paths. So what we could add here is a question: "We will delete the following: ... -> are you ok with us proceeding? [Y/N]". What do you think, does that make sense? Maybe I am overreacting a bit, but still.

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
```diff
@@ -0,0 +1,70 @@
...
+  uninstallResult <- liftIO (try removeWaspFiles :: IO (Either SomeException ()))
+  either
+    (throwError . CommandError "Unable to uninstall" . show)
+    pure
+    uninstallResult
+
+  cliSendMessageC $ Msg.Success "Uninstalled Wasp CLI"
+
+removeWaspFiles :: IO ()
+removeWaspFiles = do
```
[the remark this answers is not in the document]
Martinsos: Depends if they are used somewhere else or just here! But ok, that is easy to move around.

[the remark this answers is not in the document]
Martinsos: I would maybe use a bit different names: `let waspDirAbsPath` and `let waspBinAbsPath`. But no biggie, doesn't matter much.

Or I would even be tempted to do `(doesDirectoryExist . fromAbsDir)`, and `(removeDirectoryRecursive . fromAbsDir)`, and so on, since it is really property of these functions to define what kind of path they want to take in. If they were typed, we wouldn't have to do any conversion at all, but if we have to, we can do it right here where we call them.

I haven't been doing this in the past, but now I am getting to like it as an approach.

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
```diff
@@ -0,0 +1,70 @@
...
+  homeDir <- getHomeDir
+  let waspDir = getWaspBinariesDir homeDir
+      waspBin = getWaspScript homeDir
+
+  -- Deleting dir with Wasp binaries
+  doesWaspDirExist <- liftIO $ doesDirectoryExist $ SP.fromAbsDir waspDir
+  when doesWaspDirExist $ do
+    removeDirectoryRecursive $ SP.toFilePath waspDir
+
+  -- Deleting the script referencing Wasp binary
```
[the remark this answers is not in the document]
Martinsos: I agree. You can mention on the path to wasp bin script itself this, if you want to describe what it is, but even that might not be necessary.

Martinsos:
I forgot to comment about the knowledge of installation paths.

So one way to do it is to use XDG as @sodic said, so that might be the way to go.

Another way could be having installer script pass on the paths to the wasp CLI. This is already happening for waspc data dir -> it could also pass paths to bin and whole wasp installation. That way knowledge is transfered. But it does sound a bit too complex when I think about it, even if relatively logical.

Last approach is hardcoding it here and there, so duplicating, but putting comments on it saying that this is relying on installer script and if we change something here we need to change it there and vice versa -> yeah, not the nicest solution, but I think it works completely fine in practice and doesn't complicate things.

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
```diff
@@ -0,0 +1,70 @@
...
+uninstall :: Command ()
+uninstall = do
+  cliSendMessageC $
+    Msg.Info "Uninstalling Wasp CLI ..."
+
+  uninstallResult <- liftIO (try removeWaspFiles :: IO (Either SomeException ()))
+  either
+    (throwError . CommandError "Unable to uninstall" . show)
+    pure
+    uninstallResult
```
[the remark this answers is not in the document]
infomiho: I went for hte ultimate simplification:
```
removeWaspFiles
```
since any thrown errors would be outputed anyways :)

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
```diff
@@ -0,0 +1,70 @@
...
+  when doesWaspScriptExist $ do
+    removeFile $ SP.fromAbsFile waspBin
+
+getHomeDir :: IO (SP.Path' SP.Abs (SP.Dir HomeDir))
+getHomeDir = fromJust . SP.parseAbsDir <$> getHomeDirectory
+
+getWaspBinariesDir :: SP.Path' SP.Abs (SP.Dir HomeDir) -> SP.Path' SP.Abs (SP.Dir String)
+getWaspBinariesDir homeDir =
+  homeDir
+    SP.</> [SP.reldir|.local/share/wasp-lang|]
```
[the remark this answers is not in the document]
infomiho: Let's do everything the hard-coded way, and then iterate on both ends: installation and installation 👍

waspc/cli/src/Wasp/Cli/Command/Telemetry/Common.hs
```diff
@@ -2,17 +2,14 @@ module Wasp.Cli.Command.Telemetry.Common
...
 
 import StrongPath (Abs, Dir, Path', reldir)
 import qualified StrongPath as SP
 import qualified System.Directory as SD
-
-data UserCacheDir
-
-getUserCacheDirPath :: IO (Path' Abs (Dir UserCacheDir))
-getUserCacheDirPath = SD.getXdgDirectory SD.XdgCache "" >>= SP.parseAbsDir
+import Wasp.Cli.Command.Utils.File (UserCacheDir, getUserCacheDirPath)
```
sodic, line 12: We generally prefer to list symbols in a file from top to bottom: https://github.com/wasp-lang/wasp/pull/500#discussion_r824602425.

So when we encounter a file where this isn't the case, we shuffle them around a bit. Nothing to do here, this file is already in order, just FYI for future references.

waspc/cli/src/Wasp/Cli/Command/Telemetry/Common.hs
```diff
@@ -25,4 +22,7 @@ ensureTelemetryCacheDirExists = do
   return telemetryCacheDirPath
 
 getTelemetryCacheDirPath :: Path' Abs (Dir UserCacheDir) -> Path' Abs (Dir TelemetryCacheDir)
-getTelemetryCacheDirPath userCacheDirPath = userCacheDirPath SP.</> [reldir|wasp/telemetry|]
+getTelemetryCacheDirPath userCacheDirPath = getWaspCacheDirPath userCacheDirPath SP.</> [reldir|telemetry|]
+
+getWaspCacheDirPath :: Path' Abs (Dir UserCacheDir) -> Path' Abs (Dir TelemetryCacheDir)
+getWaspCacheDirPath userCacheDirPath = userCacheDirPath SP.</> [reldir|wasp|]
```
sodic, line 28: Don't forget about Strong Path conventions: https://github.com/wasp-lang/wasp/pull/953#discussion_r1071593511.

TL;DR: Drop the `SP.` prefix.

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
```diff
@@ -0,0 +1,63 @@
...
+-- {home}/.local/share/wasp-lang
+-- {home}/.local/bin/wasp
+-- {home}/.cache/wasp
+uninstall :: Command ()
+uninstall = do
+  cliSendMessageC $ Msg.Info "Uninstalling Wasp ..."
+
+  liftIO removeWaspFiles
+
+  cliSendMessageC $ Msg.Success "Uninstalled Wasp"
```
sodic, line 28: Too much spacing here my taste, but that's subjective.

waspc/cli/src/Wasp/Cli/Command/Utils/File.hs
```diff
@@ -0,0 +1,26 @@
...
+
+deleteFileIfExists :: FilePath -> IO ()
+deleteFileIfExists dirPath = do
+  doesDirExist <- SD.doesFileExist dirPath
+  when doesDirExist $ SD.removeFile dirPath
+
+data UserCacheDir
+
+getUserCacheDirPath :: IO (SP.Path' SP.Abs (SP.Dir UserCacheDir))
+getUserCacheDirPath = SD.getXdgDirectory SD.XdgCache "" >>= SP.parseAbsDir
```
sodic, line 26: Good idea getting these out! We can make them even better by swapping `FilePath` with Strong Path types. 

Context: Martin's comment in https://github.com/wasp-lang/wasp/pull/953#discussion_r1081063787 sparked a discussion in the office. We concluded that Strong Path gets clunky when we have to integrate it with regular file system functions and devised a plan to solve it for good: #972. This is a separate issue, but helps explain why it makes sense to change the signatures of these functions.

waspc/cli/src/Wasp/Cli/Command/Utils/File.hs
```diff
@@ -0,0 +1,26 @@
...
+deleteDirectoryIfExists dirPath = do
+  doesDirExist <- SD.doesDirectoryExist dirPath
+  when doesDirExist $ SD.removeDirectoryRecursive dirPath
+
+deleteFileIfExists :: FilePath -> IO ()
+deleteFileIfExists dirPath = do
+  doesDirExist <- SD.doesFileExist dirPath
+  when doesDirExist $ SD.removeFile dirPath
+
+data UserCacheDir
```
sodic, line 23: I recommend moving this type to the top.

waspc/cli/src/Wasp/Cli/Command/Utils/User.hs
```diff
@@ -0,0 +1,16 @@
...
+
+import Data.Maybe (fromJust)
+import qualified StrongPath as SP
+import System.Directory
+  ( getHomeDirectory,
+  )
+
+data HomeDir
+
+getHomeDir :: IO (SP.Path' SP.Abs (SP.Dir HomeDir))
```
sodic, line 15: I'd bundle this thing together with `Wasp.Cli.Command.Utils.File` under something like `Wasp.Cli.Command.FileSystem`, dropping the utils folder entirely.

Also, remember StrongPath conventions i mentioned the other day:  https://github.com/wasp-lang/wasp/pull/953#discussion_r1071593511 (drop the `SP.` :).

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
```diff
@@ -0,0 +1,63 @@
...
+        "",
+        "We will also remove the following files:",
+        "  " ++ waspBinFile',
+        "",
+        "Are you sure you want to continue? [y/N]"
+      ]
+
+  answer <- getLine
+
+  when (answer /= "y") $ die "Aborted."
```
sodic, line 53: Are we sure we don't want to be case insensitive? I don't know what's the usual convention. Btw, the message says `y/N`, is the different capitalization on purpose?

Minor nitpick: I'd also drop the spacing  on lines 50 and 52.

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
```diff
@@ -0,0 +1,63 @@
...
+  cliSendMessageC $ Msg.Success "Uninstalled Wasp"
+
+removeWaspFiles :: IO ()
+removeWaspFiles = do
+  homeDir <- getHomeDir
+  userCacheDir <- getUserCacheDirPath
+
+  let waspInstallationDir' = fromAbsDir $ homeDir </> waspInstallationDir
+      waspBinFile' = fromAbsFile $ homeDir </> waspBinFile
+      waspCacheDir = fromAbsDir $ getWaspCacheDirPath userCacheDir
```
sodic, line 37: The apostrophe naming thing is something I always recommend avoiding. It carries no information and gives us an easy way out of coming up with a more useful name. In this case, I recommend something like:
```haskell
absWaspInstallationDir = ...
absWaspBinFile = ...
absWaspCacheDir = ...
````

File: waspc/cli/src/Wasp/Cli/Command/Uninstall.hs [hunk 18]
```diff
@@ -0,0 +1,63 @@
...
+
+  deleteDirectoryIfExists waspInstallationDir'
+  deleteFileIfExists waspBinFile'
+  deleteDirectoryIfExists waspCacheDir
+
+waspInstallationDir :: Path' (Rel HomeDir) Dir'
+waspInstallationDir = [reldir|.local/share/wasp-lang|]
+
+waspBinFile :: Path' (Rel HomeDir) File'
+waspBinFile = [relfile|.local/bin/wasp|]
```
sodic, line 63: We move this to `Wasp.Cli.Command.FileSystem` (assuming you agree with the other comment that mentions it).

sodic (commented):
Great work with this!

I like having those two helper functions, it's pretty annoying to always have to check whether something exists before deleting it.

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
infomiho: I think it's fine to require exact char `y`, we are deleting files after all. 

Also, I've seen the `y/N` used in some other places which indicates that `N` is the default choice.

Screenshot 2023-01-25 at 09 29 46

waspc/cli/src/Wasp/Cli/Command/Utils/File.hs
infomiho: > Good idea getting these out! We can make them even better by swapping FilePath with Strong Path types.

We are using StrongPath types here already, I'm not sure what else we can do to make it better here? :)

sodic: Swap out `FilePath` with SP Paths:
```haskell
deleteDirectoryIfExists :: (Path a) b (Dir c) -> IO ()
deleteFileIfExists :: (Path a) b (File c) -> IO ()
```
You can even set the signatures to something less polymorphic if you think it makes sense. This makes sure that we can only pass directories to `deleteDirectoryIfExists` and only files to `deleteFileIfExists`.

Also, you have a typo in `abswWspBinFile`.

waspc/cli/src/Wasp/Cli/Command/FileSystem.hs
```diff
@@ -0,0 +1,43 @@
...
+getUserCacheDirPath = SD.getXdgDirectory SD.XdgCache "" >>= SP.parseAbsDir
+
+getHomeDir :: IO (Path' Abs (Dir HomeDir))
+getHomeDir = fromJust . SP.parseAbsDir <$> getHomeDirectory
+
+waspInstallationDir :: Path' (Rel HomeDir) Dir'
+waspInstallationDir = [reldir|.local/share/wasp-lang|]
+
+waspBinFile :: Path' (Rel HomeDir) File'
+waspBinFile = [relfile|.local/bin/wasp|]
```
sodic, line 43: Sorry, I didn't catch this on my first read-through. We have a loose convention for naming relative paths, these two would be `waspInstallationDirInHomeDir` and `waspExecutableInHomeDir`. It's a little Hungarian, but makes stuff easier to read.

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
```diff
@@ -0,0 +1,52 @@
...
+  liftIO removeWaspFiles
+  cliSendMessageC $ Msg.Success "Uninstalled Wasp"
+
+removeWaspFiles :: IO ()
+removeWaspFiles = do
+  homeDir <- getHomeDir
+  userCacheDir <- getUserCacheDirPath
+
+  let absWaspInstallationDir = fromAbsDir $ homeDir </> waspInstallationDir
+      abswWspBinFile = fromAbsFile $ homeDir </> waspBinFile
```
sodic, line 33: Typo in this variable name :)

waspc/cli/src/Wasp/Cli/Command/Utils/File.hs
infomiho: Sorry, I didn't consider the whole file! Now that makes sense, and I've made the needed adjustment. I've went with your suggestion and did the most general typing.

sodic (approved):
All good as far as I'm concerned!

waspc/cli/src/Wasp/Cli/Command/Telemetry/Common.hs
Martinsos: Good one! It is a typical Clean Code recommendation. I wonder if we should put that into out Coding Style guidelines.

waspc/cli/src/Wasp/Cli/Command/Utils/File.hs
Martinsos: @sodic your StrongPath-fu is getting impressive, but still not 100% there :D :
It would be better to limit the "standard" of these paths to `System`, since if you are doing operations on the file system, you want the file paths standard to match the file system standard (I don't want to use Posix path to delete stuff on Win -> it won't work correctly).

So ideally signatures would be:
```hs
deleteDirectoryIfExists :: Path System r (Dir d) -> IO ()
deleteFileIfExists :: Path System r (File f) -> IO ()
```

These can be then shortened a bit with alias `Path'`:
```hs
deleteDirectoryIfExists :: Path' r (Dir d) -> IO ()
deleteFileIfExists :: Path' r (File f) -> IO ()
```

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
```diff
@@ -0,0 +1,52 @@
...
+uninstall :: Command ()
+uninstall = do
+  cliSendMessageC $ Msg.Info "Uninstalling Wasp ..."
+  liftIO removeWaspFiles
+  cliSendMessageC $ Msg.Success "Uninstalled Wasp"
+
+removeWaspFiles :: IO ()
+removeWaspFiles = do
+  homeDir <- getHomeDir
+  userCacheDir <- getUserCacheDirPath
```
Martinsos, line 30: Seeing some inconsistency here: one function has `Path` at the end, the other does not. I am not sure what is better, but let's stick to one approach, at least here.

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
```diff
@@ -0,0 +1,52 @@
...
+
+  let waspInstallationDir = homeDir </> waspInstallationDirInHomeDir
+      waspBinFile = homeDir </> waspExecutableInHomeDir
```
Martinsos, line 33: Inconsistencty: Bin vs exe.

Martinsos (changes requested):
@infomiho we are very close now! I left some fresh comments, check them out.

I also advise resolving existing ones that are solved, or marking them approiatelly with thumbs up and rockets :D.

waspc/cli/src/Wasp/Cli/Command/FileSystem.hs
```diff
@@ -0,0 +1,45 @@
...
+getUserCacheDirPath = SD.getXdgDirectory SD.XdgCache "" >>= SP.parseAbsDir
+
+getHomeDir :: IO (Path' Abs (Dir HomeDir))
+getHomeDir = fromJust . SP.parseAbsDir <$> getHomeDirectory
+
+waspInstallationDirInHomeDir :: Path' (Rel HomeDir) Dir'
+waspInstallationDirInHomeDir = [reldir|.local/share/wasp-lang|]
+
+waspExecutableInHomeDir :: Path' (Rel HomeDir) File'
+waspExecutableInHomeDir = [relfile|.local/bin/wasp|]
```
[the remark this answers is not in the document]
infomiho: Added this issue to capture the intention of using `getXdgDirectory` https://github.com/wasp-lang/wasp/issues/980

waspc/cli/src/Wasp/Cli/Command/FileSystem.hs
```diff
@@ -0,0 +1,45 @@
...
+deleteDirectoryIfExists dirPath = do
+  let dirPathStr = SP.toFilePath dirPath
+  exists <- SD.doesDirectoryExist dirPathStr
+  when exists $ SD.removeDirectoryRecursive dirPathStr
+
+deleteFileIfExists :: Path a b (File c) -> IO ()
+deleteFileIfExists filePath = do
+  let filePathStr = SP.toFilePath filePath
+  exists <- SD.doesFileExist filePathStr
+  when exists $ SD.removeFile filePathStr
```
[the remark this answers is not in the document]
infomiho: The module was already exposed, so no issue there. I've moved the util functions there 👍

waspc/cli/src/Wasp/Cli/Command/Telemetry/Common.hs
```diff
@@ -25,4 +22,7 @@ ensureTelemetryCacheDirExists = do
   return telemetryCacheDirPath
 
 getTelemetryCacheDirPath :: Path' Abs (Dir UserCacheDir) -> Path' Abs (Dir TelemetryCacheDir)
-getTelemetryCacheDirPath userCacheDirPath = userCacheDirPath SP.</> [reldir|wasp/telemetry|]
+getTelemetryCacheDirPath userCacheDirPath = getWaspCacheDirPath userCacheDirPath </> [reldir|telemetry|]
+
+getWaspCacheDirPath :: Path' Abs (Dir UserCacheDir) -> Path' Abs (Dir TelemetryCacheDir)
```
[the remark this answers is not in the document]
infomiho: We already introduced a `FileSystem.hs` in this PR, so it could go there as well.

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
infomiho: Great catch, I'll remove the suffix 👍

waspc/cli/src/Wasp/Cli/Command/FileSystem.hs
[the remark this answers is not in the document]
Martinsos: Ok great! Have you also added comments to installer.sh, to the code?

Martinsos (approved):
@infomiho great, I approved! But I did leave a couple small comments that should be taken care of. Pls take care of those and then you can merge. I approved in advance because I think solving those comments is obvious and you don't need my review again after you get it done.

waspc/cli/src/Wasp/Cli/Command/FileSystem.hs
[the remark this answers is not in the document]
infomiho: I'll do this after I merge this 👍

PR #956, opened 2023-01-17
Generate all JS import statements in Haskell

### LINKED ISSUE #574: Refactor how Wasp generates JavaScript imports
Wasp currently generates JavaScript imports in two distinct ways:
- For generating imports from user code ("ext imports"), we generate the entire import statements inside Haskell: https://github.com/wasp-lang/wasp/blob/main/waspc/src/Wasp/Generator/JsImport.hs
- For (I believe) all other use cases, we use templates. Imports inside templates look something like `import {= importWhat =} from {= importPath =}`, while the values come from the generator. Check the router for an exmple: https://github.com/wasp-lang/wasp/blob/main/waspc/data/Generator/templates/react-app/src/router.js

Perhaps it makes sense to centralize/unify the logic for generating imports. If anything, it's less confusing and more easily extendable (for example, if we ever end up generating something other than JavaScript). Then again, it makes sense to wait until we decide what we're going to do about our templates. Are they staying as-is for the time being, or are we planning to do some refactoring (e.g., moving things to the runtime)?

infomiho:
Fixes #574

This PR refactors four templates to receive full import statements instead of piece by piece:
- router.js [FE]
- operations/_query.js
- operations/_action.js
- operations/index.js

Left to do:
- [x] fix failing tests due to using single quote vs. double quotes

Possible improvements:
- [x] avoid sending `FilePath` and send a more typed path


waspc/data/Generator/templates/server/src/routes/operations/index.js
```diff
@@ -6,13 +6,13 @@ import auth from '../../core/auth.js'
...
 
 {=# operationRoutes =}
-import {= importIdentifier =} from '{= importPath =}'
+{=& importStatement =}
 {=/ operationRoutes =}
 
 const router = express.Router()
 
 {=# operationRoutes =}
-router.post('{= routePath =}', {=# isUsingAuth =} auth, {=/ isUsingAuth =} {= importIdentifier =})
```
infomiho, line 15: Style cleanup: The output went from 
`router.post('bla',  auth,  bla)` to `router.post('bla', auth, bla)`

waspc/src/Wasp/Generator/JsImport.hs
```diff
@@ -8,19 +11,66 @@ import qualified StrongPath as SP
...
+
+mkImportStatement :: AS.ExtImport.ExtImportName -> FilePath -> ImportStatement
+mkImportStatement extImportName path =
+  ImportStatement {_extImportName = extImportName, _path = path, _importAlias = Nothing}
+
+mkImportStatementWithAlias :: AS.ExtImport.ExtImportName -> FilePath -> String -> ImportStatement
+mkImportStatementWithAlias extImportName path importAlias =
+  ImportStatement {_extImportName = extImportName, _path = path, _importAlias = Just importAlias}
+
+mkImportStatementFromRelPath ::
```
infomiho, line 68: Note: this function is based on the old `getJsImportDetailsForExtFnImport` to enable drop in replacement with the new `genJsImport`

waspc/e2e-test/test-outputs/waspMigrate-golden/files.manifest
```diff
@@ -3,7 +3,6 @@ waspMigrate/.wasp/out/.waspchecksums
 waspMigrate/.wasp/out/Dockerfile
 waspMigrate/.wasp/out/db/migrations/migration_lock.toml
 waspMigrate/.wasp/out/db/migrations/no-date-foo/migration.sql
-waspMigrate/.wasp/out/db/package.json
```
infomiho, line 6: @sodic This must mean we dropped having `package.json` in the `db` folder sometime between last time e2e tests were generated and this PR.

Martinsos: Ok we concluded this was caused by a prisma package.json bug

Martinsos (changes requested):
Good start @infomiho !

It took me a bit to find my way around, but I think I get it now.
So there are two main novelties:
1. You added support for import alias.
2. You added support for not only providing paths relative to ext src dir, but also paths relative to the file that does import.

Those were needed to cover all the use cases we had in our code.

I think this is good, but I also think we can improve our code to make it simpler / easier to understand. I left a bunch of comments trying to explain what I think could be clearer, so I hope I managed to convey what I wanted. Give it a try and ask me if something is unclear!

waspc/src/Wasp/Generator/JsImport.hs
```diff
@@ -8,19 +11,66 @@ import qualified StrongPath as SP
...
 import Wasp.Generator.ExternalCodeGenerator.Common (GeneratedExternalCodeDir)
 
-getJsImportDetailsForExtFnImport ::
-  -- | Path to generated external code directory, relative to the directory in which file doing the importing is.
-  Path Posix (Rel a) (Dir GeneratedExternalCodeDir) ->
-  AS.ExtImport.ExtImport ->
+data ImportStatement = ImportStatement
+  { _extImportName :: AS.ExtImport.ExtImportName,
+    _path :: FilePath,
+    _importAlias :: Maybe String
```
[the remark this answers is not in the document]
infomiho: Yep, we had it implicitly because for routes in `router.js` we used page name instead of the actual import name when generating the `import` statements.

waspc/src/Wasp/Generator/JsImport.hs
```diff
@@ -1,26 +1,122 @@
...
+    importStatement = getImportStatement importWhat importFrom
+
+type JsImportStatement = String
+
+getImportStatement :: JsImportWhat -> Path Posix (Rel Dir') File' -> JsImportStatement
+getImportStatement importWhat importPath =
+  "import " ++ importWhat ++ " from '" ++ normalizedPath ++ "'"
+  where
+    filePath = SP.fromRelFileP importPath
+    -- TODO: check if StrongPath can generate "./" prefixed paths
```
infomiho, line 95: Since I've decided to use `[reldir|../|]` to indicate relative path to the `root` folder when generating the `importPath` ([line](https://github.com/wasp-lang/wasp/pull/956/files#diff-883fead5eb518073046ab034da6eb5e9918a6544075416dd8c51ef415db098e5R45)) an edge case popped up. 

When the file importing from `ext-src` is at top level e.g. `import Page from ./ext-src/Page.jsx` and we encode this like `[reldir|./|] </> webAppExtDir </> pathToFile` the `StrongPath` lib renders it like `ext-src/Page.jsx` (without the `./` prefix) which makes it an absolute import in JS 😢 

@Martinsos Is there a better way force the `./` prefix?

Martinsos: Ha yes -> well I think this is ok from the side of StrongPath, because having ./ at the front doesn't really make much sense path-wise -> it is a redundancy, gives no info. So while StrongPath will maintain ../ in front, ./ is normalized away when normalizing the path.

So if you want to make sure to force it, I don't think there is a better option but to check if ./ exists at the start and add it if needed.

waspc/src/Wasp/Generator/WebAppGenerator.hs
```diff
@@ -236,9 +231,6 @@ genIndexJs spec =
       )
   where
     maybeSetupJsFunction = AS.App.Client.setupFn =<< AS.App.client (snd $ getApp spec)
-    maybeSetupJsFnImportDetails = getJsImportDetailsForExtFnImport extClientCodeDirInWebAppSrcDirP <$> maybeSetupJsFunction
+    maybeSetupJsFnImportDetails = getClientJsImport [reldirP|./|] <$> maybeSetupJsFunction
```
Martinsos, line 234: How is it that you dropped the `extClientCodeDirInWebAppSrcDirP`?

waspc/src/Wasp/Generator/WebAppGenerator/RouterGenerator.hs
```diff
@@ -163,21 +159,14 @@ determineRouteTargetComponent spec (_, route) =
...
 
-mkNamedImportExpr :: String -> String -> String
-mkNamedImportExpr identifier alias
-  | identifier == alias = identifier
-  | otherwise = identifier ++ " as " ++ alias
+    importAlias :: String
+    importAlias = fst page
+
+    importStmt :: String
+    (_, importStmt) = getClientJsImportWithAlias [reldirP|./|] pageComponent $ Just importAlias
```
Martinsos, line 172: I would order this in opposite way, so I can read it from top down -> importStmt, then importAlias below it. Since importStmt is more interesting.

It is an idea from CleanCode -> to write the file like a story, top down, with more high level stuff above, and low level stuff below.

waspc/src/Wasp/JsImport.hs
```diff
@@ -0,0 +1,80 @@
...
+    getJsImportData,
+  )
+where
+
+import Data.Data (Data)
+import Data.List (isPrefixOf)
+import StrongPath (Dir', File', Path, Posix, Rel)
+import qualified StrongPath as SP
+
+data JsImport = JsImport
```
Martinsos, line 18: Ok, so this is general representation of JS import.

I would put a comment here explaining that, plus explaining that it is simplified and has can import only one name (be it default or listed) + potentially alias.
The reason for that is that we don't need more at the moment.

waspc/src/Wasp/JsImport.hs
```diff
@@ -0,0 +1,80 @@
...
+    JsImportField Identifier
+  deriving (Show, Eq, Data)
+
+type JsImportIdentifier = String
+
+type JsImportWhat = String
+
+type JsImportStatement = String
+
+type JsImportData = (JsImportIdentifier, JsImportStatement)
```
Martinsos, line 47: Ha this name is really hm. In general, names with `data`, `info` and similar in them are a bad pattern, since they don't really mean anything.
In this case, a better name would probably be `JsImportIdentifierAndStmt`. Or maybe even better to avoid having a `type` here if we are having hard time finding a name for it.

waspc/src/Wasp/JsImport.hs
```diff
@@ -0,0 +1,80 @@
...
+
+getJsImportData :: JsImportPath -> JsImportName -> Maybe JsImportAlias -> JsImportData
```
Martinsos, line 49: Why don't we instead have function `getJsImportStmtAndIdentifier :: JsImport -> (JsImportIdentifier, JsImportStatement)`

waspc/src/Wasp/JsImport.hs
```diff
@@ -0,0 +1,80 @@
...
+data JsImportName
+  = -- | Represents external imports like @import Identifier from "file.js"@
+    JsImportModule Identifier
+  | -- | Represents external imports like @import { Identifier } from "file.js"@
+    JsImportField Identifier
+  deriving (Show, Eq, Data)
+
+type JsImportIdentifier = String
+
+type JsImportWhat = String
```
Martinsos, line 43: I know I gave it this name long time ago but looking at it now I think it is a pretty bad name :D. I am not sure how to better name it though.

waspc/src/Wasp/JsImport.hs
```diff
@@ -0,0 +1,80 @@
...
+type JsImportAlias = String
+
+data JsImportName
+  = -- | Represents external imports like @import Identifier from "file.js"@
+    JsImportModule Identifier
+  | -- | Represents external imports like @import { Identifier } from "file.js"@
+    JsImportField Identifier
+  deriving (Show, Eq, Data)
+
+type JsImportIdentifier = String
```
Martinsos, line 41: We have this, but we also have Identifier above. Maybe let's stick to just Identifier?

waspc/src/Wasp/JsImport.hs
Martinsos, line 49: I am just somehow tempted to take this one, and two functions below: getImportStatement and getFirstPartOfJsImport and put them all into one function with stuff in `where`, because having this guys standalone is a bit weird. But ok, we can leave it like this (I know you copied this from before, I wrote it like this before).

Martinsos: Btw do we have tests for this? If not, we should add some, this is a nice thing to test!

waspc/src/Wasp/Generator/JsImport.hs
```diff
@@ -7,68 +7,64 @@ module Wasp.Generator.JsImport
...
   )
 
-getServerJsImport :: Path Posix (Rel GeneratedExternalCodeDir) Dir' -> EI.ExtImport -> JsImportData
-getServerJsImport relPathToRoot extImport = getServerJsImportWithAlias relPathToRoot extImport Nothing
+type RelPathToRoot = Path Posix (Rel GeneratedExternalCodeDir) Dir'
 
-getClientJsImport :: Path Posix (Rel GeneratedExternalCodeDir) Dir' -> EI.ExtImport -> JsImportData
-getClientJsImport relPathToRoot extImport = getClientJsImportWithAlias relPathToRoot extImport Nothing
+-- | Generates JS import statement based on generated external code dir (server)
+getServerJsImport :: RelPathToRoot -> EI.ExtImport -> JsImportData
```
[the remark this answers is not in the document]
Martinsos: Although I see now it returns JsImportData, not just JsImport.
So I would consider naming it `jsImportStmtAndIdentifierFromExtImport`, or going with `jsImportFromExtImport` but then have it return JsImport and then additionaly use `getJsImportStmtAndIdentifier` on top of it.

Martinsos (commented):
@infomiho getting closer! This is looking better now but I still left some comments. They are mostly focused on naming -> ensuring that names are in sync with what is actually function doing / returning. I believe forcing ourselves to give very precise names also forces us to write better code -> it uncovers weak points in our abstractions, forcing us to improve them (e.g. if you can't name something well, it means you haven't modeled it well), while also making our code more readable. Big part of coding is just finding right names I think.

waspc/src/Wasp/Generator/WebAppGenerator.hs
infomiho: We construct the path to the `web-app/ext-src` inside of the `getClientJsImport` so we don't need it anymore. 

https://github.com/wasp-lang/wasp/pull/956/files#diff-883fead5eb518073046ab034da6eb5e9918a6544075416dd8c51ef415db098e5R15

waspc/src/Wasp/Generator/JsImport.hs
```diff
@@ -7,68 +7,64 @@ module Wasp.Generator.JsImport
...
+  ( JsImportAlias,
     JsImportData,
     JsImportName (JsImportField, JsImportModule),
+    JsImportPath,
     getJsImportData,
   )
 
-getServerJsImport :: Path Posix (Rel GeneratedExternalCodeDir) Dir' -> EI.ExtImport -> JsImportData
-getServerJsImport relPathToRoot extImport = getServerJsImportWithAlias relPathToRoot extImport Nothing
+type RelPathToRoot = Path Posix (Rel GeneratedExternalCodeDir) Dir'
```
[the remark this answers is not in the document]
infomiho: It's a relative dir to any of our generated external code dir, I'll make the name more informative 👍

waspc/src/Wasp/Generator/ServerGenerator/OperationsRoutesG.hs
```diff
@@ -65,10 +67,18 @@ genOperationRoute spec operation tmplFile = return $ C.mkTmplFdWithDstAndData tm
...
 
     operationImportPath =
-      C.toESModulesImportPath $
-        SP.fromRelFileP $
-          relPosixPathFromOperationsRoutesDirToSrcDir
-            </> fromJust (SP.relFileToPosix $ operationFileInSrcDir operation)
+      relPosixPathFromOperationsRoutesDirToSrcDir
+        </> fromJust (SP.relFileToPosix $ operationFileInSrcDir operation)
+
+    operationEsImportPath =
```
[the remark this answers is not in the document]
infomiho: It was based on the function we introduced recently `C.toESModulesImportPath`, so I wanted to singal that it's "ES module ready" import.

waspc/src/Wasp/JsImport.hs
infomiho: I've kept only the `JsImportIdentifier`

waspc/src/Wasp/JsImport.hs
infomiho: Looking at JS AST tools, they call this part "importClause". I'll go with that 👍

Screenshot 2023-01-31 at 16 00 10

waspc/src/Wasp/JsImport.hs
infomiho: I've gone with using the tuple everywhere explictly :)

waspc/src/Wasp/JsImport.hs
infomiho: We don't have tests atm, I'll add them 👍

waspc/src/Wasp/Generator/JsImport.hs
```diff
@@ -1,26 +1,70 @@
...
+    getJsImportData,
+  )
+
+type RelPathToRoot = Path Posix (Rel GeneratedExternalCodeDir) Dir'
+
+-- | Generates JS import statement based on generated external code dir (server)
+getServerJsImport :: RelPathToRoot -> EI.ExtImport -> JsImportData
+getServerJsImport relPathToRoot extImport = getServerJsImportWithAlias relPathToRoot extImport Nothing
+
+getServerJsImportWithAlias ::
```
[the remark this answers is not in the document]
infomiho: Great suggestion! I've went with `applyJsImportAlias` and this made things much cleaner. It enabled me to only add the alias in one case and not have duplicate function names.

waspc/src/Wasp/Generator/ServerGenerator/OperationsRoutesG.hs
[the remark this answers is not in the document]
Martinsos: Ha hm ok! I wonder if we should still call it just JsImportPath, I feel like ES is just too much out of nowhere. But up to you!

waspc/src/Wasp/Generator/JsImport.hs
```diff
@@ -1,26 +1,34 @@
...
-  Path Posix (Rel a) (Dir GeneratedExternalCodeDir) ->
-  AS.ExtImport.ExtImport ->
-  -- | (importIdentifier, importStmt)
-  --   - importIdentifier -> Identifier via which you can access ext js function after you import it with importStmt.
-  --   - importStmt -> Javascript import statement via which you should do the import.
-  (String, String)
-getJsImportDetailsForExtFnImport relPosixPathToExtCodeDir extImport = (importIdentifier, importStmt)
+-- | Returns a getter function that can be used to get a JsImport from an ExtImport
+mkJsImportGetterFromExtSrcDir ::
+  Path Posix (Rel ()) (Dir GeneratedExternalCodeDir) ->
```
Martinsos, line 21: Since type doesn't say what this Path is relative to (and that is ok), I would explain in comments what it is relative to.

waspc/src/Wasp/Generator/ServerGenerator/JsImport.hs
```diff
@@ -0,0 +1,30 @@
...
+
+-- | Wrapper function to avoid needing to import getJsImportStmtAndIdentifier from Wasp.JsImport
+--  in most cases.
+getJsImportStmtAndIdentifier ::
+  RelDirToExternalCodeDir ->
+  EI.ExtImport ->
+  (JsImportStatement, JsImportIdentifier)
+getJsImportStmtAndIdentifier relDirToExternalCodeDir = JI.getJsImportStmtAndIdentifier . getJsImport relDirToExternalCodeDir
+
+-- | Generates a JsImport from an ExtImport and relative to server ext code dir.
```
Martinsos, line 23: Redundant comment, types are saying the exact same thing (which is super nice).

waspc/src/Wasp/Generator/ServerGenerator/JsImport.hs
```diff
@@ -0,0 +1,30 @@
...
+getJsImport ::
```
Martinsos, line 24: I would potentially rename this one to `jsImportFromExtImport`. Or `extImportToJsImport`.

waspc/src/Wasp/Generator/JsImport.hs
```diff
@@ -1,26 +1,34 @@
...
-  -- | Path to generated external code directory, relative to the directory in which file doing the importing is.
-  Path Posix (Rel a) (Dir GeneratedExternalCodeDir) ->
-  AS.ExtImport.ExtImport ->
-  -- | (importIdentifier, importStmt)
-  --   - importIdentifier -> Identifier via which you can access ext js function after you import it with importStmt.
-  --   - importStmt -> Javascript import statement via which you should do the import.
-  (String, String)
-getJsImportDetailsForExtFnImport relPosixPathToExtCodeDir extImport = (importIdentifier, importStmt)
+-- | Returns a getter function that can be used to get a JsImport from an ExtImport
+mkJsImportGetterFromExtSrcDir ::
```
Martinsos, line 20: I would rename this one to `jsImportFromExtImport` or `extImportToJsImport`. Probably the first one. And would keep its name the same as the functions you have in ServerGenerator and ClientGenerator for doing this same thing. And then I would also remove those extra parentheses in the type definition, although I told you earlier those can make sense -> but I don't think they make sense here. Why? Because instead of looking at it as producing another function, which could be valid but I don't think is the best option in this case, you can look at it as a more general version of those functions you have in Server and Client Generators. Let's call this one A, and we can call those B and C. So B and C are the same function, just in different context, which enables them to just call the A with only first argument, which comes from their context, and the rest are propagated. So B and C are just specializations of A. And therefore it is even ok that they all have the same name, because they have the additional context of module they are imported from.
I think this is the cleanest way do it. This way, function A doesn't even have to know that B and C exist. While it kind of does in your case -> it doesn't know about them particularly, but it does have this notion that it is producing another function that will be used somewhere else, while it doesn't need to know that. Ok I expressed this last part relatively clumsily so take it with a grain of salt.

waspc/src/Wasp/Generator/ServerGenerator/JsImport.hs
```diff
@@ -0,0 +1,30 @@
...
+import Wasp.Generator.ServerGenerator.ExternalCodeGenerator (extServerCodeDirInServerSrcDir)
+import Wasp.JsImport
+  ( JsImport,
+    JsImportIdentifier,
+    JsImportStatement,
+  )
+import qualified Wasp.JsImport as JI
+
+-- | Wrapper function to avoid needing to import getJsImportStmtAndIdentifier from Wasp.JsImport
+--  in most cases.
```
Martinsos, line 16: Hm this comment confuses me, I would just remove it, I think this is clear from the code.

Martinsos (changes requested):

waspc/src/Wasp/JsImport.hs
```diff
@@ -0,0 +1,87 @@
...
+-- | Represents the left side of the import statement e.g. { Name } or { Name as Alias } or NameForDefault
+type JsImportClause = String
+
+type JsImportStatement = String
+
+getJsImport :: JsImportPath -> JsImportName -> JsImport
+getJsImport importPath importName = JsImport importPath importName Nothing
+
+applyJsImportAlias :: Maybe JsImportAlias -> JsImport -> JsImport
+applyJsImportAlias importAlias (JsImport importPath importName _) = JsImport importPath importName importAlias
```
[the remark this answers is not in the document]
infomiho: Nice, I knew about that update syntax but never used it much 👍

waspc/src/Wasp/Generator/ServerGenerator/JsImport.hs
infomiho: I've went with `extImportToJsImport` since it reads nicer to me. Conversion from `ExtImport` to `JsImport` :

waspc/src/Wasp/Generator/JsImport.hs
infomiho: > I would rename this one to `jsImportFromExtImport` or `extImportToJsImport. Probably the first one. And would keep its name the same as the functions you have in ServerGenerator and ClientGenerator for doing this same thing. 
>
> And then I would also remove those extra parentheses in the type definition, although I told you earlier those can make sense -> but I don't think they make sense here. Why? Because instead of looking at it as producing another function, which could be valid but I don't think is the best option in this case, you can look at it as a more general version of those functions you have in Server and Client Generators. 
> 
> <details>
> <summary>Extra info</summary>
> Let's call this one A, and we can call those B and C. So B and C are the same function, just in different context, which enables them to just call the A with only first argument, which comes from their context, and the rest are propagated. So B and C are just specializations of A. And therefore it is even ok that they all have the same name, because they have the additional context of module they are imported from.
> </details>
> 
> I think this is the cleanest way do it. This way, function A doesn't even have to know that B and C exist. While it kind of does in your case -> it doesn't know about them particularly, but it does have this notion that it is producing another function that will be used somewhere else, while it doesn't need to know that. Ok I expressed this last part relatively clumsily so take it with a grain of salt.

Makes sense 👍 general function exists in `Wasp.Generator` context and specialized functions in `Wasp.Generator.ServerGenerator` and `Wasp.Generator.WebAppGenerator` contexts.

waspc/src/Wasp/Generator/JsImport.hs
```diff
@@ -1,26 +1,34 @@
...
+  Path Posix (Rel ()) (Dir GeneratedExternalCodeDir) ->
+  ( RelDirToExternalCodeDir ->
+    EI.ExtImport ->
+    JsImport
+  )
+mkJsImportGetterFromExtSrcDir rootDir relDirToExternalCodeDir extImport = getJsImport importPath importName
```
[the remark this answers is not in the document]
infomiho: I've named it `pathToExternalCodeDir` since the `relPathFromImportingJsFileToExternalCodeDir` is something we get when we combine `pathToExternalCodeDir` and the second argument `relDirToExternalCodeDir`

waspc/src/Wasp/Generator/ServerGenerator/JsImport.hs
Martinsos: Ok, good job! Just a subjective reasoning why I prefer starting with the produced thing instead of the "ingredient", due to two reasons:
1. When reading code I more often care about what is function producing, in cases like this.
2. I might have multiple ways to create JsImport, so in that case I find it nice that all those functions start in a same way, e.g. with `makeJsImport` and then one can continue with `fromX`, another can be `fromY`, another can have no suffix because it creates it from scratch, ... .

waspc/src/Wasp/JsImport.hs
```diff
@@ -0,0 +1,89 @@
...
+-- | Represents a JS import data type that can be used to generate import statements
+--   in generated code. It doesn't fully support all types of JS imports (multiple imports)
+--   but this is enough for our current use case.
+data JsImport = JsImport
+  { -- | Path from which we are importing.
+    _path :: JsImportPath,
+    -- | What is being imported. NOTE: We don't currenly support multiple names in one statement,
+    --   that's why it's "name" and not "names".
+    _name :: JsImportName,
+    -- | Alias for the imported name
```
Martinsos, line 30:
```suggestion
    -- | Alias for the imported name.
```

waspc/test/Generator/JsImportTest.hs
```diff
@@ -0,0 +1,26 @@
...
+            JI._importAlias = Nothing
+          }
+  where
+    pathToExtCodeDir = [SP.reldirP|ext-src|]
+    relDirToExternalCodeDir = [SP.reldirP|../|]
+    extImport =
+      ExtImport
+        { name = ExtImportModule "test",
+          path = [SP.relfileP|folder/test.js|]
+        }
```
Martinsos, line 26: It is slightly weird to put this in `where` because it applies to whole `spec_GeneratorJsImportTest` function, so if you add more tests and describes and so on, it becomes too widely scoped, in the sense of that you have defined test setup in a place that is in scope of other tests also.

So instead, I would move these to be defined via `let` in either the specific test or specific describe, however you estimate they are most likely to be used in the future. In this case probably just best to put it into that specific test, and it can be taken out a level above if needed in the future.

waspc/test/JsImportTest.hs
```diff
@@ -0,0 +1,53 @@
...
+  describe "applyJsImportAlias" $ do
+    it "applies alias to JsImport" $ do
+      applyJsImportAlias
+        (Just "alias")
+        (makeJsImport pathToJsFile (JsImportModule "test"))
+        `shouldBe` JsImport
+          { _path = pathToJsFile,
+            _name = JsImportModule "test",
+            _importAlias = Just "alias"
+          }
```
Martinsos, line 27: What about:

```suggestion
      let jsImport = makeJsImport pathToJsFile (JsImportModule "test")
      applyJsImportAlias (Just "testAlias") jsImport
        `shouldBe` jsImport { _importAlias = Just "testAlias" }
```

This way you are testing exactly what you wanted to test. Yes, this comes down to replicating implementation, but that is because implementation is trivial. When implementation is super trivial, one might wonder is test even useful / needed? The only thing that test does is just repeat implementation. This does make it more robust to unexpected changes -> if you specify in two places how something works, you need to change implementation in both places, which means it is resistant to mistake / accidental change in just one place. So I would argue it can be ok to do a test even for such trivial functions, although I would probably also be ok with even dropping the test.

Martinsos: Btw, one way to make this test better is to actually use property testing (QuickCheck).
I think this is a good opportunity to go for it, give it a try. You would basically define Arbitrary class for JSImport and somewhat modify this test then to test for many different possible js imports.

waspc/test/JsImportTest.hs
```diff
@@ -0,0 +1,53 @@
...
+  describe "generateJsImportStatement" $ do
```
Martinsos, line 28: generateJsImportStatement -> why this name? We don't have any function with this name, do we?

waspc/test/JsImportTest.hs
```diff
@@ -0,0 +1,53 @@
...
+    it "generates import statement for default import" $ do
```
Martinsos, line 29: what about identifier, we are also testing that aren't we?

waspc/test/JsImportTest.hs
Martinsos, line 28: I think there are some opportunities to make the tests below easier to read / more maintainable. Some ideas:
1. All `it`s share same prefix: `generates import statement` -> that should be extracted into common `describe`.
2. Extracting some more of common logic. You did some of it with `pathToJsFile`, but you stopped there and I think you can go a bit further, I commented more on it below.

Maybe there are some other opportunities also.
We shouldn't overdo it, but still, testing code should also be subject to most of the rules we use in the normal code, which means keeping it DRY and SOLID and whatever other cool acronyms :D: in other words, keeping it clean. I think it can be ok to sometimes do it a bit less than we would do in the tested code though, but not too much.
The thing is, test code also needs to be maintained and readable -> if it is not, as a result you get that people don't want to change normal code because they don't want to deal with test code, or it stops being clear from test code what was actually the behaviour you wanted to test, ... .

waspc/test/JsImportTest.hs
```diff
@@ -0,0 +1,53 @@
...
+      getJsImportStmtAndIdentifier
+        (makeJsImport pathToJsFile (JsImportModule "test"))
+        `shouldBe` ("import test from './ext-src/folder/test.js'", "test")
```
Martinsos, line 32: `./ext-src/folder/test.js` here is repeated in many places below. It would be good to remove that duplication.
Also, it is connected with `pathToJsFile`, but that connection is not clear.

waspc/test/JsImportTest.hs
```diff
@@ -0,0 +1,53 @@
...
+        `shouldBe` ("import { test } from './ext-src/folder/test.js'", "test")
+    it "generates import statement for named import with alias" $ do
+      getJsImportStmtAndIdentifier
+        ( applyJsImportAlias
+            (Just "alias")
+            (makeJsImport pathToJsFile (JsImportField "test"))
+        )
+        `shouldBe` ("import { test as alias } from './ext-src/folder/test.js'", "alias")
+  where
+    pathToJsFile :: Path Posix (Rel Dir') File'
```
Martinsos, line 52: Nit.: Maybe `testImportPath` or `mockImportPath` or `testJsFilePathInExtSrc` is better.

Martinsos (changes requested):
We are super close, all the structure is there, but still there are a few things to look at, mostly just making core more readable / explicit.

waspc/src/Wasp/Generator/JsImport.hs
```diff
@@ -1,26 +1,33 @@
...
-import qualified Wasp.AppSpec.ExtImport as AS.ExtImport
+import qualified Wasp.AppSpec.ExtImport as EI
 import Wasp.Generator.ExternalCodeGenerator.Common (GeneratedExternalCodeDir)
+import Wasp.JsImport
+  ( JsImport,
+    JsImportName (JsImportField, JsImportModule),
+    makeJsImport,
+  )
+
+type RelDirToExternalCodeDir = Path Posix (Rel GeneratedExternalCodeDir) Dir'
```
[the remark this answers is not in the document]
infomiho: I'd like to keep the type since it's used in 4 places and helps not to drag around a more complex definition. 

The concept is "relative directory path from import location to external code dir" e.g. "../../" which says the import is happening two levels below external code dir. 

`RelDirFromImportLocationToExternalCodeDir` captures that for me.

waspc/src/Wasp/Generator/JsImport.hs
```diff
@@ -1,26 +1,33 @@
...
-    importFrom = "./" ++ SP.fromRelFileP (relPosixPathToExtCodeDir </> SP.castRel (AS.ExtImport.path extImport))
-    (importIdentifier, importWhat) =
-      case AS.ExtImport.name extImport of
-        AS.ExtImport.ExtImportModule defaultImport -> (defaultImport, defaultImport)
-        AS.ExtImport.ExtImportField namedImport -> (namedImport, "{ " ++ namedImport ++ " }")
+    userDefinedPath = SP.castRel $ EI.path extImport
+    importName = extImportNameToJsImportName $ EI.name extImport
+    importPath = SP.castRel $ relDirToExternalCodeDir </> pathToExternalCodeDir </> userDefinedPath
+
+extImportNameToJsImportName :: EI.ExtImportName -> JsImportName
```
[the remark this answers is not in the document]
infomiho: Makes sense, it's not used independently, I'll move it inside 👍

waspc/src/Wasp/Generator/ServerGenerator/JsImport.hs
```diff
@@ -0,0 +1,28 @@
...
+import Wasp.Generator.ServerGenerator.ExternalCodeGenerator (extServerCodeDirInServerSrcDir)
+import Wasp.JsImport
+  ( JsImport,
+    JsImportIdentifier,
+    JsImportStatement,
+  )
+import qualified Wasp.JsImport as JI
+
+getJsImportStmtAndIdentifier ::
+  RelDirToExternalCodeDir ->
```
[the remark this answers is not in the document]
infomiho: I've updated the type to `PathFromImportLocationToExtCodeDir`, it reads a bit better now, but I agree it's still little hard to know out of the box what to pass in.

I'll provide an example above the type definition 👍

waspc/test/Generator/JsImportTest.hs
infomiho: Makes sense 👍 I went with `where` instinctively  but `let` makes it easier to scope it properly

waspc/test/JsImportTest.hs
infomiho: I'm disliking test results that have *any* logic or computation inside. That's why I've went with the verbose construction of `JsImport` and repeated result strings in many places in the `getImportStmtAndIdentifier` tests.

infomiho: We talked about this offline, and we agreed that using a setter it's okay to construct the expected data. The aim is to keep the tests maintaintable and making the code, cleaner.

waspc/test/JsImportTest.hs
infomiho: Github Copilot got creative with its recommandation and I didn't notice it, sorry about that 👍

waspc/test/JsImportTest.hs
infomiho: We agreed not to do property testing here since it's hard to define the properites to test and we already test all of the behaviours we need.

waspc/src/Wasp/Generator/JsImport.hs
```diff
@@ -1,26 +1,33 @@
...
-  -- | Path to generated external code directory, relative to the directory in which file doing the importing is.
-  Path Posix (Rel a) (Dir GeneratedExternalCodeDir) ->
-  AS.ExtImport.ExtImport ->
-  -- | (importIdentifier, importStmt)
-  --   - importIdentifier -> Identifier via which you can access ext js function after you import it with importStmt.
-  --   - importStmt -> Javascript import statement via which you should do the import.
-  (String, String)
-getJsImportDetailsForExtFnImport relPosixPathToExtCodeDir extImport = (importIdentifier, importStmt)
+extImportToJsImport ::
+  -- | Path to directory with external code, relative to generated code source dir (server or web app).
```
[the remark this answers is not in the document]
infomiho: We added a class `GeneratedSourceCodeDir` so we can constrain the `Rel` to a polymorphic type `GeneratedSourceCodeDir  d => Rel d` and we defined instances for that class for `ServerSourceDir` and `WebAppSourceDir`. This way we control precisely which dirs can be used with the  `extImportToJsImport` function.

waspc/src/Wasp/Generator/ServerGenerator/OperationsRoutesG.hs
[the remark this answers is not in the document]
Martinsos: bump

waspc/src/Wasp/Generator/ServerGenerator/JsImport.hs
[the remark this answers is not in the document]
Martinsos: Awesome, that name is already much better!

Martinsos (changes requested):
@infomiho nice, this is know a really nice PR! Amazing stuff you did here with types and nice names :D!

We are 99% there, the only thing left is small stuff. I still didn't approve because there are couple of thigs that are not 100% clear, but once you take care of everything that is left I think it is very likely we can just merge it and that is it.

waspc/src/Wasp/Generator/JsImport.hs
```diff
@@ -1,26 +1,34 @@
...
   )
 where
 
-import StrongPath (Dir, Path, Posix, Rel, (</>))
+import StrongPath (Dir, Dir', Path, Posix, Rel, (</>))
 import qualified StrongPath as SP
-import qualified Wasp.AppSpec.ExtImport as AS.ExtImport
+import qualified Wasp.AppSpec.ExtImport as EI
+import Wasp.Generator.Common (GeneratedCodeSourceDir)
 import Wasp.Generator.ExternalCodeGenerator.Common (GeneratedExternalCodeDir)
```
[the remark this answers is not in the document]
infomiho: I'll add this to my TODO to research and I'll create a PR 👍

waspc/src/Wasp/Generator/ServerGenerator/JobGenerator.hs
```diff
@@ -101,9 +97,6 @@ genAllJobImports spec =
         [ "name" .= jobName
         ]
 
-relPosixPathFromJobFileToExtSrcDir :: Path Posix (Rel ServerSrcDir) (Dir GeneratedExternalCodeDir)
```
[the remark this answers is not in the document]
infomiho: It's not really needed IMHO. 

When you use `getJsImportStmtAndIdentifier` from the `ServerGenerator` it's generating an import statement with a path relative to `GeneratedExternalCodeDir` in `ServerSrcDir` anyways.

That's also encoded in the first argument's type `PathFromImportLocationToExtCodeDir` "just tell me a relative path".

So, this whole definition is not necessary, we could only encode it with a definition like `relPosixPathFromJobFileToExtSrcDir = [reldirP|../|]` but it feels kinda redundant since that's what the first argument encodes anyways. Let me know what do you think.

waspc/src/Wasp/Generator/ServerGenerator/OperationsG.hs
```diff
@@ -121,10 +119,6 @@ operationFileInSrcDir :: AS.Operation.Operation -> Path' (Rel C.ServerSrcDir) Fi
 operationFileInSrcDir (AS.Operation.QueryOp name _) = queryFileInSrcDir name
 operationFileInSrcDir (AS.Operation.ActionOp name _) = actionFileInSrcDir name
 
-relPosixPathFromOperationFileToExtSrcDir :: Path Posix (Rel Dir') (Dir GeneratedExternalCodeDir)
```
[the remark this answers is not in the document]
infomiho: Wrote a comment here: https://github.com/wasp-lang/wasp/pull/956/files#r1103689737

waspc/src/Wasp/Generator/WebAppGenerator.hs
```diff
@@ -236,9 +231,6 @@ genIndexJs spec =
       )
   where
     maybeSetupJsFunction = AS.App.Client.setupFn =<< AS.App.client (snd $ getApp spec)
-    maybeSetupJsFnImportDetails = getJsImportDetailsForExtFnImport extClientCodeDirInWebAppSrcDirP <$> maybeSetupJsFunction
-    (maybeSetupJsFnImportIdentifier, maybeSetupJsFnImportStmt) =
+    maybeSetupJsFnImportDetails = getJsImportStmtAndIdentifier [reldirP|./|] <$> maybeSetupJsFunction
```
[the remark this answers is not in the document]
infomiho: The current `getJsImportStmtAndIdentifier` implementation is based around using relative paths and `Extimport` data. My intention was to avoid constructing the relative path to the `ext-src` every time from scratch. If you are using the `getJsImportStmtAndIdentifier` from `WebAppGenerator`, it will know where to import the JS file from.

Again, smilar comment: https://github.com/wasp-lang/wasp/pull/956/files#r1103689737

waspc/src/Wasp/Generator/WebAppGenerator/RouterGenerator.hs
```diff
@@ -24,7 +22,8 @@ import Wasp.Generator.Monad (Generator)
 import Wasp.Generator.WebAppGenerator.Common (asTmplFile, asWebAppSrcFile)
 import qualified Wasp.Generator.WebAppGenerator.Common as C
 import Wasp.Generator.WebAppGenerator.ExternalAuthG (ExternalAuthInfo (..), frontendLoginUrl, gitHubAuthInfo, googleAuthInfo, serverOauthRedirectHandlerUrl)
-import Wasp.Generator.WebAppGenerator.ExternalCodeGenerator (extClientCodeDirInWebAppSrcDir)
```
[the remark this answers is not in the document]
infomiho: The `getJsImportStmtAndIdentifier` imports `extClientCodeDirInWebAppSrcDir` and it's a "proxy" towards that value.

It uses your relative path like `../` and joins it with the path of `extClientCodeDirInWebAppSrcDir`.

waspc/src/Wasp/Generator/ServerGenerator/OperationsRoutesG.hs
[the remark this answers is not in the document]
infomiho: I've updated the respective variables to `pathToOperationFile` and `operationImportPath`

waspc/src/Wasp/Generator/JsImport.hs
```diff
@@ -15,19 +15,19 @@ import Wasp.JsImport
...
-type PathFromImportLocationToExtCodeDir = Path Posix (Rel GeneratedExternalCodeDir) Dir'
+type PathFromImportLocationToSrcDir d = Path Posix (Rel ()) (Dir d)
 
 extImportToJsImport ::
   GeneratedSrcDir d =>
   Path Posix (Rel d) (Dir GeneratedExternalCodeDir) ->
-  Path Posix (Rel GeneratedExternalCodeDir) Dir' ->
+  PathFromImportLocationToSrcDir d ->
   EI.ExtImport ->
   JsImport
```
Martinsos, line 25: Just a quick idea, no need to do it but wanted to offer it in case you like it:
```hs
type ImportLocation = ()

extImportToJsImport ::
  GeneratedSrcDir srcDir =>
  Path Posix (Rel srcDir) (Dir GeneratedExternalCodeDir) ->
  Path Posix (Rel ImportLocation) (Dir srcDir)
  EI.ExtImport ->
  JsImport
```

it gives some uniformity to types used in this signature (it is a bit weird when one is behind type alias and another on is not) but still captures the info that we are dealing with import location here.

What was also a bit weird with the old `type` was that it says it goes to SrcDir but doesn't enforce constraint on its own so you could use it somewhere and not put constraint on it.

Martinsos (approved):

waspc/src/Wasp/Generator/JsImport.hs
infomiho: I'll address this suggestion in a separate PR 👍

PR #966, opened 2023-01-23
Fix for misaligned contents of auth buttons.

matijaSos:
Screenshot 2023-01-23 at 12 10 04

Screenshot 2023-01-23 at 12 09 17

Now it looks the same both with and without tailwind. Tailwind still affects e.g. button fonts via its global settings, so that's something we'll need to decide what to do about in the future.


shayneczyzewski (approved):
Awesome job @matijaSos, thanks again for the help here! Confirmed it looks great both with and without TW, and that the div container is clickable. Wooohoooo! 🚀 🎨

maksim36ua:
Great improvement, waiting for it to be included in the next version of Wasp!

PR #977, opened 2023-01-25
Password field is required only if `usernameAndPassword` auth is used

### LINKED ISSUE #876: Do not require `password` on `userEntity` when only social logins are used
Right now, our social login system depends on the `userEntity` entity. That entity has a `password` field requirement. We should relax this requirement if you are not using `usernameAndPassword`.

infomiho:
Fixes #876

It's a template modification:
- don't the generate `password` field if not needed
- don't validate `password` field if not needed

Left to do:
- [ ] investigate what happens when existing users would upgrade to new Wasp version
    - is it a breaking change and what can we do about it
- [ ] fix e2e tests


infomiho:
If the user had `password` previously already setup, he would get this error from Prisma with this change when signing up with Google:
```
Argument password for data.password is missing.
```

shayneczyzewski (commented):
Hey @infomiho, great job on this so far! It looks pretty minimal and clean so far, which is great.

I'm sad to report (since I introduced it, sorry!), there is one more bit of coupling between social login and username and password: the username.

We actually try to help the user and set the `username` for them by providing default `getUserFieldsFn` implementations for each provider:
- https://github.com/wasp-lang/wasp/blob/2182b9088ddc45d9c7197f09b8913f8eb9dce54e/waspc/data/Generator/templates/server/src/routes/auth/passport/google/defaults.js
- https://github.com/wasp-lang/wasp/blob/2182b9088ddc45d9c7197f09b8913f8eb9dce54e/waspc/data/Generator/templates/server/src/routes/auth/passport/github/defaults.js

They both currently make use of `generateAvailableDictionaryUsername` to help set an untaken randomized username.

So, I think we have a few options here:
(1) we just make `username` required on the user entity when they use social login (not ideal, but fast and non-breaking). We can add the check here: https://github.com/wasp-lang/wasp/blob/26f45942f8c76c84c69c1518362bffad312839ea/waspc/src/Wasp/AppSpec/Valid.hs#L156-L160
(2) we can make the coupling explicit, and add a flag for these to that say something like `generateUsername: <bool>`, which would then make use of the existing defaults or not depending on what is set (plus add a check to Valid), or
(3) we change the defaults to be a no-op, and just provide the existing implementation as an example in the docs _if_ they want to generate a username. But they would have to make it an override.

I am leaning toward (3), but that is definitely a breaking change, which is fine. 🤔 What do you think? 

Also, whatever we decide, we will need to update these docs: https://wasp-lang.dev/docs/language/features#social-login-providers-oauth-20---google-github where they talk about the user model requirements being both username and password.

infomiho:
@shayneczyzewski I thinking there are few user types here:
1. User is using Wasp for the first time, he just wants "nice login with Google" out of the box `new-user 🟢`
2. User moved beyond trying out Wasp and is now serious about making the project in Wasp `pro-user 🔴`
3. Our existing users who already have the existing implementation `existing-user 🔵`

For `new-user 🟢` type of users, the less work they have to do, the more delightful the experience. Set `auth: google` and get something nice looking with profile picture and name. For example, a solution would be to have some smart defaults. 

For `pro-user 🔴` the best is giving them total freedom to customize their user and social entities to their liking. All of the fields are optional.

For `existing-user 🔵` ideally nothing breaks with the upgrade. I understand that pre-1.0 software will have changes in the API. But we should either inform them with a nice error message what they need to change or try to be backwards compatible.

I'll try to work with these cases in mind to satisfy `new-user 🟢` and `pro-user 🔴` 100% and see what we can do for `existing-user 🔵`

shayneczyzewski:
> @shayneczyzewski I think there are few user types here:
> 
> 1. User is using Wasp for the first time, he just wants "nice login with Google" out of the box `new-user 🟢`
> 2. User moved beyond trying out Wasp and is now serious about making the project in Wasp `pro-user 🔴`
> 3. Our existing users who already have the existing implementation `existing-user 🔵`
> 
> For `new-user 🟢` type of users, the less work they have to do, the more delightful the experience. Set `auth: google` and get something nice looking with profile picture and name. For example, a solution would be to have some smart defaults.
> 
> For `pro-user 🔴` the best is giving them total freedom to customize their user and social entities to their liking. All of the fields are optional.
> 
> For `existing-user 🔵` ideally nothing breaks with the upgrade. I understand that pre-1.0 software will have changes in the API. But we should either inform them with a nice error message what they need to change or try to be backwards compatible.
> 
> I'll try to work with these cases in mind to satisfy `new-user 🟢` and `pro-user 🔴` 100% and see what we can do for `existing-user 🔵`

I think keeping those distinct personas in mind is a great idea, @infomiho! And I think you are prioritizing the right use cases 👍🏻 Also, that markdown art you just did is `fire 🔥 ` :D I'm going to borrow that!

PR #982, opened 2023-01-31
Fix structural typing bug with operation types

sodic:


waspc/data/Generator/templates/server/src/entities/index.ts
```diff
@@ -4,15 +4,40 @@ import {
...
+{=/ entities =}
+
+export type _Entity = 
+  {=# entities =}
+  | {= internalTypeName =}
+  {=/ entities =}
+  | never
+
+type WithName<E extends Entity, Name extends EntityName> = 
+  E & { _entityName: Name }
```
sodic, line 43: Should we be creating "internal" files for all the stuff we don't want users to import?

They'll still be able to import anything we export, but importing from `something/_internal` will probably send the right signal.

waspc/data/Generator/templates/server/src/types/index.ts
```diff
@@ -33,25 +31,26 @@ type AuthenticatedOperation<Entities extends WaspEntity[], Result> = (
 type {= userViewName =} = Omit<{= userEntityName =}, 'password'>
 {=/ isAuthEnabled =}
 
-type Operation<Entities extends WaspEntity[], Result> = (
+type Operation<Entities extends _Entity[], Result> = (
   args: any,
   context: {
-      entities: EntityMap<Entities>,
+    entities: Expand<EntityMap<Entities>>,
```
sodic, line 37: @infomiho I ended up doing this in the end: https://github.com/wasp-lang/wasp/pull/945#discussion_r1084175543

The types look much nicer for the user (they never see the `Expand` type, which was my main concern).

waspc/data/Generator/templates/server/src/entities/index.ts
shayneczyzewski: Ideally, it would probably be better if they were somehow more hidden or differently scoped to indicate they should not use them, but since it is easier to just put them all into a single file I think it is probably fine to leave as is.

shayneczyzewski (commented):
Hey @sodic - this was a neat problem to review. Nice job! It is also quite surprising that you and Miho discovered it to begin with! TS is not my strength (yet 💪🏻) so I will provide a "weak" review, and just let @infomiho do the "strong" one if he can.

Overall, the approach feels like it makes sense. One thing I am not 100% sure on is if users import and use the non-internal ones, can they themselves ever hit the snag we are solving for them (I guess if they did, then they would know why and solve)?

Also, do we need to use a similar approach anywhere else we expose entities in the system (I guess for now though, it is just the context)?

Lastly, I do have entities in a context for Jobs, will we need to do something similar for that? I don't think that part has been part of any of the TS stuff yet (which is fine, I can in the future if needed)? https://github.com/wasp-lang/wasp/blob/main/waspc/data/Generator/templates/server/src/jobs/_job.js#L11-L14

Nice work!

waspc/data/Generator/templates/server/src/types/index.ts
infomiho: What do you think about the old idea of using 
```ts
type ContextWithEntities<Entities extends _Entity[]> = {
  entities: Expand<EntityMap<Entities>>
}
```
and then Operation looks like this
```ts
type Operation<Entities extends _Entity[], Result> = (
  args: any,
  context: ContextWithEntities<Entities>,
) => Promise<Result>
```
and AuthenticatedOperation looks like (notice the Expand around UserView)
```ts
type AuthenticatedOperation<Entities extends _Entity[], Result> =  (
  args: any,
  context: ContextWithEntities<Entities> & { user: Expand<UserView> },
) => Promise<Result>
```

infomiho: Even if we don't go with the `ContextWithEntities` type, I would still recomand wrapping `UserView` with `Expand` because it makes it easier for the user to see the fields.

waspc/data/Generator/templates/server/src/entities/index.ts
infomiho: It makes sense for the me to keep internal stuff in `_taggedEntities.ts` or something like that. It's not uncommon in JS world to have internal files prefixed with `_`.

One argument against it is: what happens when user clicks on the `_Task` type to see what's there. If the user is taken to a isolated file with only `_Task` which imports `Task` then the user needs to again click on the `Task` to see the definition. Keep `_Task` and `Task` in the same file helps. 

If we don't care that much about the second part, then I'm okay with moving this to an internal file, which will keep the entity types file cleaner.

infomiho:
I've played around with the generated TS and have some minor comments. Overall, it fixes the bug well 👍

waspc/data/Generator/templates/server/src/types/index.ts
sodic: Yes, thanks for reminding me. I had to rearrange the placements of `Expand`, but ended up getting this on hover:
image
Which I think is ideal.

waspc/data/Generator/templates/server/src/entities/index.ts
infomiho: Did we decide if we want to hide this in an internal file?

sodic: Yes, it's hidden now. Also, the entire `types` modules is now internal.

infomiho (approved):
LGTM

shayneczyzewski (approved):
I don't have a lot to add here given it's mostly TS magic to me, so if you and Miho are good with it, so am I. :)

waspc/data/Generator/templates/server/src/_types/index.ts
```diff
@@ -0,0 +1,55 @@
+{{={= =}=}}
```
sodic, line 1: I couldn't get Git to treat this change as a move. Compare with "deleted file" `types/index.ts` to see the true differences.

PR #988, opened 2023-02-03
Blog: truncate 2 posts

vincanger:
### Description

quick fix to truncate two blog posts


infomiho:
There is also a blog post by @Martinsos where we should add truncate `2022-12-01-beta-ide-improvements.md`

infomiho (approved):

Issue #975, opened 2023-01-25
Improve type support for operation arguments


sodic:
#969 introduced type support to operations (for the context object and optionally the return value). However, we still don't have type support for `args`:
```typescript
import { Task } from '@wasp/entities'
import { GetTasks} from '@wasp/queries'

const getTasks: GetTasks<Task[]> = (args, context) => { 
  // The function knows the type of 'context' 
  // It also knows it must return a value of type Task[]
  // But it doesn't know the type of 'args', it treats it as 'any'
}
```
We should implement support for `args`. We want something like this:
```typescript
import { Task } from '@wasp/entities'
import { GetTasks} from '@wasp/queries'

type Payload = Pick<Task, 'isDone'>;

const getTasks: GetTasks<Payload, Task[]> = (args, context) => { 
  // The function knows args has type 'Payload'
  // The function also knows everything it knew before :)
}
```

sodic:
@basdirks I know we went through it on Discord, but good thinking posting it here for visibility purposes :smiley: 

Anyway, here goes:

> We can also omit the second type variable, giving us GetTasks<Payload>, right?

That's correct. We should be able to omit both type arguments (they default to `unknown`).

> If we have the user point us to the argument type in their .wasp, we could also omit Payload inside the getTasks definition, because we could bake it into GetTasks during generation.

Also correct, but let's not do that just yet (i.e., let's stick to TS-only, no changes to the syntax and semantics of Wasp). 

The proposed additions only come into play once we get into full-stack type safety (#976).  Implementing it now would be much more work with no apparent benefits. Once we start working on #976, we'll see whether we want to move these types into the `.wasp` file

>  In order to add the Payload type variable, do we extend Query -> Operation, as we can type args to be Payload?

That is _exactly_ what we do :)

> If so, will the default type of Payload be unknown, like with Result?

Yes!

PR #989, opened 2023-02-04
Type operation input argument

basdirks:
Fixes [#975](https://github.com/wasp-lang/wasp/issues/975).

This types the input of `Query`, `Action`, `Operation`, `AuthenticatedQuery`, `AuthenticatedAction`, `AuthenticatedOperation`, and generated queries and actions.

```ts
type Operation<Entities extends WaspEntity[], Input, Result> = (
    args: Input,
    context: {
        entities: EntityMap<Entities>,
    },
}) => Promise<Result>;
```

Users can now specify the type of the input of an operation.

```ts
//                                         resolves to `Pick<Task, 'id'>`
//                                                       ↓
export const getTask: GetTask<Pick<Task, 'id'>> = async (where, context) => {
  return Task.findUnique({ where, include: { user: true } });
};
```

The order of type variables breaks all user-defined operations that have an explicit `Result` type that doesn't match `Input`.

```ts
// before:
export type GetFoo<Result> = // ...
export const getBarFoo: GetFoo<Foo> = async ({ bar: Bar }, context) => // ...

// after:
export type GetFoo<Input, Result> = // ...
//                        now `Input` :D     not `Foo` :(
//                             ↓             ↓
export const getBarFoo: GetFoo<Foo> = async (bar /* : Bar */, context) => // ...
```

We could make this backwards compatible by adding `Input` to the end of the type variable list, but it doesn't belong there.

I will edit the change log after initial feedback :)

### DIFF
--- a/waspc/data/Generator/templates/server/src/types/index.ts
+++ b/waspc/data/Generator/templates/server/src/types/index.ts
@@ -5,24 +5,24 @@ import {
   {=# entities =}
   type {= name =},
   {=/ entities =}
- } from "../entities"
+} from "../entities"
 
-export type Query<Entities extends WaspEntity[] = [], Result = unknown> = Operation<Entities, Result>
+export type Query<Entities extends WaspEntity[] = [], Input = unknown, Result = unknown> = Operation<Entities, Input, Result>
 
-export type Action<Entities extends WaspEntity[] = [], Result = unknown> = Operation<Entities, Result>
+export type Action<Entities extends WaspEntity[] = [], Input = unknown, Result = unknown> = Operation<Entities, Input, Result>
 
 {=# isAuthEnabled =}
-export type AuthenticatedQuery<Entities extends WaspEntity[] = [], Result = unknown> = 
-  AuthenticatedOperation<Entities, Result>
+export type AuthenticatedQuery<Entities extends WaspEntity[] = [], Input = unknown, Result = unknown> = 
+  AuthenticatedOperation<Entities, Input, Result>
 
-export type AuthenticatedAction<Entities extends WaspEntity[] = [], Result = unknown> = 
-  AuthenticatedOperation<Entities, Result>
+export type AuthenticatedAction<Entities extends WaspEntity[] = [], Input = unknown, Result = unknown> = 
+  AuthenticatedOperation<Entities, Input, Result>
 
-type AuthenticatedOperation<Entities extends WaspEntity[], Result> = (
-  args: any,
+type AuthenticatedOperation<Entities extends WaspEntity[], Input, Result> = (
+  args: Input,
   context: {
-      user: {= userViewName =},
-      entities: EntityMap<Entities>,
+    user: {= userViewName =},
+    entities: EntityMap<Entities>,
   },
 ) => Promise<Result>
 
@@ -33,10 +33,10 @@ type AuthenticatedOperation<Entities extends WaspEntity[], Result> = (
 type {= userViewName =} = Omit<{= userEntityName =}, 'password'>
 {=/ isAuthEnabled =}
 
-type Operation<Entities extends WaspEntity[], Result> = (
-  args: any,
+type Operation<Entities extends WaspEntity[], Input, Result> = (
+  args: Input,
   context: {
-      entities: EntityMap<Entities>,
+    entities: EntityMap<Entities>,
   },
 ) => Promise<Result>
 
--- a/waspc/data/Generator/templates/server/src/types/index.ts
+++ b/waspc/data/Generator/templates/server/src/types/index.ts
@@ -7,24 +7,24 @@ import {
   {=/ entities =}
 } from "../entities"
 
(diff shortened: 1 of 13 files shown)


basdirks:
About type variable order: I think `Result` is almost always `Entity` or `Entity[]`. `Input` is less obvious, although it will often be some `Partial<Entity>`.

waspc/data/Generator/templates/server/src/types/index.ts
```diff
@@ -5,24 +5,24 @@ import {
...
 {=# isAuthEnabled =}
-export type AuthenticatedQuery<Entities extends WaspEntity[] = [], Result = unknown> = 
-  AuthenticatedOperation<Entities, Result>
+export type AuthenticatedQuery<Entities extends WaspEntity[] = [], Input = unknown, Result = unknown> = 
+  AuthenticatedOperation<Entities, Input, Result>
 
-export type AuthenticatedAction<Entities extends WaspEntity[] = [], Result = unknown> = 
-  AuthenticatedOperation<Entities, Result>
+export type AuthenticatedAction<Entities extends WaspEntity[] = [], Input = unknown, Result = unknown> = 
+  AuthenticatedOperation<Entities, Input, Result>
```
sodic, line 19: Please change the default type of `Input` into `never`. This will allow subtyping when the input's type isn't specified.
### Explanation
Function types are _variant_ for result types, but _contravariant_ for argument types. Variance tells us whether applying a type constructor to a type preservers or flips the subtype/supertype relationship. More precisely (taken from the SO answer linked below):
> Variance has to do with how a generic type `F<T>` varies with respect to its type parameter `T`. If you know that `T extends U`, then variance will tell you whether you can conclude that `F<T> extends F<U>`, conclude that `F<U> extends F<T>`, or neither, or both.

Sources:
 - https://stackoverflow.com/a/66411491
 - https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)

### Example
Try compiling the following snippets:
```typescript
export const createTask: CreateTask<unknown> = async (task: Pick<Task, 'description'>, context) => {
  // body
}
```
```typescript
export const createTask: CreateTask<never> = async (task: Pick<Task, 'description'>, context) => {
  // body
}
```

waspc/examples/todoApp/src/server/queries.ts
```diff
@@ -6,7 +6,7 @@ import {
   GetTasks
 } from '@wasp/queries/types'
 
-export const getTasks: GetTasks<Task[]> = async (args, context) => {
+export const getTasks: GetTasks = async (_args, context) => {
```
sodic, line 9: Nice work renaming this!

waspc/examples/todoApp/src/server/queries.ts
```diff
@@ -23,19 +23,19 @@ export const getTasks: GetTasks<Task[]> = async (args, context) => {
   return tasks
 }
 
-export const getNumTasks: GetNumTasks<number> = async (args, context) => {
+export const getNumTasks: GetNumTasks = async (_args, context) => {
   return context.entities.Task.count()
 }
 
-export const getTask: GetTask<Task> = async ({ id }, context) => {
+export const getTask: GetTask<Pick<Task, 'id'>> = async (where, context) => {
```
sodic, line 30: I'd argue that changing this one from `{ id }` to `where` communicates less information than before, but I guess it could make sense now that we have the type. Your call!

sodic (changes requested):
Excellent work! One small change and I think we're good to go :smile:

waspc/data/Generator/templates/server/src/types/index.ts
basdirks: Good catch, will fix.

waspc/examples/todoApp/src/server/queries.ts
basdirks: I agree, but I like how passing the argument wholesale illustrates that quite often `args` is indeed equivalent to a `where` clause, and it just slots in so neatly, now that it is typed. Still a tough call, but since you're leaving it up to me, I'll keep it like this.

sodic: Sure! It's our internal testing app anyway. We usually go crazy with stuff here, so naming is the least of our concerns.

waspc/data/Generator/templates/server/src/types/index.ts
basdirks: Can you verify my update?

sodic (commented):
Just noticed two final things:
- Some of the return types are still called `Result`. Let's rename them all to `Output` for consistency
- Let's remove the default type arguments from `Query`, `Action`, `AuthenticatedQuery`, and `AuthenticatedAction`. These are internal types (i.e,. the users won't use them), and we already specify the default arguments one level above (in `actions/types.ts` and `ueries/types.ts`).

sodic (approved):
Perfect!

Thanks for helping out :smiley:

PR #994, opened 2023-02-08
Adds env validation and shared templates folder

### LINKED ISSUE #756: Ensure the REACT_APP_API_URL and WASP_WEB_CLIENT_URL env vars are valid URLs
Wasp users can control their backend server's URL by setting the `REACT_APP_API_URL` to a desired base URL.

We should check whether they specified a valid, absolute URL. To see what I mean, try running the app with the following two settings:
```bash
export REACT_APP_API_URL=something
export REACT_APP_API_URL=htp://something
```

In such cases, we should detect the error and warn the user with an appropriate message. Let's start by implementing this inside JavaScript (i.e., during runtime). We can use [this SO post](https://stackoverflow.com/questions/5717093/check-if-a-javascript-string-is-a-url) for inspiration.

Perhaps later we can extend the CLI to check for this before it runs the app (i.e., on `wasp start` and `wasp compile`)

infomiho:
Fixes #756

Ensures on runtime that env variables `REACT_APP_API_URL` and `WASP_WEB_CLIENT_URL` are valid URLs. 

- For the web app, it runs as a check before building or running the app - so it fails in the CLI rather in browser.
- For the server, it crashes the server with an error message. 

### Validator
It tries to construct a `new URL()` and it checks for `http` or `https` protocol existance. It seemed enough? 

### Shared templates
This PR also introduces a `shared` templates folder since the validator is shared between server and web app parts.

### Left to do
- [x] Tidy up the shared templates folder implementation and usage in the Server and WebApp generators.


waspc/src/Wasp/Generator/WebAppGenerator.hs
```diff
@@ -51,6 +52,8 @@ genWebApp spec = do
   sequence
     [ genFileCopy [relfile|README.md|],
       genFileCopy [relfile|tsconfig.json|],
+      genSharedFileCopy [relfile|validators.js|] [relfile|scripts/validators.mjs|],
```
infomiho, line 55: Maybe it makes sense to wrap these two lines into `genEnvValidation` or something similar.

waspc/src/Wasp/Generator/Shared.hs
```diff
@@ -0,0 +1,29 @@
+module Wasp.Generator.Shared (
```
infomiho, line 1: This piece was based heavily on existing Server and WebApp generators `Common` modules, there might be a chance to create some more useful "shared-specific" helper functions.

waspc/src/Wasp/Generator/ServerGenerator.hs
```diff
@@ -309,3 +311,12 @@ getPackageJsonOverrides = map buildOverrideData (designateLastElement overrides)
     designateLastElement l =
       map (\(x1, x2, x3) -> (x1, x2, x3, False)) (init l)
         ++ map (\(x1, x2, x3) -> (x1, x2, x3, True)) [last l]
+
+genValidators :: Generator [FileDraft]
+genValidators =
+  return
+    [ genSharedTmplToServerSrcDirCopy [relfile|validators.js|] [relfile|validators/validators.js|]
```
sodic, line 318: I suggest changing `validators.js` to `index.js`. It gives us cleaner imports.

waspc/src/Wasp/Generator/ServerGenerator.hs
```diff
@@ -309,3 +311,12 @@ getPackageJsonOverrides = map buildOverrideData (designateLastElement overrides)
...
+    ]
+
+genSharedTmplToServerSrcDirCopy :: Path' (Rel S.SharedTemplatesDir) File' -> Path' (Rel C.ServerSrcDir) File' -> FileDraft
+genSharedTmplToServerSrcDirCopy src = S.mkTmplFdWithDst src . (</>) C.serverSrcDirInProjectRootDir
```
sodic, line 322: **Nit:** I think the pointfree style makes things less readable in this case, so I'd probably specify both arguments explicitly (we'd get to name the other argument too).

If you decide to stick with the pointfree approach, I recommend changing `(</>) C.serverSrcDirInProjectRootDir` to `(</> C.serverSrcDirInProjectRootDir)`.

waspc/src/Wasp/Generator/Shared.hs
```diff
@@ -0,0 +1,31 @@
...
+import Wasp.Generator.Common (ProjectRootDir)
+import Wasp.Generator.FileDraft (FileDraft, createTemplateFileDraft)
+import Wasp.Generator.Templates (TemplatesDir)
+
+data SharedTemplatesDir
+
+mkTmplFdWithDst :: Path' (Rel SharedTemplatesDir) File' -> Path' (Rel ProjectRootDir) File' -> FileDraft
+mkTmplFdWithDst src dst = mkTmplFdWithDstAndData src dst Nothing
+
+-- | Path in templates directory where shared templates reside.
```
sodic, line 18: I also vote for dropping this comment (I'm aware you probably copied it over from existing generators). Still, I think the information given by the StrongPath type is plenty :)

waspc/src/Wasp/Generator/Shared.hs
```diff
@@ -0,0 +1,31 @@
...
+  )
+where
+
+import qualified Data.Aeson as Aeson
+import StrongPath (Dir, File', Path', Rel, reldir, (</>))
+import Wasp.Generator.Common (ProjectRootDir)
+import Wasp.Generator.FileDraft (FileDraft, createTemplateFileDraft)
+import Wasp.Generator.Templates (TemplatesDir)
+
+data SharedTemplatesDir
```
sodic, line 13: Hm, we now have both `Shared` and `Common`.
Sure, one is a module with common functionalities while the other is a module with functionalities related to the shared templates, but the names are still pretty similar (and possibly confusing).

I'm not sure whether it's better to come up with better names, or think about unifying them (they do share a semantic connection, after all).

If we opt for the second option, we can place `SharedTemplatesDir` and `sharedTemplatesDirInTemplatesDir` in _Generator/Common.hs_, and introduce a few changes to the file draft functions already present in _ServerGenerator.hs_ and _WebAppGenerator.hs_.

I'll use the `WebAppGenerator` as an example. This is what we currently have:
```haskell
-- file: Generator/WebAppGenerator/Common.hs

mkTmplFdWithDst :: Path' (Rel WebAppTemplatesDir) File' -> Path' (Rel WebAppRootDir) File' -> FileDraft
mkTmplFdWithDst src dst = mkTmplFdWithDstAndData src dst Nothing
```
Instead, we could have something like this:

```haskell
-- file: Generator/Common.hs

data SharedTemplatesDir


-- file: Generator/WebAppGenerator/Common.hs

class TemplatesDirForWebApp d

instance TemplatesDirForWebApp SharedTemplatesDir
instance TemplatesDirForWebApp WebAppTemplatesDir

mkTmplFdWithDst :: TemplatesDirForWebApp d => Path' (Rel d) File' -> Path' (Rel WebAppRootDir) File' -> FileDraft
mkTmplFdWithDst src dst = mkTmplFdWithDstAndData src dst Nothing
```
Doing so would allow us to use the same polymorphic function for working with any template file that's applicable to WebApp, be it from _shared/templates_ or _web-app/templates_.

sodic (commented):
Nice work with this and thanks for adding the shared folder! I needed it too but decided it was a niche requirement :).

Can you please verify whether the `WASP_WEB_CLIENT_URL` is or isn't sensitive to a trailing slash? If it is, we should handle it.

waspc/data/Generator/templates/react-app/package.json
```diff
@@ -6,8 +6,9 @@
   {=& depsChunk =},
   {=& devDepsChunk =},
   "scripts": {
-    "start": "react-scripts start",
-    "build": "react-scripts build",
+    "validate-env": "node ./scripts/validate-env.mjs",
+    "start": "npm run validate-env && react-scripts start",
+    "build": "npm run validate-env && react-scripts build",
```
sodic, line 11: Why can't the server fail like this too?

waspc/src/Wasp/Generator/Shared.hs
infomiho: This is an interesting idea and I quite like it. The `Shared.hs` module definetly felt a bit to me since it's not really standalone `SharedGenerator`.

I'll try using this new try we learned with creating `instances` to limit our polymorphism 🔥

waspc/data/Generator/templates/react-app/package.json
infomiho: I wanted to keep things inside of the `src` folder so I don't have to use `mjs`, but maybe in the name of consistency it's better to do same things in Web App and Server.

waspc/data/Generator/templates/react-app/scripts/validate-env.mjs
```diff
@@ -0,0 +1,7 @@
+// NOTE: validators are copied from shared templates
```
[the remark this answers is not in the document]
infomiho: Since we are outside of the `src` dir and working with plain Node.js scripts, I needed to use the `mjs` extensison to indicate that this file is in `ES module` format.

waspc/data/Generator/templates/shared/validators.js
```diff
@@ -0,0 +1,23 @@
+export function isValidAbsoluteURL(rawUrl) {
```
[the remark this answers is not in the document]
infomiho: This is plain JS file, so I can't type it. And since it will live outside the project structure as a plain Node.js script, I wouldn't like to add Typescript here.

waspc/data/Generator/templates/shared/validators.js
```diff
@@ -0,0 +1,23 @@
...
+    try {
+        new URL(rawUrl);
+    } catch (e) {
+        // Invalid URL
+        return false;
+    }
+
+    const urlParts = rawUrl.split("//");
```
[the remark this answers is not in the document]
infomiho: Added an explainer comment above 👍

waspc/data/Generator/templates/server/src/config.js
```diff
@@ -1,5 +1,7 @@
 {{={= =}=}}
 import _ from 'lodash'
+{=! NOTE: validators are copied from shared templates =}
```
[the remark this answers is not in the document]
infomiho: I've went with the first approach so the resulting structure is:
```
scripts/
-- validate-env.mjs
-- shared/
---- validators.mjs
```

waspc/src/Wasp/Generator/Shared.hs
infomiho: I've created a "generic" function for making FileDrafts with destination and data that accepts one of the two template folders `SharedTemplatesDir` and `WebAppTemplatesDir`. 

Then we export two functions `mkTmplWithDst` and `mkSharedTmplWithDst`, it should be clear when the template is coming from the shared folder.

waspc/data/Generator/templates/react-app/package.json
infomiho: I've updated the server to work exactly like the web app.

waspc/data/Generator/templates/shared/validators.js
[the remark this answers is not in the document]
sodic: My bad, thought it was TS

waspc/data/Generator/templates/shared/validators.js
```diff
@@ -0,0 +1,15 @@
+export function isValidAbsoluteURL(rawUrl) {
+    try {
+        const url = new URL(rawUrl);
+        /*
+            URL constructor will parse some invalid absolute URLs as valid URLs
+            so we need to do some extra checks.
+
+            Example: "localhost:3000" will be parsed as URL with protocol of
+            "localhost:" and host of "3000"
+        */
```
sodic, line 10: Nice explanation!

waspc/src/Wasp/Generator/ServerGenerator.hs
```diff
@@ -323,3 +324,10 @@ getPackageJsonOverrides = map buildOverrideData (designateLastElement overrides)
     designateLastElement l =
       map (\(x1, x2, x3) -> (x1, x2, x3, False)) (init l)
         ++ map (\(x1, x2, x3) -> (x1, x2, x3, True)) [last l]
+
+genEnvValidationScript :: Generator [FileDraft]
+genEnvValidationScript =
+  return
+    [ C.mkTmplFd [relfile|scripts/validate-env.mjs|],
+      C.mkSharedTmplFdWithDst [relfile|validators.js|] [relfile|scripts/shared/validators.mjs|]
```
sodic, line 332: Really minor, but why not change the the template's extension to `mjs` as well?

waspc/src/Wasp/Generator/ServerGenerator/Common.hs
```diff
@@ -33,6 +35,12 @@ data ServerTemplatesDir
 
 data ServerTemplatesSrcDir
 
+class ValidServerTemplatesDir d
+
+instance ValidServerTemplatesDir ServerTemplatesDir
```
sodic, line 40: **Nit:** Maybe a better name would be `ServerTemplatesLocation`. We basically have to make something up to avoid clashing with `ServerTemplatesDir` (which would obviously be the most appropriate name), but I think `Valid` is a bit weird (don't know how to argument why).

waspc/data/Generator/templates/react-app/scripts/validate-env.mjs
```diff
@@ -0,0 +1,7 @@
+import { isValidAbsoluteURL } from './shared/validators.mjs';
+
+console.info("🔍 Validating environment variables...");
+if (process.env.REACT_APP_API_URL && !isValidAbsoluteURL(process.env.REACT_APP_API_URL)) {
+    throw 'Environemnt variable REACT_APP_API_URL is not a valid absolute URL';
+}
```
sodic, line 6: This block is identical to the one in `server` (modulo env var name). Can we put it in `validators.mjs` as well? We could use it like `throwIfInvalidUrl(process.env.REACT_API_URL)`.

Btw, we're throwing strings. Aren't they usually wrapped in `Error`s? Is this an output thing?

waspc/src/Wasp/Generator/Shared.hs
sodic: I now see this whole isn't playing out as I had hoped. My original suggestion obviously wouldn't work because of the error this code throws:
```haskell
mkAnyValidTmplFdWithDstAndData ::
  ValidWebAppTemplatesDir d =>
  Path' (Rel d) File' ->
  Path' (Rel WebAppRootDir) File' ->
  Maybe Aeson.Value ->
  FileDraft
mkAnyValidTmplFdWithDstAndData relSrcPath relDstPath tmplData =
  createTemplateFileDraft
    (webAppRootDirInProjectRootDir </> relDstPath)
    (webAppTemplatesDirInTemplatesDir </> relSrcPath)
    tmplData
``` 
So, unfortunately, it doesn't look like the type class is helping us that much.

<details> 
<summary>Maybe we should just keep it simple and ditch the polymorphism:
</summary>

```haskell
mkTmplFdWithDstAndData ::
  Path' (Rel WebAppTemplatesDir) File' ->
  Path' (Rel WebAppRootDir) File' ->
  Maybe Aeson.Value ->
  FileDraft
mkTmplFdWithDstAndData relSrcPath = 
  mkAnyValidTmplFdWithDstAndData (webAppTemplatesDirInTemplatesDir </> relSrcPath)

mkSharedTmplFdWithDst :: Path' (Rel SharedTemplatesDir) File' -> Path' (Rel WebAppRootDir) File' -> FileDraft
mkSharedTmplFdWithDst relSrcPath relDstPath =
  mkAnyValidTmplFdWithDstAndData
    (sharedTemplatesDirInTemplatesDir </> relSrcPath)
    relDstPath
    Nothing
    
mkAnyValidTmplFdWithDstAndData ::
  Path' (Rel TemplatesDir) File' ->
  Path' (Rel WebAppRootDir) File' ->
  Maybe Aeson.Value ->
  FileDraft
mkAnyValidTmplFdWithDstAndData relSrcPath relDstPath tmplData =
  createTemplateFileDraft
    (webAppRootDirInProjectRootDir </> relDstPath)
    relSrcPath
    tmplData
```

</details>

<details> 


> **Warning**
> I strongly prefer the previous option and am 85% joking with this one :)

<summary>Or, of course, we could double down:
</summary>

```haskell
class ValidWebAppTemplatesDir d where
  pathInTemplatesDir :: Path' (Rel TemplatesDir) (Dir d)

instance ValidWebAppTemplatesDir WebAppTemplatesDir where
  pathInTemplatesDir = webAppTemplatesDirInTemplatesDir

instance ValidWebAppTemplatesDir SharedTemplatesDir where
  pathInTemplatesDir = sharedTemplatesDirInTemplatesDir
  
-- All higher level functions would now be polymorphic and work automatically:

mkTmplFdWithDst :: ValidWebAppTemplatesDir d => Path' (Rel d) File' -> Path' (Rel WebAppRootDir) File' -> FileDraft
mkTmplFdWithDst src dst = mkTmplFdWithDstAndData src dst Nothing

mkTmplFdWithData :: ValidWebAppTemplatesDir d => Path' (Rel d) File' -> Aeson.Value -> FileDraft
mkTmplFdWithData src tmplData = mkTmplFdWithDstAndData src (SP.castRel src) (Just tmplData)

mkTmplFdWithDstAndData ::
  ValidWebAppTemplatesDir d =>
  Path' (Rel d) File' ->
  Path' (Rel WebAppRootDir) File' ->
  Maybe Aeson.Value ->
  FileDraft
mkTmplFdWithDstAndData relSrcPath relDstPath tmplData =
  createTemplateFileDraft
    (webAppRootDirInProjectRootDir </> relDstPath)
    (pathInTemplatesDir </> relSrcPath)
    tmplData
```

But we would then sometimes have to pay the verbosity tax when calling these functions (by specifying types for all arguments, one way or the other). For example:
```haskell
genEnvValidationScript :: Generator [FileDraft]
genEnvValidationScript =
  return
    [ C.mkTmplFd [relfile|scripts/validate-env.mjs|],
      C.mkTmplFdWithDst [relfile|validators.js|] [relfile|scripts/shared/validators.mjs|]
    ]
```
Would have to become:

```haskell
genEnvValidationScript :: Generator [FileDraft]
genEnvValidationScript =
  return
    [ C.mkTmplFd ([relfile|scripts/validate-env.mjs|] :: Path' (Rel WebAppTemplatesDir) File'),
      C.mkTmplFdWithDst
        ([relfile|validators.js|] :: Path' (Rel WebAppTemplatesDir) File')
        [relfile|scripts/shared/validators.mjs|]
    ]
```

</details>

With all that said, if you're tired of making changes, keeping what you currently have is fine too.

waspc/src/Wasp/Generator/WebAppGenerator/Common.hs
```diff
@@ -89,8 +96,20 @@ mkTmplFdWithDstAndData ::
...
+mkSharedTmplFdWithDst relSrcPath relDstPath = mkAnyValidTmplFdWithDstAndData sharedTemplatesDirInTemplatesDir relSrcPath relDstPath Nothing
+
+mkAnyValidTmplFdWithDstAndData ::
+  ValidWebAppTemplatesDir d =>
+  Path' (Rel TemplatesDir) (Dir d) ->
+  Path' (Rel d) File' ->
+  Path' (Rel WebAppRootDir) File' ->
+  Maybe Aeson.Value ->
+  FileDraft
+mkAnyValidTmplFdWithDstAndData templatesDir relSrcPath relDstPath tmplData =
```
sodic, line 111: I think the naming is a bit off here.

`mkAnyValidTmplFdWithDstAndData` (has prefix `AnyValid`) specializes into:
  - `mkSharedTmplFdWithDst` (has prefix `Shared`)
  - `mkTmplFdWithDstAndData` (no prefix, implying it's the general one of the three, but is actually specific).

I suggest:

`mkTmplFdWithDstAndData`specializes into:
  - `mkSharedTmplFdWithDstAndData`
  - `mkWebAppTmplFdWithDstAndData`

waspc/data/Generator/templates/server/src/config.js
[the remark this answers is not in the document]
sodic: Crap, didn't see this on time, I wanted to suggest option 2... sorry :sweat_smile: 

Shared stuff might contain more than just the the scripts stuff (e.g., types, application logic). We already have some template (and generated code) duplication with types this could help reduce (e.g., entity definitions).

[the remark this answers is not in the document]
infomiho: Scripts and src in my mind are separate scopes and I wanted to avoid cross-imports from one scope to another e.g. `import validators from '../src/shared/validators.mjs`

For example in the `scripts` scope we have a `shared` folder, but also in the `src` folder we have a `shared` folder. Does that make sense?

waspc/src/Wasp/Generator/ServerGenerator.hs
infomiho: It's slightly related to the idea of `scripts` and `src` being two different scopes. For some other use case in` src` where we would use the validators, the extension won't be `mjs` but rather `js`. 

So, it's the renaming of the file that actually makes this file work in `scripts` scope. It's kinda implicit, if you have a better way of doing it, I'll hapiliy consider it.

waspc/src/Wasp/Generator/ServerGenerator/Common.hs
infomiho: > We basically have to make something up to avoid clashing with ServerTemplatesDir
Precisely :joy:

I like your suggestion, I'll take it.

waspc/src/Wasp/Generator/WebAppGenerator/Common.hs
infomiho: I wanted to avoid renaming the most used function to `mkWebAppTmplFdWithDstAndData`, I liked your idea about ditching polymorphism and just created a concrete function related to using the shared templates. 

I'll see how the renaming goes and take this comment into account 👍 "more specialized function shouldn't look like the more general function due to a missing prefix"

waspc/data/Generator/templates/react-app/scripts/validate-env.mjs
infomiho: That's an interesting idea, there is some logic here (if the variable is set then check it) which would be good if it was extracted.

I dropped the `Error` wrapper to get a slightly nicer output IMHO. What do you think:
```js
// Wrapped with Error
Web app(stderr): file:///Users/ilakovac/wasp-lang/wasp/waspc/examples/todoApp/.wasp/out/web-app/scripts/validate-env.mjs:5
Web app(stderr):     throw new Error('Environemnt variable REACT_APP_API_URL is not a valid absolute URL');
Web app(stderr):           ^
Web app(stderr):
Web app(stderr): Error: Environemnt variable REACT_APP_API_URL is not a valid absolute URL
Web app(stderr):     at file:///Users/ilakovac/wasp-lang/wasp/waspc/examples/todoApp/.wasp/out/web-app/scripts/validate-env.mjs:5:11
Web app(stderr):     at ModuleJob.run (node:internal/modules/esm/module_job:193:25)
Web app(stderr):     at async Promise.all (index 0)
Web app(stderr):     at async ESMLoader.import (node:internal/modules/esm/loader:530:24)
Web app(stderr):     at async loadESM (node:internal/process/esm_loader:91:5)
Web app(stderr):     at async handleMainPromise (node:internal/modules/run_main:65:12)
Web app(stderr):
Web app(stderr): Node.js v18.12.1
```

```js
// Plain string
Web app(stderr): node:internal/process/esm_loader:97
Web app(stderr):     internalBinding('errors').triggerUncaughtException(
Web app(stderr):                               ^
Web app(stderr): Environemnt variable REACT_APP_API_URL is not a valid absolute URL
Web app(stderr): (Use `node --trace-uncaught ...` to show where the exception was thrown)
Web app(stderr):
Web app(stderr): Node.js v18.12.1
```

Now that I look at it side by side, it seems that the `Error` wrapped one provides a location of the throw, which can be useful.

waspc/src/Wasp/Generator/ServerGenerator/Common.hs
infomiho: If we were to ditch the polymorphism, then this suggestion becomes no-op. Nice name though.

waspc/src/Wasp/Generator/WebAppGenerator/Common.hs
infomiho: There are now two independnt functions `mkTmplFdWithDst` like before (which lives in either `Server` or `WebApp` context) and a special one `mkSharedTmplFdWithDst` which lives alongside. the frist one, but it signals that it uses the shared templates vs. regular templates.

waspc/src/Wasp/Generator/Shared.hs
infomiho: I've went with the route of ditching polymorphism, keeping this simple.

We just introduced a `SharedTemplatesDir` concept and created a function for crafting `FileDrafts` from shared templates in `WebApp` and `Server` contexts. We expand on that as we use `shared` templates more.

shayneczyzewski:
This is looking really cool @infomiho I like the direction of adding a new shared template notion (TBH, I'm not in love with `universal`, but re-using `shared` is probably not good either, and `common` may be confusing, so I think it's probably fine).

This approach will work with both envars in a `.env` file or just in the environment as regular environment variables, right? Really excited to have this. Will give a more thorough review once closer (just ping when ready). 👍🏻

infomiho:
@shayneczyzewski the name is a bit  weird at first, but I've seen in the wild when people reference "full-stack" JS.

Yep, it validates on run-time what it sees in `env`, wherever it comes from.

waspc/data/Generator/templates/server/src/config.js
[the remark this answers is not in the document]
sodic: Talked about this, we're using `shared` and `universal` to distinguish between the two top-level shared folders (user-defined and ours).

@infomiho Should we rename the user-defined one to `ext-shared`?

waspc/data/Generator/templates/react-app/scripts/validate-env.mjs
sodic: There's still some repetition here, why not just:
```
throwIfInvalidUrl(process.env.REACT_API_URL)
```
And report the message inside (it should be the same for all env vars).

waspc/data/Generator/templates/server/tsconfig.json
```diff
@@ -16,5 +15,6 @@
     "moduleResolution": "node",
     "outDir": "dist",
     "allowJs": true
-  }
+  },
+  "include": ["src"]
```
sodic, line 19: Is the web app missing this as well?

waspc/data/Generator/templates/universal/url.ts
```diff
@@ -0,0 +1,3 @@
+export function stripTrailingSlash(url?: string): string | undefined {
```
sodic, line 1: **Nit:** If I'm not mistaken, the signature is a bit underspecified. We have:
```typescript
(x: string | undefined) => string | undefined
```
While the "correct" signature would be:
```typescript
(x: string) => string | (x: undefined) => undefined
```

waspc/data/Generator/templates/universal/validators.js
```diff
@@ -0,0 +1,21 @@
...
+        return url.protocol === "http:" || url.protocol === "https:";
+    } catch (e) {
+        return false;
+    }
+}
+
+export function throwIfNotValidAbsoluteURL(rawUrl, errorMessage) {
+    if (rawUrl && !isValidAbsoluteURL(rawUrl)) {
+        throw new Error(errorMessage);
+    }
```
sodic, line 20: When `rawUrl` is `undefined`, we don't throw an error? Is this expected behavior?

waspc/src/Wasp/Generator/WebAppGenerator/Common.hs
```diff
@@ -89,8 +90,15 @@ mkTmplFdWithDstAndData ::
...
-mkTmplFdWithDstAndData srcPathInWebAppTemplatesDir dstPathInWebAppRootDir tmplData =
+mkTmplFdWithDstAndData relSrcPath relDstPath tmplData =
   createTemplateFileDraft
-    (webAppRootDirInProjectRootDir </> dstPathInWebAppRootDir)
-    (webAppTemplatesDirInTemplatesDir </> srcPathInWebAppTemplatesDir)
+    (webAppRootDirInProjectRootDir </> relDstPath)
+    (webAppTemplatesDirInTemplatesDir </> relSrcPath)
     tmplData
+
+mkUniversalTmplFdWithDst :: Path' (Rel UniversalTemplatesDir) File' -> Path' (Rel WebAppRootDir) File' -> FileDraft
```
sodic, line 99: The signature should probably be polymorphic (`File a` and `File b` instead of `File'`), right? The same should apply to the rest of these functions.

@Martinsos am I missing something, I see all old functions use units.

waspc/src/Wasp/Generator/ServerGenerator.hs
sodic: It's fine as far as I'm concerned.

waspc/data/Generator/templates/universal/validators.js
infomiho: This is indeed the expected behaviour since we have fallbacks for `REACT_APP_API_URL` and `WASP_WEB_CLIENT_URL` env variables. If the user doesn't specify the env variables, it should not throw the erorr.

For this use case, this is fine IMHO, if we start using this function more, maybe we can refactor it to be more explicit about not throwing in that case or change the behaviour.

waspc/data/Generator/templates/universal/url.ts
infomiho: We agreed offline that we are okay with the current signature 👍

waspc/data/Generator/templates/react-app/scripts/validate-env.mjs
infomiho: I wrote this function without thinking about `env` variables, that's why I wanted to pass in the full error message. 

I could reduce it to `throwIfInvalidUrl (value, name)` and put inside ```throw new Error(`${name} is not valid absolute URL`)``` 

Then we could send in `throwIfInvalidUrl(process.env.REACT_API_URL, 'Env variable REACT_API_URL')`

waspc/data/Generator/templates/server/src/config.js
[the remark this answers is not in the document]
infomiho: @sodic That's a nice move, since we already have `ext-src` and use the `ext` to signal user-defined code. 

I'll make that change in this PR. Should we then still stick to `universal`? I'm on the fence tbh, it helps to further distinguish between `shared` and `ext-shared`, but then again, we do have `ext-src` and `src`.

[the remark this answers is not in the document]
sodic: You can make the choice, I'd probably go with shared but wouldn't go out of my way to change it.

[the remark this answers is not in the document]
infomiho: We agreed that we'll keep `universal` for now, since going with `ext-shared` will require rewriting of paths for imports in the `client` and `server` dirs.

Screenshot 2023-02-20 at 15 20 44

waspc/data/Generator/templates/react-app/scripts/validate-env.mjs
infomiho: I've updated it here: https://github.com/wasp-lang/wasp/pull/994/commits/e6000d2ffd8328c706d9d95cbb377c3c4db6361b

sodic (approved):

PR #997, opened 2023-02-09
Adds a workaround for GHCup CI issue

infomiho:
Adds the recommended fix for Haskell setup error in our CI. 

Reference: https://discourse.haskell.org/t/incident-github-actions-ci-failure-ghcup/5761

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -64,6 +64,11 @@ jobs:
       - name: Checkout the repo
         uses: actions/checkout@v3
 
+      - name: Workaround runner image issue
+        if: matrix.os == 'ubuntu-20.04'
+        # https://github.com/actions/runner-images/issues/7061
+        run: sudo chown -R $USER /usr/local/.ghcup
+
       - name: Set up Haskell
         id: setup-haskell-cabal
         uses: haskell/actions/setup@v2
@@ -104,9 +109,9 @@ jobs:
         if: matrix.os == 'ubuntu-20.04' || matrix.os == 'macos-latest'
         run: ./tools/install_deploy_package_to_data_dir.sh
 
-      - name: Build external dependencies 
-        run: cabal build --enable-tests --enable-benchmarks --only-dependencies 
-      
+      - name: Build external dependencies
+        run: cabal build --enable-tests --enable-benchmarks --only-dependencies
+
       - name: Build wasp code
         run: cabal build all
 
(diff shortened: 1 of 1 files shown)


sodic (approved):

.github/workflows/ci.yaml
```diff
@@ -104,16 +109,16 @@ jobs:
...
+        run: cabal build --enable-tests --enable-benchmarks --only-dependencies
+
       - name: Build wasp code
         run: cabal build all
 
       - name: Set up Node
         uses: actions/setup-node@v3
         with:
-          node-version: '18'
+          node-version: '18.12'
```
Martinsos, line 121: You pinned this down to make sure it works now, due to it pulling too new version of `npm`, 9 instead of 8?

While this is ok as a temporary fix, I would consider putting it back to `18` if we decide to drop the restrictions on `npm`, because we want to have the same experience as users have. It is good that we got the error in this case, when users also got it, I think we want to keep this behaviour.

.github/workflows/ci.yaml
```diff
@@ -64,6 +64,11 @@ jobs:
       - name: Checkout the repo
         uses: actions/checkout@v3
 
+      - name: Workaround runner image issue
+        if: matrix.os == 'ubuntu-20.04'
+        # https://github.com/actions/runner-images/issues/7061
```
Martinsos, line 69: Sounds good, but I would also add a TODO here that says: This is a fix for thing X and once it is fixed, pls remove this step. You can also mention estimated time of it being fixed: for example it should be fixed by the end of Feb 2023.

Why? With the comment you left, it will take very long time till somebody checks that link -> there is no reason to look at it.

But with the comment I suggested above, the next time somebody looks at this piece of code, they will likely follow up on it, especially if they see that the mentioned date has passed, so we will have this fix no longer than we should. And I think that is good, I would rather not unnecessary have it here if it is not needed.

.github/workflows/ci.yaml
infomiho: I'll update the version when we fix the issue of using Node.js 18.14 👍

.github/workflows/ci.yaml
infomiho: Makes sense, I'll add it in the next update to the workflow (when I'll be changing the Node.js version back to `18`)

.github/workflows/ci.yaml
Martinsos: Ok! For the future, it is always good to leave the TODO if you have future plans or something is temporary -> you could forget for some reason, and then we would have this here but wouldn't know we should change it or why it was done this way.

PR #1001, opened 2023-02-11
Relaxes npm version to be any version

infomiho:
Removes the constraint on the `npm` version the user need to have installed.

### Background

With the change from Node.js `18.12` to `18.14`, Node.js ships a [different major npm version](https://nodejs.org/en/blog/release/v18.14.0/#updated-npm-to-9-3-1) (version 9). And this breaks the experience for our users who use Node.js 18.14 which is the default that gets installed when running `nvm install 18`.

### DIFF
--- a/waspc/src/Wasp/SemanticVersion.hs
+++ b/waspc/src/Wasp/SemanticVersion.hs
@@ -25,13 +25,11 @@ data Version
         minor :: Natural,
         patch :: Natural
       }
-  | AnyVersion
   deriving (Eq, Ord)
 
 -- | We rely on this `show` implementation to produce valid semver representation of version.
 instance Show Version where
   show (Version mjr mnr ptc) = printf "%d.%d.%d" mjr mnr ptc
-  show AnyVersion = "*"
 
 data Operator
   = Equal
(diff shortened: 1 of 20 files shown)


waspc/src/Wasp/SemanticVersion.hs
```diff
@@ -19,11 +19,12 @@ import Text.Printf (printf)
 
 -- Implements SemVer (semantic versioning) by following spec from https://github.com/npm/node-semver .
 
-data Version = Version
-  { major :: Natural,
-    minor :: Natural,
-    patch :: Natural
-  }
+data Version
+  = Version
```
Martinsos, line 23: Can we format to be as it was?

Martinsos (approved):
LGTM, nice!

waspc/src/Wasp/SemanticVersion.hs
infomiho: Yep, it's failing the CI, I had to turn of autoformat for a second (package.json get messed up otherwise)

infomiho:
We'll do the fix directly on the `release` branch

PR #1004, opened 2023-02-13
Updates type for relative import paths

infomiho:
The types read much nicer now, it's clear from `Path Posix (Rel ImportLocation) (Dir WebAppSrcDir)` that we are importing something from `WebAppSrcDir` relative to the import location.


Martinsos (approved):
LGTM!

PR #1005, opened 2023-02-14
[WIP] Adds support for root client component

### LINKED ISSUE #77: Enable custom wrapping of top-lvl React component
If Wasper wants to, for example, add Redux to their app, they will want to wrap the top-lvl React component into Provider for Redux. In that case, they will need some kind of mechanism to do that -> it could be a JS function that receives top lvl component and returns the same component but wrapped into whatever they like -> basically a decorator.

infomiho:
### Root component
Fixes #77

This PR introduces `rootComponent` in the `app.client` dictionary in Wasp lang which will help users to define their root React component.

`rootComponent` is of type `ExtImport` which is an import statement from user defined files. 

This enables the user to do:
1. Define a `src/client/App.jsx` which exports a component
2. Use the component to setup various providers e.g. `Redux` or `ChakraProvider` https://chakra-ui.com/getting-started

We then use their root component to wrap the `router` in the app's `index.js`.

### User in React Query cache bug
Since we now have a root component that's mounted when there is a change from logged in to logged out state, a bug bubbled up: after we performed the `login` action we didn't properly invalidate the query cache (user wasn't updated and thus component weren't rerendered). Fixed [here](https://github.com/wasp-lang/wasp/pull/1005/files#diff-01340e49338218c93087c78525c0899e764359a2db4833628e06447e80a2b36eR20).

### [BREAKING] Auth changes
Instead of using the `history` API to push pages, we now use the prefered (?) `<Redirect />` React Router component to redirect after login and signup. [<AuthSuccessRedirect />](https://github.com/wasp-lang/wasp/pull/1005/files#diff-225186dd44ab268657d5bed00d9df8778b25408fe8748b0aad4747ee15d4b3e2R1)

This is a BREAKING CHANGE for our users that expected the redirect behaviour happening after logging in with our `Login` form.

They now have to return the `<AuthSuccessRedirect />` component like [this](https://github.com/wasp-lang/wasp/pull/1005/files#diff-ea888a61f3842fcbc16dca741a1c425c4c09a9b54d5b22acbeee918a182d613dR13)

### Left to do:
- [x] Cover the change with unit tests
- [x] Cover the change with E2E tests
- [ ] Update the docs
- [x] Fix a bug with `Logout` button not showing up on login


waspc/data/Generator/templates/react-app/src/index.js
```diff
@@ -4,22 +4,26 @@ import ReactDOM from 'react-dom'
...
+{=& rootComponent.importStatement =}
+{=/ rootComponent.isDefined =}
 
 startApp()
 
 async function startApp() {
-  {=# doesClientSetupFnExist =}
-  await {= clientSetupJsFnIdentifier =}()
-  {=/ doesClientSetupFnExist =}
+  {=# setupFn.isDefined =}
```
infomiho, line 24: Ideally the check would be just `{=# setupFn =}`, but Mustache doesn't trigger `{=^ setupFn =}` when it's `undefined` vs. `false`, which is weird. So, that's why we use this `isDefined` crutch to make it work in all cases.

waspc/examples/todoApp/src/client/App.jsx
```diff
@@ -0,0 +1,25 @@
+import logout from "@wasp/auth/logout.js";
```
infomiho, line 1: Having a root component opened up a place for `Main.css` import and having a `Logout` button present in the header of each page.

Martinsos:
@infomiho Nice! Btw is this one ready for review? If not yet, then it is good to keep it a draft so people know.

infomiho:
I'll be adding the same couple of changes through couple of separate PRs.

PR #1006, opened 2023-02-14
Allow Fly to build locally

shayneczyzewski:
### Description

Allows `wasp deploy fly` users to specify if server/client should be built locally (using their own Docker) by optionally passing `--build-locally` to `launch` and/or `deploy`. Before, it would _always_ just run on the Fly builder. This may not be ideal if the build process requires lots of resources/time, and they cannot use the Free Builder for it.

It also fixes up the use of shared options and how those are expressed in the interfaces.

### DIFF
--- a/waspc/packages/deploy/src/providers/fly/deploy/deploy.ts
+++ b/waspc/packages/deploy/src/providers/fly/deploy/deploy.ts
@@ -50,7 +50,7 @@ export async function deploy(options: DeployOptions): Promise<void> {
 		const inferredBaseName = getInferredBasenameFromServerToml(tomlFilePaths);
 		const deploymentInfo = createDeploymentInfo(inferredBaseName, undefined, options, tomlFilePaths);
 		await buildWasp();
-		await deployServer(deploymentInfo, options.buildServerRemotely);
+		await deployServer(deploymentInfo, options.buildServerLocally);
 	}
 
 	if (!clientTomlExistsInProject(tomlFilePaths)) {
@@ -61,11 +61,11 @@ export async function deploy(options: DeployOptions): Promise<void> {
 		const inferredBaseName = getInferredBasenameFromClientToml(tomlFilePaths);
 		const deploymentInfo = createDeploymentInfo(inferredBaseName, undefined, options, tomlFilePaths);
 		await buildWasp();
-		await deployClient(deploymentInfo, options.buildClientRemotely);
+		await deployClient(deploymentInfo, options.buildClientLocally);
 	}
 }
 
-async function deployServer(deploymentInfo: DeploymentInfo, buildServerRemotely: boolean) {
+async function deployServer(deploymentInfo: DeploymentInfo, buildServerLocally: boolean) {
 	waspSays('Deploying your server now...');
 
 	cdToServerBuildDir(deploymentInfo.options.waspProjectDir);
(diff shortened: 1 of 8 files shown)


waspc/packages/deploy/src/providers/fly/index.ts
```diff
@@ -20,6 +20,10 @@ class FlyCommand extends Command {
 			.option('--initial-cluster-size <initialClusterSize>', 'flyctl postgres create option', '1')
 			.option('--volume-size <volumeSize>', 'flyctl postgres create option', '1');
 	}
+	addLocalBuildOption(): this {
+		return this.option('--build-server-locally', 'force the server Docker container to build locally', false)
```
Martinsos, line 24: What happens otherwise? It always get's built remotely?
When I read "forces", it sounds like normally it might build locally or it might build remotely, but now we made sure it builds locally -> but is that correct?
We could mention default behaviour here if there is one.

waspc/packages/deploy/src/providers/fly/index.ts
```diff
@@ -20,6 +20,10 @@ class FlyCommand extends Command {
...
+			.option('--build-client-locally', 'force the client Docker container to build locally', false);
```
Martinsos, line 25: Just curious, how is it that we have two commands, and not just one `--build-locally` that would build everything locally? When might one want to build one thing locally and other thing not?

Martinsos (approved):
2 tiny comments, LGTM otherwise!

waspc/packages/deploy/src/providers/fly/deploy/deploy.ts
```diff
@@ -61,11 +61,11 @@ export async function deploy(options: DeployOptions): Promise<void> {
 		const inferredBaseName = getInferredBasenameFromClientToml(tomlFilePaths);
 		const deploymentInfo = createDeploymentInfo(inferredBaseName, undefined, options, tomlFilePaths);
 		await buildWasp();
-		await deployClient(deploymentInfo);
+		await deployClient(deploymentInfo, options.buildClientLocally);
 	}
 }
 
-async function deployServer(deploymentInfo: DeploymentInfo) {
+async function deployServer(deploymentInfo: DeploymentInfo, buildServerLocally: boolean) {
```
infomiho, line 68: Nit: I prefer prefer using option objects vs. providing `boolean` params.

```
await deployServer({}, true)
```
vs.
```
await deployServer({}, { buildClientLocally: true })
```

Your code is clear to me because you are using it like this:
```
await deployClient(deploymentInfo, options.buildClientLocally);
```
but I'm just looking out for the future, and making this code easier to extend with additional options if we need to.

shayneczyzewski: Good point, can do 👍🏻

waspc/packages/deploy/src/providers/fly/index.ts
shayneczyzewski: The current implementation defaults to building remotely, which is mentioned in the docs (I can mention here too). This flag forces local. The default behavior of `flyctl deploy` without any flags is to build locally _if_ Docker was running, but otherwise build remotely. I am not a big fan of that for Wasp users to help support and debug build failures, etc., so was thinking it may benefit us to make it 100% deterministic, so we just default to remote but they can override and build locally if the need arises. Does that make sense?

waspc/packages/deploy/src/providers/fly/index.ts
shayneczyzewski: Unsure, I can collapse it to one option for now and we can split if needed.

waspc/packages/deploy/src/providers/fly/index.ts
shayneczyzewski: Actually, to probably better address your point, I can make it read something like `builds the Docker container locally instead of remotely` or something like that?

infomiho (approved):

Issue #897, opened 2022-12-22
Implement proper production support


sodic:
Summary:
- We have dependencies (things we need when running the code), and we have dev dependencies (things we need when writing the code, e.g., TypeScript).
- TypeScript is specified as a dev dependency (as it should be), meaning we shouldn't install it when building Wasp for production. The same goes for _nodemon_, _@types_ packages, etc. One can usually avoid installing dev dependencies by running the command `npm install --production` (this creates a "production build).
- Wasp does not make this distinction, it always calls the same code which ends up calling the regular `npm install`, regardless of the Wasp command at hand (`start`, `build`, `compile`, ...).

To make Wasp production ready, we must:
- Start treating dev dependencies differently than regular dependencies
- Put each dependency into the appropriate category (e.g. TypeScript is a dev dependency, Express is a regular dependency)
- Ensure our code can run without any dev dependencies (e.g., we currently don't compile TypeScript when building for production)

Martinsos:
Is this blocked by the restructuring, or can we go after it right now?

infomiho:
@sodic is this now solved after 0.12.0?

sodic:
I'm not sure, some parts of it may be fixed, but I think not (I think we still treat them the same way when building for production).

Most (if not all) points are still applicable, I'd say.

PR #1008, opened 2023-02-15
Build TS in Docker, not when starting app

shayneczyzewski:
### Description

This PR moves the TS build process (`npx tsc`) into the Dockerfile, thus happening at build time and not when the app starts. This should reduce the memory requirements for running the app.

In the future, we will fix how production assets are built and shipped here: https://github.com/wasp-lang/wasp/issues/897


waspc/data/Generator/templates/Dockerfile
```diff
@@ -47,6 +47,7 @@ WORKDIR /app
 COPY --from=server-builder /app/server/node_modules ./server/node_modules
 COPY server/ ./server/
 COPY db/ ./db/
+RUN cd server && npm run build
```
shayneczyzewski, line 50: I couldn't run this in `server-builder` since I need the source code, which is copied over here. I didn't want to fundamentally restructure this Dockerfile, and instead do the smallest change that will work, since we will rethink this process in the near future.

Martinsos: I don't quite understand this!
So we need to build a TS project. Why not build it in server-builder? I guess you would need to copy some extra files then, but I guess that wouldn't be too hard? I don't think this is too bad either, it is probably a small impact, but I am curious.

Martinsos (approved):
Approved but also left a small comment!

waspc/data/Generator/templates/Dockerfile
```diff
@@ -47,6 +47,7 @@ WORKDIR /app
 COPY --from=server-builder /app/server/node_modules ./server/node_modules
```
infomiho, line 47: Just out of curiosity, I've noticed this line in the `Dockerfile`

```
RUN apk add --no-cache python3
```

Is there a valid reason to install `python3` or is this some sort of leftover?

waspc/data/Generator/templates/Dockerfile
infomiho: There is an added benefit if we did it in the `server-builder` step, we could copy only the `node_modules` and `dist` folder to the next step (avoiding copying over the `src` and everything else).

But we should still separate the copying of `src` and `package*.json` files to keep the layer caching benefits.

waspc/data/Generator/templates/Dockerfile
Martinsos: We installed it because we had people doing stuff where they needed python3, and it was missing, and it is such a common dep that people need that we decided to just ship with it.

waspc/data/Generator/templates/Dockerfile
shayneczyzewski: Ok ok, you all convinced me, I will just refactor it a bit to do all the building where it probably should. The 1 line fix was just so appealing though :D

waspc/data/Generator/templates/Dockerfile
```diff
@@ -25,16 +25,15 @@ RUN apk add --no-cache openssl1.1-compat
 FROM base AS server-builder
 RUN apk add --no-cache build-base libtool autoconf automake
 WORKDIR /app
+COPY server/ ./server/
 # Install npm packages, resulting in node_modules/.
-{=# usingServerPatches =}
-COPY server/patches ./server/patches
-{=/ usingServerPatches =}
-COPY server/package*.json ./server/
```
shayneczyzewski, line 32: We no longer have to selectively, or conditionally in the case of patches, copy a subset of the `server` dir over. Just copy it all, since we need the source later anyways for TS compilation.

waspc/data/Generator/templates/Dockerfile
```diff
@@ -45,7 +44,8 @@ RUN apk add --no-cache python3
 ENV NODE_ENV production
 WORKDIR /app
 COPY --from=server-builder /app/server/node_modules ./server/node_modules
-COPY server/ ./server/
+COPY --from=server-builder /app/server/dist ./server/dist
+COPY --from=server-builder /app/server/package*.json ./server/
```
shayneczyzewski, line 48: In addition to the `dist` dir, we also need the `package.json` files to run the `npm run` command.

shayneczyzewski (commented):
Ok @Martinsos and @infomiho, made your updates (plus one small other cleanup). Feel free to do a final check if you'd like. Thanks!

waspc/data/Generator/templates/Dockerfile
```diff
@@ -25,16 +25,15 @@ RUN apk add --no-cache openssl1.1-compat
...
-{=# usingServerPatches =}
-COPY server/patches ./server/patches
-{=/ usingServerPatches =}
-COPY server/package*.json ./server/
 RUN cd server && npm install
 {=# usingPrisma =}
 COPY db/schema.prisma ./db/
 RUN cd server && {= serverPrismaClientOutputDirEnv =} npx prisma generate --schema='{= dbSchemaFileFromServerDir =}'
 {=/ usingPrisma =}
+# Compile TS (depends on Prisma, if in use)
```
infomiho, line 35: Nit: This comment feels a bit off

The fact that we are compiling TS here, it's not really important, we are building the server in general. If I understood the comment correctly, you wanted to convey the idea "we need to put this AFTER the `prisma generate` bit". Maybe write it like that 🤷

infomiho (approved):
Looking good, I'll test it out locally as well.

waspc/data/Generator/templates/Dockerfile
shayneczyzewski: Good point, I can drop the compile/build note and just say we need to do this after Prisma entities were generated. 👍🏻 Will update comment shortly

Martinsos (approved):
LGTM! Nice job!

PR #1009, opened 2023-02-16
Adds support for root component

### LINKED ISSUE #77: Enable custom wrapping of top-lvl React component

infomiho:
### Root component
Fixes #77

This PR introduces `rootComponent` in the `app.client` dictionary in Wasp lang which will help users to define their root React component.

`rootComponent` is of type `ExtImport` which is an import statement from user defined files. 

This enables the user to do:
1. Define a `src/client/App.jsx` which exports a component
2. Use the component to setup various providers e.g. `Redux` or `ChakraProvider` https://chakra-ui.com/getting-started

We then use their root component to wrap the `router` in the app's `index.js`.


waspc/data/Generator/templates/react-app/src/index.js
```diff
@@ -34,7 +37,12 @@ async function render() {
   const queryClient = await queryClientInitialized
   ReactDOM.render(
     <QueryClientProvider client={queryClient}>
-      { router }
+      {=# rootComponent.isDefined =}
+      <{= rootComponent.importIdentifier =}>{router}</{= rootComponent.importIdentifier =}>
+      {=/ rootComponent.isDefined =}
+      {=^ rootComponent.isDefined =}
+      {router}
+      {=/ rootComponent.isDefined =}
```
Martinsos, line 45: Nitpick: What about:

```suggestion
      {=# rootComponent.isDefined =}
      <{= rootComponent.importIdentifier =}>
      {=/ rootComponent.isDefined =}
        {router}
      {=# rootComponent.isDefined =}
      </{= rootComponent.importIdentifier =}>
      {=/ rootComponent.isDefined =}
```

is this maybe nicer / more composable?

Martinsos (approved):

Martinsos:
Nice job -> your first extension of Wasp language itself :D!

Oh one thing is missing though -> docs, right? Let's add them in this PR, no reason to do them in separate PR.

Martinsos (changes requested):
Let's add docs?

infomiho:
I've updated the docs in the meantime, check the last commit 🙂

web/docs/language/features.md
```diff
@@ -1377,20 +1377,69 @@ This function should return the user fields to use when creating a new user upon
...
+}
+
+function App({ children }) {
+  return (
+    <div>
+      <h1>My App</h1>
+      {children}
+    </div>
+  )
+}
```
Martinsos, line 1440: WHy do we need this extra App here? Can't we just have `Root` render children directly in Provider component and that is it?

Martinsos (approved):
LGTM! Some small comments, resolve as you find fit and you can merge!

web/docs/language/features.md
infomiho: I wanted to offer a more elaborate example where we have the `Provider` wrapping and some `<App />` compnent (which is how you'll see it's done in the Redux docs).

This also works:
```jsx
export default async function Root({ children }) {
  return (
    <Provider store={store}>
      {children}
    </Provider>
  )
}
```

waspc/data/Generator/templates/react-app/src/index.js
infomiho: It will be easier to add new user defined wrappers, but will we have the need to add them in the future?

Your approach has extra whitespace in the output when not using `rootComponent` e.g. `   {router}` 😂

web/docs/language/features.md
Martinsos: Ok! It is good to give exapmles, but I sometimes have the need to reduce them if possible, to make maintainance simpler. But this is probably ok it is not so complex. Unless this might confused them -> do I have to use this App inside this thing? I don't know, your call!

waspc/data/Generator/templates/react-app/src/index.js
Martinsos: Yeah I know but I gave up on formatting, we can't do it 100% anyway.

I think what I wrote is more explicit about "we are just wrapping it into this".

web/docs/language/features.md
infomiho: I'll provide two use-case examples (Layout + Providers) and a one more elaborate where both are coming together.

waspc/data/Generator/templates/react-app/src/index.js
infomiho: Makes sense, I'll probably need to give up on output formatting as well 😢 

Maybe in the future, one of our build steps can be running everything through Prettier auto-format and then the users will get super nice output :D

Martinsos: yes that is probably what we will want to do! Or we will give up on it completely. But I like the idea of prettier.

web/docs/language/features.md
```diff
@@ -1377,20 +1377,69 @@ This function should return the user fields to use when creating a new user upon
...
+      {children}
+      <footer>
+        <p>My App footer</p>
+      </footer>
+    </div>
+  )
+}
+```
+
+You can use it to setup various providers that your application needs:
```
sodic, line 1419:
```suggestion
You can use it to set up various providers that your application needs:
```

web/docs/language/features.md
infomiho: Updated here: https://github.com/wasp-lang/wasp/pull/1009/commits/207806c7cedd0093bfe9682f4a201d82fff4edd2

waspc/data/Generator/templates/react-app/src/index.js
infomiho: Updated here: https://github.com/wasp-lang/wasp/pull/1009/commits/207806c7cedd0093bfe9682f4a201d82fff4edd2

waspc/src/Wasp/Generator/WebAppGenerator.hs
```diff
@@ -246,16 +247,28 @@ genIndexJs spec =
...
     maybeSetupJsFunction = AS.App.Client.setupFn =<< AS.App.client (snd $ getApp spec)
-    maybeSetupJsFnImportDetails = getJsImportStmtAndIdentifier relPathToWebAppSrcDir <$> maybeSetupJsFunction
-    (maybeSetupJsFnImportStmt, maybeSetupJsFnImportIdentifier) =
-      (fst <$> maybeSetupJsFnImportDetails, snd <$> maybeSetupJsFnImportDetails)
+    maybeRootComponent = AS.App.Client.rootComponent =<< AS.App.client (snd $ getApp spec)
 
-    relPathToWebAppSrcDir :: Path Posix (Rel ()) (Dir C.WebAppSrcDir)
-    relPathToWebAppSrcDir = [reldirP|./|]
+mkTmplDataForExtImport :: Maybe ExtImport -> Aeson.Value
+mkTmplDataForExtImport maybeExtImport = maybe notDefinedValue mkTmplData maybeExtImport
```
sodic, line 259: This function encapsulates a common pattern.

If I understand correctly, we're using it for all ext imports. Do you think it makes sense to put it in a more public place and reuse?

waspc/src/Wasp/Generator/WebAppGenerator.hs
```diff
@@ -246,16 +247,28 @@ genIndexJs spec =
...
+  where
+    notDefinedValue = object ["isDefined" .= False]
+
+    mkTmplData :: ExtImport -> Aeson.Value
+    mkTmplData extImport =
+      let jsImportDetails = getJsImportStmtAndIdentifier relPathToWebAppSrcDir extImport
+          (jsImportStmt, jsImportIdentifier) = jsImportDetails
```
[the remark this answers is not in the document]
infomiho: That's true, nice catch 👍

waspc/src/Wasp/Generator/WebAppGenerator.hs
infomiho: This can be extracted to a common function in `Wasp.Generator.JsImport` and then it can be specialized in `Wasp.Generator.WebAppGenerator.JsImport` and `Wasp.Generator.ServerGenerator.JsImport`.

In the web app templates, this pattern is actually only used in this file, but in the server templates it's used more.

I'll extract the pattern in this PR and apply it to `WebApp` only. I'll do the same for server templates in a separate PR.

waspc/src/Wasp/Generator/WebAppGenerator.hs
```diff
@@ -246,16 +247,28 @@ genIndexJs spec =
...
+       in object
+            [ "isDefined" .= True,
+              "importStatement" .= jsImportStmt,
+              "importIdentifier" .= jsImportIdentifier
+            ]
+
+relPathToWebAppSrcDir :: Path Posix (Rel ()) (Dir C.WebAppSrcDir)
```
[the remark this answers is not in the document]
infomiho: You are correct, this seems like we should use polymorphism here. I've gone with a parameter name of `importLocation` to make it more informative (which I liked the `ImportLocation` approach as well).

waspc/src/Wasp/Generator/WebAppGenerator.hs
infomiho: We now have:
- `Wasp.Generator.JsImport` -> `jsImportToTmplData` which generates approprisate `JSON` for usage in templates when we need to do something with import statements
- `Wasp.Generator.WebAppGenerator.JsImport` -> `extImportToJsImportTmplData` which uses the above method to convert an `ExtImport` to `JSON` template data

We should add in a separate PR:
- `Wasp.Generator.ServerGenerator.JsImport` -> `extImportToJsImportTmplData`
- use the new function to refactor the way use generate template data for JS import statements

waspc/src/Wasp/Generator/WebAppGenerator.hs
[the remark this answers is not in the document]
sodic: Seems like `relPathToWebAppSrcDir` wasn't fixed.

waspc/src/Wasp/Generator/WebAppGenerator.hs
sodic: To keep in line with naming conventions for functions (i.e., they should be verbs) and what we already have (like this guy [here](https://github.com/wasp-lang/wasp/blob/868362d103469b9cfc6b1ed5b66c26eae7e33da5/waspc/src/Wasp/Generator/Common.hs#L58)*), I suggest renaming this to something like `makeImportJson`.

The "functions should be verbs" applies to other places as well.

\* Although that name could be improved too :)

infomiho: We have a number of these "tranform" functions in our codebase, and I kinda like this convetion of `<A>To<B>` naming.

Some examples I found in our codebase:
- `kebabToCamelCase`
- `sourceSpanToRegion`
- `lspPositionToOffset`
- `extImportToJsImport` (that's mine)

What about `extImportToImportJson`?

waspc/src/Wasp/Generator/WebAppGenerator.hs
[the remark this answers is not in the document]
infomiho: Right you are, I fixed it in all other places where `ImportLocation` was mentioned, but not the exact spot where you commented 🤦‍♂️

waspc/src/Wasp/Generator/WebAppGenerator.hs
sodic: That's true, we have several of those.

Anyway, choose whatever you like best.

Purely FYI, this is my order of preference:
1. `makeExtImportJson`
2. `extImportToImportJson`
3. `extImportToJsImport`

PR #595, opened 2022-05-11
Support for the client setup function

sodic:
Fixes #593

PR #1014, opened 2023-02-20
Fixes user query cache invalidation

infomiho:
With the usage of `useAuth` in the root component, a bug appeared: we didn't invalidate `/auth/me` query properly. 

The fix is one line, but in this PR I've updated the example app a bit (logout button in the header, some styling changes, nothing major).

### Before (look at the logout button)

https://user-images.githubusercontent.com/2223680/220100901-2e0339a6-6405-406f-a06e-a01465d9becd.mov

### After

https://user-images.githubusercontent.com/2223680/220086977-aec69c1d-fdc8-4cb3-a310-12dcad788e30.mov

### DIFF
--- a/waspc/data/Generator/templates/react-app/src/auth/login.js
+++ b/waspc/data/Generator/templates/react-app/src/auth/login.js
@@ -1,4 +1,4 @@
-import { removeQueries } from '../operations/resources'
+import { invalidateAndRemoveQueries } from '../operations/resources'
 import api, { setAuthToken, handleApiError } from '../api.js'
 
 export default async function login(username, password) {
@@ -9,7 +9,7 @@ export default async function login(username, password) {
     setAuthToken(response.data.token)
     // This isn't really neccessary because we remove all private queries after
     // logout, but we do it to be extra safe.
-    // 
+    //
     // For example, in future versions, users might be able to get to an SPA
     // login page while there's an active session. This code will prevent data
     // leaks in such cases.
@@ -17,7 +17,7 @@ export default async function login(username, password) {
     // TODO(filip): We are currently removing all the queries, but we should
     // remove only non-public, user-dependent queries - public queries are
     // expected not to change in respect to the currently logged in user.
-    await removeQueries()
+    await invalidateAndRemoveQueries()
   } catch (error) {
     handleApiError(error)
   }
--- a/waspc/examples/todoApp/src/client/App.jsx
+++ b/waspc/examples/todoApp/src/client/App.jsx
@@ -1,35 +1,32 @@
-import logout from "@wasp/auth/logout.js";
-import useAuth from "@wasp/auth/useAuth.js";
+import logout from '@wasp/auth/logout.js'
+import useAuth from '@wasp/auth/useAuth.js'
 
-import "./Main.css";
+import './Main.css'
 
 export function App({ children }) {
-    const { data: user } = useAuth();
+  const { data: user } = useAuth()
 
-    return (
-        <div className="app border-spacing-2 p-4">
-            <header className="flex justify-between">
-                <h1 className="font-bold text-3xl mb-5">
-                    <a href="/">ToDo App</a>
-                </h1>
-                {user && (
-                    <div className="flex gap-3 items-center">
-                        <div>
-                            Hello, <a href="/profile">{user.username}</a>
-                        </div>
-                        <div>
-                            <button
-                                className="btn btn-primary"
-                                onClick={logout}
-                            >
-                                Logout
-                            </button>
-                        </div>
-                    </div>
-                )}
-            </header>
-            <main>{children}</main>
-            <footer className="mt-8 text-center">Created with Wasp</footer>
-        </div>
-    );
+  return (
+    <div className="app border-spacing-2 p-4">
+      <header className="flex justify-between">
+        <h1 className="font-bold text-3xl mb-5">
+          <a href="/">ToDo App</a>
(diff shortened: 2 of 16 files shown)


waspc/data/Generator/templates/react-app/src/auth/login.js
```diff
@@ -9,15 +9,15 @@ export default async function login(username, password) {
...
+    //
     // For example, in future versions, users might be able to get to an SPA
     // login page while there's an active session. This code will prevent data
     // leaks in such cases.
     //
     // TODO(filip): We are currently removing all the queries, but we should
     // remove only non-public, user-dependent queries - public queries are
     // expected not to change in respect to the currently logged in user.
-    await removeQueries()
+    await invalidateAndRemoveQueries()
```
infomiho, line 20: By only removing the query, we removed the `Observer`s from all the components that used `user` object, and then the component didn't rerender probably. 

Invalidating and *then* removing the query, makes sure that the `Observer`s are recreated in the components and rerenders happen.

waspc/examples/todoApp/src/client/App.jsx
```diff
@@ -1,10 +1,35 @@
...
+    const { data: user } = useAuth();
+
     return (
-        <div className="p-6">
-            <header className="mb-6">
-                <h1 className="text-3xl font-bold">ToDo App</h1>
+        <div className="app border-spacing-2 p-4">
+            <header className="flex justify-between">
+                <h1 className="font-bold text-3xl mb-5">
+                    <a href="/">ToDo App</a>
```
infomiho, line 13: There is an issue with the usage of `rootComponent` and navigation around the app. 

We are rendering it like this:
```
<App>{router}</App>
```
This means that the `App` component is outside the `<Router>...</Router>` which prevents us from using `<Link />` component in the `rootComponent` which is the proper way to navigate around the app which prevent full page reload.

Related issue: https://github.com/wasp-lang/wasp/issues/1015

waspc/examples/todoApp/src/client/Todo.tsx
```diff
@@ -1,193 +1,216 @@
-import React, { useState, FormEventHandler, ChangeEventHandler } from 'react'
```
infomiho, line 1: All these changes are just `Prettier` formatting, safe to skip.

waspc/examples/todoApp/src/client/Main.css
```diff
@@ -9,13 +9,17 @@
     a {
         @apply underline text-blue-600 hover:text-blue-800 visited:text-purple-600;
     }
-    input[type='submit'] {
-        @apply btn btn-blue;
+    input[type="text"],
+    input[type="password"] {
+        @apply rounded;
```
infomiho, line 14: Tiny style changes, making the example app look a tiny bit nicer.

waspc/examples/todoApp/src/client/Todo.tsx
sodic: The thing is: we doesn't use semicolons, and we use two spaces for indentation. I'm not necessarily a fan of either of these choices, but it's better to stick with conventions :sweat_smile:

waspc/data/Generator/templates/react-app/src/auth/login.js
sodic: I think this comment is worth adding to the code (inside the `invalidateAnndRemoveQueries` function). You can even link this explanation in there: https://github.com/wasp-lang/wasp/pull/595#discussion_r877364290

waspc/examples/todoApp/src/client/pages/Main.js
```diff
@@ -1,18 +1,7 @@
...
-    <>
-      <button className='btn btn-blue' onClick={logout}>
-        Logout
-      </button>
-      <Todo />
-    </>
-  )
-}
+    return <Todo />;
+};
```
sodic, line 5: If you want to clean up even further, you can completely remove this file (as it boils down to `export default Todo from "../Todo"`).

sodic (commented):
Looks good, minor comments.

waspc/examples/todoApp/src/client/Todo.tsx
infomiho: We should then define a `Prettier` config to make it explicit which conventions we expect. 

I can revert this change, ofc, since it's very aggressive and out-of-place.

waspc/examples/todoApp/src/client/pages/Main.js
infomiho: One of those is a `page` and the other is a `component`, I think it's okay to keep the distinction and have a place in the future to put some content besides the `<Todo />` component.

waspc/examples/todoApp/src/client/Todo.tsx
infomiho: I'll add the `prettier.config.js` so it's explicit :)

waspc/data/Generator/templates/react-app/src/auth/login.js
infomiho: Should we consider removing the `removeQueries` function since it's not used anywhere anymore? It will only confuse us in the future.

infomiho: I've updated the comment and removed `removeQueries` since we have no use for it right now and it's a potential footgun.

sodic (approved):

Issue #268, opened 2021-06-28
Allow defining custom API routes (http)


Martinsos:
Right now dev can define operations (actions, queries) which are then consumed from client via RPC that works via http.
However, they can't define custom http API roues at the moment! They might need this, so we should offer them support for this.

This basically means they should be able to drop one level of abstraction below the operations and directly define http routes (in Express right now).

Martinsos:
There are couple approaches to consider here:
1. We could somehow extend Operations in a way to provide access to headers in them -> basically give more control in them. Although there is probably a limit to what we can do here, we can't go completely custom, but maybe access to headers is already a big thing?
2. We can give them access to wasp express app / routes in app.server.setupFn, so they can define their own routes there.
3. We can provide a mechanism for injecting express middleware into server app in general.
4. We can have special mechanism directly for defining custom HTTP routes.

4 is probalby the strongest solution, 2 is a quick way out, while 1 and 2 are probably useful in their own regard but don't address the problem directly / completely.

Martinsos:
Regarding middleware -> we talked a lot about it in some other capacities, e.g. Permissions, so I will also link to that https://github.com/wasp-lang/wasp/issues/584 . @shayneczyzewski was big part of those previous conversations so also might have something to say here.

Martinsos:
Here is request from a Waspeteer that is related to this:
> Trying to implement uploading files for my app. Has anyone tried doing this with wasp? Is there a good way to do this within wasp? Any good libraries?
>
> Everything seems to go through multer, however I don’t see how we can add express middleware through wasp file or pass through headers to actions. Unless I am missing something?

shayneczyzewski:
I think we all agree our main choices are escape hatch into Express (super easy, but lose lots of info), or we provide Wasp syntax.

The Wasp syntax could take many forms, but one quick one that may be nice is if we just provided them a way to say something like:
```c
post { path: "/foo/<id:integer>", to: import { sayHi } from "@server/actions.js" }
```

So we would let them specify the HTTP verb as a keyword, and in the path we can also encode some type info. This may also dovetail into the top level data schema issue https://github.com/wasp-lang/wasp/issues/642

Martinsos:
One more idea: What if we would allow them to define Operations with custom API?

Think about it as a hybrid between custom API routes and normal Operations -> Custom Operation.

[...]

PR #1017, opened 2023-02-20
Add `app` and `server` to server `setupFn`

shayneczyzewski:
### Description

This PR adds `app` and `server` to the server `setupFn`, mainly as an escape hatch. Users can then do arbitrary Express endpoints, or things like websockets.

It is the start of #268, but the real finish there is adding some proper `api` DSL support.


waspc/examples/todoApp/src/server/serverSetup.js
```diff
@@ -5,7 +5,8 @@ let someResource = undefined
 
 export const getSomeResource = () => someResource
 
-const setup = async () => {
+const setup = async (app, _server) => {
```
infomiho, line 8: I would recommend adding a type for the `ServerSetupFn` like:

```ts
type ServerSetupFn = (context: ServerSetupFnContext) => void;

type ServerSetupFnContext = {
  app: express.Application,
  server: http.Server,
};
```
which the user can import like `import { ServerSetupFnContext } from '@wasp/types';` or similar.

shayneczyzewski: Yep, can do. 👍🏻 This is just a WIP skeleton at the moment but will beef it out shortly before tagging for review.

waspc/data/Generator/templates/server/src/types/index.ts
```diff
@@ -0,0 +1,13 @@
+import express from 'express'
+import http from 'http'
+
+export type Application = express.Application
+
+export type Server = http.Server
```
shayneczyzewski, line 6: Thoughts on exporting type aliases like these? I would like people to be able to pass around/use the components of `ServerSetupFnContext` and still have proper typing, but _without_ having to import the actual `express`/`node` imports themselves. I think it makes it nice they can get it all from the same import path, but curious on the TS expert pros/cons here. The downside is they will still need to know it is actually an `express.Application` if they want to use it with other libraries, look up docs, etc. but I hope that is an ok use for aliases? Thanks!

shayneczyzewski: For example, see `function addCustomRoute(app: Application)` in `serverSetup.ts`.

waspc/examples/todoApp/src/server/serverSetup.js
shayneczyzewski: Added some types, thanks!

waspc/data/Generator/templates/server/src/server.ts
```diff
@@ -21,12 +22,13 @@ const startServer = async () => {
   const port = normalizePort(config.port)
   app.set('port', port)
 
+  const server = http.createServer(app)
+
   {=# doesServerSetupFnExist =}
-  await {= serverSetupJsFnIdentifier =}()
+  const serverSetupFnContext: ServerSetupFnContext = { app, server }
```
infomiho, line 28: @shayneczyzewski Why are we exposing the `server` object? I'm not sure for which use case it's needed

infomiho: Makes sense, I though that `socket.io` didn't need the server 🤦‍♂️

Martinsos: When I was doing this for an experiment, I did it as described here: https://discord.com/channels/686873244791210014/920312576133443634/1070052697572638870 -> so I passed `app` to setupFn, but I didn't move the server creation above it and then pass it.

I think it is cool that you did, that gives people more stuff to work with!

But it does change behavior a bit -> before, `setupFn` was allowing you to do stuff before `server` is created. Now, it allows you to set up stuff after server is created. I have no idea if we loose anything with that? Is there a use case where it might be important to do something before `server` is created from `app`? If we are not sure now, that is ok, we can just go with this and if we learn there is use case for it, we can add another function called `preSetupFn` or smth, that will be called just before the server.

One thing I thought was important was to add routes to `app` before the `server(app)` is called, but I guess I was wrong because I am sure you tested this and it works, even though you add stuff to `app` after the `server(app)` is called. Interesting. But does it work because it should, or by accident?

Maybe we can get some answers here: https://nodejs.org/api/http.html#class-httpserver .

waspc/data/Generator/templates/server/src/types/index.ts
infomiho: I like the reexport in our context, so users can import from `@wasp/types` 👍 

I would maybe write it a bit differently:
```
import { Application } from 'express';
import { Server } from 'http';

export type ServerSetupFn = (context: ServerSetupFnContext) => Promise<void>

 export type ServerSetupFnContext = {
   app: Application,
   server: Server,
 }
 
 export { Application } from 'express';
 export { Server } from 'http';
```

Martinsos: Ok I will leave this to @sodic and @infomiho .
I need to take some time these days to learn TS but it is not yet that day :D!

Martinsos (commented):
@shayneczyzewski this is it from me, I will let @sodic or @infomiho handle the TS side of this!

waspc/data/Generator/templates/server/src/types/index.ts
shayneczyzewski: Sounds good, thanks @infomiho for a better convention to use!

waspc/data/Generator/templates/server/src/server.ts
shayneczyzewski: Good question @Martinsos - I don't think there is any technical reason why we have to keep it after, but you are right it does change the previous implied semantics a bit from user-level docs (which I massaged to say "before the server starts accepting requests"). But I don't think there is any actual technical impact in doing this swap. The custom route I setup in the `setupFn` still worked even when `createServer` happened before. I am not aware of any other potential impacts of this.

I believe `createServer` is just registering a request handler callback (of which an Express instance itself is). That callback should be using things internally that can be set either before or after that registration (i.e. it shouldn't even know it was registered to me). But I will do more testing and digging to be 100% sure (including doing a Socket.IO test too).

web/docs/language/features.md
```diff
@@ -1537,11 +1537,34 @@ app MyApp {
...
 #### `setupFn: ServerImport` (optional)
 
-`setupFn` declares a JS function that will be executed on server start. This function is expected to be async and will be awaited before server continues with its setup and starts serving any requests.
+`setupFn` declares a JS function that will be executed on server start. This function is expected to be async and will be awaited before the server starts accepting any requests.
 
-It gives you an opportunity to do any custom setup, e.g. setting up additional database or starting cron/scheduled jobs.
+It gives you an opportunity to do any custom setup, e.g. setting up additional database/websockets or starting cron/scheduled jobs.
 
-The javascript function should be async, takes no arguments and its return value is ignored.
+This async JS function may accept a context object containing an `app` and `server` and should return nothing. In short, it should conform to a TypeScript signature like the following:
```
infomiho, line 1544:
```suggestion
The `setupFn` function receives the `express.Application` and `http.Server` as part of its context. The `app` and `server` instances can be useful for any custom server routes or setting up `socket.io`.
```

I would maybe mention `socket.io` for anyone searching through the docs.

waspc/data/Generator/templates/server/src/server.ts
shayneczyzewski: It's not pretty (we have to specify the client/server URLs since different domains) but you can get Socket.IO up and running with access to `server`: https://gist.github.com/shayneczyzewski/165b4266d77a2a75454f605dd55b5db7

PR #1021, opened 2023-02-22
Add missing dotenv dependency

infomiho:
This didn't happen before while testing the `env` PR (https://github.com/wasp-lang/wasp/pull/994), but running the `wrun wasp-cli start` with the `main` branch now throws this:
```
Error: Cannot find module 'dotenv/config'
```

Adding the `dotenv` dependency explicitly fixes the issue.


Martinsos (approved):
LGTM, weird that we didn't have it before!?

PR #1024, opened 2023-02-22
Invalidate queries after oAuth login

infomiho:
Redirect to `onAuthSucceededRedirectTo` happens before `auth/me` query is invalidated.

### DIFF
--- a/waspc/data/Generator/templates/react-app/src/auth/helpers/user.ts
+++ b/waspc/data/Generator/templates/react-app/src/auth/helpers/user.ts
@@ -1,0 +1,14 @@
+import { setAuthToken } from '../../api.js'
+import { invalidateAndRemoveQueries } from '../../operations/resources'
+
+export async function setupAuth({ token }: { token: string }): Promise<void> {
+    setAuthToken(token)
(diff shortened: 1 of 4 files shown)


waspc/data/Generator/templates/react-app/src/auth/helpers/user.ts
```diff
@@ -0,0 +1,14 @@
+import { setAuthToken } from '../../api.js'
+import { invalidateAndRemoveQueries } from '../../operations/resources'
+
+export async function setupAuth({ token }: { token: string }): Promise<void> {
```
infomiho, line 4: I wasn't really sure which name to use here, this function is setting up our auth/user/session after login.

- `setupAuth`
- `setupUser`
- `prepareAuth`
- `initSession`

infomiho: Also, if the file should be `helpers/user.ts` or `helpers/auth.ts` 🤷

sodic: I think `initSession`, `initializeSession`, or `startSession` are the best.

PR #1027, opened 2023-02-23
Increase clickable area for auth buttons

infomiho:
Padding was applied to the outer `div` and not the actionable `a` tag. This results in users clicking on the edge of the button, and nothing happening 😢

### Before (clickable area)

Screenshot 2023-02-23 at 14 26 38

### After

Screenshot 2023-02-23 at 14 26 11


matijaSos (approved):
lgtm

PR #1030, opened 2023-02-23
Adds new complex e2e test

### LINKED ISSUE #701: Add e2e test for a complex app with all features
**Is your feature request related to a problem? Please describe.**
Originally, the e2e golden tests were meant to test basic sequences of CLI command outputs on disk. They then started testing specific features, like Jobs (but via a hacky `printf` command line mechanism for creating/updating files).

However, we'd like to have some way to make sure all the kitchen sink features remain stable, and e2e should ideally support this. What we would like to do for this case is create a golden test for a complex app where we can add pretty much every feature to it, but while it is all already on disk before the test starts. This would not test so much CLI commands themselves, but rather, the compiled output of a complex project. This would be helpful as we modify the parser, for example, or just want to make sure features remain stable as we refactor, and it would scale better than one sample project per feature under test.

NOTE: As stated, we would need to create the golden dir on disk first. The existing tests create them all with `wasp new`. We would not do that, but instead, only test the output of `wasp compile`.  However, I think there are some directory creation/naming assumptions currently baked into `runGoldenTest`, and the fact that it assumes we are creating things on the fly. Would need to test it out some to figure it out and probably refactor it a bit.

infomiho:
Closes #701

This is an e2e test that does the following:
- creates a new Wasp project
- creates a `.env.server` file with variables needed for Google auth
- adds dependencies (`redux` and `react-redux`)
- adds client setup (setup fn and root component) and server setup
- adds auth with Google
- adds a job (sets PostgreSQL as db system)
- adds one action and one query
- compiles the project

Are there any other features we need to cover?


waspc/e2e-test/ShellCommands.hs
```diff
@@ -71,11 +72,14 @@ createFile content relDirFp filename = return $ combineShellCommands [createPare
 -- NOTE: This is fragile and will likely break in future. Assumes `app` decl is first line and by default
 --       we do not have a `db` field. Consider better alternatives.
 setDbToPSQL :: ShellCommandBuilder ShellCommand
-setDbToPSQL =
+-- Change DB to postgres by adding string at specific line so it still parses.
+setDbToPSQL = insertLineAtWaspFile "  db: { system: PostgreSQL }," 2
+
+insertLineAtWaspFile :: String -> Int -> ShellCommandBuilder ShellCommand
+insertLineAtWaspFile line atLineNumber =
```
shayneczyzewski, line 79: Nice extraction here 👍🏻

shayneczyzewski: In the future, is the plan to still possibly do the wasp file manipulation in memory in Haskell? Where there any insights/blockers you thought of for the future on that one we may want to document now while still fresh?

waspc/e2e-test/Util.hs
```diff
@@ -0,0 +1,8 @@
+module Util where
+
+import Control.Applicative (liftA2)
+
+infixr 5 <++>
+
+(<++>) :: Applicative f => f [a] -> f [a] -> f [a]
```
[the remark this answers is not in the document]
Martinsos: I guess they could have it by exposing it in Cabal?

Martinsos (approved):
LGTM! Left some minor comments.

waspc/e2e-test/ShellCommands.hs
infomiho: Even manipulating it as a in-memory string will be an improvement :)

@shayneczyzewski and I started talking about doing it because I thought I had some concurency issues and some things got written to the `.wasp` file and some didn't. I thought that's due to doing the file manipulation via shell commands.

But that wasn't the casue, I had a bug in a different place (running `wasp compile` in the middle of the process) and that messed things up.

Things work as expected, `.wasp` file is consturcted properly even with shell commands. That's why I decided to keep the improvement of doing it in-memory for some future PR :)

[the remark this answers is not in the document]
infomiho: Great suggestions, I've made the params swap and renamed the function.

PR #1033, opened 2023-02-27
`release` ->  `main` (0.8.2)

sodic:


sodic:
Forgot we were doing this manually from the CLI.

PR #1043, opened 2023-03-03
Adds `api` for defining custom HTTP routes

### LINKED ISSUE #268: Allow defining custom API routes (http)

shayneczyzewski:
### Description

Adds new syntax for defining a new API endpoint:
```c
api fooBar {
  fn: import { fooBar } from "@server/apis.js",
  entities: [Task],
  httpRoute: (GET, "/foo/bar")
  // auth: <bool>
}
```

The JS function gets access to req/res, and a typed context (with entities):
```ts
import { FooBar } from '@wasp/apis/types'

export const fooBar : FooBar = (req, res, context) => {
  res.set('Access-Control-Allow-Origin', '*') // Example of modifying headers to override CORS middleware.
  res.json({ msg: `Hello, ${context.user?.username || "unknown"}!` })
}
```

This also gives them an `auth` field and access to `context.user` like Operations do.

### Notes:
- Right now they still use our default middleware stack. In the future we can try to name them and make it easier to modify. In the meantime they can manipulate headers of `res` as needed.
- We just pass the `path` straight through, so they can use anything Express supports. But we don't do any path/query/body parsing for them.

### TODO:
- [x] Update docs
- [x] Deploy test
- [x] Update changelog
- [x] Add e2e updates

Fixes #268

Also adds syntax highlighting to VS Code extension here: https://github.com/wasp-lang/vscode-wasp/pull/9


waspc/examples/todoApp/src/server/apis.ts
```diff
@@ -0,0 +1,7 @@
+import { Request, Response } from '@wasp/types'
+import { fooBarContext } from '@wasp/apis/types'
+
+export function fooBar(req: Request, res: Response, context: fooBarContext) {
```
shayneczyzewski, line 4: What do people think of providing a context type (`waspc/data/Generator/templates/server/src/apis/types.ts`) instead of some wrapper function?

waspc/data/Generator/templates/server/src/routes/index.js
```diff
@@ -16,5 +17,6 @@ router.get('/', function (req, res, next) {
 router.use('/auth', auth)
 {=/ isAuthEnabled =}
 router.use('/{= operationsRouteInRootRouter =}', operations)
+router.use(apis)
```
shayneczyzewski, line 20: By putting at the end, it should not be allowed to interfere with any of our routes defined before it. Route setting in Express appears to be first set wins.

waspc/examples/todoApp/src/server/apis.ts
shayneczyzewski: Also, should we uppercase the start of the typename?

Martinsos: @sodic will probably be the best person to give advice here!

waspc/data/Generator/templates/server/src/routes/apis/index.js
```diff
@@ -0,0 +1,25 @@
...
+import { handleRejection } from '../../utils.js'
+
+{=# apiRoutes =}
+{=& importStatement =}
+{=/ apiRoutes =}
+
+const router = express.Router()
+
+{=# apiRoutes =}
+router.{= routeMethod =}('{= routePath =}', handleRejection((req, res) => {
```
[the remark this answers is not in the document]
Martinsos: Looks good to me!

waspc/data/Generator/templates/server/src/routes/index.js
Martinsos: Oh intersting, ok! Maybe we could add a comment here explaining that? if you think it is redundant, ignore this comment.

waspc/src/Wasp/AppSpec/Api.hs
```diff
@@ -0,0 +1,32 @@
...
+
+instance IsDecl Api
+
+data HttpRoute = HttpRoute
+  { method :: HttpMethod,
+    path :: String
+  }
+  deriving (Show, Eq, Data)
+
+data HttpMethod = ALL | GET | POST | PUT | DELETE
```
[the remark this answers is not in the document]
Martinsos: I think these are enough for now -> we can always add more later if needed.

waspc/examples/todoApp/todoApp.wasp
```diff
@@ -105,6 +105,15 @@ query getTasks {
   entities: [Task]
 }
 
+api fooBar {
+  fn: import { fooBar } from "@server/apis.js",
+  entities: [Task],
+  httpRoute: {
```
Martinsos, line 111: We could use a tuple here if you like it better, so it is:
```
  httpRoute: (GET, "/foo/bar")
```

We use it for npm deps, so you can take a look at them to see how it is implemented.

But I am also ok with the dict if you like it better!

waspc/examples/todoApp/src/server/apis.ts
infomiho: Yep, the convention is to uppercase the types in Typescript. 

> What do people think of providing a context type (waspc/data/Generator/templates/server/src/apis/types.ts) instead of some wrapper function?

Could you provide code examples, I'm not 100% I understand what you mean?

waspc/src/Wasp/Generator/ServerGenerator/ApiRoutesG.hs
```diff
@@ -0,0 +1,68 @@
...
+    getTmplData (name, api) =
+      object
+        [ "name" .= name,
+          "entities" .= getApiEntitiesObject api
+        ]
+
+getApiEntitiesObject :: Api -> [Aeson.Value]
+getApiEntitiesObject api = maybe [] (map (makeJsonWithEntityData . AS.refName)) (Api.entities api)
+
+relPathFromApisDirToServerSrcDir :: Path Posix (Rel importLocation) (Dir C.ServerSrcDir)
```
Martinsos, line 67: This is rel path from `src/routes/apis` to `src`, is it?
It is not a path from `src/apis` to `src`?

Could we then maybe name it `relPathFromApisRoutesToServerSrcDir`?

It is also a bit tricky that this is defined here, but then you have `src/routes/apis` up there -> these two are connected but it is not clear I think.
One option is to do something like `routesApisDirInServerSrcDir = [reldir|routes/apis|] :: Path Posix (Rel C.ServerSrcDir) (Dir ())`.
Then to get the inverse of it (`../../`), I believe there is a function in StrongPath for that, or if not there is one in our codebase.

waspc/src/Wasp/AppSpec/Api.hs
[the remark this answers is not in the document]
infomiho: Express supports a bunch of HTTP request methods, but I think we should be fine with the most used ones (you already included those). 

https://expressjs.com/en/5x/api.html#routing-methods

waspc/src/Wasp/Generator/ServerGenerator/ApiRoutesG.hs
```diff
@@ -0,0 +1,68 @@
...
+  return $ C.mkTmplFdWithDstAndData tmplFile dstFile (Just tmplData)
+  where
+    apis = map snd $ AS.getApis spec
+    tmplData = object ["apiRoutes" .= map getTmplData apis]
+    tmplFile = C.asTmplFile [relfile|src/routes/apis/index.js|]
+    dstFile = SP.castRel tmplFile :: Path' (Rel ServerRootDir) File'
+
+    getTmplData :: Api -> Aeson.Value
+    getTmplData api =
+      let (jsImportStmt, jsImportIdentifier) = getJsImportStmtAndIdentifier relPathFromApisDirToServerSrcDir (Api.fn api)
```
infomiho, line 39: Too bad this is not yet merged, but we could refactor this in the future to use `extImportToImportJson` https://github.com/wasp-lang/wasp/pull/998#diff-c45766197d8c820bbae67e251a4a002a1845a2da09f9d17a89c6819919241940R18

Martinsos (approved):
@shayneczyzewski looks great to me, cool, custom api routes :)!!!

I left some comments but it is all minor.

Here is my response to your general notes:

No notion for `auth: true` and `User` in context -> I guess that is fine! They can easily send user themselves if they want, can they? Although, authentication might be harder to do if they are using our custom auth.
Hm, maybe it would be interesting to add -> how hard do you think that would be to add? But one option is certianly to go without it and we can add it later. Your assumption makes sense to me but I am not sure, it is hard to predict.

Default middleware stack -> makes sense!

`pass` being passed through -> makes sense! We can document it for now and we will improve it in the future.

All together I think it is a nice addition, and that you did the right amount of work -> obviusly we have some ideas for the future, but best to see a bit how this works and then we will be smarter on what people need / what we can improve. I think this is great as a first go at this feature.

waspc/src/Wasp/Generator/ServerGenerator/ApiRoutesG.hs
```diff
@@ -0,0 +1,68 @@
...
+    dstFile = SP.castRel tmplFile :: Path' (Rel ServerRootDir) File'
+
+    getTmplData :: (String, Api) -> Aeson.Value
+    getTmplData (name, api) =
+      object
+        [ "name" .= name,
+          "entities" .= getApiEntitiesObject api
+        ]
+
+getApiEntitiesObject :: Api -> [Aeson.Value]
```
infomiho, line 64: Is there a plan to use the explicit entities list used by the api for something?

waspc/data/Generator/templates/server/src/routes/apis/index.js
```diff
@@ -0,0 +1,25 @@
+{{={= =}=}}
```
infomiho, line 1: I would suggest we rewrite this file to Typescript to get the benefits of Typescript :)

waspc/src/Wasp/Generator/ServerGenerator/ApiRoutesG.hs
Martinsos: or by letting them import them via normal imports. Hm I thought we have a reason for this but I also can't remember!

waspc/examples/todoApp/src/server/apis.ts
shayneczyzewski: Wrapper function was incorrect. I meant I am not creating a function-level type that provides the types the context, amongst other things, like Queries: https://github.com/wasp-lang/wasp/blob/main/waspc/data/Generator/templates/server/src/queries/types.ts Here, it's just for the params.

I can uppercase the first letter too 👍🏻

waspc/examples/todoApp/todoApp.wasp
shayneczyzewski: Ah, that does look nice! I will consider swapping it to that for sure, thnx 👍🏻

shayneczyzewski: I think the only downside is it makes it harder to add more things in the future or make things optional, but I am not sure we need either of those at this point so in terms of ergonomics that may win!

waspc/src/Wasp/Generator/ServerGenerator/ApiRoutesG.hs
shayneczyzewski: Yeah, I'm not sure to be honest if we get any big gains from making it explicit (beyond looking like operations, and leaving us open to using that knowledge in the future somehow if needed). I don't know that we have a good way to let them import the Prisma query proxy object (which is different than the Prisma entity type) right now. I'll tinker with it and see if just giving them all entities in the context feels right 👍🏻 thnx for the idea

waspc/src/Wasp/Generator/ServerGenerator/ApiRoutesG.hs
shayneczyzewski: Ah yes, you are correct, that was under routes. I can update the name and bring it into the function it was used in, but I will also see how SP can help too. Thanks! 👍🏻

shayneczyzewski:
> I would maybe consider simplifying the api:
> 
> 1. If no method is provided, `GET` is assumed
> 2. `method` and `path` are top level properties
> 3. Inject all of the entities by default (I'm not sure if there is something we want to get by explicitly defining the list of entities)
> 
> This:
> 
> ```c
> api fooBar {
>   fn: import { fooBar } from "@server/apis.js",
>   entities: [Task],
>   httpRoute: {
>     method: GET,
>     path: "/foo/bar"
>   }
> }
> ```
> 
> would become this:
> 
> ```c
> api fooBar {
>   path: "/foo/bar",
>   fn: import { fooBar } from "@server/apis.js",
> }
> ```

Thanks for the review and good ideas @infomiho! I personally am not a huge fan of default HTTP methods in this context. If it was making `GET` the default for `query` then I think it would make sense, but here I do not think we should have a bias, and being explicit makes it nicer to read to me personally. Feels a bit more like RoR, which I of course like haha.

Good point about removing nesting and making it easier to write though, so what do you think about Martin's idea of: `httpRoute: (GET, "/foo/bar")` I think that reads pretty well here.

It's probably fine to just pass in all entities like you said, I can look into that. Thanks again!

shayneczyzewski:
> @shayneczyzewski looks great to me, cool, custom api routes :)!!!
> 
> I left some comments but it is all minor.
> 
> Here is my response to your general notes:
> 
> No notion for `auth: true` and `User` in context -> I guess that is fine! They can easily send user themselves if they want, can they? Although, authentication might be harder to do if they are using our custom auth. Hm, maybe it would be interesting to add -> how hard do you think that would be to add? But one option is certianly to go without it and we can add it later. Your assumption makes sense to me but I am not sure, it is hard to predict.
> 
> Default middleware stack -> makes sense!
> 
> `pass` being passed through -> makes sense! We can document it for now and we will improve it in the future.
> 
> All together I think it is a nice addition, and that you did the right amount of work -> obviusly we have some ideas for the future, but best to see a bit how this works and then we will be smarter on what people need / what we can improve. I think this is great as a first go at this feature.

Thanks for the comments and review! Yeah, I think going this far is a good first-stab compromise and then we can iterate on feedback for some future things. Plus, I like your tuple idea, I think I will swap to that! :) 👍🏻

In terms of auth, I could try to implement support for it if we think people will want it soon. The trick though is I don't think we give them an easy way to make arbitrary HTTP requests from the client and pass along the necessary JWT header (thanks to operations solving this for them already). So in practical terms, I am not sure how they could easily make use of that right now unless we add some capabilities to the client API as part of this (but we can just expose the `api` object or some wrapped version pretty quickly)). If we think it would add value to the initial release though or eventually be needed, I can try to see what that looks like, no prob! I don't think it will take too long.

Martinsos:
> I would maybe consider simplifying the api:
> 
> 1. If no method is provided, `GET` is assumed
> 2. `method` and `path` are top level properties
> 3. Inject all of the entities by default (I'm not sure if there is something we want to get by explicitly defining the list of entities)
> 
> This:
> 
> ```c
> api fooBar {
>   fn: import { fooBar } from "@server/apis.js",
>   entities: [Task],
>   httpRoute: {
>     method: GET,
>     path: "/foo/bar"
>   }
> }
> ```
> 
> would become this:
> 
> ```c
> api fooBar {
>   path: "/foo/bar",
>   fn: import { fooBar } from "@server/apis.js",
> }
> ```

I kind of like that method is not default, I think it makes it more readable.
But we can do `(GET, "some/path")` that I described in another comment.

Injecting all entities might be a bit aggressive, since most of the times I think they will not need any! I would rather then go the route of them importing them via JS. I am not sure if that breaks any of our "abstractions" though? If so, then I would stick to this here.

waspc/examples/todoApp/todoApp.wasp
Martinsos: True, but it is hard to guess now. We might also just add those next to this then.
Whatever you think is best! Once we add union types, we will be able to say it can be `(GET, "foo/bar")` OR it can be a dict. So even if you got with `(,)` for now, we still have space to add more config options in the future, once we add union types. With that in mind I think it is ok to with a pair at the moment?

Martinsos:
> 

auth -> all makes sense, I think you can judge the best if it makes sense to do it now or to add it in later PR! Or we can say we can certainly do it in later PR, and then questions is just when to tackle that one.

waspc/examples/todoApp/src/server/apis.ts
shayneczyzewski: Made type uppercase.

waspc/data/Generator/templates/server/src/routes/index.js
shayneczyzewski: Added comment 👍🏻

waspc/examples/todoApp/todoApp.wasp
shayneczyzewski: Swapped syntax to the tuple format 👍🏻

waspc/src/Wasp/Generator/ServerGenerator/ApiRoutesG.hs
shayneczyzewski: Updated name and pulled into function.

waspc/src/Wasp/Generator/ServerGenerator/ApiRoutesG.hs
shayneczyzewski: I plan to leave it explicit in the wasp file for now, since it looks/feels like Operations and we may have a future use.

waspc/data/Generator/templates/server/src/routes/apis/index.js
shayneczyzewski: Rewritten!

shayneczyzewski:
@Martinsos I added support for auth, it wasn't too bad. I will do more testing on it tomorrow though to make sure I didn't break anything in the process. :)

waspc/data/Generator/templates/server/src/apis/types.ts
```diff
@@ -0,0 +1,20 @@
+{{={= =}=}}
+
+import prisma from '../dbClient.js'
+
+{=# isAuthEnabled =}
+import { {= userEntityUpper =} } from '../entities'
```
sodic, line 6: Pro tip: use `import type`, see https://github.com/wasp-lang/wasp/pull/945#discussion_r1081399180.

waspc/examples/todoApp/src/server/apis.ts
sodic: Yes, it's better to expose the type of the entire function in this case. Just to be clear, this is what I understood you meant:
```typescript
import { FooBar } from '@wasp/apis/types'

export const fooBar: FooBar = (req, res, context) => {
  // ...
}
```
We always need a function of type `(req: Request, res: Response, context: fooBarContext) => void`:
  - If they specify a function of type `(req: string, res: Response, context: fooBarContext) => void`, they'll run into errors.
  - If they specify a function of type `(req: Request, res: SomethingElse, context: fooBarContext) => void`, they'll run into errors.
  - etc.
 
On the other hand, if we just tell them to use the type `FooBar` that's equal to `(req: string, res: Response, context: fooBarContext) => unknown`, they can't make such errors.
Also, it's only a single import that assigns correct types to everything they need - fewer imports, fewer things to remember, less code, smaller chance of making a mistake.
Finally, it's more similar to the API we already have.

The only problem is, `Response` and `Request` are generic types (and they can take quite a few type arguments):
```typescript
interface Request<
    P = core.ParamsDictionary,
    ResBody = any,
    ReqBody = any,
    ReqQuery = core.Query,
    Locals extends Record<string, any> = Record<string, any>
> extends core.Request<P, ResBody, ReqBody, ReqQuery, Locals> {}

interface Response<
    ResBody = any,
    Locals extends Record<string, any> = Record<string, any>
> extends core.Response<ResBody, Locals> {}
```
By hiding them behind `FooBar`, we take away the option to specify the argument types from our users.

The good new is, if you take a closer look, you'll see several of these type arguments are the same (e..g, `ResBody`). In other words, these two types (`Request` and `Response`) are coupled.
To prevent mistakes such a coupling could cause (i.e., to make illegal states unrepresentable), Express also exposes a function type for the handler:
```typescript
interface RequestHandler<
    P = core.ParamsDictionary,
    ResBody = any,
    ReqBody = any,
    ReqQuery = core.Query,
    Locals extends Record<string, any> = Record<string, any>
> extends core.RequestHandler<P, ResBody, ReqBody, ReqQuery, Locals> {}
```
Which is pretty  much exactly what we should be using - our `FooBar` type should decorate `RequestHandler`.

sodic (commented):
Just answered the question for now, will come back with more stuff :)

Awesome work, as always!

Martinsos:
> @Martinsos I added support for auth, it wasn't too bad. I will do more testing on it tomorrow though to make sure I didn't break anything in the process. :)

Cool, nice :)!

waspc/data/Generator/templates/server/src/apis/types.ts
```diff
@@ -0,0 +1,20 @@
...
+export type {= name =}Context = {
+  {=# isUsingAuth =}
+  user: {= userEntityUpper =},
+  {=/ isUsingAuth =}
+  entities: {
+    {=# entities =}
+    {= name =}: typeof prisma.{= prismaIdentifier =},
+    {=/ entities =}
+  }
+}
```
sodic, line 19: We have a type for this in our internal types: https://github.com/wasp-lang/wasp/blob/cb691c8dd7348553bee123b2cdcf05d75847a0b5/waspc/data/Generator/templates/server/src/_types/index.ts#L35.

waspc/data/Generator/templates/server/src/routes/apis/index.ts
```diff
@@ -0,0 +1,37 @@
...
+{=# apiRoutes =}
+{=# isUsingAuth =}
+router.{= routeMethod =}('{= routePath =}', auth, handleRejection((req: Request & { user: {= userEntityUpper =} }, res: Response) => {
+{=/ isUsingAuth =}
+{=^ isUsingAuth =}
+router.{= routeMethod =}('{= routePath =}', handleRejection((req: Request, res: Response) => {
+{=/ isUsingAuth =}
+  const context = {
+    {=# isUsingAuth =}
+    user: req.user,
```
sodic, line 25: In operations, we strip the `password` filed. Do we want to do the same here?

sodic (commented):
Ok, went through everything.

A couple of minor comments you might want to check out (the `password` one mostly). Other than that, all good!

waspc/data/Generator/templates/server/src/routes/apis/index.ts
shayneczyzewski: Good question. It gets stripped out in the middleware that ops and apis use: https://github.com/wasp-lang/wasp/blob/main/waspc/data/Generator/templates/server/src/core/auth.js#L47-L53

Just `console.log`ged it to confirm and no password in api context 👍🏻

sodic: In that case, we should correct the type. We're currently telling users they're working with `User`. You can omit the password field like this: https://github.com/wasp-lang/wasp/blob/cb691c8dd7348553bee123b2cdcf05d75847a0b5/waspc/data/Generator/templates/server/src/_types/index.ts#L28 (mind the comment :)).

waspc/data/Generator/templates/server/src/apis/types.ts
sodic: Sure, let me know if I can help (but you should be able to just import and use it).

waspc/examples/todoApp/todoApp.wasp
```diff
@@ -105,6 +105,12 @@ query getTasks {
   entities: [Task]
 }
 
+api fooBar {
+  fn: import { fooBar } from "@server/apis.js",
+  entities: [Task],
+  httpRoute: (GET, "/foo/bar")
```
sodic, line 111: Should we maybe prefix all api routes with `api`? Since operations are under `operations`?

infomiho: I vote against it since users might need ultra specific callback path and then they are doomed 😢

waspc/examples/todoApp/src/server/apis.ts
shayneczyzewski: Updated to use a function type, thanks for the help @sodic!

waspc/examples/todoApp/todoApp.wasp
sodic: Ok, makes sense

waspc/ChangeLog.md
```diff
@@ -1,5 +1,26 @@
 # Changelog
 
+## v0.9.1
+
+### Adds an `api` keyword for defining an arbitrary endpoint and URL
+Need a specific endpoint, like `/healthcheck` or `/foo/callback`? Or need complete control of the response? Use an `api` to define one by tying a JS function to any HTTP method and path! For example:
```
Martinsos, line 6: Oh yeah I like the spirit :D! This got me excited to use this feature :D!

waspc/data/Generator/templates/server/src/_types/index.ts
```diff
@@ -49,6 +67,20 @@ type PrismaDelegate = {
...
 }
 
+type Context<Entities extends _Entity[]> = Expand<{
+  entities: Expand<EntityMap<Entities>>
+}>
+
+type ContextWithUser<Entities extends _Entity[]> = Expand<Context<Entities> & UserInContext>
+
+export type UserInContext = {
+  // TODO: This type must match the logic in core/auth.js (if we remove the
```
Martinsos, line 77: Just to check, do we also have corresponding comment there?

waspc/src/Wasp/AppSpec/Api.hs
```diff
@@ -0,0 +1,34 @@
...
+import Data.Data (Data)
+import Wasp.AppSpec.Core.Decl (IsDecl)
+import Wasp.AppSpec.Core.Ref (Ref)
+import Wasp.AppSpec.Entity (Entity)
+import Wasp.AppSpec.ExtImport (ExtImport)
+
+data Api = Api
+  { fn :: ExtImport,
+    entities :: Maybe [Ref Entity],
+    httpRoute :: (HttpMethod, String),
```
Martinsos, line 20: Maybe just drop a short comment here saying that snd element in the tuple is route path, describe it a bit?

Btw, it is possible to have a Haskell record here, like you had it before, while on other hand having it as a tuple in Wasp DSL -> we do it with dependencies in AppSpec. I am not saying you have to do it, but it is a way to use syntax we want in DSL while giving it some extra structure in AppSpec when we want.

waspc/src/Wasp/Generator/ServerGenerator.hs
```diff
@@ -240,7 +243,8 @@ genRoutesDir spec =
         ( Just $
             object
               [ "operationsRouteInRootRouter" .= (operationsRouteInRootRouter :: String),
-                "isAuthEnabled" .= (isAuthEnabled spec :: Bool)
+                "isAuthEnabled" .= (isAuthEnabled spec :: Bool),
+                "areApisInUse" .= (not . null $ getApis spec)
```
Martinsos, line 247: Nitpick: Maybe this is a bit richer?
"inUse" sounds like it is not so much important if they exist, but if they are used. But really we are checking if they exist, right?

```suggestion
                "areThereAnyCustomApiRoutes" .= (not . null $ getApis spec)
```
but if you don't like this ignore it.

web/docs/language/features.md
```diff
@@ -584,6 +584,107 @@ import { isPrismaError, prismaErrorToHttpError } from '@wasp/utils.js'
   }
 ```
 
+## APIs
```
Martinsos, line 587: Wow these are some nice rich comments!

web/docs/language/syntax.md
```diff
@@ -66,6 +66,7 @@ While fundamental types are here to be basic building blocks of a language, and
 - Domain types ([source of truth](https://github.com/wasp-lang/wasp/blob/main/waspc/src/Wasp/Analyzer/StdTypeDefinitions.hs))
   - Declaration types
     - **action**
+    - **api**
```
Martinsos, line 69: oh good catch!

Martinsos (approved):
Left a couple of small comments but looks great all together, approved!

waspc/data/Generator/templates/server/src/_types/index.ts
shayneczyzewski: This was Filip's original comment I just moved. But yes, just checked that file and there is one there too 👍🏻

waspc/src/Wasp/AppSpec/Api.hs
shayneczyzewski: I'll add a short comment just to wrap this PR up. I also have helpers below already that make it feel a bit like record syntax from the outside.

waspc/data/Generator/templates/server/src/_types/index.ts
```diff
@@ -20,24 +22,40 @@ export type AuthenticatedAction<Entities extends _Entity[], Input, Output> =
...
+  Entities extends _Entity[],
+  Params extends ExpressParams,
+  ResBody,
+  ReqBody,
+  ReqQuery extends ExpressQuery,
+  Locals extends Record<string, any>
+> = (
+  req: Request<Params, ResBody, ReqBody, ReqQuery, Locals>,
+  res: Response<ResBody, Locals>,
+  context: Context<Entities>,
```
sodic, line 57: I see you decided to share the context between operations and the API after all.

Fine by me, just making sure you didn't leave it by accident (as the reasoning was that we maybe wouldn't want to accidentally change one while changing the other)..

waspc/data/Generator/templates/server/src/apis/types.ts
sodic: I see we're still importing several types without the `type` import (`Request` and `Response` from `express` for example).

This is not _that_ important, so feel free to skip going through them and fixing it.

I'm approving either way.

web/docs/language/features.md
```diff
@@ -584,6 +584,107 @@ import { isPrismaError, prismaErrorToHttpError } from '@wasp/utils.js'
...
+
+In Wasp, the default client-server interaction mechanism is through Operations. However, if you need a specific URL method/path, or a specific response, Operations may not be suitable for you. For these cases, you can use an `api`! Best of all, they should look and feel very familiar.
```
sodic, line 589: I'd suggest linking an anchor to the Operations section of the docs here.

web/docs/language/features.md
```diff
@@ -584,6 +584,107 @@ import { isPrismaError, prismaErrorToHttpError } from '@wasp/utils.js'
...
+import { FooBar } from '@wasp/apis/types'
+
+export const fooBar : FooBar = (req, res, context) => {
+  res.set('Access-Control-Allow-Origin', '*') // Example of modifying headers to override Wasp default CORS middleware.
+  res.json({ msg: `Hello, ${context.user?.username || "stranger"}!` })
+}
+```
+
+#### Declaring an API in Wasp
+After implementing your APIs in NodeJS, all that's left to do before using them is tell Wasp about it!
```
sodic, line 618: We should flip the narrative here (and I think we use a similar sentence when we talk about Queries).

The thing is, we want them to first define their Query/Action/API inside the `.wasp` file, and only then go and define its implementation. If they don't do it in this order, they lose all benefits of type support.

sodic (commented):
Great job on the docs!

The only real change I think we should make before merging is outlined here:

waspc/e2e-test/Tests/WaspComplexTest.hs
```diff
@@ -228,3 +229,28 @@ insertCodeIntoWaspFileAfterVersion = insertCodeIntoWaspFileAtLineNumber lineNumb
...
+addApi = do
+  sequence
+    [ appendToWaspFile apiDecl,
+      createFile apiFile "./src/server" "apis.ts"
+    ]
+  where
+    apiDecl =
+      unlines
+        [ "api fooBar {",
+          "  fn: import { fooBar } from \"@server/apis.js\",",
```
infomiho, line 243: I would maybe suggest adding another authenticated API route (it can have the same JS impl)

web/docs/language/features.md
```diff
@@ -584,6 +584,107 @@ import { isPrismaError, prismaErrorToHttpError } from '@wasp/utils.js'
...
+You can easily do this with the `api` declaration, which supports the following fields:
+- `fn: ServerImport` (required) - The import statement of the APIs NodeJs implementation.
+- `httpRoute: (HttpMethod, string)` (required) - The HTTP (method, path) pair, where method can be one of:
```
infomiho, line 621:
```suggestion
- `httpRoute: (HttpMethod, string)` (required) - The HTTP (method, path) pair, where the method can be one of:
```
Grammarly

web/docs/language/features.md
```diff
@@ -584,6 +584,107 @@ import { isPrismaError, prismaErrorToHttpError } from '@wasp/utils.js'
...
+  - `ALL`, `GET`, `POST`, `PUT` or `DELETE`
+  - and path is an Express path `string`.
+- `entities: [Entity]` (optional) - A list of entities you wish to use inside your API.
+We'll leave this option aside for now. You can read more about it [here](#using-entities-in-apis).
+- `auth: bool` (optional) - If auth is enabled, this will default to `true` and provide a `context.user` object. If you do not wish to attempt to parse the JWT in the Authorization Header, you may set this to `false`.
+
+Wasp APIs and their implementations don't need to (but can) have the same name. With that in mind, this is how you might declare the API that use the implementations from the previous step:
```
infomiho, line 628:
```suggestion
Wasp APIs and their implementations don't need to (but can) have the same name. With that in mind, this is how you might declare the API that uses the implementations from the previous step:
```
Grammarly

web/docs/language/features.md
```diff
@@ -584,6 +584,107 @@ import { isPrismaError, prismaErrorToHttpError } from '@wasp/utils.js'
...
+}
+```
+
+#### Using the API
+To use the API externally, you simply call the endpoint using the method and path you used. For example, if your app is running at `https://example.com` then from the above you could issue a `GET` to `https://example/com/foo/callback` (in your browser, Postman, `curl`, another web service, etc.).
+
+To use the API from your client, including with auth support, you can import the Axios wrapper from `@wasp/api` and invoke a call. For example:
+```ts
+import React, { useEffect } from 'react'
+import api from '@wasp/api'
```
infomiho, line 644: Love this piece of DX!

web/docs/language/features.md
```diff
@@ -584,6 +584,107 @@ import { isPrismaError, prismaErrorToHttpError } from '@wasp/utils.js'
...
+
+async function useCustomRoute() {
```
infomiho, line 646: The naming of the function is weird since the prefix `use` usually signifies some React hooks in actio. To avoid confusing the reader I would maybe name it `fetchCustomRoute` or `fetchFooBar`

infomiho (approved):
Great work :) It's a really cool addition to the Wasp toolbelt. 

Tested out locally, works for me. I left some minor comments.

web/docs/language/features.md
```diff
@@ -584,6 +584,107 @@ import { isPrismaError, prismaErrorToHttpError } from '@wasp/utils.js'
...
+#### Declaring an API in Wasp
+After implementing your APIs in NodeJS, all that's left to do before using them is tell Wasp about it!
+You can easily do this with the `api` declaration, which supports the following fields:
+- `fn: ServerImport` (required) - The import statement of the APIs NodeJs implementation.
+- `httpRoute: (HttpMethod, string)` (required) - The HTTP (method, path) pair, where method can be one of:
+  - `ALL`, `GET`, `POST`, `PUT` or `DELETE`
+  - and path is an Express path `string`.
+- `entities: [Entity]` (optional) - A list of entities you wish to use inside your API.
+We'll leave this option aside for now. You can read more about it [here](#using-entities-in-apis).
+- `auth: bool` (optional) - If auth is enabled, this will default to `true` and provide a `context.user` object. If you do not wish to attempt to parse the JWT in the Authorization Header, you may set this to `false`.
```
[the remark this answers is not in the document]
infomiho: Now I got it, you get the `user` object if the auth is enabled. 

This was a bit confusing for me a bit because I expected that it will return `401` if the user is not authenticaed. 

For operations, we include the user object by default, right? This means we have two strategies: for operations we include it by default, for APIs it's opt in. Which is fine, since APIs are more hands on and used for custom work.

waspc/data/Generator/templates/server/src/apis/types.ts
shayneczyzewski: Ah crap, I forgot to check others, I will update them before merging 👍🏻

waspc/data/Generator/templates/server/src/_types/index.ts
shayneczyzewski: Yeah I decided to share for now and we can separate if/when they diverge. Just wanted to keep it simple for now 👍🏻

web/docs/language/features.md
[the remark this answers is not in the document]
shayneczyzewski: Yeah it behaves the same as ops for now 👍🏻

web/docs/language/features.md
shayneczyzewski: So I was thinking of adding this note as a short stopgap until a larger doc refresh:
Screenshot 2023-03-13 at 1 24 06 PM

sodic: > So we have a bit of a chicken and egg probem when explaining this though, since if we say define it in your .wasp file first, we have to reference some file that does not yet exist. Is that a bit confusing?

It's definitely confusing, that's why I wasn't a fan of having our type support work using this hack, but it was the only way at the moment. And unfortunately I don't know how to word it either.

> Do you mean until they are both defined and the next compile, or indefinitely? I'm not sure how it could be the latter. If it is the former, is it so bad that it is temporarily unavailable until they do both and compile? Maybe a note in between sections is enough for now as a stopgap until a larger rewrite to make sure the TS generation stuff is in all the doc sequences correctly?

Until the `.wasp` file is compiled, yes. Not indefinitely.

 It doesn't seem like a big deal, but I'd argue that it's pretty important.
The main benefits of TypeScript are type checking and autocomplete. If they start implementing the API handler without those benefits, they aren't getting anything out of our TS support (and if they want to use TS, that's probably a deal breaker).

The only way it could work is if they:
1. Define the skeleton of the implementation in the TS file (and export it).
2. Define the declaration inside the `.wasp` file 
3. Go back to the TS file and only then start writing the implementation.

In other words, they'd knowingly have to postpone implementing the entire function when they first create it, and if we're asking them to do that, it's simpler to ask them to just add the query inside the `.wasp` file first.

With all that said, I still wouldn't know how to communicate this, so the note can stay.

PR #1044, opened 2023-03-03
Managed db

Martinsos:
Bearoucracy: This is new feature and I added Changelog entry + updated cabal version to 0.8.3.

Ok, so this works, we now have a way to easily run dev database for a Wasp app!

It brings two new commands: `wasp db start` and `wasp db reset`.

The idea is that you run db with `wasp db start` and that is it, nothing else to do.

Must TODO:
 - [x] Add text at start of `wasp db start` that explains what is happening.
 - [x] Fix bug when .env.server does not exist.
 - [x] Fix permissions on created dev db -> right now neither wasp nor user can delete it.
 - [x] Fix `wasp db reset` breaking:
       This will be interesting, it wants to re-generate clients, but it doesn't know how so it breaks. I think what I will want to do here is use the --skip-generate flag that they mention and then either not do anything, or run our own generation of clients -> I have to think about what is the correct approach.
 - [x] `wasp db start` -> `wasp start db`.
 - [x] Update docs.
 - [x] Test situation when .env.server does not exist.
 - [x] Test situation when `docker` is not in PATH?
 - [x] Test `wasp db reset`.
 - [x] Update CHANGELOG.
 - [x] Test on Mac and WSL2 on Win.

Probably should TODO:
 - [x] When running `wasp db start` -> have explicit check to see if `docker` is there, if not show nice error message.
 - [x] If they defined DATABASE_URL in .env.server but are running `wasp db start` -> warn them about that. But how do we know the content of .env.server? Well, we could read env files when constructing AppSpec and feed their parsed content into the AppSpec, instead of providing just URLs. Then we have them available.
 - [x] Make it so that `wasp clean` deletes only .wasp/out but not databases? And then smth like `wasp clean --also-db` deletes everything? If I don't do it now, I should at least create an issue for it.
 - [x] There is a potential issue: what if I run `wasp db start` in one project, and then `wasp start` another project -> that second project will now connect to the database of the first project, because the default connection URL is the same!
   How can I avoid this? One way could be to use the db name that is the same as project name. This could cause problems though if project changes its name, or if two projects have the same name.
   Another way might be to write down the port that used for establishing connection, and then Wasp obtains connection URL based on that file. This is what we will want to do anyway in the future, for dynamic ports support. Or something similar but we first make up unique id/hash, write that to a hidden non-version file, and then use that for db name or smth.
  Actually, since we are using projet name + path right now for volumes, we could use same thing for db name and that way avoid this issue.

For the future TODO:
 - #1049 
 - #1051 
 - #1052 
 - #1053


waspc/cli/src/Wasp/Cli/Command/Db/Start.hs
```diff
@@ -0,0 +1,63 @@
...
+  case getDbSystem decls of
+    Nothing -> noteSQLiteDoesntNeedStart
+    Just AS.App.Db.SQLite -> noteSQLiteDoesntNeedStart
+    Just AS.App.Db.PostgreSQL -> startPostgreDevDb waspProjectDir
+  where
+    getDbSystem decls = (snd <$> AS.getApp decls) >>= AS.App.db >>= AS.App.Db.system
+
+    noteSQLiteDoesntNeedStart =
+      cliSendMessageC . Msg.Info $
+        "Nothing to do! You are all good, you are using SQLite which doesn't need to be started."
```
infomiho, line 38: Maybe also add, if they wanted to use `PostgreSQL` they should change the `system` field in the `db`` declaration of the Wasp app.

waspc/cli/src/Wasp/Cli/Command/Db/Start.hs
```diff
@@ -0,0 +1,63 @@
...
+import Wasp.Cli.Command.Common
+  ( findWaspProjectRootDirFromCwd,
+    parseWaspFileC,
+  )
+import Wasp.Cli.Command.Message (cliSendMessageC)
+import Wasp.Cli.Common (WaspProjectDir, dotWaspDirInWaspProjectDir)
+import qualified Wasp.Db.Postgres as P
+import qualified Wasp.Message as Msg
+
+start :: Command ()
```
infomiho, line 24: As we talked IRL, I would maybe add a bit of explanation when the PostgreSQL db is started,

```
You are using a PostgreSQL database system in the Wasp file, so we started a local dev PostgreSQL instance for you to use.

You don't need to set up anything in the .env.server file, but if you are looking for the connection string for some database tool, here it is:

postgresql://...
```

infomiho: If have a `DATABASE_URL` set up, my app will use it. It would be great if we could check for that and output a message to the user.

waspc/src/Wasp/Db/Postgres.hs
```diff
@@ -0,0 +1,33 @@
...
+  )
+where
+
+import Text.Printf (printf)
+
+-- | TODO: Write explanation on what "dev" stands for in all these variables.
+-- Explain that the idea is that there are default settings we expect from a development database
+-- and that they are defined here.
+-- Maybe move "dev" stuff into additional module, Wasp.Db.Postgres.Dev ?
+defaultDevUser :: String
```
infomiho, line 17: Since the username, password and database name don't matter, I vote we keep it as plain as possible e.g. `user=postgres`, `dbName=postgres` and `password=devpass` :)

waspc/test/Generator/FileDraft/CopyAndModifyTextFileDraftTest.hs
```diff
@@ -0,0 +1,29 @@
...
+        `shouldBe` [(SP.toFilePath expectedDstPath, "First line\nMock text file content")]
+  where
+    (dstDir, dstPath, srcPath) =
+      ( systemSPRoot SP.</> [SP.reldir|a/b|],
+        [SP.relfile|c/d/dst.txt|],
+        systemSPRoot SP.</> [SP.relfile|e/src.txt|]
+      )
+    fileDraft = createCopyAndModifyTextFileDraft dstPath srcPath ("First line\n" <>)
+    expectedSrcPath = srcPath
+    expectedDstPath = dstDir SP.</> dstPath
```
infomiho, line 29: This goes against of the convention of importing `(</>)` from StrongPath vs. using it `SP.</>`

waspc/test/Generator/FileDraft/CopyAndModifyTextFileDraftTest.hs
```diff
@@ -0,0 +1,29 @@
...
+      let mock = write dstDir fileDraft
+      let mockLogs = Mock.getMockLogs mock Mock.defaultMockConfig
+      Mock.createDirectoryIfMissing_calls mockLogs
+        `shouldBe` [(True, SP.toFilePath $ SP.parent expectedDstPath)]
+      Mock.readFileAsText_calls mockLogs
+        `shouldBe` [SP.toFilePath expectedSrcPath]
+      Mock.writeFileFromText_calls mockLogs
+        `shouldBe` [(SP.toFilePath expectedDstPath, "First line\nMock text file content")]
+  where
+    (dstDir, dstPath, srcPath) =
```
infomiho, line 22: Is this a way in Haskell to group variables? (deconstructing a tuple on the left while defining it on the lef)

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
```diff
@@ -114,8 +105,18 @@ generatePrismaClient projectDir prismaClientOutputDirEnv jobType =
     envVars = [prismaClientOutputDirEnv]
     serverRootDir = projectDir </> serverRootDirInProjectRootDir
     prismaExecutable = absPrismaExecutableFp projectDir
-    prismaGenerateCmdArgs = ["generate", "--schema", schemaFile]
-    schemaFile = SP.fromAbsFile $ projectDir </> dbSchemaFileInProjectRootDir
+    prismaGenerateCmdArgs = ["generate", "--schema", SP.fromAbsFile schemaFile]
+    schemaFile = projectDir </> dbSchemaFileInProjectRootDir
+
+runPrismaCommandAsJob ::
```
infomiho, line 111: Nice simplification

waspc/cli/src/Wasp/Cli/Command/Db/Start.hs
```diff
@@ -0,0 +1,63 @@
...
+          [ "docker run",
+            "--rm",
+            "--publish " <> show P.defaultDevPort <> ":5432",
+            "-v " <> fromAbsDir devDbDir <> ":/var/lib/postgresql/data",
+            "--env POSTGRES_PASSWORD=" <> P.defaultDevPass,
+            "--env POSTGRES_USER=" <> P.defaultDevUser,
+            "--env POSTGRES_DB=" <> P.defaultDevDbName,
+            "postgres"
+          ]
+  liftIO $ callCommand command
```
infomiho, line 56: Maybe we can run `migrate-dev` after we start the database? 🤷 I'm not sure how much this complicates things.

infomiho:
Looking good, work on my machine well. 

I would maybe not write our dev db connection string if `DATABASE_URL` is already defined in the `.env.server`.
```env
# We wrote this
DATABASE_URL=our_dev_db
# User wrote this
DATABASE_URL=custom_user_db
```
If we don't do it, we implicitly depend on the behaviour of `dotenv` JS lib to honor the order of definiton.

waspc/cli/exe/Main.hs
```diff
@@ -102,7 +105,7 @@ printUsage =
         title "  IN PROJECT",
         cmd "    start                 Runs Wasp app in development mode, watching for file changes.",
         cmd "    db <db-cmd> [args]    Executes a database command. Run 'wasp db' for more info.",
-        cmd "    clean                 Deletes all generated code and other cached artifacts. Wasp equivalent of 'have you tried closing and opening it again?'.",
+        cmd "    clean                 Deletes all generated code and other cached artifacts, including dev db. Wasp equivalent of 'have you tried closing and opening it again?'.",
```
shayneczyzewski, line 108: Do we want clean to empty the DB as well? I feel that could be a bit surprising. I would vote for clean just deleting .wasp artifacts, and leaving DB clean to `wasp db reset`.

waspc/cli/exe/Main.hs
```diff
@@ -150,6 +155,8 @@ printDbUsage =
         "  wasp db <command> [command-args]",
         "",
         title "COMMANDS",
+        cmd "  start       Starts development database.",
```
shayneczyzewski, line 158: Should we mention Docker here explicitly? May be nice to give them the "how" as well.

waspc/cli/src/Wasp/Cli/Command/Db/Start.hs
```diff
@@ -0,0 +1,63 @@
...
+start = do
+  waspProjectDir <- findWaspProjectRootDirFromCwd
+  decls <- parseWaspFileC waspProjectDir
+
+  case getDbSystem decls of
+    Nothing -> noteSQLiteDoesntNeedStart
+    Just AS.App.Db.SQLite -> noteSQLiteDoesntNeedStart
+    Just AS.App.Db.PostgreSQL -> startPostgreDevDb waspProjectDir
+  where
+    getDbSystem decls = (snd <$> AS.getApp decls) >>= AS.App.db >>= AS.App.Db.system
```
shayneczyzewski, line 34:
```suggestion
    getDbSystem decls = AS.getApp decls >>= AS.App.db . snd >>= AS.App.Db.system
```

Just giving you an option that Hlint gave.

waspc/cli/exe/Main.hs
shayneczyzewski: Aha, I see later on about where the Docker DB data lives (.wasp) and how it was designed to behave similarly to SQLite. Ok, I think this makes sense. I still feel it may catch some people by surprise, but we can see!

waspc/cli/src/Wasp/Cli/Command/Db/Reset.hs
```diff
@@ -0,0 +1,37 @@
...
+import System.Exit (ExitCode (..))
+import Wasp.Cli.Command (Command)
+import Wasp.Cli.Command.Common (findWaspProjectRootDirFromCwd)
+import Wasp.Cli.Command.Message (cliSendMessageC)
+import qualified Wasp.Cli.Common as Common
+import qualified Wasp.Generator.DbGenerator.Jobs as DB.Jobs
+import Wasp.Generator.Job.IO (readJobMessagesAndPrintThemPrefixed)
+import qualified Wasp.Message as Msg
+
+reset :: Command ()
```
shayneczyzewski, line 19: When I ran `db reset` with PG running outside of Docker (like I normally do), I go the following error on WSL2:

```sh
🐝 --- Reseting the database... ---------------------------------------------------


Db: Environment variables loaded from .env
Db: Prisma schema loaded from ../db/schema.prisma
Db: Datasource "db": PostgreSQL database "todoapp", schema "public" at "host:5432"
Db: 
✔ Are you sure you want to reset your database? All data will be lost. … yes
Db: 
Db: Applying migration `20220413175924_switch_to_postgresql`
Db: Applying migration `20220818151104_rename_email_to_username`
Db: Applying migration `20220822154342_add_social_login`
Db: 
Db: Database reset successful
Db: 
Db: The following migration(s) have been applied:
Db: 
Db: migrations/
Db:   └─ 20220413175924_switch_to_postgresql/
Db:     └─ migration.sql
Db:   └─ 20220818151104_rename_email_to_username/
Db:     └─ migration.sql
Db:   └─ 20220822154342_add_social_login/
Db:     └─ migration.sql
Db: 
Db: Running generate... (Use --skip-generate to skip the generators)
Db(stderr): Error: Attempted to load provider value using `env(PRISMA_CLIENT_OUTPUT_DIR)` but it was not present. Please ensure that PRISMA_CLIENT_OUTPUT_DIR is present in your Environment Variables
✅ --- Database reset failed with exit code 1 -------------------------------------
```

waspc/cli/src/Wasp/Cli/Command/Db/Start.hs
shayneczyzewski: Yeah, I agree. I got confused for a second when I ran `wasp db start` and then started the app, but had my .server.env file there and it was using my regular PG DB instead of the Wasp/Docker version.

waspc/src/Wasp/Generator/ServerGenerator.hs
```diff
@@ -86,11 +87,17 @@ genDotEnv spec = return $
...
               (C.serverRootDirInProjectRootDir </> dotEnvInServerRootDir)
               srcFilePath
+              addDevDbUrlToStartIfSet
           ]
     _ -> []
+  where
+    addDevDbUrlToStartIfSet :: T.Text -> T.Text
+    addDevDbUrlToStartIfSet t = case AS.devDatabaseUrl spec of
+      Nothing -> t
+      Just url -> "DATABASE_URL=" <> T.pack url <> "\n" <> t
```
shayneczyzewski, line 100: I'm wondering what would happen if:
- the file did not exist, would we still start the server ok?
- if the file did exist, and already had that envar set?

waspc/src/Wasp/Generator/FileDraft/CopyAndModifyTextFileDraft.hs
```diff
@@ -0,0 +1,54 @@
...
+data CopyAndModifyTextFileDraft = CopyAndModifyTextFileDraft
+  { -- | Path where the file will be copied to.
+    _dstPath :: !(Path' (Rel ProjectRootDir) File'),
+    -- | Absolute path of source file to copy.
+    _srcPath :: !(Path' Abs File'),
+    _failIfSrcDoesNotExist :: !Bool,
+    _modify :: !(T.Text -> T.Text)
+  }
+
+instance Writeable CopyAndModifyTextFileDraft where
```
shayneczyzewski, line 27: This is neat! :D Nice

shayneczyzewski (commented):
This is looking really cool so far, @Martinsos. Nice work! I think people will really like having this option! I left a few comments (including one error).

One general thing I noticed while testing on WSL2 is for some reason the `DATABASE_URL` seems missing. At first, `pg-boss` failed.

```sh
Server: Starting pg-boss...
Server(stderr): pg-boss failed to start!
Server(stderr): Error: SASL: SCRAM-SERVER-FIRST-MESSAGE: client password must be a string
Server(stderr):     at Object.continueSession (/home/shayne/dev/wasp/waspc/examples/todoApp/.wasp/out/server/node_modules/pg/lib/sasl.js:24:11)
Server(stderr):     at Client._handleAuthSASLContinue (/home/shayne/dev/wasp/waspc/examples/todoApp/.wasp/out/server/node_modules/pg/lib/client.js:261:12)
Server(stderr):     at Connection.emit (node:events:513:28)
Server(stderr):     at /home/shayne/dev/wasp/waspc/examples/todoApp/.wasp/out/server/node_modules/pg/lib/connection.js:119:12
Server(stderr):     at Parser.parse (/home/shayne/dev/wasp/waspc/examples/todoApp/.wasp/out/server/node_modules/pg-protocol/dist/parser.js:40:17)
Server(stderr):     at Socket.<anonymous> (/home/shayne/dev/wasp/waspc/examples/todoApp/.wasp/out/server/node_modules/pg-protocol/dist/index.js:11:42)
Server(stderr):     at Socket.emit (node:events:513:28)
Server(stderr):     at addChunk (node:internal/streams/readable:324:12)
Server(stderr):     at readableAddChunk (node:internal/streams/readable:297:9)
Server(stderr):     at Readable.push (node:internal/streams/readable:234:10)
Server: This is coming from shared function.
Server(stderr): node:internal/process/promises:288
Server(stderr):             triggerUncaughtException(err, true /* fromPromise */);
Server(stderr):             ^
Server(stderr): 
Server(stderr): [UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "#<PgBoss>".] {
Server(stderr):   code: 'ERR_UNHANDLED_REJECTION'
Server(stderr): }
Server(stderr): 
Server(stderr): Node.js v18.12.0
Server: [nodemon] app crashed - waiting for file changes before starting...
```

Then I disabled it and saw the real error:
```sh
Server(stderr): Error: 
Server(stderr): Invalid `prisma.task.count()` invocation:
Server(stderr): 
Server(stderr): 
Server(stderr): error: Environment variable not found: DATABASE_URL.
Server(stderr):   -->  schema.prisma:4
Server(stderr):    | 
Server(stderr):  3 |   provider = "postgresql"
Server(stderr):  4 |   url      = env("DATABASE_URL")
Server(stderr):    | 
Server(stderr): 
Server(stderr): Validation Error Count: 1
```

So in order to try to restart, I tried clean, but that failed with a perms error:
```sh
cabal run wasp-cli clean
Up to date

🐝 --- Deleting .wasp/ directory... -----------------------------------------------

wasp-cli: /home/shayne/dev/wasp/waspc/examples/todoApp/.wasp/.wasp-dev-db-data: removeDirectoryRecursive:removeContentsRecursive:removePathRecursive:removeContentsRecursive:getDirectoryContents:openDirStream: permission denied (Permission denied)
```

So I think there may be some WSL2 kinks to work out (or maybe my setup is not typical). Specifically, even with the Docker DB running (via `wasp db start`) my `DATABASE_URL` does not seem to be set. I did not have any `.env.server` file if that matters (I moved it so it wouldn't use my old PG DB).

For reference, I am just running Docker Desktop on Windows and using the WSL2 engine: https://docs.docker.com/desktop/windows/wsl/

waspc/cli/src/Wasp/Cli/Command/Db/Reset.hs
```diff
@@ -0,0 +1,37 @@
...
+
+  chan <- liftIO newChan
+  ((), exitCode) <-
+    liftIO $
+      readJobMessagesAndPrintThemPrefixed chan
+        `concurrently` DB.Jobs.reset genProjectDir chan
+
+  case exitCode of
+    ExitSuccess -> cliSendMessageC $ Msg.Success "Database reset successfully!"
+    ExitFailure c ->
```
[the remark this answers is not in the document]
Martinsos: Good one, fixed!

waspc/cli/src/Wasp/Cli/Command/Db/Start.hs
Martinsos: Interesting idea but I would not add it here as I think there are better ways for them to learn about PostgreSQL support, I would rather they go through those.

waspc/cli/src/Wasp/Cli/Command/Db/Start.hs
Martinsos: @infomiho this is a cool idea, I will add it!

@infomiho & @shayneczyzewski yes I would like to add that check for existing DATABASE_URL, I have actually have it there under the TODOs, although I put it under more advanced TODOs since it is mostly important for old users + it requires a bit of effort to implement (parsing .env.server) so I am considering postponing it for a later PR.

waspc/src/Wasp/Db/Postgres.hs
Martinsos: One reason why I added this "waspUser" and "waspDb" suffixes is that I often find it very confusing when I see something like 
```
postgresql://postgres:postgres@localhost:5432/postgres
```
, I would rather see smth like
```
postgresql://postgresWaspDevUser:postgresWaspDevPass@localhost:5432/postgresWaspDevDb
```

Also, I was thinking that it is better to have "wasp" in names because it makes sure that e.g. we won't by accident connect to a db that was run by smb else and not Wasp without knowing -> because as you said, all "postgress" is so default that maybe they already have such db running from some other project and we by accident connect to it.

waspc/test/Generator/FileDraft/CopyAndModifyTextFileDraftTest.hs
Martinsos: I removed `SP.` and imported SP stuff with list!
That said, I wouldn't say we have a strict convention there -> I think we do prefer somewhat importing it as a list of stuff vs qualified because we use it quite a lot in the codebase so it becomes boring to look at `SP.` all the time, but on the other hand it can be ok situationally to use `SP.` sometimes, it is up to the code author really I think.

waspc/test/Generator/FileDraft/CopyAndModifyTextFileDraftTest.hs
Martinsos: Well not officially, I just used it to group them :D, but I actually didn't achieve much here so now I put them all into separate lines.

waspc/cli/src/Wasp/Cli/Command/Db/Start.hs
Martinsos: That would be cool! I had that down as an idea, but decided to punt on it for now, I think it could potentially complicate things quite a bit right now.

waspc/cli/exe/Main.hs
Martinsos: Maybe we could -> I was thought thinking about it more as an implementation detail, you know they don't need to know how it is started. Although it will complain if they don't have docker :D. How would you write it? "Starts development database (via Docker)"?

waspc/cli/src/Wasp/Cli/Command/Db/Start.hs
Martinsos: Ah I know, it gave it to me also but I didn't like it I thought it reduces readibility!

waspc/cli/src/Wasp/Cli/Command/Db/Reset.hs
Martinsos: Ah sorry about this, I still have to test `reset` properly, it is on my TODO list!
This will be interesting, it wants to re-generate clients. I think what I will want to do here is use the `--skip-generate` flag that they mention and then either not do anything, or run our own generation of clients -> I have to think about what is the correct approach.

waspc/src/Wasp/Generator/FileDraft/CopyAndModifyTextFileDraft.hs
Martinsos: Yeah I was not sure if I should do it or not to be honest!
Name turned out really long.
I was tempted to instead modify `CopyFileDraft` and make it have additional field that allows this modification.
But then that felt like I will be using too power full(draft) in 90% of cases so I decided to make this separate one.
Hm.

Martinsos:
> This is looking really cool so far, @Martinsos. Nice work! I think people will really like having this option! I left a few comments (including one error).
> 
> One general thing I noticed while testing on WSL2 is for some reason the `DATABASE_URL` seems missing. At first, `pg-boss` failed.
> 
> ```shell
> Server: Starting pg-boss...
> Server(stderr): pg-boss failed to start!
> Server(stderr): Error: SASL: SCRAM-SERVER-FIRST-MESSAGE: client password must be a string
> Server(stderr):     at Object.continueSession (/home/shayne/dev/wasp/waspc/examples/todoApp/.wasp/out/server/node_modules/pg/lib/sasl.js:24:11)
> Server(stderr):     at Client._handleAuthSASLContinue (/home/shayne/dev/wasp/waspc/examples/todoApp/.wasp/out/server/node_modules/pg/lib/client.js:261:12)
> Server(stderr):     at Connection.emit (node:events:513:28)
> Server(stderr):     at /home/shayne/dev/wasp/waspc/examples/todoApp/.wasp/out/server/node_modules/pg/lib/connection.js:119:12
> Server(stderr):     at Parser.parse (/home/shayne/dev/wasp/waspc/examples/todoApp/.wasp/out/server/node_modules/pg-protocol/dist/parser.js:40:17)
> Server(stderr):     at Socket.<anonymous> (/home/shayne/dev/wasp/waspc/examples/todoApp/.wasp/out/server/node_modules/pg-protocol/dist/index.js:11:42)
> Server(stderr):     at Socket.emit (node:events:513:28)
> Server(stderr):     at addChunk (node:internal/streams/readable:324:12)
> Server(stderr):     at readableAddChunk (node:internal/streams/readable:297:9)
> Server(stderr):     at Readable.push (node:internal/streams/readable:234:10)
> Server: This is coming from shared function.
> Server(stderr): node:internal/process/promises:288
> Server(stderr):             triggerUncaughtException(err, true /* fromPromise */);
> Server(stderr):             ^
> Server(stderr): 
> Server(stderr): [UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "#<PgBoss>".] {
> Server(stderr):   code: 'ERR_UNHANDLED_REJECTION'
> Server(stderr): }
> Server(stderr): 
> Server(stderr): Node.js v18.12.0
> Server: [nodemon] app crashed - waiting for file changes before starting...
> ```
> 
> Then I disabled it and saw the real error:
> 
> ```shell
> Server(stderr): Error: 
> Server(stderr): Invalid `prisma.task.count()` invocation:
> Server(stderr): 
> Server(stderr): 
> Server(stderr): error: Environment variable not found: DATABASE_URL.
> Server(stderr):   -->  schema.prisma:4
> Server(stderr):    | 
> Server(stderr):  3 |   provider = "postgresql"
> Server(stderr):  4 |   url      = env("DATABASE_URL")
> Server(stderr):    | 
> Server(stderr): 
> Server(stderr): Validation Error Count: 1
> ```
> 
> So in order to try to restart, I tried clean, but that failed with a perms error:
> 
> ```shell
> cabal run wasp-cli clean
> Up to date
> 
> 🐝 --- Deleting .wasp/ directory... -----------------------------------------------
> 
> wasp-cli: /home/shayne/dev/wasp/waspc/examples/todoApp/.wasp/.wasp-dev-db-data: removeDirectoryRecursive:removeContentsRecursive:removePathRecursive:removeContentsRecursive:getDirectoryContents:openDirStream: permission denied (Permission denied)
> ```
> 
> So I think there may be some WSL2 kinks to work out (or maybe my setup is not typical). Specifically, even with the Docker DB running (via `wasp db start`) my `DATABASE_URL` does not seem to be set. I did not have any `.env.server` file if that matters (I moved it so it wouldn't use my old PG DB).
> 
> For reference, I am just running Docker Desktop on Windows and using the WSL2 engine: https://docs.docker.com/desktop/windows/wsl/

Actually I think you just encountered legit bugs, irrelevant of WSL2!

But you also caught them well in the later comments.

One is the fact that db is created with weird permission so it is hard to delete.

Another one is that I didn't account for the possibility of .env.server not existing.

Will fix them! Thanks @shayneczyzewski !

waspc/cli/src/Wasp/Cli/Command/Db/Start.hs
Martinsos: Gave up on this but came up with a better idea that is more general, will create an issue for it (it is under TODOs in the main comment of this PR).

waspc/cli/exe/Main.hs
Martinsos: Ok, what I ended up doing was not deleting the db when it is postgres db, but sqlite one will end up deleted on `wasp clean`. 
It is just the easiest way to do it and I think in practice it will work just fine.
Only improvement could be to also void deleting sqlite on wasp clean, but that would be quite some work and is old behaviour anyway so I wouldn't worry about it for now.

waspc/cli/exe/Main.hs
Martinsos: I added the check for "is docker available", which prints a nice error message if it is not! In that case, is it ok that I don't mention Docker here?

waspc/cli/src/Wasp/Cli/Command/Db/Reset.hs
Martinsos: I just added the --skip-generate flag, that should do it, since we take care of generating clients anyway before running any db commands.

waspc/src/Wasp/Generator/ServerGenerator.hs
Martinsos: Fixed!

waspc/src/Wasp/Generator/FileDraft/CopyAndModifyTextFileDraft.hs
Martinsos: Actually I ended up not using this at the end, but I am considering leaving it in due to amount of work + potential it might be useful in the future. If you all think this is wrong decision let me know and I can remove it still.

waspc/cli/src/Wasp/Cli/Command/Db/Start.hs
```diff
@@ -0,0 +1,138 @@
...
+          P.defaultDevPass
+          P.defaultDevUser
+          dbName
+  liftIO $ callCommand command
+  where
+    dockerVolumeName = makeWaspDevDbDockerVolumeName waspProjectDir appName
+    dbName = P.makeDevDbName waspProjectDir appName
+    connectionUrl = P.makeDevConnectionUrl waspProjectDir appName
+
+throwIfExeIsNotAvailable :: String -> String -> Command ()
```
Martinsos, line 117: TODO for myself: extract this one into more common place.

waspc/src/Wasp/Db/Postgres.hs
```diff
@@ -0,0 +1,63 @@
...
+  )
+where
+
+import Data.Char (isAsciiLower, isAsciiUpper, isDigit)
+import StrongPath (Abs, Dir, Path', fromAbsDir)
+import Text.Printf (printf)
+import Wasp.Project.Common (WaspProjectDir)
+import qualified Wasp.Util as U
+
+-- | TODO: Write explanation on what "dev" stands for in all these variables.
```
Martinsos, line 18: TODO here!

waspc/src/Wasp/Generator/FileDraft/CopyAndModifyTextFileDraft.hs
shayneczyzewski: If we aren't using it, I wonder if it can easily be extracted into some other branch for future possible use? Only downside to keeping it if not used is the additional drag anytime we make `FileDraft` changes, we have another thing to update.

waspc/cli/src/Wasp/Cli/Command/Db/Start.hs
```diff
@@ -0,0 +1,129 @@
...
+        "Nothing to do! You are all good, you are using SQLite which doesn't need to be started."
+
+    throwIfThereIsCustomDbUrl :: AS.AppSpec -> Command ()
+    throwIfThereIsCustomDbUrl spec =
+      when (isThereCustomDbUrl spec) $
+        E.throwError $
+          CommandError
+            "You are using custom database already"
+            ( printf
+                ( "Wasp has detected that you have defined %s env var in your %s file.\n"
```
shayneczyzewski, line 57: Nice!!!

waspc/src/Wasp/Db/Postgres.hs
```diff
@@ -0,0 +1,15 @@
...
+  ( makeConnectionUrl,
+    postgresMaxDbNameLength,
+  )
+where
+
+import Text.Printf (printf)
+
+makeConnectionUrl :: String -> String -> Int -> String -> String
+makeConnectionUrl user pass port dbName =
+  printf "postgresql://%s:%s@localhost:%d/%s" user pass port dbName
```
shayneczyzewski, line 11: I like that it shows this, nice!

shayneczyzewski (approved):
Hey @Martinsos nice job here! I did some testing locally on WSL2 and macOS and it seems solid right now. Great improvements over the last round. There were lots of code changes and I tried to make it through most, but it may be hard to go line by line to verify it all :D Functionally it looks good. Made a few comments you can take or leave, but do check out he last concurrence check idea for reset plz.

Good from my book, but I'd still encourage others who are interested to review too in case I missed something given the size. 👍🏻

waspc/cli/exe/Main.hs
```diff
@@ -89,36 +95,40 @@ printUsage :: IO ()
 printUsage =
   putStrLn $
     unlines
+{- ORMOLU_DISABLE -}
```
infomiho, line 98: Why did you disable it?

waspc/src/Wasp/Project/Env.hs
```diff
@@ -0,0 +1,45 @@
...
+readDotEnvClient waspDir = readDotEnvFileInWaspProjectDir waspDir dotEnvServer
+
+-- | Checks if .env exists in wasp dir, and produces a warning if so.
+-- We have this function because Wasp doesn't use ".env", but still user
+-- might assume it does and then bother quite a bit trying to figure it out,
+-- so this way we warn if they assume so.
+warnIfTheDotEnvPresent :: Path' Abs (Dir WaspProjectDir) -> IO (Maybe CompileWarning)
+warnIfTheDotEnvPresent waspDir = (warningMessage <$) <$> findFileInWaspProjectDir waspDir [relfile|.env|]
+  where
+    warningMessage = "Wasp .env files should be named .env.server or .env.client, depending on their use."
```
infomiho, line 33: Great DX improvement

waspc/src/Wasp/Generator/WebAppGenerator.hs
```diff
@@ -74,15 +75,15 @@ genWebApp spec = do
...
-genDotEnv spec = return $
-  case AS.dotEnvClientFile spec of
-    Just srcFilePath
-      | not $ AS.isBuild spec ->
-          [ createCopyFileDraft
-              (C.webAppRootDirInProjectRootDir </> dotEnvInWebAppRootDir)
-              srcFilePath
-          ]
-    _ -> []
+-- Don't generate .env if we are building for production, since .env is to be used only for
```
infomiho, line 78: What if the user set some variables they might expect to see during the build phase? 

Maybe it's something like `IS_STAGE_BUILD` and they change their layout based on this?

waspc/src/Wasp/Generator/ServerGenerator.hs
```diff
@@ -83,15 +88,22 @@ genServer spec =
...
-    Just srcFilePath
-      | not $ AS.isBuild spec ->
-          [ createCopyFileDraft
-              (C.serverRootDirInProjectRootDir </> dotEnvInServerRootDir)
-              srcFilePath
-          ]
-    _ -> []
+-- Don't generate .env if we are building for production, since .env is to be used only for
+-- development.
+genDotEnv spec | AS.isBuild spec = return []
```
infomiho, line 93: Are we sure about this? The user might still expect to get this `.env.server` copied over?

waspc/src/Wasp/Generator/ServerGenerator.hs
```diff
@@ -83,15 +88,22 @@ genServer spec =
...
+        (C.serverRootDirInProjectRootDir </> dotEnvInServerRootDir)
+        (envVarsToDotEnvContent envVars)
+    ]
+  where
+    envVars = waspEnvVars ++ userEnvVars
+    userEnvVars = AS.devEnvVarsServer spec
+    waspEnvVars = case AS.devDatabaseUrl spec of
+      Just url | not isThereCustomDbUrl -> [(databaseUrlEnvVarName, url)]
+      _ -> []
+    isThereCustomDbUrl = any ((== databaseUrlEnvVarName) . fst) userEnvVars
```
infomiho, line 106: Nit: This block of code feels a bit messy, maybe extract it into a separate helper function

infomiho: `getWaspEnvVars` or similar

examples/realworld/README.md
```diff
@@ -8,13 +8,6 @@ Here, we implement it in Wasp, by following their [specification](https://realwo
 This app is deployed at https://wasp-rwa.netlify.app/ .
 
 # Development
-
-### Database
```
infomiho, line 12: If we are removing this block, why not mention `wasp start db` instead?

waspc/cli/src/Wasp/Cli/Command/Db/Start.hs
```diff
@@ -0,0 +1,129 @@
+module Wasp.Cli.Command.Db.Start
```
infomiho, line 1: The command is now in the `start` scope, shouldn't the module reflect that:

`Wasp.Cli.Command.Start.Db`

waspc/src/Wasp/Env.hs
```diff
@@ -0,0 +1,36 @@
...
+parseDotEnvFile :: Path' Abs (File ()) -> IO [EnvVar]
+parseDotEnvFile envFile =
+  Dotenv.parseFile (fromAbsFile envFile)
+    -- Parse errors are returned from Dotenv.parseFile as ErrorCall, which Wasp compiler would
+    -- report as a bug in compiler, so we instead convert these to IOExceptions.
+    `catch` \(ErrorCall msg) -> throwIO $ userError $ "Failed to parse dot env file: " <> msg
+
+envVarsToDotEnvContent :: [EnvVar] -> T.Text
+envVarsToDotEnvContent vars =
+  T.pack $ intercalate "\n" $ map (\(n, v) -> n <> "=" <> show v) vars
```
infomiho, line 36: Nit: `\(name, value) -> name <> "=" <> show value` would be more readable to me

waspc/src/Wasp/Project/Db/Dev/Postgres.hs
```diff
@@ -0,0 +1,35 @@
...
+defaultDevUser = "postgresWaspDevUser"
+
+defaultDevPass :: String
+defaultDevPass = "postgresWaspDevPass"
+
+makeDevDbName :: Path' Abs (Dir WaspProjectDir) -> String -> String
+makeDevDbName waspProjectDir appName =
+  -- We use makeDevDbUniqueId to construct a db name instead of a hardcoded value like "waspDevDb"
+  -- in order to avoid the situation where one Wasp app accidentally connects to a db that another
+  -- Wasp app has started. This way db name is unique for the specific Wasp app, and another Wasp app
```
infomiho, line 26: Is it possible to create multiple databases in an existing Docker container? I thought those env vars (`DATABASE_NAME` or something similar) you pass it on start were only used then and subsequent runs ignored them?

web/docs/cli.md
```diff
@@ -80,7 +81,7 @@ To setup Bash completion, execute `wasp completion` and follow the instructions.
 
 ### In project
  - `wasp start` runs Wasp app in development mode. It opens a browser tab with your application running, and watches for any changes to .wasp or files in `src/` to automatically reflect in the browser. It also shows messages from the web app, the server and the database on stdout/stderr.
-
+ - `wasp start db` start the database for you -> this can be very handy, since you don't need to spin up your own database or provide its connection URL to the Wasp app!
```
infomiho, line 84:
```suggestion
 - `wasp start db` starts the database for you. This can be very handy, since you don't need to spin up your own database or provide its connection URL to the Wasp app!
```

web/docs/cli.md
```diff
@@ -21,8 +21,9 @@ COMMANDS
     uninstall             Removes Wasp from your system.
   IN PROJECT
     start                 Runs Wasp app in development mode, watching for file changes.
+    start db              Starts managed development database for you.
     db <db-cmd> [args]    Executes a database command. Run 'wasp db' for more info.
-    clean                 Deletes all generated code and other cached artifacts. Wasp equivalent of 'have you tried closing and opening it again?'.
+    clean                 Deletes all generated code and other cached artifacts.                          Wasp equivalent of 'have you tried closing and opening it again?'.
```
infomiho, line 26: Does this render okay on different screen resolutions? It looks werid on my laptop 

Screenshot 2023-03-13 at 12 06 42

web/docs/language/features.md
```diff
@@ -1649,29 +1649,20 @@ Default database is `SQLite`, since it is great for getting started with a new p
...
-One of the easiest ways to run a PostgreSQL database on your own is by spinning up [postgres docker](https://hub.docker.com/_/postgres) container when you need it with the following shell command:
-```
-docker run \
-  --rm \
-  --publish 5432:5432 \
-  -v my-app-data:/var/lib/postgresql/data \
-  --env POSTGRES_PASSWORD=devpass1234 \
-  postgres
-```
+To help with this, Wasp provides `wasp start db` that starts the default db for you. Your Wasp app will automatically connect to it once you have it running via `wasp start db`, no additional configuration is needed. Wasp will require that you have `docker` installed for this though.
```
infomiho, line 1654:
```suggestion
To help with this, Wasp provides `wasp start db` that starts the default db for you. Your Wasp app will automatically connect to it once you have it running via `wasp start db`, no additional configuration is needed. This command relies on Docker being installed on your machine.
```

web/docs/language/features.md
```diff
@@ -1649,29 +1649,20 @@ Default database is `SQLite`, since it is great for getting started with a new p
...
 
 ### Migrating from SQLite to PostgreSQL
 To run Wasp app in production, you will need to switch from `SQLite` to `PostgreSQL`.
-1. Set `app.db.system` to `PostgreSQL` and set `DATABASE_URL` env var accordingly (as described [above](/docs/language/features#postgresql)).
-2. Delete old migrations, since they are SQLite migrations and can't be used with PostgreSQL: `rm -r migrations/`.
-3. Run `wasp db migrate-dev` to apply new changes and create new, initial migration. You will need to have your postgres database running while doing this (check [above](/docs/language/features#postgresql) for easy way to get it running).
+1. Set `app.db.system` to `PostgreSQL`.
+3. Delete old migrations, since they are SQLite migrations and can't be used with PostgreSQL: `rm -r migrations/`.
+3. Run `wasp start db` to start your new db running (or check instructions above if you prefer using your custom db). Leave it running, since we need it for the next step.
+4. Separately, run `wasp db migrate-dev` to apply new changes and create new, initial migration.
```
infomiho, line 1667:
```
```suggestion
4. In a different terminal, run `wasp db migrate-dev` to apply new changes and create new, initial migration.
```

waspc/cli/src/Wasp/Cli/Command/Db/Start.hs
```diff
@@ -0,0 +1,129 @@
...
+        "     " <> dockerVolumeName
+      ]
+
+  -- NOTE: POSTGRES_PASSWORD, POSTGRES_USER, POSTGRES_DB below are really used by the docker image
+  --   only when initializing the database -> if it already exists, they will be ignored.
+  --   This is how the postgres Docker image works.
+  let command =
+        printf
+          ( unwords
+              [ "docker run",
```
infomiho, line 95: Suggestion: we could add the `--name` param as well, so it looks nicer in the list of containers. We could use the same name as the db name 🤷

Screenshot 2023-03-13 at 12 16 07

infomiho:
Looking good overall @Martinsos! It works on my machine and I like the improvements done in the last round 👏

I've left some comments, most are minor, some things like the `info` output should be fixed IMHO.

One thing regarding the DX, I dislike the fact it's running in the foreground. This forces me to have an additional terminal running. It would be cool if in the future we could provide a flag `--background` and let the user know he needs to cleanup the container manually.

waspc/cli/src/Wasp/Cli/Command/Db/Reset.hs
```diff
@@ -0,0 +1,36 @@
...
+import Wasp.Cli.Command (Command)
+import Wasp.Cli.Command.Common (findWaspProjectRootDirFromCwd)
+import Wasp.Cli.Command.Message (cliSendMessageC)
+import qualified Wasp.Cli.Common as Common
+import qualified Wasp.Generator.DbGenerator.Jobs as DB.Jobs
+import Wasp.Generator.Job.IO (readJobMessagesAndPrintThemPrefixed)
+import qualified Wasp.Message as Msg
+
+reset :: Command ()
+reset = do
```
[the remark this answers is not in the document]
Martinsos: Good point I haven't thought about it!
The thing is, `prisma migrate reset` deletes the whole database and then applies all migrations to it. So once it is done, the migrations should be all set and no need to run migrate-dev.

So what is the best thing to do then? I guess just deleting that file is not enough, because then it will ask user to run migrate-dev which we don't want to do?

Or should we instead re-generate that file?

I see now that migrateDev also calls `  generatePrismaClients projectRootDir
` at its very end. Is that something we should also call here? But why hm, aren't prisma clients generated when we compile the project? And don't they depend on the schema, and not the migrations? Now I wonder why they are generated even at the end of the migrate-dev! If you can help me detangle this stuff that would be very helpful.

waspc/cli/src/Wasp/Cli/Command/Db/Start.hs
```diff
@@ -0,0 +1,129 @@
...
+import Wasp.Project.Db.Dev (makeDevDbUniqueId)
+import qualified Wasp.Project.Db.Dev.Postgres as Dev.Postgres
+import Wasp.Project.Env (dotEnvServer)
+
+-- | Starts a "managed" dev database, where "managed" means that
+-- Wasp creates it and connects the Wasp app with it.
+-- Wasp is smart while doing this so it checks which database is specified
+-- in Wasp configuration and spins up a database of appropriate type.
+start :: Command ()
+start = do
```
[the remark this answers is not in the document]
Martinsos: Uff this is weird! I would expect in this situation that `wasp start db` fails due to not being able to obtain the port 5432.
I think the solution we want here is for `wasp start db` to properly fail.
In the future we can do dynamic ports, but for now this would be the best.

The main thing to figure out is why it doesn't fail! Maybe let's go on a call together and you can show me what this looks like on your machine.

waspc/cli/src/Wasp/Cli/Command/Db/Reset.hs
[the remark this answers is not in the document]
Martinsos: Talking with Shayne, TODOs for myself:
- I should reset the checksum and then regenerate if all is successful, if not I don't do anything, it remains deleted. But I might not even want to write it, since if it is not there it will just result with deeper check that will anyway generate that file if it succeeds.
- We could also remove it when we do `wasp start db`.
- ask Filip why he put generatePrismaClients at the end of the migrateDev. Should I do it in Wasp reset? Is it connected to prisma reset asking to do generation?

waspc/cli/src/Wasp/Cli/Command/Db/Start.hs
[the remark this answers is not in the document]
Martinsos: So what Shayne said actually happens!
So we don't know, maybe it is a docker + mac thing?
I can test on LInux the same thing.
One solution could be to use non-standard port, like 5433 or something..

waspc/src/Wasp/Generator/FileDraft/CopyAndModifyTextFileDraft.hs
Martinsos: Yeah I know. Ah man. Probably best to remove it then!

waspc/cli/src/Wasp/Cli/Command/Db/Start.hs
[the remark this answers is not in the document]
Martinsos: I learned this is due to Docker havig its own network, and ports on host machine are forwarded to it, so native processes "steal" the stuff that comes to host machine ports and it doesn't get forwarded, if there is a native process! Interesting. I added a function to Wasp that checks if port 5432 is occupied and throws an error if it, so now for this situation `wasp start db` will throw an error! Yay!

waspc/cli/exe/Main.hs
Martinsos: Because I wanted to format nicely the stuff below, usage strings, so you can read it same as you would read it in the terminal! Otherwise it was a bit of a mess.

waspc/src/Wasp/Project/Env.hs
Martinsos: Ha this is actually old code, written by Shayne :D. Sorry for that I moved too much stuff around.

waspc/src/Wasp/Generator/WebAppGenerator.hs
Martinsos: Philosophy of dotenv is that it shouldn't have anything to do with build/production, only with development, check it here: https://www.npmjs.com/package/dotenv .

If we figure out we need something like what you mentioned, we can relax this and think about it, but for now we are just sticking to what they say. We also make this clear in docs so there should be no confusion.

waspc/src/Wasp/Generator/ServerGenerator.hs
Martinsos: See answer above, but short answer is that they will not expect because dotenv philosophy is what we do here + we also mention it in docs.

waspc/src/Wasp/Generator/ServerGenerator.hs
Martinsos: I am up for making it clearer but I have to admit I am not sure how to go about it.
This is stuff in the context of dot env file -> we have all env vars, which are really wasp env vars or user env vars, and then we define how to get user env vars, how to get wasp env vars, and finally a check for if there is custom db url.

What do you find unclear, how would you suggest improving it? I think taking it out into standalone helper function would not be great because it would be missing the context of the server dot env file.

examples/realworld/README.md
Martinsos: Good point, added!

waspc/cli/src/Wasp/Cli/Command/Db/Start.hs
Martinsos: I thought it is ok to leave it in Db.Start because I felt it logically belong there -> it deals with db, and it starts it. And I saw `wasp start db` as a UX/DX thing.
But I moved it now, I don't think it makes much difference at the end, and it is a bit confusing that all other commands reflect in their positioning how they are called from CLI but this one.

waspc/src/Wasp/Project/Db/Dev/Postgres.hs
Martinsos: No, not really! At least I don't think this is the idea how this Docker image should be used, nor do we do it.

Yes, they are used only once. Or to put it better, they are used only if initialized db is not already present on disk. And disk in this case is docker volume. And we have one docker volume per one wasp app.

When I say we want to avoid situation where one app connects to another one's db is following: imagine you have two Wasp apps, WA1 and WA2. In WA1 you run `wasp start db`. Then you go to WA2 and run `wasp start` there. If WA1 hosted that db on the same connection URL as WA2 is using to connect to dev db, then WA2 will happily connect to it and you will never know. Since connection url is unique in following: db name, user name, pass, port, then if all of those are the same, this can happen!
Which is why I make it so that db name is unique per app, so this can't happen.

web/docs/cli.md
Martinsos: Done!

web/docs/cli.md
Martinsos: Nice catch, it was a mistake, fixed!

web/docs/language/features.md
Martinsos: Sure, done! I am curious, why do you think this is better?

waspc/cli/src/Wasp/Cli/Command/Db/Start.hs
Martinsos: Good idea, I did it!

Martinsos:
@infomiho I took care of all the suggestions, please take a look again!

The only thing I didn't do is enabling running in the background -> I think that is an interesting idea, but thought I would leave that for the future, since this should already bring a lot of value and it is hard for me to say right now how much people cares about running it in background -> one nice thing about running it in the foreground is that you can see logs easily + keeps you aware it is running. I could open an issue for the background flag, or we can leave it be and wait till somebody else also suggests it.

Martinsos:
This should wrap it all up -> pls give it another look @shayneczyzewski @infomiho !
I did everything you suggested (check individual comments) and also added checking if port 5432 is open, to avoid situations were it is taken but Wasp doesn't recognize it.

waspc/src/Wasp/Generator/WebAppGenerator.hs
infomiho: Understood, makes sense 👍

infomiho:
Running the dev database in a different project while a dev db is already running, causes the following error:
```
docker: Error response from daemon: driver failed programming external connectivity on endpoint wasp-dev-db-test-e24d36af4c (9deb4264fb8fb03648280f12de5b8e8d127f19c0a465e7f545bd55f99dc2215e): Bind for 0.0.0.0:5432 failed: port is already allocated.
wasp-cli: callCommand: docker run --name wasp-dev-db-test-e24d36af4c --rm --publish 5432:5432 -v wasp-dev-db-test-e24d36af4c:/var/lib/postgresql/data --env POSTGRES_PASSWORD=postgresWaspDevPass --env POSTGRES_USER=postgresWaspDevUser --env POSTGRES_DB=test-e24d36af4c postgres (exit 125): failed
```

This is expected of course since we are using the same port on host machine for both containers.

I wanted to note this, but I don't expect our users using two dev dbs at the same time. Maybe add something in the docs "You can run only one dev database at a time"?

infomiho (approved):
Looking good, works on my machine, db starts well and I like the DX improvements you did from the last review.

I will patiently wait for the addition of background mode (detached container), since that's my ideal workflow. 

Left a single nit comment.

Martinsos:
> Running the dev database in a different project while a dev db is already running, causes the following error:
> 
> ```
> docker: Error response from daemon: driver failed programming external connectivity on endpoint wasp-dev-db-test-e24d36af4c (9deb4264fb8fb03648280f12de5b8e8d127f19c0a465e7f545bd55f99dc2215e): Bind for 0.0.0.0:5432 failed: port is already allocated.
> wasp-cli: callCommand: docker run --name wasp-dev-db-test-e24d36af4c --rm --publish 5432:5432 -v wasp-dev-db-test-e24d36af4c:/var/lib/postgresql/data --env POSTGRES_PASSWORD=postgresWaspDevPass --env POSTGRES_USER=postgresWaspDevUser --env POSTGRES_DB=test-e24d36af4c postgres (exit 125): failed
> ```
> 
> This is expected of course since we are using the same port on host machine for both containers.
> 
> I wanted to note this, but I don't expect our users using two dev dbs at the same time. Maybe add something in the docs "You can run only one dev database at a time"?

Ha ok good question: I was hoping they would recognize from the fact that port is taken (`Bind for 0.0.0.0:5432 failed: port is already allocated.`) that there is something else running on this port, which might be another dev db.

Btw I think we might relatively soon add dynamic port choosing, so in that case they could even run multiple dev dbs. Even though that will rarely be the case.

I feel like itwould be a bit too much in the docs, so I would maybe skip it, but I can put it if you feel strongly about it.

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -0,0 +1,175 @@
...
+startPostgreDevDb :: Path' Abs (Dir WaspProjectDir) -> String -> Command ()
+startPostgreDevDb waspProjectDir appName = do
+  throwIfExeIsNotAvailable
+    "docker"
+    "To run PostgreSQL dev database, Wasp needs `docker` installed and in PATH."
+  throwIfDevDbPortIsAlreadyInUse
+
+  cliSendMessageC . Msg.Info $
+    unlines
+      [ "✨ Starting dev database for your Wasp app (detected PostgreSQL) ✨",
```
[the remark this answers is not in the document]
shayneczyzewski: I think that updated wording sounds good. 👍🏻

PR #1045, opened 2023-03-07
Fixes tsconfig.json use of extends 

infomiho:
Our build step is broken, this fixes it.

Updates `tsconfig.json` in the templates and the e2e tests.

It's weird that my editor didn't scream at me that `extends` should be outside of `compilerOptions`.


shayneczyzewski (approved):
Did another test deploy and this one works. Thanks for finding and fixing after a late in the day notification. :) 🚀

PR #1048, opened 2023-03-09
Adds support for FE listening on multiple addresses

infomiho:
This PR enables using Wasp in environemtns where the user is not visiting the FE through `localhost:3000` 

It also allows all origins in the CORS rules when in development mode, so the backend will work with requests from both `localhost:3000` or `192.168.1.1:3000`

### DIFF
--- a/waspc/data/Generator/templates/server/src/config.js
+++ b/waspc/data/Generator/templates/server/src/config.js
@@ -16,22 +16,38 @@ const config = {
     port: parseInt(process.env.PORT) || 3001,
     databaseUrl: process.env.{= databaseUrlEnvVar =},
     frontendUrl: undefined,
+    allowedCORSOrigins: [],
     {=# isAuthEnabled =}
     auth: {
       jwtSecret: undefined
     }
     {=/ isAuthEnabled =}
   },
-  development: {
-    frontendUrl: stripTrailingSlash(process.env.WASP_WEB_CLIENT_URL) || 'http://localhost:3000',
+  development: getDevelopmentConfig(),
+  production: getProductionConfig(),
+}
+
+const resolvedConfig = merge(config.all, config[env])
+export default resolvedConfig
+
+function getDevelopmentConfig() {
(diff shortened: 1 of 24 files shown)


waspc/data/Generator/templates/server/src/config.js
```diff
@@ -16,29 +16,42 @@ const config = {
     port: parseInt(process.env.PORT) || 3001,
```
infomiho, line 16: Development config has origin `*` which allows all origins. 

Production configs uses the `WASP_WEB_CLIENT_URL` env variable like before.

shayneczyzewski: Is there a reason to prefer `*` over a specific URL for local dev? My only fear is if they do something in dev they think will work but will fail in prod for CORS reasons.

waspc/data/Generator/templates/server/src/config.js
```diff
@@ -16,29 +16,42 @@ const config = {
...
     {=# isAuthEnabled =}
     auth: {
       jwtSecret: undefined
     }
     {=/ isAuthEnabled =}
   },
-  development: {
-    frontendUrl: stripTrailingSlash(process.env.WASP_WEB_CLIENT_URL) || 'http://localhost:3000',
+  development: getDevelopmentConfig(),
+  production: getProductionConfig(),
```
shayneczyzewski, line 27: Nice job pulling it into helpers! 💪🏻

shayneczyzewski (approved):
Nice work @infomiho! Did some local testing, all looks good. Did a test deploy and still good! 🚀 Can't guarantee anything about Replit but seems like this should help. Nice job 👍🏻

waspc/data/Generator/templates/server/src/config.js
shayneczyzewski: Ah wait, nvmd 🤦🏻‍♂️ we **need** that for Replit sorry lol

sodic (approved):

PR #1050, opened 2023-03-10
Adds support for e-mail sending

infomiho:
Users can define their `emailSender` provider in their Wasp file and then use the `emailSender.send({ ... })` method in their server-side JS code.

Supported providers for now:
- SMTP
- SendGrid
- MailGun

I've kept the two extra providers, although we agreed we want to reduce to SMTP. I just wanted to hear from you guys one more time if we maybe want to support SendGrid and MailGun out of the box?


waspc/data/Generator/templates/server/src/email/core/helpers.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export function createFromEmailString({
+  email,
+  title,
+}: {
+  email: string;
+  title?: string;
+}): string {
+  if (title) {
+    return `${title} <${email}>`;
```
infomiho, line 13: SMTP from field can look like `Name <email@bla.com>` and it will look nice in the e-mail client.

waspc/data/Generator/templates/server/src/email/core/helpers.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+  return email;
+}
+
+export function getDefaultFromField(): EmailFromField {
```
infomiho, line 18: User must setup the default sender, but it he can override it when using the `emailSender.send({ ... })`

waspc/data/Generator/templates/server/src/email/core/helpers.ts
shayneczyzewski: Can we maybe rename `title` to `name`? I think that is a bit more common (but could be wrong).

shayneczyzewski (commented):
Overall this is looking really good so far @infomiho, and I like how we have multiple options out of the gate. 🥳 And the abstractions look good.

A few notes:
1) It would be good if we added some general docs about this, but also maybe even setup guides for SendGrid and Mailgun.
2) I would consider changing `title` to `name` (mainly since I see that more in provider docs).
3) Do you have any thoughts how we may want to handle dev mode? Some frameworks give you a "local mailbox" so it doesn't have to really send them in dev mode. Does nodemailer offer anything like this? Can/should we maybe just console log them in dev mode, for example by default (unless they set some ENVAR to really send)? I know sometimes people get bit by using prod-like data and actually generate dev/test emails lol.
4) Any thoughts on if we should do anything around templating for v1? Is there maybe a quick and dirty way we can support it? I feel it could go a long way if we can.

Great job so far! Will do deeper dive tomorrow 👍🏻

waspc/examples/todoApp/todoApp.wasp
```diff
@@ -34,7 +34,14 @@ app todoApp {
   },
   db: {
     system: PostgreSQL
-  }
+  },
+  emailSender: {
```
Martinsos, line 38: [Head in the clouds]: what if we would make `emailSender` a top level declaration, and then you can have multiple of them, and when something is using email sender, you have to specify which one you want to use?
I wouldn't do this now, neither am I sure that we should ever do it, but I am mentioning this because I think it might be interesting wasp-design-wise -> this idea that some of this stuff could be standalone, and we could have multiple of them. Like `emailSender`, or `server`, or even `db`.
I guess with email sender, it could make some sense, the question is would anybody ever need it -> maybe they want to use different providers for different stuff hm? Weird. Also, does it make any sense to SendGrid providers? I don't know. But if it is standalone like that and named, then it can also be passed around as a dependency, maybe it can be imported from external module, ... .
Ok, just food for thought a bit, nothing you need to do on this PR.

waspc/src/Wasp/AppSpec/App/EmailSender.hs
```diff
@@ -0,0 +1,30 @@
...
+
+data EmailProvider = SMTP | SendGrid | Mailgun
+  deriving (Eq, Data)
+
+instance Show EmailProvider where
+  show SMTP = "smtp"
+  show SendGrid = "sendgrid"
+  show Mailgun = "mailgun"
+
+data EmailFrom = EmailFrom
```
Martinsos, line 26: I think I saw you calling this `EmailFromField` in other pieces of code? What is better, `EmailFrom` or `EmailFromField`? I think I like the second one a bit better, it seems to provide a bit more information.

waspc/src/Wasp/AppSpec/App/EmailSender.hs
```diff
@@ -0,0 +1,30 @@
...
+  }
+  deriving (Show, Eq, Data)
+
+data EmailProvider = SMTP | SendGrid | Mailgun
+  deriving (Eq, Data)
+
+instance Show EmailProvider where
+  show SMTP = "smtp"
+  show SendGrid = "sendgrid"
+  show Mailgun = "mailgun"
```
Martinsos, line 24: Why are you doing this?
Since default implementation of Show would already work quite well, and this one only has everything lowercase, I am guessing you had a good reason for doing this -> some other part of code is expecting it to be lowercase.
If that is so, then `Show` is not the best thing to use -> `Show` is meant to be used as a debugging / prettifying thing, not to provide string ids that the rest of the code will rely on.
So if you need string ids, I would rather suggest implementing a function like `toStringId` or something like that.

waspc/src/Wasp/Generator/ServerGenerator/EmailG.hs
```diff
@@ -0,0 +1,143 @@
...
+    sequence
+      [ genIndex emailSender
+      ]
+      <++> genCore emailSender
+  Nothing -> return []
+  where
+    maybeEmailSender = AS.App.emailSender $ snd $ getApp spec
+
+genIndex :: EmailSender -> Generator FileDraft
+genIndex email = return $ C.mkTmplFdWithDstAndData srcPath dstPath (Just tmplData)
```
Martinsos, line 31: I thought we have a `mkTmpl` function that assumes same location in templates and destination -> if we do, maybe you can use that one, so you don't have to repeat it twice in the `where`?

waspc/src/Wasp/Generator/ServerGenerator/EmailG.hs
```diff
@@ -0,0 +1,143 @@
...
+    _ -> return []
+
+genMailgun :: EmailSender -> Generator [FileDraft]
+genMailgun email =
+  case AS.EmailSender.provider email of
+    AS.EmailSender.Mailgun ->
+      sequence
+        [ copyTmplFile [relfile|email/core/providers/mailgun.ts|]
+        ]
+    _ -> return []
```
Martinsos, line 97: Could you unite all these into one function called `genEmailSender` (I made this name up, you might have a better idea of what is a good name here, I had a bit of trouble figuring it out from the context) and then it would have one big `case` which would check which provider it is and based on it return correct filedraft?

You could reuse than the `copyTmplFile [relfile|email/core/providers/` parts also I believe.

waspc/src/Wasp/Generator/ServerGenerator/EmailG.hs
```diff
@@ -0,0 +1,143 @@
...
+mailgunVersionRange :: SV.Range
+mailgunVersionRange = SV.Range [SV.backwardsCompatibleWith (SV.Version 0 5 1)]
+
+mailgunDependency :: AS.Dependency.Dependency
+mailgunDependency = AS.Dependency.make ("ts-mailgun", show mailgunVersionRange)
+
+depsRequiredByEmail :: AppSpec -> [AS.Dependency.Dependency]
+depsRequiredByEmail spec =
+  concat
+    [ [nodeMailerDependency | provider == Just AS.EmailSender.SMTP],
```
Martinsos, line 138: Here is a thought:
So you have a bunch of function here in this file, and a lot of them do checks to figure out which provider is it, and then based on that, provide correct data/logic/result.

So in bunch of places we have "if smtp then X else if sendgrid then Y else if mailgun then Z" -> I am paraphrasing, but something like that.

That is not bad, it is pretty normal for functional programming, but when reading this file it did feel a bit dissipated maybe, like it is hard to put together a picture of what we end up with if we do smtp, what we end up with if we do sendgrid, ... .

Maybe, it would make sense to reorganize it a bit, and have one central function that goes smth like `getEmailSenderInfo` (ok name sucks, and maybe that indicates that this is the wrong approach) that has one big `case` and then returns all the stuff for that provider -> its depenencies, its core stuff, and maybe there is something more?

I am not sure if this is better than what yo have right now, but it sounds potentially interesting, so I think it is maybe worth a quick try and if it looks better, great.

waspc/src/Wasp/Generator/ServerGenerator/EmailG.hs
```diff
@@ -0,0 +1,143 @@
...
+      [sendGridDependency | provider == Just AS.EmailSender.SendGrid],
+      [mailgunDependency | provider == Just AS.EmailSender.Mailgun]
+    ]
```
Martinsos, line 141: One thing I think we can improve about this list of lists is that you already know it will always be just one of these, but it looks like it might be a combo of them. So I would rather we be explicit about that, and do a `case` that cases on provider and based on that returns correct dependency. Then I am reading the code and it is very clear what the output can be, while it looks like it could be a combo, but it can't. You will also get for free compiler warning when new provider is added, since it will be aware that that case wasn't handled in the `case`, while here you don't have that.

Martinsos (commented):
@infomiho this is looking great!

I think the core stuff is great, looking solid and nice.
I left some comments, mostly minor stuff.

@shayneczyzewski you asked about giving them a dev option for sending emails -> that makes sense! I used SendGrid in one project some time ago, and if I remember correctly, what you can do is get "testing" keys from them and then use those for some tests. Or was that Stripe? Yeah Stripe does it 100%, but I think it was also Sendgrid? Anyway, I agree we should have a clear opinion on how to easily do this stuff in testing, and it might be worth checking out if they offer test keys, maybe that can be an ok solution (although how would that work for SMTP hm -> I guess they would specify creds to "test" smtp). Well how do we want stuff to behave in testing, do we want to still send real emails but via "test" sender to not polute sending reputation and metrics, or do we want to not even send emails but just mock that we are sending them? Hmmmm.

waspc/data/Generator/templates/server/src/email/core/helpers.ts
```diff
@@ -0,0 +1,25 @@
+{{={= =}=}}
+
+import { EmailFromField } from "./types";
+
+export function createFromEmailString({
```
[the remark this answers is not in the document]
infomiho: A very valid comment, I'll come up with something more understandable  🤣

waspc/data/Generator/templates/server/src/email/core/providers/mailgun.ts
```diff
@@ -0,0 +1,21 @@
...
+  const mailer = new NodeMailgun(config.apiKey, config.domain);
+
+  const defaultFromField = getDefaultFromField();
+
+  return {
+    async send(email) {
+      const fromField = email.from || defaultFromField;
+      mailer.fromEmail = fromField.email;
+      mailer.fromTitle = fromField.title;
+      mailer.init();
```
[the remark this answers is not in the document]
infomiho: The way the Mailgun lib is written you can't override the from field when sending the e-mail but only on init.

I've gone into their source code and the init method only sets some fields, which makes calling the `init` multiple times harmless 👍

waspc/src/Wasp/AppSpec/App/EmailSender.hs
infomiho: Thank you for the tip! I thought `Show` instance is the right way to do this.

I had some earlier ideas to use this key (lowercase name) to map things in the templates, but I decided not to do it. I can remove this custom impl and just derive `Show` instead.

waspc/src/Wasp/Generator/ServerGenerator/EmailG.hs
infomiho: That's very nice idea, I'll try to rewrite this piece to have all of the provider data grouped in a place and then we just switch between them.

infomiho:
@shayneczyzewski I think having a "dev mode" that just kinda outputs things in the console instead of sending things is a great idea. 

When I was researching other libs, I found that one of them (https://github.com/forwardemail/email-templates) does things like this:
1. If it's `process.env.NODE_ENV === "DEVELOPMENT"` then don't send e-mails by default
2. If the user passes an override like `send: true` it will send it irregardless of dev or prod mode
3. In production, send e-mails

So, we might go for something similar. Like a property in the Wasp file like `areEmailsSentInDev: true`?

@Martinsos Thanks for the great comments, I'll restructure the code a bit, I feel like it will be much cleaner!

shayneczyzewski:
> @shayneczyzewski I think having a "dev mode" that just kinda outputs things in the console instead of sending things is a great idea.
> 
> When I was researching other libs, I found that one of them (https://github.com/forwardemail/email-templates) does things like this:
> 
> 1. If it's `process.env.NODE_ENV === "DEVELOPMENT"` then don't send e-mails by default
> 2. If the user passes an override like `send: true` it will send it irregardless of dev or prod mode
> 3. In production, send e-mails
> 
> So, we might go for something similar. Like a property in the Wasp file like `areEmailsSentInDev: true`?
> 

Yeah, I think that is the right approach. One suggestion may be to consider also keeping the "send in dev" flag as an envar too, instead of in the .wasp file. The reason is we already decoupled the connection string to envars, and I would imagine they would probably change those envars depending on the environment (dev, stage, etc.). Since wasp doesn't yet support multiple envs, I think keeping the send behavior an envar colocates it with what it is operating on (the SMTP account info) and makes it more flexible/transparent from a configuration point of view (to me at least).

waspc/data/Generator/templates/server/src/email/core/helpers.ts
infomiho: Makes sense, I've renamed it to `name` since a bit more common then `title`

waspc/examples/todoApp/todoApp.wasp
```diff
@@ -34,7 +34,14 @@ app todoApp {
...
+    provider: SendGrid,
+    defaultFrom: {
+      title: "Hello",
+      email: "mihovil@ilakovac.com"
+    },
```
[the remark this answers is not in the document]
infomiho: If it's not required here, then we would need to make it a required field when using the Typescript function.

If we want to then provide a convenience of optionally defining the default sender, the Typescript would be more complicated. I'll investigate how can it be done, if it's not that complicated, I'll see to make the `defaultFrom` optional.

waspc/src/Wasp/AppSpec/App/EmailSender.hs
infomiho: Agreed 👍

waspc/src/Wasp/Generator/ServerGenerator/EmailG.hs
infomiho: Forgot about that, thanks 👍

waspc/src/Wasp/Generator/ServerGenerator/EmailG.hs
infomiho: I've tried to do that in `Wasp.Generator.ServerGenerator.EmailSender.Providers`, grouping of all relevant data for each of the providers.

shayneczyzewski (commented):
Overall this is looking really nice, @infomiho. I left a few comments for your consideration and one possible template error. Otherwise, I think it's looking good to go! 👍🏻

waspc/data/Generator/templates/server/src/email/index.ts
```diff
@@ -0,0 +1,28 @@
+{{={= =}=}}
+
+import { initEmailSender } from "./core/index.js";
+
+{=# isSmtpProviderUsed =}
+const emailProvider = {
+    type: "smtp",
+    host: process.env.SMTP_HOST,
```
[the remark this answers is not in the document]
Martinsos: Good analysis!

One benefit of big config.js is that you can easily see, in one place, how app is configured. You know all the env vars are there. You can also easily play there with defining different configs -> dev, prod, staging, ... . This was popular approach some time ago (>5 years), maybe it is not these days though.

That said, I don't want to overcomplicate things, so ok let's leave it as it is. Once we come to the point where we will formalize ENV vars more, we can look into all this in more details.

waspc/examples/todoApp/todoApp.wasp
Martinsos: @infomiho bump on this one!

waspc/src/Wasp/Generator/ServerGenerator/EmailG.hs
Martinsos: You haven't implemented this yet, right? Or maybe I misremembered and there is no such function?

Martinsos (approved):

waspc/data/Generator/templates/server/src/email/core/helpers.ts
```diff
@@ -0,0 +1,37 @@
...
+// { email: "test@test.com, name: "Test" } -> "Test <test@test.com>"
+export function formatFromField({
+  email,
+  name,
+}: {
+  email: string;
+  name?: string;
+}): string {
+  if (name) {
+    return `${name} <${email}>`;
```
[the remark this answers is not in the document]
infomiho: If they don't provide a `defaultFrom` the type is updated to make the `from` field required. But that helps only if they are in TS world. But in JS they can do what they want with any of the fields.

waspc/data/Generator/templates/server/src/email/core/helpers.ts
infomiho: > I thought defaultFrom was optional? Do you mean if they set it, then they must also set email?
This comment is outdated, sorry about that. `defaultFrom` is optional.

I'll check out the error 👍

waspc/examples/todoApp/todoApp.wasp
infomiho: I've read the comment, and didn't respond due to the note:

> Ok, just food for thought a bit, nothing you need to do on this PR.

waspc/examples/todoApp/src/server/jobs/sendTaskUpdate.js
```diff
@@ -0,0 +1,32 @@
+import { emailSender } from '@wasp/email/index.js'
+
+export async function sendTaskUpdateFn(args, context) {
```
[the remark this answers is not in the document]
infomiho: That's a good suggestion, it's a leftover from me testing it out locally.

waspc/data/Generator/templates/server/src/email/core/helpers.ts
infomiho: The error was caused by me sending `Nothing` as a value to the templates (instead of falling back to empty strings for example) and we can't do that. I've updated the code to used `fromMaybe "" ...` 

https://github.com/wasp-lang/wasp/pull/1050/commits/bd9070262216de3c0c9048f87a42e4f87373d909

shayneczyzewski (approved):
Updates look good to me, nice job! 🥳 🚀

PR #1055, opened 2023-03-14
Adds `vitest` for testing client code

shayneczyzewski:
### Description

This PR adds `vitest` support for testing client code. It supports both unit tests and React component tests.

A few notes:
- It does not provide any hiding of the fact that it is `vitest` based, and users should just import from that directly for now. In the future, we may put it behind some `@wasp/test` namespace, but that will depend on how we implement server-side testing and if that is even important.
- Since `wasp test client` also runs `watch` to recompile code, it should not be run alongside `wasp start`.
- Any extra args supplied to `wasp test client <...>` will just be passed to `vitest`.
- It does not allow any test section config overrides of vite.config.ts (similar to our current Vite support).

TODO:
- [ ] update docs again based on any API changes during PR review


waspc/examples/todoApp/src/client/Todo.test.tsx
```diff
@@ -0,0 +1,29 @@
+import { test, expect } from 'vitest'
+import { screen } from '@testing-library/react'
+
+import { mockQuery, renderWrapped } from '@wasp/vitest.helpers'
```
shayneczyzewski, line 4: This probably needs a better name than `vitest.helpers`. Any ideas?

waspc/cli/src/Wasp/Cli/Command/Test.hs
```diff
@@ -0,0 +1,49 @@
...
+import Wasp.Cli.Command.Watch (watch)
+import qualified Wasp.Cli.Common as Common
+import Wasp.Lib (ProjectRootDir)
+import qualified Wasp.Lib as Lib
+import qualified Wasp.Message as Msg
+
+test :: [String] -> Command ()
+test [] = throwError $ CommandError "Not enough arguments." "Expected: wasp test client <args>"
+test ("client" : args) = watchAndtest $ Lib.testWebApp args
+test ("server" : _args) = throwError $ CommandError "Invalid arguments." "Server testing not yet implemented."
```
shayneczyzewski, line 25: Are we good with "client" and "server"? I know we sometimes call it web app too, but I don't think we have a definitive term used and client felt more natural/shorter here.

waspc/examples/todoApp/src/client/Todo.test.tsx
infomiho: I would go with `@wasp/test` for now?

infomiho: Maybe put everything related to testing in a `testing` folder.

```c
web-app/
-- testing/
---- vitest.setup.ts
---- index.ts // or helpers.ts
```

waspc/data/Generator/templates/react-app/src/vitest.helpers.tsx
```diff
@@ -0,0 +1,40 @@
...
+export const server = setupServer()
+
+export function initCallbacks() {
+  beforeAll(() => server.listen())
+  afterEach(() => server.resetHandlers())
+  afterAll(() => server.close())
+}
+
+export function mockQuery(query: Query<any, any>, resJson: any): void {
+  const route = query.queryCacheKey[0]
```
infomiho, line 34: This bit feel a bit fragile - using the cache key abstraction to get the path of the query.

I would suggest that we add another key on the query called `route` or similar (it can have the same value) but then the expectation is clear and changing query keys won't affect testing.

infomiho:
Here's a suggestion around the `Query` type. 

Typescript complains that `queryCacheKey` is not defined on it (which it isn't).

Screenshot 2023-03-23 at 13 56 53

This is how we can define a function type, but which also has some properties on it:
```ts
// web-app/src/queries/index.d.ts

export type Query<Input, Output> = {
  (args: Input): Promise<Output>
  queryCacheKey: string[]
}

// web-app/src/queries/core.d.ts

export type Query<Input, Output> = {
  (args: Input): Promise<Output>
  queryCacheKey: string[]
}
```

And we probably want to have both in the same file (by rewriting `core.js` and `index.js` to Typescript, but I understand it's maybe not in the scope of this PR).

waspc/src/Wasp/Generator/WebAppGenerator/Test.hs
```diff
@@ -0,0 +1,15 @@
...
+import StrongPath (Abs, Dir, Path', (</>))
+import Wasp.Generator.Common (ProjectRootDir)
+import qualified Wasp.Generator.Job as J
+import Wasp.Generator.Job.Process (runNodeCommandAsJob)
+import qualified Wasp.Generator.WebAppGenerator.Common as Common
+
+testWebApp :: [String] -> Path' Abs (Dir ProjectRootDir) -> J.Job
+testWebApp args projectDir = do
+  let webAppDir = projectDir </> Common.webAppRootDirInProjectRootDir
+  runNodeCommandAsJob webAppDir "npx" ("vitest" : args) J.WebApp
```
infomiho, line 15: I'm a bit confused if we are running the tests using the `npm run test` or `npx vitest`? Is one of those a leftover?

waspc/data/Generator/templates/react-app/vitest.setup.ts
```diff
@@ -0,0 +1,8 @@
+import matchers from '@testing-library/jest-dom/matchers'
```
infomiho, line 1: Is this setup file used implicitly?

infomiho: Another question, can we configure `vitest` to maybe reduce the amount of output on each test run?
Screenshot 2023-03-23 at 14 08 28

waspc/data/Generator/templates/react-app/src/vitest.helpers.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { rest } from 'msw'
+import { setupServer } from 'msw/node'
+import { BrowserRouter as Router } from 'react-router-dom'
+import { render } from '@testing-library/react'
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
+import { beforeAll, afterEach, afterAll } from 'vitest'
+import { Query } from './queries'
+import config from './config'
+
+export function renderWrapped(ui: React.ReactElement): any {
```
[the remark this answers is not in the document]
shayneczyzewski: Hmm, it won't always be an "app" component though, right? I guess swapping to `renderInContext` could more accurately describe it.

waspc/data/Generator/templates/react-app/src/vitest.helpers.tsx
shayneczyzewski: True, we could explicitly track the route separately. Will do

waspc/data/Generator/templates/react-app/vitest.setup.ts
shayneczyzewski: This file is referenced by `vite.config.ts`.

That output is from `screen.debug()`, mainly just for showing the render works. Most users won't leave that in there.

waspc/data/Generator/templates/react-app/src/vitest.helpers.tsx
```diff
@@ -0,0 +1,40 @@
...
+    rerender: (rerenderUi: React.ReactElement) =>
+      rerender(
+        <QueryClientProvider client={client}><Router>{rerenderUi}</Router></QueryClientProvider>
+      )
+  }
+}
+
+export const server = setupServer()
+
+export function initCallbacks() {
```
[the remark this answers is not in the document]
shayneczyzewski: Good point. The rub is I use `server` in both `initCallbacks`, which is our setup function, and also `mockQuery` uses it which they need.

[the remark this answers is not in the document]
shayneczyzewski: However, this does make me wonder if creating a server this way opens the door for issues if vitest can run multiple tests in parallel, then we'd have an issue with this shared global server. 🤔 I may need to rethink this setup a bit to ensure its safe concurrently.

shayneczyzewski:
I believe I implemented all your feedback @infomiho, ready for another check. Thanks!

waspc/data/Generator/templates/react-app/src/queries/core.d.ts
```diff
@@ -1,3 +1,7 @@
-export type Query<Input, Output> = (args: Input) => Promise<Output>
+export type Query<Input, Output> = {
+  (args: Input): Promise<Output>
+  queryCacheKey: string[]
+  route: { method: string, path: string}
```
shayneczyzewski, line 4: I should probably make `method` a proper type... does anyone know of anything in Node/Express/Axios I can make use of for a `'GET'` etc. constant equivalent?

waspc/data/Generator/templates/react-app/src/vitest.helpers.tsx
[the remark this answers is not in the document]
shayneczyzewski: Refactored it into a function like we chatted about to encapsulate it all 👍🏻

waspc/data/Generator/templates/react-app/src/queries/core.d.ts
infomiho: I would suggest defining a type like `type Method = "GET" | "POST" | ...`

infomiho:
https://user-images.githubusercontent.com/2223680/227963715-7748505a-d685-45ce-8069-cd36eaae8ef6.mov

When I exit the test watcher with `CTRL + C` it outputs an error message which I didn't expect because I wanted to stop it.

waspc/cli/src/Wasp/Cli/Command/Test.hs
```diff
@@ -0,0 +1,49 @@
...
+test ("client" : args) = watchAndTest $ Wasp.Generator.testWebApp args
+test ("server" : _args) = throwError $ CommandError "Invalid arguments." "Server testing not yet implemented."
+test _ = throwError $ CommandError "Invalid arguments." "Expected: wasp test client <args>"
+
+watchAndTest :: (Path' Abs (Dir ProjectRootDir) -> IO (Either String ())) -> Command ()
+watchAndTest testRunner = do
+  waspRoot <- findWaspProjectRootDirFromCwd
+  let outDir = waspRoot </> Common.dotWaspDirInWaspProjectDir </> Common.generatedCodeDirInDotWaspDir
+
+  cliSendMessageC $ Msg.Start "Starting compilation and setup phase. Hold tight..."
```
infomiho, line 33: Since the output is long:
```
🐝 --- Starting compilation and setup phase. Hold tight... ------------------------


✅ --- Your wasp project has successfully compiled. -------------------------------


🐝 --- Listening for file changes... ----------------------------------------------


🐝 --- Running tests... -----------------------------------------------------------
```

Maybe consider removing this message and merging the two from below into `Watching for file changes and running tests ...`

waspc/data/Generator/templates/react-app/src/queries/core.d.ts
infomiho, line 4:
```suggestion
  route: { method: string, path: string }
```

waspc/data/Generator/templates/react-app/src/queries/index.d.ts
```diff
@@ -1,6 +1,10 @@
 import { UseQueryResult } from "@tanstack/react-query";
 
-export type Query<Input, Output> = (args: Input) => Promise<Output>
+export type Query<Input, Output> = {
+    (args: Input): Promise<Output>
+    queryCacheKey: string[]
+    route: { method: string, path: string}
```
infomiho, line 6:
```suggestion
    route: { method: string, path: string }
```

waspc/data/Generator/templates/react-app/src/test/vitest/helpers.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { BrowserRouter as Router } from 'react-router-dom'
+import { render, RenderResult, cleanup } from '@testing-library/react'
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
+import { beforeAll, afterEach, afterAll } from 'vitest'
+import { Query } from '../../queries'
+import config from '../../config'
+
+export function renderInContext(ui: React.ReactElement): RenderResult {
+  const client = new QueryClient()
+  const { rerender, ...result } = render(
```
infomiho, line 13: @shayneczyzewski Why are we repeating `QueryClientProvider` twice in this file?

waspc/data/Generator/templates/react-app/src/test/vitest/helpers.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+    switch (route.method) {
+      case 'GET':
+        server.use(rest.get(url, responseHandler))
+        break
+      case 'POST':
+        server.use(rest.post(url, responseHandler))
+        break
+      default: throw new Error(`Unsupported method ${route.method}`)
+    }
```
infomiho, line 51: Maybe replace the switch case with an object mapping approach. This way Typescript will yell at you if you use something that's not a method. I would keep the runtime check as well, it doesn't hurt to check one more time.

```suggestion
    type HttpMethod = 'GET' | 'POST'
    type Handler = () => void
    const handlers: Record<HttpMethod, Handler> = {
      GET: () => {
        server.use(rest.get(url, responseHandler))
      },
      POST: () => {
        server.use(rest.post(url, responseHandler))
      },
    }
    const setupMock = handlers[route.method]
    if (!setupMock) {
      throw new Error(`Unsupported method ${route.method}`)
    }
    setupMock()
```

waspc/cli/src/Wasp/Cli/Command/Test.hs
```diff
@@ -0,0 +1,49 @@
...
+import Wasp.Cli.Command.Compile (compile)
+import Wasp.Cli.Command.Message (cliSendMessageC)
+import Wasp.Cli.Command.Watch (watch)
+import qualified Wasp.Cli.Common as Common
+import qualified Wasp.Generator
+import Wasp.Generator.Common (ProjectRootDir)
+import qualified Wasp.Message as Msg
+
+test :: [String] -> Command ()
+test [] = throwError $ CommandError "Not enough arguments." "Expected: wasp test client <args>"
```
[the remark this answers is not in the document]
shayneczyzewski: I'll have to check how the other titles work, but yeah maybe we can just remove `:` globally.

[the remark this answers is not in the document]
infomiho: Or you can remove the dot from your title :) but what ever feels right after checking out the other titles 👍

web/docs/testing.md
```diff
@@ -0,0 +1,104 @@
...
+import useBaseUrl from '@docusaurus/useBaseUrl';
+
+import AddExternalAuthEnvVarsReminder from './_addExternalAuthEnvVarsReminder.md'
+
+:::info
+Wasp is in beta, so keep in mind there might be some kinks / bugs, and possibly some changes with testing support in the future.
+If you encounter any issues, reach out to us on [Discord](https://discord.gg/rzdnErX) and we will make sure to help you out!
+:::
+
+# Web App
```
infomiho, line 13:
```suggestion
# Testing your React app
```

shayneczyzewski:
> When I exit the test watcher with `CTRL + C` it outputs an error message which I didn't expect because I wanted to stop it.

I noticed the same. It seems to show an error when you had any test fail. But if they all succeeded it doesn't show that. I am unsure if that is just how vitest watch behaves (return error code if any test ever failed) or a quirk in our exit code handling. 🤷🏻‍♂️ hopefully not too big a deal for v1

web/docs/testing.md
```diff
@@ -0,0 +1,104 @@
+---
```
infomiho, line 1: After you merge the main branch, I think you should see a `guides` folder, I would suggest we keep all the guides there. I didn't migrate anything yet there, but we can start putting new things there.

web/docs/testing.md
```diff
@@ -0,0 +1,104 @@
...
+import AddExternalAuthEnvVarsReminder from './_addExternalAuthEnvVarsReminder.md'
+
+:::info
+Wasp is in beta, so keep in mind there might be some kinks / bugs, and possibly some changes with testing support in the future.
+If you encounter any issues, reach out to us on [Discord](https://discord.gg/rzdnErX) and we will make sure to help you out!
+:::
+
+# Web App
+
+Wasp enables you to quickly and easily write both unit tests and React component tests for your frontend code. Becase we already use [Vite](https://vitejs.dev/), we were able to add many additional testing capabilities by giving you an integrated [Vitest](https://vitest.dev/) experience out of the box.
```
infomiho, line 15:
```suggestion
Wasp enables you to quickly and easily write both unit tests and React component tests for your frontend code. Because we already use [Vite](https://vitejs.dev/), we also support testing the web app by giving you an integrated [Vitest](https://vitest.dev/) experience out of the box.
```

web/docs/testing.md
```diff
@@ -0,0 +1,104 @@
...
+})
+```
+
+#### React Testing Helpers
+
+Wasp provides two React testing helpers:
+- `mockQuery`: Takes a Wasp Query (or an object of type `{ route: { method: string; path: string; } }`) to mock and the JSON data to return. This is helpful if your Query uses `useQuery`. Behind the scenes, this uses `msw` to create a server request handler that responds with the provided JSON to an HTTP request for the operation's endpoint. Request handlers are cleared after each test.
+- `renderInContext`: Takes a React component, wraps it inside a `QueryClientProvider` and `Router`, and renders it.
+
+# Server
```
infomiho, line 102:
```suggestion
# Testing your server-side code
```

web/docs/testing.md
```diff
@@ -0,0 +1,104 @@
...
+
+  screen.debug()
+})
+```
+
+#### React Testing Helpers
+
+Wasp provides two React testing helpers:
+- `mockQuery`: Takes a Wasp Query (or an object of type `{ route: { method: string; path: string; } }`) to mock and the JSON data to return. This is helpful if your Query uses `useQuery`. Behind the scenes, this uses `msw` to create a server request handler that responds with the provided JSON to an HTTP request for the operation's endpoint. Request handlers are cleared after each test.
+- `renderInContext`: Takes a React component, wraps it inside a `QueryClientProvider` and `Router`, and renders it.
```
infomiho, line 100: This makes it a bit more readable for me:
```suggestion
- `mockQuery`: Takes a Wasp Query to mock and the JSON data to return.
  - This is helpful if your Query uses `useQuery`.
  - Behind the scenes, this uses `msw` to create a server request handler that responds with the provided JSON to an HTTP request for the operation's endpoint.
  - Request handlers are cleared after each test.
- `renderInContext`: Takes a React component, wraps it inside a `QueryClientProvider` and `Router`, and renders it.
```

shayneczyzewski:
> > When I exit the test watcher with `CTRL + C` it outputs an error message which I didn't expect because I wanted to stop it.
> 
> I noticed the same. It seems to show an error when you had any test fail. But if they all succeeded it doesn't show that. I am unsure if that is just how vitest watch behaves (return error code if any test ever failed) or a quirk in our exit code handling. 🤷🏻‍♂️ hopefully not too big a deal for v1

It is actually a behavior of vitest @infomiho. If you are in watch mode and any test run fails, it will report an exit code of 1 even if it ends with a successful test run. Just tried in terminal.

Screenshot 2023-03-27 at 11 49 07 AM

waspc/cli/src/Wasp/Cli/Command/Test.hs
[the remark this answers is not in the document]
shayneczyzewski: Yeah I just checked everywhere and they don't have punctuation, so removing the period in this one :)

waspc/data/Generator/templates/react-app/src/queries/index.d.ts
```diff
@@ -1,6 +1,12 @@
 import { UseQueryResult } from "@tanstack/react-query";
 
-export type Query<Input, Output> = (args: Input) => Promise<Output>
+import { type HttpMethod } from "../types";
+
+export type Query<Input, Output> = {
+    (args: Input): Promise<Output>
+    queryCacheKey: string[]
+    route: { method: HttpMethod, path: string }
+  }
```
infomiho, line 9: Small nit:

```suggestion
}
```

waspc/data/Generator/templates/react-app/src/test/vitest/helpers.tsx
```diff
@@ -0,0 +1,72 @@
...
+        server.use(rest.delete(url, responseHandler))
+      },
+    }
+
+    const setupMock = handlers[route.method]
+    if (!setupMock) {
+      throw new Error(`Unsupported query method for mocking: ${route.method}. Supported method strings are: ${Object.keys(handlers).join(', ')}.`)
+    }
+
+    setupMock()
```
sodic, line 68: You could simplify this block to:
```typescript
const handlers: Record<HttpMethod, Parameters<typeof server.use>[0]> = {
  GET: rest.get(url, responseHandler),
  POST: rest.post(url, responseHandler),
  PUT: rest.put(url, responseHandler),
  DELETE: rest.delete(url, responseHandler),
}

server.use(handlers[route.method])
```

## What about checking for the wrong method?
I've removed the part where you check for whether the correct method was specified because TypeScript takes care of it.

I assume you wanted to make sure to return an informative message even when the caller isn't using TypeScript. If that's the case, it's definitely better to keep the check.  However, we should perform it directly and as soon as possible (instead of relying on a map lookup to fail).

To perform this check without having to define a handler mapping first, you'll need a way to access the allowed methods in **the value space** (the type `HttpMethod` only introduces them to the type space).

This is where enums really shine (we've already used them [here](https://github.com/wasp-lang/wasp/pull/961#discussion_r1086804020)):
```typescript
enum HttpMethod {
    Get = 'GET',
    Post = 'POST',
    Put = 'PUT',
    Delete = 'DELETE',
}

// ...

function mockQuery({ route }: { route: QueryRoute }, resJson: any): void {
  if(!Object.values(HttpMethod).includes(route.method)) {
    // throw error
  }

  // continue knowing that route.method is a real method
  // ...

  const handlers: Record<HttpMethod, Parameters<typeof server.use>[0]> = {
    [HttpMethod.Get]: rest.get(url, responseHandler),
    [HttpMethod.Post]: rest.post(url, responseHandler),
    [HttpMethod.Put]: rest.put(url, responseHandler),
    // You can still use the string directly if you like
    // but the above method is easier to auto-refactor
    DELETE: rest.delete(url, responseHandler),
  }

  server.use(handlers[route.method])
}
```
> **Note**
> Enums in TypeScript are pretty weird. [This StackOverflow answer](https://stackoverflow.com/a/50396312) explains them nicely. If you didn't need runtime checks, a union type (like the one Miho recommended and you have now) would be a better choice.

waspc/data/Generator/templates/react-app/src/test/vitest/helpers.tsx
sodic: If we have a link that points to the docs where they recommend this, I suggest adding it to the code.

waspc/data/Generator/templates/react-app/src/test/vitest/helpers.tsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react'
```
sodic, line 1: You can do a named import of `ReactElement` to avoid dragging in the entire `React` namespace.

waspc/data/Generator/templates/react-app/src/test/vitest/helpers.tsx
```diff
@@ -0,0 +1,72 @@
...
+  beforeAll(() => server.listen())
+  afterEach(() => {
+    server.resetHandlers()
+    cleanup()
+  })
+  afterAll(() => server.close())
+
+  function mockQuery({ route }: { route: QueryRoute }, resJson: any): void {
+    const url = `${config.apiUrl}${route.path}`
+    const responseHandler = (_req, res, ctx) => {
```
sodic, line 41: We're not using `strict: true`, in our TS config because we're trying to match the TS config we generate for our users.

Still, for best development experience, I recommend setting it to `true` while working. It's going to help you catch many additional type errors. 

No type errors here, btw. I just noticed it was off thanks to the implicit `any`. If the function was longer, it would be a good idea to do something like this and get proper type support inside:
```typescript
const responseHandler: ResponseResolver<any, RestContext, any> = (_req, res, ctx) => {
  return res(ctx.json(resJson))
}
```

waspc/data/Generator/templates/react-app/src/test/vitest/helpers.tsx
```diff
@@ -0,0 +1,72 @@
...
+      rerender(
+        <QueryClientProvider client={client}><Router>{rerenderUi}</Router></QueryClientProvider>
+      )
+  }
+}
+
+type QueryRoute = Query<any, any>['route']
+type Handler = () => void
+
+export function mockServer() {
```
sodic, line 29: Mind the module boundaries :)
https://github.com/wasp-lang/vscode-wasp/pull/6#discussion_r970774084

sodic (commented):
Great work on getting this out so quickly!

Left a lot of comments, but none of them are related to the PR's functionality (just code style and organization). Only a couple of those are important to address (maybe just one lol). Anyway, you'll know what to do.

waspc/data/Generator/templates/react-app/src/test/vitest/helpers.tsx
shayneczyzewski: It was mostly random people on SO or in GH discussions. I can at least point to that code as a reference in the comment. 👍🏻

waspc/cli/src/Wasp/Cli/Command/Test.hs
```diff
@@ -0,0 +1,48 @@
...
+  watchOrStartResult <- liftIO $ do
+    ongoingCompilationResultMVar <- newMVar (warnings, [])
+    let watchWaspProjectSource = watch waspRoot outDir ongoingCompilationResultMVar
+    watchWaspProjectSource `race` testRunner outDir
+
+  case watchOrStartResult of
+    Left () -> error "This should never happen, listening for file changes should never end but it did."
+    Right startResult -> case startResult of
+      Left testError -> throwError $ CommandError "Test failed" testError
+      Right () -> return ()
```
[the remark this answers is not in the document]
shayneczyzewski: Good observation. They do have a few lines of similarity, but I'm not entirely sure what the abstraction would look like or do here for you, or even what common lines we could remove. I will give it more thought though, since it does feel like we may be able to do more here. I wonder if `watch` itself is the command that needs some work. 🤔

web/docs/guides/testing.md
```diff
@@ -0,0 +1,107 @@
...
+  screen.debug()
+})
+```
+
+#### React Testing Helpers
+
+Wasp provides two React testing helpers:
+- `mockQuery`: Takes a Wasp Query to mock and the JSON data to return.
+  - This is helpful if your Query uses `useQuery`.
+  - Behind the scenes, this uses `msw` to create a server request handler that responds with the provided JSON to an HTTP request for the operation's endpoint.
```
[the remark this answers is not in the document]
shayneczyzewski: We do above it in the dropdown, but I can add another link here as well 👍🏻

waspc/data/Generator/templates/react-app/src/test/vitest/helpers.tsx
shayneczyzewski: Great tips here @sodic, thanks, and using it in deploy is definitely coming back to me! :D Eventually it will become natural haha 🙏🏻

waspc/data/Generator/templates/react-app/src/test/index.ts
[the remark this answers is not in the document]
shayneczyzewski: Yeah, I think for right now it is probably easiest to keep it under src so that:
- all the helpers on the Haskell side continue to work since they are mostly based off src, and
- we don't have to change the "@wasp" resolution process for imports

I think its technically possible to pull it out, but it would be quite a bit of work we may want to punt on until/unless this becomes a problem, or we do a larger organizational refactor.

waspc/data/Generator/templates/react-app/src/types.ts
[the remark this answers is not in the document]
sodic: We decided to keep it in the react-app folder for now.

web/docs/guides/testing.md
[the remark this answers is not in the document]
sodic: Oh, didn't see that one. You didn't need to add it then 😄 , but it can stay.

web/docs/guides/testing.md
```diff
@@ -0,0 +1,107 @@
...
+In fact, anything after `wasp test client` gets passed to Vitest directly, so check out [their CLI docs](https://vitest.dev/guide/cli.html) for more.
+
+:::warning
+You should not run `wasp test` while running `wasp start`, as both will attempt to compile and write your project to `.wasp/out`.
+:::
+
+## Examples
+### Unit Tests
+
+```ts title=src/client/Todo.test.tsx
```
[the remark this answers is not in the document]
sodic: It's still .tsx on the file system, but no biggie.

waspc/data/Generator/templates/react-app/src/operations/index.ts
```diff
@@ -0,0 +1,21 @@
+import api, { handleApiError } from '../api'
+import { HttpMethod } from '../types'
+
+export type OperationRoute = { method: HttpMethod, path: string }
+
+export async function callOperation(operationRoute: OperationRoute, args: any) {
+  if (operationRoute.method !== HttpMethod.Post) {
+    throw new Error(`Unsupported operation method: ${operationRoute.method}. Only POST is supported.`)
+  }
```
sodic, line 9: You can catch this in the type system (especially if it's an internal function). Just change the argument type to `operationRoute: OperationRoute & { method: 'POST' }`.

shayneczyzewski: True, I will update the type, thnx! I mainly kept the runtime checks since the calling functions (our own) are still JS.

web/docs/guides/testing.md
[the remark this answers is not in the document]
shayneczyzewski: Don't they need a `.tsx` extension if they use JSX? Helpers has `<QueryClientProvider` and Todo.test.tsx has `<App><Todo /></App>` inside.

PR #1058, opened 2023-03-15
Fixes localstorage clearing bug

### LINKED ISSUE #1042: Logout clears the `localStorage` thus affecting user defined keys
When we perform the [logout action](https://github.com/wasp-lang/wasp/blob/main/waspc/data/Generator/templates/react-app/src/auth/logout.js#LL5) it clears [everything](https://github.com/wasp-lang/wasp/blob/main/waspc/data/Generator/templates/react-app/src/api.js#L31) in `localStorage`, not only things related to auth, which is affecting the user's ability to keep things related to his app in `localStorage` 

### Ideally
We would only remove specific `localStorage` keys we set there. 

```js
const authKeys = [WASP_APP_AUTH_TOKEN_NAME];

authKeys.forEach((key) => localStorage.removeItem(key));
```

infomiho:
Instead of clearing everything, clear only the keys we set.

Fixes #1042


waspc/data/Generator/templates/react-app/src/api.js
```diff
@@ -7,6 +7,10 @@ const api = axios.create({
 
 const WASP_APP_AUTH_TOKEN_NAME = "authToken"
 
+const storageKeys = [
```
Martinsos, line 10: What is a bit tricky is that if we modify this code to have some more keys, a dev needs to remember to list them here, and they could forget to do so.

Small improvement would be to add a comment here that explains what this list is and why it is ipmortant to keep it up to date.

A bit better improvement could be to add some kind of abstraction that takes care of this, I am not sure what that could look like though. Maybe it allows setting only those keys that are defined here.

One option could be to also scope all wasp stuff under one big key, and then we always delete everything under that key. I am not sure if that is a bit weird / bad though, feels like it could use a bit of scrutiny. It is at the end also some kind of abstraction.

waspc/data/Generator/templates/react-app/src/api.js
```diff
@@ -25,10 +29,16 @@ export const clearAuthToken = () => {
   window.localStorage && window.localStorage.removeItem(WASP_APP_AUTH_TOKEN_NAME)
 }
 
-export const clearLocalStorage = () => {
+export const cleanupStorage = () => {
```
Martinsos, line 32: So ok, this is now not any local storage, this is instead storage for the API thingy right? Ok hm! Maybe we can name it even more descriptive in that case, smth like cleanupApiCache ? Surprising that we had such bad abstraction before hm, I think I did that, pretty bad!

waspc/data/Generator/templates/react-app/src/auth/logout.js
```diff
@@ -1,8 +1,8 @@
-import { clearLocalStorage } from '../api.js'
+import { cleanupStorage } from '../api.js'
 import { invalidateAndRemoveQueries } from '../operations/resources'
 
 export default async function logout() {
-  clearLocalStorage()
+  cleanupStorage()
```
Martinsos, line 5: What do we really want to do here?
We want to clear any of the local storage / cache connected to the user? But also connected to the Wasp? What would be the full sentence that very precisely defines what we are doing here? I think this is important to nail down, and then we can make sure that function we call here reflects that and is a correct abstraction for it.
So let's nail down the intention here first, and then based on that we might even want to take that functino out of api.js -> that is what it looks like to me right now.

Martinsos (changes requested):

waspc/data/Generator/templates/react-app/src/api.js
infomiho: I've updated the code to have an abstraction which introduces "key prefix" for each thing put in the `localStorage`.

We put stuff under `wasp:{key}` and remove only keys that start with `wasp:`

waspc/data/Generator/templates/react-app/src/auth/logout.js
infomiho: I've renamed it to be `removeLocalUserData`

Martinsos:
@infomiho actually I think this is a bit more complicated than we thought!

So what to we want to do? As you stated now, we want to `removeLocalUserData`: remove local user data. Yes, we are logging out user, so we want to remove all their data.

So what is wrong with cleaning up the whole local storage? Well, we concluded that we might delete a bit too much stuff that way. What stuff? I am not sure actually -> I guess stuff that is not specific to the user, but that is specific to the global app stuff? Maybe some public data, that is not private to the user? @infomiho is this correct, could you give me an example?

The problem is, I don't see how we can know what belongs to user and what not. Yes, we can prefix stuff that Wasp puts in with `wasp` and delete that, that we know belongs to user, but what if user put some extra stuff into local storage? What if developer using Wasp put some extra stuff into local storage, some private user info, but they put it manually, and it doesn't get deleted with logout? Maybe some other token they obtained from somewhere?
That sounds like we have an issue then.

What can we do about this then?
1. Be aggressive as we have been so far and delete all local data. We might delete too much, but ok at least nothing sensitive can remain. I wonder now if we should also clear other stuff, like cookies or something (can that even be cleaned like that?).
2. Explicitly have user define what gets deleted -> they have to provide their own function, which forces them to think about it and be explicit about it.
3. Something in between -> we could by default be aggressive, but let them override it and define their own handle for it.

I think first step is better understanding the use cases we have here, especially the one you encountered, which made you open this issue at the end, and then also taking into account that we want to make sure not sensitive stuff remains, be it Wasp that put it or developer using Wasp, and then work with that.

infomiho:
### Example

> So what is wrong with cleaning up the whole local storage? Well, we concluded that we might delete a bit too much stuff that way. What stuff? I am not sure actually -> I guess stuff that is not specific to the user, but that is specific to the global app stuff? Maybe some public data, that is not private to the user? @infomiho is this correct, could you give me an example?

I'll provide you with an example I encountered from building Phrase Tutor. When users land on the page for the first time, I display an intro banner. I don't want to display it each time they visit. I use `localStorage` to set a flag `isDisplayed=true` to avoid showing it again.

With the current Wasp implementation we get this:

https://user-images.githubusercontent.com/2223680/225578723-028abbf3-e9e0-456c-bc0a-b021d882e22b.mov

(after logging out, I get the banner again!)

### Data ownership

> I guess that here I am questioning the very idea of the issue this is fixing: is it ok to delete only some user data on logout? How do we know what we need to delete on logout? What if we don't delete enough, and sensitive data remains?

I would think about the problem like this: localStorage is a useful mechanism that developers like to use. Wasp uses it to achieve a part of its functionality i.e. storing the auth token. Developers using `localStorage` for their needs usually have full control over it. 

Wasp interfering with that (clearing it) makes Wasp a no-go for some developers, since they have no way of overriding the logout behaviour (other than forking Wasp or writing their own custom `logout` action). 

If we think about it in terms of data ownership, I think the proposed solution makes sense. Wasp owns the auth, therefore it owns the data and it is responsible to clear it on log out. On the other hand, if the developer puts something else in `localStorage`, they are responsible for it now. If it's something private, they need to take care of it on logout.

### Outro
In the future, if we want to offer a managed storage abstraction which the developer can use to mark some data as sensitive and then we can clear it as well. That would make sense to me. But I think it's outside of the scope of this bug fix (for me it's a bug currently since it prevented me from using `localStorage` reliably)

waspc/data/Generator/templates/react-app/src/api.js
sodic: > Surprising that we had such bad abstraction before hm, I think I did that, pretty bad

image

sodic (commented):
> I would think about the problem like this: localStorage is a useful mechanism that developers like to use. Wasp uses it to achieve a part of its functionality i.e. storing the auth token. Developers using localStorage for their needs usually have full control over it.
>
> Wasp interfering with that (clearing it) makes Wasp a no-go for some developers, since they have no way of overriding the logout behaviour [...]
>
> Wasp owns the auth, therefore it owns the data and it is responsible to clear it on log out. On the other hand, if the developer puts something else in `localStorage`, they are responsible for it now.

I think this is spot on - we wouldn't be dropping databases and clearing the RAM just because Wasp uses a part of it.

> In the future, if we want to offer a managed storage abstraction which the developer can use to mark some data as sensitive and then we can clear it as well. That would make sense to me. But I think it's outside of the scope of this bug fix (for me it's a bug currently since it prevented me from using localStorage reliably)

And I was going to recommend this as an end-solution. An abstraction that offers the methods `set` and `setPublic`. There are libraries that take care of this, I think.

Martinsos (approved):
Ok, looking good enough to me, I will let @sodic take it over from here!

waspc/data/Generator/templates/react-app/src/storage.ts
```diff
@@ -0,0 +1,44 @@
+export type DataStore = {
+    set(key: string, value: unknown): void;
+    get(key: string): unknown | undefined;
```
sodic, line 3: You don't need the `| undefined` part because `unknown` is a set of all possible values: `unknown | undefined` is the same thing as `unknown`.

waspc/data/Generator/templates/react-app/src/storage.ts
```diff
@@ -0,0 +1,44 @@
...
+    };
+
+    function ensureLocalStorageIsAvailable(): void {
+        if (!window.localStorage) {
+            throw new Error('Local storage is not available.');
+        }
+    }
+}
+
+export const storage = createLocalStorageDataStore('wasp');
```
sodic, line 44: **Nit:** Empty line missing.

waspc/data/Generator/templates/react-app/src/storage.ts
```diff
@@ -0,0 +1,44 @@
...
+            ensureLocalStorageIsAvailable();
+            Object.keys(localStorage).forEach((key) => {
+                if (key.startsWith(prefix)) {
+                    localStorage.removeItem(key);
+                }
+            });
+        },
+    };
+
+    function ensureLocalStorageIsAvailable(): void {
```
sodic, line 37: I know why you put the return statement before the function definition (public stuff first, implementation details later), but this could easily confuse people not familiar with JS's quirks.

Therefore, I recommend placing the `ensureLocalStorageIsAvailable` before the return statement, or even moving it outside of `createLocalStorageDataStore` altogether (I also realize that you put it inside to localize all the `localStorage`-specific stuff into a single function, but it's not a big deal to move it outside IMO).

waspc/data/Generator/templates/react-app/src/storage.ts
```diff
@@ -0,0 +1,54 @@
...
+            if (!isLocalStorageAvailable()) {
+                return;
+            }
+            Object.keys(localStorage).forEach((key) => {
+                if (key.startsWith(prefix)) {
+                    localStorage.removeItem(key);
+                }
+            });
+        },
+    };
```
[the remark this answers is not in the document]
sodic: Makes sense, please link the issue when you create it.

waspc/data/Generator/templates/react-app/src/api.ts
```diff
@@ -1,44 +1,44 @@
...
 })
 
-const WASP_APP_AUTH_TOKEN_NAME = "authToken"
+const WASP_APP_AUTH_TOKEN_NAME = 'authToken'
 
-let authToken = null
-if (window.localStorage) {
-  authToken = window.localStorage.getItem(WASP_APP_AUTH_TOKEN_NAME)
-}
+let authToken: string | undefined = storage.get(WASP_APP_AUTH_TOKEN_NAME) as string | undefined
```
sodic, line 11: You can drop `: string | undefined` if you want, it's inferred anyway.

waspc/data/Generator/templates/react-app/src/api.ts
```diff
@@ -1,44 +1,44 @@
...
+      `Token must be a string, but it was: {${typeof token}} ${token}.`
+    )
   }
   authToken = token
-  window.localStorage && window.localStorage.setItem(WASP_APP_AUTH_TOKEN_NAME, token)
+  storage.set(WASP_APP_AUTH_TOKEN_NAME, token)
 }
 
-export const clearAuthToken = () => {
+export const clearAuthToken = (): void => {
```
sodic, line 23: **Nit:** If you feel like it, you can change these guys to regular function (reasoning: #487). If not, just ignore the comment.

waspc/data/Generator/templates/react-app/src/api.ts
```diff
@@ -59,9 +59,12 @@ export const handleApiError = (error) => {
...
     //   HttpError is also used on server to throw errors like these.
     //   That would require copying HttpError code to web-app also and using it here.
-    const responseJson = error.response?.data
+    const responseJson = error.response?.data as {
+      message?: string
+      data?: unknown
+    }
     const responseStatusCode = error.response.status
-    const e = new Error(responseJson?.message || error.message)
+    const e: WaspError = new Error(responseJson?.message || error.message)
```
sodic, line 67: When creating custom errors, it's better to extend the `Error` class and use it like this:
```typescript
throw new WaspError(responseJson?.message ?? error.message, responseStatusCode, responseJson?.data)
```

sodic (commented):
Added several new TS-related comments. 

Thanks for having the patience and tolerating us turning this simple bug fix PR into a refactor :upside_down_face:

waspc/data/Generator/templates/react-app/src/api.ts
```diff
@@ -1,44 +1,44 @@
...
-}
+let authToken: string | undefined = storage.get(WASP_APP_AUTH_TOKEN_NAME) as string | undefined
 
-export const setAuthToken = (token) => {
+export const setAuthToken = (token: unknown): void => {
   if (typeof token !== 'string') {
-    throw Error(`Token must be a string, but it was: {${typeof token}} ${token}.`)
+    throw Error(
+      `Token must be a string, but it was: {${typeof token}} ${token}.`
+    )
```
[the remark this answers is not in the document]
infomiho: You are right, we don't need to go this defensive. We are setting it ourselves and we are setting it inside of a function that already receives a `string`. I'll simplify this code.

waspc/data/Generator/templates/react-app/src/api.ts
infomiho: Good suggestion, I didn't do it since I didn't want to touch too many lines, but while we are at it, let's do it properly.

waspc/data/Generator/templates/react-app/src/storage.ts
```diff
@@ -0,0 +1,44 @@
...
+            ensureLocalStorageIsAvailable();
+            localStorage.removeItem(`${prefix}:${key}`);
+        },
+        clear() {
+            ensureLocalStorageIsAvailable();
+            Object.keys(localStorage).forEach((key) => {
+                if (key.startsWith(prefix)) {
+                    localStorage.removeItem(key);
+                }
+            });
```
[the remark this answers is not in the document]
infomiho: I'll take a look at the pattern, thanks for the info!

waspc/data/Generator/templates/react-app/src/api.ts
```diff
@@ -59,9 +59,12 @@ export const handleApiError = (error) => {
     // TODO: We might want to use HttpError here instead of just Error, since
     //   HttpError is also used on server to throw errors like these.
     //   That would require copying HttpError code to web-app also and using it here.
-    const responseJson = error.response?.data
+    const responseJson = error.response?.data as {
+      message?: string
+      data?: unknown
+    }
```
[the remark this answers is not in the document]
infomiho: Ah, that's cool. Thanks for the pro tip.

sodic (approved):
Looks good, nice work!

waspc/data/Generator/templates/react-app/src/storage.ts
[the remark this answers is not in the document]
infomiho: Created the issue: https://github.com/wasp-lang/wasp/issues/1082

Issue #891, opened 2022-12-20
In Wasp, offer support for sending emails + using it for sending reset password emails


Martinsos:

infomiho:
Closed via https://github.com/wasp-lang/wasp/pull/1087

Issue #691, opened 2022-08-11
How best to add emails for use in Auth?


shayneczyzewski:
We currently lack the ability to do email verification and password reset for the `EmailAndPassword` Auth method. This RFP-lite proposes a direction to solve this.

We could leverage [Nodemailer](https://nodemailer.com/about/) for the underlying ability to send emails using different email providers (e.g., Sendgrid). The assumption when email is used is this would always be an external concern. Wasp would not attempt to spin up any local SMTP server when running locally, due to the complexity and the fact that most emails would not actually likely be delivered anyways due to spam policies.

We could also use [nodemailer-express-handlebars](https://www.npmjs.com/package/nodemailer-express-handlebars) to handle the email templating.

Users could enable email-related functionality in Wasp like so (using the Google AuthN PR updated syntax):
```css
app todoApp {
  title: "ToDo App",
  auth: {
    userEntity: User,
    methods: {
      emailAndPassword: {
        emailValidation: {
          emailTemplate: ValidateEmail,
          linkExpiration: "10m" // some Simple Duration string
        },
        passwordReset: {
          // just use all defaults, but can override any like above
        }
      }
    },
    onAuthFailedRedirectTo: "/login"
  },
  email: {
    templatesDir: "@ext/email/templates",
    configFn: import { config } from "@ext/email/config.js",
    deliveryJob: Foo // optional, if they want to send via a job instead of async call
  }
}

emailTemplate ValidateEmail {
  from: "blah@mydomain.com",
  subject: "Please validate your email",
  textTemplateName: "validation.txt.handlebars", // assumed under `app.email.templatesDir`
  htmlTempateName: "validation.html.handlebars"
}
```

```js
export function config() {
  return {
    perform_deliveries: process.env.NODE_ENV === 'production', // optional, if false show in console instead
    smtp_settings: { // required
      host: process.env.SMTP_HOST,
      port: process.env.SMTP_PORT,
      ...
    }
  }
}
```

```html
<!-- Provided a context containing { user, magicLinkUrl } -->
<!DOCTYPE html>
<html>
<head>
  <meta charset="utf-8">
  <meta http-equiv="x-ua-compatible" content="ie=edge">
  <title>Please Validate Email</title>
</head>
<body>
  <h2>Hello {{user.name}}!</h2>
  <p>Please click <a href="{{{magicLinkUrl}}}">here</a> to validate your email. Thanks!</p>
</body>
</html>
```

### Signup

[...]

PR #1087, opened 2023-03-28
Adding e-mail auth (e-mail verfication, password reset)

infomiho:
### E-mail auth

Adds e-mail & password auth method:
- e-mail field validation
- support for e-mail verification (`isEmailVerified`) using JWT token
- support for password reset using JWT token
- sending of e-mails for e-mail verification and password reset
- templating of e-mails
- redirect after e-mail is verified
- route to open when doing password reset

### TODO

We still need to figure out the following:
- [x] can we allow using username & password and e-mail & password at the same time?
  - [x] how can we make that work?
- [x] validate that a e-mail sender is defined if the e-mail & password provider is used  

Closes:
- #891
- #691


waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
```diff
@@ -0,0 +1,31 @@
...
+    fromField: EmailFromField;
+    getVerificationEmailContent: GetVerificationEmailContentFn;
+}) {
+    return handleRejection(async (req, res) => {
+        const userFields = req.body || {};
+    
+        const user = await createUser(userFields);
+    
+        const verifyEmailToken = await createEmailVerificationToken(user);
+        const verificationLink = `${waspServerConfig.backendUrl}/auth/email/verify-email?token=${verifyEmailToken}`;
```
infomiho, line 22: We need to now know about the backend URL as well and the user needs to provide it through the env variables.

waspc/examples/todoApp/src/server/auth/email.ts
```diff
@@ -0,0 +1,26 @@
+import {
+  GetPasswordResetEmailContentFn,
+  GetVerificationEmailContentFn,
+} from '@wasp/types'
+
+export const getPasswordResetEmailContent: GetPasswordResetEmailContentFn = ({
```
infomiho, line 6: User defined e-mail templates

waspc/examples/todoApp/migrations/20230328100125_add_email_verified/migration.sql
```diff
@@ -0,0 +1,8 @@
+/*
+  Warnings:
+
+  - Added the required column `isEmailVerified` to the `User` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- AlterTable
+ALTER TABLE "User" ADD COLUMN     "isEmailVerified" BOOLEAN NOT NULL;
```
shayneczyzewski, line 8: Do we also want to track other things, like number of emails sent, or last email send time, etc. so we don't allow users to spam emails via our apps?

shayneczyzewski: Also, maybe we should put this on a different object? Unsure, but that would avoid users having to add lots of fields to user if we just use an FK.

waspc/examples/todoApp/src/server/auth/email.ts
shayneczyzewski: In the future maybe we can use some sort of templating for writing the text/html strings more ergonomically. Maybe something simple like loading Handlebars/Mustache templats, or maybe even rendering React components to strings (https://legacy.reactjs.org/docs/react-dom-server.html)?

waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
```diff
@@ -0,0 +1,31 @@
...
+    getVerificationEmailContent,
+}: {
+    emailSender: EmailSender;
+    fromField: EmailFromField;
+    getVerificationEmailContent: GetVerificationEmailContentFn;
+}) {
+    return handleRejection(async (req, res) => {
+        const userFields = req.body || {};
+    
+        const user = await createUser(userFields);
```
shayneczyzewski, line 19: I worry a bit about using the real user model here with pending/to be verified accounts. I think there may be flows that are not desireable (beyond the ability to just spam the system and clutter the user table).

waspc/data/Generator/templates/server/src/auth/providers/email/resetPassword.ts
```diff
@@ -0,0 +1,19 @@
+import { handleRejection } from "../../../utils.js";
+import { findUserBy, updateUserPassword, verifyToken } from "../../utils.js";
+
+export const resetPassword = handleRejection(async (req, res) => {
+    const args = req.body || {};
+    const { token, newPassword } = args;
+    try {
+        const { id: userId } = await verifyToken(token);
```
shayneczyzewski, line 8: How long are tokens valid for? Should we make it shorter than our default expiration?

waspc/data/Generator/templates/server/src/auth/providers/email/login.ts
```diff
@@ -0,0 +1,20 @@
+import { verifyPassword } from "../../../core/auth.js";
+import { handleRejection } from "../../../utils.js";
+import { findUserBy, createAuthToken } from "../../utils.js";
+
+export const login = handleRejection(async (req, res) => {
+    const args = req.body || {};
+    const user = await findUserBy<'email'>({ email: args.email });
```
shayneczyzewski, line 7: Don't we also need to make sure they are verified?

waspc/data/Generator/templates/server/src/auth/utils.ts
```diff
@@ -12,3 +18,72 @@ export const authConfig = {
...
+      throw new HttpError(422, 'Validation failed', { message: e.message })
+    } else if (isPrismaError(e)) {
+      throw prismaErrorToHttpError(e)
+    } else {
+      throw new HttpError(500)
+    }
+  }
+}
+
+export async function updateUserPassword(userId: {= userEntityUpper =}Id, password: string): Promise<void> {
```
shayneczyzewski, line 65: Just seeing this update function made me wonder if we provide any methods to update other user fields. We may want to put some restrictions in auth validations on updating the `User.email` since then they could take over any other email after they verified theirs.

shayneczyzewski (commented):
This is cool @infomiho, nice work! I did not yet run it or go through the code in a super detailed way, but rather just made some high-level questions/comments so we can discuss some design considerations as quickly as possible. Happy to hop on a call to chat/brainstorm if helpful. I will start testing and reviewing in more detail tomorrow 👍🏻

waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
infomiho: The users won't see this URL at all since it will redirect to the client URL right after it processes the token it received.

In the RFC we had two options:
1. Load the client page first 🔵
    1. Client page loads with `?token`
    2. It sends a backend request with the token
    3. User is verified
2. Go directly to the backend 🟠
    1. Backend loads with `?token`
    2. Backend redirects to a success or failure page
    
  The second 🟠 one seemed simpler at the time. But now that I implemented it, I see that we need to introduce a mandatory env variable called `BACKEND_URL` to get it working properly. Maybe the first 🔵 one is better for the user after all?

waspc/examples/todoApp/src/server/auth/email.ts
infomiho: I agree 100%, we should consider templating for e-mails in some future iteration. Then this will benefit the users when using e-mail auth. There is thing interesting library for creating e-mails with React: https://react.email/docs/utilities/render

waspc/data/Generator/templates/server/src/auth/providers/email/requestPasswordReset.ts
```diff
@@ -0,0 +1,36 @@
...
+        const args = req.body || {};
+        const user = await findUserBy<'email'>({ email: args.email });
+    
+        if (!user) {
+            return res.status(400).json({ error: 'User not found' });
+        }
+    
+        const token = await createPasswordResetToken(user);
+        const passwordResetLink = `${waspServerConfig.frontendUrl}${clientRoute}?token=${token}`;
+        emailSender.send({
```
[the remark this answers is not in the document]
infomiho: Ideally, we would `await` the result here and do something with the info. If the API call fails, then maybe return a `500` here. But if the API call succeeds, but the e-mail delivery fails then it's a different kind of problem. 

The issue is that various e-mail providers return different kind of "success" messages, and with most you don't get a definite answer other than "it's scheduled for sending successfully".

waspc/examples/todoApp/migrations/20230328075236_add_email/migration.sql
```diff
@@ -0,0 +1,17 @@
...
+  - A unique constraint covering the columns `[email]` on the table `User` will be added. If there are existing duplicate values, this will fail.
+  - Added the required column `email` to the `User` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- DropIndex
+DROP INDEX "User_username_key";
+
+-- AlterTable
+ALTER TABLE "User" DROP COLUMN "username",
+ADD COLUMN     "email" TEXT NOT NULL;
```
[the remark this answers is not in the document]
infomiho: We can do that 👍 and it makes our existing docs valid for e-mail & password auth for the most part. 

When we tackle the "provider data" issue and make it so that every provider has its own piece of the database, then they would split.

waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
infomiho: > using the real user model here with pending/to be verified accounts

I can see a scenario where somebody knows my e-mail and create an account with my e-mail.

1. They create an account with "miho@xyz.com"
2. I get an e-mail (spam)
3. I later on want to create my real account
4. The e-mail is taken and i would need to use "reset password" to get the actual password

That sounds inconvenient, true. But it seems like it would work itself out.  

> beyond the ability to just spam the system and clutter the user table

That's also a problem with username & password auth as well.

waspc/data/Generator/templates/server/src/auth/providers/email/resetPassword.ts
infomiho: I've updated the code to make the tokens valid for 30 minutes.

waspc/data/Generator/templates/server/src/auth/providers/email/login.ts
infomiho: There are two options we considered:
1. Unverified users have access by default and the developer decides within their actions what to do
2. Unverified users should not be allowed to log in at all

The first flow sounds to me like it could be better with providing extra onboarding steps while the user didn't verify their e-mail. I'm not sure which one we want to go with. Maybe we put it as an option in the Wasp file?

waspc/data/Generator/templates/server/src/auth/utils.ts
infomiho: Changing the email should imply that the user must verify the e-mail again. That's true!

Do we want to expose a function "updateUserEmail" for the user to use or somehow do it inside of Prisma middleware? 🤷

waspc/data/Generator/templates/server/src/auth/providers/config/email.ts
```diff
@@ -0,0 +1,77 @@
...
+            fromField,
+            clientRoute: '{= passwordResetClientRoute =}',
+            getPasswordResetEmailContent: _waspGetPasswordResetEmailContent,
+        }));
+        router.post('/reset-password', resetPassword);
+        router.get('/verify-email', getVerifyEmailRoute({
+            onVerifySuccessRedirectTo: '{= onVerifySuccessRedirectTo =}',
+        }));
+
+        return router;
```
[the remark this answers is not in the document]
infomiho: We mount this router when we are going through the provider configs. We use the `createRouter` to get a router which we setup under the `/auth` route. https://github.com/wasp-lang/wasp/blob/6dbeedca60edb55315be68477d67e34ae7b1cb5c/waspc/data/Generator/templates/server/src/auth/providers/index.ts#L24

waspc/data/Generator/templates/server/src/auth/providers/email/resetPassword.ts
infomiho: TODO: make sure we redirect to a user defined route when the token is invalid or expired

waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
infomiho: The thrid option might be: don't consider an e-mail taken until verified.

infomiho: TODO: research how others are doing e-mail verification

waspc/data/Generator/templates/server/src/auth/utils.ts
infomiho: Additional option: provide an option `allowEmailModification` which is by default `false`, and based on that flag we implement Prisma middleware.

infomiho:
@shayneczyzewski ready for another round
- added a rate limiter
- uses `username` instead of `email` as field
- if an existing user with unverified e-mail is detected, they are deleted when a new signup is attempted (Supabase auth does it similarly)

shayneczyzewski (commented):
Hi @infomiho I spent some time playing with it locally yesterday and today, it is looking better with the most recent changes! Nice job!

Here are some stream of conscious notes I took while testing. I will do more on this Monday, while also working on removing the username & pass coupling with oauth. 👍🏻 

## NOTES

1. Clicking "Sign up" didn't provide any info to user about what was happening. It looks like under the covers it was successful though. :thinking:

   ```sh
   Server: Test email (not sent): {
   Server:   from: { email: 'shayne.czyzewski@gmail.com', name: 'ToDO App' },
   Server:   to: 'thelastemail@gmail.com',
   Server:   subject: 'Verify your email',
   Server:   text: 'Click the link below to verify your email: http://localhost:3000/email-verification-?token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6MywiaWF0IjoxNjgwMjAyOTE2LCJleHAiOjE2ODAyMDQ3MTZ9.YskUq4e123cHplCXK2yWLXqNaPkC8piFF0pHRuKJwc8',
   Server:   html: '\n' +
   Server:     '        <p>Click the link below to verify your email</p>\n' +
   Server:     '        <a href="http://localhost:3000/email-verification-?token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6MywiaWF0IjoxNjgwMjAyOTE2LCJleHAiOjE2ODAyMDQ3MTZ9.YskUq4e123cHplCXK2yWLXqNaPkC8piFF0pHRuKJwc8">Verify email</a>\n' +
   Server:     '    '
   Server: }
   Server: POST /auth/email/signup 200 137.200 ms - 16
   Server: POST /auth/email/login 401 1.881 ms - -
   Server: GET /auth/me 401 0.297 ms - -
   ```

   We will probably want to work on the redirects and/or provide a response they can use for a toast message.

   Same with request password reset.

2. We may want to avoid leaking the info that:

      > Error: User with this email already exists. - Details: User with this email already exists.

when you try to sign up with an email in the system. Although I'm not sure what we can say, but this reminds me of some security article I read discouraging that since attackers can then figure out what emails are registered with sites.

3. It appeared like the timestamps about email sending were not being updated in either send or non-send mode.

shayneczyzewski:
Here is what I'm working on to remove the oauth dep with `usernameAndPassword` we will need for this @infomiho: https://github.com/wasp-lang/wasp/pull/1094

waspc/data/Generator/templates/server/src/middleware/rateLimiter.ts
```diff
@@ -0,0 +1,39 @@
+import { Request, Response, NextFunction } from "express";
+import { RateLimiterMemory } from "rate-limiter-flexible";
+
+// There are two limiters
+// - rateLimiter: 10 requests per second
+// - oncePerMinuteLimiter: 2 requests per minute
```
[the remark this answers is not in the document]
infomiho: I've decided to not introduce the generic limiter, but only for the two routes we have for the e-mail auth. I want to better support rate limiting customisation for the general case. 

Now we only have a rate limiter for signup and password reset and just let through everything else like before.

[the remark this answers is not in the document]
infomiho: Issue for rate limiting improvement https://github.com/wasp-lang/wasp/issues/1098

waspc/data/Generator/templates/server/src/core/auth/prismaMiddleware.js
```diff
@@ -59,18 +60,25 @@ export const registerAuthMiddleware = (prismaClient) => {
...
+  { validates: PASSWORD_FIELD, message: 'password must be at least 8 characters', validator: password => password.length >= 8 },
+  { validates: PASSWORD_FIELD, message: 'password must contain a number', validator: password => /\d/.test(password) },
+]
+
+{=# isUsernameAndPasswordAuthEnabled  =}
+userValidations.push({ validates: USERNAME_FIELD, message: 'username must be present', validator: username => !!username })
+{=/ isUsernameAndPasswordAuthEnabled  =}
+{=# isEmailAuthEnabled =}
+userValidations.push({ validates: USERNAME_FIELD, message: 'email must be present', validator: email => !!email })
+userValidations.push({ validates: USERNAME_FIELD, message: 'email must be valid', validator: email => isValidEmail(email) })
```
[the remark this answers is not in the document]
infomiho: Since our UI will probably say "E-mail" and the fact that this is called `username` will be an implementation detail, I'd rather keep the existing message.

waspc/examples/todoApp/migrations/20230328100331_update_default_value/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "User" ALTER COLUMN "isEmailVerified" SET DEFAULT false;
```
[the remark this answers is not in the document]
infomiho: I'll clean up the migration 👍 

Thanks for the heads up

infomiho:
@shayneczyzewski this is ready for another round, changes:

- Compiler forbids usage of `username & password` and `email` auth at the same time
- Model for `email` auth uses now **optional** `email` and `password` fields
- Validation of `email` auth fields happens in the routes vs. at the Prisma middleware
- I've tested the following scenarios locally:
    - Signup with Google then login
    - Signup with email then login (verification required & not)
    - Reset password
    - Signup with Google then reset password
    - Signup with email then login with Google ⚠️ (this doesn't work since we aren't doing anything to link the accounts)
 - Adds timestamp whenever we send the verification and password reset e-mails to limit that to **once per minute**
 - General fixes and error handling improvements
 - I've deleted the migrations in the example app and I propose we keep only **one migration file** to make it easier for development 🤷‍♂️

waspc/data/Generator/templates/server/src/core/auth/prismaMiddleware.js
```diff
@@ -2,7 +2,9 @@
 import { hashPassword } from '../auth.js'
 import AuthError from '../AuthError.js'
 
+{=# isUsernameAndPasswordAuthEnabled  =}
 const USERNAME_FIELD = 'username'
+{=/ isUsernameAndPasswordAuthEnabled  =}
```
shayneczyzewski, line 7: We could have an error if username is on the User entity, but usernameandPassword auth is _not_ enabled (due to use below). Maybe we just don't wrap this?

waspc/examples/todoApp/src/server/auth/google.js
```diff
@@ -1,19 +1,15 @@
-import { generateAvailableUsername } from '@wasp/core/auth.js'
-
 export function config() {
   console.log('Inside user-supplied Google config')
   return {
     clientID: process.env['GOOGLE_CLIENT_ID'],
     clientSecret: process.env['GOOGLE_CLIENT_SECRET'],
-    scope: ['profile'],
+    scope: ['profile', 'email'],
```
shayneczyzewski, line 6: Nice! We will probably want to note this in the docs somewhere as well 👍🏻

waspc/src/Wasp/Analyzer/StdTypeDefinitions.hs
```diff
@@ -32,6 +31,7 @@ makeEnumType ''JobExecutor
 makeDeclType ''Job
 makeEnumType ''HttpMethod
 makeDeclType ''Api
+makeDeclType ''App
```
shayneczyzewski, line 34: Was there a reason to swap this ordering?

waspc/data/Generator/templates/server/src/auth/providers/local/login.ts
```diff
@@ -1,35 +1,25 @@
...
 
-const prisma = new Prisma.PrismaClient()
+import { findUserBy, createAuthToken } from '../../utils.js'
 
 export default handleRejection(async (req, res) => {
   const args = req.body || {}
 
-  // Try to fetch user with the given username.
-  const user = await prisma.{= userEntityLower =}.findUnique({ where: { username: args.username } })
+  const user = await findUserBy<'email'>({ email: args.email })
```
shayneczyzewski, line 10: This appears to break `usernameAndPassword` auth.

shayneczyzewski (commented):
This is looking great @infomiho! I only found one main bug (just `usernameAndPassword` doesn't work due to an email for local in get user). Otherwise, the testing is coming back clean (including just auth working still)! I will do another round of testing later today 👍🏻 Great work 🥳

waspc/data/Generator/templates/react-app/src/auth/email/actions/passwordReset.ts
```diff
@@ -0,0 +1,20 @@
+{{={= =}=}}
+import api, { handleApiError } from '../../../api';
+
+export async function requestPasswordReset(data: { email: string; }): Promise<{ success: boolean }> {
+    try {
+        const response = await api.post('{= requestPasswordResetPath =}', data).catch(handleApiError);
```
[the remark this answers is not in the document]
infomiho: A great _catch_, I'll remove it!

waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
```diff
@@ -0,0 +1,64 @@
...
+    getVerificationEmailContent: GetVerificationEmailContentFn;
+}) {
+    return async function signup(
+        req: Request<{ email: string; password: string; }>,
+        res: Response,
+    ): Promise<Response<{ success: true } | { success: false; message: string }>> {
+        const userFields = req.body;
+        ensureValidEmailAndPassword(userFields);
+        
+        userFields.email = userFields.email.toLowerCase();
```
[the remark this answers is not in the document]
infomiho: I've lowercased everything 👍

waspc/data/Generator/templates/server/src/auth/providers/local/login.ts
infomiho: Should be fixed now 👍

waspc/data/Generator/templates/server/src/core/auth/prismaMiddleware.js
infomiho: Updated to use `isUsernameOnUserEntity` to keep it consistent.

waspc/src/Wasp/Analyzer/StdTypeDefinitions.hs
infomiho: Since it's template Haskell it expected `Route` to be defined before `App` because `App` now uses `Ref Route` for the email auth definition, something @Martinsos helped me with.

waspc/data/Generator/templates/react-app/src/auth/email/actions/passwordReset.ts
```diff
@@ -0,0 +1,20 @@
...
+        return response.data;
+    } catch (e: unknown) {
+        handleApiError(e);
+    }
+}
+
+export async function resetPassword(data: { token: string; password: string; }): Promise<{ success: boolean }> {
+    try {
+        const response = await api.post('{= resetPasswordPath =}', data);
+        return response;
```
shayneczyzewski, line 16: Just spot checking changed files. Does this need a `.data` too?

shayneczyzewski:
Great job @infomiho all the major stuff and testing looks good! I even tried out a few sample apps and they seem to still work just fine. I think the main discussion points are:
- Should we wait to merge this until the Matija Auth UI stuff lands on `main` and can be integrated into this branch? It is probably ok to combine it all in a separate PR like you mentioned, but `main` will be "dirty" until then and we will need to be pretty certain that can get done quickly to make sure `main` is RC-ready again soon.
- I think we can merge before the docs are ready (as long as we get them done by next week).

🤔 any other considerations before this could be merged? Want to run this by anyone else for 👀 too or just use them for final testing?

Nice work on this! 🥳 Functionally and code wise it feels good to go for me 🚀

waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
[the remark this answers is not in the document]
shayneczyzewski: Nice! The` one place we may want to not do so is `data/Generator/templates/server/src/auth/providers/local/login.ts` if the old way did not, just for backward compatibility.

waspc/data/Generator/templates/react-app/src/auth/email/actions/passwordReset.ts
infomiho: Right you are 👍

waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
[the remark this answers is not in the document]
infomiho: I think this is small change and we are pre 1.0, so if you agree, I'll keep the change

shayneczyzewski (approved):
approving now, group agrees as do I we can just merge this in and add docs/auth UI after, followed by final testing 👍🏻 nice work! 🥳

PR #1090, opened 2023-03-30
Full-stack compile-time type safety 

### LINKED ISSUE #976: Automatic client-side support for operation inputs and outputs
Blocked by #975.

If a user specifies the input (arguments) and output (return value) types for an operation, we want to know about it when we import it on the client.

### Idea
Assume we have this on the server
```typescript
import { Task } from '@wasp/entities'
import { GetTasks} from '@wasp/queries'

type Payload = Pick<Task, 'isDone'>;

// `Payload` is the type of args, `Task[]` is the query's return type.
const getTasks: GetTasks<Payload, Task[]> = (args, context) => { 
  // ...
}
```
We want to be able to do this on the client:
```typescript
import getTasks from '@wasp/queries/getTasks.js'

// Type of 'tasks' automatically inferred to 'Task[]'
// Type of payload inferred to 'Payload', typo (done -> isDone) detected.
const { data: tasks, isError, error } = useQuery(getTasks, { done: true })
```

sodic:
Fixes #976.

To keep TypeScript from lying to the user, we had to add a data serialization/transformation layer. I've decided to use `superjson` for the time being (thanks @Zeko369!).

In the future, I hope we can make this transformer configurable (as tRPC does).

### Motivating example
To understand why we had to use Superjson, let's see what would happen if we implemented #976 without it.

Imagine you had the following query:
```typescript
export const getPerson: GetPerson<{ id: number }, { name: string, dob: Date }> = ({ id }, context) => {
  // You do some processing and return a value of type { name: string, dob: Date }.
  // All is well with the type checker.
}
```
Then, you do this on the client:
```typescript
import getPerson from '@wasp/queries/getPerson'

const person = getPerson({ id: 12 })
```
TypeScript will tell you that `person` has type `{ name: string, dob: Date }`. This is probably the type you wanted, but not the one you actually have:
1. Your query returns a value of type `{ name: string, dob: Date }`, so far so good
2. Experss middleware serializes it into a JSON, but JSON serializes `Date`s into strings
3. The payload is sent over the network.
4. Axios deserializes the payload back into an object of type `{ name: string, dob: string }`, but the type checker things you have an object of type `{ name: string, dob: Date }`.
5. Chaos

### Todo list:
  - [x] Ensure proper serialization/deserialization to ensure types match runtime types
  - [x] Clean up the Haskell code and reorganize it
  - [x] Implement full-stack type safety for actions
  - [x] Discuss whether server imports are a problem or not (could be creating technical debt)
  - [ ] Require operation `Input` and `Output` types to extend superjson's serializable type


waspc/src/Wasp/Generator/WebAppGenerator/OperationsGenerator.hs
```diff
@@ -57,20 +65,54 @@ genActions spec =
...
             .= ( ServerGenerator.operationsRouteInRootRouter
                    ++ "/"
                    ++ ServerOperationsRoutesG.operationRouteInOperationsRouter operation
                ),
           "entitiesArray" .= makeJsArrayOfEntityNames operation
         ]
+          ++ operationTypeData operation
     operation = AS.Operation.QueryOp queryName query
 
+operationTypeData :: AS.Operation.Operation -> [Pair]
```
sodic, line 83: I'm going to reconsider where to put this entire function (and make it nicer too).

waspc/src/Wasp/Generator/WebAppGenerator/OperationsGenerator.hs
```diff
@@ -57,37 +64,72 @@ genActions spec =
...
+    serverOperationFileFromWebAppQueriesDir =
+      webAppRootDirFromWebAppQueriesDir </> serverOperationFileFromWebAppRootDir
+
+    operationImportPath =
+      fromJust $
+        SP.parseRelFileP $
+          toViteImportPath $
+            SP.fromRelFileP $
+              fromJust $
+                SP.relFileToPosix serverOperationFileFromWebAppQueriesDir
```
sodic, line 120: Todo: this is ridiculous.

waspc/data/Generator/templates/react-app/src/queries/_query.ts
```diff
@@ -0,0 +1,11 @@
...
+import { createQuery } from './core'
+{=& operationTypeImportStmt =}
+
+
+const query = createQuery<{= operationTypeName =}>(
+  '{= queryRoute =}',
+  {=& entitiesArray =},
+)
+
+export default query
```
sodic, line 11: Although we decided to abandon default exports (see #546 for details), we're still postponing it. It requires many  breaking changes in our API (and the docs).

In the meantime, I separated the export from the definition because it helps with VS Code's "on hover" feature.

When exporting the value directly:
image

When exporting the variable after the assignment:
image

waspc/data/Generator/templates/universal/types.ts
```diff
@@ -0,0 +1,16 @@
...
+  ? (...args: A) => R
+  : T extends infer O
+  ? { [K in keyof O]: O[K] }
+  : never
+
+// TypeScript's native Awaited type exhibits strange behavior in VS Code (see
+// todo for details). Until it's fixed, we're using our own type for this.
+export type _Awaited<T extends Promise<unknown>> = T extends Promise<infer V>
+  ? V
+  : never
```
sodic, line 16: Take a look at [this playground](https://www.typescriptlang.org/play?noImplicitAny=false&strictNullChecks=false&strictFunctionTypes=false&strictPropertyInitialization=false&strictBindCallApply=false&noImplicitThis=false&noImplicitReturns=false&alwaysStrict=false&ts=4.9.5#code/C4TwDgpgBAYlC8UAUBDATgcwM4C4oDsBXAWwCMI0AaKAYwHt9gIAPYPIsigSigQD4oABTR1iASywQAPIXwBrfHQDu+PgChQkKAHEEydNnYQAbhWr1GLNgRLk0PeAOGiJ02QuWq1G8NABKEMCEaPhwiACCSihiTAAmUgFBIQAqvlIwfAIA9FlQWAAWdIQANrFQ5FAA5O6KKpXUisBVKPgglT5aicH4uhFRMRDxXSlp2plQOXmFJWUV1fK1+PUEdE2VLW1AA) and try the same snippet in VS Code. This strange behavior causes a problem for operations with unspecified inputs.

#### Example
Let's say your backend query definition looks like this:
```typescript
export const getTasks: GetTasks = async (_args, context) => {
  // ...
}
```
This is what we want TypeScript to infer on the client:



Unfortunately, if we use the built-in `Awaited` type, this happens:

waspc/data/Generator/templates/react-app/src/test/vitest/helpers.tsx
```diff
@@ -26,10 +26,11 @@ export function renderInContext(ui: ReactElement): RenderResult {
 }
 
 type QueryRoute = Query<any, any>['route']
+type MockQuery = <Input, Output, MockData extends Output>(query: Query<Input, Output>, resJson: MockData) => void
```
sodic, line 29: @shayneczyzewski I've changed this signature to improve type safety. Check [this playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAigrhATiAPASQHZjsANFAeR22AD4oBeKAbwCgoGoAKAQ0QHMBnALikxICUvAAqIA9gFsAlpwgoiwEqVoBfWrQAmEAMYAbNtABmcDNuBSxGKBLHaA1vCSp+OfApL4AsrbtQIAD2AIDA1OQmIcUiYARwRkXkdkdCxXcMVI-EQITgApTkteb3shKAA3MSkNAG51ALAxRGAoUEgoABUWTl8qOgZK3gw4CQAjJHooLU5tRCkwcwKoTmAZjHZxmQARSwheYbExXQgWDHG4WUQ0DQGh0cRVdS09AyhtSyWodghgDq6eWDjUOVKvgfnYANoAXWUj30WRebyawE6dj+oMhmh0sOgrww7zAbHMLF0oL+wik9hQoPwACJKtSoAAfKDUybTWbzDD0pm0zhbDAQamkdG0Gz2RIgJifb7Izj4JG-ASMJUMAD0KuaAAsZIsNWI4LoNM1wNiNTo7PhSkTKtYfCKfOLJV8Sfh8Y0pESSYqoGrNdrOLr9YaMGJEcaXqb7PgpBhLbpraK7A9Mc8ce8pQR+W0AO5iAiINoarI7f5OFBLFbsfAARkZUAATLWAMzQ5Nw1NNbZ57NiXg1pl1jFPNsIqDbbsFot5gBiesQvdrDaZjdrABZ1AmHenMzmu4WIBB8J38zmvcrGD7gFqwv69Qaja1tBHzWUrYaE3axQDHcAMxBu7ui0PbcxAnfdp1nRULyvHVbyDEN7xNM0owwFhSjjN8fCAA) for the demo.

waspc/data/Generator/templates/react-app/src/queries/index.d.ts
```diff
@@ -1,14 +1,10 @@
...
-
 export type Query<Input, Output> = {
-    (args: Input): Promise<Output>
-    queryCacheKey: string[]
-    route: { method: HttpMethod, path: string }
+    (queryCacheKey: string[], args: Input): Promise<Output>
 }
 
-export function useQuery<Input, Output, Error = unknown>(
+export function useQuery<Input, Output>(
```
sodic, line 7: We always throw something that resembles a general `Error`, so let's keep it simple for now.

waspc/data/Generator/templates/react-app/src/queries/core.d.ts
```diff
@@ -1,3 +1,13 @@
...
 
-export function createQuery<Input, Output>(queryRoute: string, entitiesUsed: any[]): Query<Input, Output>
+export function createQuery<BackendQuery extends GenericBackendQuery>(
+  queryRoute: string,
+  entitiesUsed: any[]
+): QueryFor<BackendQuery>
+
+type QueryFor<BackendQuery extends GenericBackendQuery> = Expand<
+  Query<Parameters<BackendQuery>[0], _Awaited<ReturnType<BackendQuery>>>
+>
```
sodic, line 11: Applying `Expand` to this type results in a better DX.

On-hover with `Expand`:

image

On-hover without `Expand`: 

image

waspc/data/Generator/templates/react-app/src/actions/index.ts
```diff
@@ -7,7 +7,10 @@ import {
 } from '@tanstack/react-query'
 import { Query } from '../queries';
 
-export type Action<Input, Output> = (args?: Input) => Promise<Output>;
+export type Action<Input, Output> = 
+  [Input] extends [never] ? 
+  (args?: unknown) => Promise<Output> :
+  (args: Input) => Promise<Output>
```
sodic, line 13: To understand why and what's going on here, we need to establish a couple of things.

### Default input and output types for Operations
Users can specify their operation's `Input` and `Output` types, like this:
```typescript
export const updateTaskIsDone: UpdateTask<Pick<Task, "id">, Task> = async ({ id }, context) => {
  // ...
}
```
We also want to allow them to omit input and output types. Sure, they won't get full-stack type safety, but it helps rapid prototyping:
```typescript
export const updateTaskIsDone: UpdateTask = async (_args, context) => {
  // ...
}
```
The main question is: _What should be the default input and output types?_

Assuming we don't want to use `any` unless absolutely necessary, the correct answer is: `never` for the input, `unknown` for the output. Read https://github.com/wasp-lang/wasp/pull/989#discussion_r1101155001 to understand why.

### Frontend behavior with omitted input types
Assuming the user decides not to specify input and output types:
```typescript
export const updateTaskIsDone: UpdateTask = async (_args, context) => {
  // ...
}
```
We want the client RPC call to either accept nothing or anything. More precisely, all following calls should pass the type checker:
```typescript
updateTaskIsDone() 
updateTaskIsDone(12)
updateTaskIsDone("asdf")
```
To implement such behavior, we need to detect whether an action's `Input` type is `never` and type the argument appropriately:
1. `Input` type is `never` (i.e. the user hasn't specified an input type), the action should take a single optional argument of type `unknown`
2. `Input` type isn't `never` (i.e., the user has specified an input type), the action should simply use the user-specified input type.

### Disabling the distribution of `never`
All that's left to answer is _Why do we wrap the types inside an array?_.

The following type wouldn't work:
```typescript
export type Action<Input, Output> = 
  Input extends never ? 
  (args?: unknown) => Promise<Output> :
  (args: Input) => Promise<Output>

type X = Action<never, number> // X is 'never', not (args?: unknown) => Promise<number>
```
This is because conditional types (in our case, `Action`) distribute over unions (in our case, the empty union `never`). We need a way to prevent the distribution, and that's what the array wrapper is for.

Read https://github.com/microsoft/TypeScript/issues/31751#issuecomment-498526919 for more details.

waspc/data/Generator/templates/react-app/src/actions/_action.ts
sodic, line 1: Mostly the same as former `_action.js`, but Git didn't detect it as a rename.

waspc/data/Generator/templates/react-app/src/auth/useAuth.ts
```diff
@@ -1,12 +1,14 @@
 import { useQuery } from '../queries'
 import api, { handleApiError } from '../api'
 import { HttpMethod } from '../types'
+// todo(filip): turn into a proper import
+import { type SanitizedUser } from '../../../server/src/_types/' 
 
 export default function useAuth(queryFnArgs, config) {
```
sodic, line 7: I've decided not to go with the module boundaries principle (https://github.com/wasp-lang/vscode-wasp/pull/6#discussion_r970774084) to avoid unnecessary verbosity.

This short function that's basically a wrapper to two other (explicitly typed) public functions: `useQuery` and `getMe`. If that changes, we'll make the type signature explicit.

waspc/data/Generator/templates/react-app/src/operations/index.ts
```diff
@@ -1,12 +1,17 @@
...
+ } from 'superjson'
 
 export type OperationRoute = { method: HttpMethod, path: string }
 
 export async function callOperation(operationRoute: OperationRoute & { method: HttpMethod.Post }, args: any) {
   try {
-    const response = await api.post(operationRoute.path, args)
-    return response.data
+    const superjsonArgs = superjsonSerialize(args)
+    const response = await api.post(operationRoute.path, superjsonArgs)
```
sodic, line 13: We're still relying on `axios` to fully serialize a "semi-serialized" superjson object under the hood.

More precisely, Axios will turn `{ json: "...", meta: "..." }` (object) into `'{"json": "...", "meta": "..." }'` (string).

waspc/data/Generator/templates/server/src/actions/_action.ts
```diff
@@ -7,7 +7,7 @@ import prisma from '../dbClient.js'
           consider in the future if it is worth removing this duplication. =}
 
 export default async function (args, context) {
-  return {= jsFn.importIdentifier =}(args, {
+  return {= jsFn.importIdentifier =}(args as never, {
```
sodic, line 10: This is pretty ridiculous, but is the quickest possible way to fix a type issue caused by the requirements described here: https://github.com/wasp-lang/wasp/pull/1090/files#r1159732471.

We'll come up with a more long-term solution eventually.

waspc/data/Generator/templates/server/src/routes/operations/_action.js
```diff
@@ -1,17 +1,22 @@
...
+  deserialize as superjsonDeserialize,
+  serialize as superjsonSerialize,
+} from 'superjson'
 import { handleRejection } from '../../utils.js'
-
 {=& operationImportStmt =}
 
 export default handleRejection(async (req, res) => {
-  const args = req.body || {}
+  const args = (req.body && superjsonDeserialize(req.body)) || {}
```
sodic, line 10: TODO: I should probably be catching something here.

sodic, line 10: Our deserialization process on the server is analogous to the one on the client (https://github.com/wasp-lang/wasp/pull/1090/files#r1159749811).

More precisely:
1. When deserializing, Express will turn `'{"json": "...", "meta": "..." }'` (string) to `{ json: "<serialized_data>", meta: ".<metadata>" }` (object) , and we'll do the rest with `superjsonDeserialize`. The end result is `data`.
2. When serializing, we will turn `data` into `{ json: "<serialized_data>", meta: ".<metadata>" }` (object) using `superjsonSerialize`, and Express will do the rest. The end result is `'{"json": "<serialized_data>", "meta": "<metadata>" }'` (string).

waspc/src/Wasp/Generator/WebAppGenerator/Common.hs
```diff
@@ -97,3 +111,8 @@ mkUniversalTmplFdWithDst relSrcPath relDstPath =
     (webAppRootDirInProjectRootDir </> relDstPath)
     (universalTemplatesDirInTemplatesDir </> relSrcPath)
     Nothing
+
+toViteImportPath :: Path Posix (Rel r) (File f) -> Path Posix (Rel r) (File f)
+toViteImportPath = fromJust . SP.parseRelFileP . dropExtension . SP.fromRelFileP
+  where
+    dropExtension = fst . splitExtension
```
sodic, line 118: This signature differs from the method `toESModulesImport` path we have in `ServerGenerator`.

I don't think it's a big deal, as we'll want to overhaul the entire story wrt these extensions in the near future. I can change the signature to `FilePath -> FilePath`, but then we get this: https://github.com/wasp-lang/wasp/pull/1090#discussion_r1157836550 (which is arguably even worse).

waspc/src/Wasp/Generator/WebAppGenerator/OperationsGenerator.hs
sodic, line 1: This file relies on several symbols imported from `ServerGenerator.Something`, but it's not without precedent (i..e, we've already had a server import). So that's the excuse I'm using :)

Just kidding, I don't have a better idea for structuring this code. If you think this coupling is a problem, suggestions are welcome.

waspc/src/Wasp/Generator/WebAppGenerator/OperationsGenerator.hs
```diff
@@ -20,10 +22,14 @@ import qualified Wasp.AppSpec.Query as AS.Query
 import Wasp.Generator.FileDraft (FileDraft)
 import Wasp.Generator.Monad (Generator)
 import qualified Wasp.Generator.ServerGenerator as ServerGenerator
+import Wasp.Generator.ServerGenerator.Common (serverSrcDirInServerRootDir)
+import Wasp.Generator.ServerGenerator.OperationsG (operationFileInSrcDir)
```
Martinsos, line 26: Maybe import these qualified in a way that makes it clear this is coming from Server. For `serverSrcDirInServerRootDir` it is clear, but for `operationFileInSrcDir` it is not.

waspc/data/Generator/templates/react-app/src/auth/useAuth.ts
```diff
@@ -1,12 +1,14 @@
 import { useQuery } from '../queries'
 import api, { handleApiError } from '../api'
 import { HttpMethod } from '../types'
+// todo(filip): turn into a proper import
+import { type SanitizedUser } from '../../../server/src/_types/'
```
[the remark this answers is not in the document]
infomiho: You don't really _need_ to use the new import style, it only helps you to define a JS import statement in Haskell and then print it in a JS file.

If want to ensure that this import is importing from the proper place (server/src) you can Strong Path it and define a new `JSImport` and then output it in the template. This would move the information about this weird import in Haskell, which would be a win for us IMHO.

Martinsos (approved):

waspc/data/Generator/templates/react-app/src/auth/useAuth.ts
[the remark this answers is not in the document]
infomiho: Nit: remove trailing`/`

waspc/data/Generator/templates/react-app/src/test/vitest/helpers.tsx
shayneczyzewski: Sounds good to me, thanks!

waspc/data/Generator/templates/react-app/src/test/vitest/helpers.tsx
```diff
@@ -40,14 +41,15 @@ export function mockServer(): {
   })
   afterAll(() => server.close())
 
-  function mockQuery({ route }: { route: QueryRoute }, resJson: any): void {
+  const mockQuery : MockQuery = (query, mockData) => {
```
infomiho, line 44: Nit
```suggestion
  const mockQuery: MockQuery = (query, mockData) => {
```

waspc/data/Generator/templates/react-app/src/test/vitest/helpers.tsx
```diff
@@ -65,3 +67,9 @@ export function mockServer(): {
 
   return { server, mockQuery }
 }
+
+type InternalQuery<Input, Output> = {
```
infomiho, line 71: We are hiding the `route` from user?

waspc/data/Generator/templates/server/src/_types/index.ts
```diff
@@ -72,20 +73,11 @@ type Context<Entities extends _Entity[]> = Expand<{
 }>
 
 {=# isAuthEnabled =}
-type ContextWithUser<Entities extends _Entity[]> = Expand<Context<Entities> & UserInContext>
+type ContextWithUser<Entities extends _Entity[]> = Expand<Context<Entities> & { user: SanitizedUser}>
```
infomiho, line 76: Nit 
```suggestion
type ContextWithUser<Entities extends _Entity[]> = Expand<Context<Entities> & { user: SanitizedUser }>
```

waspc/data/Generator/templates/server/src/routes/operations/_action.js
infomiho: Maybe write a helper function called `getRequestBody` or similar so it's more readable.

waspc/data/Generator/templates/universal/types.ts
```diff
@@ -0,0 +1,16 @@
+// This is a helper type used exclusively for DX purposes. It's a No-op for the
+// compiler, but expands the type's representatoin in IDEs (i.e., inlines all
+// type constructors) to make it more readable for the user.
+//
+// It expands this SO answer to funcitons: https://stackoverflow.com/a/57683652
```
infomiho, line 5:
```suggestion
// It expands this SO answer to functions: https://stackoverflow.com/a/57683652
```

waspc/examples/todoApp/src/client/App.tsx
```diff
@@ -1,18 +1,25 @@
...
+  const { data: date } = useQuery(getDate)
 
   return (
     <div className="app border-spacing-2 p-4">
       <header className="flex justify-between">
         <h1 className="font-bold text-3xl mb-5">
           <Link to="/">ToDo App</Link>
         </h1>
+        <h2>
+          Your site was loaded at: {date?.toLocaleString()}
```
infomiho, line 21: Like in the good old PHP websites. 
Screenshot 2023-04-06 at 16 55 28

waspc/examples/todoApp/src/client/pages/Task.tsx
```diff
@@ -13,9 +13,9 @@ type TaskPayload = Pick<Task, "id" | "isDone">
 const Todo = (props: any) => {
   const taskId = parseInt(props.match.params.id)
 
-  const { data: task, isFetching, error } = useQuery<Pick<Task, "id">, Task>(getTask, { id: taskId })
```
infomiho, line 16: Nice

waspc/examples/todoApp/todoApp.wasp
```diff
@@ -29,7 +29,7 @@ app todoApp {
     setupFn: import setup from "@server/serverSetup.js"
   },
   client: {
-    rootComponent: import { App } from "@client/App.jsx",
+    rootComponent: import { App } from "@client/App",
```
infomiho, line 32: We wanted to keep the extensions in the `main.wasp` so it's not confusing that we MUST include them in backend but they are OPTIONAL in the client. 🤷‍♂️

waspc/data/Generator/templates/react-app/src/actions/index.ts
shayneczyzewski: I appreciate the depth of this comment. The TS is still a bit of magic to me that enables it, but I appreciate how much thought went into this! :D

waspc/data/Generator/templates/react-app/src/test/vitest/helpers.tsx
sodic, line 1: Most of the changes in the file are prettier formatting.

waspc/data/Generator/templates/react-app/src/test/vitest/helpers.tsx
```diff
@@ -9,27 +10,39 @@ import { Query } from '../../queries'
...
 import { HttpMethod } from '../../types'
 
+export type Route = { method: HttpMethod; path: string }
+
+export type MockQuery = <Input, Output, MockOutput extends Output>(
+  query: Query<Input, Output>,
+  resJson: MockOutput
+) => void
+
+export type MockApi = (route: Route, resJson: unknown) => void
```
sodic, line 20: The rule of thumb for exporting types is: _If users can dig it out themselves (e.g., using utility types like `Parameters`), make it public._

waspc/data/Generator/templates/react-app/src/auth/useAuth.ts
```diff
@@ -1,16 +1,19 @@
+import { deserialize as superjsonDeserialize } from 'superjson'
 import { useQuery } from '../queries'
 import api, { handleApiError } from '../api'
 import { HttpMethod } from '../types'
+// todo(filip): turn into a proper import
+import { type SanitizedUser as User } from '../../../server/src/_types/' 
 
-export default function useAuth(queryFnArgs, config) {
+export default function useAuth(queryFnArgs?: unknown, config?: any) {
```
sodic, line 8: It seems like `useQuery` will send `queryFnArgs` to `getMe`, but `getMe` doesn't expect them so they end up ignored.

Todo: make an issue for this or investigate.

waspc/data/Generator/templates/react-app/src/test/vitest/helpers.tsx
sodic: Yes, for the time being, reasoning here: https://github.com/wasp-lang/wasp/pull/1055#discussion_r1150429257 (spoiler: not much of a reasoning)

waspc/examples/todoApp/todoApp.wasp
sodic: Changing it back... and the worst thing is: I think it was my suggestion :smile:

shayneczyzewski (approved):
Functionally looks good to me! Great job @sodic! Much of the TS is still over my head, but I can see the benefits of this already. :D 🥳 🚀

infomiho (approved):
LGTM, it's really cool to get type in FE just like that 🤯

sodic:
I'm merging to get RC out asap, but will take a look at what else to do tomorrow.

PR #1092, opened 2023-03-31
Middleware customization

shayneczyzewski:
## Current implementation

Currently, we add all of our middleware to the `app` very early on and as such it would apply to _every route_. Once installed, they cannot be removed. For example:

```js
const app = express()

app.use(helmet())
app.use(cors({
  // TODO: Consider allowing users to provide an ENV variable or function to further configure CORS setup.
  origin: config.allowedCORSOrigins,
}))
app.use(logger('dev'))
app.use(express.json())
app.use(express.urlencoded({ extended: false }))
app.use(cookieParser())

app.use('/', indexRouter)
```

## Goals and use cases

The goal of this PR is to give users 3 ways to optionally customize this:

### 1) Customizations a global level (i.e. applying to every `query`, `action`, and `api`). They can add new middleware, or remove some of our defaults.

Sample use cases:
  - Modify our default CORS setup, use different logging, custom rate limiting, etc.

### 2) On a per-`api` level, they can add new middleware or remove some of the output of (1).

Sample use cases:
  - In the Stripe callback, you cannot use `express.json` and must use the raw body parser since they need to calculate the checksum of the body unmodified.

## Approach

The only way I could come up with to implement this is to install middleware on a per-router level. I introduce a notion of "named middleware" that users can then modify. For example:

```ts
export type MiddlewareConfig = Map<string, express.RequestHandler>

export type MiddlewareConfigFn = (middlewareConfig: MiddlewareConfig) => MiddlewareConfig

// This is the set of middleware Wasp supplies by default.
const defaultGlobalMiddlewareConfig: MiddlewareConfig = new Map([
  ['helmet', helmet()],
  ['cors', cors({ origin: config.allowedCORSOrigins })],
  ['logger', logger('dev')],
  ['express.json', express.json()],
  ['express.urlencoded', express.urlencoded({ extended: false })],
  ['cookieParser', cookieParser()]
])
```

Users can then define `MiddlewareConfigFn` for either the global level, per-`api`, or per-path. Exe:

```c
app todoApp {
   ...
  server: {
    setupFn: import setup from "@server/serverSetup.js",
    middlewareConfigFn: import { serverMiddlewareFn } from "@server/serverSetup.js"
  },
}

apiNamespace fooBar {
  middlewareConfigFn: import { fooBarNamespaceMiddlewareFn } from "@server/apis.js",
  path: "/foo/bar"
}

api fooBar {
  fn: import { fooBar } from "@server/apis.js",
  middlewareConfigFn: import { fooBarMiddlewareFn } from "@server/apis.js",
  entities: [Task],
  httpRoute: (GET, "/foo/bar")
}
```

```ts
export const serverMiddlewareFn: MiddlewareConfigFn = (middleware) => {
  middleware.set('custom.global',
    (_req, _res, next) => {
      console.log('custom global middleware')
      next()
    }
  )
  return middleware
}

export const fooBarMiddlewareFn: MiddlewareConfigFn = (middleware) => {...}
```

We then install the (possibly modified) global defaults in the index router like so:

```js
const router = express.Router()
const middleware = globalMiddlewareForExpress()

router.get('/', middleware, function (_req, res, _next) {
  res.json('Hello world')
})

router.use('/auth', middleware, auth)
router.use('/operations', middleware, operations)
// NOTE: Keep user-defined api routes last so they cannot override our routes.
// Additionally, do not add middleware to these routes here. Instead, we add
// it later to allow for middleware customization.
router.use(apis)

export default router
```

And the `api` specific middleware mods like so:

```js
const router = express.Router()

// path-specific
router.use('/foo/bar', globalMiddlewareForExpress(fooBarnamespaceMiddlewareConfigFn))

// api-specific
const fooBarMiddleware = globalMiddlewareForExpress(fooBarmiddlewareConfigFn)
router.get(
  '/foo/bar',
  [auth, ...fooBarMiddleware],
  handleRejection(
    (
```

Releated VS Code update: https://github.com/wasp-lang/vscode-wasp/pull/12


Martinsos:
@shayneczyzewski looks good to me and well thought out!

Main thoughts:
1. If they modify the global middleware, they could possibly mess up Wasp quite a bit, if some other parts of Wasp are relying on having some of that middleware present, right? How are we feeling about that? I guess that can be fine, if we warn them that by doing it they are risking messing up stuff. Would be even better if we could somehow guarantee they can't mess things up, beacuse that would make it easier for them to use it -> imagine user having to remove that JSON parsing and not being sure if they broke something, that is pretty tricky hm. I wonder now if we should even allow them to customize global middleware so directly. Not sure how to go about this. One option could be to give them some kind of configuration they can configure, but that is made in such way that it can't mess up stuff for us. This does sound quite tricky to design though, hard to guess what they need and also to ensure it doesn't mess anything up on our side. At the end they can probably always add some kind of middleware that messes some stuff up.
2. Sounds a bit bad that they can't define different GET and POST routers for the same route? I have hard time judging how bad is that though, but at first it sounds a bit invasive from our side. Maybe if we said that happens ONLY if they use custom middleware, that is a bit better then? I have to admit I don't understand the intricacies of Express here, probably beacuse I am missing Express knowledge, but I am guessing there is no way to specify middleware for all of the possible methods and then still provide specific implemetnation for each method?

waspc/data/Generator/templates/server/src/app.js
```diff
@@ -1,32 +1,19 @@
...
 
-app.use(helmet())
-app.use(cors({
-  // TODO: Consider allowing users to provide an ENV variable or function to further configure CORS setup.
-  origin: config.allowedCORSOrigins,
-}))
-app.use(logger('dev'))
-app.use(express.json())
-app.use(express.urlencoded({ extended: false }))
-app.use(cookieParser())
```
Martinsos, line 24: What if some middleware we still installed here always, like logger and helmet and stuff, but then some other gets applied per-route basis? Because maybe for some of them it makes sense to use different rules?

waspc/data/Generator/templates/server/src/middleware.ts
```diff
@@ -0,0 +1,41 @@
...
+const _defaultMiddleware: MiddlewareConfig = new Map([
+  ['helmet', helmet()],
+  ['cors', cors({ origin: config.allowedCORSOrigins })],
+  ['logger', logger('dev')],
+  ['express.json', express.json()],
+  ['express.urlencoded', express.urlencoded({ extended: false })],
+  ['cookieParser', cookieParser()]
+])
+
+const defaultMiddleware = {=& globalMiddlewareConfigFnImportAlias =}(_defaultMiddleware)
```
Martinsos, line 30: I am having a bit of trouble understanding how is this working, what is this `globalMiddlewareConfigFnImportAlias`?

waspc/data/Generator/templates/server/src/middleware.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export function getDefaultMiddleware(): MiddlewareConfig {
+  // Return a clone so they can't mess up the Map for any other routes.
+  return new Map(defaultMiddleware)
```
Martinsos, line 34: Are you sure this will ensure they can't mess it up? Is this clone deep enough, maybe they can modify something still? Maybe we don't have to protect it that deeply though, just wondering.

waspc/data/Generator/templates/server/src/middleware.ts
Martinsos: Ah ok I kind of get it now! For some reason I found it pretty tricky to understand this file.
Do you think we could somehow make it easier to grasp? I found it somewhat hard to understand the relationships / what is important what is not.
Maybe some comments, different naming, reordering stuff or similar could help?

waspc/data/Generator/templates/server/src/middleware.ts
```diff
@@ -0,0 +1,41 @@
...
+  ['cors', cors({ origin: config.allowedCORSOrigins })],
+  ['logger', logger('dev')],
+  ['express.json', express.json()],
+  ['express.urlencoded', express.urlencoded({ extended: false })],
+  ['cookieParser', cookieParser()]
+])
+
+const defaultMiddleware = {=& globalMiddlewareConfigFnImportAlias =}(_defaultMiddleware)
+
+export function getDefaultMiddleware(): MiddlewareConfig {
```
Martinsos, line 32: This is really a global middleware, right? So we should probably call it that way.
Also, it is not really default, is it? Because it can get modified by that globalMiddlewareConfigFnImportAlias function, so then it is not default any more?

Martinsos: File should probably also be renamed. Actually what I guess is confusing that in this file concepts of middleware in general + global middleware are mixed?

When dealing with API routes you handle modification of default middleware right there, in place, while for global middleware do you do it in here in this file -> I would maybe also do that rather directly in place, or create a separate file/module for it.

waspc/data/Generator/templates/server/src/routes/apis/index.ts
```diff
@@ -2,18 +2,35 @@
...
+
 {=/ apiRoutes =}
 
 const router = express.Router()
 
+/*
+  NOTE: Since some middleware needs to run _before_ the method, like CORS for GET/POST requests,
+  we need to use `Router.use`. However, since this applies to all methods for that route, this
+  means we can only support _unique_ route paths.
+*/
```
Martinsos, line 31: This confuses me a bit, I would think that we can somehow go about supporting different methods.

Problem is actually that middleware is applied per route, not per method, is that it?
Maybe then we also shouldn't apply it per method in Wasp, but apply it per route instead? You know, have something like `server-route "/foo/bar" { middlewareFn: ... }`? I made up this syntax, but I guess something like this? This does complicate things a bit though hm, but maybe is a direction we should actually explore?

waspc/data/Generator/templates/server/src/app.js
shayneczyzewski: I tried to document our defaults a bit, and made the note that `express.json` was the main one we need for Wasp to work. Plus a warning :) Hopefully this will be enough for them on the global front.

waspc/data/Generator/templates/server/src/middleware.ts
shayneczyzewski: Tried to clean this up and comment a bit 👍🏻

waspc/data/Generator/templates/server/src/middleware.ts
shayneczyzewski: Pretty sure. We also chatted on this and it sounds fine 👍🏻

waspc/data/Generator/templates/server/src/middleware.ts
shayneczyzewski: Renamed file and variable names. Created one entrypoint for mods.

waspc/data/Generator/templates/server/src/routes/apis/index.ts
shayneczyzewski: This is no longer needed with `namespace`, which is similar to your `server-route`. Open to tweaking that name/properties.

waspc/data/Generator/templates/server/src/middleware/globalMiddleware.ts
```diff
@@ -0,0 +1,46 @@
...
+import config from '../config.js'
+
+{=# globalMiddlewareConfigFnDefined =}
+{=& globalMiddlewareConfigFnImportStatement =}
+{=/ globalMiddlewareConfigFnDefined =}
+{=^ globalMiddlewareConfigFnDefined =}
+const {=& globalMiddlewareConfigFnImportAlias =} = (m: MiddlewareConfig) => m
+{=/ globalMiddlewareConfigFnDefined =}
+
+export type MiddlewareConfig = Map<string, express.RequestHandler>
```
shayneczyzewski, line 17: Kept as a `Map` instead of `ReadonlyMap` just to make it easier to work with in those `MiddlewareConfigFn` for users. We clone the defaults so they are not impacted.

web/docs/guides/middleware-customization.md
```diff
@@ -0,0 +1,134 @@
...
+- [CORS](https://github.com/expressjs/cors#readme): CORS is a package for providing a middleware that can be used to enable [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) with various options.
+- [Morgan](https://github.com/expressjs/morgan#readme): HTTP request logger middleware.
+- [express.json](https://expressjs.com/en/api.html#express.json) / [body-parser](https://github.com/expressjs/body-parser#bodyparserjsonoptions): Parses incoming request bodies in a middleware before your handlers, making the result available under the `req.body` property. ⚠️ This is required for Wasp Operations to function properly.
+- [express.urlencoded](https://expressjs.com/en/api.html#express.urlencoded) / [body-parser](https://expressjs.com/en/resources/middleware/body-parser.html#bodyparserurlencodedoptions): Returns middleware that only parses urlencoded bodies and only looks at requests where the `Content-Type` header matches the type option.
+- [cookieParser](https://github.com/expressjs/cookie-parser#readme): Parse Cookie header and populate `req.cookies` with an object keyed by the cookie names.
```
Martinsos, line 17: Cool! O, so which of these they must not touch? You mentioned body-parser is needed for Operations, what about the rest? Are any of these freely removable if they want? I guess Helmet yes, and Morgan also, waht about the rest?

Oh should we drop a NOTE in the code in JS where we actually define these midleware, to update docs if that part is changed, since it is public in a way?

web/docs/guides/middleware-customization.md
```diff
@@ -0,0 +1,134 @@
...
+# Customizing Express server middleware
+
+Wasp comes with a minimal set of useful Express middleware in every application. While this is good for most users, we realize some may wish to add, modify, or remove some of these choices both globally, or on a per-`api`/path basis.
+
+## Default global middleware
+
+- [Helmet](https://helmetjs.github.io/): Helmet helps you secure your Express apps by setting various HTTP headers. It's not a silver bullet, but it can help!
+- [CORS](https://github.com/expressjs/cors#readme): CORS is a package for providing a middleware that can be used to enable [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) with various options.
+- [Morgan](https://github.com/expressjs/morgan#readme): HTTP request logger middleware.
+- [express.json](https://expressjs.com/en/api.html#express.json) / [body-parser](https://github.com/expressjs/body-parser#bodyparserjsonoptions): Parses incoming request bodies in a middleware before your handlers, making the result available under the `req.body` property. ⚠️ This is required for Wasp Operations to function properly.
```
Martinsos, line 15: What does express.json / body-parser mean, as regarding to the `/`? Is that "both"?

web/docs/guides/middleware-customization.md
```diff
@@ -0,0 +1,134 @@
...
+- [cookieParser](https://github.com/expressjs/cookie-parser#readme): Parse Cookie header and populate `req.cookies` with an object keyed by the cookie names.
+
+## Customization
+
+You have three places where you can customize middleware:
+1. global: here, any changes will apply by default *to all operations (`query` and `action`) and `api`.* This is helpful if you wanted to add support for multiple domains to CORS, for example. ⚠️ Please treat modifications to global middleware with extreme care!
+2. per-api: you can override middleware for a specific api route (exe: `POST /webhook/callback`). This is helpful if you want to disable JSON parsing, for example.
+3. per-path: this is helpful if you need to customize middleware for all methods for a given path. This is helpful for things like "complex CORS requests," which may need to apply to both `OPTIONS` and `GET`, or to apply some middleware to a _set of `api` routes_.
```
Martinsos, line 24: Can they also affect operations with this? They can if, `namespace` is targetting the path behind which operation operates? Ok, I guess they couldn't really know that path hm, they would have to hit it by accident. ok ok ok.

web/docs/guides/middleware-customization.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
+```ts title=src/server/serverSetup.js
+import cors from 'cors'
+import { MiddlewareConfigFn } from '@wasp/middleware'
+import config from '@wasp/config.js'
+
+export const serverMiddlewareFn: MiddlewareConfigFn = (middleware) => {
+  // Example of adding an extra domains to CORS.
+  middleware.set('cors', cors({ origin: [config.frontendUrl, 'https://example1.com', 'https://example2.com'] }))
```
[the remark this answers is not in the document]
Martinsos: Aha, making config part of API here. Hmmmm. This asks a good question, what is part of API? We don't really have a way to track it, we kind of know it, but we don't have it "secured" in a code -> I could fiddle with `config.js` and forget we promised users it will be there!
In general yeah I think we can expose it, it is a way to get to env vars. Although they don't know much about config, we would have to tell them what is in it then, how would they know it has `frontendUrl`?. Uff, ok this gets complex now hm. Maybe we can leave it a bit greyish for now? Or do you have a better idea how to go about it?

web/docs/guides/middleware-customization.md
```diff
@@ -0,0 +1,134 @@
...
+router.post('/webhook/callback', webhookCallbackMiddleware, ...)
+```
+:::
+
+## 3. Customize path-path middleware
+
+If you would like to modify the middleware for all API routes under some common path, you can do something like:
+
+```c title=todoApp.wasp
+namespace fooBar {
```
Martinsos, line 120: So this is cool! How are you feeling about `namespace` name? I find it pretty generic, maybe too generic. Is there a good, more specific name? That makes it clear we are dealing with `api` here? `apiRoute`? Even `apiNamespace` or `apiPath` is better than just `namespace` I think.

Martinsos (commented):
@shayneczyzewski this looks pretty good! I think all is here, there are some possible kinks to check before finalizing it, but in general I think we are there.
It seems to me like this is pretty detailed and useful, as a feature. It is a bit low-level maybe, I would in theory love it to be a bit more high level, but in practice I think this it. We can see how it evolves with time, this is the best place to be right now I think.

shayneczyzewski (commented):
Thanks for another round @Martinsos! I will implement the feedback today/tomorrow👍🏻

waspc/data/Generator/templates/server/src/middleware/globalMiddleware.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export type MiddlewareConfigFn = (middleware: MiddlewareConfig) => MiddlewareConfig
```
infomiho, line 19: Is the parameter `middleware` or `middlewareConfig`? :)

waspc/data/Generator/templates/server/src/middleware/globalMiddleware.ts
```diff
@@ -0,0 +1,46 @@
...
+
+// This is the set of middleware Wasp supplies by default.
+const defaultGlobalMiddleware: MiddlewareConfig = new Map([
```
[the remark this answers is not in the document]
infomiho: Ah, didn't see this comment. +1 on `middleware` naming being a bit confusing it's either a `config` or `enabledMiddleware` or `middlewareConfig`

waspc/src/Wasp/Generator/ServerGenerator/ApiRoutesG.hs
```diff
@@ -39,30 +38,49 @@ genApiRoutes :: AppSpec -> Generator FileDraft
...
+    getApiRoutesTmplData :: (String, Api.Api) -> Aeson.Value
+    getApiRoutesTmplData (apiName, api) =
+      object
+        [ "routeMethod" .= map toLower (show $ Api.method api),
+          "routePath" .= Api.path api,
+          "importStatement" .= jsImportStmt,
+          "importIdentifier" .= jsImportIdentifier,
+          "entities" .= getApiEntitiesObject api,
+          "usesAuth" .= isAuthEnabledForApi spec api,
+          "routeMiddlewareConfigFnDefined" .= isJust maybeMiddlewareConfigFnImport,
```
infomiho, line 72: Maybe consider grouping things into object e.g. 
```
"routeMiddlewareConfigFn" .= object [ "isDefined" .= ... ]
```
to make it a bit less busy

waspc/src/Wasp/Generator/ServerGenerator/JsImport.hs
```diff
@@ -29,6 +30,14 @@ getJsImportStmtAndIdentifier ::
   (JsImportStatement, JsImportIdentifier)
 getJsImportStmtAndIdentifier pathFromImportLocationToExtCodeDir = JI.getJsImportStmtAndIdentifier . extImportToJsImport pathFromImportLocationToExtCodeDir
 
+getAliasedJsImportStmtAndIdentifier ::
```
infomiho, line 33: Have you looked at `extImportToImportJson` and maybe provide a fn like that for aliased imports?

web/docs/guides/middleware-customization.md
infomiho, line 24:
```suggestion
3. per-path: this is helpful if you need to customize middleware for all methods for a given path. This is helpful for things like "complex CORS requests" which may need to apply to both `OPTIONS` and `GET`, or to apply some middleware to a _set of `api` routes_.
```

web/docs/guides/middleware-customization.md
```diff
@@ -0,0 +1,134 @@
...
+  ['helmet', helmet()],
+  ['cors', cors({ origin: config.allowedCORSOrigins })],
+  ['logger', logger('dev')],
+  ['express.json', express.json()],
+  ['express.urlencoded', express.urlencoded({ extended: false })],
+  ['cookieParser', cookieParser()]
+])
+```
+
+Here is the [definition of config](https://github.com/wasp-lang/wasp/blob/main/waspc/data/Generator/templates/server/src/config.js).
```
infomiho, line 45: I would avoid linking to our internal templates (which contain Mustache and are barely readable by us 😂 ) maybe list the config options?

web/docs/guides/middleware-customization.md
infomiho: +1 on `apiNamespace`, I would also think about what is this thing? For now it's a path definition which has custom middleware as one of its options. Will we expand in the future? Maybe add some sort of auth option to control access to a subset of routes? If yes, which name makes sense? 

Laravel calls this things `routeGroup` for example 🤷  https://laravel.com/docs/10.x/routing#route-groups

infomiho: @shayneczyzewski don't forget the VS Code extension update as well 
Screenshot 2023-04-19 at 21 35 56

web/docs/guides/middleware-customization.md
```diff
@@ -0,0 +1,134 @@
...
+
+```c title=todoApp.wasp
+namespace fooBar {
+  middlewareConfigFn: import { fooBarNamespaceMiddlewareFn } from "@server/apis.js",
+  path: "/foo/bar"
+}
+```
+
+:::note
+The JS/TS implementation is not shown, as it would be similar to the examples above.
```
infomiho, line 127: Why not? For completeness sake and if somebody links to this piece directly, they didn't really read the previous part.

infomiho:
Tested it out locally. The implementation seems sound 👍

I like what this enables:
- CORS escape hatch
- Stripe web hooks
- Auth middleware per route prefix

Left some minor comments, the biggest unknown for me is the `namespace` naming.

### TS complains if I don't put the explicit type on the `Map`

Screenshot 2023-04-19 at 21 36 38
Screenshot 2023-04-19 at 21 41 17

waspc/data/Generator/templates/server/src/middleware/globalMiddleware.ts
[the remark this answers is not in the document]
Martinsos: makes sense @shayneczyzewski let's add Config suffix to those then!

web/docs/guides/middleware-customization.md
Martinsos: Ok, whatever you think makes the most sense here!

web/docs/guides/middleware-customization.md
Martinsos: I would certainly prefix it with `api` for now. If in the future we come up with better DSL for all this stuff then fine, but for now I think it is important to make it clear that it is related to the `api`. So `apiNamespace` or `apiRoute` is ok with me.

Martinsos (approved):
I am giving approve, since all of my concerns have been addressed or will be addressed!

waspc/src/Wasp/Generator/ServerGenerator/JsImport.hs
shayneczyzewski: Do you mean in addition to this one? Can we just add that once we need it, or do you think it is needed now?

web/docs/guides/middleware-customization.md
shayneczyzewski: I will nix it, I point out the only real value they should want to use in an example, and we want to keep config a sort of gray/mysterious area. :)

web/docs/guides/middleware-customization.md
shayneczyzewski: I'll just go with `apiNamespace` for v1. Using `apiRoute` feels like it could confuse people with our existing `route` notion. I think this should probably feel pretty distinct and vague until we know how it will evolve, and a name we don't love will force us to come back to it lol Sound ok?

web/docs/guides/middleware-customization.md
shayneczyzewski: Just to close the loop here,  after some testing, only `CORS` and `express.json` appear to be required. Updated docs with warnings for both of those.

PR #1095, opened 2023-04-01
Db seeding

Martinsos:
New feature! I took care of Changelog, cabal version, docs, tests.

So now you can do `wasp db seed` or `wasp db seed <name>` to run the seeding function, which does some stuff on the database!

We use `prisma seed` in the background, although from what I understand it does so little that we could have just implemented it without prisma, but I decided to still go with them just in case.

One thing that prisma doesn't offer but we do is having multiple seeds, so you can easily pick which one you want at specific moment.

Also they by default run seeding after reseting the database or after some migrations, but I turned that off, I think it is too aggresive, and I think people should be able to reset the db without seeding running in case they want to do something manually with it.

Check it out!

@infomiho already helped me with types so  maybe you could take a final look at TS types @sodic !


waspc/data/Generator/templates/server/package.json
```diff
@@ -10,8 +10,8 @@
     "build-and-start": "npm run build && npm run start",
     "watch": "nodemon --exec 'npm run build-and-start || exit 1'",
     "validate-env": "node -r dotenv/config ./scripts/validate-env.mjs",
+    "db-seed": "npm run build && NODE_PATH=dist node -r dotenv/config dist/dbSeed.js",
     "db-migrate-prod": "prisma migrate deploy --schema=../db/schema.prisma",
-    "db-migrate-dev": "prisma migrate dev --schema=../db/schema.prisma",
```
Martinsos, line 14: I realized we don't use this so I deleted it.

waspc/data/Generator/templates/server/package.json
```diff
@@ -30,6 +30,7 @@
   "engines": {
     "node": "{=& nodeVersionRange =}"
   },
+  "prisma": {=& prisma =},
```
Martinsos, line 33: This is how we specify seed script for Prisma.

waspc/examples/todoApp/src/server/dbSeeds.ts
Martinsos, line 1: One thing I wonder here is if we could offer them some nice utility functions for stuff like creating user (and removing password from it) and for constructing context for actions, based on user and prismaClient, so they don't have to do that manually.

I am not sure how to go about it exactly, so any ideas are welcome!

For users for example, I would probably have to offer them createUser function based on the types of auth they use, right? So for more complex stuff like social login and similar, I don't even know how to easily construct it here, is that even possible unless some heavy mocking is done (how would that even work)? It is easy if it is username/pass, but what if they are not using that?

waspc/examples/todoApp/src/server/dbSeeds.ts
```diff
@@ -0,0 +1,35 @@
+import { createTask } from './actions.js'
+import type { DbSeedFn } from '@wasp/dbSeed/types.js'
+
+async function createUser (prismaClient, data) {
+  const { password, ...newUser } = await prismaClient.user.create({ data })
+  return newUser
+}
+
+export const devSeedSimple: DbSeedFn = async (prismaClient) => {
```
Martinsos, line 9: I also wonder here if it would be better to pass `entities` instead of `prismaClient`, since they are used to `entities` in Operations? But, then `entities` would need to contain all the entitites in them!
Also, they still might want to import prisma client at some point if they need some fancy stuff like raw queries. We do have docs for importing it though. So I am not sure hm.

waspc/examples/todoApp/src/server/dbSeeds.ts
shayneczyzewski: Yeah that is a good point, helpers for complex entity relationships would be very nice to have. I think creating users is probably a good place to start.

waspc/src/Wasp/Generator/ServerGenerator/Db/Seed.hs
```diff
@@ -0,0 +1,67 @@
...
+dbSeedScriptInServerTmplSrcDir :: Path' (Rel C.ServerTemplatesSrcDir) (File ())
+dbSeedScriptInServerTmplSrcDir = [relfile|dbSeed.ts|]
+
+pathFromDbSeedScriptToServerSrc :: Path Posix (Rel ()) (Dir C.ServerSrcDir)
+pathFromDbSeedScriptToServerSrc = [reldirP|./|]
+
+getPackageJsonPrismaSeedField :: AppSpec -> Maybe String
+getPackageJsonPrismaSeedField spec =
+  case getDbSeeds spec of
+    Just (_ : _) -> Just "npm run db-seed"
```
shayneczyzewski, line 45:
```suggestion
    Just _ -> Just "npm run db-seed"
```

waspc/examples/todoApp/src/server/dbSeeds.ts
shayneczyzewski: Prisma client is probably ok here.

shayneczyzewski (commented):
Looking good @Martinsos! Did a first pass on code and tested it, only had one major question (about the first seed being the default). I will do another round/deeper dive tomorrow as a final check 👍🏻

waspc/examples/todoApp/migrations/20230328122947_init/migration.sql
```diff
@@ -0,0 +1,41 @@
+-- CreateTable
```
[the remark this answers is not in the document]
Martinsos: I had that weird problem where it was complaining about some migration file missing so I just blew them away. I do have to admit I did that without thinking too much, I was just trying to get past that error. Do you think that was wrong? What should I do instead?

waspc/examples/todoApp/src/server/dbSeeds.ts
Martinsos: Ok thanks that helps! I could have also let them import it, but I prefer this, this way it is clear we want them to use it plus it makes mocking possibly easier.

waspc/examples/todoApp/src/server/dbSeeds.ts
Martinsos: I do have to admit I am not sure how to go about it in a simple way. Can I somehow easily offer them those functions right now? Do you have any suggestion? Maybe some piece of auth code I can just wrap a bit and ta-da, here it is? Can it even be done easily for social auth at all, without some mocking? How would we do mocking even?

waspc/ChangeLog.md
```diff
@@ -2,6 +2,41 @@
...
+    { description: "Chase the cat" },
+    { user: newUser, entities: { Task: prismaClient.task } }
+  )
+}
+
+//...
+```
+
+`wasp db seed` runs the first(default) seed, while `wasp db seed <name>` will run a seed with specific name.  
+For example, for the case above, you could do `wasp db seed prodSeed`.
```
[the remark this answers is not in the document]
Martinsos: You are right! I just wanted to make it simple, but you make a good point, it is certianly not intuitive and it makes you think all of them will run. I will play a bit with it in the direction of what you suggested and make it more intuitive!

waspc/cli/src/Wasp/Cli/Command/Db/Seed.hs
```diff
@@ -0,0 +1,43 @@
...
+seed maybeSeedName = do
+  waspProjectDir <- findWaspProjectRootDirFromCwd
+  let genProjectDir =
+        waspProjectDir </> Common.dotWaspDirInWaspProjectDir </> Common.generatedCodeDirInDotWaspDir
+
+  appSpec <- analyze waspProjectDir
+  when (isNothing $ getDbSeedFn appSpec) $
+    (E.throwError . CommandError "No seed specified") $
+      "You haven't specified any database seeding functions, so there is nothing to run!\n"
+        <> "To do so, set app.db.seed in your Wasp config."
```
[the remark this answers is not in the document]
Martinsos: Good point, will improve this!

waspc/src/Wasp/Generator/ServerGenerator/Db/Seed.hs
Martinsos: The difference is that `(_ : _)` doesn't match empty list -> it needs to have at least one element. While `_` would also match an empty list.
This will probably go away after I make changes that you suggested for not keeping everything in one list.

waspc/ChangeLog.md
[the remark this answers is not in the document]
Martinsos: I have been looking more into it now, and there is one thing -> the list is simple.
In the sense that they don't have to think much about when specifying seeds.
Yes, only first will run by default, and that is not obvious, but it is really easy to specify seeds.

While if we have `default` and `others`, they have to start immediatelly thinking about where to put and how to organize the seeds. Also `others` is a bit weird name, not terrible, but also not great, and I can't come up with a better name.

I will try writing down a couple of approaches:
1.
```
seeds: [<extimport>]
```
This is what we have now. It is not clear that first one is default, but that is kind of  "a happy little accident". It does make defining them really simple, not much thinking.

2.
I could do she same thing as (1), just remove the idea of default seed -> you always have to specify the name of the seed in `wasp db seed <name>`. DSL is kept simple, the idea of "default" is lost, but at least there is no confusion.

3. 
```
seeds: {
  default: <extimport>,
  others: [<extimport>]
}
```
Here it is very clear what the default is. But I find it a bit cumbersome to separate them between the default and others. It could be a hard decision, and it has to be intentional.

Also, do we allow them to not specify a `default`, but specify `others`? So:
```
seeds: {
  others: [<extimport>]
}
```
which would mean they don't want to have a default, they just have seeds.
Should we allow this if we are going with this? Hm.

4.
```
seeds: {
  default: "devDbSimple",
  fns: [<extimport>]
```
This could be one way to do it. We have `fns` as the main thing, and one can choose one from them to be the default one. `default` would be optional field, `fns` would be required. We would have to check in validation step that `default` matches one of the identifiers in the `fns` field.


I am a bit torn between all of these! You are right that `seeds` being a list makes behaviour of default one being first somewhat unintuitive and therefore change in order can make it confusing. On the other hand, it works nicely when you have just one seed, which is often going to be the case I guess, and also this is something used in development, so it is not like they will be surprised in production with a sudden change. So I am at the end tempted to leave it as it is merely due to how simple it is.

Let's say they don't expect the first element to be run by default. They run `db seed`, they see that first seed was used, and now they know! Maybe that is not terrible?

waspc/cli/src/Wasp/Cli/Command/Db/Seed.hs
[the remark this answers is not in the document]
Martinsos: Fixed

waspc/examples/todoApp/migrations/20230328122947_init/migration.sql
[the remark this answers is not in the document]
Martinsos: Ok, I should try to reverse migrations as they were before PR and just add new one if needed, instead of resetting them completely.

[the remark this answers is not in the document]
Martinsos: Fixed! Worked from first try, I just checked out those from main and all was good.

waspc/src/Wasp/Generator/ServerGenerator/Db/Seed.hs
shayneczyzewski: No need to apologize, it was an incorrect suggestion :) no offense taken

waspc/examples/todoApp/src/server/dbSeeds.ts
shayneczyzewski: Yeah I am a bit stumped as well, since it depends a bit on the shape of their user model. :/ The helper would probably just end up taking the same data as them just setting it up in Prisma/via entities. Maybe we just punt for now until people ask for specific helpers.

waspc/ChangeLog.md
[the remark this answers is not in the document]
shayneczyzewski: Nice! You will also want to update the changelog comment too I think:

> `wasp db seed` runs the first(default) seed

waspc/cli/exe/Main.hs
```diff
@@ -167,6 +170,9 @@ printDbUsage =
               "",
         title "COMMANDS",
         cmd   "  reset         Drops all data and tables from development database and re-applies all migrations.",
+        cmd   "  seed [name]   Executes the db seed function (specified via app.db.seed).",
+        cmd   "                If there are multiple seeds, you can specify a seed to execute by providing its name.",
+        cmd   "                By default it will execute the first seed.",
```
shayneczyzewski, line 175: This default behavior changed.

waspc/cli/src/Wasp/Cli/Command/Db/Seed.hs
```diff
@@ -0,0 +1,94 @@
...
+import Control.Monad.IO.Class (liftIO)
+import Data.List (intercalate)
+import Data.List.NonEmpty (NonEmpty ((:|)))
+import qualified Data.List.NonEmpty as NE
+import StrongPath ((</>))
+import Text.Printf (printf)
+import qualified Wasp.AppSpec as AS
+import qualified Wasp.AppSpec.App as AS.App
+import qualified Wasp.AppSpec.App.Db as AS.Db
+import qualified Wasp.AppSpec.ExtImport as AS
```
shayneczyzewski, line 16: Probably accident to import `as AS`.

web/docs/language/features.md
```diff
@@ -1742,42 +1750,123 @@ app MyApp {
...
+
+async function createUser (prismaClient, data) {
+  const { password, ...newUser } = await prismaClient.user.create({ data })
+  return newUser
+}
+```
+
+#### Running seed functions
+
+Run `wasp db seed` to run the default seed function (the first one in the `app.db.seeds` list).
```
shayneczyzewski, line 1859: May need to update "default" notion here.

shayneczyzewski (approved):
Looks great @Martinsos! I just noticed a few references to "default" you probably want to tidy up, but everything looks good. Oh, and the select when nothing provided is pretty cool! 😎 Nice job! 🚀  🌱

Martinsos:
Uff @shayneczyzewski nice catch I completely forgot to update the docs!
Ok this should be it now, I updated the docs also!

PR #1102, opened 2023-04-04
Relax dependency between oauth and usernameAndPassword

shayneczyzewski:
### Description

This relaxes the relationship between oauth and usernameAndPassword. Before, we implicitly required the User entity to have username/password fields since we checked/set them. Now, we only do is _if they exist on the User entity_. This should be backwards compatible, but also now allow for oauth to be used with no assumptions on the User entity fields beyond an id (which we can later relax the type of).

### DIFF
--- a/waspc/src/Wasp/AppSpec/Valid.hs
+++ b/waspc/src/Wasp/AppSpec/Valid.hs
@@ -5,6 +5,7 @@ module Wasp.AppSpec.Valid
     ValidationError (..),
     getApp,
     isAuthEnabled,
+    doesUserEntityContainField,
   )
 where
 
@@ -193,3 +194,11 @@ getDbSystem spec = AS.Db.system =<< AS.App.db (snd $ getApp spec)
 -- | This function assumes that @AppSpec@ it operates on was validated beforehand (with @validateAppSpec@ function).
 isPostgresUsed :: AppSpec -> Bool
 isPostgresUsed = (Just AS.Db.PostgreSQL ==) . getDbSystem
+
+-- | This function assumes that @AppSpec@ it operates on was validated beforehand (with @validateAppSpec@ function).
+doesUserEntityContainField :: AppSpec -> String -> Maybe Bool
+doesUserEntityContainField spec fieldName = do
+  auth <- App.auth (snd $ getApp spec)
+  let (_userEntityName, userEntity) = AS.resolveRef spec (Auth.userEntity auth)
+  let userEntityFields = Entity.getFields userEntity
+  Just $ any (\field -> Entity.Field.fieldName field == fieldName) userEntityFields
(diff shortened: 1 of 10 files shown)


waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -193,3 +194,12 @@ getDbSystem spec = AS.Db.system =<< AS.App.db (snd $ getApp spec)
 -- | This function assumes that @AppSpec@ it operates on was validated beforehand (with @validateAppSpec@ function).
 isPostgresUsed :: AppSpec -> Bool
 isPostgresUsed = (Just AS.Db.PostgreSQL ==) . getDbSystem
+
+-- | This function assumes that @AppSpec@ it operates on was validated beforehand (with @validateAppSpec@ function).
+-- If there is no user entity, it returns Nothing.
+doesUserEntityContainField :: AppSpec -> String -> Maybe Bool
+doesUserEntityContainField spec fieldName = do
+  auth <- App.auth (snd $ getApp spec)
+  let (_userEntityName, userEntity) = AS.resolveRef spec (Auth.userEntity auth)
```
infomiho, line 203: You could go for `snd` like above :)

waspc/src/Wasp/Generator/ServerGenerator/Auth/OAuthAuthG.hs
```diff
@@ -38,31 +38,32 @@ import Wasp.Generator.ServerGenerator.JsImport (extImportToImportJson)
...
+genCreateRouter :: AS.AppSpec -> AS.Auth.Auth -> Generator FileDraft
+genCreateRouter spec auth = return $ C.mkTmplFdWithData [relfile|src/auth/providers/oauth/createRouter.ts|] (Just tmplData)
   where
     tmplData =
       object
         [ "userEntityUpper" .= (userEntityName :: String),
           "userEntityLower" .= (Util.toLowerFirst userEntityName :: String),
-          "externalAuthEntityLower" .= (Util.toLowerFirst externalAuthEntityName :: String)
+          "externalAuthEntityLower" .= (Util.toLowerFirst externalAuthEntityName :: String),
+          "isPasswordOnUserEntity" .= (doesUserEntityContainField spec "password" == Just True)
```
infomiho, line 66: Nit: I would maybe add a local variable `isPasswordOnUserEntity` and use that to make the `tmplData` less busy

infomiho:
I think this is a good solution for the time being 👍 

I'll test it out locally today, but I think it should work okay by looking at the code.

waspc/src/Wasp/Generator/ServerGenerator/Auth/OAuthAuthG.hs
shayneczyzewski: Updated 👍🏻

waspc/src/Wasp/AppSpec/Valid.hs
shayneczyzewski: I didn't have a strong pref so I just updated with suggestion :) 👍🏻

PR #1105, opened 2023-04-05
Adds scaffolding based on giget

infomiho:
Adds the `--template <name>` option to the `wasp new` command which uses [gitget](https://github.com/unjs/giget) under the hood to fetch templates from the https://github.com/wasp-lang/starters repository.

### How it works
1. The `<name>` represents the folder name in the repo.
2. After we fetch the template we replace some placeholders
    - replacing `__waspProjectName__`, `__waspAppName__` with user provided info (like we did with the regular `wasp new` commend)
    - replacing `__waspVersion__` with current version

### Templates
We can add new templates all the time to the repo and users will be able to use them via the Wasp CLI.

### DIFF
--- a/waspc/cli/exe/Main.hs
+++ b/waspc/cli/exe/Main.hs
@@ -38,7 +38,7 @@ main :: IO ()
 main = withUtf8 . (`E.catch` handleInternalErrors) $ do
   args <- getArgs
   let commandCall = case args of
-        ["new", projectName] -> Command.Call.New projectName
+        ("new" : newArgs) -> Command.Call.New newArgs
         ["start"] -> Command.Call.Start
         ["start", "db"] -> Command.Call.StartDb
         ["clean"] -> Command.Call.Clean
@@ -62,7 +62,7 @@ main = withUtf8 . (`E.catch` handleInternalErrors) $ do
   telemetryThread <- Async.async $ runCommand $ Telemetry.considerSendingData commandCall
 
   case commandCall of
-    Command.Call.New projectName -> runCommand $ createNewProject projectName
+    Command.Call.New newArgs -> runCommand $ createNewProject newArgs
     Command.Call.Start -> runCommand start
     Command.Call.StartDb -> runCommand Command.Start.Db.start
     Command.Call.Clean -> runCommand clean
@@ -104,7 +104,10 @@ printUsage =
               "",
         title "COMMANDS",
         title "  GENERAL",
-        cmd   "    new <project-name>    Creates new Wasp project.",
+        cmd   "    new <name> [args]     Creates a new Wasp project.",
+              "    OPTIONS:",
+              "    --template <template-name>",
+              "",
         cmd   "    version               Prints current version of CLI.",
         cmd   "    waspls                Run Wasp Language Server. Add --help to get more info.",
         cmd   "    completion            Prints help on bash completion.",
@@ -125,6 +128,7 @@ printUsage =
               "",
         title "EXAMPLES",
               "  wasp new MyApp",
+              "  wasp new MyApp --template waspello",
               "  wasp start",
               "  wasp db migrate-dev",
               "",
--- a/waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
+++ b/waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
@@ -3,9 +3,15 @@ module Wasp.Cli.Command.CreateNewProject
   )
 where
 
+import Control.Concurrent (newChan)
+import Control.Monad (when)
 import Control.Monad.Except (throwError)
 import Control.Monad.IO.Class (liftIO)
+import Data.Functor ((<&>))
 import Data.List (intercalate)
+import Data.Maybe (isJust)
+import qualified Data.Text as T
+import GHC.IO.Exception (ExitCode (..))
 import Path.IO (copyDirRecur, doesDirExist)
 import StrongPath (Abs, Dir, Path, Path', System, parseAbsDir, reldir, relfile, (</>))
 import StrongPath.Path (toPathAbsDir)
@@ -14,21 +20,26 @@ import qualified System.FilePath as FP
 import Text.Printf (printf)
 import Wasp.Analyzer.Parser (isValidWaspIdentifier)
 import Wasp.Cli.Command (Command, CommandError (..))
+import Wasp.Cli.Command.Message (cliSendMessageC)
 import qualified Wasp.Data as Data
+import qualified Wasp.Generator.Job as J
+import Wasp.Generator.Job.Process (runNodeCommandAsJob)
+import qualified Wasp.Message as Msg
 import Wasp.Project (WaspProjectDir)
-import Wasp.Util (indent, kebabToCamelCase)
+import Wasp.Util (indent, kebabToCamelCase, whenM)
 import qualified Wasp.Util.IO as IOUtil
 import qualified Wasp.Util.Terminal as Term
 import qualified Wasp.Version as WV
 
 data ProjectInfo = ProjectInfo
   { _projectName :: String,
-    _appName :: String
+    _appName :: String,
+    _template :: Maybe String
   }
 
-createNewProject :: String -> Command ()
-createNewProject projectNameCandidate = do
-  projectInfo <- parseProjectInfo projectNameCandidate
+createNewProject :: [String] -> Command ()
+createNewProject newArgs = do
+  projectInfo <- parseProjectInfo newArgs
   createWaspProjectDir projectInfo
   liftIO $ printGettingStartedInstructions $ _projectName projectInfo
   where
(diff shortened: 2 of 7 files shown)


waspc/cli/exe/Main.hs
```diff
@@ -38,7 +38,7 @@ main :: IO ()
 main = withUtf8 . (`E.catch` handleInternalErrors) $ do
   args <- getArgs
   let commandCall = case args of
-        ["new", projectName] -> Command.Call.New projectName
+        ("new" : newArgs) -> Command.Call.New newArgs
```
Martinsos, line 41: To maintain a bit more semantics I would maybe go for:
```hs
("new" : projectName : args) -> Command.Call.New projectName args
```

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
```diff
@@ -59,17 +74,26 @@ parseProjectInfo name
...
-  if dirExists
-    then throwProjectCreationError $ show absWaspProjectDir ++ " is an existing directory"
-    else liftIO $ do
-      initializeProjectFromSkeleton absWaspProjectDir
-      writeMainWaspFile absWaspProjectDir projectInfo
+
+  when
+    dirExists
+    $ throwProjectCreationError
+    $ show absWaspProjectDir ++ " is an existing directory"
```
Martinsos:
```suggestion
  when dirExists $
    throwProjectCreationError $ show absWaspProjectDir ++ " is an existing directory"
```

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
```diff
@@ -14,21 +20,26 @@ import qualified System.FilePath as FP
...
+import Wasp.Util (indent, kebabToCamelCase, whenM)
 import qualified Wasp.Util.IO as IOUtil
 import qualified Wasp.Util.Terminal as Term
 import qualified Wasp.Version as WV
 
 data ProjectInfo = ProjectInfo
   { _projectName :: String,
-    _appName :: String
+    _appName :: String,
+    _template :: Maybe String
```
Martinsos, line 37: I found it a tiny bit weird that you put template here, I imagined ProjectInfo to be about the names, but on the other hand maybe this is fine! I don't have super great ideas how to keep it outside so I am ok with this. @sodic maybe you have an opinion on this?

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
```diff
@@ -59,17 +74,26 @@ parseProjectInfo name
...
-  if dirExists
-    then throwProjectCreationError $ show absWaspProjectDir ++ " is an existing directory"
-    else liftIO $ do
-      initializeProjectFromSkeleton absWaspProjectDir
-      writeMainWaspFile absWaspProjectDir projectInfo
+
+  when
+    dirExists
+    $ throwProjectCreationError $
+      show absWaspProjectDir ++ " is an existing directory"
```
Martinsos, line 84:
```suggestion
  when dirExists $
    throwProjectCreationError $ show absWaspProjectDir ++ " is an existing directory"
```

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
```diff
@@ -103,5 +127,71 @@ writeMainWaspFile waspProjectDir (ProjectInfo projectName appName) = IOUtil.writ
           "}"
         ]
 
+createProjectFromTemplate :: Path System Abs (Dir WaspProjectDir) -> ProjectInfo -> Command ()
+createProjectFromTemplate absWaspProjectDir ProjectInfo {_appName = appName, _projectName = projectName, _template = maybeTemplateName} = do
```
Martinsos, line 131: Interesting that this compiles, I would expect Haskell to require  ( ) around `ProjectInfo { ...}`.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
```diff
@@ -103,5 +127,71 @@ writeMainWaspFile waspProjectDir (ProjectInfo projectName appName) = IOUtil.writ
...
+      where
+        executeCmd = newChan >>= runNodeCommandAsJob dirWhereProjectIsCreated "npx" args J.Cli
+        args = ["giget@latest", templatePath, projectDir]
+
+    -- gitget doesn't fail if the template dir doesn't exist in the repo, so we need to check if the directory exists.
+    ensureTemplateWasFetched :: Command ()
+    ensureTemplateWasFetched = do
+      whenM
+        (liftIO $ IOUtil.isDirectoryEmpty absWaspProjectDir)
+        $ throwProjectCreationError "Are you sure that the template exists? 🤔 Check the list of templates here: https://github.com/wasp-lang/starters"
```
Martinsos, line 172:
```suggestion
      whenM (liftIO $ IOUtil.isDirectoryEmpty absWaspProjectDir) $
        throwProjectCreationError "Are you sure that the template exists? 🤔 Check the list of templates here: https://github.com/wasp-lang/starters"
```

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
```diff
@@ -103,5 +127,71 @@ writeMainWaspFile waspProjectDir (ProjectInfo projectName appName) = IOUtil.writ
...
+      let replacedContent =
+            foldl
+              (\acc (placeholder, value) -> T.replace (T.pack placeholder) (T.pack value) acc)
+              mainWaspFileContent
+              placeholders
+
+      IOUtil.writeFileFromText absMainWaspFile replacedContent
+      where
+        absMainWaspFile = absWaspProjectDir </> [relfile|main.wasp|]
+        placeholders =
```
Martinsos, line 187:
```suggestion
        replacements =
```
maybe better?

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
```diff
@@ -103,5 +127,71 @@ writeMainWaspFile waspProjectDir (ProjectInfo projectName appName) = IOUtil.writ
...
+          [ ("__waspAppName__", appName),
+            ("__waspProjectName__", projectName),
+            ("__waspVersion__", waspVersionStr)
+          ]
+
+waspVersionStr :: String
+waspVersionStr = "^%s" `printf` show WV.waspVersion
```
Martinsos, line 194: I think in theory this is not version, but something like version range? In our SemanticVersion.hs module this stuff is relatively clear so maybe we can understand it from there. So if we can make it better let's do it, if not then ok.

waspc/src/Wasp/Generator/Job.hs
```diff
@@ -28,4 +28,4 @@ data JobMessageData
 
 data JobOutputType = Stdout | Stderr deriving (Show, Eq)
 
-data JobType = WebApp | Server | Db deriving (Show, Eq, Ord, Bounded, Enum)
+data JobType = WebApp | Server | Db | Cli deriving (Show, Eq, Ord, Bounded, Enum)
```
Martinsos, line 31: Woah woah new job type here, in Generator :D. These job types are normally to indicate the context in which Job is running. I am not sure how much sense it makes to introduce CLI, what is that, what does it stand for?

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
```diff
@@ -103,5 +127,71 @@ writeMainWaspFile waspProjectDir (ProjectInfo projectName appName) = IOUtil.writ
...
+        -- gh: prefix means Github repo
+        waspTemplatesRepo = "gh:wasp-lang/starters"
+
+    fetchTemplate :: String -> String -> Path System Abs (Dir dir) -> Command ()
+    fetchTemplate templatePath projectDir dirWhereProjectIsCreated = do
+      liftIO executeCmd >>= \case
+        ExitSuccess -> return ()
+        ExitFailure _ -> throwProjectCreationError "Failed to create project from template."
+      where
+        executeCmd = newChan >>= runNodeCommandAsJob dirWhereProjectIsCreated "npx" args J.Cli
```
Martinsos, line 164: Actually let's not use `runNodeCommandAsJob` because that gets into Generator territory and we don't really need Generator here. But it is useful because it checks for npm and stuff right? Hm. Or maybe we should extract runNodeCommandAsJob out of Generator? Do we have a function runNodeCommand that doesn't run as a job? Should we create that one?

Martinsos (changes requested):

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
infomiho: I'll change it to `templateName` to be more inline with other info in there.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
infomiho: Agreed 👍

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
infomiho: It's defined like this:
```hs
waspVersion :: SV.Version
waspVersion = SV.Version (toEnum major) (toEnum minor) (toEnum patch)
  where
    DV.Version [major, minor, patch] _ = Paths_waspc.version
```

I'll call it `waspCompilerVersionStr`

infomiho: But we are adding `^` in front ... that's true

Issue #1114, opened 2023-04-07
Improve the error messaging on Login & Signup


infomiho:
Instead of showing the ugly `401` message, we should show a nicer "Your login credentials are invalid" or similar.

Martinsos:
In case when Username already exists, I got "Save failed".

![image](https://user-images.githubusercontent.com/1536647/230616450-7f6a8384-e309-4d34-9563-a62cba1b43a4.png)

PR #1120, opened 2023-04-11
Improves error messages in auth

infomiho:
Fixes https://github.com/wasp-lang/wasp/issues/1114

Screenshot 2023-04-11 at 11 17 57
Screenshot 2023-04-11 at 11 17 53
Screenshot 2023-04-11 at 11 17 39

### DIFF
--- a/waspc/data/Generator/templates/server/src/utils.js
+++ b/waspc/data/Generator/templates/server/src/utils.js
@@ -32,7 +32,7 @@ export const prismaErrorToHttpError = (e) => {
   if (e instanceof Prisma.PrismaClientKnownRequestError) {
     if (e.code === 'P2002') {
       return new HttpError(422, 'Save failed', {
-        message: `A record with the same ${e.meta.target.join(', ')} already exists.`,
+        message: `user with the same ${e.meta.target.join(', ')} already exists`,
         target: e.meta.target
       })
     } else {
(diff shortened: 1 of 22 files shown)


Martinsos (approved):
Looking good! I approved but still I open one/two questions below to consider a bit.

The only thing I am not completely sure about is the message "username is already taken". Two things I am not sure about here:
1. Is it ok to let them know that username is taken? In theory, we just exposed the fact that user with such username exists, which means they could use this to discover if somebody is user of a platform, which feels a bit like security issue. I wonder what others do about this. If it is username, then I guess there is no other option, you need to tell them that username is taken, sure. But if it is email, then an option is to pretend that all is good and say that you sent them an email, while not doing anything -> but uff that sounds terrible hm :D. Ok I am talking with myself here, but: it sounds to me like it is a bit of a security issue, but I don't see a way to go around it. I wonder what is the general recommendation for this.
2. You are actually returning and displaying directly the error from Prisma. Could this be an issue, what if that error reveals something about the database that we don't want to reveal?

waspc/data/Generator/templates/server/src/utils.js
```diff
@@ -32,7 +32,7 @@ export const prismaErrorToHttpError = (e) => {
   if (e instanceof Prisma.PrismaClientKnownRequestError) {
     if (e.code === 'P2002') {
       return new HttpError(422, 'Save failed', {
-        message: `A record with the same ${e.meta.target.join(', ')} already exists.`,
+        message: `user with the same ${e.meta.target.join(', ')} already exists`,
```
shayneczyzewski, line 35: Will this always be "user"? Couldn't we get the same error from a failed save in a different context/entity?

shayneczyzewski: Makes sense. I'd prob add a TODO so we remember 👍🏻

PR #1123, opened 2023-04-11
Fix full-stack type support errors

sodic:


waspc/data/Generator/templates/server/src/_types/index.ts
```diff
@@ -24,7 +24,7 @@ export type AuthenticatedAction<Entities extends _Entity[], Input, Output> =
 type AuthenticatedOperation<Entities extends _Entity[], Input, Output> = (
   args: Input,
   context: ContextWithUser<Entities>,
-) => Promise<Output>
+) => Output | Promise<Output>
```
sodic, line 27: We want to allow users to write Queries and Actions that aren't async (if they don't need to be).

waspc/data/Generator/templates/server/src/actions/_action.ts
```diff
@@ -7,7 +7,7 @@ import prisma from '../dbClient.js'
           consider in the future if it is worth removing this duplication. =}
 
 export default async function (args, context) {
-  return {= jsFn.importIdentifier =}(args as never, {
+  return ({= jsFn.importIdentifier =} as any)(args, {
```
sodic, line 10: Ugly, but this assertion allows users not to specify `args` and `context` in their query/action implementations (if they don't need them)/

infomiho: We had to do that for server setup function, so it's a useful hack we have to use sometimes 👍

infomiho (approved):

PR #1125, opened 2023-04-12
Add better intro docs

infomiho:
Merges the two intro pages "What is Wasp?" which talked about the DSL and "How it works" which had a nice diagram of the compiler. It is less focused on the DSL and talks about what Wasp can do for the developer.

I've moved the "Introduction" pages into the `introduction` folder.


web/docs/introduction/what-is-wasp.md
```diff
@@ -0,0 +1,174 @@
...
+---
+
+import ImgWithCaption from '../../blog/components/ImgWithCaption'
+
+We will give a brief overview of what Wasp is, how it works on a high level and when to use it.
+
+## Wasp is a tool to build modern web applications
+
+It is an opinionated way of building **full-stack web applications**. It takes care of all three
+major parts of a web application: **client** (front-end), **server** (back-end) and **deployment**.
```
shayneczyzewski, line 12: Should we include DB here too? Or is that implied in backend/server?

web/docs/introduction/what-is-wasp.md
```diff
@@ -0,0 +1,174 @@
...
+Define your app in the Wasp config and get:
+- login and signup with Auth UI components,
+- full-stack type safety,
+- e-mail sending,
+- async processing jobs,
+- React Query powered data fetching,
+- security best practices,
+- and more. 
+
+You don't need to write any code for these features, Wasp will take care of it for you 🤯
```
shayneczyzewski, line 42: I would maybe also consider adding that we continue to update/improve/maintain the "Wasp-owned code" over time, not just spit something out that _you_ have to then maintain after initial project generation. (aka a counterpunch at T3 and similar)

shayneczyzewski (approved):
These updates look and sound great to me @infomiho I personally approve but know others are more invested in this stuff, so I would wait for their 👍🏻 too :) Great job!

web/docs/introduction/what-is-wasp.md
infomiho: I've copy pasted this part from before. I guess the intention was to emphasise **deployment** as a thing you get with Wasp.

web/docs/introduction/what-is-wasp.md
infomiho: That's a good idea, I'll add it 👍

PR #1137, opened 2023-04-14
Docs intro fix

Martinsos:
Docs fix! I improved examples in the "Introductions" somewhat to be a bit more realistic / correct. I think they are still not too big, but possibly provide a bit more context / idea of how Wasp works.

Also some small fixes.

@infomiho let me know if this is ok.


infomiho (approved):
LGTM

Issue #1108, opened 2023-04-06
Improve our scaffolding experience


infomiho:
We are offering users a way to select a template to start their Wasp project from.

The user can write `wasp new <project-name> -t <template-name>` but they need to know about which templates they can use.

### Current situation
We are mentioning https://github.com/wasp-lang/starters as a place where they can take a look at the list of all available templates.

### What could we do better
We could maybe fetch the list of templates and output them in the CLI somehow. What would be even cooler, they can interactively select one of the options.

Issue #1134, opened 2023-04-13
Ensure that Wasp nicely warns if `node` is not installed / usable


Martinsos:
We have `runNodeCommandAsJobWithExtraEnv` which is used for running `node` commands and it checks that node version is correct. It should also check that `node` itself exists and return nice message if not. It does try to do that right now:
```hs
      `catchIOError` ( \e ->
                         if isDoesNotExistError e
                           then return (ExitFailure 1, "", "Command 'node' not found.")
                           else ioError e
                     )

```
but, we noticed that in some cases instead we get following behaviour (when `node` is not installed):

```
:bee: --- Starting compilation and setup phase. Hold tight... ------------------------


:bee: --- Compiling wasp project... --------------------------------------------------


:bee: --- Starting npm install... ----------------------------------------------------

wasp-bin: node: readCreateProcessWithExitCode: posix_spawnp: failed (Undefined error: 0)
```

@sodic got that behaviour on macbook (m2) without `node`.

Instead, we want a much nicer error message, something like "It seems that you don't have `node` installed! Wasp requires `node` to be installed and in PATH in order to run.".

craigmc08:
I think the behavior @sodic encountered is related to https://github.com/haskell/process/issues/251 (which has been fixed as of the latest version of `process`, v1.6.17.0). I suggest we increase the minimum version in `waspc` from 1.6.13 to 1.6.17 and see if that issue still happens. I don't have access to any m2 devices so I can't test this myself.

If that doesn't fix the error, when something that is not an IO error happens while running `node --version`, we could display that IO error *and* say that they should make sure node is installed. The output from the example you gave could instead end with something like

```txt
An unknown error occurred while running `node --version`:
    wasp-bin: node: readCreateProcessWithExitCode: posix_spawnp: failed (Undefined error: 0)
Check that `node` is installed and in PATH.
```

---

> So we should look into redefining Wasp's dependency on node, and as this issue says, ensure that we warn user clearly and on time if their node is not installed and provide some instructions.

We already have all places that use node going through `Wasp.Generator.Job.Process.runNodeCommandAsJob[WithExtraEnv]`, which checks node version. There is still potential for someone to go around that and run a node command through some ad-hoc system, but I'm not sure of a way to automatically prevent that.

[...]

PR #1140, opened 2023-04-17
Improve the scaffolding experience

infomiho:
Closes https://github.com/wasp-lang/wasp/issues/1108

- Uses `optparse-applicative` for CLI options (testing it on a smaller piece, since in the future we will rewrite commands in it)
- Adds an interactive mode with template options (`wasp new` without options)
- Adds loading of templates from the Github repo instead of hard coding it

https://user-images.githubusercontent.com/2223680/232472938-2edc71ce-acc9-4501-873b-9358f26f8129.mov


waspc/cli/src/Wasp/Cli/Question.hs
```diff
@@ -0,0 +1,86 @@
...
+import Text.Read (readMaybe)
+import qualified Wasp.Util.Terminal as Term
+
+queryNotNull :: String -> IO String
+queryNotNull question = do
+  ensureNotNull . query $ question
+
+queryWithOptions :: String -> [String] -> IO (Maybe String)
+queryWithOptions question options = do
+  putStrLn optionsStr
```
Martinsos, line 20: When I am reading this function, seeing just `optionsStr` makes it very unclear for me as to what this is.

waspc/cli/src/Wasp/Cli/Question.hs
```diff
@@ -0,0 +1,86 @@
...
+
+  handleEmptyAnswer answer
+    >>= \case
+      Just _ ->
+        handleOptionIndex answer
+          >>= \case
+            Just option -> return $ Just option
+            Nothing -> handleOptionName answer
+      -- Optional
+      Nothing -> return Nothing
```
Martinsos, line 32: So I haven't tried to understand it fully, but I think you can make this somewhat simpler by possibly making more functions pure + using the short circuiting capability of the Maybe Monad. I am not 100% sure, but it looks too convoluted at the moment.

If you want to give it a try on your own with this suggestion in mind, give it a try, and if you run out of ideas and it still looks complex, we can try it together.

waspc/cli/src/Wasp/Cli/Question.hs
```diff
@@ -0,0 +1,86 @@
+module Wasp.Cli.Question
```
Martinsos, line 1: Interesting, I like how you isolated this!

I think you wanted to name the module `Query`, not `Question`? Although even `Query` is a bit vague, but at least it seems to be more in line with the naming you use for the functions in the module.

So you created your module for queries toward user, that makes sense. I have a feeling that some advanced haskeller would probably make this a monad or applicative or something hehe :D, but we don't have to overdo it for now, I think this is a good start.

I did find it somewhat hard to read though, maybe we can clean it up somewhat?

Some ideas:
1. It is unusual to see functions that are not verbs. Can work, but a bit confusing.
2. `type Question = String` and `type Answer = String` could already help somewhat.
3. Make sure stuff is more in context / more readable by providing better names, better order / structure of functions, ... .

Martinsos (changes requested):
Exciting to see this happening, and good job on doing more Haskell work @infomiho ! I hope it was fun!

From the DX perspective:

0. I like the simplicity and colors, I think it is a good start!

1. I find it confusing how it right now lists all the templates first and then tells you to choose the template.
So right now it looks like this:
```
Enter the project name (e.g. my-project) ▸ MyTest
(1) saas
(2) waspello
Choose a template (optional)  ▸ <blink-blink>
Created new Wasp app in ./MyTest directory!
To run it, do:

    cd MyTest
    wasp start
```

Instead, could it be better to do something like this?
```
Enter the project name (e.g. my-project) ▸ MyTest

Choose the starting template:
(1) vanilla (just the basics)
(2) saas (basics + auth + stripe)
Enter the number (e.g. 1) ▸ 1

Created new Wasp app in ./MyTest directory!
To run it, do:

    cd MyTest
    wasp start
```

This immediately pulls the question of the optionality of choosing a template -> it is not explained what does it mean it is optional: what happens if I don't choose a template? It still does get templated from something, right? So does it make sense to just call that a "basic" or "vanilla" template? Or do we handle that as a special case?
Also, right now it is not very clear how to say "I am skipping" for this optional option.

The rest of the comments are mostly just "clean code" and haskell stuff (or mix of it).

shayneczyzewski:
Nice work so far @infomiho. I like how this will be able to quickly get people set up, and pave the way toward more options, like TS or JS, add auth, add jobs, etc.

Since Martin took the first stab at the code, I mainly just played with it for now.

1) I do think it would be nice to have a "basic"/"vanilla"/"standard"/"minimal" option that does what ours does now, and make the choice required when you do `wasp new` without a name.

2) I think we probably want to handle the case of the internet (or let's be real, GitHub) being down. When I turned my internet off, I got the following after about 5-10 seconds:

```sh
$ wasp-cli new 
wasp-cli: HttpExceptionRequest Request {
  host                 = "api.github.com"
  port                 = 443
  secure               = True
  requestHeaders       = [("Accept","application/json"),("User-Agent","wasp-lang/wasp")]
  path                 = "/repos/wasp-lang/starters/git/trees/main"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
  proxySecureMode      = ProxySecureWithConnect
}
 (ConnectionFailure Network.Socket.getAddrInfo (called with preferred socket type/protocol: AddrInfo {addrFlags = [AI_ADDRCONFIG], addrFamily = AF_UNSPEC, addrSocketType = Stream, addrProtocol = 0, addrAddress = 0.0.0.0:0, addrCanonName = Nothing}, host name: Just "api.github.com", service name: Just "443"): does not exist (Temporary failure in name resolution))
```

Maybe in this case we can just offer the standard creation, or just do it automatically?

3) Additionally, I noticed when the internet was down I could also not do `wasp new fooBar`. Can we make a named project bypass the template lookup? It seems like it doesn't require it and would be nice to not pay that penalty even if everything is working.

4) Can/do we want some way to have some templates show up in the CLI, but others in the repo not? For example, having `waspello` show up in the list of options feels kinda arbitrary to me. Should it maybe just not live under starters? Or can we add some metadata so the GitHub HTTP pull can filter it out? Not quite sure if that is possible though 🤷🏻‍♂️ 

Great stuff so far, will circle back to code tomorrow. 👍🏻

Martinsos:
Quick addition on top of what @shayneczyzewski said:
1. Ouf internet, ok that is tricky, I am not sure. I don't want to overcomplicate situation. Maybe we can just say they need internet? Or we warn them there is none and they can't choose templates and go with default one.
2. I guess waspello was there just as a placeholder, but yeah I would also remove it and make sure we have only a couple of standard starters that are well maintained.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/Options.hs
```diff
@@ -0,0 +1,95 @@
...
+import Wasp.Util (indent, kebabToCamelCase)
+
+data NewProjectArgs = NewProjectArgs
+  { _optsProjectName :: Maybe String,
+    _optsTemplateName :: Maybe String
+  }
+
+type Templates = [String]
+
+getProjectInfo :: Arguments -> Command ProjectInfo
```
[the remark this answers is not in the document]
infomiho: `Arguments` is just a list of any arguments since it's a `[String]`

I've rewritten this a bit.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/Options.hs
```diff
@@ -0,0 +1,95 @@
...
+  createProjectInfoFromArgs newProjectArgs templates
+
+createProjectInfoFromArgs :: NewProjectArgs -> Maybe Templates -> Command ProjectInfo
+createProjectInfoFromArgs (NewProjectArgs projectNameArg templateNameArg) maybeTemplates =
+  case (projectNameArg, templateNameArg, maybeTemplates) of
+    -- Non-interactive mode
+    (Just projectName, templateName, _) ->
+      createProjectInfo projectName templateName
+    -- Interactive mode
+    (_, _, Just templates) -> do
```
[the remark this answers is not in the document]
infomiho: Now it's called `createNewProjectDescription` and it handles the creation process based on `args` and list of `templates`. I didn't want to complicate since this is the central function that based on the available data decides either to create the new project description or interactively ask the user 🤷

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/Options.hs
```diff
@@ -0,0 +1,95 @@
...
+getNewProjectArgs newArgs = extractArgs $ execParserPure defaultPrefs argsParserInfo newArgs
+
+argsParserInfo :: Opt.ParserInfo NewProjectArgs
+argsParserInfo = Opt.info (argsParser Opt.<**> Opt.helper) Opt.fullDesc
+
+argsParser :: Opt.Parser NewProjectArgs
+argsParser = NewProjectArgs <$> Opt.optional projectNameParser <*> Opt.optional templateNameParser
+
+projectNameParser :: Opt.Parser String
+projectNameParser = Opt.strArgument $ Opt.metavar "PROJECT_NAME"
```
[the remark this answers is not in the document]
infomiho: That's the name that `optparse-applicative` prints in help and error messages when referring to this argument.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/Options.hs
```diff
@@ -0,0 +1,95 @@
...
+
+templateNameParser :: Opt.Parser String
+templateNameParser =
+  Opt.strOption $
+    Opt.long "template"
+      <> Opt.short 't'
+      <> Opt.metavar "TEMPLATE_NAME"
+      <> Opt.help "Template to use for the new project"
+
+extractArgs :: Opt.ParserResult NewProjectArgs -> Command NewProjectArgs
```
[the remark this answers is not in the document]
infomiho: > Opt parse is looking cool! What does it give us exactly?

By defining the options we want, we can compose them and (later) also use their commands. This gives us:
- validation
- automatic command listing
- help per command
- bash / zsh / fish completions 

It's a very powerful thing and it will level up our CLI interface and give the mentioned stuff above for free.

waspc/cli/src/Wasp/Cli/Question.hs
infomiho: I've stolen the naming from the `summoner` https://github.com/kowainik/summoner/blob/main/summoner-cli/src/Summoner/Question.hs

I'll clean a bit what I think it makes sense.

waspc/cli/src/Wasp/Cli/Question.hs
```diff
@@ -0,0 +1,86 @@
...
+      putStrLn $ Term.applyStyles [Term.Red] "This field cannot be empty."
+      ensureNotNull action
+    else return result
+
+prompt :: IO String
+prompt = do
+  putStrFlush $ Term.applyStyles [Term.Yellow] " ▸ "
+  T.unpack . T.strip . T.pack <$> getLine
+
+-- | Explicit flush ensures prompt messages are in the correct order on all systems.
```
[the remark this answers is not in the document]
infomiho: I've seen it here: https://github.com/kowainik/summoner/blob/main/summoner-cli/src/Summoner/Question.hs

waspc/cli/src/Wasp/Cli/Question.hs
infomiho: I'm not sure how to use
> using the short circuiting capability of the Maybe Monad
in this context, I'll push the code I have and if you can help me with it 🤷

infomiho:
@shayneczyzewski @Martinsos I've updated the code based on the comments, did a better split regarding the modules and some naming improvements.

- I've also handled the failing internet connection -> it falls back to the bundled template we had before. It's good enough IMHO.
- I've also added the "basic" template to the starters repo https://github.com/wasp-lang/starters/tree/main/basic
- I've also updated the output to make it easier to read
- The template is now a required field which has the first option as the default
- If you only provide the project name, it will ask you to pick a template


https://user-images.githubusercontent.com/2223680/232799878-601a6d6c-24ff-4423-8426-d997a8c17e2e.mov

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/Options.hs
```diff
@@ -0,0 +1,95 @@
...
+import Wasp.Cli.Command.CreateNewProject.Http (getTemplatesJson)
+import qualified Wasp.Cli.Question as Question
+import Wasp.Util (indent, kebabToCamelCase)
+
+data NewProjectArgs = NewProjectArgs
+  { _optsProjectName :: Maybe String,
+    _optsTemplateName :: Maybe String
+  }
+
+type Templates = [String]
```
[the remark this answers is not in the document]
Martinsos: Yup yup, but how are strings templates? Are they content of templates? Are they paths to templates? Are they template names? Could be anything!

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/Options.hs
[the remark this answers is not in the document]
Martinsos: Right, but you probably don't want to get any arguments here, you want some specific ones -> like all of the arguments from the `new` command -> so I wanted to emphaseize that should be documented, be it via comments or naming.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/Options.hs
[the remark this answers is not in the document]
Martinsos: Ok that sounds awesome, looking forward to having it for the whole CLI!

waspc/cli/src/Wasp/Cli/Question.hs
Martinsos: Well maybe they don't have the best naming :D. It is good to use them for inspiration, but we should be able to support it logically.

Query being a verb and noun -> ah sure, it is a verb also, but for me what made it sound like a noun in this context was the addition of `WithOptions` and `NotNull`. But I guess it depends on perspective hm. Still, not very clear from the name what the function does.

waspc/cli/src/Wasp/Cli/Question.hs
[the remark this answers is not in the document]
Martinsos: Hm, still it doesn't mean it is correct. Although Kowainik pplz are good.
I think they wanted to say what I wrote, just wrote the comment badly.

Martinsos (commented):
@infomiho I see you left some comments, but is it ready for new round o are you still working on it? Because I see you also haven't answered / emojied all of my comments. So I will be waiting till you give me explicit "go" (in the meantime I commented back on a couple of things though).

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/Options.hs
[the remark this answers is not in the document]
infomiho: That's a very valid point, I didn't realize that's what you were saying :)

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/Http.hs
```diff
@@ -0,0 +1,58 @@
+module Wasp.Cli.Command.CreateNewProject.Http
```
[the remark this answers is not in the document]
infomiho: I've rewritten this piece in `Templates.hs`, I hope it makes more sense now.

waspc/cli/src/Wasp/Cli/Question.hs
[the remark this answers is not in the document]
infomiho: Ah yes, that is true. I've updated the comment 👍

waspc/cli/src/Wasp/Cli/Question.hs
infomiho: I hope it's better with `Interactive` module name and `askForInput` and `askToChoose`

infomiho:
@Martinsos @shayneczyzewski  Ready for another round

shayneczyzewski:
Hey @infomiho nice work! I gave it another testing whirl, again leaving the code mostly alone to Martin's feedback. Nice improvements! 👏🏻 Here is what I came up with via WSL Ubuntu testing, if you want me to verify on MacOS just let me know.

- When I first run it asks me to install `giget`... is there a way we can hide that, or let `npx` or other solve that for us?
Screenshot 2023-04-19 101230

- When internet is disabled, but you supply a `-t <template>` arg, it just hangs. Can the timeout apply here too?
Screenshot 2023-04-19 113502

- I keep getting these node experimental warnings (I'm on 18.12 if that matters):
Screenshot 2023-04-19 113656

- Did we previously warn the user in a nicer way if the dir already existed before we tried to create the project? Probably not so we don't have to here, but since in this area it may be a nice addition if time allows.
Screenshot 2023-04-19 113756

shayneczyzewski:
> Comment regarding Waspello, it's just one of the starters in the repo, whatever we put in the repo, it will be listed. I think we can change this indepedent of this PR.

I agree, and would recommend we remove it before rolling this out 👍🏻

infomiho:
@shayneczyzewski I've fixed the first two things:
- `npx` shouldn't ask for permission (calling it with `npx --yes`
- network failures are better detected - if there are not templates, falls back to the default. If the user provided a template name, we warn him that we are having trouble fetching it and we are using the default template

As for the experimental warnings, I'm not sure. I'll investigate. 

I'll check for the dir existence as soon as possible and make it a bit nicer 👍

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/ProjectDescription.hs
```diff
@@ -0,0 +1,91 @@
...
+import Wasp.Cli.Terminal (asWaspWarningMessage)
+import Wasp.Util (indent, kebabToCamelCase)
+
+data NewProjectDescription = NewProjectDescription
+  { _projectName :: String,
+    _appName :: String,
+    _templateName :: Maybe String
+  }
+
+initNewProjectDescription :: Arguments -> Command NewProjectDescription
```
Martinsos, line 30: Again me with the name, but I am not sure what to think about `init`, it sounds like it is initializing something, and I don't see what initialization means in this context, it is not like we have a multi-stage thing here.

How about `obtainNewProjectDescription`? Sounds like there is some work to be done here, with the obtaining, and we are doing some work to at the end give you new project description.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/ProjectDescription.hs
```diff
@@ -0,0 +1,91 @@
...
+initNewProjectDescription newArgs = do
+  newProjectArgs <- parseNewProjectArgs newArgs
+  templates <- liftIO getStarterTemplates
```
Martinsos, line 33: Are these just template names? Here it looks like they are templates, but I realize from the createNewProjectDesription that it takes just names, so there are just names? Let's reflect that in the name? I guess the function `getStarterTemplates` should also reflect that in the name. I would probably rename it to `fetchStarterTemplateNames`.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/ProjectDescription.hs
```diff
@@ -0,0 +1,91 @@
...
+          waspWarns (asWaspWarningMessage "Could not download Wasp templates, using the default template.")
+      mkNewProjectDescription projectName Nothing
+  where
+    askProjectName :: Command String
+    askProjectName = liftIO $ Interactive.askForRequiredInput "Enter the project name (e.g. my-project)"
+
+    askTemplateName :: StarterTemplateNames -> Command (Maybe String)
+    askTemplateName templateNames = liftIO $ Interactive.askToChoose "Choose a starter template" $ templatesToList templateNames
+
+    ensureValidTemplateName :: String -> StarterTemplateNames -> Command (Maybe String)
```
Martinsos, line 72: Why Maybe here, in return type? Actually why do you return anything?

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/Templates.hs
```diff
@@ -0,0 +1,79 @@
...
+  )
+where
+
+import Control.Exception (try)
+import Data.Aeson (FromJSON, parseJSON, withObject, (.:))
+import qualified Network.HTTP.Simple as HTTP
+
+data StarterTemplateNames = StarterTemplateNames [String]
+
+getStarterTemplates :: IO (Maybe StarterTemplateNames)
```
Martinsos, line 15: I would document that Nothing means we failed to fetch templates.
Or, you might go a bit different, and let it throw exceptions, and then in the caller you can decide to drop them all if you want.
Also, would rename to `getStarterTemplatesNames`.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/Templates.hs
```diff
@@ -0,0 +1,79 @@
...
+    Right response -> extractTemplateNames . HTTP.getResponseBody $ response
+    Left (_ :: HTTP.HttpException) -> Nothing
+  where
+    templatesRepoInfoURL :: HTTP.Request
+    templatesRepoInfoURL = "https://api.github.com/repos/wasp-lang/starters/git/trees/main"
+
+    -- Each folder in the "wasp-lang/starters" repo is a template.
+    extractTemplateNames :: Either HTTP.JSONException RepoInfo -> Maybe StarterTemplateNames
+    extractTemplateNames (Left _) = Nothing
+    extractTemplateNames (Right body) = Just . StarterTemplateNames . map _path . filter isFolder $ _objects body
```
Martinsos, line 30: Weird that it takes Either and then returns Maybe where Left and Nothing are directly mapped. Then you probably just want a function that takes normal value and returns normal value and then the caller should handle case when there are no input values.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/Templates.hs
```diff
@@ -0,0 +1,79 @@
...
+
+    isFolder :: RepoObject -> Bool
+    isFolder = (== Folder) . _type
+
+templatesToList :: StarterTemplateNames -> [String]
+templatesToList (StarterTemplateNames templates) = templates
+
+isValidTemplateName :: String -> StarterTemplateNames -> Bool
```
Martinsos, line 38: Sounds like you are checking if it is correctly formatted to be a template name. But you are actually checking if it is one of the existing templates. Not terrible, but a bit misleading.

waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -0,0 +1,80 @@
+module Wasp.Cli.Interactive
```
Martinsos, line 1: Love the new naming! Interactive is pretty clear, and `askToChoose` and `askForInput` are great.

Martinsos (changes requested):
@infomiho most left more clean code suggestions!

General stuff:
1. Do we now have both "basic" template up there, among the templates, and also locally in the CLI then? If so, should we make sure we don't have that duplication and have just one? What would make sense to me is to keep the one we have locally embedded into the CLI, and not have the one up there, but when we offer them to choose a template, we offer them basic + whatever is up there, they don't need to know what is remote what is local.
2. When there is no internet, we should let them know they are missing out: We can tell them that we couldn't fetch templates, so instead we are using a basic template.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
```diff
@@ -153,13 +127,14 @@ createProjectFromTemplate absWaspProjectDir ProjectInfo {_appName = appName, _pr
         Right _ -> return ()
       where
         executeCmd = callCommand $ unwords command
-        command = ["npx", "giget@latest", templatePath, projectDir]
+        command = ["npx", "--yes", "giget@latest", templatePath, projectDir]
```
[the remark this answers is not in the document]
infomiho: > but we do need to handle it somehow here, that is for sure

I think we should rely on `node` being a requirement for even running Wasp CLI and we have an issue here: https://github.com/wasp-lang/wasp/issues/1134 

We can tackle this in a subsequent PR, which can go after this one :)

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/ArgumentsParser.hs
```diff
@@ -0,0 +1,43 @@
...
+          <> Opt.short 't'
+          <> Opt.metavar "TEMPLATE_NAME"
+          <> Opt.help "Template to use for the new project"
+
+    parserResultToNewProjectArgs :: Opt.ParserResult NewProjectArgs -> Command NewProjectArgs
+    parserResultToNewProjectArgs (Opt.Success success) = return success
+    parserResultToNewProjectArgs (Opt.Failure failure) = throwProjectCreationError $ show help
+      where
+        (help, _, _) = Opt.execFailure failure "wasp new"
+    parserResultToNewProjectArgs (Opt.CompletionInvoked _) = throwProjectCreationError "Completion invoked when parsing 'wasp new', but this should never happen"
```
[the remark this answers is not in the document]
infomiho: This function is called like this `parserResultToNewProjectArgs $ execParserPure ...`.

I've imagined it as being a "transform" function from the `optparse-applicative` parse result to `Command NewProjectArgs`

If we return here `Either` instead, we would need to do something like this when calling it
```
let newProjectArgsOrError = parserResultToNewProjectArgs $ execParserPure ...
case newProjectArgsOrError of
Right newProjectArgs -> return newProjectArgs
Left error -> throwProjectCreationError error
```

Which is not that bad, but I'm not sure which approach is more readable?

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/Common.hs
```diff
@@ -0,0 +1,24 @@
...
+import Wasp.Cli.Command (Command, CommandError (..))
+import Wasp.Project (WaspProjectDir)
+
+throwProjectCreationError :: String -> Command a
+throwProjectCreationError = throwError . CommandError "Project creation failed"
+
+throwInvalidTemplateNameUsedError :: Command a
+throwInvalidTemplateNameUsedError = throwProjectCreationError "Are you sure that the template exists? 🤔 Check the list of templates here: https://github.com/wasp-lang/starters"
+
+getAbsoluteWaspProjectDir :: String -> Command (Path System Abs (Dir WaspProjectDir))
```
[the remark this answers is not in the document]
infomiho: I've just moved this function to the `Common` module and didn't really implement it or read through it. So, sorry about that. I can rewrite it to make it more understandable, sure 👍 

It was moved from `CreateNewProject` to `CreateNewProject.Common` which means it was very local to the new project creation context. If it's not used anywhere else does it still make sense to move it to a more generic location?

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/ProjectDescription.hs
infomiho: `obtain` === `get` in my mind, and that's the name we had before. I've moved away from that on your suggestion that's not **only** getting stuff, but also doing something. So, `init` felt right since it's doing some prep work like parsing the args, getting the template names and then finally creating the `NewProjectDescription`

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/ProjectDescription.hs
infomiho: I thought it was clear from the type signature 
```hs
getStarterTemplates :: IO (Maybe StarterTemplateNames)
```
that it returned names 🤷 

I can adjust the name a bit more to make it more clear.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/Templates.hs
infomiho: Maybe we can do something like `data StarterTemplateNames = StarterTemplateNames [String] | NoTemplateNamesAvailable` and be explicit about the failure to fetch? 

I'm not sure if that's an overkill.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/Templates.hs
infomiho: > Then you probably just want a function that takes normal value and returns normal value and then the caller should handle case when there are no input values.

I'm not sure what you mean here exactly, could you give me an example?

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
```diff
@@ -125,11 +97,15 @@ writeMainWaspFile waspProjectDir (ProjectInfo projectName appName _) = IOUtil.wr
           "}"
         ]
 
-createProjectFromTemplate :: Path System Abs (Dir WaspProjectDir) -> ProjectInfo -> Command ()
-createProjectFromTemplate absWaspProjectDir ProjectInfo {_appName = appName, _projectName = projectName, _templateName = maybeTemplateName} = do
-  cliSendMessageC $ Msg.Start "Creating project from template..."
+createProjectFromTemplate :: Path System Abs (Dir WaspProjectDir) -> NewProjectDescription -> Command ()
+createProjectFromTemplate absWaspProjectDir NewProjectDescription {_appName = appName, _projectName = projectName, _templateName = maybeTemplateName} = do
```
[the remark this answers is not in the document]
infomiho: Sure, that's something we'd encounter in TS world as well.

```ts
// getName gets called for properly defined user (good)
function getName(user: User): string {
  return user.name;
}
if (!user) {
  throw ...
}
getName(user)

// vs.

// getName gets called even when user is not defined (bad)
function getName(user?: User): string {
  if (!user) {
    throw ...
  }
  return user.name;
}
// check is inside
getName(undefined)
```

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/ProjectDescription.hs
infomiho: Rewrote this a bit 👍

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/Templates.hs
infomiho: I've introduced a `StarterTemplateNamesFetchResult` type which can either be success or failure. Then it's the same as `Either` but with extra meaning.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/Common.hs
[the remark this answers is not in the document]
infomiho: I've moved the core of the logic into `Wasp.Cli.FileSystem` and used that in `CreateNewProject.Common` since I wanted to handle the failure as well.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/ProjectDescription.hs
infomiho: I've ended up axing this function, I've just pulled the dependency fetching / preparing out.

infomiho:
@Martinsos @shayneczyzewski when you have time, it's ready for another pass:
- split the templates into remote and local ones (only one local template - the one we had bundled before)
- better error handling if there is not internet
- checks for Node.js before executing `npx`
- hides `npx` command output (not really useful)
- general naming improvements

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/ProjectDescription.hs
```diff
@@ -0,0 +1,104 @@
...
+          let isTemplateNameArgProvided = isJust templateNameArg
+              isTemplateNameFound = isJust maybeTemplateName
+          when
+            (isTemplateNameArgProvided && not isTemplateNameFound)
+            throwInvalidTemplateNameUsedError
+
+        getOrAskTemplateName :: Maybe StarterTemplateName -> Command StarterTemplateName
+        getOrAskTemplateName = \case
+          Just templateName -> return templateName
+          Nothing -> liftIO $ Interactive.askToChoose "Choose a starter template" templateNames
```
shayneczyzewski, line 82: We don't have to do it for this round, but I'm just wondering if we are in non-interactive mode and they don't specify a template, should we just use the default? That would allow any existing scripts to run as-is.

As a reminder, we will also want to be sure we update all references to `wasp new foo` in the docs to supply a `-t basic` after them, as well as label this a breaking change IMO. 👍🏻

shayneczyzewski (commented):
Great work @infomiho! From a functionality perspective, I think things are pretty much wrapped up well! I gave it another go and only really found the `cabal install` thing, plus I am still getting the node experimental warning (see below).
Screenshot 2023-04-25 105718

As a few final reminders, I think we will want to:
- make sure to update the docs with any new parameters (like `wasp new example -t basic`) so it all still works as expected
- remove the waspello project from the templates repo
- consider marking this a breaking change (plus update changelog)

Other than that, I think functionally we are in good shape, and will leave it to you and Martin to close out any remaining comments/obtain final approval. 👍🏻 Nice work! 🥳

waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -0,0 +1,86 @@
...
+    optionsToChooseFrom = intercalate "\n" $ prependIndexToOption options
+    prependIndexToOption = zipWith (\i o -> showIndexWithStyle i ++ show o ++ noteForDefault i) indexes
+    noteForDefault i = if i == 1 then " (default)" else ""
+    showIndexWithStyle i = Term.applyStyles [Term.Yellow] $ "[" ++ show (i :: Int) ++ "] "
+    indexes = [1 .. length options]
+
+askForInput :: String -> IO String
+askForInput question = putStr (Term.applyStyles [Term.Bold] question) >> prompt
+
+ensureNotNull :: Foldable t => IO (t a) -> IO (t a)
```
Martinsos, line 68: I would rename this one to `repeatIfNull`, since it will repeat the IO action until it returns result that is not null. I didnt' expec this behaviour from the name but I read it from the implementation.

Another way to go about this is to implement a bit more general function, smth like
```hs
repeatUntil :: (a -> Bool) -> String -> IO a -> IO a
repeatUntil predicate errMsg action = ...
```

which you can then call pretty nicely from `askForRequiredInput`.

waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -0,0 +1,86 @@
...
+      putStrLn $ Term.applyStyles [Term.Red] "Invalid selection, write the name or the index of the option."
+      askToChoose question options
+
+    findElemByString option = find (\o -> show o == option) options
+
+    optionsToChooseFrom = intercalate "\n" $ prependIndexToOption options
+    prependIndexToOption = zipWith (\i o -> showIndexWithStyle i ++ show o ++ noteForDefault i) indexes
+    noteForDefault i = if i == 1 then " (default)" else ""
+    showIndexWithStyle i = Term.applyStyles [Term.Yellow] $ "[" ++ show (i :: Int) ++ "] "
+    indexes = [1 .. length options]
```
Martinsos, line 63: I ended up going to town with this one. I made it as "haskellish" as possible, check it out! It wasn't so bad before, but, now it should look more declarative, utilize maybes more, and functions are doing less, so all together should be simpler / nicer. It also uses a couple of a bit advanced haskell concepts (not too advanced, but not super beginner), so check those out also, they are useful to know and easy to learn.

```hs
askToChoose :: forall o. Show o => String -> NonEmpty o -> IO o
askToChoose _ (option :| []) = return option
askToChoose question options = do
  putStrLn $ Term.applyStyles [Term.Bold] question
  putStrLn showIndexedOptions
  answer <- prompt
  getOptionMatchingAnswer answer & maybe printErrorAndAskAgain return
  where
    getOptionMatchingAnswer :: String -> Maybe o
    getOptionMatchingAnswer "" = Just defaultOption
    getOptionMatchingAnswer answer =
      getOptionByIndex answer `mplus` getOptionByName answer

    getOptionByIndex :: String -> Maybe o
    getOptionByIndex idxStr =
      case readMaybe idxStr of
        Just idx | idx >= 1 && idx <= length options -> Just $ options NE.!! (idx - 1)
        _invalidIndex -> Nothing

    getOptionByName :: String -> Maybe o
    getOptionByName name = find ((== name) . show) options

    printErrorAndAskAgain = do
      putStrLn $ Term.applyStyles [Term.Red] "Invalid selection, write the name or the index of the option."
      askToChoose question options

    showIndexedOptions :: String
    showIndexedOptions =
      intercalate "\n" $ showIndexedOption <$> zip [1 ..] (NE.toList options)
      where
        showIndexedOption (idx, option) =
          showIndex idx <> " " <> show option <> (if isDefaultOption option then " (default)" else "")
        showIndex i = Term.applyStyles [Term.Yellow] $ "[" ++ show (i :: Int) ++ "]"

    defaultOption :: o
    defaultOption = NE.head options

    isDefaultOption :: o -> Bool
    isDefaultOption option = show option == show defaultOption
```

waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -0,0 +1,86 @@
...
+    askUserToChoose = do
+      putStrLn $ Term.applyStyles [Term.Bold] question
+      putStrLn optionsToChooseFrom
+      answer <- prompt
+
+      handleOptionIndex answer >>= \case
+        Just option -> return option
+        Nothing -> handleOptionName answer
+
+    handleOptionIndex answer = do
```
Martinsos, line 34: handleOptionIndex is not the best name -> how does it handle it, what does it do?

waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -0,0 +1,86 @@
...
+        -- The default option is the first one
+        parseIndexOrDefault input = if null input then Just 1 else readMaybe input :: Maybe Int
+
+    handleOptionName answer = do
+      let option = findElemByString answer
+      case option of
+        Just o -> return o
+        Nothing -> invalidOptionAction
+
+    invalidOptionAction = do
```
Martinsos, line 53: If it is action, it is better to say what it does: `printErrorAndAskToChooseAgain` or smth like that.

waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -0,0 +1,86 @@
...
+import qualified Data.Text as T
+import System.IO (hFlush, stdout)
+import Text.Read (readMaybe)
+import qualified Wasp.Util.Terminal as Term
+
+askForRequiredInput :: String -> IO String
+askForRequiredInput question = do
+  ensureNotNull . askForInput $ question
+
+askToChoose :: Show a => String -> [a] -> IO a
```
Martinsos, line 19: NOTE: I left some comments below, but ended up writing a whole new implementation fro this function for you to check out, as we discussed, so check the comments on this functions for idea what I thought can be improved, and then at the end of the function see the new implementation that should improve that stuff + some more.

waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -0,0 +1,86 @@
...
+
+      case index of
+        Just i -> do
+          if i `elem` indexes
+            then return $ Just $ options !! (i - 1)
+            else Just <$> invalidOptionAction
+        Nothing -> return Nothing
+      where
+        -- The default option is the first one
+        parseIndexOrDefault input = if null input then Just 1 else readMaybe input :: Maybe Int
```
Martinsos, line 45: The notion that default options is the first option repeats, because you reason about it also when adding " (default)" to option when printing it, so that is duplication that should be removed.

waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -0,0 +1,86 @@
...
+        Nothing -> invalidOptionAction
+
+    invalidOptionAction = do
+      putStrLn $ Term.applyStyles [Term.Red] "Invalid selection, write the name or the index of the option."
+      askToChoose question options
+
+    findElemByString option = find (\o -> show o == option) options
+
+    optionsToChooseFrom = intercalate "\n" $ prependIndexToOption options
+    prependIndexToOption = zipWith (\i o -> showIndexWithStyle i ++ show o ++ noteForDefault i) indexes
```
Martinsos, line 60: Probably should be named `prependIndexesToOptions` (I added "s" at the end).
It returns String though, that is not very clear.

Martinsos (commented):
@infomiho did another round of review!

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/ProjectDescription.hs
infomiho: I think this is a valid suggestion 👍 if they run `wasp new` only then it's interactive, if they do `wasp new <name>` it shouldn't go into the interactive mode

waspc/data/Cli/templates/basic/main.wasp
```diff
@@ -0,0 +1,11 @@
+app __waspAppName__ {
```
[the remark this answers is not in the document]
infomiho: I've renamed the folder from `new` to `basic`, I'll fix this 👍

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/Common.hs
```diff
@@ -0,0 +1,130 @@
+module Wasp.Cli.Command.CreateNewProject.StarterTemplates.Common
```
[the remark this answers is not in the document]
infomiho: Agreed

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/Remote.hs
```diff
@@ -0,0 +1,46 @@
...
+  liftIO $ replaceTemplatePlaceholdersInWaspFile appName projectName absWaspProjectDir
+  where
+    getURLToGithubTemplate :: String -> Command String
+    getURLToGithubTemplate tmplName = return $ waspTemplatesRepo ++ "/" ++ templateFolderName
+      where
+        templateFolderName = tmplName
+        -- gh: prefix means Github repo
+        waspTemplatesRepo = "gh:wasp-lang/starters"
+
+    fetchGithubTemplateToDisk :: Path' Abs (Dir WaspProjectDir) -> String -> Command ()
```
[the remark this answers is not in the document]
infomiho: It will create the folder if it doesn't exist 🙂

waspc/cli/src/Wasp/Cli/Interactive.hs
infomiho: I've used `<|>` instead of `mplus` since it makes the code more clear for me. Great changes :)

waspc/cli/src/Wasp/Cli/GithubRepo.hs
```diff
@@ -0,0 +1,129 @@
...
+type GithubRepo = (GithubRepoOwner, GithubRepoName)
+
+type GithubRepoOwner = String
+
+type GithubRepoName = String
+
+fetchFolderFromGithubRepoToDisk ::
+  forall d.
+  GithubRepo ->
+  String ->
```
Martinsos, line 36: So `pathToFolderInRepo` is path but you used `String`.
If I got it right, we know it is relative path to directory -> I believe I saw you using `SP.parseRelDir` in one place, so obviously you expect it to be that.

You actually use it in two places total:
1. In first place you expect it to be rel dir path on the disk. So we can do `Path' Rel' Dir'`.
2. In second place you need it to be Posix, because you are using it URL. Also relative, and also dir. So `Path Posix Rel' Dir'`.

So I would suggest that in fetchFolderFromGithubRepoToDisk you say it is `Path Posix Rel' Dir'`, or maybe you can use `Path Posix (Rel a) (Dir b)` to give whoever is passing this URL a bit more freedom. But it should be Posix, because it is `pathToFolderInRepo`, online repo, so URLs are the thing online, and Posix usually makes more sense then.

Then, in your first place where you use it, you can do `relDirToSystem`, and later in the second place you can do `fromRelDirP`, since you need posix there.

Ah crap I just realized I don't have `relDirToSystem` in StrongPath hm. I have `relDirToPosix`, but I don't have `relDirToWindows` or `relDirToSystem` (ther is issue for those on strong path gh repo). The trick is you can easily switch rel path from one standard to another because you just need to swap all separators with correct separator.

Well, there are a couple of options then:
1. Update StrongPath to have `relDirToSystem` :D (but you also need to publish it which you can't yeah, blah).
2. Implement it here in Utils or somewhere, by importing StrongPath.Internal, I think that should work. Later we can transfer it to actual StrongPath library.
3. Type it as `Path System (Rel a) (Dir b)` for now, and turn it into Posix in that place where you need Posix. Leave a TODO saying that this should semantically be better to be typed as Posix, but well at the end it doesn't make a big difference.

If any of this stuff is a bit confusing, ask @sodic for help, he has seen enough StrongPath to become a StrongPath expert.

waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -0,0 +1,109 @@
...
+    isDefaultOption :: o -> Bool
+    isDefaultOption option = showOption option == showOption defaultOption
+
+askForInput :: String -> IO String
+askForInput question = putStr (Term.applyStyles [Term.Bold] question) >> prompt
+
+repeatIfNull :: Foldable t => IO (t a) -> IO (t a)
+repeatIfNull action = repeatUtil null "This field cannot be empty." action
+
+repeatUtil :: (a -> Bool) -> String -> IO a -> IO a
```
Martinsos, line 91:
```suggestion
repeatUntil :: (a -> Bool) -> String -> IO a -> IO a
```

waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -0,0 +1,109 @@
...
+askToChoose question options = do
+  putStrLn $ Term.applyStyles [Term.Bold] question
+  putStrLn showIndexedOptions
+  answer <- prompt
+  getOptionMatchingAnswer answer & maybe printErrorAndAskAgain return
+  where
+    getOptionMatchingAnswer :: String -> Maybe o
+    getOptionMatchingAnswer "" = Just defaultOption
+    getOptionMatchingAnswer answer =
+      getOptionByIndex answer <|> getOptionByName answer
```
Martinsos, line 56: Nice

waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -0,0 +1,109 @@
...
+
+{-
+  Why are we doing this?
+
+  Using a list of Strings for options results in the Strings being wrapped in quotes
+  when printed. We want to avoid this so users can just type the name of the option
+  without having to type the quotes as well.
+
+  We do this by overriding the default Show instance for Strings with our own instance
+  that just returns the String itself.
```
Martinsos, line 31: We don't override the default Show instance, we override the Option instance.

I would maybe emphasize that to solve the thing with Strings in quotes we introduced Option, which has default instance for every type implementing Show by just using Show, and only for String we do sometihng a bit different.

waspc/cli/src/Wasp/Cli/GithubRepo.hs
```diff
@@ -0,0 +1,129 @@
...
+
+    -- Which point in repo history to download (a branch or commit hash)
+    repoReference :: String
+    repoReference = "main"
+
+    downloadFile :: String -> Path' Abs (Dir d) -> String -> IO (Path' Abs (File f))
+    downloadFile downloadUrl destinationDir fileName = do
+      let destinationPath = destinationDir </> fromJust (SP.parseRelFile fileName)
+      simpleHttp downloadUrl >>= BL.writeFile (SP.fromAbsFile destinationPath)
+      return destinationPath
```
Martinsos, line 60: It feels here like this function could have easily received the destinationPath already assembled, instead of assembling it on its own and then returning it. Seems like a responsibility that it got accidentally by code evolution, not because it should have it really. Yeah, I would assemble it outside and let it receive one argument less, also wouldn't return anything.

waspc/cli/src/Wasp/Cli/GithubRepo.hs
```diff
@@ -0,0 +1,129 @@
...
+  where
+    downloadFolderToDisk :: Path' Abs (Dir d) -> IO ()
+    downloadFolderToDisk tempDir = do
+      downloadFile githubArchiveUrl tempDir archiveFileName >>= unpackArchive tempDir
+      copyRepoFolderToDestination tempDir pathToFolderInRepo destinationOnDisk
+      where
+        githubArchiveUrl = intercalate "/" ["https://github.com", githubOrg, repoName, "archive", archiveFileName]
+        archiveFileName = repoReference ++ ".tar.gz"
+
+    -- Which point in repo history to download (a branch or commit hash)
```
Martinsos, line 52: Ok nice explanation, important!

```suggestion
    -- Which point in repo history to download (a branch or commit hash).
```

waspc/cli/src/Wasp/Cli/GithubRepo.hs
```diff
@@ -0,0 +1,129 @@
...
+    repoReference :: String
+    repoReference = "main"
+
+    downloadFile :: String -> Path' Abs (Dir d) -> String -> IO (Path' Abs (File f))
+    downloadFile downloadUrl destinationDir fileName = do
+      let destinationPath = destinationDir </> fromJust (SP.parseRelFile fileName)
+      simpleHttp downloadUrl >>= BL.writeFile (SP.fromAbsFile destinationPath)
+      return destinationPath
+
```
Martinsos, line 61: Maybe you can document just a bit -> what will it do in that destination dir, will it unpack the contents in it directly, or wrap them all in some kind of dir maybe instead?

waspc/cli/src/Wasp/Cli/GithubRepo.hs
```diff
@@ -0,0 +1,129 @@
...
+    unpackArchive :: Path' Abs (Dir d) -> Path' Abs (File f) -> IO ()
+    unpackArchive destinationDir pathToTheTarFile = do
+      Tar.unpack (SP.fromAbsDir destinationDir) . Tar.read . GZip.decompress =<< BL.readFile (SP.fromAbsFile pathToTheTarFile)
+
+    copyRepoFolderToDestination :: Path' Abs (Dir d) -> String -> Path' Abs (Dir d) -> IO ()
+    copyRepoFolderToDestination archiveRootDir pathToFolderInRepo' destinationOnDisk' = do
+      copyDirRecur (toPathAbsDir pathToDownloadededFolder) (toPathAbsDir destinationOnDisk')
+      where
+        pathToDownloadededFolder = archiveRootDir </> rootArchiveFolder </> fromJust (SP.parseRelDir pathToFolderInRepo')
+        -- Github repo tars have a root folder that is named after the repo name and the reference (branch or tag).
```
Martinsos, line 71: Ok, this is knowledge specific to the fact that the repo folder was obtained from unarchiving a GH repo archive.
But function itself in its name doesn't specify that assumption. it does use `archiveRootDir` for variable name so that helps a tiny bit, but not enough I think.

Idea: rename this function to `getPathToFolderInGhArchive` or smth like that and it should return path, and then when you need it, you can directly call `copyDirRecur` on it.

Or go a bit further and make these functions you have in `where` more prominent and better named and again, where you need this, directly use those + `copyDirRecur`.

My point is that this function probably shouldn't exist, it is too specific and assumes a lot and is hard to name therefore, it is better to break into smaller pieces, make sure those are named very well, and compose them as needed and when needed. That should result with the most readable code + it will be composable.

waspc/cli/src/Wasp/Cli/GithubRepo.hs
```diff
@@ -0,0 +1,129 @@
...
+        rootArchiveFolder = fromJust (SP.parseRelDir $ repoName ++ "-" ++ repoReference)
+
+    withTempDir :: FilePath -> (Path' Abs (Dir d) -> IO a) -> IO a
+    withTempDir dirName action = do
+      tempDir <- getUserCacheDir <&> (</> fromJust (SP.parseRelDir dirName))
+
+      createDirectoryIfMissing True $ SP.fromAbsDir tempDir
+      result <- action tempDir
```
Martinsos, line 79: It can be tricky if exception is thrown here, you will never do clean up then.
you should catch it and in "finally" make sure you do cleanup (removing tmp dir).
There is no finally in Haskell, but there is a family of functions called `bracket` that play that role, so use that. Search through codebase for the usage example, and you will probably want to use the same import for it.

waspc/cli/src/Wasp/Cli/GithubRepo.hs
```diff
@@ -0,0 +1,129 @@
...
+
+    copyRepoFolderToDestination :: Path' Abs (Dir d) -> String -> Path' Abs (Dir d) -> IO ()
+    copyRepoFolderToDestination archiveRootDir pathToFolderInRepo' destinationOnDisk' = do
+      copyDirRecur (toPathAbsDir pathToDownloadededFolder) (toPathAbsDir destinationOnDisk')
+      where
+        pathToDownloadededFolder = archiveRootDir </> rootArchiveFolder </> fromJust (SP.parseRelDir pathToFolderInRepo')
+        -- Github repo tars have a root folder that is named after the repo name and the reference (branch or tag).
+        rootArchiveFolder = fromJust (SP.parseRelDir $ repoName ++ "-" ++ repoReference)
+
+    withTempDir :: FilePath -> (Path' Abs (Dir d) -> IO a) -> IO a
```
Martinsos, line 74: Nice one!

waspc/cli/src/Wasp/Cli/GithubRepo.hs
```diff
@@ -0,0 +1,129 @@
...
+
+type GithubRepoName = String
+
+fetchFolderFromGithubRepoToDisk ::
+  forall d.
+  GithubRepo ->
+  String ->
+  Path' Abs (Dir d) ->
+  IO (Either String ())
+fetchFolderFromGithubRepoToDisk (githubOrg, repoName) pathToFolderInRepo destinationOnDisk = do
```
Martinsos, line 39: This whole function, well its helper functions in `where`, are a bit convoluted I think -> there are some assumptions going around, especially around the archive, what it looks like, ... . I commented on some specific things but there might be some more that I just couldn't pinpoint yet, so I suggest trying to look into it very carefully, looking for those assumptions, trying to make every function as standalone, readable and simple as possible, while being explicit about what it depends on, what it outputs, ... . So that data flow is clear.

waspc/src/Wasp/Util/Node/Version.hs
```diff
@@ -0,0 +1,79 @@
+module Wasp.Util.Node.Version
```
Martinsos, line 1: Reminder to revert this.

waspc/src/Wasp/Util/Node/Command.hs
```diff
@@ -0,0 +1,32 @@
+module Wasp.Util.Node.Command where
```
Martinsos, line 1: Reminder to revert this.

Martinsos (commented):
This is now looking really nice! There are still some things that can possibly be improved, but in general it is very close to being done I think!

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
```diff
@@ -3,185 +3,64 @@ module Wasp.Cli.Command.CreateNewProject
...
-      putStrLn $ Term.applyStyles [Term.Bold] ("    cd " ++ projectFolder)
-      putStrLn $ Term.applyStyles [Term.Bold] "    wasp start"
+-- It receives all of the arguments that were passed to the `wasp new` command.
+createNewProject :: Arguments -> Command ()
+createNewProject args = do
+  newProjectArgs <-
+    parseNewProjectArgs args
+      & \case
+        Right newProjectArgs -> return newProjectArgs
+        Left err -> throwProjectCreationError err
```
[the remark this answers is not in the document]
infomiho: The old `&` and `either` trick! Cool stuff

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
```diff
@@ -3,185 +3,64 @@ module Wasp.Cli.Command.CreateNewProject
...
-waspVersionBounds :: String
-waspVersionBounds = show (SV.backwardsCompatibleWith WV.waspVersion)
-
-throwProjectCreationError :: String -> Command a
-throwProjectCreationError = throwError . CommandError "Project creation failed"
+    printGettingStartedInstructions :: NewProjectName -> IO ()
+    printGettingStartedInstructions (NewProjectName projectName) = do
+      let projectFolder = projectName
+{- ORMOLU_DISABLE -}
+      putStrLn $ Term.applyStyles [Term.Green] $ "Created new Wasp app in ./" ++ projectFolder ++ " directory!"
```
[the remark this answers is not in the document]
infomiho: > You could just use basename to get the name of the dir, but then you are assuming that project dir is in current directory :D, which is not such a terrible assumption, but it is still an assumption.

Since we are constructing the path to the project from a `folderName` and prepending the `cwd` to it, I would say it holds.

It feels a bit weird to pass in the `projectName` inside and get the `abs path` out. Which we then parse to get the relative dir to the `cwd` but ... we know how we constructed the `abs path`.

Alternatively, we could return the `abs path` + `relative path` (or something like `projectDirInCwd`) in `ProjectDescription` 🤷

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/ProjectDescription.hs
```diff
@@ -0,0 +1,141 @@
...
+obtainNewProjectDescriptionInteractively :: Maybe String -> [StarterTemplateName] -> Command NewProjectDescription
+obtainNewProjectDescriptionInteractively templateNameArg availableTemplates = do
+  projectName <- liftIO $ Interactive.askForRequiredInput "Enter the project name (e.g. my-project)"
+  absWaspProjectDir <- validateAndGetAbsProjectDir projectName
+
+  selectedTemplate <- maybe (liftIO askForTemplateName) findTemplateNameByStringOrThrow templateNameArg
+  mkNewProjectDescription projectName absWaspProjectDir selectedTemplate
+  where
+    findTemplateNameByStringOrThrow templateName = findTemplateNameByString availableTemplates templateName & maybe throwInvalidTemplateNameUsedError return
+    askForTemplateName = Interactive.askToChoose "Choose a starter template" $ fromList availableTemplates
```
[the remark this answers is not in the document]
infomiho: Nice rewrite!

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/ProjectDescription.hs
```diff
@@ -0,0 +1,141 @@
...
+        throwError err = throwProjectCreationError $ "Failed to get absolute path to Wasp project dir: " ++ show err
+    ensureProjectDirDoesNotExist :: String -> Path' Abs (Dir WaspProjectDir) -> Command ()
+    ensureProjectDirDoesNotExist projectDirName absWaspProjectDir = do
+      dirExists <- doesDirExist $ toPathAbsDir absWaspProjectDir
+
+      when dirExists $
+        throwProjectCreationError $
+          "Directory \"" ++ projectDirName ++ "\" is not empty."
+
+throwIfUserProvidedInvalidTemplateNameViaArgs :: Maybe String -> Maybe StarterTemplateName -> Command ()
```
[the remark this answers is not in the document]
infomiho: Agreed, the function felt bad, but I couldn't see the forrest from the tree 😄

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/Github.hs
```diff
@@ -0,0 +1,6 @@
+module Wasp.Cli.Command.CreateNewProject.StarterTemplates.Github where
```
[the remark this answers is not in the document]
infomiho: Makes sense 👍

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/Common.hs
```diff
@@ -0,0 +1,29 @@
...
+getAbsPathToNewProjectDirInCwd projectDir = do
+  absPathOrError <- liftIO $ getAbsPathToDirInCwd projectDir
+
+  case absPathOrError of
+    Right absPathToNewProjectInCwd ->
+      return absPathToNewProjectInCwd
+    Left err ->
+      throwProjectCreationError $ "Failed to get absolute path to Wasp project dir: " ++ show err
+
+waspVersionBounds :: String
```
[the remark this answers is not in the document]
infomiho: I thought I changed that, maybe some code moving around happened. Sorry about that, I'll go with the `defaultWaspVersionBounds` since it makes sense to me 👍

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/ProjectDescription.hs
```diff
@@ -0,0 +1,141 @@
...
+    - Template name is optional, if not provided, we use the default template.
+  2. Interactively
+
+     wasp new
+
+    - Project name is required.
+    - Template name is required, we ask the user to choose from available templates.
+-}
+obtainNewProjectDescription :: NewProjectArgs -> [StarterTemplateName] -> Command NewProjectDescription
+obtainNewProjectDescription (NewProjectArgs projectNameArg templateNameArg) starterTemplateNames =
```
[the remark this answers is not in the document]
infomiho: Very valid concern and I agree 👍 

Updated the code

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/ProjectDescription.hs
```diff
@@ -0,0 +1,141 @@
...
+  projectName <- liftIO $ Interactive.askForRequiredInput "Enter the project name (e.g. my-project)"
+  absWaspProjectDir <- validateAndGetAbsProjectDir projectName
+
+  selectedTemplate <- maybe (liftIO askForTemplateName) findTemplateNameByStringOrThrow templateNameArg
+  mkNewProjectDescription projectName absWaspProjectDir selectedTemplate
+  where
+    findTemplateNameByStringOrThrow templateName = findTemplateNameByString availableTemplates templateName & maybe throwInvalidTemplateNameUsedError return
+    askForTemplateName = Interactive.askToChoose "Choose a starter template" $ fromList availableTemplates
+
+validateAndGetAbsProjectDir :: String -> Command (Path' Abs (Dir WaspProjectDir))
```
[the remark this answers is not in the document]
infomiho: Makes sense 👍 

I like the notion of "available folder" since it signals "we are checking if it's available".

waspc/cli/src/Wasp/Cli/GithubRepo.hs
infomiho: I've rewritten the file quite a bit to make it more simple and all the functions are more agnostic.

waspc/cli/src/Wasp/Cli/GithubRepo.hs
infomiho: We agreed to use a thrid-party lib that handles this 👍

waspc/src/Wasp/Util/Node/Version.hs
infomiho: Moved this to `Wasp.Generator.Node.Version`

waspc/src/Wasp/Util/Node/Command.hs
infomiho: Removed this

waspc/cli/src/Wasp/Cli/GithubRepo.hs
infomiho: We went through this offline and it should be much better

waspc/cli/src/Wasp/Cli/GithubRepo.hs
infomiho: I've named it `targetRepoFolderName` now and it is representing a folder name.

infomiho: I've updated the function to accept a StrongPath

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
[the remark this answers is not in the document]
infomiho: I've used `basename` to get the folder name from the abs path and wrote down the assumption above the function definition. 

I did implement the "get relative path to folder from cwd" but I didn't like how it all looked like. So I went with the simplest option IMHO.

PR #1146, opened 2023-04-18
Adds building of `examples/todoApp` to CI

shayneczyzewski:
### Description

Adds building of `examples/todoApp` to CI. This should help catch any TS errors.

Example of a build failure: https://github.com/wasp-lang/wasp/actions/runs/4734719184/jobs/8404025469 (this was what would have happened on `main` before the fix was made).

### DIFF
--- a/waspc/tools/ensure_todoapp_builds.sh
+++ b/waspc/tools/ensure_todoapp_builds.sh
@@ -22,6 +22,6 @@ echo "Ensure client builds"
 cd .wasp/build/web-app
 npm run build
 
-echo "Ensure server builds""
+echo "Ensure server builds"
 cd ../server
 npm run build
--- a/waspc/tools/ensure_todoapp_builds.sh
+++ b/waspc/tools/ensure_todoapp_builds.sh
@@ -1,0 +1,27 @@
+#!/bin/sh -e
+
+# Compiles the todoApp and checks `npm run build` on both client and server.
+
+# Gets the directory of where this script lives.
+dir=$(CDPATH= cd -- "$(dirname -- "$0")" && pwd)
+echo "Running ensure_todoapp_builds.sh from $dir"
+
+cd "$dir/../examples/todoApp"
+echo "cd into $(pwd)"
+
+# Compile example app.
+WASP_BINARY_PATH="$(cabal list-bin wasp-cli)"
+CABAL_PROJECT_ROOT_PATH="$(cabal list-bin wasp-cli | sed s/\\/dist-newstyle.*//)"
+WASP_COMMAND="waspc_datadir=$CABAL_PROJECT_ROOT_PATH/data $WASP_BINARY_PATH"
+echo "Wasp command: $WASP_COMMAND"
+
+eval "$WASP_COMMAND build"
+
+# Make sure they build.
+echo "Ensure client builds"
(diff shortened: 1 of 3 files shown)


waspc/tools/ensure_todoapp_builds.sh
```diff
@@ -0,0 +1,27 @@
...
+echo "Running ensure_todoapp_builds.sh from $dir"
+
+cd "$dir/../examples/todoApp"
+echo "cd into $(pwd)"
+
+# Compile example app.
+WASP_BINARY_PATH="$(cabal list-bin wasp-cli)"
+CABAL_PROJECT_ROOT_PATH="$(cabal list-bin wasp-cli | sed s/\\/dist-newstyle.*//)"
+WASP_COMMAND="waspc_datadir=$CABAL_PROJECT_ROOT_PATH/data $WASP_BINARY_PATH"
+echo "Wasp command: $WASP_COMMAND"
```
Martinsos, line 16: Have you tried just running `cabal run wasp-cli build`? For me that works to build the waspc/examples/todoApp.

waspc/tools/ensure_todoapp_builds.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh -e
+
+# Compiles the todoApp and checks `npm run build` on both client and server.
```
Martinsos, line 3: Maybe provide a bit of context of why we care about this. You can explain that this is what is done during deploymen and that we want to make sure we have a way to test it without doign actual deployment, or smth like that?

waspc/tools/ensure_todoapp_builds.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh -e
```
Martinsos, line 1: You put this in separate file so we can also run it locally?

Martinsos (approved):

Martinsos:
Great, this will help!!!

waspc/tools/ensure_todoapp_builds.sh
shayneczyzewski: Ah yeah, that may work too, I'll try :D

waspc/tools/ensure_todoapp_builds.sh
shayneczyzewski: We could, but just easier to manage this way mainly.

Martinsos:
A thought here: this building step (running `npm build` in both projects, client and server), when does it happen? The only other place we do it at is Dockerfile, isn't it?

So what we really have is `wasp build` generating a code in .wasp/build, and then that code, we want to go there in those projects and build them, because that is what we finally want to deploy.

I am wondering if we should in some way extract this "building step" then, make it a standalone thing. Because in this PR we are really repeating logic that we have in that Dockerfile, I think? The idea of "building" the stuff in .wasp/build .

So maybe we should recognize that, and have a special thing for doing that, and then we can test that in CI + call it in Dockerfile + maybe Wasp users can call if they want, whatever.

I am not sure how it would be capture though. Maybe in .sh script at the top level of .wasp/build? If only `npm run build` go into it, maybe that is redundant then? What about prisma generation, I saw that step also in Dockerfile, should that be part of this "build" step?

PR #1147, opened 2023-04-18
Fixes Haskell list to JS array utility

infomiho:
Notice that we have double quotes around User and profile (we generate these JS arrays in Haskell) 
```ts
const query = createQuery<GetUsers>(
  'operations/get-users',
  ['"User"'],
)
```
```ts
const _waspOAuthConfig: OAuthConfig = {
    clientID: process.env.GOOGLE_CLIENT_ID,
    clientSecret: process.env.GOOGLE_CLIENT_SECRET,
    scope: ['"profile"'],
};
```


waspc/e2e-test/test-outputs/waspComplexTest-golden/waspComplexTest/.wasp/out/server/src/auth/providers/config/google.ts
```diff
@@ -11,7 +11,7 @@ const _waspUserDefinedConfigFn = undefined
 const _waspOAuthConfig: OAuthConfig = {
     clientID: process.env.GOOGLE_CLIENT_ID,
     clientSecret: process.env.GOOGLE_CLIENT_SECRET,
-    scope: ['"profile"'],
+    scope: ['profile'],
```
infomiho, line 14: I ran the e2e tests with `accept all` blindly and it negated the usefulness of e2e tests

Martinsos (approved):
Oh crap good one! We really should have write unit tests for this one.

PR #1168, opened 2023-04-26
Updates imports for throwInvalidCredentialsError

infomiho:
It fixes the circular dependency issue we had.

This is the simplest solution I could come up with, moving the `throwInvalidCredentialsError` from `auth/utils.ts` to `core/auth.js`. -> this just highlights the need for a cleanup in auth code which we will do soon.

Side note: can we detect circular imports like this automatically?

### DIFF
--- a/waspc/ChangeLog.md
+++ b/waspc/ChangeLog.md
@@ -2,6 +2,9 @@
 
 ## v0.10.3
 
+### Bug fixes
+- Fixed a bug with circular imports in JS code which prevented database seeding from working properly.
+
 ### Express middleware customization
 We now offer the ability to customize Express middleware:
 - globally (impacting all actions, queries, and apis by default)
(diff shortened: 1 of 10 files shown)


Martinsos (approved):
Sounds good!
Those are good questions: I think there are tools for detecting circular dependencies, we should investigate running one of those on our codebase. This is popular tool for it: https://www.npmjs.com/package/madge . Maybe you can create an issue for this also, to use this tool.
As for the structure of auth -> can you open an issue that links to this issue and suggest refactoring of auth JS code? And label it with "refactoring" and "shouldfix".

infomiho:
Here's an issue for the refactor: https://github.com/wasp-lang/wasp/issues/1169

I suggest we merge this into main, then cherry pick it to release and release v0.10.3

PR #1175, opened 2023-04-27
Small docs fixes

sodic:
Two fixes are included:
 - A cherry picked commit from `main` that fixes an import issue in the tutorial.
 - A commit that updates our docs to account for non-async operations.


shayneczyzewski (approved):
LGTM thanks for the doc updates! 🚀

PR #1186, opened 2023-05-02
WIP: Headless testing

infomiho:


shayneczyzewski (commented):
This is really cool @infomiho! I will do a deeper dive tomorrow, but had two initial questions:

1) It looks like we are going from about 7min to 10min runtime now on Ubuntu CI. Is there any way we can optimize it? Perhaps cache things like the browser artifacts (if not cached)? Or run things in parallel more (if not already)? I don't think it's terrible to add 40% for this platform, but we will just want to be mindful of continuing to increase overall runtime (where e2e was probably the first big culprit) I think in future.

2) Is there a way we can avoid having a duplicate example app, and just use the one under `waspc/examples` directly for testing? Or is this a slightly modified version? I just worry about skew between them and the potential maintenance burden of having two now (and possibly forgetting to test something we would otherwise in a different one).

Great work, will run/review more tomorrow 👍🏻

.github/workflows/ci.yaml
```diff
@@ -126,12 +126,55 @@ jobs:
...
+      # - name: Ensure todoApp builds
+      #   if: matrix.os == 'ubuntu-20.04' || matrix.os == 'macos-latest'
+      #   run: ./tools/ensure_todoapp_builds.sh
+
+      - name: Headless - Cache Node Modules
+        uses: actions/cache@v3
+        if: matrix.os == 'ubuntu-20.04'
+        with:
+          path: ~/.npm
+          key: node-modules-${{ hashFiles('./headless-test/package-lock.json') }}
```
shayneczyzewski, line 141: Should we preface all keys with `headless-` maybe? I imagine collisions with other steps are near impossible, but possibly just for identification's sake?

shayneczyzewski: Also, the logs say `Cache restored from key: node-modules-` but not the hash. Is that a possible bug?

shayneczyzewski: https://github.com/wasp-lang/wasp/actions/runs/4893035316/jobs/8735506230#step:14:16

shayneczyzewski (commented):
Overall this is looking really good to me @infomiho, nice job again! Left a few comments.

I agree it is probably best then to just keep this app separate and headless friendly. The other option could be to just have a custom headless .wasp file and overwrite the existing todoApp version before starting testing, if all the other code is assumed to be the same, but unsure if that mitigates all the downsides.

shayneczyzewski:
> @shayneczyzewski I've cached the dependencies now, most of the time is spent in booting the example app now (about a minute) and tests are running for <30s most of the time.

Awesome job on the caching! 🏃🏻💨 It is much faster 👍🏻

shayneczyzewski (approved):
Nice job on this @infomiho! 🥳 🚀

waspc/headless-test/README.md
```diff
@@ -0,0 +1,26 @@
+## Headless Test
```
Martinsos, line 1: I think we could improve this file to be more informative! Some ideas:
1. Headless Test as a name is not very descriptive. If I asked you what is Headless test, how would you describe it to me? I would probably describe it as "It is when we take the whole wasp app and try to run it, you know clicking and stuff, end to end, see that it works as it should". Based on that, I would name it "app-e2e-test". We already have e2e-test/ dir btw, but that one is for cli (maybe it should be named `cli-e2e-test`). So this kind of fits in with that. I undrestand that "headless testing" is coming from "headless browser testing" but that is really an implementation detail of how these e2e tests are performed. I could use "headless" browser testing to test anything where I need to click in browser, so it is a technique, not precise explanation of what we are testing that way.
2. I would emphasize here in README that what we are doing here is taking an actual Wasp app, running it with wasp, and then running a suite of e2e tests on it, to confirm that basic functionality works as it should. And then list the signup, login and stuff you listed above.
3. You tell me to check out the test file -> why? What should I check out? If you want to say that is where most of the testing code is defined, you can say it that way.

waspc/headless-test/README.md
```diff
@@ -0,0 +1,26 @@
...
+### How to run
+
+Running headless tests:
+```
+$ cd headless-test
+$ npm install
+$ npx playwright test
+```
+
+If something breaks, maybe the example app won't run. Try running it and see if there are any errors:
```
Martinsos, line 19: I don't understand this sentence. "If something breaks" -> what? Like in setting up playwright, or smth else, what could it be? "the example app won't run" -> you mean because the test won't run? Ok, sure, if test broke before that, that makes sense.
"Try running it and see if there are any errors" -> why would I do that, what will I see? What kind of errors?

waspc/headless-test/examples/todoApp/Dockerfile
```diff
@@ -0,0 +1 @@
+## HELLO!
```
Martinsos, line 1: Hi :D.

Why even include this file?

.github/workflows/ci.yaml
```diff
@@ -133,6 +133,54 @@ jobs:
...
+      - name: Headless - Install Dependencies
+        id: headless-install-dependencies
+        if: matrix.os == 'ubuntu-20.04'
+        run: |
+          cd headless-test
+          npm ci
+
+      - name: Headless - Store Playwright's Version
+        id: headless-store-playwright-version
+        if: matrix.os == 'ubuntu-20.04'
```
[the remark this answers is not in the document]
infomiho: I did try to solve the issue of grouping similar steps, it annoyed me that I had to repeat conditions, prefix the names etc. 

The issue with defining a new job is: Github Actions use a new isolated runner for each job - meaning we would need to set up Haskell, build Wasp etc. in the new job which I wanted to avoid.

.github/workflows/ci.yaml
```diff
@@ -133,6 +133,54 @@ jobs:
...
+
+      - name: Headless - Cache Playwright Browsers for Playwright's Version
+        id: headless-cache-playwright-browsers
+        if: matrix.os == 'ubuntu-20.04'
+        uses: actions/cache@v3
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-browsers-${{ env.PLAYWRIGHT_VERSION }}
+
+      - name: Headless - Setup Playwright
```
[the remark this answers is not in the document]
infomiho: I can adjust the name, sure 👍

.github/workflows/ci.yaml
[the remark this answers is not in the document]
Martinsos: Ah ok, I was hoping there is an easy way to have them reuse the runner / container, but it seems like there is not. Ok then, what can we do! Maaaaybe we could use yaml feature, I think they have some way to define variables or something? So we at least have one single place where we need to update that ubuntu version if we want to? Maybe that is worth giving a quick look.

Martinsos:
Let's also add `run test:headless` command to `run` script! And we can also have it trigger when `run test` runs.

Martinsos:
Sometimes docker processes running the database remain open, occuping port 5432. Can be tricky if you are running it locally. We might want to look into ensuring those docker containers get stopped.

PR #1190, opened 2023-05-03
Expose correct `User` type on the frontend

sodic:
Usage:
```typescript
import { User } from '@wasp/auth/types'

// ...

export const ProfilePage = ({ user }: { user: User }) => {
  // ...
}
```
I didn't change anything in the docs since they don't cover this feature properly anyway. We'll have to do a rewrite of that part.


infomiho:
Do we need to update the docs to mention this?

Zeko369:
You're all making the changes I wanted to propose 😅

PR #1196, opened 2023-05-09
WIP: Adds `--org` and launch secrets to `wasp deploy`

### LINKED ISSUE #1099: Add extra optional flags to `wasp deploy fly launch`
1) One user requested the ability to specify an `--org`. This appears to only be relevant for `flyctl launch`.
2) We should also allow users to specify envars for `launch` as well. Otherwise, if they have social auth enabled, for example, it will fail on the final deploy step. It is not hard to then add the envars and redeploy, but just not ideal.

shayneczyzewski:
### Description

⚠️ Draft, not yet ready for review ⚠️

Fixes #1099

TODO:
- [ ] final testing
- [ ] update docs with new options

### DIFF
--- a/waspc/packages/deploy/src/providers/fly/CommonOptions.ts
+++ b/waspc/packages/deploy/src/providers/fly/CommonOptions.ts
@@ -2,6 +2,7 @@ export interface CommonOptions {
 	waspExe: string;
 	waspProjectDir: string;
 	flyTomlDir?: string;
+	org?: string;
 }
 
 export interface DbOptions {
@@ -12,4 +13,9 @@ export interface DbOptions {
 
 export interface LocalBuildOptions {
 	buildLocally: boolean;
+}
+
+export interface SecretsOptions {
+	serverSecret: string[];
+	clientSecret: string[];
 }
--- a/waspc/packages/deploy/src/providers/fly/DeploymentInfo.ts
+++ b/waspc/packages/deploy/src/providers/fly/DeploymentInfo.ts
@@ -1,10 +1,10 @@
 import { TomlFilePaths } from './helpers/tomlFileHelpers.js';
 import { CommonOptions } from './CommonOptions.js';
 
-export type DeploymentInfo = Readonly<{
+export type DeploymentInfo<OptionType extends CommonOptions> = Readonly<{
 	baseName: string;
 	region?: string;
-	options: CommonOptions;
+	options: OptionType;
 	tomlFilePaths: TomlFilePaths;
 	clientName: string;
 	clientUrl: string;
@@ -13,12 +13,12 @@ export type DeploymentInfo = Readonly<{
 	dbName: string;
 }>;
 
-export function createDeploymentInfo(
+export function createDeploymentInfo<OptionType extends CommonOptions>(
 	baseName: string,
 	region: string | undefined,
-	options: CommonOptions,
+	options: OptionType,
 	tomlFilePaths: TomlFilePaths,
-): DeploymentInfo {
+): DeploymentInfo<OptionType> {
 	return {
 		baseName,
 		region,
--- a/waspc/packages/deploy/src/providers/fly/cmd/cmd.ts
+++ b/waspc/packages/deploy/src/providers/fly/cmd/cmd.ts
@@ -20,14 +20,18 @@ export async function cmd(flyctlArgs: string[], options: CmdOptions): Promise<vo
 	const tomlFilePaths = getTomlFilePaths(options);
 	const commonOps = getCommonOps(options.context, options.waspProjectDir, tomlFilePaths);
 
-	await runFlyctlCommand(commonOps, flyctlArgs);
+	await runFlyctlCommand(commonOps, flyctlArgs, options.org);
 }
 
-async function runFlyctlCommand(commonOps: CommonOps, flyctlArgs: string[]) {
+async function runFlyctlCommand(commonOps: CommonOps, flyctlArgs: string[], org?: string): Promise<void> {
 	commonOps.cdToBuildDir();
 	deleteLocalToml();
 	if (commonOps.tomlExistsInProject()) {
 		commonOps.copyProjectTomlLocally();
+	}
+
+	if (org) {
+		flyctlArgs.push('--org', org);
 	}
 
 	try {
--- a/waspc/packages/deploy/src/providers/fly/index.ts
+++ b/waspc/packages/deploy/src/providers/fly/index.ts
@@ -23,6 +23,13 @@ class FlyCommand extends Command {
 	addLocalBuildOption(): this {
 		return this.option('--build-locally', 'build Docker containers locally instead of remotely', false);
 	}
+	addSecretsOptions(): this {
+		function collect(value: string, previous: string[]) {
+			return previous.concat([value]);
+		}
+		return this.option('--server-secret <serverSecret>', 'secret to set on the server app (of form FOO=BAR)', collect, [])
+			.option('--client-secret <clientSecret>', 'secret to set on the client app (of form FOO=BAR)', collect, []);
+	}
 }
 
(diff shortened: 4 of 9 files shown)


waspc/packages/deploy/src/providers/fly/CommonOptions.ts
```diff
@@ -13,3 +14,8 @@ export interface DbOptions {
 export interface LocalBuildOptions {
 	buildLocally: boolean;
 }
+
+export interface SecretsOptions {
+	serverSecret: string[];
```
infomiho, line 19: Maybe go with plural `serverSecrets`, is a `string` the best way to represent a list of key values?
```ts
type Secret { key: string; value: string }
```

waspc/packages/deploy/src/providers/fly/DeploymentInfo.ts
```diff
@@ -13,12 +13,12 @@ export type DeploymentInfo = Readonly<{
 	dbName: string;
 }>;
 
-export function createDeploymentInfo(
+export function createDeploymentInfo<OptionType extends CommonOptions>(
```
infomiho, line 16:
```suggestion
export function createDeploymentInfo<Options extends CommonOptions>(
```

waspc/packages/deploy/src/providers/fly/cmd/cmd.ts
```diff
@@ -20,16 +20,20 @@ export async function cmd(flyctlArgs: string[], options: CmdOptions): Promise<vo
 	const tomlFilePaths = getTomlFilePaths(options);
 	const commonOps = getCommonOps(options.context, options.waspProjectDir, tomlFilePaths);
 
-	await runFlyctlCommand(commonOps, flyctlArgs);
+	await runFlyctlCommand(commonOps, flyctlArgs, options.org);
```
infomiho, line 23: Can we put `org` in `flyctlArgs` before, so we don't need another function param?

waspc/packages/deploy/src/providers/fly/index.ts
```diff
@@ -23,6 +23,13 @@ class FlyCommand extends Command {
 	addLocalBuildOption(): this {
 		return this.option('--build-locally', 'build Docker containers locally instead of remotely', false);
 	}
+	addSecretsOptions(): this {
+		function collect(value: string, previous: string[]) {
+			return previous.concat([value]);
+		}
+		return this.option('--server-secret <serverSecret>', 'secret to set on the server app (of form FOO=BAR)', collect, [])
```
infomiho, line 30: So I have to do `--server-secret` for each server secret?

Looking at what Fly's CLI supports - providing a list of key value pairs, maybe we can support that like `--server-secret="ONE=1 TWO=2"`?
```
❯ flyctl secrets
Secrets are provided to applications at runtime as ENV variables. Names are
		case sensitive and stored as-is, so ensure names are appropriate for
		the application and vm environment.

Usage:
  flyctl secrets [command]

Available Commands:
  import      Set secrets as NAME=VALUE pairs from stdin
  list        List application secret names, digests and creation times
  set         Set one or more encrypted secrets for an application
  unset       Unset one or more encrypted secrets for an application
  ```

File: waspc/packages/deploy/src/providers/fly/setup/setup.ts [hunk 5]
```diff
@@ -43,37 +43,74 @@ export async function setup(baseName: string, region: string, options: CommonOpt
...
+		`JWT_SECRET=${randomString}`,
+		// NOTE: Normally these would just be envars, but flyctl
+		// doesn't provide a way to set envars that persist to fly.toml.
+		'PORT=8080',
+		`WASP_WEB_CLIENT_URL=${deploymentInfo.clientUrl}`,
+	];
+
+	if (deploymentInfo.options.serverSecret.length > 0) {
+		deploymentInfo.options.serverSecret.forEach(secret => {
+			secretsArgs.push(secret);
```
infomiho, line 78: I see that you are passing it in as-is (strings in, strings passed to Fly CLI), maybe consider parsing the user input into `Secret` type to make sure that they provided valid secrets and then we stringily it here again.

waspc/packages/deploy/src/providers/fly/setup/setup.ts
```diff
@@ -43,37 +43,74 @@ export async function setup(baseName: string, region: string, options: CommonOpt
...
-	await $`flyctl launch --no-deploy --name ${deploymentInfo.clientName} --region ${deploymentInfo.region}`;
+	await $`flyctl launch --no-deploy  ${launchArgs}`;
 
 	// goStatic listens on port 8043 by default, but the default fly.toml assumes port 8080.
 	replaceLineInLocalToml(/internal_port = 8080/g, 'internal_port = 8043');
 
 	copyLocalClientTomlToProject(deploymentInfo.tomlFilePaths);
 
+	if (deploymentInfo.options.clientSecret.length > 0) {
+		await $`flyctl secrets set ${deploymentInfo.options.clientSecret}`;
```
infomiho, line 112: I know secrets does a redeploy, is there a way to avoid two deployments?

waspc/packages/deploy/src/providers/fly/setup/setup.ts
shayneczyzewski: Can be chewed on more here: https://github.com/wasp-lang/wasp/pull/1196#discussion_r1196473220

waspc/packages/deploy/src/providers/fly/cmd/cmd.ts
shayneczyzewski: I think there was a reason I didn't add it, but let me go back and verify why.

shayneczyzewski: Good catch, setting it earlier now, thnx 👍🏻

waspc/packages/deploy/src/providers/fly/DeploymentInfo.ts
```diff
@@ -1,10 +1,10 @@
 import { TomlFilePaths } from './helpers/tomlFileHelpers.js';
 import { CommonOptions } from './CommonOptions.js';
 
-export type DeploymentInfo = Readonly<{
+export type DeploymentInfo<Option extends CommonOptions> = Readonly<{
```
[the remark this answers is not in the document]
shayneczyzewski: I don't think we can. Some commands would error if given too much or little. Also, we transform some args before supplying them. And lastly, the ability to get the TS-goodness from knowing what they are/should be is nice. Unless maybe I missed some of the nuances in the question.

shayneczyzewski:
> Left a couple of comments.
> 
> I'm not 100% sure what this change does (i.e., what we had before vs. what we have now). so some of my suggestions might not make sense.
> 
> @shayneczyzewski could you list a "before and after" for all affected commands to help me understand?

Thanks for the comments, and sure thing. The main addition here is just two optional options:

Before:
```sh
wasp deploy fly launch shayne-test1 mia
```

After:
```sh
wasp deploy fly launch shayne-test1 mia -- --server-secret GOOGLE_CLIENT_ID=foo --server-secret GOOGLE_CLIENT_SECRET=bar --org hive
```

Secrets only apply to `launch`, and `org` applies to two other commands too.

waspc/packages/deploy/src/providers/fly/index.ts
infomiho: Makes sense 👍 Docker has a similar behaviour for their `--env` option as well

infomiho (approved):
LGTM, I'll test it out locally as well

Edit: working as expected ✅

waspc/packages/deploy/src/providers/fly/CommonOptions.ts
shayneczyzewski: The interface names are a convention of commander.js. Our option name is `--server-secret` which they will put into a variable `serverSecret`.

PR #1201, opened 2023-05-12
fix: cli start db error message on inactive docker

### LINKED ISSUE #1135: Have `wasp start db` produce nice warning message if docker daemon is not running
This is what it currently looks like:

```

Additional info:
     postgresql://postgresWaspDevUser:postgresWaspDevPass@localhost:5432/SaaSTemplate-f587f704d3
     wasp-dev-db-SaaSTemplate-f587f704d3

docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
See 'docker run --help'.
wasp-bin: callCommand: docker run --name wasp-dev-db-SaaSTemplate-f587f704d3 --rm --publish 5432:5432 -v wasp-dev-db-SaaSTemplate-f587f704d3:/var/lib/postgresql/data --env POSTGRES_PASSWORD=postgresWaspDevPass --env POSTGRES_USER=postgresWaspDevUser --env POSTGRES_DB=SaaSTemplate-f587f704d3 postgres (exit 125): failed
```

While I thought message here is clear ("Is the docker daemon running") it turns out, based on feedback, that it is easy to miss. So it would be much better if we detected this explicitly with Wasp and provided a nice, explicit error message about it + some instructions on how to go about it.

We will probably want to run `docker info` and parse its output to determine if docker daemon is running or not.

modotte:
### Description

Improve `wasp start db` error message whenever docker daemon isn't running by using existing error handling mechanism and guiding user to relevant docker troubleshooting docs.

fixes #1135


waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -132,6 +133,8 @@ startPostgreDevDb waspProjectDir appName = do
           Dev.Postgres.defaultDevPass
           Dev.Postgres.defaultDevUser
           dbName
+
+  throwIfDockerDaemonIsNotRunning
```
Martinsos, line 137: We will want to put this check together with other checks at the start of this function, so I would put it right after

```
throwIfExeIsNotAvailable
    "docker"
    "To run PostgreSQL dev database, Wasp needs `docker` installed and in PATH."
```

That way we have to docker checks first, then a check for occupied port, and then if all of them pass, we go on and run the database!

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -157,6 +160,30 @@ startPostgreDevDb waspProjectDir appName = do
...
 
+    throwIfDockerDaemonIsNotRunning :: Command ()
+    throwIfDockerDaemonIsNotRunning = do
+      whenM (liftIO isDockerStopped) throwDockerDaemonIsNotRunningError
+      where
+        isDockerStopped = do
+          (exitCode, errorString, _) <-
+            -- For docker formatting reference, see https://pkg.go.dev/text/template#hdr-Functions
+            readProcessWithExitCode "docker" ["info", "--format", "{{ index .ServerErrors 0 }}"] ""
+          pure $ exitCode == ExitSuccess && errorString == "Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?\n"
```
Martinsos, line 171: What if exit code is not success? I know that is very unlikely to happen since we just checked that `docker` command is accessible, but still, it is safest to assume it can happen.
If that does happen and it is not success, `isDockerStopped` will be set to False, which doesn't make much sense I believe, because it seems as if everything is ok while it is really not.

One way we can fix it is by turning `isDockerStopped` into `isDockerDaemonRunning` and modifying the conditions accordingly.

Another thing is that checking for `unix:///var/run/docker.sock` seems a bit fragile. What if they are running this on Windows? Ok true, we don't support windows yet hm, but we might in the future. I also wonder if this is same across all the systems (linux, mac).
Also, what if we have a different error message, something else went wrong?

Again, solution might be to turn things around -> instead of looking for specific error, we can look at specific success condition. If we get that success condition, we know that all is ok and we can set `isDockerRunning` to True. If not, we set it to False, so all errors are captured under that.

What might help is if you give me a bit more info on what Docker outputs in different situations and your reasoning about it, then we can reason about it together.

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -157,6 +160,30 @@ startPostgreDevDb waspProjectDir appName = do
...
+
+        throwDockerDaemonIsNotRunningError =
+          E.throwError $
+            CommandError
+              "Docker daemon is not running"
+              ( unlines
+                  [ "Wasp can't run dev database for you because docker daemon isn't running",
+                    "in the background. Have you tried executing `systemctl start docker` before running",
```
Martinsos, line 179: `systemctl` makes sense only for OSes using systemd. Mac doesn't use it if I am correct, and not all Linux OSes use it. So probably best to skip this specific advice and let's just go for those links.

Martinsos (changes requested):
Looking good @modotte !

It is good overall, nice job -> there are a couple of smaller changes that might improve robustness / messaging of our approach, I commented on those, let's take care of those and it will be ready to merge!

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
modotte: Hi! Thanks for the review :smile:!

> What if exit code is not success? I know that is very unlikely to happen since we just checked that docker command is accessible, but still, it is safest to assume it can happen.
If that does happen and it is not success, isDockerStopped will be set to False, which doesn't make much sense I believe, because it seems as if everything is ok while it is really not.

> Another thing is that checking for unix:///var/run/docker.sock seems a bit fragile. What if they are running this on Windows? Ok true, we don't support windows yet hm, but we might in the future. I also wonder if this is same across all the systems (linux, mac).
Also, what if we have a different error message, something else went wrong?

You're right! It's fragile and certainly ignores the cases such as if the user chose to start the daemon at different socket location over the default or standard `tcp://127.0.0.1:2376` on Windows.

> What might help is if you give me a bit more info on what Docker outputs in different situations and your reasoning about it, then we can reason about it together.

My thought was that `docker info --format {{ index .ServerErrors 0 }}` should only return the error string whenever docker daemon is not running and I assumed the `unix:///var/run/docker.sock` string would be all the same across the OSes and cannot be tampered, which I now found to be incorrect. 

For example, it doesn't take care of situation where `docker` client needs root access to even run the `docker info` command in some systems where docker access isn't allowed for non-root users, which confirms your first point.

Also, this is just one of the many docker daemon errors I'm aware of, so there are probably numerous , `docker` throwIf* handlers that need to be written if I choose to stick with current code, which I don't think is ideal.

So yes :smiley:, I think it's much better if we go with your suggestion to check if docker is running and throw the error otherwise.

modotte: Regarding my own question about multiple docker throwIfs, I think I'll just use `system` to check for `0` from `docker info` and let docker prints its actual cause of failure but also properly return Wasp's error to mark something did clearly *fail*.

modotte: Regarding about my handling different kind of docker error with multiple throwIfs comment, I think it's perhaps better if we just redirect actual docker error message into Wasp's error handling to give user some info of what exactly is failing?

Something like this, probably?
image

but maybe I'm just over-complicating things.

Martinsos: Hm I am also not sure -> we have to assume different errors can happen, and we have to figure out how to handle them.

Right now, if they try to run `start db` but docker deamon is not running, they actually get error message from docker that says that docker deamon is possibly not running. It is just that it is not a very prominent message and if you don't read it, you will not know what happened. Which is why I didn't implement this at the beginning, beacuse I assumed people will just read that, but then couple of them didn't and so this issue was created.

So we said let's run `docker info` and if all is ok and if it says under Server that stuff is running, then all is ok, if not we let them know what is happening.

The problem is that we can't cover all the "not" reasons.

I have two ideas to proceed:
1. Don't do anything, we close this PR. They will get error message from `start db` and they should read it. Let them find their way about it. Maybe not the best experience, but at least we won't mess anything up.
2. We call `docker info` and look for clean signs that docker deamon is not running. If we detect those signs, we report that. If we don't find them and there is no error, we don't say anything, conclude all is good. If instead there is an error but we are not sure what it is about, we forward it and way "hey, we ran docker info to check if your docker deamon is running but got this error while doing that, check it out", smth like that.

modotte: Ah, that's alright then. I'm actually not too sure about how `docker info` works and I'd wish the return types and output are all well-documented for the command. I think I'll close this PR, for now, to let someone else with more docker experience fix the issue. Thx for the review btw! :slightly_smiling_face:

Martinsos: Thank you for the effort! I think this is reasonable -> to be honest I also at the moment am not sure how to tackle it, and with what we just learned, would probably vote myself for not handling it. It just feels like too much work and the result will still not be super solid.

PR #1203, opened 2023-05-17
Adds WebSocket support

shayneczyzewski:
### Description

Adds support for WebSockets (via Socket.IO), adds a `useSocket` abstraction, and demonstrates a use of shared types on the client and server.

Example from docs (https://github.com/wasp-lang/wasp/blob/e4745a7bcc20ffe3a90b48a290cca949603665d0/web/docs/guides/websockets.md) integrated into example app. You can chat on profile page. We can remove if not helpful.

TODO:
- [x] cleanup
- [x] more testing, deploy, etc.
- [x] add docs
- [ ] update changelog


waspc/data/Generator/templates/react-app/src/socket.ts
```diff
@@ -0,0 +1,4 @@
+import { io } from 'socket.io-client'
+import config from './config'
+
+export const socket = io(config.apiUrl, { autoConnect: false })
```
shayneczyzewski, line 4: I think it is better to only connect when `useSocket` is used (which takes care of connecting by default). Thoughts?

shayneczyzewski: Actually, reversed that since handling connection logic across multiple components is not worth the extra complexity.

waspc/src/Wasp/Generator/ServerGenerator.hs
```diff
@@ -399,3 +408,38 @@ genMiddleware spec =
...
+
+genWebSockets :: AppSpec -> Generator [FileDraft]
+genWebSockets spec
+  | AS.WS.areWebSocketsUsed spec =
+      sequence
+        [ genWebSocketTs spec
+        ]
+  | otherwise = return []
+
+genWebSocketTs :: AppSpec -> Generator FileDraft
```
Martinsos, line 425: Quite a lot of web socket functions, should we extract them into separate module?

Second, why does the name here end with `Ts`, does that symbolize that it uses Typescript, or that these are Types?
If the former, I would question if we need that, I haven't seen us using it before. If latter, I would expand it to full name, `Types`.

web/docs/guides/websockets.md
```diff
@@ -0,0 +1,129 @@
+---
+title: WebSockets
+---
+import useBaseUrl from '@docusaurus/useBaseUrl';
+
+# WebSocket support
+Wasp provides a fully integrated WebSocket experience by utilizing [Socket.IO](https://socket.io/) on the client and server. We handle making sure your URLs are correctly setup, CORS is enabled, and provide a useful `useSocket` abstraction for use in React components.
+
```
Martinsos, line 8: Maybe we could mention here, as a NOTE, that if you want, you can still roll your own sockets, via `serverSetupFn`?

web/docs/guides/websockets.md
```diff
@@ -0,0 +1,129 @@
...
+## Enable in Wasp file
```
Martinsos, line 9: Could we, before we dive into details, give a quick overview of needed steps?
We could say smth like:
1. Define your socket logic on the server.
2. Declare you are using sockets in wasp file + connect it to the server logic you defined.
3. Use them on the client, in React, via `useSocket`.
4. Optional: defines types for messages.

Maybe there shouldn't be (4), this is just an idea.

web/docs/guides/websockets.md
```diff
@@ -0,0 +1,129 @@
...
+export interface InterServerEvents { }
+
+// Data that is attached to the socket.
+// NOTE: Wasp automatically injects the JWT into the connection,
+// and if present/valid, the server adds a user to the socket.
+export interface SocketData extends WaspSocketData { }
+```
+
+## Add server support
+On the server, you will get `io` argument and `context` for your WebSocket function. You can use this `io` object to register callbacks for all the regular Socket.IO events.
```
Martinsos, line 46: Will they know what is this `io` object? Is it something they would normally get from socket.io? Could we link to what this object is, maybe link to its type on socket.io or its docs or something?

Martinsos: You don't say much about `context`?

web/docs/guides/websockets.md
```diff
@@ -0,0 +1,129 @@
...
+  })
+}
+```
+
+## Add client support
+Client access to WebSockets is provided by the `useSocket` abstraction. It returns:
+- `socket: Socket` for sending and receiving events.
+- `isConnected: boolean` for showing a display of the Socket.IO connection status.
+  - Note: Wasp automatically connects establishes a WebSocket connection from the client to the server, so you do not need to explicitly `socket.connect()` or `socket.disconnect()`. All components using `useSocket` share the same underlying `socket`.
+- `refreshAuthToken: () => void` for refreshing the auth token on the `socket` when login status changes.
```
Martinsos, line 76: Should we use this somewhere in our Auth logic?

web/docs/guides/websockets.md
```diff
@@ -0,0 +1,129 @@
...
+    socket.on('chatMessage', async (msg) => {
+      console.log('message: ', msg)
+      io.emit('chatMessage', { id: uuidv4(), username, text: msg })
+    })
+  })
+}
+```
+
+## Add client support
+Client access to WebSockets is provided by the `useSocket` abstraction. It returns:
```
Martinsos, line 72: What if they want to use `socket` directly, without `useSocket`, should we allow them that? What if they want to make connection conditional, and have it connected only sometimes / when used? Hm this sounds like reasonable request, have socket connected only when used. I am not saying we have to support all this, but would love to explore use cases a bit.

Ok let's imagine this: I have an app which has a chat in it. Chat is used only sometimes. So with this PR, all instances of my Wasp app will maintain active connection, even though possibly only 1% of them is using the socket? Could that be an issue? Is there a limit to number of socket connections that can even be opened?

Martinsos (commented):
Looking good @shayneczyzewski , I am excited about this coming in!
Left first round of comments!

waspc/data/Generator/templates/server/src/webSocket.ts
```diff
@@ -0,0 +1,60 @@
...
+        {=# entities =}
+        {= name =}: typeof prisma.{= prismaIdentifier =},
+        {=/ entities =}
+      }
+    }
+  ) => Promise<void> | void
+
+type ServerType = Parameters<typeof {= webSocket.fn.importIdentifier =}> [0]
+
+export async function init(server: http.Server): Promise<void> {
```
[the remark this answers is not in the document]
shayneczyzewski: I can try to comment it a bit for people new to Socket.IO and about how our type plays into it.

`io` is the canonical name for a Socket.IO `Server`. It is used heavily in docs and in real user code. Question is do we use consistent library conventions internally, or Wasp-etize it. I'd probably vote for former.

I can extract middleware into helper 👍🏻

web/docs/guides/websockets.md
shayneczyzewski: Yep I can add a quick overview

web/docs/guides/websockets.md
shayneczyzewski: I think they should, but I can link to Socket.IO docs for `io`. I can mention `context` will contain entities. And yeah, I can note the user on socket 👍🏻

shayneczyzewski (commented):
Thanks @Martinsos replied to comments, and will start on those I can 👍🏻 appreciate it!

web/docs/guides/websockets.md
shayneczyzewski: Ok, so I pulled `refreshAuthToken` up to the module level, exported it and `socket`, and added an `autoConnect` option. So this effectively does 1 and 2, and they can now support your use case by setting `autoConnect: false` and then manually calling `socket.connect()`/`socket.disconnect()` in the component that needs it 👍🏻

shayneczyzewski: But by default, we will still just have `autoConnect` be `true`

waspc/data/Generator/templates/react-app/src/webSocket.ts
```diff
@@ -0,0 +1,56 @@
+import { useState, useEffect } from 'react'
+import { io, Socket } from 'socket.io-client'
+import { DefaultEventsMap, EventsMap } from '@socket.io/component-emitter'
+
+import { getAuthToken } from './api'
+import config from './config'
+
+// TODO: In the future, it would be nice if users could configure this somehow.
+const socket = io(config.apiUrl)
```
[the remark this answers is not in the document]
Martinsos: Ok nice to see they do it in similar way!

Ok, so one socket should be enough -> makes sense, I wasn't sure about that.

As per connection -> so by default it is always connected?
They mention here, that if only part of your app uses sockets, you might want to disconnect it: https://socket.io/how-to/use-with-react#disconnection . I also wonder if in such case they might want to disconnect it, to not maintain the connection all the time if they don't need it. I was thinking that we could have something smart, like reference counting, that knows when no component is using socket for longer than X seconds and then disconnects it manually, that could be interesting to have! They could configure smth like that. Maybe for the future, but could be nice to document it.
What we document hm: the fact that it is always connected which they might not want, and that they can handle that manually via `socket`, or that we could handle it for them with smart reference counting, with timer.

As for putting stuff in a hook -> ok, makes sense! I would still expose them globally also, but we can leave hook for a convenience. Although they might a feeling that this is some kind of "local" socket hm. So we should document well that it is all the same socket, global socket.

waspc/data/Generator/templates/server/src/webSocket.ts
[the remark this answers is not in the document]
Martinsos: Ok, let's go for the conventions then! Maybe just a bit of comments would help then, for people who don't konw much about Socket.IO .

waspc/data/Generator/templates/server/src/webSocket.ts
```diff
@@ -0,0 +1,60 @@
...
+import http from 'http'
+import { Server, Socket } from 'socket.io'
+import { EventsMap, DefaultEventsMap } from '@socket.io/component-emitter'
+
+import { getUserFromToken } from './core/auth.js'
+import config from './config.js'
+import { WaspSocketData } from './universal/types.js'
+import prisma from './dbClient.js'
+
+{=& webSocket.fn.importStatement =}
```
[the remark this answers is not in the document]
Martinsos: Yes I think that would help!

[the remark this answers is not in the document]
infomiho: `userWebSocketFn`

waspc/examples/todoApp/src/client/pages/ProfilePage.tsx
```diff
@@ -13,10 +15,36 @@ export const ProfilePage = ({
...
+  const { socket, isConnected } = useSocket<ServerToClientEvents, ClientToServerEvents>()
+  const inputRef = useRef<HTMLInputElement>(null)
+
   useEffect(() => {
     fetchCustomRoute()
+
+    socket.on('chatMessage', logMessage)
+
+    return () => {
+      socket.off('chatMessage', logMessage)
```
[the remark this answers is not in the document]
infomiho: This is a really cool idea @Martinsos and it goes in line with some other React libs for sockets that I saw.

I'd love to do something like `const { isConnected, registerHandler } = useSocket()`

and then `registerHandler('event', (even) => {})`

We would then use `useEffect` inside and the user would avoid a common pitfall (not using useEffect keep registering the handlers on rerenders -> happened to me today)

waspc/examples/todoApp/src/client/pages/ProfilePage.tsx
```diff
@@ -13,10 +15,36 @@ export const ProfilePage = ({
 }: {
   user: User
 }) => {
+  const [messages, setMessages] = useState<{ id: string, username: string, text: string }[]>([]);
+  const { socket, isConnected } = useSocket<ServerToClientEvents, ClientToServerEvents>()
+  const inputRef = useRef<HTMLInputElement>(null)
```
infomiho, line 20: I know this is just an example app but I've seen the `useState` pattern with inputs more often:
```
const [message, setMessage] = useState('');
...
<input value={value} onChange={event => setMessage(event.target.value)} />
```

web/docs/guides/websockets.md
infomiho: The most advanced users that might want that will probably figure out that they can do it anyways? It might confuse the not so advanced users.

web/docs/guides/websockets.md
```diff
@@ -0,0 +1,129 @@
...
+
+  webSocket: {
+    fn: import { webSocketFn } from "@server/webSocket.js"
+    // Optional: entities: [...]
+  },,
+}
+```
+
+## Optional: Add shared types for client and server use
+If TypeScript is your thing, you can add shared interface definitions to use on both the client and server. This allows you to type all the messages you send and receive!
```
infomiho, line 24:
```suggestion
If you are using Typescript you can add shared interface definitions to use on both the client and server. This allows you to type all the messages you send and receive!
```

web/docs/guides/websockets.md
```diff
@@ -0,0 +1,129 @@
...
+}
+
+export interface ClientToServerEvents {
+  chatMessage: (msg: string) => void;
+}
+
+export interface InterServerEvents { }
+
+// Data that is attached to the socket.
+// NOTE: Wasp automatically injects the JWT into the connection,
```
infomiho, line 40: Saying `JWT` here feels bit low level, I would maybe say "auth info" or something like that so people don't need to think too hard.

waspc/data/Generator/templates/react-app/src/webSocket.ts
```diff
@@ -0,0 +1,56 @@
...
+  ClientToServerEvents extends EventsMap = DefaultEventsMap
+>(): {
+  socket: Socket<ServerToClientEvents, ClientToServerEvents>;
+  isConnected: boolean;
+  refreshAuthToken: () => void
+} {
+
+  const [isConnected, setIsConnected] = useState(socket.connected)
+
+  // TODO: Any better way to do this?
```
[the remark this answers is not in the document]
shayneczyzewski: Yeah true, I was just wondering if that change was required for this v1, but I think it hopefully won't be too much extra work and would clean it up a bit. Ok, I'll look into adding a way to register a listener for auth events 👍🏻

web/docs/guides/websockets.md
```diff
@@ -0,0 +1,129 @@
...
+    })
+  })
+}
+```
+
+## Add client support
+Client access to WebSockets is provided by the `useSocket` abstraction. It returns:
+- `socket: Socket` for sending and receiving events.
+- `isConnected: boolean` for showing a display of the Socket.IO connection status.
+  - Note: Wasp automatically connects establishes a WebSocket connection from the client to the server, so you do not need to explicitly `socket.connect()` or `socket.disconnect()`. All components using `useSocket` share the same underlying `socket`.
```
infomiho, line 75:
```suggestion
  - Note: Wasp automatically connects and establishes a WebSocket connection from the client to the server, so you do not need to explicitly `socket.connect()` or `socket.disconnect()`. All components using `useSocket` share the same underlying `socket`.
```

web/docs/guides/websockets.md
```diff
@@ -0,0 +1,129 @@
...
+
+```ts title=src/client/Chat.tsx
+import React, { useEffect, useRef, useState } from 'react'
+import { useSocket } from '@wasp/webSocket'
+import { ClientToServerEvents, ServerToClientEvents } from '../shared/webSocket'
+
+export const ProfilePage = () => {
+  const [messages, setMessages] = useState<{ id: string, username: string, text: string }[]>([]);
+  const { socket, isConnected } = useSocket<ServerToClientEvents, ClientToServerEvents>()
+  const inputRef = useRef<HTMLInputElement>(null)
```
infomiho, line 87:
```suggestion
  const [message, setMessage] = useState('')
```

web/docs/guides/websockets.md
```diff
@@ -0,0 +1,129 @@
...
+
+  function logMessage(msg: { id: string, username: string, text: string }) {
+    setMessages((priorMessages) => [msg, ...priorMessages])
+  }
+
+  function handleSubmit(e: React.FormEvent<HTMLFormElement>) {
+    e.preventDefault()
+
+    if (inputRef.current !== null) {
+      socket.emit('chatMessage', inputRef.current.value)
```
infomiho, line 105:
```suggestion
      socket.emit('chatMessage', message)
```

waspc/data/Generator/templates/react-app/src/webSocket.ts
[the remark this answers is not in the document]
shayneczyzewski: The change I recently made allows them to control `autoConnect`. So they can do the `socket.connect/disconnect` stuff if they want it only active at certain points. I would personally vote to not bother us doing RC during this round. But I will make a note in the docs about connection handling and this option. I think it should be easier for them to manage the connection if they need to vs us trying to handle it generically in this initial round.

web/docs/guides/websockets.md
```diff
@@ -0,0 +1,129 @@
...
+  const connectionIcon = isConnected ? '🟢' : '🔴'
+
+  return (
+    <>
+      <h2>Chat</h2>
+      <div>
+        <form onSubmit={handleSubmit}>
+          <div className="flex space-x-4 place-items-center">
+            <div>{connectionIcon}</div>
+            <div><input type="text" ref={inputRef} /></div>
```
infomiho, line 120:
```suggestion
            <div><input type="text" value={message} onChange={event => setMessage(event.target.value)} /></div>
```

waspc/examples/todoApp/src/shared/webSocket.ts
```diff
@@ -0,0 +1,13 @@
+import { WaspSocketData } from '@wasp/universal/types'
```
infomiho, line 1: We could also do full-stack types here:
```ts
// Server is coming from Socket.IO
import { webSocketFn, Server } from '../../../server/src/ext-src/websocket'

type WebSocketFn = typeof webSocketFn

type Events = Parameters<WebSocketFn>[0] extends Server<infer F, infer S> ? [F, S] : never

const MainPage = ({ user }) => {
  const { socket, isConnected } = useSocket<Events[1], Events[0]>();
```

This is very rough but it works:
1. uses the type of the user defined webSocket function
2. Infers the two generic arguments the user passed when defining the fn

infomiho: That way we avoid having the user import the same types in the client and the server, they can just define it on the server

infomiho (commented):
Looking good! 

Works for me locally, I've played around with it and it's nice that I don't have to write things too much by myself.

Two main suggestions:
- going for full-stack types if we can (I think we can from playing around a bit) and not do the double shared folder imports
- going for some helper for registering handlers on the FE (`registerHandler`) which takes care of `useEffect` nuance for me

waspc/examples/todoApp/src/client/pages/ProfilePage.tsx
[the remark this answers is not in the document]
shayneczyzewski: Ok, sounds good, I can add that then 👍🏻

waspc/examples/todoApp/src/client/pages/ProfilePage.tsx
shayneczyzewski: On change will fire for every keystroke though, right? Also, I was wanting to handle both enter to submit and a button style.

waspc/examples/todoApp/src/shared/webSocket.ts
shayneczyzewski: Interesting. So would Wasp be the one doing the above, or the user? I am unsure what we would expose or how the user would still parameterize `useSocket` in a clean way based on the example?

Also, does this mean we would be including `Server` (and thus the non-client `'socket.io'`) on the client, too?

waspc/examples/todoApp/src/client/pages/ProfilePage.tsx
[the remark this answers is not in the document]
Martinsos: I just thought maybe that is nicer instead of writing their own useEffect that has to do on and off! But if you think it is a bit too much we don't have to do it.

waspc/src/Wasp/Generator/ServerGenerator.hs
Martinsos: Aha so we did use it hm! Ok, whatever you like the bets! I would be ok with nixing it also.

web/docs/guides/websockets.md
Martinsos: Yes, or if they have some special needs. But you decide, if you think it is redundant and clutters the docs here, let's skip, it is ok with me.

web/docs/guides/websockets.md
Martinsos: yes, on login for example.

Although, if we do the listener thing in api.ts, then that should be enough.

web/docs/guides/websockets.md
Martinsos: oh ok that is pretty cool now, sounds good!

Could you elaborate a bit more on the clustering? I don't know the problem you are talking about, maybe you can explain it to me in a bit more detail here? Sounds to me like we should create an issue for this so we can track it in GH?

waspc/examples/todoApp/src/shared/webSocket.ts
Martinsos: Oh fancy I don't understand it fully but it sounds great :D.

web/docs/guides/websockets.md
shayneczyzewski: I have a note in the code with a link, but essentially the servers will need to talk to each other if you want all messages to go to everyone. Additionally, if using long polling, we would need sticky session.

For example:
- https://socket.io/docs/v4/using-multiple-nodes
- https://socket.io/docs/v4/cluster-adapter/

If it needs a full issue just let me know 👍🏻 But as of now, this works for number of server containers = 1, not > 1.

waspc/data/Generator/templates/react-app/src/webSocket.ts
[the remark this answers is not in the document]
Martinsos: Sounds good!

web/docs/guides/websockets.md
```diff
@@ -0,0 +1,133 @@
...
+import useBaseUrl from '@docusaurus/useBaseUrl';
+
+# WebSocket support
+Wasp provides a fully integrated WebSocket experience by utilizing [Socket.IO](https://socket.io/) on the client and server. We handle making sure your URLs are correctly setup, CORS is enabled, and provide a useful `useSocket` abstraction for use in React components.
+
+To get started, you need to:
+1. Define your WebSocket logic on the server
+1. Declare you are using WebSockets in your Wasp file, and connect it with your server logic
+1. Use WebSockets on the client, in React, via `useSocket`
+1. Optionally define the types for messages for use on the client and server
```
Martinsos, line 13:
```suggestion
1. Define your WebSocket logic on the server.
2. Declare you are using WebSockets in your Wasp file, and connect it with your server logic.
3. Use WebSockets on the client, in React, via `useSocket`.
4. Optionally define the types for messages for use on the client and server.
```

web/docs/guides/websockets.md
Martinsos: Ok sounds good -> please create a full Github issue for this, so we can track it!

waspc/data/Generator/templates/react-app/src/api.ts
```diff
@@ -40,6 +51,20 @@ api.interceptors.response.use(undefined, (error) => {
   return Promise.reject(error)
 })
 
+// This handler will run on other tabs (not the active one), and will
```
Martinsos, line 54: Aha, so this basically listens for changes to storage coming from somewhere else and then updates the state based on that. And main purpose of this is to detected changes in other tabs.

So what happens when somebody calls setAuthToken -> won't that trigger this listener also, and then it will get triggered redundantly? So we will already set `authToken` variable in this module, and we will already emit the event 'authToken.set', but then this will trigger again and those things will happen again? I guess it doesn't hurt, but it is duplicate work + our event is being sent out twice which is a bit weird.

waspc/data/Generator/templates/react-app/src/webSocket.ts
```diff
@@ -0,0 +1,71 @@
...
+import config from './config'
+
+// TODO: In the future, it would be nice if users could pass more
+// options to `io`, likely via some `configFn`.
+export const socket = io(config.apiUrl, { autoConnect: {= autoConnect =} })
+
+function refreshAuthToken() {  
+  // NOTE: In the future, we should consider making this explicit in the Wasp file when
+  // we make the change for how auth works in Operations.
+  // For now, it is fine, and can be ignored on the server if not needed.
```
Martinsos, line 17: Is this comment still relevant? If so, what does it mean, I don't understand it.

waspc/data/Generator/templates/react-app/src/webSocket.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  function registerHandler<Event extends keyof ServerToClientEvents>
+    (event: Extract<Event, string>, cb: ServerToClientEvents[Event]) {
+    useEffect(() => {
+      socket.on(event, cb)
+      return () => {
+        socket.off(event, cb)
+      }
+    }, [])
+  }
```
Martinsos, line 68: Is this valid regarding the rules that hooks must obey? I thought that hook needs to be used in component at the top level, or in another hook. No branching.

What we do is have a hook that returns a function which then calls useEffect (another hook) inside of it, I find that a bit weird and potentially breaking those rules. They are calling a hook, but it looks like it is not a hook, ... .

> useEffect is a Hook, so you can only call it at the top level of your component or your own Hooks. You can’t call it inside loops or conditions. If you need that, extract a new component and move the state into it.

Maybe it passes React checks somehow, but it doesn't look to me like it clearly obeys these rules.

To me it sounds like the better way would be that we have a separate hook called `useSocketListener` that can be used to register a listener.

Martinsos: I just showed this code to chatGPT + I gave it that piece of react docs and it suggested on its own that we create another hook, like I suggested above. I am not sure if that is much of evidence because I kind of pushed it into that line of thinking and chatGPT is very agreeable but still :D.

shayneczyzewski: Yeah I agree, I felt funny doing it this way too, but couldn't see any other way to do what @infomiho suggested and it seemed to work fine. Conceptually I think it is probably "safe"/valid, but perhaps for safety, a separate hook is preferable though. I'll extract 👍🏻

waspc/data/Generator/templates/react-app/src/webSocket.ts
shayneczyzewski: It is just meant to say, "when we fix how `auth: true` works for operations and unauthenticated calls, we should mirror that here" if that makes sense. It was something you noticed a while back I think.

waspc/data/Generator/templates/react-app/src/api.ts
shayneczyzewski: No, the active tab will call the API functions, but won't get this handler. This only runs in the non-active windows (aka the one _not_ calling API functions).

waspc/data/Generator/templates/react-app/src/api.ts
```diff
@@ -1,29 +1,40 @@
...
+
 import config from './config'
 import { storage } from './storage'
 
 const api = axios.create({
   baseURL: config.apiUrl,
 })
 
+// Used to allow API clients to register for auth token change events.
+export const events = new EventEmitter()
```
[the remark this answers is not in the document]
shayneczyzewski: Yeah, I can see what that would look like 👍🏻

waspc/data/Generator/templates/react-app/src/webSocket.ts
Martinsos: I also don't see a better way to do what @infomiho suggested, but on the other hand the idea of `useSocketListener` hook sounds good to me!

waspc/data/Generator/templates/react-app/src/webSocket.ts
Martinsos: Ah yes! Could you please update the comment a bit to be more clear about this, and link to the github issue that speaks about that problem we have with `auth` on operations? Also, mention in that GH issue that we should also take a look at sockets then while working on it.

waspc/data/Generator/templates/react-app/src/webSocket.ts
shayneczyzewski: He was going to take a look to make sure I was doing it right, but thought it was valid. If so, maybe we can just leave as is?

waspc/data/Generator/templates/react-app/src/api.ts
Martinsos: Oh interesting! What magic is that? It doesn't emit that event to itself?

waspc/data/Generator/templates/react-app/src/api.ts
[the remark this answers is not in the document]
Martinsos: Ok, if it is too complex feel free to punt on it though.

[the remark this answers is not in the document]
shayneczyzewski: I think the typed ones may be, but I will at least make them string constants 👍🏻

waspc/data/Generator/templates/react-app/src/webSocket.ts
Martinsos: I have to admit I don't see how it could be valid with that approach. But I am not a React expert hm.

I would just go for the separate hook. But if you prefer waiting for @infomiho to give his take on this, we can also do that!

waspc/data/Generator/templates/react-app/src/api.ts
shayneczyzewski: By design it seems https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event

> Note: This won't work on the same page that is making the changes — it is really a way for other pages on the domain using the storage to sync any changes that are made. Pages on other domains can't access the same storage objects.

web/docs/guides/websockets.md
shayneczyzewski: https://github.com/wasp-lang/wasp/issues/1228

waspc/data/Generator/templates/react-app/src/webSocket.ts
shayneczyzewski: Ok, after I extracted it, I can kinda see how these should be equivalent now. In the former case, we are just providing back a reference to a function (a custom hook in both cases), not really different from it being at the top level. The downside with a separate hook is it's more verbose since it requires adding the type:

Screenshot 2023-06-01 at 1 12 49 PM
Screenshot 2023-06-01 at 1 13 02 PM

Martinsos: Hm too bad that we have to repeat that type but I guess we can't avoid it.

I am ok also with the first option in general, the problem is that I am suspicious it breaks the rules what React requires from us for using hooks, so I wonder if it will misbehave down the road possibly. I could ask in our Discord.

waspc/data/Generator/templates/react-app/src/webSocket.ts
Martinsos: Ok, feel free to resolve the comment once that is taken care of!

waspc/data/Generator/templates/react-app/src/webSocket.ts
Martinsos: Ok did some more research:

Rules from React docs:

> Only Call Hooks at the Top Level
> Don’t call Hooks inside loops, conditions, or nested functions. Instead, always use Hooks at the top level of your React function, before any early returns. By following this rule, you ensure that Hooks are called in the same order each time a component renders. That’s what allows React to correctly preserve the state of Hooks between multiple useState and useEffect calls. (If you’re curious, we’ll explain this in depth [below](https://legacy.reactjs.org/docs/hooks-rules.html#explanation).)
> 
> Only Call Hooks from React Functions
> Don’t call Hooks from regular JavaScript functions. Instead, you can:
> 
> ✅ Call Hooks from React function components.
> ✅ Call Hooks from custom Hooks (we’ll learn about them [on the next page](https://legacy.reactjs.org/docs/hooks-custom.html)).
> By following this rule, you ensure that all stateful logic in a component is clearly visible from its source code.

By having this extra step where we return a hook from another function (although it is hook), I think we violate these rules.

@Zeko369 also showed on Discord that this can break easily actually, so that confirms it: https://discord.com/channels/686873244791210014/1080864617347162122/1114104310645669928 .

Martinsos: Start of Discord convo: https://discord.com/channels/686873244791210014/1080864617347162122/1113918843774054440 .

waspc/examples/todoApp/src/shared/webSocket.ts
Martinsos: Question -> for every usage of `useSocket`, they will have to provide these types, right? Isn't that a bit cumbersome?

What if we instead let them provide those types to Wasp, and then Wasp will make sure to pre-type the `useSocket`, so they don't have to specify these types on it?

@infomiho ?

waspc/data/Generator/templates/react-app/src/webSocket.ts
shayneczyzewski: Pulled up to top level 👍🏻

waspc/examples/todoApp/src/client/pages/ProfilePage.tsx
```diff
@@ -16,14 +16,14 @@ export const ProfilePage = ({
...
-  const { socket, isConnected, registerHandler } = useSocket<ServerToClientEvents, ClientToServerEvents>()
+  const { socket, isConnected } = useSocket<ServerToClientEvents, ClientToServerEvents>()
   const inputRef = useRef<HTMLInputElement>(null)
 
   useEffect(() => {
     fetchCustomRoute()
   }, [])
 
-  registerHandler('chatMessage', logMessage)
+  useSocketListener(socket, 'chatMessage', logMessage)
```
Martinsos, line 26: Does it have to receive socket? Socket is global and known, so it could just obtain it on its own?

shayneczyzewski: This is what we use to infer the type. It either has to take a socket (as suggested in the Discord chat) or we have to accept the type (like I had before). Pulling off socket feels cleaner to me.

waspc/data/Generator/templates/react-app/src/webSocket.ts
Martinsos: The only thing I am curious about now is, if we actually return it from a useSocket hook -> so if useSocket hook returns useSocketListener hook, which is weird, but -> does that mean that useSocketListener hook has some extra powers, like has the types inferred? Although there is also that idea that I mentioned above, that all the types could be already provided by Wasp in the "backstage" if we registered them with Wasp, since we only have one socket and we actually know the types for it. Hmmmmmmmmmmm.

Martinsos: If we provided types to Wasp, in Wasp DSL, via ext imports, then Wasp could, in generated Wasp code, specify that these functions have those types!

waspc/examples/todoApp/src/client/pages/ProfilePage.tsx
Martinsos: Ok, that does make sense. Also it is kind of clear from the code the connection between the two.

waspc/data/Generator/templates/react-app/src/webSocket.ts
shayneczyzewski: Would that be kinda clunky from a DX perspective to have 4 imports (presumably so we can differentiate them regardless of type name)?

Martinsos: Maybe. They could maybe also be provided via JS, via some kind of "hook" (not React hook) and then you import (in Wasp) just that one thing, if we feel bad about having too many ext imports in Wasp. But it improves DX of `useSocket` potentially, because you don't have to provide types there.
On the other hand, they could create their own hook, useSocketsTyped which just calls useSocket with types and they can then use that everywhere. Hm.

infomiho:
I'll be reviewing this next week one more time 👍

waspc/data/Generator/templates/react-app/src/api.ts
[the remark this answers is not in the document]
infomiho: I would consider wrapping the `EventEmitter` to:
- allow us to type allowed events
- abstract away that we are using `EventEmitter` 

Something like this:
```ts
export const EventType = {
  SET_AUTH_TOKEN: 'authToken.set',
  CLEAR_AUTH_TOKEN: 'authToken.clear',
} as const;

// 'authToken.set' | 'authToken.clear'
type EventType = typeof EventType[keyof typeof EventType]

function createPubSub() {
  const events = new EventEmitter();

 return {
    emit(event: EventType): void {
      events.emit(event);
    },
    on(event: EventType, handler: () => void): void {
      events.on(event, handler) 
   }
  }
}

export const events = createPubSub();
```

waspc/data/Generator/templates/react-app/src/api.ts
```diff
@@ -40,6 +55,23 @@ api.interceptors.response.use(undefined, (error) => {
   return Promise.reject(error)
 })
 
+// This handler will run on other tabs (not the active one calling API functions),
+// and will ensure they know about auth token changes.
+// Ref: https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event
+// "Note: This won't work on the same page that is making the changes — it is really a way
+// for other pages on the domain using the storage to sync any changes that are made."
+window.addEventListener('storage', (event) => {
```
infomiho, line 63: This is really cool API, I didn't know it existed 👍

waspc/data/Generator/templates/react-app/src/storage.ts
```diff
@@ -1,4 +1,5 @@
 export type DataStore = {
+    prefixedKey(key:string): string;
```
infomiho, line 2:
```suggestion
    getPrefixedKey(key: string): string;
```

Since it's a function let's use a verb

waspc/data/Generator/templates/react-app/src/webSocket.ts
```diff
@@ -0,0 +1,76 @@
...
+    socket.disconnect()
+    socket.connect()
+  }
+}
+
+refreshAuthToken()
+apiEvents.on(EventType.SET_AUTH_TOKEN, refreshAuthToken)
+apiEvents.on(EventType.CLEAR_AUTH_TOKEN, refreshAuthToken)
+
+export function useSocket<
```
infomiho, line 31: I still think we can achieve the following (maybe now, maybe in v2):
- this file imports the user defined types from the server file (like we did for operations)
- it glues these types to this fn and the user gets full stack type safety

IMHO, if we have it for operations, it would be a missed opportunity not to do it here.

I can attempt to do this and I'll make a commit if it works well.

waspc/data/Generator/templates/react-app/src/webSocket.ts
```diff
@@ -0,0 +1,76 @@
...
+  ServerToClientEvents extends EventsMap = DefaultEventsMap,
+  ClientToServerEvents extends EventsMap = DefaultEventsMap
+>(): {
+  socket: Socket<ServerToClientEvents, ClientToServerEvents>;
+  isConnected: boolean;
+} {
+  const [isConnected, setIsConnected] = useState(socket.connected)
+
+  useEffect(() => {
```
infomiho, line 40: I think there is an opportunity to use context here:
- define a `WebSocketContext`
- wrap our app in the `WebSocketProvider`
    ```jsx
    <WebSocketProvider value={{ scoket, isConnected }}>
      {children}
    <WebSocketProvider>
    ```
- keep the state in the `WebSocketContext`
- `useSocket` gets it's socket from there (kinda like global `useState`)
    ```ts
    const { socket } = useContext(WebSocketContext);
    ```
- then ... `useSocketListener` can use `useSocket` under the hood and the user doesn't even need to think about providing the `socket` to `useSocketListener`

TLDR: we maybe can reduce the API of `useSocketListener` to `event` + `cb`

I can try to rewrite this to that form and commit if it works well.

waspc/data/Generator/templates/server/src/webSocket.ts
```diff
@@ -0,0 +1,69 @@
...
+    context: {
+      entities: {
+        {=# entities =}
+        {= name =}: typeof prisma.{= prismaIdentifier =},
+        {=/ entities =}
+      }
+    }
+  ) => Promise<void> | void
+
+type ServerType = Parameters<typeof {= userWebSocketFn.importIdentifier =}> [0]
```
infomiho, line 32:
```suggestion
type ServerType = Parameters<typeof {= userWebSocketFn.importIdentifier =}>[0]
```

infomiho (commented):
I have two suggestions:
- using context to simplify dealing with the `socket`
- maybe implement full-stack type safety so user gets the types on FE from the type he defined in the server handler

waspc/data/Generator/templates/react-app/src/webSocket.ts
Martinsos: I think it is worth giving it a try, for sure!

waspc/data/Generator/templates/react-app/src/webSocket.ts
Martinsos: Sounds interesting!

Btw I think this could work even without us using `context` though, right? I mean we can do all this right now, by using `useSocket` in `useSocketListener`? What do we get with `context` that we don't have now? I am not against `context`, just wondering.

waspc/data/Generator/templates/react-app/src/api.ts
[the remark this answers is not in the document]
Martinsos: Interesting -> although this look exactly the same as event emitter, interface wise? So I am not sure we got a lot here regarding abstracting away, it is the same interface really.

I would say the only thing we got are types. But isn't there a typed event emitter out there in the world, do we really have to write our own? Should we maybe just name it TypedEventEmitter then and have that in the codebase?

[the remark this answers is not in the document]
infomiho: > Interesting -> although this look exactly the same as event emitter, interface wise? So I am not sure we got a lot here regarding abstracting away, it is the same interface really.

If we wanted to swap out the lib we are using, it would be easier :)

This `events` lib is older and it wasn't written with Typescript in mind. Maybe we can find a different one? Or be without types, that's okay with me since it's a limited use case.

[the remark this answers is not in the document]
Martinsos: I think this events lib comes is builtin? But you know what, I am ok with any approach here, whatever you choose, including what you suggested above.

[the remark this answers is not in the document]
infomiho: I think this doesn't work since `EventEmitter` is not available in the browser. It _seems_ to work in our code because we installed `@types/node`, but this code shouldn't work in any browser.

I'll see if we can use a different library/API

web/docs/guides/websockets.md
```diff
@@ -4,67 +4,48 @@ title: WebSockets
...
 app todoApp {
   // ...
 
   webSocket: {
-    fn: import { webSocketFn } from "@server/webSocket.js"
-    // Optional: entities: [...]
-    // Optional: autoConnect: <bool>
-  },,
+    fn: import { webSocketFn } from "@server/webSocket.js",
+    entities: [], // optional
```
Martinsos, line 28: I would consider just passing all the entities here automatically, and not having user to specify them. Because we have one global websocket, for whole app, so I don't see much use in explicitly mentioning which entities it uses.
If not now, we can possibly do it in the future, but I think it makes sense to do from the DX side.

web/docs/guides/websockets.md
```diff
@@ -75,55 +56,95 @@ export const webSocketFn: WebSocketDefinition<
...
+  const [messages, setMessages] = useState<
+    { id: string; username: string; text: string }[]
+  >([])
+  // The "socket" instance is typed with the types you defined on the server.
+  const { socket, isConnected } = useSocket()
 
-  useSocketListener(socket, 'chatMessage', logMessage)
+  // This is a type-safe event handler: "chatMessage" event and its payload type
+  // are defined on the server.
+  useSocketListener('chatMessage', logMessage)
```
Martinsos, line 111: Nice, it is all already typed, awesome.

In the future I could imagine us wanting to define different sockets, that work with different sets of events -> but ok, that is for the future. Might be worth creating an issue for it, if we have good idea what to write. If you have some ideas here, pls open an issue, if not nvm.

Martinsos (approved):
LGTM, left some comments but overall looking great!
Will let @sodic do the final approval.

web/docs/guides/websockets.md
infomiho: > I would consider just passing all the entities here automatically, and not having user to specify them. 

That's a good point, I remember us talking about it. 

I'll make the changes to receive all the available entities inside of context.

waspc/examples/todoApp/src/client/pages/ProfilePage.tsx
```diff
@@ -13,10 +14,38 @@ export const ProfilePage = ({
 }: {
   user: User
 }) => {
+  const [messages, setMessages] = useState<
+    { id: string; username: string; text: string }[]
+  >([])
+  const { socket, isConnected } = useSocket()
```
sodic, line 20: I'm very happy that this supports full stack type safety, that was my first suggestion after seeing the example in the PR's description. Nice work to everyone involved!

waspc/examples/todoApp/src/server/webSocket.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export interface ServerToClientEvents {
+  chatMessage: (msg: { id: string; username: string; text: string }) => void
+}
+
+export interface ClientToServerEvents {
+  chatMessage: (msg: string) => void
+}
+
+export interface InterServerEvents {}
```
sodic, line 12: Why are we exporting these types? Does making them private break full-stack type safety? I've removed the exports and everything seems to be working fine.

waspc/data/Generator/templates/server/src/webSocket.ts
sodic, line 1: Can we format this file (after generation) and apply the formatting here?

waspc/data/Generator/templates/server/src/webSocket.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export interface WaspSocketData {
+  user?: any;
+}
+
+type WebSocketFn = typeof {= userWebSocketFn.importIdentifier =}
+type ServerType = Parameters<WebSocketFn>[0]
+type Events = ServerType extends Server<infer ClientToServerEvents, infer ServerToClientEvents>
+  ? [ClientToServerEvents, ServerToClientEvents]
+  : [DefaultEventsMap, DefaultEventsMap]
```
sodic, line 40: How come we're handling this case? Won't the code break (in runtime) if those types aren't correct anyway?

waspc/data/Generator/templates/react-app/src/webSocket.ts
```diff
@@ -0,0 +1,32 @@
...
+  event: Event,
+  handler: ServerToClientEvents[Event],
+) {
+  const { socket } = useContext(WebSocketContext);
+  useEffect(() => {
+    // TODO(miho): This is a hack for type errors we are getting
+    // from socket.on when passing in "handler" directly
+    const handlerInstance: any = (event: any) => {
+      handler(event);
+    };
```
sodic, line 26: Can we elaborate on the error and list a proper fix (if we know it)?

waspc/data/Generator/templates/react-app/src/storage.ts
```diff
@@ -7,6 +8,7 @@ export type DataStore = {
 
 function createLocalStorageDataStore(prefix: string): DataStore {
     return {
+        getPrefixedKey(key) { return `${prefix}:${key}`; },
```
sodic, line 11: Can we refactor the other functions to use this function for generating prefixed keys? I recommend defining them outside of the returned object (to avoid dealing with `this`).

waspc/data/Generator/templates/react-app/src/webSocket.ts
```diff
@@ -0,0 +1,32 @@
+import { useContext, useEffect } from 'react'
+import { WebSocketContext } from './webSocket/WebSocketProvider';
+
+import type {
+  ClientToServerEvents as _ClientToServerEvents,
+  ServerToClientEvents as _ServerToClientEvents
+} from '../../server/src/webSocket'
```
sodic, line 7: Why are we aliasing the types, and then reassigning them to their original names?

waspc/src/Wasp/Generator/ServerGenerator/WebSocketG.hs
```diff
@@ -0,0 +1,70 @@
...
+  where
+    maybeWebSocket = AS.App.webSocket $ snd $ getApp spec
+
+mkWebSocketFnImport :: Maybe WebSocket -> Aeson.Value
+mkWebSocketFnImport maybeWebSocket =
+  let maybeWebSocketFn = AS.App.WS.fn <$> maybeWebSocket
+   in extImportToImportJson relPathToServerSrcDir maybeWebSocketFn
+  where
+    relPathToServerSrcDir :: Path Posix (Rel importLocation) (Dir C.ServerSrcDir)
+    relPathToServerSrcDir = [reldirP|./|]
```
sodic, line 70: Since we already have a `where` block, let's put `maybeWebSocketFn` inside as well.

waspc/src/Wasp/Generator/WebAppGenerator.hs
```diff
@@ -287,3 +301,22 @@ genEnvValidationScript =
...
+          genWebSocketProvider spec
+        ]
+  | otherwise = return []
+  where
+    genFileCopy = return . C.mkSrcTmplFd
+
+genWebSocketProvider :: AppSpec -> Generator FileDraft
+genWebSocketProvider spec = return $ C.mkTmplFdWithData tmplFile tmplData
+  where
+    maybeWebSocket = webSocket (snd (getApp spec))
```
sodic, line 319: Nit: I'd probably use `$` here.

sodic (commented):
Great work! Left several comments, nothing major.

waspc/data/Generator/templates/react-app/src/webSocket.ts
infomiho: Not doing it this way breaks the "import only types from server" magic trick we are using for some reason. Vite starts complaining about *server-side code*

infomiho: I'm not sure if there is something smarter we could be doing to avoid this.

waspc/data/Generator/templates/react-app/src/webSocket.ts
infomiho: I get the following error:
```
Argument of type 'ServerToClientEvents[Event]' is not assignable to parameter of type 'FallbackToUntypedListener<Event extends "disconnect" | "connect" | "connect_error" ? SocketReservedEvents[Event] : Event extends "chatMessage" ? ServerToClientEvents[Event] : never>'.
  Type '(msg: { id: string; username: string; text: string; }) => void' is not assignable to type 'FallbackToUntypedListener<Event extends "disconnect" | "connect" | "connect_error" ? SocketReservedEvents[Event] : Event extends "chatMessage" ? ServerToClientEvents[Event] : never>'.
```
I haven't been able to figure out why that happens after digging through socket-io's source code.

And I haven't been able to find a fully working Typescript socket.io example, but I did find this workaround.

infomiho: Screenshot 2023-06-16 at 12 22 53

waspc/data/Generator/templates/server/src/webSocket.ts
infomiho: Sure, that's a good idea 👍

waspc/data/Generator/templates/server/src/webSocket.ts
infomiho: I was thinking of users that don't use `WebSocketDefinition` in their definition or use Javascript. 

Let me test it a bit more and see if it's needed 👍

infomiho: I just realised I needed this check anyways to `infer` the `ClientToServerEvents` and `ServerToClientEvents` types. And in the case it doesn't infer, just return the default ones. Does this make sense?

infomiho: Screenshot 2023-06-16 at 12 39 43
Screenshot 2023-06-16 at 12 39 47
Screenshot 2023-06-16 at 12 40 00

waspc/data/Generator/templates/react-app/src/webSocket.ts
infomiho: I've added a more detailed comment 👍

waspc/data/Generator/templates/react-app/src/webSocket.ts
infomiho: It seems that exporting the types from this file directly works as well 👍

waspc/examples/todoApp/src/client/pages/ProfilePage.tsx
```diff
@@ -13,10 +14,38 @@ export const ProfilePage = ({
 }: {
   user: User
 }) => {
+  const [messages, setMessages] = useState<
+    { id: string; username: string; text: string }[]
```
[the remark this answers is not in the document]
infomiho: That would be nice, the user could hack it around with getting the ServerToClient events and using the `Parameters`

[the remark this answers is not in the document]
infomiho: We agreed to not doing anything fancy for now 👍 

Users can define their payload types in the `shared` dir and import them if they wish on the client and the server.

waspc/src/Wasp/Generator/WebAppGenerator.hs
sodic: ... For the `getApp $ spec` part as well. Still irrelevant, you can resolve this one as you see fit.

waspc/data/Generator/templates/react-app/src/webSocket.ts
```diff
@@ -0,0 +1,44 @@
+import { useContext, useEffect } from 'react'
+import { WebSocketContext } from './webSocket/WebSocketProvider';
+
+export type {
+  ClientToServerEvents,
+  ServerToClientEvents,
+} from '../../server/src/webSocket'
+
+import type { ServerToClientEvents } from '../../server/src/webSocket'
```
sodic, line 9: Just please move the `import` to the top of the file.

waspc/data/Generator/templates/react-app/src/webSocket.ts
sodic: How come you didn't use `as keyof ServerToClientEvents` in the end?

web/docs/typescript.md
```diff
@@ -418,4 +418,100 @@ import type { GetAllQuery, GetQuery, CreateAction, UpdateAction, DeleteAction }
...
+}
+```
+
+### Using the WebSocket on the client
+
+After you have defined the WebSocket function on the server, you can use it on the client. The `useSocket` hook will give you the `socket` instance and the `isConnected` boolean. The `socket` instance is typed with the types you defined on the server.
+
+The `useSocketListener` hook will give you a type-safe event handler. The event name and its payload type are defined on the server.
+
+You can additonally use the `ClientToServerPayload` and `ServerToClientPayload` helper types to get the payload type for a specific event (it extract the message type from the event handler type).
```
sodic, line 472: I'd remove the stuff in parentheses, they don't need to know how it does it

waspc/data/Generator/templates/server/src/webSocket.ts
sodic, line 1: Feels a little strange to have a folder called `webSocket` and a file called `webSocket.ts` at the same level (we're not in Haskell, after all :). Maybe we can call this one `webSocket/index.ts` or something?

waspc/data/Generator/templates/react-app/src/webSocket.ts
infomiho: I've forgot about that change 🤦‍♂️ I did it in the generated code and not in the template.

infomiho: Updated the code to use the cast 👍

PR #1209, opened 2023-05-18
Adds Wasp and Prisma syntax highlighting

infomiho:
- adds PrismJS language definition for Wasp and Prisma (taken from their docs)
- it changes the syntax highlighting theme to `github` (it highlights more tokens)
Before:
Screenshot 2023-05-18 at 16 48 03

After:
Screenshot 2023-05-18 at 16 48 06
- changes syntax highlighting language in the docs from`c` and similar to `wasp`
- fixes some minor issues

### DIFF
--- a/web/docs/guides/auth-ui.md
+++ b/web/docs/guides/auth-ui.md
@@ -14,7 +14,7 @@ Based on your `main.wasp` file on the authentication providers you enabled, the 
 
 For example, if you only enabled e-mail authentication:
 
-```c title="main.wasp"
+```wasp title="main.wasp"
 app MyApp {
   title: "My app",
   //...
@@ -33,7 +33,7 @@ We'll get this:
 
 And then we enable Google and Github:
 
-```c title="main.wasp" {7-8}
+```wasp title="main.wasp" {7-8}
 app MyApp {
   title: "My app",
   //...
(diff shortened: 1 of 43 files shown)


web/docs/guides/auth-ui.md
```diff
@@ -60,7 +60,7 @@ Let's take a look at the components that are available for you to use.
 
 ### Login form
 
-Useful for <span style={{ background: '#fee', padding: '0.2rem 0.5rem', borderRadius: '0.375rem' }}>username & password</span> and <span style={{ background: '#eef', padding: '0.2rem 0.5rem', borderRadius: '0.375rem' }}>email</span> authentication.
```
infomiho, line 63: Uses a class instead of inline CSS. Fixes dark mode colors.

web/src/theme/prism-include-languages.js
```diff
@@ -0,0 +1,22 @@
+import siteConfig from "@generated/docusaurus.config";
```
infomiho, line 1: This file gets autogenerated when you "eject" to add custom languages to Docosaurus

Martinsos: Probably best to add this comment directly to this file!

Martinsos (approved):
@infomiho great work, looks good to me! This will be very nice for our syntax highlighting :).

It would be cool if we could have kept the dark code though, but ok, if this gives better highlighting then I guess it is worth it.

I left just one comment, everything else looks fine.

I do wonder how are we going to remember to keep these language definitions up to date hm, that could be a bit tricky. Maybe leave some comments somewhere? Or we can just let it be and hopefully, we will remember, that is ok also ok for now I think, we can fix the problem once it happens.

web/src/theme/prism-include-languages.js
infomiho: I'll move it there 👍

infomiho:
@Martinsos I'll add a note in the https://github.com/wasp-lang/vscode-wasp repo to make sure to update the Wasp syntax highlighting for the web when updating the VS Code syntax as well. I think it will be a good reminder and it should work for us since we already have the habit of updating the docs / VS Code syntax highlighting when adding new features.

PR #1210, opened 2023-05-18
Improve waspc node version checking and error messages

### LINKED ISSUE #1134: Ensure that Wasp nicely warns if `node` is not installed / usable

craigmc08:
### Description

- [X] Fixes weird IO error when node is missing on MacOS by changing minimum version of `process` (see discussion in #1134)
- [X] Improves error message if more weird IO errors happen when checking node's version
- [X] Improves error messages for node versions
- [X] Make node version check happen earlier and only once per CLI run

Fixes #1134

### DIFF
--- a/waspc/cli/exe/Main.hs
+++ b/waspc/cli/exe/Main.hs
@@ -31,8 +31,10 @@ import qualified Wasp.Cli.Command.Telemetry as Telemetry
 import Wasp.Cli.Command.Test (test)
 import Wasp.Cli.Command.Uninstall (uninstall)
 import Wasp.Cli.Command.WaspLS (runWaspLS)
+import Wasp.Cli.Message (cliSendMessage)
 import Wasp.Cli.Terminal (title)
 import qualified Wasp.Generator.Node.Version as NodeVersion
+import qualified Wasp.Message as Message
 import Wasp.Util (indent)
 import qualified Wasp.Util.Terminal as Term
 import Wasp.Version (waspVersion)
@@ -70,7 +72,7 @@ main = withUtf8 . (`E.catch` handleInternalErrors) $ do
   -- See https://github.com/wasp-lang/wasp/issues/1134#issuecomment-1554065668
   NodeVersion.checkNodeVersion >>= \case
     Left errorMsg -> do
-      putStrLn $ Term.applyStyles [Term.Red] errorMsg
+      cliSendMessage $ Message.Failure "Node requirement not met" errorMsg
       exitFailure
     Right _ -> pure ()
 
--- a/waspc/cli/exe/Main.hs
+++ b/waspc/cli/exe/Main.hs
@@ -70,7 +70,7 @@ main = withUtf8 . (`E.catch` handleInternalErrors) $ do
   -- not needed for every command, but checking for every command was decided
   -- to be more robust than trying to only check for commands that require it.
   -- See https://github.com/wasp-lang/wasp/issues/1134#issuecomment-1554065668
-  NodeVersion.checkNodeVersion >>= \case
+  NodeVersion.getAndCheckNodeVersion >>= \case
     Left errorMsg -> do
       cliSendMessage $ Message.Failure "Node requirement not met" errorMsg
       exitFailure
--- a/waspc/cli/exe/Main.hs
+++ b/waspc/cli/exe/Main.hs
@@ -8,6 +8,7 @@ import Data.Char (isSpace)
 import Data.List (intercalate)
 import Main.Utf8 (withUtf8)
 import System.Environment (getArgs)
+import System.Exit (exitFailure)
 import Wasp.Cli.Command (runCommand)
 import Wasp.Cli.Command.BashCompletion (bashCompletion, generateBashCompletionScript, printBashCompletionInstruction)
 import Wasp.Cli.Command.Build (build)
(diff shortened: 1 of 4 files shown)


infomiho:
1. Running with the wrong Node version
```
Your node version does not meet Wasp's requirements! You are running node 16.20.0.
Wasp requires node ^18.12.0 to be installed and in PATH. Check Wasp docs for more details: https://wasp-lang.dev/docs/quick-start#requirements.
```

-> I've noticed we didn't use the waspScreams format we usually use for errors
2. Running without Node
```
Running 'node --version' failed:
  `node` command not found!
Wasp requires node ^18.12.0 to be installed and in PATH. Check Wasp docs for more details: https://wasp-lang.dev/docs/quick-start#requirements.
```
-> Looking good!

waspc/cli/exe/Main.hs
```diff
@@ -62,6 +66,16 @@ main = withUtf8 . (`E.catch` handleInternalErrors) $ do
 
   telemetryThread <- Async.async $ runCommand $ Telemetry.considerSendingData commandCall
 
+  -- Before calling any command, check that the node requirement is met. Node is
+  -- not needed for every command, but checking for every command was decided
+  -- to be more robust than trying to only check for commands that require it.
+  -- See https://github.com/wasp-lang/wasp/issues/1134#issuecomment-1554065668
```
sodic, line 72: Nice comment!

waspc/src/Wasp/Generator/Node/Version.hs
```diff
@@ -13,25 +14,56 @@ import qualified System.Process as P
...
+import Wasp.Util (indent)
 
+-- | Gets the installed node version, if any is installed, and checks that it
+-- meets Wasp's version requirement.
+--
+-- Returns a string representing the error
+-- condition if node's version could not be found or if the version does not
+-- meet the requirements.
+checkNodeVersion :: IO (Either String SV.Version)
+checkNodeVersion =
```
sodic, line 26: Perhaps a more informative name is `getNodeVersionOrErrorMsg` (or something along those lines).

waspc/src/Wasp/Generator/Node/Version.hs
```diff
@@ -49,11 +81,21 @@ parseNodeVersion nodeVersionStr =
       return $ SV.Version mjr mnr ptc
     _ -> Nothing
 
+nodeNotFoundMessage :: String
+nodeNotFoundMessage = "`node` command not found!"
+
+nodeUnknownErrorMessage :: IOError -> String
+nodeUnknownErrorMessage err =
```
sodic, line 88: Maybe `makeUnknownNodeErrorMessage` since we're dealing with a function? I haven't programmed in Haskell for a while and forgot most of the conventions, so correct me if I'm saying nonsense.

sodic (approved):
Left a couple of stylistic comments, but generally all good.

Nice work!

waspc/src/Wasp/Generator/Node/Version.hs
craigmc08: That makes sense, it will make it consistent with `makeNodeVersionMismatchMessage` at the very least.

waspc/src/Wasp/Generator/Node/Version.hs
```diff
@@ -13,25 +14,56 @@ import qualified System.Process as P
...
 import qualified Wasp.SemanticVersion as SV
+import Wasp.Util (indent)
 
+-- | Gets the installed node version, if any is installed, and checks that it
+-- meets Wasp's version requirement.
+--
+-- Returns a string representing the error
+-- condition if node's version could not be found or if the version does not
+-- meet the requirements.
+getAndCheckNodeVersion :: IO (Either String SV.Version)
```
Martinsos, line 25: I am a bit late to the party, but one question here is: is this a right place to define this stuff? This is in Generator, and it makes sense because it was only Genreator that required node. But now, we are importing this logic in the rest of the app, in CLI, and using it there. We could say: that is because CLI now explicitly depends on this specific Generator! But maybe what would be nicer is extracting Node logic into higher level, something like src/Wasp/Node.hs .
Specifically, extracting the concept of Node version on which Wasp depends on, and also extracting the checks for it, that might make sense -> extracting it into src/Wasp/Node.hs as I mentioned.

I am not 100% sure about this, but it feels like the right thing to do. This way we say "not only generator depends on node, but whole Wasp", and that is really true, because Wasp even uses node to do deployment via `wasp deploy`. This is relatively new also, and we never captured that really in this way.

@sodic @craigmc08 let me know what you think.

sodic: We could move it, makes sense if @craigmc08 is up for it?

Martinsos: No problem, let's create new PR!

PR #1216, opened 2023-05-23
Auth UI component refactor

infomiho:
- breaks down the huge `Auth.tsx` component into bunch of smaller components

### TODO
- [x] conditionally generate Typescript enum for possible Auth component states
- [ ] ~use React Query for login, signup, reset password, email verification actions?~


waspc/src/Wasp/Generator/WebAppGenerator/Auth/AuthFormsG.hs
```diff
@@ -0,0 +1,128 @@
...
+    [ genAuthComponent auth,
+      genTypes auth,
+      copyTmplFile [relfile|auth/forms/Login.tsx|],
+      copyTmplFile [relfile|auth/forms/Signup.tsx|],
+      copyTmplFile [relfile|stitches.config.js|]
+    ]
+    <++> genEmailForms auth
+    <++> genInternalAuthComponents auth
+  where
+    copyTmplFile = return . C.mkSrcTmplFd
```
craigmc08, line 29:
```suggestion
genAuthForms :: AS.Auth.Auth -> Generator [FileDraft]
genAuthForms auth =
  sequence
    [ genAuthComponent auth,
      genTypes auth,
	  genFileCopy [relfile|auth/forms/Login.tsx|],
	  genFileCopy [relfile|auth/forms/Signup.tsx|],
	  genFileCopy [relfile|stitches.config.js|]
    ]
    <++> genEmailForms auth
    <++> genInternalAuthComponents auth
  where
	genFileCopy = return . C.mkSrcTmplFd
```

To keep naming consistent with other generator files (for example https://github.com/wasp-lang/wasp/blob/main/waspc/src/Wasp/Generator/WebAppGenerator.hs#L76)

waspc/src/Wasp/Generator/WebAppGenerator/Auth/AuthFormsG.hs
```diff
@@ -0,0 +1,128 @@
...
+genEmailForms auth =
+  genConditionally isEmailAuthEnabled $
+    sequence
+      [ copyTmplFile [relfile|auth/forms/ResetPassword.tsx|],
+        copyTmplFile [relfile|auth/forms/ForgotPassword.tsx|],
+        copyTmplFile [relfile|auth/forms/VerifyEmail.tsx|]
+      ]
+  where
+    copyTmplFile = return . C.mkSrcTmplFd
+    isEmailAuthEnabled = AS.Auth.isEmailAuthEnabled auth
```
craigmc08, line 59:
```suggestion
genEmailForms :: AS.Auth.Auth -> Generator [FileDraft]
genEmailForms auth =
  genConditionally isEmailAuthEnabled $
    sequence
      [ genFileCopy [relfile|auth/forms/ResetPassword.tsx|],
        genFileCopy [relfile|auth/forms/ForgotPassword.tsx|],
        genFileCopy [relfile|auth/forms/VerifyEmail.tsx|]
      ]
  where
    genFileCopy = return . C.mkSrcTmplFd
    isEmailAuthEnabled = AS.Auth.isEmailAuthEnabled auth
```

To keep naming consistent with other generator files (for example https://github.com/wasp-lang/wasp/blob/main/waspc/src/Wasp/Generator/WebAppGenerator.hs#L76)

waspc/src/Wasp/Generator/WebAppGenerator/AuthG.hs
```diff
@@ -29,7 +26,7 @@ genAuth spec =
     Nothing -> return []
     Just auth ->
       sequence
-        [ copyTmplFile [relfile|auth/logout.js|],
+        [ copyTmplFile [relfile|auth/logout.ts|],
           copyTmplFile [relfile|auth/helpers/user.ts|],
           copyTmplFile [relfile|auth/types.ts|],
```
craigmc08, line 31: Rename `copyTmplFile` to `genFileCopy` to keep naming consistent with other generator files (for example https://github.com/wasp-lang/wasp/blob/main/waspc/src/Wasp/Generator/WebAppGenerator.hs#L76).

I would put a suggestion here like the places in AuthFormsG.hs but GH doesn't let me since some of the lines are unchanged.

craigmc08 (changes requested):
@infomiho Looks pretty good! I mostly have a couple naming/formatting nitpicks.

A couple more question that don't fit on individual files:

1. Naming the folder where the code for the implementation of the login forms `internal` feels a bit odd to me, maybe `lib`? But `internal` also signals to users not to import from there, so I'm not sure.

2. Should `auth/forms/Auth.tsx` in the react-app template be in `auth/forms/internal/`? It looks like something that users shouldn't use directly.

waspc/data/Generator/templates/react-app/src/auth/forms/internal/email/useEmail.ts
```diff
@@ -0,0 +1,49 @@
...
+  const [passwordFieldVal, setPasswordFieldVal] = useState('')
+
+  async function handleSubmit() {
+    try {
+      if (isLogin) {
+        await login({ email: emailFieldVal, password: passwordFieldVal })
+        onLoginSuccess()
+      } else {
+        await signup({ email: emailFieldVal, password: passwordFieldVal })
+        if (isEmailVerificationRequired) {
```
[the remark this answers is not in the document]
infomiho: I wanted to reduce the number of files that need to be templates. 

Here, we have a function called `useEmail` which is exported from `internal/email/useEmail.ts` and since it receives a param `isEmailVerificationRequired` there is not need for it to be a Mustache template. 

Every time we have a Mustache template, VS Code complains and doesn't color the Typescript nicely, so, the less we use then, the better :)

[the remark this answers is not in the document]
craigmc08: That makes sense!

waspc/data/Generator/templates/react-app/src/auth/forms/Auth.tsx
```diff
@@ -1,31 +1,18 @@
...
-type ErrorMessage = {
-  title: string;
-  description?: string;
-};
+import {
+  type State,
+  type CustomizationOptions,
+  type ErrorMessage,
+} from './types'
+import { LoginSignupForm } from './internal/common/LoginSignupForm'
```
matijaSos, line 11: nice breakdown

waspc/data/Generator/templates/react-app/src/auth/forms/Auth.tsx
```diff
@@ -42,590 +29,67 @@ const HeaderText = styled('h2', {
...
   // If we do it in useEffect(), then there is a glitch between the default color and the
   // user provided one.
   const customTheme = createTheme(appearance ?? {})
 
-  const cta = isLogin ? 'Log in' : 'Sign up'
   const titles: Record<State, string> = {
     login: 'Log in to your account',
     signup: 'Create a new account',
+    {=# isEmailAuthEnabled =}
     "forgot-password": "Forgot your password?",
```
matijaSos, line 56: nice, you introduced labels. Is that something users can access already and customize? I imagine sth like this: https://supabase.com/docs/guides/auth/auth-helpers/auth-ui#custom-labels

But we don't need to have it right now ofc. Maybe open an issue for it, might be a nice first issue for somebody

waspc/data/Generator/templates/react-app/src/auth/forms/Auth.tsx
```diff
@@ -42,590 +29,67 @@ const HeaderText = styled('h2', {
...
+      {successMessage && <MessageSuccess>{successMessage}</MessageSuccess>}
+      <AuthContext.Provider value={{ isLoading, setIsLoading, setErrorMessage, setSuccessMessage }}>
+        {(state === 'login' || state === 'signup') && (
+          <LoginSignupForm
+            state={state}
+            socialButtonsDirection={socialButtonsDirection}
+          />
+        )}
+        {=# isEmailAuthEnabled =}
+        {state === 'forgot-password' && (<ForgotPasswordForm />)}
```
matijaSos, line 86: is this some TS thing when you are using string but it's actually a type (State)? So you can't put any string as a state, compiler would complain?

waspc/data/Generator/templates/react-app/src/auth/forms/Auth.tsx
```diff
@@ -42,590 +29,67 @@ const HeaderText = styled('h2', {
...
-  }
-  {=/ isAnyPasswordBasedAuthEnabled =}
 
   // TODO(matija): this is called on every render, is it a problem?
   // If we do it in useEffect(), then there is a glitch between the default color and the
   // user provided one.
   const customTheme = createTheme(appearance ?? {})
 
-  const cta = isLogin ? 'Log in' : 'Sign up'
   const titles: Record<State, string> = {
```
matijaSos, line 52: aha so `titles` is used both as titles but also as a state of the component? That was a bit hard to follow at first. maybe I'd make it more explicit (although I'm not sure how).

matijaSos (commented):
very nice breakdown and making everything modular. Left a few questions but nothing major.

waspc/data/Generator/templates/react-app/src/auth/forms/Auth.tsx
infomiho: Yes sir 👍

waspc/data/Generator/templates/react-app/src/auth/forms/Auth.tsx
infomiho: > aha so titles is used both as titles but also as a state of the component? That was a bit hard to follow at first. maybe I'd make it more explicit (although I'm not sure how).

This says: "based on the State define a title" which means, for the "login" state, the title should be "Login" and so on. It's a common thing in Typescript to do it this way with `Record<SomeEnum, string>` which forces you to define the titles for each state (exhaustive pattern matching, similar to Haskell)

waspc/data/Generator/templates/react-app/src/auth/forms/Auth.tsx
infomiho: Here's an issue for further deepening the customization options https://github.com/wasp-lang/wasp/issues/1251

PR #1218, opened 2023-05-23
Display error message when database is not running

### LINKED ISSUE #1051: During development, produce a nice error message for user if database is not running but it should be
When development database is not running and user runs a Wasp command that requires database (like `wasp start` or `wasp db migrate-dev`), right now they will get a bit different error messages each time, coming from different source -> whoever failed to connected to the db. It might come from Server, from Prisma, ... .

Instead, it would be nice if we had more control over it and if we could print a nice error message that informs them that db is not running and how they can run it ("You can run `wasp start db`!").

Good way to do this could be to run `prisma db execute` with `SELECT 1` query and just check if it returned successfully. We could then run this check at the start of any Wasp command that requires db to be running, and if it fails, print nice error message and abort the rest of the wasp command. It is good to do this via prisma because then we don't care if it is sqlite or postgres or whatever.

craigmc08:
### Description

Adds an error message that is shown immediately when a wasp command is run if that command needs access to the database and the database is not running.

Additionally, adds a more generic framework to the `Command` monad that can be used for checking for other external dependencies (for example, `node`).

Fixes #1051 


waspc/cli/src/Wasp/Cli/Command/Requires.hs
```diff
@@ -0,0 +1,71 @@
...
+-- want it to be (e.g. marking a requirement as met without actually checking
+-- it).
+class Monad m => MonadRequires m where
+  -- | @checkRequirement requirement check@ runs @check@ if the requirement
+  -- has not been met yet. @check@ should return whether the requirement is
+  -- met.
+  --
+  -- @checkRequirement requirement check@ returns @True@ when the requirement
+  -- is met, and will not run the check if the requirement has already been
+  -- checked and met.
```
Martinsos, line 58: Seems like you duplicated the comment here, kind of?

I think it would be enough to just say:
```
@checkRequirement requirement doCheck@ returns True if specified requirement is met, otherwise False. It will run `doCheck` to check this, unless it already knows from before that requirement is met.
```

waspc/cli/src/Wasp/Cli/Command/Requires.hs
```diff
@@ -0,0 +1,71 @@
...
+runRequiresT :: Monad m => RequiresT m a -> m a
+runRequiresT m = evalStateT (unRequiresT m) []
+
+-- | The Requires monad keeps track of which 'CommandRequirements' have been
+-- checked and met.
+--
+-- NOTE: We go through the trouble of creating this abstraction so we can make
+-- sure the set of satisfied requirements is never modified in a way we do not
+-- want it to be (e.g. marking a requirement as met without actually checking
+-- it).
```
Martinsos, line 50: Although it might be convenient in some cases to mark requirement as met without going through the interface here -> maybe I just have that information from some other part of my program and don't want to run the explicit check again.

waspc/cli/src/Wasp/Cli/Command/Requires.hs
```diff
@@ -0,0 +1,71 @@
...
+data CommandRequirement
+  = -- | Requires that it is possible to connect to the prisma database.
+    DbConnection
+  deriving (Eq, Show)
+
+-- | A monad transformer that holds a list of satisfied 'CommandRequirement's
+-- that is updated whenever a new requirement is checked.
+--
+-- This is the canonical implementation of 'MonadRequires'.
+newtype RequiresT m a = RequiresT {unRequiresT :: StateT [CommandRequirement] m a}
```
Martinsos, line 38: If you went for making RequiresT more general, than you could make it take a type parameter that represents the requirements, instead of it being hardcoded like this.

waspc/cli/src/Wasp/Cli/Command/Requires.hs
Martinsos: What may be important to mention here is that this interface assumes that once a requirement is met, it stays met.

waspc/cli/src/Wasp/Cli/Command/Start.hs
```diff
@@ -26,6 +26,8 @@ start = do
   waspRoot <- findWaspProjectRootDirFromCwd
```
Martinsos, line 26: Hm, a thought -> could we use this Require mechanism to obtain the waspRoot easily?

Imagine that we could say that command requires to be in wasp project. That would be checked, but only that -> waspRoot would also be obtained.

I guess you would need to expand it a bit, so that we don't store just the fact that requirement is met, but also its "payload"/"value", but that shouldn't be hard.

I am not sure if we should do it though hm. Sometimes it is nice when commands are explicit about what they need in their arguments.

waspc/src/Wasp/Generator/DbGenerator/Operations.hs
```diff
@@ -134,6 +135,15 @@ dbSeed genProjectDir seedName = do
     ExitSuccess -> Right ()
     ExitFailure c -> Left $ "Failed with exit code " <> show c
 
+dbIsRunning ::
+  Path' Abs (Dir ProjectRootDir) ->
+  IO Bool
+dbIsRunning genProjectDir = do
+  chan <- newChan
+  exitCode <- DbJobs.dbExecuteTest genProjectDir chan
```
Martinsos, line 143: Seems like we don't really need a Job here, since we don't do anything with that Chan, right?
But I guess you used a Job because you had `runPrismaCommandAsDbJob`, which is very convenient? I guess it would be better if we had just normal `runPrismaCommand` or smth like that hm. But ok, we would still want to take over control over stdout possibly, hm. Ok, I guess if this works fine we can do it like this (although it feels like a tiny bit overkill, but I guess refactoring it would be even more of an overkill).

waspc/src/Wasp/Generator/DbGenerator/Operations.hs
```diff
@@ -134,6 +135,15 @@ dbSeed genProjectDir seedName = do
     ExitSuccess -> Right ()
     ExitFailure c -> Left $ "Failed with exit code " <> show c
 
+dbIsRunning ::
```
Martinsos, line 138:
```suggestion
isDbRunning ::
```

waspc/src/Wasp/Generator/DbGenerator/Operations.hs
```diff
@@ -134,6 +135,15 @@ dbSeed genProjectDir seedName = do
...
+  Path' Abs (Dir ProjectRootDir) ->
+  IO Bool
+dbIsRunning genProjectDir = do
+  chan <- newChan
+  exitCode <- DbJobs.dbExecuteTest genProjectDir chan
+  -- TODO: should we inspect the error it gives?
```
Martinsos, line 144: Ha good question. What do we do with it if we inspect it? I guess we would just propagate it all the way to user and say "hey, we failed to connect to the databse, and it returned this exit code".
So it depends, if you feel like there is a decent way to propagate this exit code back to user, we could go with that. If not and it feel unneccessary, then let's turn this TODO into a NOTE explaining our reasoning and it is fine.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
```diff
@@ -113,6 +114,19 @@ seed projectDir seedName =
     projectDir
     (const ["db", "seed"])
 
+-- | Runs `prisma db execute --stdin --schema <path to db schema>`, for checking
+-- if the DB is active. Runs the command in the generated server code directory
+-- so it has access to the database URL.
```
Martinsos, line 119:
```suggestion
-- | Checks if the DB is running and connectable by running
-- `prisma db execute --stdin --schema <path to db schema>`.
--  Runs the command in the generated server code directory so it has access to the database URL.
```

waspc/cli/src/Wasp/Cli/Command/Requires.hs
```diff
@@ -0,0 +1,71 @@
...
+
+    -- This module exports the lower-level parts of the command requirement
+    -- API. To use these features to establish a requirement for a command,
+    -- use 'Wasp.Cli.Command.Common.commandRequires'. For convenience,
+    -- 'CommandRequirement' and its constructors are also exported from that
+    -- module.
+    --
+    -- Call @commandRequires SomeRequirementConstructor@ anywhere in your command,
+    -- preferably near the beginning of the command implementation and definitely
+    -- before using the requirement.
```
Martinsos, line 14: I would try to make this comment a bit more friendly to read.

```suggestion
  ( -- * Command Requirements

    -- Command might have certain requirements that must be met before it runs.
    -- This module exports the lower-level parts of the command requirement
    -- API. To use these features to specify a requirement for a command,
    -- use 'Wasp.Cli.Command.Common.commandRequires'. For convenience,
    -- 'CommandRequirement' and its constructors are also exported from that
    -- module.
    --
    -- Call @commandRequires SomeRequirementConstructor@ anywhere in your command,
    -- preferably near the beginning of the command implementation and definitely
    -- before using the requirement.
```

Martinsos (commented):
Well @craigmc08 I see you went full Craig on this: I imagined we could solve this by having a function that checks this and we call it in the correct two places, but you did the whole monad transformer thing.

Taking into account that we could have likely solved this more easily, I think this is a bit of overengineering, but on the other hand I do like the result we get, could be useful maybe for some other stuff.

But in that case, if we are to keep it, we need to make it really good and easy to read and use. Because we will forget about it, come back to it, and we will not know how to find our way. So the simpler, cleaner and non-obstrusive we can make it, the better. It is already pretty good but I left some comments on how we could maybe improve it more.

Martinsos (commented):
Well @craigmc08 I see you went full Craig on this: I imagined we could solve this by having a function that checks this and we call it in the correct two places, but you did the whole monad transformer thing.

Taking into account that we could have likely solved this more easily, I think this is a bit of overengineering, but on the other hand I do like the result we get, could be useful maybe for some other stuff.

But in that case, if we are to keep it, we need to make it really good and easy to read and use. Because we will forget about it, come back to it, and we will not know how to find our way. So the simpler, cleaner and non-obstrusive we can make it, the better. It is already pretty good but I left some comments on how we could maybe improve it more.

waspc/cli/src/Wasp/Cli/Command/Start.hs
craigmc08: I think that's an interesting idea! I talk about your idea here in one of my other comments, since its a solution for a couple other issues.

But I agree that its a bit nicer when its more explicit, and I'm not sure if we should leave this Requires thing for requirements that don't have data associated with them (like with waspRoot).

waspc/cli/src/Wasp/Cli/Command/Requires.hs
```diff
@@ -0,0 +1,71 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+module Wasp.Cli.Command.Requires
+  ( -- * Using command requirements
+
+    -- This module exports the lower-level parts of the command requirement
+    -- API. To use these features to establish a requirement for a command,
+    -- use 'Wasp.Cli.Command.Common.commandRequires'. For convenience,
```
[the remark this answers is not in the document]
craigmc08: Agreed, it would ideally be exported from `Wasp.Cli.Command`. Let's discuss this in the thread in `Wasp.Cli.Command.Common`: https://github.com/wasp-lang/wasp/pull/1218/files#r1217807700

waspc/cli/src/Wasp/Cli/Command/Requires.hs
craigmc08: Hmm, true. I don't think its worth changing this yet if we don't have a situation where it would be useful to do this.

craigmc08 (commented):
> Taking into account that we could have likely solved this more easily, I think this is a bit of overengineering, but on the other hand I do like the result we get, could be useful maybe for some other stuff.

Definitely a bit overengineered, but I was thinking about the node version check at the same time and thought that since we have 2 requirements like this, it might be worth creating some abstraction for it. It will also be useful for if more system requirements are needed in the future.

waspc/cli/src/Wasp/Cli/Command.hs
```diff
@@ -27,3 +43,88 @@ runCommand cmd = do
...
+data Requirement where
+  Requirement :: Requirable r => r -> Requirement
+
+class Typeable r => Requirable r where
+  -- | Check if the requirement is met and return a value representing that
+  -- requirement.
+  --
+  -- This function must always return a value: if the requirement is not met,
+  -- throw a 'CommandError'.
+  checkRequirement :: Command r
```
craigmc08, line 56: I'm unsure about whether this type class should be exported from this module or not: doing so allows new requirables to be defined without modifying this file. Is that useful?

Also, should we move all the requirables into another module? Like `Wasp.Cli.Command.Requirables` or something? Or is leaving them in this module OK?

waspc/cli/src/Wasp/Cli/Command.hs
```diff
@@ -27,3 +43,88 @@ runCommand cmd = do
...
+          )
+
+-- | Require a Wasp project to exist near the current directory. Get the
+-- project directory by pattern matching on the result of 'require':
+--
+-- @
+-- do
+--   WaspRootRequirement waspRoot <- require
+-- @
+newtype WaspRootRequirement = WaspRootRequirement (SP.Path' SP.Abs (SP.Dir WaspProjectDir)) deriving (Typeable)
```
craigmc08, line 103: I'm not sure what to name this: when I went through and modified all the places that need to use this, it was split around 50/50 between calling the path `waspRoot` and `waspProjectDir`. We should choose one and stick with it everywhere.

waspc/cli/src/Wasp/Cli/Command.hs
```diff
@@ -1,24 +1,40 @@
...
 
 module Wasp.Cli.Command
   ( Command,
     runCommand,
     CommandError (..),
+
+    -- * Requirements
+    require,
+    DbConnectionRequirement (DbConnectionRequirement),
+    WaspRootRequirement (WaspRootRequirement),
```
craigmc08, line 12: I would prefer to not export the constructors for the requirements, because it would be nice to prevent people from constructing them without going through `require`. I tried using a unidirectional pattern synonym:

```hs
data DbConnectionRequirement = DbConnectionRequirementConstr
pattern DbConnectionRequirement <- DbConnectionRequirementConstr
```

But GHC complains about the pattern being able to fail when you match on it, and `{-# COMPLETE #-}` wasn't helping.

I don't think this is a big deal though, we can just trust people to not misuse these.

infomiho (commented):
This looks really cool on the usage side of things. I've tried to understand as much as I could.

- `Command` has also `StateT` with a list of requirements.
- `Typeable` and `cast` allow to match by type which requirement is being required.
- When checking for requirement, if it's not already resolved, run `checkRequirement`
- If some requirement is not met, we are throwing a `CommandError`

Fun code to read, I'll let @Martinsos approve this one when ready :)

waspc/cli/src/Wasp/Cli/Command.hs
Martinsos: Got it, makes sense. I remember I used https://stackoverflow.com/questions/8172548/is-it-possible-to-export-constructors-for-pattern-matching-but-not-for-construc for similar problems -> can't remember if I used View Patterns or Pattern Synonyms, maybe the former I think.
Interesting that it complains though hm! I think I got it working ok.
But ok, it is not terrible if we expose constructors, I agree.

waspc/cli/src/Wasp/Cli/Command.hs
Martinsos: Yeah, the issues is that we use both.
I would say WaspProjectDir then.

waspc/cli/src/Wasp/Cli/Command.hs
```diff
@@ -29,3 +43,88 @@ runCommand cmd = do
...
+require =
+  Command (gets (mapMaybe cast)) >>= \case
+    (req : _) -> return req
+    [] -> do
+      -- Requirement hasn't been met, so run the check
+      req <- checkRequirement
+      Command $ modify (Requirement req :)
+      return req
+
+data DbConnectionRequirement = DbConnectionRequirement deriving (Typeable)
```
Martinsos, line 76:
```suggestion
data DbConnectionRequirement = DbConnectionRequirement deriving (Typeable)
```
```suggestion
data DbConnectionEstablished = DbConnectionEstablished deriving (Typeable)
```

waspc/cli/src/Wasp/Cli/Command.hs
```diff
@@ -29,3 +43,88 @@ runCommand cmd = do
...
+          )
+
+-- | Require a Wasp project to exist near the current directory. Get the
+-- project directory by pattern matching on the result of 'require':
+--
+-- @
+-- do
+--   WaspRootRequirement waspRoot <- require
+-- @
+newtype WaspRootRequirement = WaspRootRequirement (SP.Path' SP.Abs (SP.Dir WaspProjectDir)) deriving (Typeable)
```
Martinsos, line 103:
```suggestion
newtype InWaspProject = InWaspProject (SP.Path' SP.Abs (SP.Dir WaspProjectDir)) deriving (Typeable)
```

waspc/cli/src/Wasp/Cli/Command.hs
Martinsos: I would probably extract it, just because the concept is somewhat standalone and it is now occupying very big part of this somewhat file.
I would move it to `Command/Require.hs` module. Or Requirement.

waspc/cli/src/Wasp/Cli/Command.hs
```diff
@@ -29,3 +43,88 @@ runCommand cmd = do
 -- TODO: What if we want to recognize errors in order to handle them?
 --   Should we add _commandErrorType? Should CommandError be parametrized by it, is that even possible?
 data CommandError = CommandError {_errorTitle :: !String, _errorMsg :: !String}
+
+data Requirement where
```
Martinsos, line 47: Aha, this is just here to be a heterogeneous container, ok.

waspc/cli/src/Wasp/Cli/Command.hs
Martinsos: While extracting it, I would consider writing one top comment in that module that explains the usage of `require` in a simple, approachable way. Kind of like if you are orally explaining it to somebody.

```
There are things that we need asserted in our code, and possibly also obtain 
connected value, e.g. ensuring we are in wasp project dir (+ learning what path to 
that dir is), checking if database connection is established, ... .
For stuff like that, we might want to ask the question multiple times, but we 
really need to do the actual check in the background only once. This is where 
`require` comes in play, as it has state which ensures that check is done only the 
first time, and after that it just delivers the results of last check. So now you 
can easily test for certain assumptions/requirements in any command.

For example, you can check if we are in wasp project root dir by doing:
EXAMPLE

All the current requirements are implemented in this module. To implement a new 
create your new type and make it an instance of Requirable, and that is it.
```

Something like this, with this level of simplicity / verbosity.

Martinsos (approved):
Left some comments, but am also approving, since I don't think I need to add anything after they are taken care of! Nice job with the `require`!

Martinsos:
@craigmc08 one general question: What if we used simpler approach for `require`, where we just have a `StateT Requirements`, where `data Requirements = Requirements { dbConnectionEstablished :: Bool, InWaspProjectDir :: Maybe <StrongPath> }`, and two functions ` checkDbConnectionEstablished :: Command Bool` and `checkInWaspProjectDir :: Command (Maybe <StrongPath>)`?
Would that work similarly well, but be much simpler regarding the type machinery?

waspc/cli/src/Wasp/Cli/Command.hs
craigmc08: Unfortunately there is no good way to completely remove the requirement code from `Command.hs`. Since the requirables depend on the Command monad, and the Command monad depends on Requirement, they can't be fully separated. I went for leaving the core requirement code (Requirement type, Requirable class, and require function) in `Command.hs` and moved the instances of requirable to a new module.

craigmc08:
> Would that work similarly well, but be much simpler regarding the type machinery?
I am guessing the answer could be that current approach is more extensible in theory, but is it in practice?

Yeah, I think that would work as well @Martinsos! Extensibility -> could be useful if we want a new requirement that's not as universally useful (i.e. only one command or one group of related commands cares about it), but I'm not sure that will ever happen.

The one benefit I'm thinking of is that having it use the Typeable stuff means that the actual requirement types are not related to the Command monad, other than running in them -> its more reasonable to have them defined in a separate file vs. with the simpler way it would feel a little strange defining that stuff outside of `Command.hs`.

Martinsos:
> > Would that work similarly well, but be much simpler regarding the type machinery?
> > I am guessing the answer could be that current approach is more extensible in theory, but is it in practice?
> 
> Yeah, I think that would work as well @Martinsos! Extensibility -> could be useful if we want a new requirement that's not as universally useful (i.e. only one command or one group of related commands cares about it), but I'm not sure that will ever happen.
> 
> The one benefit I'm thinking of is that having it use the Typeable stuff means that the actual requirement types are not related to the Command monad, other than running in them -> its more reasonable to have them defined in a separate file vs. with the simpler way it would feel a little strange defining that stuff outside of `Command.hs`.

Makes sense @craigmc08 -> I guess the tricky part is weighing what we care more about here, flexibility vs simplicity. I have to say that currently I am a bit more inclined toward simplicity here, although I would also be ok with sticking what we have, it is ok.

I will let you do the final call -> I am ok with whatever you choose. THen we can merge this!

Martinsos:
Very nice analysis @infomiho ! But what the heck :D?

So you found that those `prints` never happen. Maybe add an `hFlush stdout` after them, or use Debug.trace, in case they just didn't get the chance to be printed.

So doing that prisma check seems to be the issue. One thing I an imagine being tricky is that it assumes that schema is present in the generated project. Imagine completely fresh wasp project -> nobody compiled it yet -> so there is no .wasp/out dir, and no generated prisma schema. While that is something we might want to think about @craigmc08 -> thiknig if we need to ensure that explicitly or it will always be taken care of already, and how we can possibly make that less fragile -> I don't think this is the cause here, because `wasp db start` should have already taken care of that. Hm, or should it have? I am not sure, we should check that also.

One thing I am confused about here is that we don't have any errors printed! How is that? Are we surpressing errors at the wrong place, not forwarding something? Because I don't see much space here for infinite stuff happening, what could be looping?

To summarize:
1. I would do a bit more "console.log" debugging, but would consider putting `hFlush stdout` after any printing, and/or using `Debug.trace`. I would also put some more of those prints, for example before / after the db requirement commands.
2. We should consider the fact that our db connection check expects Wasp project to be freshly generated. Are we ensuring that properly? Even if not, is it actually a problem right now?
3. Are we surpressing some errors somewhere, why are we not seeing any?

craigmc08:
I figured out that, for some reason, when `stdin` is left open when running `db start` and a command that runs the DB status check, the DB status check hangs. I *think* this has to do with jobs inheriting `stdin` (https://github.com/wasp-lang/wasp/blob/main/waspc/src/Wasp/Generator/Job/Process.hs#L41), and can confirm that changing that to `CP.ClosedStream` fixes it, but I don't have the time currently to test if that change causes any problems. I'll open an issue for it.

In the meantime, I replaced the use of `concurrently` with a script that does essentially the same thing but closes the stdin streams of the db and app processes.

waspc/cli/src/Wasp/Cli/Command/Start.hs
```diff
@@ -26,12 +26,12 @@ start = do
   InWaspProject waspRoot <- require
   let outDir = waspRoot </> Common.dotWaspDirInWaspProjectDir </> Common.generatedCodeDirInDotWaspDir
 
-  DbConnectionEstablished <- require
-
   cliSendMessageC $ Msg.Start "Starting compilation and setup phase. Hold tight..."
 
   warnings <- compile
 
+  DbConnectionEstablished <- require
```
Martinsos, line 33: Ok, good, this should indeed happen after compilation.
Would be interesting to consider if this is something we could enforce at the type level. I am sure we could, but I guess the question is how pratical it would be.

waspc/headless-test/start.js
```diff
@@ -0,0 +1,32 @@
...
+    console.log(`\x1b[0m\x1b[33m[${name}][err]\x1b[0m ${data}`);
+  });
+  proc.on('exit', () => {
+    done();
+  });
+}
+
+let numDone = 0;
+const cb = () => {
+  if (numDone == 1) process.exit(0);
```
Martinsos: So this ensures that the execution stops after both things have been executed?

I have to admit code confused me a bit, I think it would be cleaner like this:
```
let numDone = 0;
const cb = () => {
  numDone++;
  if (numDone == 2) process.exit(0);
}
```

Even better might be having `spawn` function increase the global counter `numRunning` and then callback reduces that number and does exit when it is 0.

waspc/headless-test/start.js
```diff
@@ -0,0 +1,32 @@
...
+  // process hangs.
+  // See https://github.com/wasp-lang/wasp/pull/1218#issuecomment-1599098272.
+  proc.stdin.destroy();
+
+  readline(proc.stdout).on('line', data => {
+    console.log(`\x1b[0m\x1b[33m[${name}][out]\x1b[0m ${data}`);
+  });
+  readline(proc.stderr).on('line', data => {
+    console.log(`\x1b[0m\x1b[33m[${name}][err]\x1b[0m ${data}`);
+  });
```
Martinsos, line 20: So here we listen to stdout and wrap it in some stuff, and same for stderr. Why?

waspc/headless-test/start.js
```diff
@@ -0,0 +1,32 @@
+const cp = require('child_process');
```
Martinsos, line 1: This file is still pretty misterious to me, I think it needs some cleaning up / better comments @craigmc08 .

Martinsos:
@craigmc08 and @infomiho great job on figuring this out!

I have to admit I still don't quite understand the cause of the problem. I think this might be due to me somewhat lacking more knowledge on `stdin` and how these pipes/outputs behaves reagrding things like inheritance and open/closed and stuff.

Is there any way you could explain this simpler to me @craigmc08 ?

craigmc08:
> I have to admit I still don't quite understand the cause of the problem. I think this might be due to me somewhat lacking more knowledge on `stdin` and how these pipes/outputs behaves regarding things like inheritance and open/closed and stuff.
> 
> Is there any way you could explain this simpler to me?

I can try:

When we create a process through `runProcessAsJob`, it inherits stdin from the parent process. This means that, while the new process is running, all data sent to stdin of wasp-cli is also sent to the child process.

The prisma command used to check if the DB is running continuously reads stdin for SQL commands until there is no more input. Because it inherits stdin from wasp-cli, it keeps waiting on stdin for more input or for the pipe to be closed. This is why the `start.js` file calls `proc.stdin.destroy()` on the new processes.

(I'm still not sure why this doesn't happen when running wasp-cli from inside the terminal.)

PR #1223, opened 2023-05-25
Add test framework for language server autocompletion

craigmc08:
### Description

- Adds a framework for writing autocompletion tests for `waspls`
- Adds some tests for current autocompletion functionality
  - A bug where autocompletion doesn't work if a `.wasp` file starts with a comment was discovered and fixed
- Refactors some LSP code to make it testable in pure code

The tests are golden file tests, located in `waspls/test/Wasp/LSP/completionTests`. You write a file an indicate where the cursor is and the test system will generate a `.golden` file or compare it to the existing one to see if anything has changed. An example of a test case:

```wasp
//! test/completion
app todoApp {
  wasp: {
    version: "^0.10.5", 
  },
  title: "todo!",
}

route MainRoute { path: "/", to: | }
                                 ^
page MainPage {
  component: import { MainPage } from "@client/MainPage.jsx",
}
```

which tests the autocompletion after `to:` in the definition of `MainRoute`.


waspc/waspls/test/Wasp/LSP/CompletionTest.hs
```diff
@@ -0,0 +1,145 @@
...
+            "}"
+          ]
+   in "Completion Items:\n" ++ unlines fmtedCompletionItems
+
+-- | Parses a completion test case into a pair of the wasp source code to
+-- run completion on and the position to get the completion list at.
+--
+-- = Format
+--
+-- A normal wasp file, but with two addition
```
Martinsos, line 64: Ok, so after reading this, I understand this is just a wasp source file with a position marked in it. It could easily be just a normal wasp source, and then additionaly a pair of ints that indicate the position in it, but you went for this different way of marking the position in the source itself to make it nicer to read and figure out where the position is, otherwise we would always have to manually figure it out -> that is pretty cool!

waspc/waspls/test/Wasp/LSP/CompletionTest.hs
```diff
@@ -0,0 +1,145 @@
...
+-- @
+-- //! test/completion
+-- app todoApp {
+--   |
+--   ^
+-- }
+-- @
+--
+-- This test case checks completions after the 2 spaces on the 2nd line.
+readCompletionTest :: String -> (String, LSP.Position)
```
Martinsos, line 89: How about `parseCompletionTestInput`? Or `parseWaspSourceWithCursor`, if you go with that name.
`parse`, because it actually returns new things -> new source without the cursor, and cursor position. I would say that is proper parsing, not just "reading" something.

waspc/waspls/test/Wasp/LSP/CompletionTest.hs
```diff
@@ -0,0 +1,145 @@
...
+    before :: Int -> [a] -> [a]
+    before = take
+
+    after :: Int -> [a] -> [a]
+    after idx = drop (idx + 1)
+
+-- | Create a test case from a .wasp/.golden pair, running runCompletionTest
+-- on the .wasp file to get the .golden output. See 'readCompletionTest' for
+-- the format of a test case.
+makeCompletionTest :: FilePath -> TestTree
```
Martinsos, line 133: Ok, so this is given an test input file and based on it creates a golden test for completion. It will create golden file with same name but with .golden extension next to the test input file. Makes sense to me!

I would change name to `makeGoldenCompletionTest` or `makeGoldenCompletionTestFromTestInputFile`. Maybe the last one is the best even.

I would also consider using StrongPath -> it has to be File, it has to be System, and now is it absolute or relative -> I guess we don't care? But, even if we don't care, would be nice if user of this method knows we don't care, instead of wondering.
So, for example we can do `Path' a File'` then instead of `FilePath`.

As for comments I would also update them a tiny bit:
1. "create a test case from a .wasp/.golden pair" -> That confused me a bit. .wasp/.golden looked like a path to me. So we could say "from a pair of <test_name>.wasp, <test_name>.golden files". Although, we could argue it really makes a test from just the input file. But it does also use golden file if it is there, true. Ok. Also, since this function uses runCompletionTest, I would put it first -> first I learn about how to make a completion test, then about how it actually runs. For the sake of the "story".

waspc/waspls/test/Wasp/LSP/CompletionTest.hs
```diff
@@ -0,0 +1,145 @@
...
+          ]
+   in "Completion Items:\n" ++ unlines fmtedCompletionItems
+
+-- | Parses a completion test case into a pair of the wasp source code to
+-- run completion on and the position to get the completion list at.
+--
+-- = Format
+--
+-- A normal wasp file, but with two addition
+-- 1) Begins with "//! test/completion" (this is future-proofing in case we
```
Martinsos, line 65: This maybe feels a tiny bit redundant at the moment, we can easily add it later if it would be helpful, and right now it doesn't help much? Although, maybe it is nice for the person checking out file by accident (for example in PR) since it indicates "hey, this is a special kind of file"?
We could have also done this by modifying the extension possibly: these files could have `.wasp.withcursor` extension for example, instead. Or `.wasp.completiontest`.
But this is also good, makes sense!

waspc/waspls/test/Wasp/LSP/CompletionTest.hs
```diff
@@ -0,0 +1,145 @@
...
+import qualified Wasp.Analyzer.Parser.Lexer as Lexer
+import Wasp.LSP.Completion (getCompletionsAtPosition)
+import Wasp.LSP.ServerState (ServerState (ServerState, _cst, _currentWaspSource, _latestDiagnostics))
+
+-- | Run test cases in ./completionTests directory
+--
+-- See 'readCompletionTest' for the format of a test case.
+test_CompletionLists :: IO TestTree
+test_CompletionLists = do
+  inputFiles <- findByExtension [".wasp"] "./waspls/test/Wasp/LSP/completionTests"
```
Martinsos, line 26: What is a bit tricky here, is that when looking at `./waspls/test/...` path, I wonder where it starts, what is it relative to. We could comment that here, to make it easier for the reader? Or is there some other way to indicate that? We could play with StrongPath, but in this case it would be a bit too much work I think, easier to just comment it.

Martinsos, line 26:
```suggestion
  completionTestInputFiles <- findByExtension [".wasp"] "./waspls/test/Wasp/LSP/completionTests"
```

waspc/waspls/test/Wasp/LSP/CompletionTest.hs
```diff
@@ -0,0 +1,145 @@
...
+  return $
+    testGroup "Wasp.LSP.Completion.getCompletionsAtPosition" $
+      makeCompletionTest <$> inputFiles
+
+-- | Takes a completion test case and produces the list of completion items
```
Martinsos, line 31: When somebody says "test case", I imagine both input of test and expected result. So I wouldn't use "test case" here, but instead "completion test input". Or "wasp source with cursor". I suggested below to introduce either `type` or `newtype` for it so that should help.

Martinsos: What is also a bit confusing here is that you say it produces a list of completion items, but I see it produces a `String`. Ok, I am guessing that list of completion items is encoded as a `String` then, but still, I wonder a bit if this is a mistake in documentation. So maybe we can explain why is it just a String.

waspc/waspls/test/Wasp/LSP/CompletionTest.hs
```diff
@@ -0,0 +1,145 @@
...
+runCompletionTest :: String -> String
+runCompletionTest source =
+  let (code, position) = readCompletionTest source
```
Martinsos, line 34:
```suggestion
  let (waspSource, cursorPosition) = readCompletionTest source
```

waspc/waspls/test/Wasp/LSP/CompletionTest.hs
```diff
@@ -0,0 +1,145 @@
...
+      fmtCompletionItem item =
+        concat
+          [ "  label={",
+            show (item ^. LSP.label),
+            "} kind={",
+            show (item ^. LSP.kind),
+            "} detail={",
+            show (item ^. LSP.detail),
+            "}"
+          ]
```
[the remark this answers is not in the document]
Martinsos: I would also maybe move that prefix of two spaces into line below where you combine them all together, with a `map ("  " <>)`, since that is the place where we care about how they all look together and therefore indentation matters due to the title "Completion Items:\n".

waspc/waspls/test/Wasp/LSP/CompletionTest.hs
```diff
@@ -0,0 +1,145 @@
...
+   in "Completion Items:\n" ++ unlines fmtedCompletionItems
```
Martinsos, line 57:
```suggestion
   in "Completion items:\n" ++ unlines fmtedCompletionItems
```

waspc/waspls/test/Wasp/LSP/completionTests/route_to.golden
```diff
@@ -0,0 +1,4 @@
+Completion Items:
+  label={"todoApp"} kind={Just CiVariable} detail={Just ":: app (declaration type)"}
```
Martinsos, line 2: This is how it actually works right now? This is a bit wrong, right -> because it offers any identifier in Wasp, even those that don't fit here? It is ok, just trying to remind myself.

Martinsos (changes requested):
Super cool @craigmc08 , I love the idea with marking cursor in the wasp source file, that makes tests easy to read!

One thing: for other golden tests that we have (and have a bunch of those), we somehow told Githbu that they are generated, so it knows to show them as closed in the PR. You can still see them, but they are closed by default which does help when reviewing, so they don't take away too much attention. Should we do the same thing for these golden files? I can't remember how we did it, it is probably in some gitconfig file at top level of repo, so give it a look, I can help if you can't find it.

Martinsos:
Btw @craigmc08 we have new thing for how we do reviews: normally only creator of comment resolves comment, so they have opportunity to check the work that was done. On the other hand, PR author should put :rocket: on comments that they have solved and the code is commited. That is the signal for reviewer of "Done and commited!".
Also, all comments should be commented back on OR given a :+1: , this way reviewer know that PR author has seen the comment and read it, and can easily detect comments that were by accident skipped.

So comment or :+1: for read/accepted comments, :rocket: for solved, only comment author resolves their comment.

waspc/waspls/test/Wasp/LSP/CompletionTest.hs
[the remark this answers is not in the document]
craigmc08: I've never actually seen `printf` in Haskell before, cool!

waspc/waspls/test/Wasp/LSP/CompletionTest.hs
```diff
@@ -0,0 +1,145 @@
...
+
```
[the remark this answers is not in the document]
craigmc08: Added `newtype CompletionTestInput` to the top of the file, and moved the comment explaining the format to that newtype.

waspc/waspls/test/Wasp/LSP/completionTests/route_to.golden
craigmc08: Yep! In this version it just lists all of the identifiers in the file that have a declaration type, which is something to improve in the future.

waspc/waspls/test/Wasp/LSP/CompletionTest.hs
craigmc08: > I would change name to makeGoldenCompletionTest or makeGoldenCompletionTestFromTestInputFile. Maybe the last one is the best even.

Changed name to `makeGoldenCompletionTest`.

> As for comments I would also update them a tiny bit:

👍 

> I would also consider using StrongPath

I just tried that out and it doesn't feel worth it to me. It adds a lot of noise where the paths are used/created from and only improves the readability of `makeGoldenCompletionTest`. I've added to the doc comment for `makeGoldenCompletionTest` what the `FilePath` is expected to be, let me know if you think that is good enough.

craigmc08 (commented):
@Martinsos Most of the your comments are addressed, just a couple that I'm looking for some more feedback on before I make changes for them.

> One thing: for other golden tests that we have (and have a bunch of those), we somehow told [Github] that they are generated, so it knows to show them as closed in the PR.

Looks like that's setting `linguist-generated=true` in https://github.com/wasp-lang/wasp/blob/main/.gitattributes. I've done so for these golden files. :+1:

waspc/waspls/src/Control/Monad/Log/Pure.hs
```diff
@@ -0,0 +1,27 @@
...
+import Control.Monad.Log.Class (MonadLog (logM))
+import Control.Monad.Writer.Strict (MonadWriter (tell), WriterT, runWriterT)
+import Data.Functor.Identity (Identity (runIdentity))
+
+type Log = LogT Identity
+
+runLog :: Log a -> (a, [String])
+runLog m = runIdentity $ runLogT m
+
+newtype LogT m a = LogT (WriterT [String] m a)
```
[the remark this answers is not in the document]
Martinsos: Ok sure, let's rename to Control.Monad.Log.

Log and LogT -> let's then follow what Write and others do.

Great for instance!

waspc/waspls/test/Wasp/LSP/CompletionTest.hs
Martinsos: StrongPath -> ok sure, if you feel it clutters the whole thing too much, then comments can also do the trick! It is localized function anyway, relatively isolated.

waspc/waspls/test/Wasp/LSP/CompletionTest.hs
```diff
@@ -0,0 +1,157 @@
...
+-- app todoApp {
+--   |
+--   ^
+-- }
+-- @
+--
+-- This test checks completions after the 2 spaces on the 2nd line.
+newtype CompletionTestInput = CompletionTestInput String
+
+-- | Run test cases in ./completionTests directory
```
Martinsos, line 54:
```suggestion
-- | Run test cases in ./completionTests directory.
```

waspc/waspls/test/Wasp/LSP/CompletionTest.hs
```diff
@@ -0,0 +1,157 @@
...
+import qualified Wasp.Analyzer.Parser.Lexer as Lexer
+import Wasp.LSP.Completion (getCompletionsAtPosition)
+import Wasp.LSP.ServerState (ServerState (ServerState, _cst, _currentWaspSource, _latestDiagnostics))
+
+-- | A string containing the input to a completion test. It represents wasp
+-- source code with a cursor position.
+--
+-- = Format
+--
+-- A normal wasp file, but with two addition
```
Martinsos, line 27:
```suggestion
-- A normal wasp file, but with two additions:
```

waspc/waspls/test/Wasp/LSP/CompletionTest.hs
```diff
@@ -0,0 +1,157 @@
...
+      makeGoldenCompletionTest <$> completionTestInputFiles
+
+-- | Create a test case from a pair of <test_name>.wasp and <test_name>.golden
+-- file, using 'runCompletionTest' on the .wasp file to get the output to
+-- compare with or store in the .golden file.
+--
+-- The given filepath should be to a completion test input file (ending in .wasp)
+-- inside the completionTests folder.
+makeGoldenCompletionTest :: FilePath -> TestTree
+makeGoldenCompletionTest inputFp =
```
Martinsos, line 74:
```suggestion
makeGoldenCompletionTest testInputFile =
```

Two reasons for this change:
1. You use `Fp` suffix here, but not for `goldenFile`. So best to sync those. And Fp is  a bit harder to guess as to what it is.
2. testInput is more descriptive and we already use that term around.

waspc/waspls/test/Wasp/LSP/CompletionTest.hs
```diff
@@ -0,0 +1,157 @@
...
+        Just x -> x
+
+    -- (String, column, line) triples, where String is a line such that the
+    -- following line is spaces followed by a ^ and nothing else
+    candidateLines = mapMaybe (toCandidateLine 0) linePairs
+    -- Pairs of consecutive lines in the testInput
+    linePairs = zip3 (lines testInput) (drop 1 $ lines testInput) [0 ..]
+
+    -- Check if the candidate line contains a '|' at the specified column
+    isMarkedLine :: (String, Int, Int) -> Bool
```
Martinsos, line 138: So isMarkedLine is a line with cursor? Why not call it `cursorLine`?

waspc/waspls/test/Wasp/LSP/CompletionTest.hs
```diff
@@ -0,0 +1,157 @@
...
+-- See 'CompletionTestInput' for the format of the test input.
+parseCompletionInput :: CompletionTestInput -> (String, LSP.Position)
+parseCompletionInput (CompletionTestInput testInput) =
+  withPreambleAssert "//! test/completion" testInput (unlines code, position)
+  where
+    -- Drops the marked line AND the following line (which is blank except for the ^)
+    code = before markedLineIdx (lines testInput) ++ [markedLine] ++ after (markedLineIdx + 1) (lines testInput)
+    position = LSP.Position (fromIntegral markedLineIdx) (fromIntegral markedColIdx)
+
+    markedLine = before markedColIdx rawMarkedLine ++ after markedColIdx rawMarkedLine
```
Martinsos, line 125: "marked" line semantic confused me somewhat. I would maybe go for "lineWithCursor" and then also have function `lineWithCursorCleanedUp` or something like that.

Martinsos (commented):
@craigmc08 left some comments, looking good I think we are almost done, the only thing to review a bit more is `parseComletionInput`, I didn't have to review its inner workings in details but I think yo ucould improve readibility, I left only some comments but not all. Pls give it another go and try to make it a bit more readable / cleaner, and other comments, and that should be it.

waspc/waspls/test/Wasp/LSP/CompletionTest.hs
```diff
@@ -0,0 +1,179 @@
...
+    withPrefixAssert :: String -> String -> a -> a
+    withPrefixAssert preamble str x
+      | (preamble ++ "\n") `isPrefixOf` str = x
+      | otherwise = error $ "test expected to begin with preamble: " ++ preamble
+
+    before :: Int -> [a] -> [a]
+    before = take
+
+    after :: Int -> [a] -> [a]
+    after idx = drop (idx + 1)
```
Martinsos, line 179: I found this to be a bit imperatively written / hard to read, so I ended up giving it a go of my own:

```hs
-- | Parses a completion test case into a pair of the wasp source code to
-- run completion on and the position to get the completion list at.
--
-- See 'CompletionTestInput' for the format of the test input.
parseCompletionInput :: CompletionTestInput -> (String, LSP.Position)
parseCompletionInput (CompletionTestInput waspSourceWithCursor) =
  if not (hasCompletionTestPreamble waspSourceWithCursor)
    then error missingCompletionTestPreambleErrorMsg
    else (waspSource, cursorPosition)
  where
    waspSource =
      unlines . concat $
        [ linesBeforeCursor,
          [cursorLineWithCursorRemoved],
          linesAfterCursor
        ]

    cursorPosition = LSP.Position (fromIntegral cursorLineIdx) (fromIntegral cursorColumnIdx)

    linesBeforeCursor = take cursorLineIdx (lines waspSourceWithCursor)

    -- NOTE: We have +2 here to ensure we skip the "      ^   " line after the cursor.
    linesAfterCursor = drop (cursorLineIdx + 2) (lines waspSourceWithCursor)

    cursorLineWithCursorRemoved =
      let cursorLine = lines waspSourceWithCursor !! cursorLineIdx
       in take cursorColumnIdx cursorLine <> drop (cursorColumnIdx + 1) cursorLine

    (cursorLineIdx, cursorColumnIdx) =
      fromMaybe (error "parseCompletionTest: found no cursor!") . msum $
        (\(lineIdx, linePair) -> (lineIdx,) <$> cursorColIdxInConsecutiveLinesPair linePair)
          <$> zip [0 ..] (zip (lines waspSourceWithCursor) (drop 1 $ lines waspSourceWithCursor))

    -- Given two consecutive lines from the test input, find cursor if there is one
    -- and return the index of the column it is in.
    cursorColIdxInConsecutiveLinesPair :: (String, String) -> Maybe Int
    cursorColIdxInConsecutiveLinesPair (line1, line2) = do
      cursorColIdx <- elemIndex '|' line1
      if (length line2 > cursorColIdx)
        && all (\(idx, c) -> if idx == cursorColIdx then c == '^' else isSpace c) (zip [0 ..] line2)
        then return cursorColIdx
        else Nothing

    hasCompletionTestPreamble :: String -> Bool
    hasCompletionTestPreamble source = (completionTestPreamble <> "\n") `isPrefixOf` source

    missingCompletionTestPreambleErrorMsg =
      "test expected to begin with preamble: " <> completionTestPreamble

    completionTestPreamble = "//! test/completion"
```

I removed `withPrefixAssert` because I found it too abstract, hard to understand when I see it up there what it does, so I inlined it instead.
As for the rest of the code, I organized it more declaratively, so more top down versus steps from the bottom to the top.

Take a look if you like it and use it if so (or parts of it)!

Btw I am starting to think this could have been made a bit easier if we used something like `{CURSOR}` instead of `|` and then `^` in the next line :D. It does look visually appealing, but it does also force that blank line and is harder to parse. Ok, it also has much less chance (0?) of occuring naturaly in Wasp code, but still, we control this code so in practice there will not happen that we have {CURSOR} somewhere in it.
But ok, we already have this and it is kind of nice so if you want to keep the | + ^ I think that is ok, it is cool at the end.

craigmc08: > As for the rest of the code, I organized it more declaratively, so more top down versus steps from the bottom to the top.

Ah, that's funny! I just reorganized it to be more imperative because it felt like it might be more readable (if not less Haskellish). But I'll switch it back to be declarative again.

I've pulled a bunch of the ideas from your suggested code too.

Martinsos: Ah sorry about that! Sometimes it can be more readable if imperative, but I do like when declarative code goes from top to bottom, I think that is one of its main attractive attributes, especially when order of actions is not important.

waspc/waspls/test/Wasp/LSP/CompletionTest.hs
```diff
@@ -0,0 +1,170 @@
...
+    cursorColumnIdxFromLinePair :: (Int, (String, String)) -> Maybe (Int, Int)
+    cursorColumnIdxFromLinePair (lineIdx, (line1, line2)) = do
+      cursorColIdx <- elemIndex '|' line1
+      -- Check that the line ends after the column with the | from the previous line
+      guard $ length line2 == cursorColIdx + 1
+      -- Check that ^ is at the same column as | from the previous line
+      guard $ (line2 !! cursorColIdx) == '^'
+      -- Check that only spaces precede the ^
+      guard $ all isSpace $ take cursorColIdx line2
+      return (lineIdx, cursorColIdx)
```
Martinsos, line 163: Well while we are at it we can go all the way then :D:

```suggestion
    -- Given two consecutive lines from the test input, along with the line index
    -- of the first line, look for a cursor in the line pair and return the index
    -- of the line and column the cursor is at.
    cursorLineIdxAndColumnIdxInLinePair :: (Int, (String, String)) -> Maybe (Int, Int)
    cursorLineIdxAndColumnIdxInLinePair (line1Idx, (line1, line2)) = do
      cursorColIdx <- elemIndex '|' line1
      guard $ line2 == replicate cursorColIdx ' ' <> '^'
      return (lineIdx, cursorColIdx)
```

Ok this is a bit more sensitive because it expects ' ' and not any space but in practice I don't think it matters.

Notice I updated the name of the function and also comments to reflect it also returns line, not just column.

Btw nice idea with `guard`, I forgot about it / never knew it well enough, but is a great fit for this.

Martinsos (approved):
@craigmc08 I approved!
I left one last comment, please take a look at that one, do with it what you like, and then feel free to merge.
Very nice all together, and very nice code at the end I think.

Martinsos (approved):
@craigmc08 I approved!
I left one last comment, please take a look at that one, do with it what you like, and then feel free to merge.
Very nice all together, and very nice code at the end I think.

PR #1226, opened 2023-05-30
move `Wasp/Generator/Node/Version` to `Wasp/Node/Version`

craigmc08:
Moves the `Wasp/Generator/Node/Version` to `Wasp/Node/Version` since now modules outside of Generator depend on Node's version.

As discussed in https://github.com/wasp-lang/wasp/pull/1210/files#r1206824035.


infomiho (approved):
LGTM

Martinsos (approved):
Looking good! Feel free to merge @craigmc08 .
We don't do merge commits unless necessary, instaed we usually squash all commits into one and rebase that onto `main`. Sometimes we don't squash them all but just some, if we want to preserve some of the commits.
Easiest thing to do therefore is to just click on `squash and merge`.

PR #1230, opened 2023-06-02
[waspls] Add dictionary autocomplete + signatures

### LINKED ISSUE #1173: Dict field autocomplete for Wasp LSP

craigmc08:
### Description

- Adds autocomplete for keys in dictionaries
  - `Wasp.LSP.Completions.DictKeyCompletion`
- Adds type signatures in dictionaries to display what type is expected at the current field
  - `Wasp.LSP.SignatureHelp`
- `Wasp.LSP.TypeInference` for finding types in the concrete syntax tree

To test out the new features, `cabal build` waspc in this branch and set the VSCode Wasp extension's executable path to the build output of Cabal. Cabal shows the location of the executable on the last line of its output, where it says `Linking <filepath> ...`.

Closes #1173 

3. [X] **New feature** (non-breaking change which adds functionality).

### DIFF
--- a/waspc/waspls/src/Wasp/LSP/Completions/Common.hs
+++ b/waspc/waspls/src/Wasp/LSP/Completions/Common.hs
@@ -1,0 +1,51 @@
+{-# LANGUAGE TemplateHaskell #-}
+
+module Wasp.LSP.Completions.Common
+  ( CompletionProvider,
+    CompletionContext (..),
+    src,
+    cst,
+    makeBasicCompletionItem,
+  )
+where
+
+import Control.Lens (makeClassy)
+import qualified Data.Text as Text
+import qualified Language.LSP.Types as LSP
+import Wasp.Analyzer.Parser.CST (SyntaxNode)
+import Wasp.Analyzer.Parser.CST.Traverse (Traversal)
+
+-- | A function that providers 'LSP.CompletionItems' at a location
+type CompletionProvider m = Traversal -> m [LSP.CompletionItem]
+
+-- | The context given to 'CompletionProvider's, via @MonadReader CompletionContext@
+data CompletionContext = CompletionContext
+  { _src :: String,
+    _cst :: [SyntaxNode]
+  }
+  deriving (Eq, Show)
+
+makeClassy 'CompletionContext
+
+-- | Create a completion item containing only a label.
+makeBasicCompletionItem :: Text.Text -> LSP.CompletionItem
+makeBasicCompletionItem name =
+  LSP.CompletionItem
+    { _label = name,
+      _kind = Nothing,
+      _tags = Nothing,
+      _detail = Nothing,
+      _documentation = Nothing,
+      _deprecated = Nothing,
+      _preselect = Nothing,
+      _sortText = Nothing,
+      _filterText = Nothing,
+      _insertText = Nothing,
+      _insertTextFormat = Nothing,
+      _insertTextMode = Nothing,
+      _textEdit = Nothing,
+      _additionalTextEdits = Nothing,
+      _commitCharacters = Nothing,
+      _command = Nothing,
+      _xdata = Nothing
+    }
--- a/waspc/waspls/src/Wasp/LSP/Completions/Common.hs
+++ b/waspc/waspls/src/Wasp/LSP/Completions/Common.hs
@@ -15,10 +15,10 @@ import qualified Language.LSP.Types as LSP
 import Wasp.Analyzer.Parser.CST (SyntaxNode)
 import Wasp.Analyzer.Parser.CST.Traverse (Traversal)
 
--- | A function that providers 'LSP.CompletionItems' at a location
+-- | A function that providers 'LSP.CompletionItems' at a location.
 type CompletionProvider m = Traversal -> m [LSP.CompletionItem]
 
--- | The context given to 'CompletionProvider's, via @MonadReader CompletionContext@
+-- | The context given to 'CompletionProvider's, via @MonadReader CompletionContext@.
 data CompletionContext = CompletionContext
   { _src :: String,
     _cst :: [SyntaxNode]
@@ -27,7 +27,8 @@ data CompletionContext = CompletionContext
 
 makeClassy 'CompletionContext
 
--- | Create a completion item containing only a label.
+-- | Create a completion item containing only a label. Use lenses and 'Control.Lens.(?~)'
+-- to set more fields, if desired.
 makeBasicCompletionItem :: Text.Text -> LSP.CompletionItem
 makeBasicCompletionItem name =
   LSP.CompletionItem
--- a/waspc/waspls/src/Wasp/LSP/Completions/Common.hs
+++ b/waspc/waspls/src/Wasp/LSP/Completions/Common.hs
@@ -16,9 +16,8 @@ import Wasp.Analyzer.Parser.CST (SyntaxNode)
 import Wasp.Analyzer.Parser.CST.Traverse (Traversal)
 
 -- | A function that providers 'LSP.CompletionItems' at a location.
-type CompletionProvider m = Traversal -> m [LSP.CompletionItem]
+type CompletionProvider m = CompletionContext -> Traversal -> m [LSP.CompletionItem]
 
--- | The context given to 'CompletionProvider's, via @MonadReader CompletionContext@.
 data CompletionContext = CompletionContext
   { _src :: String,
     _cst :: [SyntaxNode]
--- a/waspc/waspls/src/Wasp/LSP/Completions/ExprCompletion.hs
+++ b/waspc/waspls/src/Wasp/LSP/Completions/ExprCompletion.hs
@@ -5,7 +5,6 @@ where
 
 import Control.Lens ((?~), (^.))
 import Control.Monad.Log.Class (MonadLog (logM))
-import Control.Monad.Reader.Class (MonadReader, asks)
 import Data.Maybe (maybeToList)
 import qualified Data.Text as Text
 import qualified Language.LSP.Types as LSP
@@ -13,7 +12,7 @@ import qualified Language.LSP.Types.Lens as LSP
 import Wasp.Analyzer.Parser.CST (SyntaxNode)
 import qualified Wasp.Analyzer.Parser.CST as S
 import Wasp.Analyzer.Parser.CST.Traverse
-import Wasp.LSP.Completions.Common (CompletionContext, CompletionProvider, makeBasicCompletionItem)
+import Wasp.LSP.Completions.Common (CompletionProvider, makeBasicCompletionItem)
 import qualified Wasp.LSP.Completions.Common as Ctx
 import Wasp.LSP.Syntax (findChild, isAtExprPlace, lexemeAt)
 
@@ -21,17 +20,15 @@ import Wasp.LSP.Syntax (findChild, isAtExprPlace, lexemeAt)
 -- and return them as completion items.
 --
 -- TODO: include completions for enum variants (use standard type defs from waspc)
-getCompletions :: (MonadReader CompletionContext m, MonadLog m) => CompletionProvider m
-getCompletions location =
+getCompletions :: (MonadLog m) => CompletionProvider m
+getCompletions context location =
   if not (isAtExprPlace location)
     then do
       logM "[ExprCompletion] not at expression"
       return []
     else do
       logM "[ExprCompletion] at expression"
-      src <- asks (^. Ctx.src)
-      syntax <- asks (^. Ctx.cst)
-      let declNames = findDeclNames src syntax
+      let declNames = getDeclNamesAndTypes (context ^. Ctx.src) (context ^. Ctx.cst)
       logM $ "[ExprCompletion] declnames=" ++ show declNames
       return $
         map
@@ -43,8 +40,8 @@ getCompletions location =
           declNames
 
 -- | Search through the CST and collect all @(declName, declType)@ pairs.
-findDeclNames :: String -> [SyntaxNode] -> [(String, String)]
-findDeclNames src syntax = traverseForDeclNames $ fromSyntaxForest syntax
+getDeclNamesAndTypes :: String -> [SyntaxNode] -> [(String, String)]
+getDeclNamesAndTypes src syntax = traverseForDeclNames $ fromSyntaxForest syntax
   where
     traverseForDeclNames :: Traversal -> [(String, String)]
     traverseForDeclNames t = case kindAt t of
--- a/waspc/waspls/src/Wasp/LSP/Completions/ExprCompletion.hs
+++ b/waspc/waspls/src/Wasp/LSP/Completions/ExprCompletion.hs
@@ -1,0 +1,60 @@
+module Wasp.LSP.Completions.ExprCompletion
+  ( getCompletions,
+  )
+where
+
+import Control.Lens ((?~), (^.))
+import Control.Monad.Log.Class (MonadLog (logM))
+import Control.Monad.Reader.Class (MonadReader, asks)
+import Data.Maybe (maybeToList)
+import qualified Data.Text as Text
+import qualified Language.LSP.Types as LSP
+import qualified Language.LSP.Types.Lens as LSP
+import Wasp.Analyzer.Parser.CST (SyntaxNode)
+import qualified Wasp.Analyzer.Parser.CST as S
+import Wasp.Analyzer.Parser.CST.Traverse
+import Wasp.LSP.Completions.Common (CompletionContext, CompletionProvider, makeBasicCompletionItem)
+import qualified Wasp.LSP.Completions.Common as Ctx
+import Wasp.LSP.Syntax (findChild, isAtExprPlace, lexemeAt)
+
+-- | If the location is at an expression, find declaration names in the file
+-- and return them as autocomplete suggestions
+--
+-- TODO: include completions for enum variants (use standard type defs from waspc)
+getCompletions :: (MonadReader CompletionContext m, MonadLog m) => CompletionProvider m
+getCompletions location =
+  if not (isAtExprPlace location)
+    then do
+      logM "[ExprCompletion] not at expression"
+      return []
+    else do
+      logM "[ExprCompletion] at expression"
+      src <- asks (^. Ctx.src)
+      syntax <- asks (^. Ctx.cst)
+      let declNames = findDeclNames src syntax
+      logM $ "[ExprCompletion] declnames=" ++ show declNames
+      return $
+        map
+          ( \(name, typ) ->
+              makeBasicCompletionItem (Text.pack name)
+                & (LSP.kind ?~ LSP.CiVariable)
+                & (LSP.detail ?~ Text.pack (":: " ++ typ ++ " (declaration type)"))
+          )
+          declNames
+
+-- | Search through the CST and collect all @(declName, declType)@ pairs.
+findDeclNames :: String -> [SyntaxNode] -> [(String, String)]
+findDeclNames src syntax = traverseForDeclNames $ fromSyntaxForest syntax
+  where
+    traverseForDeclNames :: Traversal -> [(String, String)]
+    traverseForDeclNames t = case kindAt t of
+      S.Program -> maybe [] traverseForDeclNames $ down t
+      S.Decl ->
+        let declNameAndType = maybeToList $ getDeclNameAndType t
+         in declNameAndType ++ maybe [] traverseForDeclNames (right t)
+      _ -> maybe [] traverseForDeclNames $ right t
+    getDeclNameAndType :: Traversal -> Maybe (String, String)
+    getDeclNameAndType t = do
+      nameT <- findChild S.DeclName t
+      typeT <- findChild S.DeclType t
+      return (lexemeAt src nameT, lexemeAt src typeT)
--- a/waspc/waspls/src/Wasp/LSP/Completions/ExprCompletion.hs
+++ b/waspc/waspls/src/Wasp/LSP/Completions/ExprCompletion.hs
@@ -53,6 +53,9 @@ findDeclNames src syntax = traverseForDeclNames $ fromSyntaxForest syntax
         let declNameAndType = maybeToList $ getDeclNameAndType t
          in declNameAndType ++ maybe [] traverseForDeclNames (right t)
       _ -> maybe [] traverseForDeclNames $ right t
+
+    -- @getDeclNameAndType t@ expects 't' to be at a 'S.Decl' node. It finds the
+    -- lexemes for the 'S.DeclName' and 'S.DeclType'.
     getDeclNameAndType :: Traversal -> Maybe (String, String)
     getDeclNameAndType t = do
       nameT <- findChild S.DeclName t
--- a/waspc/waspls/src/Wasp/LSP/Completions/ExprCompletion.hs
+++ b/waspc/waspls/src/Wasp/LSP/Completions/ExprCompletion.hs
@@ -18,7 +18,7 @@ import qualified Wasp.LSP.Completions.Common as Ctx
 import Wasp.LSP.Syntax (findChild, isAtExprPlace, lexemeAt)
 
 -- | If the location is at an expression, find declaration names in the file
--- and return them as autocomplete suggestions
+-- and return them as completion items.
 --
 -- TODO: include completions for enum variants (use standard type defs from waspc)
 getCompletions :: (MonadReader CompletionContext m, MonadLog m) => CompletionProvider m
--- a/waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
+++ b/waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
@@ -24,7 +24,7 @@ import Wasp.Analyzer.Type (Type)
 import qualified Wasp.Analyzer.Type as Type
 import Wasp.LSP.ServerState (ServerState, cst, currentWaspSource)
 import Wasp.LSP.Syntax (locationAtOffset, lspPositionToOffset)
-import Wasp.LSP.TypeInference (ExprKey (DictKey, List, Tuple), findExprPathToLocation, findTypeForPath)
+import Wasp.LSP.TypeInference (ExprPathStep (DictKey, List, Tuple), findExprPathToLocation, findTypeForPath)
 import Wasp.LSP.Util (hoistMaybe)
 
 -- | Configuration for 'LSP.Options', used in "Wasp.LSP.Server".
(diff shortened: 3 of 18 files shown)


infomiho:
Cool stuff!

Quick questions from testing it locally:
1. It offers existing keys, can we avoid that?
Screenshot 2023-06-02 at 17 32 48

2. Can we say "Javascript import" or similar instead of external import
Screenshot 2023-06-02 at 17 32 35

3. Can we format the dicts (not to be on the same line)?

I'll be testing it out more throughly on Monday :)

waspc/waspls/src/Wasp/LSP/Completions/ExprCompletion.hs
```diff
@@ -0,0 +1,63 @@
...
+import qualified Wasp.Analyzer.Parser.CST as S
+import Wasp.Analyzer.Parser.CST.Traverse
+import Wasp.LSP.Completions.Common (CompletionContext, CompletionProvider, makeBasicCompletionItem)
+import qualified Wasp.LSP.Completions.Common as Ctx
+import Wasp.LSP.Syntax (findChild, isAtExprPlace, lexemeAt)
+
+-- | If the location is at an expression, find declaration names in the file
+-- and return them as completion items.
+--
+-- TODO: include completions for enum variants (use standard type defs from waspc)
```
infomiho, line 23: Is this something you'd like to resolve in this PR?

waspc/waspls/src/Wasp/LSP/Completions/ExprCompletion.hs
```diff
@@ -0,0 +1,63 @@
...
+          )
+          declNames
+
+-- | Search through the CST and collect all @(declName, declType)@ pairs.
+findDeclNames :: String -> [SyntaxNode] -> [(String, String)]
+findDeclNames src syntax = traverseForDeclNames $ fromSyntaxForest syntax
+  where
+    traverseForDeclNames :: Traversal -> [(String, String)]
+    traverseForDeclNames t = case kindAt t of
+      S.Program -> maybe [] traverseForDeclNames $ down t
```
infomiho, line 51: Do we need the app name as completions items? I think we can easily add it later if needed, but now it feels weird getting a completion suggestion with the app name 

Screenshot 2023-06-06 at 11 05 22

waspc/waspls/src/Wasp/LSP/Completions/ExprCompletion.hs
```diff
@@ -0,0 +1,63 @@
...
+      S.Decl ->
+        let declNameAndType = maybeToList $ getDeclNameAndType t
+         in declNameAndType ++ maybe [] traverseForDeclNames (right t)
+      _ -> maybe [] traverseForDeclNames $ right t
+
+    -- @getDeclNameAndType t@ expects 't' to be at a 'S.Decl' node. It finds the
```
infomiho, line 57: I'm out of my depth here, but is there a way to limit the input to `S.Decl` on the type level?

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
```diff
@@ -0,0 +1,261 @@
...
+  m LSP.SignatureHelp
+getSignatureHelpAtPosition position = do
+  src <- gets (^. currentWaspSource)
+  gets (^. cst) >>= \case
+    Nothing ->
+      -- No CST in the server state, can't create a signature.
+      return emptyHelp
+    Just syntax -> do
+      let offset = lspPositionToOffset src position
+      let location = toOffset offset (fromSyntaxForest syntax)
```
infomiho, line 70: This reads a bit weird:
1. You get the `offset`
2. Then you use the `toOffset` function to get the location

Is a better name for that function `toLocation`?

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
```diff
@@ -0,0 +1,261 @@
...
+              { _signatures =
+                  LSP.List
+                    [ LSP.SignatureInformation
+                        { _parameters = Just $ LSP.List params,
+                          -- NOTE: VSCode higlights the 0th parameter if
+                          -- activeParameter is Nothing, so it's set to -1 when
+                          -- we don't want any param highlighted.
+                          _activeParameter = Just $ fromMaybe (-1) activeParam,
+                          _label = Text.pack $ showFragments signatureFragments,
+                          _documentation = Nothing
```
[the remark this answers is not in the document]
infomiho:
```haskell
LSP.SignatureHelpDocMarkup $
    LSP.MarkupContent {
        LSP._kind = LSP.MkMarkdown,
        LSP._value = Text.pack $ "```\n" ++ showFragments signatureFragments ++ "\n```"
    }
```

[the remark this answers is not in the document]
infomiho: Screenshot 2023-06-06 at 11 37 07

Excuse my poor attempt, but it could work

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
```diff
@@ -0,0 +1,261 @@
...
+fragmentParam (Text _) = Nothing
+fragmentParam (Param key _) = Just key
+
+instance IsString SignatureFragment where
+  fromString string = Text string
+
+-- | Convert the container type of a signature to a list of fragments.
+--
+-- To avoid unreadably long signatures, dictionary types inside of the container
+-- type are written @{ ... }@.
```
infomiho, line 195:
```suggestion
-- type are written as @{ ... }@.
```

waspc/waspls/src/Wasp/LSP/Syntax.hs
```diff
@@ -50,14 +72,17 @@ toOffset targetOffset start = go $ bottom start
...
-    parentIs k = Just k == parentKind t
-    hasLeft k = k `elem` map kindAt (leftSiblings t)
+isAtExprPlace =
+  anyP
+    [ allP [parentIs S.DictEntry, hasLeft S.DictKey],
+      allP [parentIs S.Decl, hasLeft S.DeclType, hasLeft S.DeclName],
+      parentIs S.Tuple
+    ]
+
+-- (parentIs S.DictEntry && hasLeft S.DictKey)
```
infomiho, line 82: Left over comment

infomiho (commented):
I left a couple of comments, it seems to work well 👍

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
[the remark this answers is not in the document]
craigmc08: This feels too much like a VSCode-specific hack to be worth it, especially since it's just duplicating information. Other editors with LSP clients seem to support multi-line signatures. Helix for example:

image

And if we look at how other LSP servers use SignatureHelp, they run into the same problem with VSCode's single line signature strings. Rust for example:

image
 
I think we should stick with what the specification says to use these fields for.

Anyways, I added newlines back to the signature string so that editors that support it will display it over multiple lines.

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
```diff
@@ -0,0 +1,261 @@
...
+    -- A left fold is used because offset needs to increase as you go right across
+    -- the list. This means the parameter list needs to be reversed after the fold
+    -- so that they are in the same order as the parameters in the fragments.
+    go :: (LSP.UInt, [LSP.ParameterLabel]) -> SignatureFragment -> (LSP.UInt, [LSP.ParameterLabel])
+    go (offset, labels) (Text text) = (offset + fromIntegral (length text), labels)
+    go (offset, labels) (Param _ text) =
+      let end = offset + fromIntegral (length text)
+       in (end, LSP.ParameterLabelOffset offset end : labels)
+
+    labelToInfo :: LSP.ParameterLabel -> LSP.ParameterInformation
```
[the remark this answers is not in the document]
craigmc08: That's just how VSCode renders it, nothing we can do about it.

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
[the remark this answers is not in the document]
Martinsos: Oho nice nice!

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
infomiho: Sounds better 👍

Martinsos:
Is there any way to get it to suggest all possible dictionary keys without me writing a first letter? I am in a declaration, let's say `app`, and I just want to know what I can write next, how would that work, is that possible?

waspc/waspls/src/Wasp/LSP/Completions/ExprCompletion.hs
Martinsos: @infomiho so we don't yet have smart completion in the sense that it suggests only stuff that matches by type: right now it suggests all identifiers.
While Craig is implementing it now to correctly work for dictionary keys, this will still not make it correctly work for values (by selecting only those that match by type).

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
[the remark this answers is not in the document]
Martinsos: Ah that sucks. But why does it look good for Typescript then, in @infomiho 's screenshot above?

waspc/waspls/src/Wasp/LSP/Completions/ExprCompletion.hs
infomiho: We can just omit the app declaration type :) that's my suggestion here

craigmc08:
> Is there any way to get it to suggest all possible dictionary keys without me writing a first letter? I am in a declaration, let's say `app`, and I just want to know what I can write next, how would that work, is that possible?

It's up to each editor to choose when to ask the LSP server for completions. In VSCode you can (by default) press <kbd>Ctrl+Space</kbd> to trigger completions.

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
```diff
@@ -0,0 +1,261 @@
...
+-- is at a parameter, and the parameter is described in relation to the container
+-- with an 'ExprKey'.
+data Signature
+  = -- | Inside a container, but not at any parameter inside of it.
+    InContainer !Type
+  | -- | Inside a container at the parameter specified by the 'ExprKey'. This
+    -- is named \"param\" to be consistent with the naming conventions used by
+    -- the LSP.
+    AtParam !Type !ExprKey
+  deriving (Eq, Show)
```
Martinsos, line 122: This might be a bit nicer:

```suggestion
data Signature
  = -- | Type describes container, while ExprKey desribes param inside of container.
    InContainerAtParam !Type !ExprKey
  | InContainer !Type
  deriving (Eq, Show)
```

Although, would it make more sense to go with:
```suggestion
data Signature = Signature {
  _container :: !Type,
  _param :: !(Maybe ExprKey)
}
```

Then I don't think you even need any comments, almost.

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
```diff
@@ -0,0 +1,261 @@
...
+
+signatureParam :: Signature -> Maybe ExprKey
+signatureParam (InContainer _) = Nothing
+signatureParam (AtParam _ key) = Just key
+
+-- | @'findSignatureAtLocation' sourceCode location@ runs type inference at the
+-- given location to try to find a 'Signature'.
+--
+-- To do this, two types are inferred: the /tip type/ and the /parent type/. The
+-- tip type is the type expected exactly at the cursor location and the parent
```
Martinsos, line 136: Why tipType, why not locationType or smth like that?

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
```diff
@@ -0,0 +1,261 @@
...
+  guard $ not $ null exprPath
+  case exprPath of
+    [path] -> InContainer <$> hoistMaybe (findTypeForPath [path])
+    path -> do
+      -- Using init/last here is OK since we know @path@ has at least 2 elements.
+      tipType <- hoistMaybe $ findTypeForPath path
+      parentType <- hoistMaybe $ findTypeForPath $ init path
+      if isContainerType tipType
+        then return $ InContainer tipType
+        else return $ AtParam parentType (last path)
```
Martinsos, line 160: You can extract return up front.

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
```diff
@@ -0,0 +1,261 @@
...
+    isContainerType _ = False
+
+-- | A fragment of the text representation of a signature, with information
+-- for finding the spans of parameters inside of the text representation.
+--
+-- This is used as an intermediate form between 'Signature' and the response
+-- format that the LSP specifies.
+data SignatureFragment
+  = -- | A plaintext fragment of a signature.
+    Text !String
```
Martinsos, line 175:
```suggestion
    Plaintext !String
```

to help avoid confusion with standard `Text`.

Martinsos: Maybe even `PlaintextFragment`

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
```diff
@@ -0,0 +1,261 @@
...
+  | -- | A fragment that contains the text for a parameter of the signature. The
+    -- parameter is identified by the 'ExprKey'.
+    Param !ExprKey !String
```
Martinsos, line 178: `ParamFragment`?

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
```diff
@@ -0,0 +1,261 @@
...
+-- fragment.
+showFragments :: [SignatureFragment] -> String
+showFragments = concatMap fragmentText
+
+-- | Search through a list of fragments to find the parameters.
+--
+-- The LSP wants the starting and ending offset of each parameter in the signature
+-- help text (computed in 'showFragments'), so this function also has to keep
+-- track of the 0-based offset into the final text.
+getParamsFromFragments :: [SignatureFragment] -> [LSP.ParameterInformation]
```
Martinsos, line 230:
```suggestion
getLspParamsInfoFromFragments :: [SignatureFragment] -> [LSP.ParameterInformation]
```

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
Martinsos, line 1: This is quite a bit file! You probably want to extract Signature and SignatureFragment out of it, each into its own module?

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
```diff
@@ -0,0 +1,261 @@
...
+-- NOTE: This function computes the index into the parameter list, not the
+-- fragment list. This is why plain 'Text' fragments are filtered out of the
+-- list before indexing (via @mapMaybe fragmentParam@).
+findActiveParameterIndex :: [SignatureFragment] -> ExprKey -> Maybe LSP.UInt
+findActiveParameterIndex fragments key =
+  fromIntegral <$> elemIndex key (mapMaybe fragmentParam fragments)
+
+-- | Lift a 'Maybe' into a 'MaybeT' monad transformer.
+hoistMaybe :: Monad m => Maybe a -> MaybeT m a
+hoistMaybe = MaybeT . pure
```
Martinsos, line 261: Uf I am finding it a bit hard to understand this, could you help me a bit?

So `MaybeT m a` is a monad transformer stack, therefore a monad that has special powers of `Maybe` on the outside, but has powers of monad `m` on the inside. So it is a combo of `Maybe` and some `m` monad.

Now this function takes a `Maybe a` and returns `MaybeT m a`.
Ok, so the value `a` remains the same, sure, we just change the "wrapper" around it.

So we do `MaybeT . pure`.

Signature of pure is `pure :: a -> m a`, and it just wraps anything into a any monad.

Signature of `MaybeT` is `MaybeT :: m a -> MaybeT m a`. Ah no, that is wrong -> ok that is what confused me. So signature is `MaybeT :: m (Maybe a) -> MaybeT m a`. So basically it takes a "normal" combo of Maybe and some monad and turns it into a transformer, a powered up monad.

Ok, makes then, all the types fit of course.

So I really take `Maybe` value and pump it up to also have these aditional powers from this `m` monad.

But why `hoist` then? Why not `liftMaybe`? Does that already have some other meaning reserved? what about `maybeToMaybeT`? Or `liftToMaybeT`?

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
```diff
@@ -0,0 +1,261 @@
...
+findSignatureAtLocation src location = runMaybeT $ do
+  exprPath <- hoistMaybe $ findExprPathAtLocation src location
+  lift $ logM $ "[SignatureHelp] at expr path " ++ show exprPath
+  guard $ not $ null exprPath
+  case exprPath of
+    [path] -> InContainer <$> hoistMaybe (findTypeForPath [path])
+    path -> do
+      -- Using init/last here is OK since we know @path@ has at least 2 elements.
+      tipType <- hoistMaybe $ findTypeForPath path
+      parentType <- hoistMaybe $ findTypeForPath $ init path
```
Martinsos, line 157: Is there a reason why you do this check here before you do `if isContainerType`?
Because there is a branch where you won't need this result. So you could move this line into second branch. Unless you want to shortcircuit here if parentType can't be found.

waspc/waspls/src/Wasp/LSP/Syntax.hs
```diff
@@ -27,21 +31,39 @@ lspPositionToOffset srcString (J.Position l c) =
...
       -- return the last node instead.
       | otherwise = maybe at go $ at & next
 
+-- | Check if all of the supplied predicates are true
+allP :: Foldable f => f (a -> Bool) -> a -> Bool
+allP preds x = all ($ x) preds
+
+-- | Check if any of the supplied predicates are true
+anyP :: Foldable f => f (a -> Bool) -> a -> Bool
+anyP preds x = any ($ x) preds
```
Martinsos, line 59: Maybe move these functions to some Util module? They are in no way special for this module, right?

waspc/waspls/src/Wasp/LSP/TypeInference.hs
```diff
@@ -0,0 +1,132 @@
...
+import Wasp.Analyzer.Parser.CST.Traverse (Traversal)
+import qualified Wasp.Analyzer.Parser.CST.Traverse as T
+import Wasp.Analyzer.StdTypeDefinitions (stdTypes)
+import Wasp.Analyzer.Type (Type)
+import qualified Wasp.Analyzer.Type as Type
+import Wasp.Analyzer.TypeDefinitions (DeclType (dtBodyType), getDeclType)
+import Wasp.LSP.Syntax (lexemeAt)
+
+-- | Infer the type at a location in the CST, or 'Nothing' if type could not be
+-- determined.
```
Martinsos, line 26: I think this comment is redundant (which is great).

waspc/waspls/src/Wasp/LSP/TypeInference.hs
```diff
@@ -0,0 +1,132 @@
...
+inferTypeAtLocation :: String -> Traversal -> Maybe Type
+inferTypeAtLocation src location = case findExprPathAtLocation src location of
+  Just path -> findTypeForPath path
+  _ -> Nothing
```
Martinsos, line 30: Can't this be just:
```suggestion
inferTypeAtLocation src location = findTypeForPath <$> findExprPathAtLocation src location
```

waspc/waspls/src/Wasp/LSP/TypeInference.hs
```diff
@@ -0,0 +1,132 @@
...
+-- app todoApp {
+--   auth: {
+--     usernameAndPassword: |
+--                          ^
+--   }
+-- }
+-- @
+--
+-- The path to the cursor would be @[Decl "app", Key "auth", Key "usernameAndPassword"]@.
+type ExprPath = [ExprKey]
```
Martinsos, line 47: I like the term `Path`, but I am not so sure about `Expr` -> why `Expr`? Path to Expr?  Why not CodePath? Or CursorPath? or CstPath, whatever is the best fit?

waspc/waspls/src/Wasp/LSP/TypeInference.hs
```diff
@@ -0,0 +1,132 @@
...
+
+-- | A single step of an 'ExprPath'.
+data ExprKey
```
Martinsos, line 50: Why `ExprKey` name? Some ideas for a nicer name: `ExprPathStep`, `ExprPathLevel`.

waspc/waspls/src/Wasp/LSP/TypeInference.hs
```diff
@@ -0,0 +1,132 @@
...
+  = -- | @Decl declType@. Enter a declaration of type @declType@.
+    Decl !String
+  | -- | @Key key@. Enter a dictionary *and* it's key @key@.
+    Key !String
```
Martinsos, line 54: This is declaration key? Should we maybe call it DeclKey then, just to make it a bit more clear what it is?

waspc/waspls/src/Wasp/LSP/TypeInference.hs
```diff
@@ -0,0 +1,132 @@
...
+  | -- | Enter a value inside a list.
+    List
+  | -- | @Tuple idx@. Enter the @idx@-th value inside of a tuple.
+    Tuple !Int
+  deriving (Eq, Show)
+
+-- | Try to get an expression path for the given location, returning 'Nothing'
+-- if no path can be determined.
```
Martinsos, line 62: These two lines are redundant.

waspc/waspls/src/Wasp/LSP/TypeInference.hs
```diff
@@ -0,0 +1,132 @@
...
+--
+-- This function only depends on the syntax to the left of the location, and
+-- tries to be as lenient as possible in finding paths.
+findExprPathAtLocation :: String -> Traversal -> Maybe ExprPath
```
Martinsos, line 66:
```suggestion
findExprPathToLocation :: String -> Traversal -> Maybe ExprPath
```
Or
```suggestion
findExprPathOfLocation :: String -> Traversal -> Maybe ExprPath
```

waspc/waspls/src/Wasp/LSP/TypeInference.hs
```diff
@@ -0,0 +1,132 @@
...
+        S.List -> (List :) <$> go t' -- Inside a list.
+        S.Tuple -> do
+          -- Inside a tuple, number of expression nodes to the left is the tuple
+          -- index that @t@ is part of.
+          let nExprsBefore = length $ filter (S.syntaxKindIsExpr . T.kindAt) $ T.leftSiblings t
+          (Tuple nExprsBefore :) <$> go t'
+        _ -> go t' -- Found some other node, just ignore it and continue the tree.
+
+-- | Get the type in 'stdTypes' for the expression path. The path must start
+-- with a 'Decl', otherwise 'Nothing' is returned. If the path does not exist in
```
Martinsos, line 98:
```suggestion
-- with a 'Decl', otherwise 'Nothing' is returned. If the path's decl does not exist in
```

Martinsos (commented):
Looking great @craigmc08 ! Left some comments but nothing big, I believe this will get merged relatively quickly.

waspc/waspls/src/Wasp/LSP/TypeInference.hs
Martinsos: It's used in places as more than just part of an ExprPath -> but you say in comments it is `A single step of an 'ExprPath'`. If it is used as something more, maybe we should adjust comments.

Also, the need to say `A single step of an ExprPath` in comments is what kind of tells me it should probably be named that way, `ExprPathStep`.

If you are sure "key" is the right term, ok, but the only way I have seen it in our code so far was as dict key, and its usage in ExprPath / Signature is new and unknown to me, therefore devoid of meaning.

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
Martinsos: How would "tip" fit into the tree methapor though?

Martinsos (approved):
@craigmc08 did another quick round!

Keep in mind to use :+1: for "ok" / "got it" / "will do", and :rocket: for "done" -> it seems to me you are using just :+1: , which sends me wrong signal because I assume you haven't done that yet so am afraid to resolve it or you might forget about it if you counted on seeing that comment later.

waspc/waspls/src/Wasp/LSP/Completions/Common.hs
```diff
@@ -0,0 +1,52 @@
...
+import Control.Lens (makeClassy)
+import qualified Data.Text as Text
+import qualified Language.LSP.Types as LSP
+import Wasp.Analyzer.Parser.CST (SyntaxNode)
+import Wasp.Analyzer.Parser.CST.Traverse (Traversal)
+
+-- | A function that providers 'LSP.CompletionItems' at a location.
+type CompletionProvider m = Traversal -> m [LSP.CompletionItem]
+
+-- | The context given to 'CompletionProvider's, via @MonadReader CompletionContext@.
```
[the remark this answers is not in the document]
craigmc08: I think `MonadReader` is better if you have a lot of nested functions that need access to the same information, especially when some of the functions don't need it. That way you avoid having to pass the parameters down every level. Maybe a similar idea to using Contexts in React?

So I think making the decision comes down to: is it getting tedious to "thread" this argument through all these functions?

craigmc08:
> Keep in mind to use 👍 for "ok" / "got it" / "will do", and 🚀 for "done"

Guess I must've misinterpreted what to use those for! I'll make sure to add :rocket: when I'm done with things now.

waspc/waspls/src/Wasp/LSP/Completions/Common.hs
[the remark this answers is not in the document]
Martinsos: I think that makes sense. Probably it would make sense for `Generator` in Wasp. That said, I really like seeing which arguments function receives, so I feel like situation has to get really ugly before I would go for doing this hah! But ok, if there are more parameters then it also makes sense (although they could always be grouped hm).

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
Martinsos: Ok, sounds good! Btw isn't tip of the branch usually called "leaf"? Or did you have something different in mind?

Martinsos:
> > Keep in mind to use +1 for "ok" / "got it" / "will do", and rocket for "done"
> 
> Guess I must've misinterpreted what to use those for! I'll make sure to add rocket when I'm done with things now.

Sounds good! If you have a better idea for how to communicate this, let me know, but this is the best way we figured out so far. Btw no emoji is required if you write an actual comment explaining the situation, but often it is cumbersome to write a commend on every thing that is done.

waspc/waspls/src/Wasp/LSP/SignatureHelp.hs
craigmc08: Yeah that's true. I can't recall any reason I chose tip instead of leaf.

waspc/waspls/src/Wasp/LSP/Completions/Common.hs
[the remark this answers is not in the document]
craigmc08: I think `AppSpec` would be a decent candidate for putting inside a `MonadReader` in generator, could allow you to simplify stuff like:

```hs
genApp spec = genWebApp spec <++> genServer spec <++> genDb spec <++> genDockerFiles <++> genConfigFiles spec
```

into

```hs
genApp = genWebApp <++> genServer <++> genDb <++> genDockerFiles <++> genConfigFiles
```

but I agree it still doesn't really feel necessary here and also it feels a little magical. Potentially it would be good if you realize you actually need some more pieces of `AppSpec` in a function that you only passed parts of it to, without Reader you would have to go back and update all the function signatures with the new type that you want to pass down.

waspc/waspls/src/Wasp/LSP/Completions/ExprCompletion.hs
infomiho: We can skip this if there is not time now 👍

infomiho (approved):
Good to go

PR #1235, opened 2023-06-07
Type safe links experiment

infomiho:
Work in progress type safe `<Link>` component

### DIFF
--- a/waspc/src/Wasp/Generator/WebAppGenerator/RouterGenerator.hs
+++ b/waspc/src/Wasp/Generator/WebAppGenerator/RouterGenerator.hs
@@ -8,7 +8,8 @@ where
 import Data.Aeson (ToJSON (..), object, (.=))
 import qualified Data.Aeson as Aeson
 import Data.List (find)
-import Data.Maybe (fromMaybe)
+import Data.List.Extra (splitOn)
+import Data.Maybe (fromMaybe, mapMaybe)
 import StrongPath (Dir, Path, Rel, reldir, reldirP, relfile, (</>))
 import StrongPath.Types (Posix)
 import Wasp.AppSpec (AppSpec)
@@ -51,6 +52,7 @@ instance ToJSON RouterTemplateData where
 
 data RouteTemplateData = RouteTemplateData
   { _urlPath :: !String,
+    _urlParams :: ![String],
     _targetComponent :: !String
   }
 
@@ -58,6 +60,7 @@ instance ToJSON RouteTemplateData where
   toJSON routeTD =
     object
       [ "urlPath" .= _urlPath routeTD,
+        "urlParams" .= _urlParams routeTD,
         "targetComponent" .= _targetComponent routeTD
       ]
 
@@ -95,7 +98,7 @@ genRouter spec = do
       targetPath
       (Just $ toJSON templateData)
   where
-    routerPath = [relfile|router.jsx|]
+    routerPath = [relfile|router.tsx|]
     templateData = createRouterTemplateData spec
     targetPath = C.webAppSrcDirInWebAppRootDir </> asWebAppSrcFile routerPath
 
@@ -135,9 +138,12 @@ createExternalAuthProviderTemplateData maybeAuth (method, provider) =
 createRouteTemplateData :: AppSpec -> (String, AS.Route.Route) -> RouteTemplateData
 createRouteTemplateData spec namedRoute@(_, route) =
   RouteTemplateData
-    { _urlPath = AS.Route.path route,
+    { _urlPath = path,
+      _urlParams = extractUrlParams path,
       _targetComponent = determineRouteTargetComponent spec namedRoute
     }
+  where
+    path = AS.Route.path route
 
 -- NOTE: This should be prevented by Analyzer, so use error since it should not be possible
 determineRouteTargetComponent :: AppSpec -> (String, AS.Route.Route) -> String
@@ -151,7 +157,8 @@ determineRouteTargetComponent spec (_, route) =
     targetPage =
       fromMaybe
         ( error $
-            "Can't find page with name '" ++ targetPageName
+            "Can't find page with name '"
+              ++ targetPageName
               ++ "', pointed to by route '"
               ++ AS.Route.path route
               ++ "'"
@@ -182,3 +189,12 @@ createPageTemplateData page =
 
 relPathToWebAppSrcDir :: Path Posix (Rel importLocation) (Dir C.WebAppSrcDir)
 relPathToWebAppSrcDir = [reldirP|./|]
+
+extractUrlParams :: String -> [String]
+extractUrlParams urlPath =
+  mapMaybe
+    ( \case
+        ':' : paramName -> Just paramName
+        _ -> Nothing
+    )
+    (splitOn "/" urlPath)
--- a/waspc/data/Generator/templates/react-app/src/router.tsx
+++ b/waspc/data/Generator/templates/react-app/src/router.tsx
@@ -1,6 +1,6 @@
 {{={= =}=}}
-import React from 'react'
-import { Route, Switch, BrowserRouter as Router } from 'react-router-dom'
+import React, { useMemo } from 'react'
+import { Route, Switch, BrowserRouter as Router, Link as RouterLink } from 'react-router-dom'
 {=# rootComponent.isDefined =}
 {=& rootComponent.importStatement =}
 {=/ rootComponent.isDefined =}
@@ -16,6 +16,19 @@ import createAuthRequiredPage from "./auth/pages/createAuthRequiredPage"
 {=# isExternalAuthEnabled =}
 import OAuthCodeExchange from "./auth/pages/OAuthCodeExchange"
 {=/ isExternalAuthEnabled =}
+
+type Routes = 
+  {=# routes =}
+| {to: "{= urlPath =}", params{=^ urlParams =}?{=/ urlParams =}: {{=# urlParams =} {= . =}: string | number;{=/ urlParams =} }}
+  {=/ routes =}
+| never
+
+export function Link({ to, params, children }: Routes & { children: React.ReactNode }) {
+  const toWithParams = useMemo(() => {
+    return Object.keys(params).reduce((acc, key) => acc.replace(`:${key}`, params[key]), to)
+  }, [to, params])
+  return <RouterLink to={toWithParams}>{children}</RouterLink>
+}
 
 const router = (
   <Router>
--- a/waspc/test/Generator/WebAppGeneratorTest.hs
+++ b/waspc/test/Generator/WebAppGeneratorTest.hs
@@ -80,7 +80,7 @@ spec_WebAppGenerator = do
                     (SP.toFilePath Common.webAppSrcDirInWebAppRootDir </>)
                     [ "logo.png",
                       "index.tsx",
-                      "router.jsx"
+                      "router.tsx"
                     ]
                 ]
 


waspc/data/Generator/templates/react-app/src/router/routes.ts
```diff
@@ -0,0 +1,47 @@
...
+  hash?: Hash,
+) {
+  const interpolatedPath = params
+  ? path.split('/').map((part) => {
+      if (part.startsWith(':')) {
+        const paramName = part.slice(1)
+        return params[paramName]
+      }
+      return part
+    }).join('/')
```
sodic, line 42: I recommend extracting this into a named function.

waspc/data/Generator/templates/react-app/src/router/routes.ts
```diff
@@ -0,0 +1,47 @@
...
+  {=/ routes =}
+}
+
+type OptionalRouteOptions = {
+  search?: Search;
+  hash?: Hash;
+}
+
+type ParamValue = string | number
+type Params = Record<string, ParamValue>
```
sodic, line 25: **Extreme nitpick, but interesting (at least to me :))**

This is not wrong, but an index signature is probably a better choice. When typing objects whose keys are unbounded primitive values, it's more idiomatic to use index signatures:
```typescript
type Params = { [name: string]:  ParamValue };
```
While the two options are entirely equivalent from a type checking perspective, there is a difference is in how programmers perceive them:

*   A `Record` means "_you're guaranteed to have a defined mapping for every possible key_" (explained in [this comment](https://github.com/optimoroute/optimo/pull/2245#discussion_r710017104)). This semantic implication "works" only for bounded types (e.g., `'A' | 'B' | 'C'`). If a `Record` is used with an unbounded type (e.g., a `string`, as is the case here), its core idea falls apart.
*   An index signature doesn't aim to provide any semantic guarantees regarding the mapping's "completeness". In other words, it describes only a single entry (i.e., the type of the key and the type of the value). **Index signatures also have the benefit of providing information about what the key represents** (in this case, `eventName`). This can be very useful when the key is a primitive value and its type name carries no domain information (e.g., a `string` or a `number`) as is the case here.

[This Stack Overflow post](https://stackoverflow.com/a/54100359) does an excellent job of describing the differences.

The same goes for all `Record`s over unbounded types.

waspc/data/Generator/templates/react-app/src/router/routes.ts
```diff
@@ -0,0 +1,47 @@
...
+type Search = string[][] | Record<string, string> | string | URLSearchParams
+type Hash = string
```
sodic, line 27: Why the type alias?

waspc/data/Generator/templates/react-app/src/router/routes.ts
```diff
@@ -0,0 +1,47 @@
+{{={= =}=}}
+export type Routes = 
+  {=# routes =}
+| {to: "{= urlPath =}", params{=^ urlParams =}?{=/ urlParams =}: {{=# urlParams =}{= . =}: string | number;{=/ urlParams =}}}
```
sodic, line 4: Why are `params` always optional, even for routes that don't use them?

waspc/data/Generator/templates/react-app/src/router/Link.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export function Link(
+  { to, params, search, hash, ...restOfProps }: Omit<RouterLinkProps, "to" | "params" | "search" | "hash">
+  & {
+    search?: Record<string, string>;
+    hash?: string;
+  }
+  & Routes
+) {
+  const toWithParams = useMemo(() => {
```
sodic, line 15: Nitpick: I'd rename it to `toPropWithParams` simply because `to` is a verb.

sodic, line 15: How come we used `useMemo` (might be correct, I always forget the rules)?

sodic (commented):
Tested it out, seems to work nicely. I just have a couple of questions/suggestions in the comments

waspc/data/Generator/templates/react-app/src/router/routes.ts
infomiho: Haskell inspired me, removed it 👍

waspc/data/Generator/templates/react-app/src/router/Link.tsx
```diff
@@ -0,0 +1,19 @@
+import { useMemo } from 'react'
+import { Link as RouterLink } from 'react-router-dom'
+import { type Routes, interpolatePath } from './routes'
+
+type RouterLinkProps = Parameters<typeof RouterLink>[0]
+
+export function Link(
+  { to, params, search, hash, ...restOfProps }: Omit<RouterLinkProps, "to" | "params" | "search" | "hash">
```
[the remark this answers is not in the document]
infomiho: Removed it 👍  I wanted to future proof it, but there isn't any real need for that

waspc/src/Wasp/Generator/WebAppGenerator/RouterGenerator.hs
```diff
@@ -87,15 +93,37 @@ instance ToJSON ExternalAuthProviderTemplateData where
         "authProviderEnabled" .= _authProviderEnabled externalProviderTD
       ]
 
-genRouter :: AppSpec -> Generator FileDraft
-genRouter spec = do
+genRouter :: AppSpec -> Generator [FileDraft]
+genRouter spec =
+  sequence
+    [ genRouterTsx spec,
```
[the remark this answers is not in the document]
infomiho: I suppose you meant `createRouterTemplateData` being called twice, and I agree, I'll try to organise stuff so we can avoid calling it twice 👍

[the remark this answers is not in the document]
infomiho: As for everything using the same definition, I haven't figured it out 🤷

infomiho:
@sodic can you take another look, I've addressed the comments?

waspc/data/Generator/templates/react-app/src/router/routes.ts
sodic: Got it. Does this have implications on the functionality (i.e., will the user have worse type checking support because of it)?

waspc/data/Generator/templates/react-app/src/router/linkHelpers.ts
```diff
@@ -0,0 +1,49 @@
...
+function interpolatePathParams(path: string, params: Params) {
+  const startsWithSlash = path.startsWith("/");
+  const interpolatedPath = path
+    .split("/")
+    .map((part) => {
+      if (part.startsWith(":")) {
+        const paramName = getParamName(part);
+        return params[paramName];
+      }
+      return part;
```
sodic, line 32: We might have misunderstood each other here. I meant extracting this part into a named function as well:
```typescript
if (part.startsWith(':')) {
    const paramName = getParamName(part)
    return params[paramName]
}
return part
```
Seems a little long for a lambda and easily nameable. If we can put a name on it, I say go for it. I know you're using a closure, but placing the definition inside the parent function should take care of that.

waspc/src/Wasp/Generator/WebAppGenerator/RouterGenerator.hs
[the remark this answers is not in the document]
sodic: > We usually call this file (and its function) index.ts and genIndexTs and place it inside the router dir. How come we didn't do that here?

What about this part?

waspc/data/Generator/templates/react-app/tsconfig.json
```diff
@@ -4,7 +4,9 @@
     // Temporary loosen the type checking until we can address all the errors.
     "jsx": "preserve",
     "allowJs": true,
-    "strict": false
+    "strict": false,
+    // Allow importing pages with the .tsx extension.
+    "allowImportingTsExtensions": true
```
sodic, line 9: Why did we need to add this one?

waspc/data/Generator/templates/react-app/src/router/linkHelpers.ts
```diff
@@ -0,0 +1,49 @@
...
+    })
+    // Drops empty strings
```
sodic, line 34: I recommend removing all the comments explaining what the code does (especially for non-complex stuff).

If you think the code is unclear and would benefit from an explanation, it's better to extract it into a named function or assign the complex expression to a named variable.

Great video on the topic: https://www.youtube.com/watch?v=Bf7vDBBOBUA

waspc/data/Generator/templates/react-app/src/router/linkHelpers.ts
```diff
@@ -0,0 +1,49 @@
+export type ParamValue = string | number
+export type Params = { [name: string]: ParamValue }
+export type Search =
+  | string[][]
+  | Record<string, string>
```
sodic, line 5: Same story like here: https://github.com/wasp-lang/wasp/pull/1235#discussion_r1300022520.

The semantic implications of `Record`s and index signatures are a nitpick, but I think naming a key would be beneficial here.

The same applies to other similar `Record`s.

waspc/ChangeLog.md
sodic, line 1: I really like how you worded the docs. Nice work!

waspc/examples/crud-testing/src/client/MainPage.tsx
```diff
@@ -1,13 +1,16 @@
...
+import { Link, routes } from "@wasp/router";
 
 import { tasks as tasksCrud } from "@wasp/crud/tasks";
-import { User, Task } from "@wasp/entities";
+import { User } from "@wasp/entities";
 
 const MainPage = ({ user }: { user: User }) => {
   const { data: tasks, isLoading } = tasksCrud.getAll.useQuery();
+
+  type Task = NonNullable<typeof tasks>[0];
```
sodic, line 12: Nice trick to avoid importing the entity.

Small tip: Writing `number` instead of `0` is more idiomatic, but doesn't really change anything.

web/docs/advanced/links.md
```diff
@@ -0,0 +1,130 @@
...
+
+  return (
+    <div>
+      {tasks.map((task) => (
+        <Link
+          key={task.id}
+          to="/task/:id"
+      {/* 👆 You must provide a valid path here */} 
+          params={{ id: task.id }}>
+      {/* 👆 All the params must be correctly passed in */}
```
sodic, line 34: I'd skip the pointer emojis, but to each their own :)

web/docs/advanced/links.md
```diff
@@ -0,0 +1,130 @@
...
+
+The `Link` component accepts the following props:
+- `to` <Required />
+
+  - A valid Wasp Route path from your `main.wasp` file.
+
+- `params: { [name: string]: string | number }` <Required /> (if the path contains params)
+
+  - An object with keys and values for each param in the path.
+  - For example, if the path is `/task/:id`, then the `params` prop must be `{ id: 1 }`. Wasp supports required and optional params..
```
sodic, line 87:
```suggestion
  - For example, if the path is `/task/:id`, then the `params` prop must be `{ id: 1 }`. Wasp supports required and optional params.
```

web/docs/advanced/links.md
```diff
@@ -0,0 +1,130 @@
...
+
+- `search: string[][] | Record<string, string> | string | URLSearchParams`
+
+  - Any valid input for `URLSearchParams` constructor.
+  - For example, you can pass an object like this: `{ sortBy: 'date' }` and it will be converted to `?sortBy=date`.
```
sodic, line 92:
```suggestion
  - For example, the object `{ sortBy: 'date' }` becomes `?sortBy=date`.
```

web/docs/advanced/links.md
sodic, line 1: Great docs here as well!

sodic (approved):
Looks good to me. I left a couple of extra suggestions. Take the ones you like

waspc/src/Wasp/Generator/WebAppGenerator/RouterGenerator.hs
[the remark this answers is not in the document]
sodic: > As for everything using the same definition, I haven't figured it out 🤷

This is fine since we're generating code. If we ever decide to improve it, a good question to ask is "If this was an app I'm writing, how would I handle route duplication" and then generate that. My guess is probably a `routes` dictionary with snake-case names converted to camelCase names. The same dict could also contain params.

From that, I think we could generate everything else (including the types). But no need to do it now IMO.

waspc/data/Generator/templates/react-app/tsconfig.json
infomiho: We commented on this IRL, the fact that the router file change from `jsx` to `tsx` made the pattern of defining Wasp pages with `.tsx` incorrect. 

For example, if the user defined `MyPage.tsx` and it got imported by Wasp in `router.tsx`, this wouldn't allow building for production.

We agreed to keep the router as `router.jsx` and just reexport TS types from it (this seems to work just fine) to avoid modifying the `tsconfig.json` if we don't have to. 

We will rethink this in the future when we go for solving the general issue of imports with ts vs. js vs none extensions.

waspc/data/Generator/templates/react-app/src/router/linkHelpers.ts
infomiho: Commented IRL: we copied the type from the built-in `URLSearchParams` class

web/docs/advanced/links.md
infomiho: If we plan on restricting emojis in docs in the new style-guide for docs, I'll remove it, np 👍

waspc/examples/crud-testing/src/client/MainPage.tsx
infomiho: Nice stuff, I didn't know that 👍

web/docs/advanced/links.md
infomiho: I'll leave it for since we have them elsewhere, but I'll remove them if we start removing emojis from all parts of the docs.

waspc/src/Wasp/Generator/WebAppGenerator/RouterGenerator.hs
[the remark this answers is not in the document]
infomiho: > > We usually call this file (and its function) index.ts and genIndexTs and place it inside the router dir. How come we didn't do that here?

> What about this part?

I've just renamed the file from `router.jsx` to `router.tsx` so I can't speak to the previous file organisation 🤷 this feels okay 

> From that, I think we could generate everything else (including the types). But no need to do it now IMO.

I've done this in the latest iteration 👍 thanks for the push

waspc/data/Generator/templates/react-app/src/router/routes.ts
infomiho: Now with the latest iteration, it's a bit different, but to answer the initial question => users get all the benefits they expect from type-safe routes.

waspc/data/Generator/templates/react-app/src/router/types.ts
```diff
@@ -0,0 +1,33 @@
...
+  hash?: string;
+};
+
+export type ParamValue = string | number
+export type Params = { [name: string]: ParamValue }
+export type Search =
+  | string[][]
+  | Record<string, string>
+  | string
+  | URLSearchParams
```
sodic, line 33: Missing empty line.

waspc/data/Generator/templates/react-app/src/router/linkHelpers.ts
```diff
@@ -0,0 +1,46 @@
...
+
+function getPathPartsMapper(params: Params) {
+  return function mapPathPart(part: string): string | number | undefined {
+    if (part.startsWith(":")) {
+      const paramName = extractParamNameFromPathPart(part);
+      return params[paramName];
+    }
+    return part;
+  }
+}
```
sodic, line 35: I find this a little complicated. Perhaps we had a misunderstanding, I meant "just put the definition of `mapPathPart` under `interpolatePathParams`.

If you prefer to do it with a higher-order function, that's fine too. I 'd still recommend against it since the "fixed" value (i.e., `params`) is always the same within the call of `interpolatePathParams`. In other words, its closure does not need to support being changed.

sodic (approved):

PR #1236, opened 2023-06-07
Narrows down what can be saved when signing up

infomiho:


waspc/ChangeLog.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## v0.10.7
+
+### Security fixes
+- Wasp signup saves only a subset of fields from the signup form to the database. This prevents users from injecting arbitrary data into the database.
```
Martinsos, line 6: This is a breaking change, because Waspers might have used this functionality to store some extra fields, so let's mark it as a breaking change and bump Wasp version also! Maybe we can also shortly explain here what they need to do if they used more fields -> "you can't pass extra fields, if you did, you will have to set those after sign up, in a second step that you will manually do. In the future we will allow defining additional fields that will become part of signup form and signup action" -> something like that.

Martinsos (approved):
Great! Just pls take a look at that comment I left on Changelog.

waspc/ChangeLog.md
infomiho: > might have used this functionality to store some extra fields, so let's mark it as a breaking change

I've changed it to *Breaking change* but we never claimed we support extra fields so it's not really breaking a promise. But I remember what Filip says: "the way the users use it, that's the official way"

Is the Changelog really the right place for this:

> you can't pass extra fields, if you did, you will have to set those after sign up, in a second step that you will manually do. In the future we will allow defining additional fields that will become part of signup form and signup action

It feels more like an addition to the docs. I'll add it to the docs.

infomiho: Updated Changelog and docs

Martinsos:
Yup, @sodic said that correctly :D.

Looking good, good call adding it to docs!

I would still add that sentence about "second step" also to changelog, it is convenient.

Anyway all good with me, merge when ready!

PR #1248, opened 2023-06-09
Upgraded React and Vite to latest version (React 18).

### LINKED ISSUE #723: Update React to version 18, plus a few other React libs
Once React 18 support hits a few more popular libraries and is used by most of the ecosystem, we should look into updating to it.

Since this update will likely be a breaking change, we should also consider updating:
- ~react-scripts to version 5~ we are using Vite now
- React Router to version 6

Martinsos:
Replaces React 17 with React 18!

Fixes #723 .

TODO:
 - [ ] Update Changelog and cabal version.
 - [ ] Check example apps / starters and upgrade them as necessary. Realworld will certainly need updating due to old MUI.


waspc/data/Generator/templates/react-app/tsconfig.node.json
```diff
@@ -1,6 +1,7 @@
 {
   "compilerOptions": {
     "composite": true,
+    "skipLibCheck": true,
```
infomiho, line 4: @Martinsos Why did we. do this change?

waspc/src/Wasp/Generator/WebAppGenerator.hs
```diff
@@ -135,10 +135,10 @@ npmDepsForWasp spec =
         AS.Dependency.fromList
           [ -- TODO: Allow users to choose whether they want to use TypeScript
             -- in their projects and install these dependencies accordingly.
-            ("vite", "^4.1.0"),
-            ("typescript", "^4.9.3"),
-            ("@types/react", "^17.0.53"),
-            ("@types/react-dom", "^17.0.19"),
+            ("vite", "^4.3.9"),
+            ("typescript", "^4.9.3"), -- Once upgraded to Typescript 5, consider using moduleResolution: 'bundler' in tsconfig.node.json, since that is what Vite now by default uses (we have it set to 'Node' right now because 'bundler' requires TS >= 5). Also, upgrade @tsconfig/vite-react to 2.0.0 (couldn't do it now because it also uses 'bundler').
```
infomiho, line 139: Maybe we merge @sodic's PR first and then we can do this as well in this PR?

infomiho:
When testing with existing apps, I had to do `wasp clean` to make sure there aren't weird interaction going on between the old `package-lock.json` and the new dep version changes.

infomiho:
- `tutorials/todoApp` - works
- `tutorials/ItWaspsOnMyMachine` - works

waspc/data/Generator/templates/react-app/tsconfig.node.json
Martinsos: New vite adds that when you scaffold new project with it. I read about what it is, and was convinced that it makes sense during development, as it can speed things up.
I am ok with removing it, but if that is their default, I concluded it makes sense we also have it as default.

Martinsos:
@infomiho kudos for testing all that!
We are in a pretty good shape then!
ItWaspsOnMyMachine -> I think we can slowly deprecate that one, as we haven't used it much lately.
waspello and realworld -> we can create GH issues to update those, and proceed in the meantime.

sodic:
Since Miho already took care of everything, I'll handle the conflicts and merge it.

waspc/src/Wasp/Generator/WebAppGenerator.hs
sodic: I merged the TS version update and upgraded `vite-react` to 2.0.0. We inherit from `vite-react`'s config, and version 2.0.0 sets `module: bundler`, so we're good.

waspc/data/Generator/templates/react-app/tsconfig.node.json
sodic: If it's the default for Vite, let's keep it.

Issue #769, opened 2022-10-17
Have Wasp enforce unique username on Prisma schema level


Martinsos:
We currently do it on the JS level.

Martinsos:
Duplicate of #1164

PR #1257, opened 2023-06-15
Ensure Auth entity has a unique username 

### LINKED ISSUE #1164: Wasp requires "username" in User to be unique, but doesn't check for that
This results with errors later down the road, if this expectation is broken -> I forgot where exactly, but it was rather inconvenient.

This is easy to fix so let's do it.

sodic:
Fixes #1164

I couldn't find any JS code that deals with this: https://github.com/wasp-lang/wasp/issues/769#issue-1411767135

### DIFF
--- a/waspc/src/Wasp/AppSpec/Valid.hs
+++ b/waspc/src/Wasp/AppSpec/Valid.hs
@@ -171,13 +171,13 @@ validateAuthUserEntityHasCorrectFieldsIfUsernameAndPasswordAuthIsUsed spec = cas
       usernameAttributeValidationErrors
         | isFieldUnique "username" userEntity == Just True = []
         | otherwise =
-          [ GenericValidationError $
-              "The field 'username' on entity '"
-                ++ userEntityName
-                ++ "' (referenced by "
-                ++ authUserEntityPath
-                ++ ") must be marked with the '@unique' attribute."
-          ]
+            [ GenericValidationError $
+                "The field 'username' on entity '"
+                  ++ userEntityName
+                  ++ "' (referenced by "
+                  ++ authUserEntityPath
+                  ++ ") must be marked with the '@unique' attribute."
+            ]
       userEntityFields = Entity.getFields userEntity
       authUserEntityPath = "app.auth.userEntity"
       (userEntityName, userEntity) = AS.resolveRef spec (Auth.userEntity auth)


waspc/src/Wasp/AppSpec/Entity.hs
```diff
@@ -45,5 +47,15 @@ getPslModelBody = pslModelBody
...
+doesFieldHaveAttribute :: String -> String -> Entity -> Maybe Bool
+doesFieldHaveAttribute fieldName attrName entity =
+  doesPslFieldHaveAttribute attrName <$> findPslFieldByName fieldName entity
+
+findPslFieldByName :: String -> Entity -> Maybe PslModel.Field
+findPslFieldByName fieldName Entity {pslModelBody = PslModel.Body elements} =
+  find isField [field | (PslModel.ElementField field) <- elements]
+  where
+    isField PslModel.Field {_name = name} = name == fieldName
+
```
sodic, line 59: @infomiho @Martinsos 

I wasn't sure how to decide what to put in here vs. in `Psl/Util.hs`.

Also, this is the only file that uses `Psl/Util.hs`, should we maybe remove it (especially considering the conversation about keeping `PslModel`  stuff hidden)?

sodic:
@Martinsos What did you mean by this comment: https://github.com/wasp-lang/wasp/issues/769#issue-1411767135. Should I also remove some JS code?

waspc/src/Wasp/AppSpec/Entity.hs
Martinsos: I am missing a bit of context so it is somewhat hard for me to make a decisive statement here, but:
1. If we have functions operating on PSL stuff and not knowing anything about Entity, then it sounds good to have them in PslUtil.hs .

For future, it might help if you give more context. For exmaple, what is in Psl/Util.hs, why do you think it shouldn't be there (besides this being the only place using it), what are pros/cons of each approach (aka "what have you tried"), what do getIdField and getIdBlockAttribute do to make them candidates for this refactoring.

Martinsos:
> @Martinsos What did you mean by this comment: [#769 (comment)](https://github.com/wasp-lang/wasp/issues/769#issue-1411767135). Should I also remove some JS code?

Unfortunately I can't remember, I wish I wrote more info. I guess we, when creating user via Signup, check there, in the Prisma request, if username is unique? But we don't enforce it in the schema itself?

waspc/src/Wasp/AppSpec/Entity.hs
Martinsos: Also, I like the idea of exposing Entity + ENtity.Filed interface and hiding PSL if possible. If not possible / practical, we can let some PSL seep out.
I can't judge right now how possible / practical it is without diving deeper, but I believe you will know that better!

infomiho: I liked the split of having `Entity` using helper functions from `Psl/Util.hs`.

Example of such fn:
```hs
getIdField :: Entity -> Maybe PslModel.Field
getIdField = findIdField . getPslModelBody
```

All the low level work on the PSL schema is hidden away in the `PslUtil` module, while Entity is composing those function to get out what they need without worrying about the implementation. 

For this PR, I think the split could go like this:
1. `Psl/Util.hs` implements functions for checking if certain field in PSL has some attribute attached `isAttributePresentOnField` 
2. `Entity.hs` uses that function to come up with a function `isFieldUnique` (looking for `@unique`)
3. `Valid.hs` can then do something like `Entity.isFieldUnique "username"`

waspc/test/AppSpec/EntityTest.hs
```diff
@@ -8,10 +8,17 @@ import qualified Wasp.Psl.Ast.Model as PslModel
 spec_AppSpecEntityTest :: Spec
 spec_AppSpecEntityTest = do
   describe "getIdField" $ do
-    it "gets primary field from entity when it exists" $ do
```
infomiho, line 11: We have 19 test files where we write the first letter as upper case and 21 files where we write it as lower case.

Obviously we have two competing standards 😄 @sodic @Martinsos

waspc/src/Wasp/AppSpec/Entity.hs
sodic: The main idea was as follows:

If we're hiding PSL as an implementation detail (this is the core assumption), there probably aren't many reasons for having a `Utils` file (in its current location).

The only module what should ever be using those utils is `Entity` (since all PSL stuff is a "hidden" implementation detail), so we might as well move all those functions there (instead of having them globally available from a different part of the codebase).

Such an organisation colocates all relevant code and discourages/prevents us from acting on PSL directly.

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -149,14 +150,31 @@ validateAuthUserEntityHasCorrectFieldsIfUsernameAndPasswordAuthIsUsed spec = cas
...
+      usernameTypeValidationErrors =
+        validateEntityHasField
+          userEntityName
+          userEntityFields
+          ("username", Entity.Field.FieldTypeScalar Entity.Field.String, "String")
+      usernameAttributeValidationErrors
+        | doesFieldHaveAttribute "username" "unique" userEntity == Just True = []
+        | otherwise =
+          [ GenericValidationError $
+              "The field 'username' on Entity referenced by " ++ userEntityName ++ " must have the '@unique' attribute."
```
infomiho, line 173: I would go for less technical error message like
```
The username field on the User entity should be marked as unique (add the @unique attribute on the username field).
```

If some new users this super technical message, it might confuse them IMHO

infomiho:
Tested it out locally, it works. Some minor comments regard the error message and code organisation.

waspc/src/Wasp/AppSpec/Valid.hs
sodic: I agree, but the rest of the file uses this style so I decided against changing it to save time.

infomiho: Consider changing this message and the others (I see only one other: https://github.com/wasp-lang/wasp/blob/afe90a20e6427e720e9fc39678c38eff5ae8bb35/waspc/src/Wasp/AppSpec/Valid.hs#L245) if you have the time 👍

waspc/src/Wasp/AppSpec/Entity.hs
sodic: I've added the `isFieldUnique` function and left the other stuff as is for now.

Martinsos: @sodic not so important for this PR, but I would do the opposite -> I would put PSL specific stuff in PSL/Utils.hs or something like that, not in `Entity`, IF it makes sense to be standalone. This is useful because it clearly indicates that this code is not Entity specific, in the sense that there are some assumptions they share. Entity would still be having that as an implementation details, that doesn't change. And the fact that nobody else is using it is also not an issue, if it is general it is general -> we might need it in some other parts of the code, even if also hidden as an implementation detail.
Otherwise we could just move the whole PSL module under Entity. But I like it better that we say "hey, look we have PSL module because Wasp knows how to work with it!" without assuming who is using it.

But this is not a biggie, we can in any case move it around if needed.

waspc/test/AppSpec/EntityTest.hs
Martinsos: Ha I also wasn't sure, but sounds good it if is nicer to read! Git messages yes, for sure, upper letter and dot at the end!

Issue #1272, opened 2023-06-21
Social login callback page sends the token twice


infomiho:
Our social flow includes redirecting from e.g. Google to our frontend with an auth token in the query params. 

When our frontend page loads, it sends a request to our backend which then returns a token to the frontend.

After upgrading to React 18 and enabling their new `StrictMode`, each component is rendered twice in development. This means, our social auth callback page is rendered twice -> which sends the auth token twice to the backend.

This is an issue because the token is only valid **once** and the second request ends with an `500` error on our server which then logs out the user.

In a nutshell, our social auth flow is broken in development (it would still work in production builds).

Martinsos:
Any ideas how to fix this?

infomiho:
We need to redesign a bit how we send that request in the component, and it should be fine. Either we'll use an `AbortController` to cancel requests or we'll modify the query after the request is sent for the first time. 

I still have to investigate. 

https://stackoverflow.com/a/72238236

PR #1275, opened 2023-06-21
Fixes oAuth redirect behaviour in development

infomiho:
Fixes https://github.com/wasp-lang/wasp/issues/1272

Based on https://stackoverflow.com/questions/73732474/handling-oauth-with-react-18-useeffect-hook-running-twice

Fixing it this way is not recommended by the docs, but they don't offer any alternatives for this particular use case (oauth redirect page): https://react.dev/learn/synchronizing-with-effects#how-to-handle-the-effect-firing-twice-in-development


Martinsos (approved):
@infomiho great! I get that this is tricky hm, but seems like good fix for now.
Could you maybe just add a comment in the code, desribing what you commented in the PR (that we did it because of StrictMode and don't know of a better solution at the moment)?

PR #1280, opened 2023-06-22
Adjust how dictionary types are displayed

craigmc08:
Adjusts the `Show` instance for `Wasp.Analyzer.Type.Type` to make the resulting text more readable.

**Before**
```
{
  foo: DictRequired string,
  bar: DictOptional number,
}
```

**After**
```
{
  foo: string,
  bar?: number
}
```


infomiho (approved):
🛳 it

PR #1283, opened 2023-06-23
Rewrite templates to include description

infomiho:
- Templates repo now has a `templates.json` file which lists all templates https://github.com/wasp-lang/starters/blob/main/templates.json
- Interactive CLI can now show an option's description
- We now include the `description` from the JSON when listing the options

Screenshot 2023-06-23 at 12 29 13

### DIFF
--- a/waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
+++ b/waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
@@ -1,9 +1,9 @@
 module Wasp.Cli.Command.CreateNewProject.StarterTemplates
-  ( getStarterTemplateInfos,
-    StarterTemplateInfo (..),
+  ( getStarterTemplates,
+    StarterTemplate (..),
     TemplateMetadata (..),
-    findTemplateNameByString,
-    defaultStarterTemplateInfo,
+    findTemplateByString,
+    defaultStarterTemplate,
   )
 where
 
@@ -12,7 +12,7 @@ import Data.Foldable (find)
 import qualified Wasp.Cli.Command.CreateNewProject.StarterTemplates.Remote.Github as Github
 import qualified Wasp.Cli.Interactive as Interactive
 
-data StarterTemplateInfo = RemoteStarterTemplate TemplateMetadata | LocalStarterTemplate TemplateMetadata
+data StarterTemplate = RemoteStarterTemplate TemplateMetadata | LocalStarterTemplate TemplateMetadata
   deriving (Eq)
 
 data TemplateMetadata = TemplateMetadata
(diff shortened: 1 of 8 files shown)


waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
```diff
@@ -1,41 +1,63 @@
...
+        )
+
+localTemplates :: [StarterTemplateInfo]
+localTemplates = [defaultStarterTemplateInfo]
+
+defaultStarterTemplateInfo :: StarterTemplateInfo
+defaultStarterTemplateInfo = LocalStarterTemplate $ TemplateMetadata {_name = "basic", _path = "basic", _description = "Simple starter template with a single page."}
+
+findTemplateNameByString :: [StarterTemplateInfo] -> String -> Maybe StarterTemplateInfo
 findTemplateNameByString templateNames query = find ((== query) . show) templateNames
```
Martinsos, line 63:
```suggestion
findTemplateByString :: [StarterTemplateInfo] -> String -> Maybe StarterTemplateInfo
findTemplateByString templateInfos query = find ((== query) . show) templateInfos
```

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/Remote/Github.hs
```diff
@@ -9,3 +11,26 @@ starterTemplateGithubRepo =
       _repoName = "starters",
       _repoReferenceName = "main"
     }
+
+starterTemplatesDataGithubFilePath :: String
```
Martinsos, line 15:
```suggestion
starterTemplatesDataGithubFilePath :: FilePath
```

waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -39,17 +38,16 @@ import qualified Wasp.Util.Terminal as Term
   without having to type the quotes as well.
 
   We introduced the Option class to get different "show" behavior for Strings and other
-  types. Option delegates to the Show instance for all other types, but for Strings it
-  just returns the String itself.
+  types. If we are using something other then String, an instance of Option needs to be defined,
```
Martinsos, line 41: Yeah I think this is sane default.

waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -85,8 +83,18 @@ askToChoose question options = do
...
+            <> (if isDefaultOption option then " (default)" else "")
+            <> showDescription option (length indexPrefix)
+          where
+            indexPrefix = showIndex idx <> " "
+
+        showIndex i = "[" ++ show (i :: Int) ++ "]"
+
+        showDescription option indentLength = case showOptionDescription option of
+          Just description -> "\n" <> replicate indentLength ' ' <> description
+          Nothing -> ""
```
Martinsos, line 97: Nice! You can rewrite this using `maybe` if you wish:
```hs
showDescription option indentLength =
  maybe "" ("\n" <> replicate indentLength ' ' <>) $
    showOptionDescription option
```

but what you have is also nice.

Martinsos (approved):
@infomiho looking great!

There is just one thing I would do: drop `Info` word. Just let it be `StarterTemplate`. You anyway use `RemoteStarterTemplate` and `LocalStartTemplate` for data constructors, you don't have Info there.
As clean code says, Info often doesn't bring much value.

If you do wish to keep it, I am ok with that, but personally I would remove it now before it spreads further.

waspc/cli/src/Wasp/Cli/Interactive.hs
infomiho: I've played around with 3 different ways of writing this and this felt like the most readable one :)

PR #1284, opened 2023-06-23
improve debouncer tests

craigmc08:
Rewrites debouncer tests so that they don't fail randomly.

Tested by running the tests 1000 times with 32 instances in parallel on an 8 core CPU to stress test the thread timing.

I was unable to figure out a way to test that the debouncer debounces actions when they are started within the debounce time, any test I tried was too flaky.


infomiho (approved):
🛳️ it

PR #1288, opened 2023-06-23
Wasp ai retry

Martinsos:


waspc/src/Wasp/Util/IO.hs
```diff
@@ -122,3 +126,22 @@ isDirectoryEmpty :: Path' Abs (Dir d) -> IO Bool
...
   (files, dirs) <- listDirectory dirPath
   return $ null files && null dirs
+
+-- | Takes number of tries so far and returns number of microseconds to pause before next try.
+type RetryDelayStrategy = (Int -> Int)
+
+-- | Runs given action, and then if it fails, retries it, up to maxNumRetries.
+--   Uses provided delayStrategy to calculate pause between tries.
+-- TODO: Write tests.
+retry :: (MonadUnliftIO m) => RetryDelayStrategy -> Int -> m (Either e a) -> m (Either e a)
```
craigmc08, line 136:
```suggestion
retry :: (MonadIO m) => RetryDelayStrategy -> Int -> m (Either e a) -> m (Either e a)
```

`MonadUnliftIO` is unnecessarily restrictive, you only use `liftIO`.

waspc/src/Wasp/Util/IO.hs
```diff
@@ -122,3 +126,22 @@ isDirectoryEmpty :: Path' Abs (Dir d) -> IO Bool
...
+  where
+    go numFailedRetries = do
+      action >>= \case
+        Right result -> pure $ Right result
+        Left e ->
+          if numFailedRetries < maxNumRetries
+            then do
+              liftIO $ threadDelay $ delayStrategy (numFailedRetries + 1)
+              go (numFailedRetries + 1)
+            else pure $ Left e
```
craigmc08, line 147: I keep confusing myself with `numFailedRetries` and if it should be called `numFailedTries` or not. On L139 it's the number of tries so far, on L143 it's the number of retries so far, and on L145 & 146 `numFailedRetries + 1` is again the number of tries so far.

I think the function is understandable how it is, it just took me a while to convince myself that there is no off-by-one error in the number of retries it will do.

waspc/src/Wasp/AI/OpenAI/ChatGPT.hs
```diff
@@ -13,13 +13,19 @@ module Wasp.AI.OpenAI.ChatGPT
...
+import Control.Monad.IO.Class (MonadIO (liftIO))
 import Data.Aeson ((.=))
 import qualified Data.Aeson as Aeson
 import Data.ByteString.UTF8 as BSU
 import Data.Text (Text)
 import GHC.Generics (Generic)
+import qualified Network.HTTP.Conduit as HTTP.C
 import qualified Network.HTTP.Simple as HTTP
+import UnliftIO (MonadUnliftIO)
+import UnliftIO.Exception (catch, throwIO)
```
craigmc08, line 26: Is there a reason you use `catch` and `throwIO` from `unliftio` instead of `base`?

craigmc08 (changes requested):

waspc/src/Wasp/AI/OpenAI/ChatGPT.hs
Martinsos: I read in article by FPComplete that it is better to use the one from unliftio so I use that one through the codebase! Although I can't say I have a super strong understanding of the matter. What is your opinion on this?

Here is the article: https://www.fpcomplete.com/haskell/tutorial/exceptions/ .

waspc/src/Wasp/Util/IO.hs
Martinsos: You are right, I rewrote it something and I think now it is cleaner, take a look pls.

Martinsos:
@craigmc08 I fixed some stuff / improved code + improved a bit how pause strategy is specified.

One thing I tried to do but failed was write tests though.

I tried going the route of introducing something called `MonadRetry`, and then say that `retry` operates with `m` that is both `MonadIO` and `MonadRetry`, but just couldn't get that to work. Ideally, I was hoping I would have some kind of `MonadRetry` that I can then mock in tests with State monad and track calls to threadDelay. Do you have any advice on how to best test this function?

Martinsos:
@craigmc08 Actually I take back what I said about tests, it seems I got them working! I overcomplicated situation a bit, made it simpler now and I think this approach will make sense. Still, please take a look, maybe you will have an idea for a better approach.

waspc/src/Wasp/AI/OpenAI/ChatGPT.hs
craigmc08: I gave that article a read over the weekend, lots of interesting stuff! Seems like there are several good reasons to use the ones from unliftio instead of from base, so lets stick with these ones.

waspc/src/Wasp/Util/IO.hs
craigmc08: Looks good now!

craigmc08 (approved):
The tests look good! Doing it by testing a list of events is smart.

PR #1290, opened 2023-06-26
Updates changelog

infomiho:


waspc/ChangeLog.md
```diff
@@ -12,8 +12,145 @@
...
+    socket.emit('chatMessage', messageText)
+    // ...
+  }
+
+  // ...
+}
+```
+
+### Automatic CRUD backend generation
+For some Entity, you can tell Wasp to automatically generate server-side logic (Queries and Actions) for creating, reading, updating and deleting such entities. As your entities update, Wasp will automatically regenerate the backend logic.
```
craigmc08, line 114:
```suggestion
You can tell Wasp to automatically generate server-side logic (Queries and Actions) for creating, reading, updating, and deleting a specific entity. As you change that entity, Wasp automatically regenerates the backend logic.
```

craigmc08 (commented):
Should a list of new features be listed at the top of the changelog along with the breaking changes and bug fixes sections? So that there is a short summary before going into detail about each feature.

PR #1292, opened 2023-06-26
[waspls] cleanup extimport code

craigmc08:


waspc/packages/ts-inspect/package.json
```diff
@@ -11,6 +11,10 @@
     "start": "node ./dist/index.js",
     "test": "jest"
   },
+  "//": [
+    "typescript is purposefully a normal dependency, and not a dev dependency:",
+    "the compiler is called at runtime to extract information from JS/TS files."
+  ],
```
craigmc08, line 17: This is apparently how people usually add comments to package.json files, since they don't support actual comments.

waspc/packages/ts-inspect/test/exports.test.ts
```diff
@@ -1,34 +1,36 @@
 import * as path from 'path';
 import { getExportsOfFiles } from "../src/exports";
 
+// TODO(before merge): run these tests in CI
```
craigmc08, line 4: Looking for help with this, I don't know what the best way to go about this is.

waspc/waspls/src/Wasp/LSP/Debouncer.hs
```diff
@@ -14,31 +14,75 @@ import Data.Foldable (traverse_)
 import Data.Hashable (Hashable)
 import qualified StmContainers.Map as STM
```
Martinsos, line 16: This is much better now!

I would maybe also add, as the module-level comment, an explanation of what "debounce" means -> You could give a short intro saying that when we want to detect an event but there is some noise in the signal that indicates the event making it look like it happened many times in short time while it really happened only once, the common technique is to use "debouncing" -> detecting spurious events and recognizing them all together as one event. This is commonly used in electronics, for example for recognizing that user pressed a button (because many "button pressed" events get sent suddenly).
This module implements a simple abstraction for dealing with such events, where you can define how to debounce them. Maybe even put a link to https://en.wikipedia.org/wiki/Switch#Contact_bounce .

craigmc08: :rocket:

I left out the part about electronics and the link to Wikipedia. I think connecting it to that is a little confusing.

waspc/packages/ts-inspect/test/exports.test.ts
Martinsos: Happy to help but give me a bit more info: which part are you not sure about, what have you tried?

I am not sure myself but I guess we would just add a step to our github actions yaml file that runs tests from this package. You can find that file at the top of our repo, in .github/workflows/ci.yaml . Or did you have some specific dillema?

craigmc08: I was mostly hoping you/someone would already know what the correct way to do this is for npm packages. But I'll look it up and see what people say.

Martinsos (approved):
@craigmc08 I believe we covered all of the comments from the original PR, so approving this! Nice work!
If there is anything you think is left unaddressed or you would like me to take additional look at, let me know and I will -> otherwise let's merge!

PR #1293, opened 2023-06-26
Wasp ai retry json

Martinsos:


infomiho (approved):
LGTM, but I feel like using functions could save us the headaches of retries. What if it fails the second time?

PR #1298, opened 2023-06-28
Show multiple parse errors from Wasp Analyzer

craigmc08:
- Adds support for returning multiple parse errors from Wasp Analyzer and displays all of these to the terminal when wasp files are compiled.
- Adjusts parser so that it returns all concrete parser errors from the parser.
- Adds new, limited error recovery to the abstract parser. It recovers at the boundaries of statements, dictionary entries, and values within lists and tuples.


Martinsos (approved):
LGTM, awesome job @craigmc08 !

PR #1299, opened 2023-06-28
Wasp ai fresh files

Martinsos:
Files now flash as they are updated!

I also fixed behaviour of folders -> they weren't expanding always -> so I added `expandedNodes` property to TreeView. Seems that `defaultExpandedNodes` is not enough.


infomiho (approved):
LGTM, I've updated the styles a bit

PR #1305, opened 2023-07-04
Help Chat GPT understand Wasp's `useAction` hook

sodic:


waspc/src/Wasp/AI/GenerateNewProject/PageComponentFile.hs
```diff
@@ -63,7 +63,6 @@ fixPageComponent newProjectDetails waspFilePath pageComponentPath = do
             - Make sure to use only queries and actions that are defined in the Wasp file (listed below)!
             - Ensure query and action imports are correct. One import per query / action, default imports,
               name of the file same as name of the query.
-            - Don't use `useAction` or `useMutation`! Use actions directly.
```
Martinsos, line 66: I would not just delete it but replace it with this:
```
 - Don't use `useMutation`, instead use `useAction`! Example of usage: `const createTaskFn = useAction(createTask);`.
```

waspc/src/Wasp/AI/GenerateNewProject/Page.hs
```diff
@@ -150,20 +152,26 @@ pageDocPrompt =
...
-        One import statement per query/action.
+        Here are the rules for importing actions and queries.
 
-        Note: there is no `useMutation` or `useAction` hook to import in Wasp. Actions are to be called directly.
+        If a query is called "myQuery", its import should look like `import myQuery from '@wasp/queries/myQuery';`.
+        It has to be a default import, and name of the file is the same as name of the query.
+        The hook for wrapping queries is called `useQuery`.
+        Use a single import statement per query.
+
+        If an action is called "myAction", its import should look like `import myAction from '@wasp/actions/myAction';`.
```
Martinsos, line 169: I wonder if you could just say, instead of repeating the whole thing, smth like `It is analogous for actions`, like we had before, I think that worked? We save a bit of tokens.

sodic: I could try, but this is probably safer. Is the token difference significant?

waspc/src/Wasp/AI/GenerateNewProject/PageComponentFile.hs
sodic: I don't think it's necessary. We already told it about the `useAction` hook the same way we told it about the `useQuery` hook. These bullet points don't mention `useQuery` (or `useAction`) and they seem to work well.

waspc/src/Wasp/AI/GenerateNewProject/Page.hs
sodic: Tried it, it doesn't really work. It does this:
```typescript
import { useQuery, useAction } from '@wasp/queries';
```

PR #1306, opened 2023-07-05
[wasp-ai] Asks GPT to list its fixes

infomiho:
### Asking the GPT to list the fixes

We are experimenting with asking the GPT to list out the fixes it did, hoping it would lead to more coherent answers.

### Outputting the original and fixed versions of the files

We added debug output of original and fixed versions of generated files.


Martinsos:
@infomiho should we close this one, since we didn't conclude it helps? Or is there something in here that we want to salvage?

infomiho:
This proved not to bring that much value, so I am okay with closing it 👍

PR #1307, opened 2023-07-06
Added prisma TS package to Wasp, for running prisma format.

Martinsos:


craigmc08 (commented):
Before I start a more thorough review, I'm wondering why this is written as a TypeScript package instead of in Haskell.

This package saves a schema to a temp file and runs `npx prisma format --schema [tmpfile]`. Can't we do this directly from Haskell? That should also have the benefit that we don't need to manually ensure this package keeps it's prisma version in-sync with waspc, since we can get the version from `Wasp.Generator.Common.prismaVersion` and specify it in the call to npx.

Martinsos:
> Before I start a more thorough review, I'm wondering why this is written as a TypeScript package instead of in Haskell.
> 
> This package saves a schema to a temp file and runs `npx prisma format --schema [tmpfile]`. Can't we do this directly from Haskell? That should also have the benefit that we don't need to manually ensure this package keeps it's prisma version in-sync with waspc, since we can get the version from `Wasp.Generator.Common.prismaVersion` and specify it in the call to npx.

The problem is that I will be using this from Wasp AI, where I am not in a wasp project, so I don't have a setting in which prisma is installed already. So I need to be able to call this while not in a Wasp project. So I thought if we have a package like this, we can install deps inside it and call prisma that way. Otherwise where do I get prisma from is the question?

craigmc08:
> Otherwise where do I get prisma from is the question?

`npx` installs the specified package and dependencies to the npm cache if there is no local/package copy of it.

Martinsos:
> > Otherwise where do I get prisma from is the question?
> 
> `npx` installs the specified package and dependencies to the npm cache if there is no local/package copy of it.

Yeah upf maybe I could have done that -> I actually started with using `@prisma/internals`, which is prisma's internal sdk, because it was supposed to have `formatSchema` function, so that is why I did the whole thing, but then it turned out that that function doesn't work well, so I switched to calling `prisma format` directly.

But you are right that I could have done this directly from haskell + use npx with specified prisma version. Ay yay!

This package could be useful for the future if we are to use any prisma-related TS packages directly, but yeah right now it is not needed. I don't have time to rewrite this to haskell though so I would certainly use it for now, but I will then not merge it into main but into wasp-ai instead and then we can figure out what to do with it.

craigmc08 (approved):
Gave the code a once over, looks good! Nice tests too.

Martinsos:
Closing, added to wasp-ai branch instead.

PR #1308, opened 2023-07-06
Help Wasp AI with imports and prevent it from generating logout/login actions

sodic:


waspc/src/Wasp/AI/GenerateNewProject/Page.hs
```diff
@@ -26,20 +27,43 @@ import Wasp.AI.GenerateNewProject.Common
...
+        opName = T.pack $ Plan.opName $ opPlan operation
+        opType = case Operation.opType operation of
+          Operation.Action -> "actions"
+          Operation.Query -> "queries"
+
 generateAndWritePage ::
   NewProjectDetails -> FilePath -> [Plan.Entity] -> [Operation] -> [Operation] -> Plan.Page -> CodeAgent Page
-generateAndWritePage newProjectDetails waspFilePath entityPlans actions queries pPlan = do
-  page <- generatePage newProjectDetails entityPlans actions queries pPlan
+generateAndWritePage newProjectDetails waspFilePath entityPlans queries actions pPlan = do
```
sodic, line 58: `actions` and `query` parameters were in a different order than arguments sent in `generateNewProject`, causing the prompt to look like this:
```
          Actions in our app:
           - { "name": "createBird",
            "jsImpl": "export const createBird = async (args, context) => {\n  if (!context.user) { throw new HttpError(401) }\n\n  const { breedId, name, pictureUrl } = args\n\n  const bird = await context.entities.Bird.create({\n    data: {\n      breed: { connect: { id: breedId } },\n      name,\n      pictureUrl,\n      user: { connect: { id: context.user.id } }\n    }\n  })\n\n  return bird\n}"
          }
           - { "name": "logout",
            "jsImpl": "export const logout = async (args, context) => {\n  context.logout()\n}"
          }

          Queries in our app:
           - { "name": "getBirds",
            "jsImpl": "export const getBirds = async (args, context) => {\n  if (!context.user) { throw new HttpError(401) }\n\n  return context.entities.Bird.findMany({\n    where: {\n      user: { id: context.user.id }\n    },\n    include: {\n      breed: true\n    }\n  })\n}"
          }
           - { "name": "getBreeds",
            "jsImpl": "export const getBreeds = async (args, context) => {\n  if (!context.user) { throw new HttpError(401) }\n\n  return context.entities.Breed.findMany()\n}"
          }
```
I've decided to settle on a convention: queries first, actions second.

waspc/src/Wasp/AI/GenerateNewProject/Plan.hs
```diff
@@ -99,6 +99,8 @@ generatePlan newProjectDetails planRules = do
         Typically, plan will have AT LEAST one query, at least one action, at least one page, and at
         least two entities. It will very likely have more than one of each, though.
 
+        DO NOT create actions for login and logout under any circumstances. They are already included in Wasp.
```
sodic, line 102: If the user's prompt mentions anything about logging in and logging out, it sometimes creates actions for those functionalities. Hopefully this helps.

Martinsos: What we can also potentially do to handle this is to add a bit of logic in Plan.hs, under `fixPlan`, that checks if any action is named `login` or `logout` or maybe contains those words -> and if there are such actions, we warn it there that we suspect those to be redundant because it is already included in wasp and that it should remove them.

Martinsos (approved):
LGTM @sodic nice ~string editing~ work!

This is looking solid -> I left just one comment as an idea for how to possibly ensure it doesn't add login/logout actions.

PR #1310, opened 2023-07-07
[waspls] implement goto definition for declarations

craigmc08:
Allows you to jump to the definition of declarations from uses of the declaration.

Demo:
decl-goto-def


Martinsos (approved):
LGTM @craigmc08 !

PR #1319, opened 2023-07-13
Martech

sodic:
Keep track of referrer and zip download data in the database.

### DIFF
--- a/wasp-ai/src/server/operations.ts
+++ b/wasp-ai/src/server/operations.ts
@@ -5,6 +5,7 @@ import { checkPendingAppsJob } from "@wasp/jobs/checkPendingAppsJob.js";
 
 export const startGeneratingNewApp: StartGeneratingNewApp<
   {
+    referrer: string;
     appName: string;
     appDesc: string;
     appPrimaryColor: string;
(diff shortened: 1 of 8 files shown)


Martinsos (approved):

wasp-ai/src/server/operations.ts
```diff
@@ -53,6 +58,22 @@ export const startGeneratingNewApp: StartGeneratingNewApp<
...
+  try {
+    await context.entities.Project.update({
+      where: { id: appId },
+      data: {
+        zipDownloadedAt: new Date(),
+      },
+    });
+  } catch (e) {
+    throw new HttpError(404, "App not found.");
+  }
```
Martinsos, line 74: This is not great @sodic because you could be hiding quite a few errors this way!

Martinsos (approved):

wasp-ai/src/server/operations.ts
Martinsos: Ok I fixed it! Also fixed another occurrence of this in the same file, where you probably copied this from.

sodic: Where copilot copied it from*. I've written 5 characters and pressed tab three times.

PR #1329, opened 2023-07-19
[docs] Removes type annotation in JS code in the tutorial

### LINKED ISSUE #1323: Todo App Tutorial Step 4 Issue
Hi all,

I'm working through the Wasp "TodoApp" tutorial, and reached an error that prevented me from moving forward on this page: https://wasp-lang.dev/docs/tutorials/todo-app/04-creating-tasks

Under the "Invoking the Action on the Frontend" section, there is a try/catch within the handleSubmit function that reads as follows:

```
const handleSubmit = async (event) => {
    event.preventDefault()
    try {
      const target = event.target
      const description = target.description.value
      target.reset()
      await createTask({ description })
    } catch (err: any) {
      window.alert("Error: " + err.message)
    }
  }
```

When I got to this step, the wasp server stopped hosting the page due to a compilation error because type declarations can only be used in TypeScript files. Removing the type as shown below allowed me to proceed.

```
   ...
    } catch (err) {
   ...
```
Hope this helps!

craigmc08:
Fixes #1323

### DIFF
--- a/web/docs/tutorials/todo-app/04-creating-tasks.md
+++ b/web/docs/tutorials/todo-app/04-creating-tasks.md
@@ -153,7 +153,7 @@ const NewTaskForm = () => {
       const description = target.description.value
       target.reset()
       await createTask({ description })
-    } catch (err: any) {
+    } catch (err) {
       window.alert("Error: " + err.message)
     }
   }


infomiho (approved):
⛴️ it

PR #1335, opened 2023-07-21
[docs] Point the page edit to release branch

infomiho:
### Description
When users click on "Edit this page" in our docs, it will point them to the `release` branch instead of the `main` branch. 

This is more correct since we publish docs from the `release` branch and it will make the PR process more correct since it will be based off of the `release` branch.


Martinsos (approved):
Great!

PR #1336, opened 2023-07-21
Add Custom Domain docs

infomiho:
### Description

Adds instructions for using a Custom domain with Wasp CLI deployment to Fly.io


Martinsos (approved):
Awesome initiative @infomiho !

One question: we still do not have "general" custom domain instructions, these are now for Fly. Could we have general ones, and then also have specific ones for Fly? Would that make sense?

infomiho:
I'll add a part about making sure to setup the env variables correctly to keep the CORS setup working 🙃

PR #1339, opened 2023-07-21
Adds APIs docs to the new structure

infomiho:
Adds APIs docs with Javscript/Typescript toggle.

It also explains how to add CORS support if you want to use the APIs from the client (a common question).


web/docs/advanced/apis.md
```diff
@@ -1,3 +1,224 @@
...
+  return (
+    <>
+      // ...
+    </>
+  )
+}
+```
+
+#### Making sure CORS works
+Since the APIs don't use the default middleware like Operations do to make APIs as flexible as possible. As a side-effect, you need to make sure CORS is enabled for your APIs to use them on the client.
```
Martinsos, line 144: `Since` here doesn't really fit.

Martinsos (approved):
Did a very quick read, left 1 comment, looking good from the distance!

Issue #110, opened 2020-10-30
Add support for tests


Martinsos:
Right now, there is no support in Wasp for testing, and we should look into adding it.

Testing external JS code should not be so hard -> we should set some conventions on how and where to write tests, probably using Jest, and we can then execute them. There might be some unique challenges regarding mocking and similar, but they should be solvable.

Testing Wasp code is different -> since it is aiming to be simple and declarative, unit tests should not be needed, and it would anyway be harder to implement them since it would need to be completely custom solution.
But, e2e tests of the whole app, those are important. We should offer good support for those, probably via Cypress or TestCafe -> this is too complex for us to do on our own, we should just delegate it to existing solutions, but somehow nicely integrate it with Wasp.

So, to recap: We should look into unit testing of JS code via Jest and into e2e testing via Cypress or Testcafe, and find a way to integrate those with Wasp.

Martinsos:
Some ideas that were suggested by community member on Discord:

> Related to testing, I recently read about Testcontainers, not sure how it fits into Wasp but the idea is neat. Can use typescript to create containers running postgres/redis/elasticsearch/etc. and control their lifecycle directly from typescript
>   For even more control over the CI/CD environment, there is also Dagger, but it is more devops-y than Testcontainers which comes with nice defaults
> https://dagger.io/

Martinsos:
We now have frontend testing support! We don't yet have support for backend.

PR #1353, opened 2023-07-24
Adds Testing docs in the new structure

craigmc08:
Content is mostly the same from the original page, but was reorganized a bit.

JS/TS toggle added to the code blocks, but the code is identical for both languages.

### DIFF
--- a/web/docs/project/testing.md
+++ b/web/docs/project/testing.md
@@ -1,3 +1,171 @@
 ---
 title: Testing
 ---
+
+:::info
+Wasp is in beta, so keep in mind there might be some kinks / bugs, and possibly some changes with testing support in the future. If you encounter any issues, reach out to us on [Discord](https://discord.gg/rzdnErX) and we will make sure to help you out!
+:::
+
+## Testing Your React App
+
+Wasp enables you to quickly and easily write both unit tests and React component tests for your frontend code. Because Wasp uses [Vite](https://vitejs.dev/), we support testing web apps through [Vitest](https://vitest.dev/).
+
+<details>
+  <summary>Included Libraries</summary>
+  <div>
+
+  [`vitest`](https://www.npmjs.com/package/vitest): Unit test framework with native Vite support.
+
+  [`@vitest/ui`](https://www.npmjs.com/package/@vitest/ui): A nice UI for seeing your test results.
+
+  [`jsdom`](https://www.npmjs.com/package/jsdom): A web browser test environment for Node.js.
+
+  [`@testing-library/react`](https://www.npmjs.com/package/@testing-library/react) / [`@testing-library/jest-dom`](https://www.npmjs.com/package/@testing-library/jest-dom): Testing helpers.
+
+  [`msw`](https://www.npmjs.com/package/msw): A server mocking library.
+
+  </div>
+</details>
+
+### Setting Up Tests
+
+In order for Wasp to pick up your tests, they should be placed within the `src/client` directory and use an extension that matches [these glob patterns](https://vitest.dev/config#include). Some of the file names that Wasp will pick up as tests:
+
+- `yourFile.test.ts`
+- `YourComponent.spec.jsx`
+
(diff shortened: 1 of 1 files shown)


web/docs/project/testing.md
```diff
@@ -1,3 +1,171 @@
...
+</details>
+
+### Setting Up Tests
+
+In order for Wasp to pick up your tests, they should be placed within the `src/client` directory and use an extension that matches [these glob patterns](https://vitest.dev/config#include). Some of the file names that Wasp will pick up as tests:
+
+- `yourFile.test.ts`
+- `YourComponent.spec.jsx`
+
+Within test files, you can import your other source files as usual. For example, if you have a component `Counter.jsx`, you test it by creating a file in the same directory called `Counter.test.jsx` and import the component with `'./Counter'`.
```
infomiho, line 37:
```suggestion
Within test files, you can import your other source files as usual. For example, if you have a component `Counter.jsx`, you test it by creating a file in the same directory called `Counter.test.jsx` and import the component with `import Counter from './Counter'`.
```
It looks a bit off just having the single quotes without the `import` part

web/docs/project/testing.md
```diff
@@ -1,3 +1,171 @@
...
+### Running Tests
+
+Running `wasp test client` will start Vitest in watch mode and recompile your Wasp project when changes are made.
+
+- If you want to see a realtime UI, pass `--ui` as an option.
+- To run the tests just once, use `wasp test client run`.
+
+All arguments after `wasp test client` are passed directly to the Vitest CLI, so check out [their documentation](https://vitest.dev/guide/cli.html) for all of the options.
+
+:::warning
```
infomiho, line 48: Screenshot 2023-07-25 at 11 16 08

Nit: You can write extra text after `warning` to change the title of the warning to something like `Be careful` which reads better IMHO than `Danger`

Screenshot 2023-07-25 at 11 17 56

web/docs/project/testing.md
```diff
@@ -1,3 +1,171 @@
...
+  - Behind the scenes, Wasp uses [`msw`](https://npmjs.com/package/msw) to create a server request handler that responds with the given JSON.
+  - Mocks are cleared after each test.
+- `mockApi`: Similar to `mockQuery`, but for [APIs](/docs/advanced/apis) instead.
+  - Instead of a Wasp Query, it takes a route of the shape: `mockApi({ method: HttpMethod.Get, path: '/foo/bar' }, { res: 'hello' })`.
+  - You can import `HttpMethod` like so: `import { HttpMethod } from '@wasp/types'`.
+- `renderInContext`: Takes a React component, wraps it inside a `QueryClientProvider` and `Router`, and renders it.
+
+## Testing Your Server-Side Code
+
+Coming soon!
```
infomiho, line 67: I would maybe expand this to say something "We currently don't support testing your server-side code, but we will be adding support soon. You can track the progress [here](https://github.com/wasp-lang/wasp/issues/110) and express your interest by commenting on the Github issue.`

web/docs/project/testing.md
```diff
@@ -1,3 +1,171 @@
...
+
+## Examples
+
+### Client Unit Tests
```
infomiho, line 71: I would maybe add a bit more examples e.g. https://github.com/wasp-lang/wasp/blob/main/waspc/examples/todoApp/src/client/pages/auth/helpers.test.ts

web/docs/project/testing.md
```diff
@@ -1,3 +1,171 @@
...
+
+test('areThereAnyTasks', () => {
+  expect(areThereAnyTasks([])).toBe(false)
+})
+```
+
+</TabItem>
+</Tabs>
+
+### React Component Tests
```
infomiho, line 102: We don't have an example for using the `mockApi` helper, maybe add as well.

infomiho: Check out the new docs merged in the parent branch, they should be better.

I see APIs an escape hatch which enables defining Express routes. It just "raw" backend endpoints.

They are used from the client by using `axios` via our `api` wrapper. So, `mockApi` should mock some JSON e.g. list of TODOs. And then the component is expected to render a list of `li`s.

PR #1366, opened 2023-07-26
[Docs] Added 'language' page

Martinsos:
I removed old language/overview page because it is redundant now.
I moved old language/syntax page under general/language .

For this, I added new section: `general`. I was thinking about putting it under `misc`, but there we have stuff that is not code-related, like contributing, contact, telemetry -> so more like meta stuff, extra stuff.

Instead, I wanted to put general stuff somewhere. I imagine under `general` we could also later add something like `cli`.

If you have better ideas where to put this I am open to that.


craigmc08:
I agree that this page doesn't fit in any other section and that "General" would be the best name for it, but I think wanting to put something in "General" is a sign that it's actually reference information -> so language/CLI docs probably don't belong in the Docs area but in an API Reference area (that we don't have yet).

infomiho:
I'll go ahead and merge this so we can reorganise it with everything else in context if needed 👍

PR #1369, opened 2023-07-26
[docs restructuring] Static Public Files

Martinsos:
Moved documentation for static public files.

I did do some changes, as it seemed to me there was some duplication and that some re-ordering might help a bit -> @infomiho please give it a look! I removed the details about the build dir in the dist directory, since those are implementation details they don't need to know about, and they don't mean much to them.

Before:
image

After:
image


infomiho (approved):
⛴️ it

PR #1372, opened 2023-07-26
[waspls] fix client directory module resolution

craigmc08:
### Description

Previously, waspls incorrectly thought that importing `@client/file.jsx` could be referencing the file `src/client/file.tsx`. But this JS/TS "merging" is only correct on the server (for reasons that are not explained by the TS documentation on module resolution). This PR fixes that by requiring exact matches for external imports beginning with `@client`.

This incorrect behavior manifested as several different bugs:
- Not reporting missing `.jsx` files when a `.tsx` file with the same name exists.
- Go-to definition going to the `.tsx` file instead of the `.jsx` file when the external import ends in `.jsx`.
- Scaffold actions suggestions scaffolding in both the `.jsx` and `.tsx`files.

This PR should fix all of those bugs.

### DIFF
--- a/waspc/waspls/src/Wasp/LSP/ExtImport/Path.hs
+++ b/waspc/waspls/src/Wasp/LSP/ExtImport/Path.hs
@@ -62,10 +62,10 @@ waspStylePathToCachePath (WaspStyleExtFilePath waspStylePath) =
     Just relPath ->
       Just $ case SP.splitRelExtension relPath of
         Nothing -> ExtFileCachePath relPath DotAnyTS
-        Just (withoutExt, ext) ->
+        Just (relPathWithoutExt, ext) ->
           if useExactExtension
-            then ExtFileCachePath withoutExt (DotExact ext)
-            else ExtFileCachePath withoutExt (widenExtension ext)
+            then ExtFileCachePath relPathWithoutExt (DotExact ext)
+            else ExtFileCachePath relPathWithoutExt (widenExtension ext)
   where
     useExactExtension = "@client" `isPrefixOf` waspStylePath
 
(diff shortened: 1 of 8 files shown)


waspc/waspls/src/Wasp/LSP/ExtImport/Path.hs
```diff
@@ -56,11 +57,18 @@ instance Hashable ExtFileCachePath where
...
-  relFile <- P.parseRelFile =<< stripPrefix "@" waspStylePath
-  let (extensionLessFile, extType) = splitExtensionType relFile
-  return $ ExtFileCachePath (SP.fromPathRelFile extensionLessFile) extType
+waspStylePathToCachePath (WaspStyleExtFilePath waspStylePath) =
+  case stripPrefix "@" waspStylePath >>= SP.parseRelFile of
+    Nothing -> Nothing
+    Just relPath ->
+      Just $ case SP.splitRelExtension relPath of
+        Nothing -> ExtFileCachePath relPath DotAnyTS
+        Just (withoutExt, ext) ->
```
infomiho, line 66:
```suggestion
        Just (pathWithoutExt, ext) ->
```

infomiho:
**Ideally:** we would inspect the file's AST and do the changes there and put everything back. It would put imports on top. And, if an import from a file exists, we would just add a new import symbol to the list. This feels like a future improvement.

>  So my thinking was: make it obvious that the import needs to be rewritten a bit by putting it with the rest of the code.

I feel that's a very valid reason to keep it as-is.

craigmc08:
> **Ideally**: we would inspect the file's AST and do the changes there and put everything back. It would put imports on top. And, if an import from a file exists, we would just add a new import symbol to the list. This feels like a future improvement.

I agree, this would definitely be the best way to do it, but will be a lot of work. And it sounds like you're okay with keeping the import in this slightly weird place, so let's stick what that.

I also addressed the comment you left.

infomiho (approved):
⛴️ it

PR #1374, opened 2023-07-27
[Docs] Data model

sodic:


infomiho:
I would suggest adding a `Options Reference` on the bottom of each page (check out some of the other pages that have it)

web/docs/database/backends.md
```diff
@@ -1,3 +1,182 @@
...
+### Connecting to an existing database
+
+If instead of using `wasp start db` you would rather spin up your own dev database or connect to some external database, you will need to provide Wasp with `DATABASE_URL` environment variable that Wasp will use to connect to it.
+
+The easiest way to provide the needed `DATABASE_URL` environment variable is by adding it to the [.env.server](https://wasp-lang.dev/docs/language/features#env) file in the root dir of your Wasp project (if that file doesn't yet exist, create it).
+
+You can also set it per command by doing `DATABASE_URL=<my-db-url> wasp ...` -> this can be useful if you want to run specific `wasp` command on a specific database.
+Example: you could do `DATABASE_URL=<my-db-url> wasp db seed myProdSeed` to seed data for a fresh staging or production database.
+
+### Migrating from SQLite to PostgreSQL
```
infomiho, line 73:
```suggestion
## Migrating from SQLite to PostgreSQL
```

web/docs/database/backends.md
```diff
@@ -1,3 +1,182 @@
...
+
+To run Wasp app in production, you will need to switch from `SQLite` to `PostgreSQL`.
+
+1. Set `app.db.system` to `PostgreSQL`.
+2. Delete old migrations, since they are SQLite migrations and can't be used with PostgreSQL: `rm -r migrations/`.
+3. Run `wasp start db` to start your new db running (or check instructions above if you prefer using your own db). Leave it running, since we need it for the next step.
+4. In a different terminal, run `wasp db migrate-dev` to apply new changes and create new, initial migration.
+5. That is it, you are all done!
+
+### Seeding
```
infomiho, line 83:
```suggestion
## Seeding
```

web/docs/database/backends.md
```diff
@@ -1,3 +1,182 @@
...
+
+**Database seeding** is a term for populating database with some initial data.
+
+Seeding is most commonly used for two following scenarios:
+
+1.  To put development database into a state convenient for testing / playing with it.
+2.  To initialize dev/staging/prod database with some essential data needed for it to be useful,
+    for example default currencies in a Currency table.
+
+#### Writing a seed function
```
infomiho, line 93:
```suggestion
### Writing a seed function
```

web/docs/database/backends.md
```diff
@@ -1,3 +1,182 @@
...
+): Promise<SanitizedUser> {
+  const { password, ...newUser } = await prismaClient.user.create({ data });
+  return newUser;
+}
+```
+
+</TabItem>
+</Tabs>
+
+#### Running seed functions
```
infomiho, line 174:
```suggestion
### Running seed functions
```

web/docs/database/backends.md
```diff
@@ -1,3 +1,182 @@
 ---
-title: Database Backends
```
infomiho, line 2: I would add at the end an `Options Reference` section with:
- `db`
- `db.system`
- `db.seeds`
- `db.prisma`
fields explained. 

When you do that, you'll see that you can delete the duplicate explanations above and keep the document simpler.

web/docs/database/backends.md
```diff
@@ -1,3 +1,182 @@
...
+title: Databases
 ---
+
+[Entities](/docs/database/entities.md), [Operations](/docs/database/operations.md) and [Automatic CRUD](/docs/database/crud.md) together make a high-level interface for working with your app's data. Still, all that data has to live somewhere, so let's see how Wasp deals with databases.
+
+## Database configuration
+
+You can tall Wasp which database to use in the `app` declaration's `db` field:
+
+```wasp
```
infomiho, line 11: I usually added Javascript/Typescript tabs for Wasp files as well to keep ti consistent. Also, the user can toggle and see, oh okay, I should put the `.js` extension even when I'm using Typescript.

web/docs/database/backends.md
```diff
@@ -1,3 +1,182 @@
...
+      import devSeed from "@server/dbSeeds.js"
+    ],
+    prisma: {
+      clientPreviewFeatures: ["extendedWhereUnique"]
+    }
+  }
+}
+```
+
+`app.db` is a dictionary with following fields (all fields are optional):
```
infomiho, line 27: So this piece would go into the `Options Reference` and here you could talk about the options in general or not at all, just link to the reference.

web/docs/database/backends.md
```diff
@@ -1,3 +1,182 @@
...
+
+  Defines seed functions that you can use via `wasp db seed` to seed your database with initial data.
+  Check out [Seeding](#seeding) section for more details.
+
+- `prisma: [PrismaOptions]`
+
+  Additional configuration for Prisma.
+  It currently only supports a single field:
+
+- `clientPreviewFeatures : string`
```
infomiho, line 45: I would inline this one more step so it's under `prisma`

web/docs/database/backends.md
```diff
@@ -1,3 +1,182 @@
 ---
-title: Database Backends
+title: Databases
 ---
+
+[Entities](/docs/database/entities.md), [Operations](/docs/database/operations.md) and [Automatic CRUD](/docs/database/crud.md) together make a high-level interface for working with your app's data. Still, all that data has to live somewhere, so let's see how Wasp deals with databases.
+
+## Database configuration
+
+You can tall Wasp which database to use in the `app` declaration's `db` field:
```
infomiho, line 9:
```suggestion
You can tell Wasp which database to use in the `app` declaration's `db` field:
```

web/docs/database/backends.md
infomiho:
```suggestion
`app.db` is a dictionary with the following fields (all fields are optional):
```

web/docs/database/backends.md
```diff
@@ -1,3 +1,182 @@
...
+  It currently only supports a single field:
+
+- `clientPreviewFeatures : string`
+
+  Allows you to define [Prisma client preview features](https://www.prisma.io/docs/concepts/components/preview-features/client-preview-features).
+
+### SQLite
+
+Default database is `SQLite`, since it is great for getting started with a new project (needs no configuring), but it can be used only in development - once you want to deploy Wasp to production you will need to switch to `PostgreSQL` and stick with it.
+Check below for more details on how to migrate from SQLite to PostgreSQL.
```
infomiho, line 52: I would link to the title

web/docs/database/backends.md
```diff
@@ -1,3 +1,182 @@
...
+
+Wasp provides `wasp start db` command that starts the default dev db for you.
+
+Your Wasp app will automatically connect to it once you have it running via `wasp start db`, no additional configuration is needed. This command relies on Docker being installed on your machine.
+
+### Connecting to an existing database
+
+If instead of using `wasp start db` you would rather spin up your own dev database or connect to some external database, you will need to provide Wasp with `DATABASE_URL` environment variable that Wasp will use to connect to it.
+
+The easiest way to provide the needed `DATABASE_URL` environment variable is by adding it to the [.env.server](https://wasp-lang.dev/docs/language/features#env) file in the root dir of your Wasp project (if that file doesn't yet exist, create it).
```
infomiho, line 68: We need to fix the link

web/docs/database/backends.md
```diff
@@ -1,3 +1,182 @@
...
+
+You can also set it per command by doing `DATABASE_URL=<my-db-url> wasp ...` -> this can be useful if you want to run specific `wasp` command on a specific database.
```
infomiho, line 70: I would avoid using `->` since it makes the docs feel like a draft. Let's use proper interpunction 😄

web/docs/database/backends.md
```diff
@@ -1,3 +1,182 @@
...
+Example: you could do `DATABASE_URL=<my-db-url> wasp db seed myProdSeed` to seed data for a fresh staging or production database.
+
+### Migrating from SQLite to PostgreSQL
+
+To run Wasp app in production, you will need to switch from `SQLite` to `PostgreSQL`.
+
+1. Set `app.db.system` to `PostgreSQL`.
+2. Delete old migrations, since they are SQLite migrations and can't be used with PostgreSQL: `rm -r migrations/`.
+3. Run `wasp start db` to start your new db running (or check instructions above if you prefer using your own db). Leave it running, since we need it for the next step.
```
infomiho, line 79:
```suggestion
3. Run `wasp start db` to start your new db running (or check the instructions above if you prefer using your db). Leave it running, since we need it for the next step.
```

web/docs/database/backends.md
```diff
@@ -1,3 +1,182 @@
...
+4. In a different terminal, run `wasp db migrate-dev` to apply new changes and create new, initial migration.
+5. That is it, you are all done!
+
+### Seeding
+
+**Database seeding** is a term for populating database with some initial data.
+
+Seeding is most commonly used for two following scenarios:
+
+1.  To put development database into a state convenient for testing / playing with it.
```
infomiho, line 89:
```suggestion
1.  To put the development database into a state convenient for testing / playing with it.
```

web/docs/database/backends.md
```diff
@@ -1,3 +1,182 @@
...
+2.  To initialize dev/staging/prod database with some essential data needed for it to be useful,
+    for example default currencies in a Currency table.
+
+#### Writing a seed function
+
+Wasp enables you to define multiple **seed functions** via `app.db.seeds`:
+
+```wasp
```
infomiho, line 97: Typescript/javascript switch

web/docs/database/entities.md
```diff
@@ -1,3 +1,66 @@
...
+The language is declarative and very intuitive. We'll also go through an example later in the text, so there's no need to go and thoroughly learn it right away. Still, if you're curious, look no further than Prisma's official documentation:
+
+- [Basic intro and examples](https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-schema)
+- [A more exhaustive language specification](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference)
+
+## Defining an Entity
+
+As mentioned, an `entity` declaration represents a database model.
+
+Each `Entity` declaration corresponds 1-to-1 to [Prisma's data model](https://www.prisma.io/docs/concepts/components/prisma-schema/data-model). Here's how you could define an entity that represents a Task:
```
infomiho, line 20: I would use `entity` vs. `Entity` to keep ti consistent with the previous sentence

web/docs/database/entities.md
```diff
@@ -1,3 +1,66 @@
...
+    id          Int     @id @default(autoincrement())
+    description String
+    isDone      Boolean @default(false)
+psl=}
+```
+
+Let's go through this declaration in detail:
+
+- `entity Task` - This tells Wasp that we wish to define an Entity (i.e., database model) called `Task`. Wasp automatically creates a table called `tasks`.
+- `{=psl ... psl=}` - Wasp treats everything that comes between the two `psl` tags as [PSL](https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-schema).
```
infomiho, line 33:
```suggestion
- `{=psl ... psl=}` - Wasp treats everything that comes between the two `psl` tags as valid [Prisma schema language](https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-schema).
```

Maybe expand from `PSL` to full name to keep in mind that this _is_ Prisma under the hood

web/docs/database/entities.md
```diff
@@ -1,3 +1,66 @@
...
+
+The above PSL definition tells Wasp to create a table for storing Tasks where each task has three fields (i.e., the `tasks` table has three columns):
+
+- `id` - An integer value serving as a primary key (i.e., ID). The database automatically generates it by incrementing the previously generated `id`.
+- `description` - A string value for storing the task's description.
+- `isDone` - A boolean value indicating the task's completion status. If you don't set it when creating a new task, the database sets it to `false` by default.
+
+### Workflow with Entities
```
infomiho, line 41:
```suggestion
### Working with Entities
```

web/docs/database/entities.md
```diff
@@ -1,3 +1,66 @@
...
+
+Let's see how you can define and work with Wasp Entities:
+
+1. Create/update some Entities in your `.wasp` file.
+2. Run `wasp db migrate-dev`. This command syncs the database model with the Entity definitions in your `.wasp` file. It does this by creating migration scripts.
+3. Migration scripts are automatically placed in the `migrations/` folder. Make sure to commit this folder into version control.
+4. Use Wasp's JavasScript API to work with the database when implementing Operations (we'll cover this in detail on the [next page](/docs/database/operations.md))
```
infomiho, line 48:
```suggestion
4. Use Wasp's JavasScript API to work with the database when implementing Operations (we'll cover this in detail in the [operations docs](/docs/database/operations.md))
```

If you refer to it a "next page" it's not immune to order changes.

web/docs/database/entities.md
```diff
@@ -1,3 +1,66 @@
...
+
+Most of the time, you will be working with Entities within the context of [Operations (Queries & Actions)](/docs/database/operations.md). We'll see how that's done on the next page.
+
+#### Using Entities directly
+
+If you need more control, you can directly interact with Entities by importing and using the [Prisma Client](https://www.prisma.io/docs/concepts/components/prisma-client/crud). We recommend sticking with conventional Wasp-provided mechanisms, only resorting to directly using the Prisma client only if you need a feature Wasp doesn't provide.
+
+You can only use the Prisma Client in your Wasp server code. You can import it like this:
+
+```js
```
infomiho, line 60: I would put this in a JS/TS switch (just to keep it consistent) and also maybe expand the snippet to include usage e.g. `prismaClient.tasks.create({ ... })`

web/docs/database/crud.md
```diff
@@ -1,3 +1,746 @@
 ---
 title: Automatic CRUD
 ---
+
+import { Required } from '@site/src/components/Required';
+import { ShowForTs } from '@site/src/components/TsJsHelpers';
+
+Writing similar [Queries and Actions](/docs/database/operations) sometimes gets tiresome, especially if they are really simple. This is where Automatic CRUD really shines!
```
infomiho, line 8: I would motivate it differently:

> If you wrote a few full-stack apps, you probably noticed that some things tend to repeat: listing some data, creating new data, editing it and deleting it. We wanted to make it easier to handle these boring bits by offering a higher level concept Automatic CRUD.

Something along those lines, change it to what sounds good to you.

web/docs/database/crud.md
```diff
@@ -1,3 +1,746 @@
...
+
+## Example: A Simple TODO App
+
+Let's create a full-app example that uses automatic CRUD. We'll stick to using the `Task` entity from the previous example, but we'll add a `User` entity and enable [username and password](/docs/language/features#username-and-password) based auth.
+
+<ImgWithCaption alt="Automatic CRUD with Wasp" source="img/crud-guide.gif" caption="We are building a simple tasks app with username based auth"/>
+
+### Creating the App
+
+We can start by running `wasp new tasksCrudApp` and then we'll add the following to our `main.wasp` file:
```
infomiho, line 70:
```suggestion
We can start by running `wasp new tasksCrudApp` and then adding the following to the `main.wasp` file:
```

web/docs/database/crud.md
```diff
@@ -1,3 +1,746 @@
...
+      },
+    },
+  });
+};
+```
+
+</TabItem>
+</Tabs>
+
+We made a custom `create` operation because we want to make sure that the task is connected to the user that is creating it. In the current iteration of CRUD operations that is not supported by default. Read more about the [default implementations](/docs/language/features#which-operations-are-supported).
```
infomiho, line 219: Update the link to the default implementations

web/docs/database/crud.md
```diff
@@ -1,3 +1,746 @@
...
+return Task.delete({ where: { id: args.id } });
+```
+
+</TabItem>
+</Tabs>
+
+:::info Current Limitations
+In the default `create` and `update` implementations, we are saving all of the data that the client sends to the server. This is not always desirable, i.e. in the case when the client should not be able to modify all of the data in the entity.
+
+[In the future](#/docs/guides/crud#future-of-crud-operations-in-wasp), we are planning to add validation of action input, where only the data that the user is allowed to change will be saved.
```
infomiho, line 577: Update the link

web/docs/database/crud.md
```diff
@@ -1,3 +1,746 @@
...
+
+  The arguments of the operation i.e. the data sent from the client.
+
+- `context`
+
+  Context contains the `user` making the request and the `entities` object with the entity that's being operated on.
+
+<ShowForTs>
+
+You can also import types for each of the functions you want to override from `@wasp/crud/{crud name}`. The available types are:
```
infomiho, line 662: I would wrap this Typescript info with `<ShowForTs>`

web/docs/database/backends.md
```diff
@@ -1,3 +1,329 @@
...
+
+## Supported Database Backends
+
+Wasp supports multiple database backends. We'll list and explain each one.
+
+### SQLite
+
+The default database Wasp uses is [`SQLite`](https://www.sqlite.org/index.html).
+
+SQLite is a great way for getting started with a new project because it doesn't require any configuration, but Wasp can only use it in development. Once you want to deploy your Wasp app to production, you'll need to switch to `PostgreSQL` and stick with it.
```
infomiho, line 15:
```suggestion
SQLite is a great way for getting started with a new project because it doesn't require any configuration, but Wasp can only use it in development. Once you want to deploy your Wasp app to production, you'll need to switch to PostgreSQL and stick with it.
```

web/docs/database/backends.md
```diff
@@ -1,3 +1,329 @@
...
+
+[Entities](/docs/database/entities.md), [Operations](/docs/database/operations.md) and [Automatic CRUD](/docs/database/crud.md) together make a high-level interface for working with your app's data. Still, all that data has to live somewhere, so let's see how Wasp deals with databases.
+
+## Supported Database Backends
+
+Wasp supports multiple database backends. We'll list and explain each one.
+
+### SQLite
+
+The default database Wasp uses is [`SQLite`](https://www.sqlite.org/index.html).
```
infomiho, line 13:
```suggestion
The default database Wasp uses is [SQLite](https://www.sqlite.org/index.html).
```

web/docs/database/backends.md
```diff
@@ -1,3 +1,329 @@
...
+
+SQLite is a great way for getting started with a new project because it doesn't require any configuration, but Wasp can only use it in development. Once you want to deploy your Wasp app to production, you'll need to switch to `PostgreSQL` and stick with it.
+
+Fortunately, migrating from SQLite to PostgreSQL is pretty simple, and we have [a guide](#migrating-from-sqlite-to-postgresql) to help you.
+
+### PostgreSQL
+
+[PostgreSQL](https://www.postgresql.org/) is a real production database.
```
infomiho, line 21: This reads poorly, we should somehow say "this is rock solid battle tested database" without implying that SQLite is not (since it has proven itself recently) .

Something I found online:
> PostgreSQL is the world’s most advanced open source database and the fourth most popular database. In development for more than 20 years, PostgreSQL is managed by a well-organized and highly principled and experienced open source community. It is an object-oriented database that is fully ACID compliant and highly extensible, enabling the community to add new features and capabilities as workload demands evolved.

But maybe more relaxed 🤷

web/docs/database/backends.md
```diff
@@ -1,3 +1,329 @@
...
+
+To use Wasp with PostgreSQL, you'll have to ensure a database instance is running during development. Wasp needs access to your database for commands such as `wasp start` or `wasp db ...` and expects to find a connection string in the `DATABASE_URL` environment variable.
```
infomiho, line 23:
```suggestion
To use Wasp with PostgreSQL, you'll have to ensure a database instance is running during development. Wasp needs access to your database for commands such as `wasp start` or `wasp db migrate-dev` and expects to find a connection string in the `DATABASE_URL` environment variable.
```

I would be specific

web/docs/database/backends.md
```diff
@@ -1,3 +1,329 @@
...
+
+We cover all supported ways of connecting to a database in [the next section](#connecting-to-a-database).
+
+### Migrating from SQLite to PostgreSQL
+
+To run your Wasp app in production, you'll need to switch from `SQLite` to `PostgreSQL`.
```
infomiho, line 29:
```suggestion
To run your Wasp app in production, you'll need to switch from SQLite to PostgreSQL.
```

web/docs/database/backends.md
```diff
@@ -1,3 +1,329 @@
...
+
+1. Set `app.db.system` to `PostgreSQL`.
+2. Delete all old migrations, since they are SQLite migrations and can't be used with PostgreSQL: `rm -r migrations/`.
```
infomiho, line 32:
```suggestion
2. Delete all old migrations, since they are SQLite migrations and can't be used with PostgreSQL:

\`\`\`
rm -r migrations/`
\`\`\`
```

web/docs/database/backends.md
```diff
@@ -1,3 +1,329 @@
...
+3. Run `wasp start db` to start your new db running (or check the instructions above if you prefer using your own database). Leave it running, since we need it for the next step.
```
infomiho, line 33: This reads poorly "start your new db running" and it references some instructions that are no longer above

web/docs/database/backends.md
```diff
@@ -1,3 +1,329 @@
...
+2. A "real" database Wasp can connect to and use in production.
+
+### Using the Dev Database Provided by Wasp
+
+The command `wasp start db` will start a default PostgreSQL dev database for you.
+
+Your Wasp app will automatically connect to it, just keep `wasp start db` running in the background.
+Also, make sure that:
+
+- You have Docker installed and in `PATH`.
```
infomiho, line 51: Link to Docker installation for connivence

web/docs/database/backends.md
```diff
@@ -1,3 +1,329 @@
...
+- The port `5432` isn't taken.
+
+### Connecting to an existing database
+
+If, instead of using `wasp start db`, you'd rather spin up your own dev database (or connect to an external one), you'll need to tell Wasp about it using the `DATABASE_URL` environment variable. Wasp will use the value of `DATABASE_URL` as a connection string.
```
infomiho, line 56:
```suggestion
If you want to spin up your own dev database (or connect to an external one), you can tell Wasp about it using the `DATABASE_URL` environment variable. Wasp will use the value of `DATABASE_URL` as a connection string.
```

web/docs/database/backends.md
```diff
@@ -1,3 +1,329 @@
...
+}
+```
+
+</TabItem>
+</Tabs>
+
+Each seed function must be an async function that takes one argument, `prismaClient`, which is a [Prisma Client](https://www.prisma.io/docs/concepts/components/prisma-client/crud) instance used to interact with the database.
+This is the same Prisma Client instance that Wasp uses internally and thus includes all of the usual features (e.g., password hashing).
+
+Since a seed function falls under server-side code, it can import other server-side functions. This is convenient because you'll normally want to seed the database using Actions.
```
infomiho, line 128:
```suggestion
Since a seed function falls under server-side code, it can import other server-side functions. This is convenient because you might want to seed the database using Actions.
```

web/docs/database/backends.md
```diff
@@ -1,3 +1,329 @@
...
+  The database system Wasp should use. It can be either `PostgreSQL` or `SQLite`.
+  The default value for the field is `SQLite` (this default value also applies if the entire `db` field is left unset).
+  Whenever you modify the `db.system` field, make sure to run `wasp db migrate-dev` to apply the changes.
+
+- `seeds: [ServerImport]`
+
+  Defines the seed functions you can use with the `wasp db seed` command to seed your database with initial data.
+  Read the [Seeding section](#seeding-the-database) for more details.
+
+- `prisma: [PrismaOptions]`
```
infomiho, line 268:
```suggestion
- `prisma: PrismaOptions`
```

web/docs/database/backends.md
sodic: Crap, forgot to change this chapter.

web/docs/database/crud.md
sodic: It is, line 659 :smile:

web/docs/database/operations/overview.md
```diff
@@ -0,0 +1,12 @@
...
+---
+
+import { Required } from '@site/src/components/Required';
+
+While Entities enable help you define your app's data model and relationships, Operations are all about working with this data.
+
+There are two kinds of Operations: [**Queries**](/docs/database/operations/queries.md) and [**Actions**](/docs/database/operations/actions.md). As their names suggest,
+Queries are meant for reading data, and Actions are meant for changing it (either by updating existing entries or creating new ones).
+
+Keep reading to find out all there is two know about Operations in Wasp.
```
infomiho, line 12:
```suggestion
Keep reading to find out all there is to know about Operations in Wasp.
```

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+Therefore, if you're already familiar with Actions, you might find reading the entire guide repetitive.
+
+We instead recommend skipping ahead and only reading [the differences between Queries and Actions](/docs/database/operations/actions#differences-between-queries-and-actions), and consulting the [Options Reference](#options-reference) as needed.
+:::
+
+## Working with Queries
+
+Queries are declared in Wasp and implemented in NodeJS. Wasp executes Queries within the server's context but also generates the code that lets you call them from anywhere in your code (client or server) via the same interface.
+In other words, you don't have to worry about building an HTTP API for the Query, handling the request on the server, or even handling and caching the responses on the client.
+Instead, simply focus on the business logic inside your Query and let Wasp take care of the rest!
```
infomiho, line 23:
```suggestion
You declare queries in the `.wasp` file and implement them using NodeJS. Wasp not only runs these queries within the server's context but also creates code that enables you to call them from any part of your codebase, whether it's on the client or server side.

This means you don't have to build an HTTP API for your query, manage server-side request handling, or even deal with client-side response handling and caching.

So, just concentrate on crafting the business logic inside your query, and let Wasp handle the rest!
```

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+
+1. Declare the Query in Wasp using the `query` declaration.
+2. Define the Query's NodeJS implementation.
+
+After completing these two steps, you'll be able to use the Query from any point in your code.
+
+### Declaring Queries
+
+We'll start by telling Wasp we wish to create a Query using a `query` declaration.
+Let's declare two Queries, one for fetching all tasks and another one for fetching tasks based on a filter (e.g., whether a task is done):
```
infomiho, line 35:
```suggestion
To create a Query in Wasp, we begin with a `query` declaration. Let's declare two Queries - one to fetch all tasks, and another to fetch tasks based on a filter, such as whether a task is done:
```

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+</TabItem>
+</Tabs>
+
+<small>
+
+If you want to know about all supported options for the `query` declaration, take a look at the [Options reference](#options-reference).
+
+</small>
+
+Wasp Queries and their implementations don't need to (but can) have the same name. We'll keep the names identical to avoid confusion.
```
infomiho, line 82:
```suggestion
The names of Wasp Queries and their implementations don't need to match, but we'll keep them the same to avoid confusion.
```

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+
+:::info
+You might have noticed that we told Wasp to import Query implementations that don't yet exist. Don't worry about that for now. We'll write the implementations imported from `queries.{js,ts}` in the next section.
+
+It's a good idea to start with the high-level concept (i.e., the Query declaration in the Wasp file) and only then deal with the implementation details (i.e., the Query's implementation in JavaScript).
+:::
+
+After declaring a Wasp Query, two crucial things happen:
```
infomiho, line 90:
```suggestion
After declaring a Wasp Query, two important things happen:
```

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+
+- Wasp **generates a server-side NodeJS function** that shares its name with the Query. This function's interface is identical to the client-side function from the previous point.
+- Wasp **generates a client-side JavaScript function** that shares its name with the Query (e.g., `getFilteredTasks`).
+  This function takes a single optional argument - an object containing any serializable data you wish to use inside the Query.
+  Wasp will send this object over the network and pass it into the Query's implementation as its first positional argument (more on this when we look at the implementations).
+  Such an abstraction works thanks to an HTTP API route handler Wasp generates on the server, which calls the Query's NodeJS implementation under the hood.
```
infomiho, line 96:
```suggestion
- Wasp **generates a server-side NodeJS function** that shares its name with the Query.
  
  This function's interface is identical to the client-side function from the previous point.

- Wasp **generates a client-side JavaScript function** that shares its name with the Query (e.g., `getFilteredTasks`).

    This function takes a single optional argument - an object containing any serializable data you wish to use inside the Query.
  Wasp will send this object over the network and pass it into the Query's implementation as its first positional argument (more on this when we look at the implementations).
  Such an abstraction works thanks to an HTTP API route handler Wasp generates on the server, which calls the Query's NodeJS implementation under the hood.
```

This section feels very heavy with details 🤯

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+
+Annotating the Queries is optional, but highly recommended. Doing so enables **full-stack type safety**. We'll see what this means when calling the Query from the client.
+
+:::tip
+Wasp uses [superjson](https://github.com/blitz-js/superjson) under the hood. In other words, you don't need to limit yourself to only sending and receiving JSON payloads.
+
+Send and receive any superjson-compatible payload (e.g., Dates, Sets, Lists, circular references, etc.) and let Wasp take care of the (de)serialization.
+
+As long as you're annotating your Queries with correct automatically generated types, TypeScript ensures your payloads are valid (i.e., that Wasp knows how to serialize and deserialize them).
+:::
```
infomiho, line 189:
```suggestion
Wasp automatically generates the types `GetTasks` and `GetFilteredTasks` based on your Wasp file's declarations:

- `GetTasks` is a generic type automatically generated by Wasp, based on the Query declaration for `getTasks`.
- `GetFilteredTasks` is also a generic type automatically generated by Wasp, based on the Query declaration for `getFilteredTasks`.

You can utilize these types to define the input and output types for your Query.

For instance, the Query `getTasks` doesn't expect any arguments (its input type is `void`), but it does return a list of tasks (its output type is `Task[]`).

On the other hand, the Query `getFilteredTasks` expects an object of type `{ isDone: boolean }`. This type is derived from the `Task` type.

While annotating the Queries is optional, it's highly recommended. Doing so enables **full-stack type safety**. We'll explore what this means when we discuss calling the Query from the client.

:::tip
Wasp uses [superjson](https://github.com/blitz-js/superjson) under the hood. This means you're not limited to only sending and receiving JSON payloads.

You can send and receive any superjson-compatible payload (like Dates, Sets, Lists, circular references, etc.) and let Wasp handle the (de)serialization.

As long as you're annotating your Queries with the correct automatically generated types, TypeScript ensures your payloads are valid (i.e., Wasp knows how to serialize and deserialize them).
:::
```

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+</small>
+
+### Using Queries
+
+To use a Query, you can import it from `@wasp` and call it directly. As mentioned, the usage doesn't change depending on whether you're on the server or the client:
+
+<Tabs groupId="js-ts">
+<TabItem value="js" label="JavaScript">
+
+```javascript
```
infomiho, line 207: I would add `title=client/SomePage.jsx` so it's clear that it's client code

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+// the payloads.
+const allTasks = await getAllTasks();
+const doneTasks = await getFilteredTasks({ isDone: true });
+```
+
+</TabItem>
+</Tabs>
+
+:::caution
+Wasp will not stop you from importing a Query's NodeJS implementation from `./queries.{js,ts}` and calling it directly.
```
infomiho, line 234: Is this even possible? `queries.ts` is server-side code?

infomiho: I would maybe delete this note completely, as it comments on some edge case.

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+
+However, it's best if you avoid doing so. Besides losing all useful features a Wasp Query provides (e.g., Entity injection), the raw Query will expect you to provide a context as a second argument (which is cumbersome).
+:::
+
+#### The `useQuery` hook
+
+When using Queries on the client, you can make them reactive with the `useQuery` hook.
+This hook comes bundled with Wasp and is a thin wrapper around the `useQuery` hook from [_react-query_](https://github.com/tannerlinsley/react-query), the only difference being in not having to supply the key (Wasp does this automatically under the hood).
```
infomiho, line 242:
```suggestion
This hook comes bundled with Wasp and is a thin wrapper around the `useQuery` hook from [_react-query_](https://github.com/tannerlinsley/react-query). The only difference is that you don't need to supply the key - Wasp handles this for you automatically.
```

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+  );
+};
+
+export default MainPage;
+```
+
+</TabItem>
+<TabItem value="ts" label="TypeScript">
+
+```tsx
```
infomiho, line 302: `title=filename`

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+This hook comes bundled with Wasp and is a thin wrapper around the `useQuery` hook from [_react-query_](https://github.com/tannerlinsley/react-query), the only difference being in not having to supply the key (Wasp does this automatically under the hood).
+
+Here's an example of calling the Queries using the `useQuery` hook:
+
+<Tabs groupId="js-ts">
+<TabItem value="js" label="JavaScript">
+
+```jsx
+import React from "react";
+import { Task } from "@wasp/entities";
```
infomiho, line 251: Importing `Task` in Javascript code

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+
+<Tabs groupId="js-ts">
+<TabItem value="js" label="JavaScript">
+
+```js title=src/server/queries.js
+import HttpError from "@wasp/core/HttpError.js";
+
+export const getAllTasks = async (args, context) => {
+  throw new HttpError(
+    403, // status coce
```
infomiho, line 379:
```suggestion
    403, // status code
```

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+    { foo: "bar" } // data
+  );
+};
+```
+
+</TabItem>
+</Tabs>
+
+If the status code is `4xx`, the client will receive a response object with the corresponding `.message` and `.data` fields and rethrow the error (with these fields included).
+To prevent information leakage, the server won't forward these fields for any other HTTP status codes.
```
infomiho, line 406:
```suggestion
If the status code is 4xx, the client will receive a response object with the corresponding .message and .data fields, and it will rethrow the error (including these fields). To prevent the leakage of information, the server won't forward these fields for any other HTTP status codes.
```

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+Declaring the Query:
+
+```wasp
+query getFoo {
+    fn: import { getFoo } from "@server/queries.js"
+    entities: [Foo]
+}
+```
+
+Makes the following import on both the server and the client:
```
infomiho, line 522: Reads weird, "Makes the following import", does it import it somewhere or what?

We mention that it is available on the server, but don't really give them an example of importing a query on the server (I don't think we should give it) so maybe no mention server here?

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+
+```js
```
infomiho, line 524: I would wrap all code blocks with tabs even though the content is the same.

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+</Tabs>
+
+### Implementing Queries
+
+The Query's implementation is a NodeJS function that takes two arguments (it can be an `async` function if you need to use the `await` keyword).
+Since both arguments are positional, you can name the parameters however you want, but we'll stick with `args` and `context`:
+
+1. `args` (type depends on the Query)
+
+   An object containing all the arguments (i.e., payload) **passed to the Query by the caller** (e.g., filtering conditions).
```
infomiho, line 562: "the data that is passed in when calling the query"

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+   Check [the usage examples](#using-the-query) to see how to pass this object to the Query.
+
+2. `context` (type depends on the Query)
+
+   An additional context object **passed into the Query by Wasp**. This object contains user session information, as well as information about entities. Check the [section about using entities in Queries](#using-entities-in-queries) to see how to use the `context` object.
```
infomiho, line 567: Maybe link to Using auth for to see `context.user` usage

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+
+2.  `Output`
+
+    The type of the Query's return value (i.e., the Query's output payload). The default value is `unknown`.
+
+The defaults were chosen to make the type signature as permissive as possible. If don't want your Query to take/return anything, use `void` as a type argument.
+
+</ShowForTs>
+
+#### Example
```
infomiho, line 592: This example shows usage of the generic type, but it offers a Javascript impl where that doesn't make sense (there are no types used).

I would suggest wrapping the whole section, starting with "Afer you [declare the query](#declaring-queries-in-wasp), Wasp generates a generic type you can use when defining its implementation." in `<ShowForTs>` and only have the example for TS.

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+  // implementation
+};
+```
+
+In this case, the Query expects to receive an object with an `id` field of type `number` (this is the type of `args`), and return a value of type `Foo` (this must match the type of the Query's return value).
+
+</TabItem>
+</Tabs>
+
+### The `useQuery` Hook
```
infomiho, line 645: Link to an example above or offer it inline

web/docs/database/operations/queries.md
```diff
@@ -0,0 +1,666 @@
...
+---
+
+import { Required } from '@site/src/components/Required';
+
+We'll explain what Queries are and how to use them. If you're looking for a detailed API specification, skip ahead to the [Options Reference](#options-reference).
+
+You can use Queries to fetch data from the server. They shouldn't modify the server's state.
+Fetching all comments on a blog post, a list of users that liked a video, information about a single product based on its ID... All of these are perfect use cases for a Query.
+
+:::tip
```
infomiho, line 12: I would maybe remove this note 🤷  let them read, they skim over the similar bits?

web/docs/database/operations/actions.md
```diff
@@ -0,0 +1,889 @@
...
+title: Actions
+---
+
+import { Required } from '@site/src/components/Required';
+
+We'll explain what Actions are and how to use them. If you're looking for a detailed API specification, skip ahead to the [Options Reference](#options-reference).
+
+Actions are very similar to [Queries](/docs/database/operations/queries.md). The key difference is that Actions are meant to modify and add data, while Queries are only supposed to read it. Adding a comment to a blog post, liking a video, updating a product's price... These are some common examples of Actions.
+
+Actions and Queries work hand-in-hand on keeping data caches fresh.
```
infomiho, line 11:
```suggestion
Actions are quite similar to [Queries](/docs/database/operations/queries.md), but with a key distinction: Actions are designed to modify and add data, while Queries are solely for reading data. Examples of Actions include adding a comment to a blog post, liking a video, or updating a product's price.

Actions and Queries work together to keep data caches up-to-date.
```

web/docs/database/operations/actions.md
```diff
@@ -0,0 +1,889 @@
...
+Actions are declared in Wasp and implemented in NodeJS. Wasp executes Actions within the server's context but also generates the code that lets you call them from anywhere in your code (client or server) via the same interface.
+In other words, you don't have to worry about building an HTTP API for the Action, handling the request on the server, or even handling and caching the responses on the client.
+Instead, simply focus on the business logic inside your Action and let Wasp take care of the rest!
+
+To create an Action, you must:
+
+1. Declare the Action in Wasp using the `action` declaration.
+2. Define the Action's NodeJS implementation.
+
+After completing these two steps, you'll be able to use the Action from any point in your code.
```
infomiho, line 31:
```suggestion
## Working with Actions

Actions are declared in Wasp and implemented in NodeJS. Wasp runs Actions within the server's context, but it also generates code that allows you to call them from anywhere in your code (either client or server) using the same interface. This means you don't have to worry about building an HTTP API for the Action, managing server-side request handling, or even dealing with client-side response handling and caching.

Instead, just focus on developing the business logic inside your Action, and let Wasp handle the rest!

To create an Action, you need to:

1. Declare the Action in Wasp using the `action` declaration.
2. Implement the Action's NodeJS functionality.

Once these two steps are completed, you can use the Action from anywhere in your code.
```

web/docs/database/operations/actions.md
```diff
@@ -0,0 +1,889 @@
...
+
+### Declaring Actions
+
+We'll start by telling Wasp we wish to create an Action using an `action` declaration.
+Let's declare two Actions, one for creating a task and another one for marking tasks as done:
```
infomiho, line 36:
```suggestion
### Declaring Actions

To create an Action in Wasp, we begin with an `action` declaration. Let's declare two Actions - one for creating a task, and another for marking tasks as done:
```

web/docs/database/operations/actions.md
```diff
@@ -0,0 +1,889 @@
...
+</TabItem>
+</Tabs>
+
+<small>
+
+If you want to know about all supported options for the `action` declaration, take a look at the [Options reference](#options-reference).
+
+</small>
+
+Wasp Actions and their implementations don't need to (but can) have the same name. We'll keep the names identical to avoid confusion.
```
infomiho, line 84:
```suggestion
The names of Wasp Actions and their implementations don't necessarily have to match. However, to avoid confusion, we'll keep them the same.
```

web/docs/database/operations/actions.md
```diff
@@ -0,0 +1,889 @@
...
+
+After declaring a Wasp Action, two crucial things happen:
+
+- Wasp **generates a server-side NodeJS function** that shares its name with the Action. This function's interface is identical to the client-side function from the previous point.
+- Wasp **generates a client-side JavaScript function** that shares its name with the Action (e.g., `createTask`).
+  This function takes a single optional argument - an object containing any serializable data you wish to use inside the Action.
+  Wasp will send this object over the network and pass it into the Action's implementation as its first positional argument (more on this when we look at the implementations).
+  Such an abstraction works thanks to an HTTP API route handler Wasp generates on the server, which calls the Action's NodeJS implementation under the hood.
+
+Generating these two functions ensures a uniform calling interface across the entire app (both client and server).
```
infomiho, line 100:
```suggestion
After declaring a Wasp Action, two key things occur:

- Wasp **creates a server-side NodeJS function** that has the same name as the Action.
  This function's interface mirrors that of the client-side function from the previous point.
- Wasp **produces a client-side JavaScript function** that also shares the Action's name (e.g., `createTask`).

  This function accepts a single optional argument - an object containing any serializable data you want to use inside the Action. Wasp will transmit this object over the network and pass it into the Action's implementation as its first positional argument (we'll discuss this more when we look at the implementations). This abstraction is facilitated by an HTTP API route handler that Wasp generates on the server, which calls the Action's NodeJS implementation behind the scenes.

These two generated functions ensure a consistent calling interface across your entire app, whether it's on the client or server side.
```

web/docs/database/operations/actions.md
```diff
@@ -0,0 +1,889 @@
...
+
+### Implementing Actions in Node
+
+Now that we've declared the Action, all that's left is implementing it.
+We've told Wasp to look for Actions' implementations in the file `src/server/actions.{js,ts}`, so that's where we should export them from.
+
+Here's how you might implement the previously declared Actions `createTask` and `markTaskAsDone`:
```
infomiho, line 107:
```suggestion
Now that we've declared the Action, what remains is to implement it. We've instructed Wasp to look for the Actions' implementations in the file `src/server/actions.{js,ts}`, so that's where we should export them from.

Here's how you might implement the previously declared Actions `createTask` and `markTaskAsDone`:
```

web/docs/database/operations/actions.md
```diff
@@ -0,0 +1,889 @@
...
+  }
+  task.isDone = true;
+};
+```
+
+:::tip
+Wasp uses [superjson](https://github.com/blitz-js/superjson) under the hood. In other words, you don't need to limit yourself to only sending and receiving JSON payloads.
+
+Send and receive any superjson-compatible payload (e.g., Dates, Sets, Lists, circular references, etc.) and let Wasp take care of the (de)serialization.
+:::
```
infomiho, line 148: This could be a "partial template" aka `_superjsonNote.md`

web/docs/database/operations/actions.md
```diff
@@ -0,0 +1,889 @@
...
+  );
+};
+```
+
+</TabItem>
+</Tabs>
+
+Since Actions don't require reactivity, they are safe to use inside components without a hook. Still, Wasp provides comes with the `useAction` hook you can use to enhance actions. Read all about it in the [Options Reference](#options-reference).
+
+### Error Handling
```
infomiho, line 345: I would maybe put this section, using entities and Prisma Error Helpers into the `Overview` section (since they are the same for actions and queries). 

I did something similar with the Social Auth docs 🤷

web/docs/database/operations/actions.md
```diff
@@ -0,0 +1,889 @@
...
+</Tabs>
+
+## Cache Invalidation
+
+One of the trickiest parts of managing a web app's state is making sure the data returned by the Queries is up to date.
+Since Wasp uses _react-query_ for Query management, we must make sure to invalidate Queries (more specifically, their cached results managed by _react-query_) whenever they become stale.
+
+It's possible to invalidate the caches manually through several mechanisms _react-query_ provides (e.g., refetch, direct invalidation).
+However, since manual cache invalidation quickly becomes complex and error-prone, Wasp offers a quicker and a more effective solution to get you started: **automatic Entity-based Query cache invalidation**.
+Because Actions can (and most often do) modify the state while Queries read it, Wasp invalidates a Query's cache whenever an Action that uses the same Entity is executed.
```
infomiho, line 547:
```suggestion
To simplify this process, Wasp offers a faster and more effective solution: **automatic Entity-based Query cache invalidation**. Since Actions often modify the state while Queries read it, Wasp automatically invalidates a Query's cache whenever an Action that uses the same Entity is executed.
```

web/docs/database/operations/actions.md
```diff
@@ -0,0 +1,889 @@
...
+
+## Differences Between Queries and Actions
+
+Actions and Queries are two very similar concepts in Wasp. They even might appear to be doing the same thing, but Wasp treats them differently and using each construct carries certain semantic implications.
+
+Here's a rundown of all the differences between Queries and Actions:
+
+1. Actions can (and most often should) modify the server's state, while Queries are only allowed to read it. Wasp relies on you following this convention when performing cache invalidations, so it's important to stick to it.
+2. Actions don't need to be reactive so you can call them directly. Still, Wasp does provide a [`useAction` React hook](#the-useaction-hook) for adding extra behavior to the Action (e.g., optimistic updates).
+3. `action` declarations in Wasp are mostly identical to `query` declarations. The only difference is in the declaration's name.
```
infomiho, line 566:
```suggestion
Actions and Queries are two closely related concepts in Wasp. They might seem to perform similar tasks, but Wasp treats them differently, and each concept represents a different thing.

Here are the key differences between Queries and Actions:

1. Actions can (and often should) modify the server's state, while Queries are only permitted to read it. Wasp relies on you adhering to this convention when performing cache invalidations, so it's crucial to follow it.
2. Actions don't need to be reactive, so you can call them directly. However, Wasp does provide a [`useAction` React hook](#the-useaction-hook) for adding extra behavior to the Action (like optimistic updates).
3. `action` declarations in Wasp are mostly identical to `query` declarations. The only difference lies in the declaration's name.
```

web/docs/database/operations/queries.md
sodic: That's exactly what I did, but forgot to import it :upside_down_face:

web/docs/database/operations/queries.md
sodic: Talked about it, decided to leave it for now.

web/docs/database/operations/actions.md
sodic: Discussed it, postponed it.

sodic:
Covered all the comments (applies ones for actions.md to queries.md and vice versa). Merging to draft branch.

PR #1379, opened 2023-07-31
Docs rst env

Martinsos:
I rewrote docs for .env vars to be more comprehensive -> they now provide more info in a more structured way.
There is one section missing yet though: section about setting env vars in the production. I left a TODO with how I think that section should be written. Once that section is finalized, that should be it.


infomiho:
I've followed up on the TODO

PR #1375, opened 2023-07-27
Updates LSP templates

infomiho:
### Description

Updates the scaffold templates:
- `query` (gives the generic params a name)

  ```ts
   
  type AddUserPayload = void
  type AddUserResult = void
  export const newStuff: AddUser<AddUserPayload, AddUserResult> = async (args, context) => {
    // Implementation goes here
  }
  ```
- `action` (gives the generic params a name)
- the folder rename from `lsp` to `Lsp` to keep it consistent with `Generator` and `Cli`

### Select what type of change this PR introduces:

1. [x] **Just code/docs improvement** (no functional change). 
2. [ ] **Bug fix** (non-breaking change which fixes an issue).
3. [ ] **New feature** (non-breaking change which adds functionality).
4. [ ] **Breaking change** (fix or feature that would cause existing functionality to not work as expected).

### Update Waspc ChangeLog and version if needed
If you did a bug fix, new feature, or breaking change, that affects waspc, make sure you satisfy the following:
1. [ ] I updated [ChangeLog.md](https://github.com/wasp-lang/wasp/blob/main/waspc/ChangeLog.md) with description of the change this PR introduces.
2. [ ] I bumped waspc version in [waspc.cabal](https://github.com/wasp-lang/wasp/blob/main/waspc/waspc.cabal) to reflect changes I introduced, with regards to the version of the latest wasp release, if the bump was needed.

PR #1381, opened 2023-07-31
[waspls] Makes LSP scaffold create missing parent directories

craigmc08:
### Description

Currently, when a scaffold code action creates a new file, waspls will crash if the parent directory of the new file does not exist. This PR fixes that crash by changing the implementation of scaffolding to create the directory (and any parent directories) if missing.

This PR also moves the scaffold templates to `data/Lsp`. #1375 changed the code to look in this directory but the fact that the templates were still in `data/lsp` was missed during code review 😅


sodic (approved):

sodic (changes requested):
I noticed several problems with the feature (not necessarily related to this PR, you might know about them already):
- The go-to-definition functionality seems to be broken now.
- The extension only detects non-existent function definitions for non-existing files. If the target file exists but doesn't export anything, the extension doesn't complain.

A few more suggestions I have (on the entire feature, not exclusively this PR):
- Adding an empty line between the type definitions and the operation's implementation.
- Renaming `Payload` to `Input` and `Result` to `Output`. Alternatively, renaming `Payload` to `Args` and keeping `Result`.

craigmc08:
> I noticed several problems with the feature

I can't replicate these issues, go-to definition and detecting missing exports from existing files are working for me. Can you describe the project its not working in and what you are doing when it's not working @sodic ?

> A few more suggestions I have

I updated the templates!

infomiho:
I've tested this locally as well, it works as expected:
- missing file & fn ✅ 
- missing fn in existing file ✅
- actions, queries and pages ✅
- missing nested file ✅

@sodic please test it out one more time locally to be extra safe

infomiho (approved):

sodic:
I'll check it out, might just be a linux thing. I'll also run it on my mac to see what's up. @craigmc08 I'll let you know in the evening (Croatian time).

sodic:
Ok, my bad.

Turns out I ran `./tools/install_packages_to_data_dir.sh` after `cabal install` and ended up with an old version of `ts-inspect`. In short, you can ignore the stuff from before. It was a local setup issue on my side.

I'll just list the things that still don't work for me.

## Spacing
The spacing for multiple exports from the same file works well in all usual cases, but seems to be broken when I combine a default export with a named export.

Scaffolding these two Queries shows the spacing issue:
  ```c
  query getMatija {
    fn: import getMatija from "@server/matija/operations.js",
    entities: []
  }

  query getMatija2 {
    fn: import { getMatija2 } from "@server/matija/operations.js",
    entities: []
  }
  ``` 
It results in:
```typescript
// ...
export default getMatija
import { GetMatija2 } from '@wasp/queries/types'
// ...
```

## LSP Errors when dealing with subdirectories
I'll list two different (but closely related) errors I've experienced. These issues only happen when importing from files in subdirectories. They don't happen for imports from source files directly under `@server`.

### Go-to-definition error
If I create a Query imported from a subdirectory of `server`, it successfully scaffolds the file, but the go-to-definition functionality doesn't work until i restart the language server.

Does this happen for anyone else?

Steps to reproduce:
1. Run `wasp new someProject` and open it in VS Code.
3. Add this Query to the wasp file: 
   ```c
   query getMatija {
         fn: import { getMatija } from "@server/people/operations.js",
         entities: []
   }
   ```
4. Use a quick action to scaffold the Query's implementation.
5. Try to go to the Query's definition.

### Scaffolding error
Importing a second (unimplemented) Query from the same file in a subdirectory of `server` doesn't seem to trigger an LSP error (red squiggly  line) or a quick action prompt.

Steps to reproduce: Do everything you did above and import one more Query from the same file.
 ```c
 query getMartin {
       fn: import { getMartin } from "@server/people/operations.js",
       entities: []
 }
 ```

craigmc08:
@sodic I could replicate all of those issues and fixed them:

The missing import diagnostics rely on VSCode telling waspls when files changed. For some reason, VSCode doesn't send that notification when the scaffold action creates new files. This is where those two bugs came from. Fixed them by having the scaffold action update the diagnostics after it creates the file.

> The spacing for multiple exports from the same file works well in all usual cases, but seems to be broken when I combine a default export with a named export.

Changed how the scaffolding action deals with newlines so that it tries to add a blank line when the file already exists. (it was due to luck that there was proper spacing between named exports)

waspc/waspls/src/Wasp/LSP/Commands/ScaffoldTsSymbol.hs
```diff
@@ -149,7 +157,29 @@ handler request respond = withParsedArgs request respond scaffold
...
         -- NOTE: we modify the file on disk instead of applying an edit through
         -- the LSP client. See "Current Limitations" above.
-        liftIO $ Text.appendFile (SP.fromAbsFile filepath) rendered
+
+        -- Create directory (and parent directories) if it doesn't exist.
+        liftIO $ Dir.createDirectoryIfMissing True $ SP.fromAbsDir $ SP.parent filepath
+
+        -- If the file already exists, we want to add a blank line between the
+        -- file contents and the new text we're writing.
+        needsLeadingNewline <- liftIO $ Dir.doesFileExist (SP.fromAbsFile filepath)
```
sodic, line 166: This is minor, but I don't think this is the right condition? The file could exist and be empty. In that case, we'll start the file with an empty line.

sodic (approved):
All the issues I've noticed have been solved, left one minor comment.

Nice work with the entire feature!

PR #1385, opened 2023-08-07
Add install command to the `run` script

sodic:
A command that makes sure we don't forget installing JavaScript packages into the data directory before calling `cabal install`.


infomiho (approved):
LGTM

PR #1393, opened 2023-08-10
[waspls] Use the wasp project root instead of the workspace root

craigmc08:
Previously, waspls used the workspace root to look for external code files, which was incorrect if you opened a workspace larger than the wasp project. This could manifest as false-positive missing import diagnostics (which also breaks a lot of other extimport-related features):

image

*Above: Example of the incorrect behavior that is now fixed*

Now, waspls uses the actual wasp project root instead of the workspace root. It finds the project root based on the location of the `.wasp` file.

This PR also updates the changelog with a description of the code scaffolding feature.

> [!NOTE]
> This does **not** add support for multi-project workspaces, similar to the following:
> ```
> .
> ├── project1/
> │   ├── src/
> │   ├── .wasproot
> │   └── main.wasp
> └── project2/
>     ├── src/
>     ├── .wasproot
>     └── main.wasp
> ```
> waspls will probably behave strangely in this scenario.

### DIFF
--- a/waspc/ChangeLog.md
+++ b/waspc/ChangeLog.md
@@ -4,6 +4,33 @@
 
 ### 🐞 Bug fixes / 🔧 small improvements
 - Wasp copied over the `.env.server` instead of `.env.client` to the client app `.env` file. This prevented using the `.env.client` file in the client app.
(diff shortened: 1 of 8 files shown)


infomiho (approved):
LGTM, tested locally

PR #1395, opened 2023-08-18
Signup Customisation [RFC]

infomiho:
## Motivation

We are looking to enable users to add additional fields to signup process when they are using Auth UI + Wasp Auth.

|Screenshot 2023-08-18 at 13 37 50|
|:-:|
|Our regular Signup form|

For example, if we have a User model that looks like this:
```prisma
entity User {=psl
  username String @unique
  password String
  address String?
psl=}
```
and want to ask user for their `address` in the Signup process, how could it look like?

## Parts of the problem
There are three pieces of the problem:
1. UI components
2. Validating extra fields on backend
3. Including the fields when creating a new User

### UI components

If the user is using Auth UI then the Auth UI components for Signup need to be extensible.

### Backend validation

Frontend validation is not enough, it needs to also happen on the backend. So, when the requests comes, the fields need to be checked e.g. `address` is present and has at least 5 characters.

### Save the fileds

When we are creating a new user, we are explicit about which fields we include (for security reasons) from the body of the request. Therefore, any extra fields, need to be specified explicitly as well. 

## Proposed solution

In this PR, we can see the above mentioned parts solved with a new field `auth.signupCustomizatiion` and an extension point in the `<Signup />` component. 

Let's see how it can be done for the `address` field from the example above.

### **main.wasp**

We defined a new dict called `signupCustomization` with one field called `additionalFields: ExtImport`

```diff title="main.wasp" {5}
app crudTesting {
  // ...
  auth: {
    userEntity: User,
    methods: {
      usernameAndPassword: {},
    },
    onAuthFailedRedirectTo: "/login",
+   signup: {
+     additionalFields: import { fields } from "@server/auth.js",
+   },
  },
}
```

### **auth.ts**

- We export an object with additional field names as keys and their definition as the value.
- The definition has two fields: `get` and `validate`, as shown below

```ts
import { defineAdditionalSignupFields } from '@wasp/auth/index.js'

export const fields = defineAdditionalSignupFields({
  // This object is type-safe and complains about defining keys that are not part
  // of the User model.

  // We also exclude the "username" and "password" fields since
  // they are already set by Wasp in the signup process

  address: (data) => {
    if (typeof data.address !== 'string') {
      throw new Error('Address is required.')
    }
    if (data.address.length < 10) {
      throw new Error('Address must be at least 10 characters long.')
    }
    return data.address
  },
})
```

### **SignupPage.tsx**

Users can extend the UI in two ways:
- providing an object describing the field (easier, but limited)
- providing a render fn that render any UI (harder, but fully flexibile)

Users extend the UI via the `additionalFields` prop. They can pass in:
- an array of objects
- an array of both objects and render fns
- just a single render fn (instead of the array)

**Declaring an array of only objects**

```tsx
import { SignupForm } from "@wasp/auth/forms/Signup";

export const SignupPage = () => {
  return (
    <div className="container">
      <main>
        <h1>Signup</h1>
        <SignupForm
          additionalFields={[
            {
              name: "address",
              label: "Address",
              type: "input",
              validations: {
                required: "Address is required",
              },
            },
          ]}
        />
      </main>
    </div>
  );
};
```

Each entry can look like this:
```ts
{
  name: string;
  label: string;
  type: "input" | "textarea";
  validations?: RegisterOptions<LoginSignupFormFields> // `react-hook-form` specific options for the `register` function
}
```

**Adding render functions**

They can add render functions inside of the `additionalFields` array. They are provided with the `react-hook-form` functions that help them to pass data from their UI to the server.

```tsx
import { SignupForm } from "@wasp/auth/forms/Signup";

import {
  FormError,
  FormInput,
  FormItemGroup,
  FormLabel,
} from "@wasp/auth/forms/internal/Form";


export const SignupPage = () => {
  return (
    <div className="container">
      <main>
        <h1>Signup</h1>
        <SignupForm
          additionalFields={[
            {
              name: "address",
              label: "Address",
              type: "input",
              validations: {
                required: "Address is required",
              },
             },
             ({ register, formState: { errors } },
                  { isLoading }) => (
                  <>
                    <FormItemGroup>
                      <FormLabel>Phone number</FormLabel>
                      <FormInput
                        {...register("phone", {
                          required: "Phone number is required",
                        })}
                        type="text"
                        disabled={isLoading}
                      />
                      {errors.phone && <FormError>{errors.phone.message}</FormError>}
                    </FormItemGroup>
                  </>
                )
          ]}
        />
      </main>
    </div>
  );
};
```

They can render **whatever** they need to, not just fields:

```tsx
import { SignupForm } from "@wasp/auth/forms/Signup";

export const SignupPage = () => {
  return (
    <div className="container">
      <main>
        <h1>Signup</h1>
        <SignupForm
          additionalFields={[
            {
              name: "address",
              label: "Address",
              type: "input",
              validations: {
                required: "Address is required",
              },
             },
             () => <div>Extra info below</div>,
          ]}
        />
      </main>
    </div>
  );
};
```

**Just using the render fn**

Or simply, they can provide a single render fn that renders any UI they need.

```tsx
import { SignupForm } from "@wasp/auth/forms/Signup";

import {
  FormItemGroup,
} from "@wasp/auth/forms/internal/Form";

export const SignupPage = () => {
  return (
    <div className="container">
      <main>
        <h1>Signup</h1>
        <SignupForm
          additionalFields={(
              { register, formState: { errors } },
              { isLoading }
            ) => (
              <>
                <FormItemGroup>
                  What ever UI they need
                </FormItemGroup>
              </>
            )}
        />
      </main>
    </div>
  );
};
```

| Screenshot 2023-08-18 at 13 24 44 |
|:-:|
|Frontend validation|

|Screenshot 2023-08-18 at 13 25 22|
|:-:|
|Backend validation|

|Screenshot 2023-08-18 at 13 25 47|
|:-:|
|Saved in the DB|

## Left to do
- [x] Agree on the API
- [x] Docs
- [x] Changelog


waspc/data/Generator/templates/react-app/src/auth/forms/internal/common/LoginSignupForm.tsx
```diff
@@ -97,40 +97,55 @@ const googleSignInUrl = `${config.apiUrl}{= googleSignInPath =}`
...
-  const cta = state === 'login' ? 'Log in' : 'Sign up';
+  const isLogin = state === 'login'
+  const cta = isLogin ? 'Log in' : 'Sign up';
   {=# isAnyPasswordBasedAuthEnabled =}
   const history = useHistory();
   const onErrorHandler = (error) => {
     setErrorMessage({ title: error.message, description: error.data?.data?.message })
   };
   {=/ isAnyPasswordBasedAuthEnabled =}
+  const hookForm = useForm<LoginSignupFormFields>()
```
infomiho, line 132: We have a new depedency called `react-hook-form` and it's an industry standard. It enables easier form values management, client side validation and it makes it easier to add additional fields.

waspc/data/Generator/templates/react-app/src/auth/forms/internal/common/LoginSignupForm.tsx
```diff
@@ -184,41 +198,45 @@ export const LoginSignupForm = ({
...
+              {...register('username', {
+                required: 'Username is required',
+              })}
               type="text"
-              required
-              value={usernameFieldVal}
-              onChange={e => setUsernameFieldVal(e.target.value)}
               disabled={isLoading}
             />
+            {errors.username && <FormError>{errors.username.message}</FormError>}
```
infomiho, line 212: As a side effect of including `react-hook-form` and client side validation becoming trivial, I've included it for existing fields as well.

waspc/data/Generator/templates/react-app/src/auth/forms/internal/common/LoginSignupForm.tsx
```diff
@@ -184,41 +198,45 @@ export const LoginSignupForm = ({
...
+              })}
               type="password"
-              required
-              value={passwordFieldVal}
-              onChange={e => setPasswordFieldVal(e.target.value)}
               disabled={isLoading}
             />
+            {errors.password && <FormError>{errors.password.message}</FormError>}
           </FormItemGroup>
+          {additionalSignupFields && additionalSignupFields(hookForm)}
```
infomiho, line 239: Render any additional fields.

Martinsos:
Nice job!

I see there is also code here, so this is not just RFC @infomiho , this is also a PR with the actual change in it?

`signupCustomization` -> I don't think you need `Customization`, just `signup` is enough. The whole .wasp file is configuration/customization.

`defineAdditionalSignupFields } from "@wasp/types/index.js"` -> this is not a type though, so should we import it from some other places? Maybe smth like `"@wasp/auth"`?

`get` receives the whole `data` object -> is that the whole user? And then we need to get `address` from it? Why is that so, why doesn't `get` already receive `data.address`? Also, can `get` have a default implementation, if they don't provide it?

Advanced idea: we could offer some default validators for stuff like address, age and similar.

Ok, so `auth.signup.additionalFields` -> this is used to do stuff on backend, like validation and creating user, right? But on frontend, they have to define their own UI, in `SignupForm` component? Looking at it, it is not super simple -> you need to know how to use react-hook-form. That is ok on one side, we give them quite some flexibility, but I also wonder if we could generate something by default for them, since we already know the fields, and we know how to validate them and all? Also, I am ok with `react-hook-form` in general, as you said it is quite popular, so that sounds good, I am just trying to avoid them writing stuff that we can easily write.

On images you provided, "address is required" error message is printed below the address field, but the on the other one, it is printed above -> first one is client validation, second one is server validation? As a user, can I avoid having to implement my own error showing on the client, could that be done automatically?

infomiho:
> signupCustomization -> I don't think you need Customization, just signup is enough. The whole .wasp file is configuration/customization.

Makes sense, changed 👍 

> defineAdditionalSignupFields } from "@wasp/types/index.js" -> this is not a type though, so should we import it from some other places? Maybe smth like "@wasp/auth"?

It's used to infer types, so it makes sense to keep it there. But the users don't need to know that. It can be imported maybe from `@wasp/auth/index.js` (it can't be `@wasp/auth` because of our module resolution being NodeNext)

> get receives the whole data object -> is that the whole user? And then we need to get address from it? Why is that so, why doesn't get already receive data.address?

I thought it might be useful to get all the fields and decide based on all of them what is the address (there might be some checkbox or something similar on the frontend e.g. `isWorkAddress`)

> Also, can get have a default implementation, if they don't provide it?

This is a weird one since I wanted to get the type from the return value of the `get` function ... this means that the type would need to stay `unknown`, which is okay I guess. I'll see what can be done.

>  Looking at it, it is not super simple -> you need to know how to use react-hook-form. 

I've added a new declarative way of doing it as well. That being said, `react-hook-form` is well known among React devs since it's quite commonly used.

>  As a user, can I avoid having to implement my own error showing on the client, could that be done automatically?

I'm not sure that we can assume which kind of validations the user might want on the client. Default validation implies we infer something from their code. But we can't import their server-side fields definition -> therefore we don't have access to that info on the client. By default, there are none. For better UX, they can add extra validation on the frontend alongside the server-side validation.

Martinsos:
> > signupCustomization -> I don't think you need Customization, just signup is enough. The whole .wasp file is configuration/customization.
> 
> Makes sense, changed 👍
> 
> > defineAdditionalSignupFields } from "@wasp/types/index.js" -> this is not a type though, so should we import it from some other places? Maybe smth like "@wasp/auth"?
> 
> It's used to infer types, so it makes sense to keep it there. But the users don't need to know that. It can be imported maybe from `@wasp/auth/index.js` (it can't be `@wasp/auth` because of our module resolution being NodeNext)

Sounds good, let's change it to smth else!

> 
> > get receives the whole data object -> is that the whole user? And then we need to get address from it? Why is that so, why doesn't get already receive data.address?
> 
> I thought it might be useful to get all the fields and decide based on all of them what is the address (there might be some checkbox or something similar on the frontend e.g. `isWorkAddress`)
> 
> > Also, can get have a default implementation, if they don't provide it?
> 
> This is a weird one since I wanted to get the type from the return value of the `get` function ... this means that the type would need to stay `unknown`, which is okay I guess. I'll see what can be done.
> 
> > Looking at it, it is not super simple -> you need to know how to use react-hook-form.
> 
> I've added a new declarative way of doing it as well. That being said, `react-hook-form` is well known among React devs since it's quite commonly used.
> 
> > As a user, can I avoid having to implement my own error showing on the client, could that be done automatically?
> 
> I'm not sure that we can assume which kind of validations the user might want on the client. Default validation implies we infer something from their code. But we can't import their server-side fields definition -> therefore we don't have access to that info on the client. By default, there are none. For better UX, they can add extra validation on the frontend alongside the server-side validation.

The rest sounds good to me!

Martinsos (commented):
@infomiho this is looking good to me!
What happens if I define both `additionalFields` and `additionalFieldsRender`, will they combine together, or get in some kind of conflict?

I am also wondering if there is a better name for `additionalFieldsRender`, but I don't have good ideas at the moment, so it might be just fine.

Sounds good all together!

What are the next steps -> docs?

infomiho:
> What happens if I define both additionalFields and additionalFieldsRender, will they combine together, or get in some kind of conflict?

They both render, one below the other. I was playing with the idea of making them mutually exclusive, but I'm not sure if it makes sense. 

> I am also wondering if there is a better name for additionalFieldsRender, but I don't have good ideas at the moment, so it might be just fine.

I'll explore what is the convention of naming for a prop that "renders stuff" but is not `children`.

> What are the next steps -> docs?

Yep, docs and changelog.

Martinsos:
> > What happens if I define both additionalFields and additionalFieldsRender, will they combine together, or get in some kind of conflict?
> 
> They both render, one below the other. I was playing with the idea of making them mutually exclusive, but I'm not sure if it makes sense.
> 
> > I am also wondering if there is a better name for additionalFieldsRender, but I don't have good ideas at the moment, so it might be just fine.
> 
> I'll explore what is the convention of naming for a prop that "renders stuff" but is not `children`.
> 
> > What are the next steps -> docs?
> 
> Yep, docs and changelog.

Ok, that sounds good, I think it is best to have them both render one after another, as you did at the moment. Although, you can't mix them then, right?

Different interface could be that you have just one field, `additionalFields`, and it can accept two types of objects -> simple ones, or complex ones, where complex ones would be what you are passing to `additionalFieldsRender` at the moment -> that way you have just one name, `additionalFields`, you give them the ability to mix the fields, and it is very clear to them that all of these will render?

infomiho:
Updated the description with the latest iteration of the API.

The API now supports providing an array of either objects or render functions. And, instead of the array of fields, users can simply pass in a single render function to render what ever UI they might want to render.

Martinsos (commented):
@infomiho sounds/looks good to me!

Zeko369:
A bit late to the party but

1. Why not have it automatically infer client field config from the backend config? Unless you want to customise it, it looks like a lot of duplication

2. Why not (besides the fact it doesn't fully work with our TS) have zod as a validator instead of doing custom validation? It would give you mulitple advantages (not only better and more advance error/typing, but also validation on the server)

> I'm not sure that we can assume which kind of validations the user might want on the client. Default validation implies we infer something from their code. But we can't import their server-side fields definition -> therefore we don't have access to that info on the client. By default, there are none. For better UX, they can add extra validation on the frontend alongside the server-side validation.
Zod also solves this since you can give error messages + complex validations (min 6 max 20 char, has 1 number and 1 symbol and doesn't contain a space and all of these can have a nice message with them + will work on client and server)

> I thought it might be useful to get all the fields and decide based on all of them what is the address (there might be some checkbox or something similar on the frontend e.g. isWorkAddress)
One last thing about zod and this, I'm kinda scared doing this since I started using zod since all of these things look like you're about to add a SQL Injection in your code (I know it's not that easy to do it with modern prisma but still), I'd def try to sanitise the user input as much as possible here and I think we should be providing a "default way" on how to do it (especially in a signup field)

PS: I noticed you have zod in one of the examples later after writing all of this

3. Don't get me wrong, I love RHF but it's quite far from a "user friendly option" in this case where you're learning a new framework and want to get an address field on the screen, how about a simpler API that just uses RHF `<Controller/>` component? so you essentially pass `value` and `setValue` to the user while also having this API for more advanced users? (register API has a bunch of footguns that new users will almost 100% find and then ask in the discord about them)

4. Just a side note, but I think you should destructure/delete password from the data object, you won't want users possibly storing it unencrypted in the DB and later shooting themselves in the food

infomiho:
> Why not have it automatically infer client field config from the backend config? Unless you want to customise it, it looks like a lot of duplication

We would need to import server code on the client, but the server code might import stuff that doesn't belong on the client. For example, they might validate their fields with some async server calls or similar. I haven't thought of a workaround. 

> Why not (besides the fact it doesn't fully work with our TS) have zod as a validator instead of doing custom validation? It would give you mulitple advantages (not only better and more advance error/typing, but also validation on the server)

It's a recommendation in the docs. Until we include `zod` ourselves by default, users have to install it. After we include it by default (for example for action input/output validation) I would make it the "official" way to validate fields in the docs.

>  SQL Injection in your code

I think using Prisma kinda solves this. We are not operating at the raw SQL level, so we should be good.

> Don't get me wrong, I love RHF but it's quite far from a "user friendly option"

I thought having two APIs: one simpler and one full blown complex one should solve the issue for novice users + advanced users. Since we are using RHF under the hood, how do you imagine providing `value` and `setValue` would look like? And let's not forget the errors as well.

> but I think you should destructure/delete password from the data object, you won't want users possibly storing it unencrypted in the DB and later shooting themselves in the food

That's a solid suggestion, I think it makes sense. I'll remove the password before passing in the data object 👍

Zeko369:
> > Why not have it automatically infer client field config from the backend config? Unless you want to customise it, it looks like a lot of duplication

> That would be awesome. I would love it if we could do it. ✨ We would need to import server code on the client, but the server code might import stuff that doesn't belong on the client. For example, they might validate their fields with some async server calls or similar. I haven't thought of a workaround. 

🤔, hmm makes sense, I'd leave it like this then and work around it later

> > Why not (besides the fact it doesn't fully work with our TS) have zod as a validator instead of doing custom validation? It would give you mulitple advantages (not only better and more advance error/typing, but also validation on the server)

> It's a recommendation in the docs. Until we include `zod` ourselves by default, users have to install it. After we include it by default (for example for action input/output validation) I would make it the "official" way to validate fields in the docs.

I'd maybe add it here, but if not then I'll be eagerly waiting for zod actions/queries

> >  SQL Injection in your code

> I think using Prisma kinda solves this. We are not operating at the raw SQL level, so we should be good.

You can still return a raw object to create/update/delete relations i.e. if you return `todo: (data) => data.todo` but it should be fine since you're probably not gonna be doing that

> > Don't get me wrong, I love RHF but it's quite far from a "user friendly option"

> I thought having two APIs: one simpler and one full blown complex one should solve the issue for novice users + advanced users. Since we are using RHF under the hood, how do you imagine providing `value` and `setValue` would look like? And let's not forget the errors as well.

Controller api provides a more "regular" input api, while register requires you to use a "uncontrolled" input (which is better I know) but it's harder for custom components / newer devs 

https://www.react-hook-form.com/api/usecontroller/controller
```tsx
<Controller
  control={control}
  name="test"
  render={({
    field: { onChange, onBlur, value, name, ref },
    fieldState: { invalid, isTouched, isDirty, error },
    formState,
  }) => (
    <Checkbox
      onBlur={onBlur} // notify when input is touched
      onChange={onChange} // send value to hook form
      checked={value}
      inputRef={ref}
    />
  )}
/>
```

> > but I think you should destructure/delete password from the data object, you won't want users possibly storing it unencrypted in the DB and later shooting themselves in the food

> That's a solid suggestion, I think it makes sense. I'll remove the password before passing in the data object 👍 

🚀

Martinsos (approved):
LGTM, once @Zeko369 's suggestions are covered, I am ok with it being merged!

web/docs/auth/overview.md
```diff
@@ -504,8 +504,382 @@ try {
...
+
+entity User {=psl
+  id Int @id @default(autoincrement())
+  username String @unique
+  password String
+  address String?
+psl=}
+```
+
+Then we'll export the `fields` object from the `server/auth/signup.js` file:
```
Martinsos, line 554:
```suggestion
Then we'll define and export the `fields` object from the `server/auth/signup.js` file:
```

web/docs/auth/overview.md
```diff
@@ -504,8 +504,382 @@ try {
...
+
+Let's see how to do both.
+
+### 1. Defining Extra Backend Fields
+
+If we want to **save** some extra fields in our signup process, we need to tell our backend they exist.
+
+We do that by defining an object where the keys represent the field name, and the values are functions that receive the data sent from the client and return the value of the field.
+
+First, we add the `auth.signup.additionalFields` field in our `main.wasp` file:
```
Martinsos, line 526:
```suggestion
First, we add the `auth.signup.additionalFields` field in our `main.wasp` file, via which we provide the path to the JS/TS file where we are going to define these extra fields.
```

web/docs/auth/overview.md
```diff
@@ -504,8 +504,382 @@ try {
...
+</Tabs>
+
+<small>
+
+Read more about the `fields` object in the [API Reference](#signup-fields-customization).
+</small>
+
+Keep in mind, that these field names need to exist on the `userEntity` you defined in your `main.wasp` file e.g. `address` needs to be a field on the `User` entity. 
+
+The field function will receive the data sent from the client and it needs to return the value of the field. If the field is invalid, the function should throw an error.
```
Martinsos, line 628:
```suggestion
The field function will receive the data sent from the client and it needs to return the value that will be saved into the database. If the field is invalid, the function should throw an error.
```

web/docs/auth/overview.md
```diff
@@ -504,8 +504,382 @@ try {
...
+</TabItem>
+</Tabs>
+</details>
+
+:::
+
+Now that we defined the fields, Wasp knows how to:
+
+1. Validate the data sent from the client
+2. Save the data to the database
```
Martinsos, line 692: Nice, this is clear. I would periods at the end of sentences though. Remember, if sentence starts with capital letter, it should end with period (or something similar).

web/docs/auth/overview.md
```diff
@@ -637,3 +1013,238 @@ The default value is `"/"`.
...
+    }
+    return address;
+  },
+});
+```
+
+</TabItem>
+</Tabs>
+
+The `fields` object is an object where the keys represent the field name, and the values are functions which receives the data sent from the client and return the value of the field.
```
Martinsos, line 1105:
```suggestion
The `fields` object is an object where the keys represent the field name, and the values are functions which receive the data sent from the client and return the value of the field.
```

web/docs/auth/overview.md
```diff
@@ -637,3 +1013,238 @@ The default value is `"/"`.
...
+
+If the field value is invalid, the function should throw an error.
+
+The field details object has the following properties:
+
```
Martinsos, line 1110: This stuff is same as above -> is there any way we could avoid this duplication?

web/docs/auth/overview.md
```diff
@@ -504,8 +504,382 @@ try {
...
+## Customizing the Signup Process
+
+Sometimes you want to include **extra fields** in your signup process, like first name and last name.
+
+In Wasp, in this case:
+
+- You need to define the new backend fields
+- You need to customize the `SignupForm`
+
+Other times, you might need to add some **extra UI** elements to the form, like a checkbox for terms of service. In this case, customizing only the UI components is enough.
```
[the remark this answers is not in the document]
infomiho: > Avoid language that invalidate struggle, such as "easy", "just", "obviously", etc. For reference, see [Words To Avoid in Educational Writing](https://css-tricks.com/words-avoid-educational-writing/).

https://github.com/vuejs/v2.vuejs.org/blob/master/writing-guide.md

web/docs/auth/overview.md
```diff
@@ -504,8 +504,382 @@ try {
 </TabItem>
 </Tabs>
 
+## Customizing the Signup Process
+
+Sometimes you want to include **extra fields** in your signup process, like first name and last name.
```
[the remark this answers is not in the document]
infomiho: We have dropped hints that they can do that:
- here https://wasp-lang.dev/docs/auth/email#fields-in-the-email-dict
- and here https://wasp-lang.dev/docs/auth/username-and-pass#customizing-the-auth-flow


I could add an extra note here as well 👍

web/docs/auth/overview.md
```diff
@@ -504,8 +504,382 @@ try {
...
+- You need to define the new backend fields
+- You need to customize the `SignupForm`
+
+Other times, you might need to add some **extra UI** elements to the form, like a checkbox for terms of service. In this case, customizing only the UI components is enough.
+
+Let's see how to do both.
+
+### 1. Defining Extra Backend Fields
+
+If we want to **save** some extra fields in our signup process, we need to tell our backend they exist.
```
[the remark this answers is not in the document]
infomiho: Sure thing, I can adjust the language, although for me the two are interchangeable. Like client and frontend.

web/docs/auth/overview.md
infomiho: If we want to keep the API reference independent, there will be some duplication always.

web/docs/auth/overview.md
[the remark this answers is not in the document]
Martinsos: Sure, that makes sense, but this wasn't for the purpose of making it sound easy, the purpose was to make a contrast compared to the previous part where we had to do two things, while here we do only one thing. When I was reading it the first time, I thought that is not very clear, that we are doing only half of the job here.
I get what they mean with not using "just", but I don't think this case falls under that. Certainly not all usages of "just" should be eliminated, just those that invalidate struggle, like "now let's just implement this thing" or "the only thing left is just running the command".
Maybe there is another way to indicate this is only part of what was needed above it?

web/docs/auth/overview.md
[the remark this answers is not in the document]
Martinsos: Ok, whatever you think makes sense!

web/docs/auth/overview.md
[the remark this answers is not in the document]
Martinsos: Right, but we do use client and server mostly. The reason why is that client and server are more precise terms. Frontend and backend can be quite relative and mean many things: backend of compiler for example.

web/docs/auth/overview.md
Martinsos: You don't think it is good? The reason I put this is because I found it unclear as to which value that field should have. You set it to `"@server/auth/signup.js"`, but why, what is the intention? Does it have to that file name, can it be something else?
If you feel it is clear, ok, leave it as it is.

web/docs/auth/overview.md
Martinsos: But this is identical, right? The more duplication we have, more chances of it going out of sync. I know we can't remove all duplication, some of it is semantical, but this seems like literal copy paste, that is why I am asking, also for the future, if not now.
We could maybe, when we have duplication like this, extract it into a React component?

web/docs/auth/overview.md
infomiho: Based on the "Show, don't tell" principle, this sentence feels redundant. It's clear from the context that we are putting a JS/TS file there and it's clear that in that file we are defining the fields.

web/docs/auth/overview.md
infomiho: Adjusted this piece as well 👍

web/docs/auth/overview.md
infomiho: > some of it is semantical, but this seems like literal copy paste

Once we pull out this content into a API Reference component, then it needs to be standalone. I could pull out this piece of repetition into a tiny React component. But then we need to make sure that tiny piece fits in both contexts over time (for example, it can't refer to something above or below itself - since it's used in two difference places) which think is worse than having some duplication.

web/docs/auth/overview.md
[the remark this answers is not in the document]
infomiho: My bad, I jumped right at the word without understanding that real meaning behind the change. The change makes a lot of sense 👍

web/docs/auth/overview.md
```diff
@@ -504,8 +504,382 @@ try {
...
 </Tabs>
 
+## Customizing the Signup Process
+
+Sometimes you want to include **extra fields** in your signup process, like first name and last name.
+
+In Wasp, in this case:
+
+- You need to define the new backend fields
+- You need to customize the `SignupForm`
```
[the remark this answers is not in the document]
infomiho: Adjusted the language 👍

web/docs/auth/overview.md
Martinsos: To me that wasn't clear from the context, not until I read further along. But ok, I will leave it to you.

web/docs/auth/overview.md
Martinsos: Sure, if it can't be the same in both places, that is fine, but in case we do have pieces that are completely the same, I think it is worth it investigating extracting them.

Zeko369 (approved):

PR #1423, opened 2023-08-31
Updates tutorial link in templates

infomiho:
Linking to wrong tutorial page in our basic template.

### DIFF
--- a/web/docusaurus.config.js
+++ b/web/docusaurus.config.js
@@ -234,7 +234,7 @@ module.exports = {
           },
           {
             from: '/docs/tutorials/todo-app',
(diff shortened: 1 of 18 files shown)


web/docusaurus.config.js
```diff
@@ -234,7 +234,7 @@ module.exports = {
           },
           {
             from: '/docs/tutorials/todo-app',
-            to: '/docs/tutorial/project-structure',
+            to: '/docs/tutorial/create',
```
infomiho, line 237: Linking to the first page of the tutorial since we want to link to the tutorial, not a specific page.

PR #1425, opened 2023-08-31
Updated default .gitignore for new Wasp project to be more robust regarding env files

### LINKED ISSUE #1411: Make .gitignore for new Wasp projects more aggressive when ignoring .env files
Right it has
```
/.env.server
/.env.client
```
which works correctly for the way .env files are used in Wasp, but, the thing is, if user by accident moves one of those files to the wrong place, it will not be git ignored any more! That `/` at the start targets only top level .env files.

In situations like these it is better to be safe then sorry, so we should be more aggressive about it, and have something like
```
.env.server
.env.client
```

I would go even step further, and do smth like
```
.env.server
.env.client
*.env
*.env.*
```
I am not 100% sure if the above is correct glob for gitignore, would have to verify, but what I am suggesting is that we should try to ignore all the .env files that are in the repo, and if that is too aggressive, they can always adjust the .gitignore file. That is why I added "redundant" `.env.server` and `.env.client`, so it is clear in gitignore that these we target specifically.

Martinsos:
Fixes #1411 .

 - [X] **Bug fix** (non-breaking change which fixes an issue).

1. [X] I updated [ChangeLog.md](https://github.com/wasp-lang/wasp/blob/main/waspc/ChangeLog.md) with description of the change this PR introduces.
2. [X] I bumped waspc version in [waspc.cabal](https://github.com/wasp-lang/wasp/blob/main/waspc/waspc.cabal) to reflect changes I introduced, with regards to the version of the latest wasp release, if the bump was needed.

### DIFF
--- a/waspc/data/Cli/templates/basic/.gitignore
+++ b/waspc/data/Cli/templates/basic/.gitignore
@@ -1,3 +1,11 @@
 /.wasp/
-/.env.server
-/.env.client
+
+# We ignore env files recognized and used by Wasp.
+.env.server
+.env.client
+
+# To be extra safe, we by default ignore any files with `.env` extension in them.
+# If this is too agressive for you, consider allowing specific files with `!` operator,
+# or modify/delete these two lines.
+*.env
+*.env.*
--- a/waspc/ChangeLog.md
+++ b/waspc/ChangeLog.md
@@ -2,9 +2,6 @@
 
 ## 0.11.3
 
-### 🐞 Bug fixes
-- Fixes API types exports for TypeScript users
-
 ### 🎉 [New Feature] Type-safe links
 
 Wasp now offers a way to link to pages in your app in a type-safe way. This means that you can't accidentally link to a page that doesn't exist, or pass the wrong arguments to a page.
@@ -48,11 +45,12 @@ import { routes } from '@wasp/router'
 const linkToTask = routes.TaskRoute({ params: { id: 1 } })
 ```
 
-## 0.11.2
+### 🐞 Bug fixes
+- Fixes API types exports for TypeScript users.
+- Default .gitignore that comes with new Wasp project (`wasp new`) is now more aggressive when ignoring .env files, ensuring they don't get committed by accident (wrong name, wrong location, ...).
 
-### 🐞 Bug fixes / 🔧 small improvements
-- Wasp copied over the `.env.server` instead of `.env.client` to the client app `.env` file. This prevented using the `.env.client` file in the client app.
-- waspls thought that importing `"@client/file.jsx"` could mean `"@client/file.tsx"`, which could hide some missing import diagnostics and cause go-to definition to jump to the wrong file.
+
+## 0.11.2
 
 ### 🎉 [New Feature] waspls Code Scaffolding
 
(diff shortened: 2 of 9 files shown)


infomiho:
@Martinsos Please merge `main` into this branch and check out the Changelog

waspc/data/Cli/templates/basic/.gitignore
```diff
@@ -1,3 +1,11 @@
 /.wasp/
-/.env.server
-/.env.client
+
+# We ignore env files recognized and used by Wasp.
+.env.server
+.env.client
+
+# To be extra safe, we by default ignore any files with `.env` extension in them.
```
infomiho, line 7: I would avoid doing this and only take care of the Wasp specific `.env` files.

Next.js's gitignore: https://github.com/vercel/next.js/blob/canary/packages/create-next-app/templates/default/ts/gitignore
Vite React's gitignore: https://github.com/vitejs/vite/blob/main/packages/create-vite/template-react-ts/_gitignore
Remix: https://github.com/remix-run/remix/blob/main/templates/remix/.gitignore

None of the above try to "protect" users that much. I would personally avoid doing this.

martinovicdev: What about giving users the option for easy switch in .gitignore? That way you aren't too protective of users, but you offer them a way to include it themselves without too much hassle.

For example:

#For ignoring .env files, uncomment the lines below. 
#.env.server
#.env.client

Martinsos: Interesting that they don't, but what are the arguments for not doing so?

I quite like the solution in this PR, as it ensures that credentials are not submitted by accident. You can easily name your file `.env.sevrer` instead of `.env.server` and then have quite some problems. Or maybe you name it `.server.env` by accident.
@matijaSos had unpleasant experience where he put `.env.server` in the wrong location and committed it by accident -> if Sengrid's bot hasn't detected it, it would stay up there for who knows how long. And he has to doctor the branch then, change API keys, ... .

Wasp is about best practices, and the best default is not to commit any .env files -> that is why we would have this .gitignore .
The only "bad" side of it is if you really want to commit something that has `.env` in it. That won't happen often though, maybe once or twice for the lifetime of the project, and it will likely be something like `example.env.server`. So the "bad" thing then is that it will be ignored by default, and user will have to realize it is due to .gitignore file and edit it, ideally by adding `!example.env.server` to it. I see that as a minor inconvenience, and a good trade off, in order to have a piece of mind to not by accident commit a dot env file.

infomiho: I understand your position and agree that Wasp is about best practices. I've tried to lay out some of the industry leading frameworks' preferences when it comes to the `.gitignore` file. 

I've stated my personal preference not to do it, but I won't push the issue too hard. This is also okay 👍

Martinsos: > 

That is interesting idea but I would rather go the opposite -> give them an easy way to opt out of protection.

Btw for .env.client and .env.server we certainly want to do the ignoring.

The question is though do we want to do it for everything that has .env in it.

matijaSos: yep, I agree, we saw in practice in happened multiple times already. With wasp it's a bit tricky because its not obvious where should the env file be. If we had only one place where it makes sense it is then I'd say it's ok to go with relaxed version, but in practice this has shown to be a problem.

So I'd rather be more restrictive for now, and it's very unlikely IMO that somebody will want to upload their env files.

matijaSos (approved):

waspc/data/Cli/templates/basic/.gitignore
Martinsos: Ok, good feedback everybody, we decided to give this more defensive approach a try, and let's see what people will say!

PR #1431, opened 2023-09-01
Add docs on how to build wasp from source on M1/M2 macs

Zeko369:
### Description

> Adds a short section to the waspc/README.md about how to run/setup rosetta on M1/M2 macs

### DIFF
--- a/waspc/README.md
+++ b/waspc/README.md
@@ -34,7 +34,32 @@ It can then also run that web app for you, deploy it (not yet but that is coming
 ### Setup
 We use the [`cabal`](https://cabal.readthedocs.io/) command line tool to build
 the project. The best way to install it is via
-[ghcup](https://www.haskell.org/ghcup/). The `ghcup tui` terminal user interface
+[ghcup](https://www.haskell.org/ghcup/) (if you're running an M1/M2 mac make sure to run it in rosetta).
+
+<details>
+<summary>M1/M2 rosetta setup</summary>
+
+1. Regular Shell: just run `arch -x86_64 zsh` in your terminal.
+2. IntelliJ: Preferences -> Tools -> Terminal -> Shell path -> `env /usr/bin/arch -x86_64 /bin/zsh --login`
+3. VSCode: add the following to your local `.vscode/settings.json` file
+
+```json
+{
+  "terminal.integrated.profiles.osx": {
+    "rosetta": {
+      "path": "/usr/bin/arch",
+      "args": ["-arch", "x86_64", "/bin/zsh"],
+      "overrideName": true
+    }
+  },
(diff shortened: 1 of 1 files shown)


waspc/README.md
```diff
@@ -34,7 +34,32 @@ It can then also run that web app for you, deploy it (not yet but that is coming
 ### Setup
 We use the [`cabal`](https://cabal.readthedocs.io/) command line tool to build
 the project. The best way to install it is via
-[ghcup](https://www.haskell.org/ghcup/). The `ghcup tui` terminal user interface
+[ghcup](https://www.haskell.org/ghcup/) (if you're running an M1/M2 mac make sure to run it in rosetta).
```
Martinsos, line 37: So what exactly do you have to run in rosetta -> `ghcup`, during installation, or `cabal` later when building the project, or both?

waspc/README.md
```diff
@@ -34,7 +34,32 @@ It can then also run that web app for you, deploy it (not yet but that is coming
...
+
+<details>
+<summary>M1/M2 rosetta setup</summary>
+
+1. Regular Shell: just run `arch -x86_64 zsh` in your terminal.
```
Martinsos, line 42: This uses rosetta for the current terminal session?

waspc/README.md
```diff
@@ -34,7 +34,32 @@ It can then also run that web app for you, deploy it (not yet but that is coming
...
+2. IntelliJ: Preferences -> Tools -> Terminal -> Shell path -> `env /usr/bin/arch -x86_64 /bin/zsh --login`
```
Martinsos, line 43: Are we ok with doing this for all of the shells that IntelliJ runs? Because this will then be also used when you are not dealing with Haskell, right? Same question for vscode below.

waspc/README.md
Zeko369: You need to install `ghcup` and then I suppose also run `cabal` with rosetta (installing ghcup in m1 mode and then trying to run it or `cabal` didn't work)

waspc/README.md
Zeko369: Yup, without any permanent changes

waspc/README.md
Zeko369: This sets it per project (so only in this folder -> `.idea` is the config for it) and same for vscode (I suggested creating the `.vscode/settings.json` to make it local)

Martinsos:
@Zeko369 could you enrich the docs with this info that I asked about in comments? So provide a bit more context of which session / project these commands affect, also probably mention .vscode/settings.json . I think we are good then!

cprecioso:
This (thankfully) is not needed anymore! We upgraded GHC (#3302) to a version that has native Apple Silicon support, so there are no workarounds needed.

PR #1443, opened 2023-09-08
Adding support for Postgres Extensions

infomiho:
### Motivation

We want to add support for `extensions` in the `schema.prisma` file.

This is how ti looks like in the `schema.prisma` file:

```diff
datasource db {
  provider = "postgresql"
  url      = env("DATABASE_URL")
+ extensions = [pgvector(map: "vector")]
}
```

### Implementation

We have added a `dbExtensions` field under `app.db.prisma` to enable users to define which extensions they want to enable. 

The API looks like this:

```diff
app pgVectorExample {
  // ...
  db: {
    system: PostgreSQL,
    prisma: {
      clientPreviewFeatures: ["postgresqlExtensions"],
+     dbExtensions: [{
+       name: "pgvector",
+       map: "vector"
+     }]
    }
  }
}
```

It's a simple 1-on-1 API that converts that dict in the Wasp file into a string that can be appended to the Prisma schema file. It's unit tested 👍 

### Example App

https://github.com/wasp-lang/wasp/assets/2223680/4b5e934a-a68e-4c80-996e-76c43446c1fa

A new example is added called `pg-vector-example` where we showcase support for the `pg_vector` extension which has gained a lot of popularity lately due to its usage for semantic search + LLM generating answers.

Make sure to start the dev db with `./run_db.sh` (it will start a Postgres DB with `pg_vector` already installed.


waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -324,6 +325,48 @@ validateCrudOperations spec =
...
+
+    isPostgresDbUsed = isPostgresUsed spec
+
+    checkIfPostgresExtensionsAreUsedWithoutPostgresDbSystem :: [ValidationError]
+    checkIfPostgresExtensionsAreUsedWithoutPostgresDbSystem = case prismaClientPreviewFeatures of
+      Nothing -> []
+      Just value ->
+        [ GenericValidationError "You enabled \"postgresqlExtensions\" in app.db.prisma.clientPreviewFeatures but your db system is not PostgreSQL."
+          | not isPostgresDbUsed && "postgresqlExtensions" `elem` value
+        ]
```
Martinsos, line 346: I know we use this pattern with list comprehension sometimes, but I find it a bit weird always to read, confusing. I would probably handle it with `if-then-else`, but I understand this also if you like it better. Up to you.

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -324,6 +325,48 @@ validateCrudOperations spec =
...
+
+    checkIfDbExtensionsAreUsedWithoutPostgresDbSystem :: [ValidationError]
+    checkIfDbExtensionsAreUsedWithoutPostgresDbSystem = case prismaDbExtensions of
+      Nothing -> []
+      Just value ->
+        [ GenericValidationError "You are using app.db.prisma.dbExtensions but your db system is not PostgreSQL."
```
Martinsos, line 352: This doesn't really it make it clera why is that wrong. Maybe you could mention that PostgreSQL is required in order to specify extensions, currently.

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -324,6 +325,48 @@ validateCrudOperations spec =
...
+    checkIfDbExtensionsAreUsedWithoutPostgresExtensions :: [ValidationError]
+    checkIfDbExtensionsAreUsedWithoutPostgresExtensions = case (prismaDbExtensions, prismaClientPreviewFeatures) of
+      (Just _extensions, Just features) ->
+        [ GenericValidationError extensionsNotEnabledMessage
+          | "postgresqlExtensions" `notElem` features
+        ]
+      (Just _extensions, Nothing) -> [GenericValidationError extensionsNotEnabledMessage]
+      _anyOtherCase -> []
+      where
+        extensionsNotEnabledMessage = "You are using app.db.prisma.dbExtensions but you didn't enable \"postgresqlExtensions\" in app.db.prisma.clientPreviewFeatures."
```
Martinsos, line 365:
```suggestion
    checkIfDbExtensionsAreUsedWithoutPostgresExtensions :: [ValidationError]
    checkIfDbExtensionsAreUsedWithoutPostgresExtensions = case (prismaDbExtensions, prismaClientPreviewFeatures) of
      (Nothing, _) -> []
      (Just _extensions, Just features) | "postgresqlExtensions" `elem` features -> []
      (Just _extensions, _) -> [GenericValidationError extensionsNotEnabledMessage]
      where
        extensionsNotEnabledMessage = "You are using app.db.prisma.dbExtensions but you didn't enable \"postgresqlExtensions\" in app.db.prisma.clientPreviewFeatures."
```

waspc/src/Wasp/Generator/DbGenerator/Prisma.hs
```diff
@@ -0,0 +1,31 @@
...
+showDbExtensions extensions = "[" <> intercalate ", " (map showDbExtension extensions) <> "]"
+  where
+    showDbExtension :: AS.Db.PrismaDbExtension -> String
+    showDbExtension
+      AS.Db.PrismaDbExtension
+        { AS.Db.name = name,
+          AS.Db.version = version,
+          AS.Db.map = extensionMap,
+          AS.Db.schema = schema
+        } = if null fields then name else name <> "(" <> fields <> ")"
```
Martinsos, line 20:
```suggestion
        } = name <> (if null fields then "" else "(" <> fields <> ")")
```

waspc/src/Wasp/Generator/DbGenerator/Prisma.hs
```diff
@@ -0,0 +1,31 @@
...
+        where
+          fields = intercalate ", " $ mapMaybe showField possibleFields
+
+          showField :: (String, Maybe String) -> Maybe String
+          showField (fieldName, maybeFieldValue) = maybeFieldValue <&> \fieldValue -> fieldName <> ": " <> show fieldValue
```
Martinsos, line 25: More readable:
```hs
          showField :: (String, Maybe String) -> Maybe String
          showField (_name, Nothing) = Nothing
          showField (name, Just value) = Just $ name <> ": " <> show value
```

Or

```hs
          showField :: (String, Maybe String) -> Maybe String
          showField (name, maybeValue) = ((name <>) . (": " <>) . show) <$> maybeValue
```

waspc/src/Wasp/Generator/DbGenerator/Prisma.hs
```diff
@@ -0,0 +1,31 @@
+module Wasp.Generator.DbGenerator.Prisma where
+
+import Data.Functor ((<&>))
+import Data.List (intercalate)
+import Data.Maybe (mapMaybe)
+import qualified Wasp.AppSpec.App.Db as AS.Db
+
+-- We want to show db extensions in the following format:
+-- [first_ext, some_other(map: "name", schema: "schema", version: "1.0.0")]
+showDbExtensions :: [AS.Db.PrismaDbExtension] -> String
```
Martinsos, line 10: We actually have part of our codebase dedicated to parsing and generating PSL, and this seems like it should belong there, so look for it and consider putting it there! It should be something like Wasp.PSL .

waspc/src/Wasp/Generator/Job/Process.hs
```diff
@@ -101,6 +101,7 @@ runNodeCommandAsJobWithExtraEnv extraEnvVars fromDir command args jobType chan =
     Right _ -> do
       envVars <- getAllEnvVars
       let nodeCommandProcess = (P.proc command args) {P.env = Just envVars, P.cwd = Just $ SP.fromAbsDir fromDir}
+      -- if head args == "generate" then error $ command ++ " " ++ show args ++ " " ++ show envVars else pure ()
```
Martinsos, line 104: Leftover?

Martinsos (approved):
Looks good! Left some comments, but minor.
As for the example app -> do we want to make it part of CI or smth like that? How will we ensure it stays updated / relevant?

waspc/src/Wasp/Generator/DbGenerator/Prisma.hs
infomiho: Makes sense 👍 

I've added it to the `Wasp.Psl.Generator.Extensions` module.

PR #1453, opened 2023-09-26
[WIP] Scaffolding of various features

infomiho:
Working on:
- [ ] Tailwind CSS scaffolding
- [ ] Auth scaffolding
- [ ] Email sender scaffolding

Biggest issues atm:
- the DX is not where we want it to be
  - we are prompting the user to edit the Wasp file by hand instead of doing it automatically
  - the blocker for automatic editing is figuring out the best way to edit Wasp's AST and pretty print it


infomiho:
This is a stale PR so I'll close it. This was an attempt that hit a blocker of not being able to edit Wasp files so the DX wasn't where we wanted it. This PR can serve as input for https://github.com/wasp-lang/wasp/issues/3308

PR #1463, opened 2023-10-03
[MAGE] add modals

vincanger:
after generating 2 apps, on returning to Mage: 
image

after each successful generation: 
image

also:
-confetti
-title
-gh login set up, but not yet implemented


wasp-ai/main.wasp
```diff
@@ -41,13 +42,17 @@ app waspAi {
     userEntity: User,
     externalAuthEntity: SocialLogin,
     methods: {
+      gitHub: {
```
infomiho, line 45: @vincanger when you deploy an update make sure to put the Github env variables

wasp-ai/src/client/pages/MainPage.jsx
```diff
@@ -68,8 +70,46 @@ const MainPage = () => {
 
   const [appAuthMethod, setAppAuthMethod] = useState(availableAuthMethods[0]);
 
+  useEffect(() => {
+    const appDetails = JSON.parse(localStorage.getItem("appDetails"));
```
infomiho, line 74: `JSON.parse` can often fail, let's put this in a try catch block with some default values

wasp-ai/main.wasp
vincanger: we're not using github login at all right now, plus I was going to leave deployment to you guys

infomiho (approved):
Let's ship it

wasp-ai/src/client/pages/ResultPage.jsx
```diff
@@ -393,6 +406,87 @@ function getCardinalNumber(number) {
   }
 }
 
+export function StarOurRepoModal({ isStarRepoOpen, setIsStarRepoOpen, appGenerationResult }) {
+  const [tokens, setTokens] = useState(0)
+  const { data: numTotalProjects } = useQuery(getNumProjects, {}, { enabled: isStarRepoOpen })
+
+  const tokenNumberStr = appGenerationResult?.project?.logs?.filter(log => log.content.toLowerCase().includes("tokens usage") === true)[0]?.content.split(':')[1]?.trim()
```
Martinsos, line 413: Not bad! I would maybe prefer regex, since it is often more robust, but this can also work.

wasp-ai/src/client/pages/ResultPage.jsx
```diff
@@ -393,6 +406,87 @@ function getCardinalNumber(number) {
...
+
+  useEffect(() => {
+    if (!!tokenNumberStr) {
+      const num = tokenNumberStr.slice(1, -1) * 1000
```
Martinsos, line 417: While the whole parsing of "tokens usage" is fragile due to output possibly changing, this part feels especially fragile, where you count on specific characters being in the right place. Btw what shape are you assuming here? " 12k" or " 3k" or something like that, and you are trying to remove the initial space and the "k" at the end?
In that case, it would be better to make that assumption more clear, e.g. you could use regex that says it expects some white space, then some digits, and then a letter "k", and then takes the number out (matches on it).

Martinsos: Btw, what you could do, is detecting if parsing failed (if that thing is not a number), and then just display initial thing instead of parsing it.

wasp-ai/src/client/pages/ResultPage.jsx
```diff
@@ -393,6 +406,87 @@ function getCardinalNumber(number) {
   }
 }
 
+export function StarOurRepoModal({ isStarRepoOpen, setIsStarRepoOpen, appGenerationResult }) {
+  const [tokens, setTokens] = useState(0)
```
Martinsos, line 410: This should be named `numTokens` or `numTokensSpent`.

wasp-ai/src/client/pages/ResultPage.jsx
Martinsos: WHy is this not in "useEffect"? If I am correct this will execute on every component render, while you use it only once, on first component render (since it is in useEffect)?

wasp-ai/src/client/pages/ResultPage.jsx
```diff
@@ -393,6 +406,87 @@ function getCardinalNumber(number) {
...
+    if (!!tokenNumberStr) {
+      const num = tokenNumberStr.slice(1, -1) * 1000
+      setTokens(num)
+    }
+  }, [tokenNumberStr])
+
+  useEffect(() => {
+    if (isStarRepoOpen) {
+      jsConfetti.addConfetti({
+        emojis: ['🐝'],
```
Martinsos, line 425: ha love it

wasp-ai/src/client/pages/ResultPage.jsx
```diff
@@ -393,6 +406,87 @@ function getCardinalNumber(number) {
...
+        <p className="text-base leading-relaxed text-gray-500">
+          This helps spread the word, so we can keep making Mage better.
+        </p>
+        <p className="text-base leading-relaxed text-gray-500">We'd very much appreciate it! 🧙</p>
+      </div>
+    </MyDialog>
+  )
+}
+
+function FormattedText({ children }) {
```
Martinsos, line 484: I would probably put this one inside `StarOurRepoModal` since it is used only there and is very specific just for it.

wasp-ai/src/server/auth.ts
```diff
@@ -13,3 +13,19 @@ export const getGoogleAuthConfig = () => {
...
+  return { email };
+};
+
+export function getGitHubAuthConfig() {
+  return {
+    clientID: process.env.GITHUB_CLIENT_ID, // look up from env or elsewhere
+    clientSecret: process.env.GITHUB_CLIENT_SECRET, // look up from env or elsewhere
+    scope: ["user"],
+  };
+}
```
Martinsos, line 31: Drop it if we are not using it right now?

wasp-ai/src/server/operations.ts
```diff
@@ -223,6 +224,12 @@ export const getStats = (async (_args, context) => {
   };
 }) satisfies GetStats<{}>;
 
+export const getNumProjects = (async (_args, context) => {
+  const { Project } = context.entities;
+  const numProjects = await Project.count();
+  return numProjects;
```
Martinsos, line 230: Could have been just
```js
return context.entities.Project.count()
```

wasp-ai/src/client/pages/MainPage.jsx
```diff
@@ -208,4 +251,34 @@ The simpler and more specific the app is, the better the generated app will be."
     </div>
   );
 };
-export default MainPage;
\ No newline at end of file
+export default MainPage;
+
+export function GhModal({ isGhModalOpen, setIsGhModalOpen }) {
```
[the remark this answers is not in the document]
vincanger: this is a simplistic version of the modal that gets shown to the user on the main page after generating two apps:
image

wasp-ai/src/client/pages/MainPage.jsx
```diff
@@ -110,6 +151,8 @@ const MainPage = () => {
     <div className="container">
       <Header currentStatus={currentStatus} isStatusVisible={true} />
 
+      <GhModal isGhModalOpen={isGhModalOpen} setIsGhModalOpen={setIsGhModalOpen} />
```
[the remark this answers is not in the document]
vincanger: see my  above comment

wasp-ai/src/client/pages/MainPage.jsx
```diff
@@ -68,12 +70,51 @@ const MainPage = () => {
...
+        setAppPrimaryColor(availableColors.find((color) => color.name === appDetails.appPrimaryColor));
+        setAppAuthMethod(availableAuthMethods.find((method) => method.value === appDetails.appAuthMethod));
+        setCreativityLevel(availableCreativityLevels.find((level) => level.value === appDetails.appCreativityLevel));
+        localStorage.removeItem("appDetails");
+      }
+    } catch (error) {
+      console.error(error);
+    }
+  }, []);
+
```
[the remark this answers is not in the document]
vincanger: I thoiught it would be nice to save their app description and repopulate it on return to the main page, so the prompt and settings are to adjust/modify on regenerations

wasp-ai/src/client/pages/MainPage.jsx
```diff
@@ -68,12 +70,51 @@ const MainPage = () => {
...
+      localStorage.setItem("appDetails", JSON.stringify({
+        appName,
+        appDesc,
+        appPrimaryColor: appPrimaryColor.name,
+        appAuthMethod: appAuthMethod.value,
+        appCreativityLevel: creativityLevel.value,
+      }));
+    } catch (error) {
+      console.error(error)
+    }
```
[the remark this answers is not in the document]
vincanger: see above comment

wasp-ai/src/client/pages/MainPage.jsx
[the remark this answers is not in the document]
Martinsos: Aha so we have two modals, one asking them here after they generate two apps, and another one once the app is done? Ok! Isn't that a bit much, bothering them with this modal after every new generation?

wasp-ai/src/client/pages/MainPage.jsx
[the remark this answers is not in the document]
Martinsos: Ah got it, that is pretty cool!

Let's improve the code so the intention here is obvious. You could have two functions, one called "locallySaveLastGenerationInput" and "locallyLoadLastGenerationInput". That way you hide a bit the implementation details, plus the code is more readable and intention is clearer. Maybe there is a bit more that could be done here but it will be more obvious after these changes.

PR #1465, opened 2023-10-03
Allow customising the Vite config 

### LINKED ISSUE #1085: Allow extending the Vite config
We would allow the user to add things to the Vite config, and we have two different approaches:

1. If the users include a `vite.config.ts` file, we drop ours and use theirs . 
    - The downside is users would need to setup everything like we did.
2. We use the `mergeConfig` API from Vite (https://vitejs.dev/guide/api-javascript.html#mergeconfig)
    - Users would provide a vite.config.ts that exports a partial config (things they wish to add or override)
    - and we merge it with the default config.

infomiho:
It allows the user to specify the overrides for the Vite config which is used for the React web app. Closes #1085

### How it works

Users need to define a `vite.config.ts` or a `vite.config.js` in their `src/client` folder. Wasp picks up this file and uses the contents of it to override the defaults.

For example, one such override would be "Does the browser open by default when a Wasp project is started?":
```js
export default {
  server: {
    open: false,
  },
};
```

### Demo app

There is a simple demo app in `waspc/examples/vite-config-demo` for testing purposes.


waspc/data/Generator/templates/react-app/vite.config.ts
```diff
@@ -1,21 +1,34 @@
+{{={= =}=}}
 /// <reference types="vitest" />
-import { defineConfig } from 'vite'
-import react from '@vitejs/plugin-react-swc'
+import { mergeConfig } from "vite";
+import react from "@vitejs/plugin-react-swc";
 
-// https://vitejs.dev/config/
-export default defineConfig({
+{=# isCustomViteConfigUsed =}
```
infomiho, line 6: If the file is defined, we import it. If not, we use an empty object.

waspc/data/Generator/templates/react-app/vite.config.ts
```diff
@@ -1,21 +1,34 @@
...
-    environment: 'jsdom',
-    setupFiles: ['./src/test/vitest/setup.ts'],
+    environment: "jsdom",
+    setupFiles: ["./src/test/vitest/setup.ts"],
   },
-})
+};
+
+// https://vitejs.dev/config/
+export default mergeConfig(
```
infomiho, line 31: Using the built-in Vite helper to merge the objects.

waspc/examples/custom-vite-demo/src/client/vite.config.js
```diff
@@ -0,0 +1,6 @@
+export default {
```
infomiho, line 1: Example overrides.

waspc/src/Wasp/Generator/WebAppGenerator/Vite.hs
```diff
@@ -0,0 +1,20 @@
+module Wasp.Generator.WebAppGenerator.Vite where
+
+import StrongPath (File', Path', Rel, relfile)
+import Wasp.AppSpec.ExternalCode (SourceExternalCodeDir)
+import qualified Wasp.AppSpec.ExternalCode as ExternalCode
+
+checkIfCustomViteConfigUsed :: [ExternalCode.File] -> Bool
```
infomiho, line 7: If we detect `vite.config.ts` or `vite.config.js` among the client external files: use the custom config to override the defaults.

waspc/data/Generator/templates/react-app/vite.config.ts
```diff
@@ -1,21 +1,34 @@
...
 /// <reference types="vitest" />
-import { defineConfig } from 'vite'
-import react from '@vitejs/plugin-react-swc'
+import { mergeConfig } from "vite";
+import react from "@vitejs/plugin-react-swc";
 
-// https://vitejs.dev/config/
-export default defineConfig({
+{=# isCustomViteConfigUsed =}
+import _waspUserProvidedConfig from "./src/ext-src/vite.config";
```
Martinsos, line 7: Mhm this is very hardcoded here, can we instead pass this path from Haskell and make it less hardcoded? I think it should not be very hard to do?

waspc/src/Wasp/AppSpec.hs
```diff
@@ -75,7 +75,8 @@ data AppSpec = AppSpec
     configFiles :: [ConfigFileRelocator],
     -- | Connection URL for a database used during development. If provided, generated app will
     -- make sure to use it when run in development mode.
-    devDatabaseUrl :: Maybe String
+    devDatabaseUrl :: Maybe String,
+    isCustomViteConfigUsed :: Bool
```
Martinsos, line 79: Since this can be determined based on `externalClientFiles`, I would not have additional field here, because it is introducing redundancy and opportunity for potential desync.

Piece of knowledge we are introducing in this PR is that there can be a vite config file defined by user at specific path(s) and we will use it.
So ideally we want a single place to encode that.

I see two ways to go about it:
1. Let `Wasp.Project` part of `Wasp` have the knowledge of what exact path is, and it will then store that path into AppSpec under `customViteConfigFilePath :: Maybe FilePath` -> ok you should use StrongPath that is relative to project source dir root, not FilePath, but you get the idea. That means that in `Wasp.Project` (or somewhere beneath it) we will have logic that searches for that file and then fills in this field in AppSpec, and from there on, AppSpec caries the info about the path and that is it.
2. Or, we can have AppSpec not have a field at all, but instead have a function called `isCustomViteConfigUsed` that takes app spec and returns Bool but doing the check through the client files. You will end up with the same type signature but it will not be field but a getter that calculates stuff on the fly. But that means that AppSpec has the knowledge of the path of the vite config file in that function, so how do other parts of code learn about that path? I guess then we add a constant somewhere in AppSpec module or submodule, which contains that path to vite config file, and that is used both by `isCustomViteConfigUsed` and also others later (the vite template file).

Of these two, I think I actually like the first one better. We could also check how we do it for other files, where do we encode their paths, but this feels right to me.

waspc/src/Wasp/Generator/WebAppGenerator/Vite.hs
Martinsos: I would move this logic to `isCustomViteConfigUsed`, as suggested above.

Also, this logic has nothing to do with the Generator, since it is instead all about how our project is defined.

waspc/src/Wasp/AppSpec.hs
infomiho: Now that I think about it, it makes sense for me to calculate the Vite config file path e.g. `Maybe Path' Rel r File f`and keep in the AppSpec. The field could be called `customViteConfigPath`

1. If it's there, use it in the template
2. If other parts of the code need it, it is also there for them to consume

It feels okay to keep it in the AppSpec to me since it's calculated at the same time as the list of external client files. Calculating it on the fly is okay to me, but it feels redundant to calculate the same thing that won't change multiple times.

waspc/data/Generator/templates/react-app/vite.config.ts
infomiho: Yep, I would generate a `JsImport` in Haskell and pass it in here 👍

waspc/src/Wasp/AppSpec.hs
Martinsos: Sounds good! Calculating multiple times sounds a bit wasteful, but it is not really expensive and it would work quite well in this case I think -> however, I also like better that this logic, this knowledge of knowing where the path to custom vite config is, is present not in AppSpec, but in the logic that creates AppSpec. So great!

waspc/src/Wasp/AppSpec.hs
```diff
@@ -76,7 +77,7 @@ data AppSpec = AppSpec
     -- | Connection URL for a database used during development. If provided, generated app will
     -- make sure to use it when run in development mode.
     devDatabaseUrl :: Maybe String,
-    isCustomViteConfigUsed :: Bool
+    customViteConfigPath :: Maybe (Path' (Rel SourceExternalCodeDir) File')
```
Martinsos, line 80: You made it relative to SourceExternalCodeDir -> that is correct, I am just wondering, do we care if it is in external code dir or not? Because we could be less specific and say it is in source project dir (so root of source project). Then it could be in ext src, or not, who knows. But ok, this is also fine on one hand, especially if you are relying on this property of it being in ext src.

waspc/src/Wasp/Project/Vite.hs
```diff
@@ -0,0 +1,37 @@
...
+import StrongPath (File', Path', Rel, reldir, relfile, (</>))
+import qualified StrongPath.Operations as SP
+import Wasp.AppSpec.ExternalCode (SourceExternalCodeDir)
+import qualified Wasp.AppSpec.ExternalCode as ExternalCode
+import qualified Wasp.Generator.ExternalCodeGenerator.Common as ECC
+import qualified Wasp.Generator.WebAppGenerator.Common as C
+import qualified Wasp.Generator.WebAppGenerator.ExternalCodeGenerator as EC
+import Wasp.JsImport (JsImport, JsImportName (JsImportModule), makeJsImport)
+
+findCustomViteConfigPath :: [ExternalCode.File] -> Maybe (Path' (Rel SourceExternalCodeDir) File')
```
Martinsos, line 14: Yeah this is looking good now!

waspc/src/Wasp/Project/Vite.hs
```diff
@@ -0,0 +1,37 @@
...
+        { _pathInExtCodeDir = path
+        } = path == pathToViteTsConfig || path == pathToViteJsConfig
+
+    pathToViteTsConfig :: Path' (Rel SourceExternalCodeDir) File'
+    pathToViteTsConfig = [relfile|vite.config.ts|]
+
+    pathToViteJsConfig :: Path' (Rel SourceExternalCodeDir) File'
+    pathToViteJsConfig = [relfile|vite.config.js|]
+
+makeCustomViteConfigJsImport :: Path' (Rel SourceExternalCodeDir) File' -> JsImport
```
Martinsos, line 31: So this should probably go into Generator.
Some indications for this:
1. All the Generator imports in this module are used in this function, so it is heavily connected to Generator.
2. Creating that JS import -> that doesn't sound like general Project level thing, it sounds like something Generator needs.

There is also a bit of suspicious code in it -> I am not sure when that import is generated, where can it be used, in which JS file? Is it absolute, relative to something? If it was in the Generator, in `where` block of place where it is used, that would give it that context. This way as it is now, it feels pulled out of context. That `reldir|src|` is also tiny bit suspicious, feels like that shouldn't be hardcoded here but instead coming from some variable.
I think once you move this to Generator it will all sit in its place.

waspc/src/Wasp/AppSpec.hs
infomiho: Either we add the `src` part in the AppSpec creation part or we add it in the generator 🤷 I think it's a bit arbitrary and maybe just keep it like this for now? I don't really any good reasons

Martinsos: It's more about constraints -> is it important for us that custom vite config is in source external code dir? If so, then let's leave it as it is. If it doesn't matter, and we are ok with it being anywhere in the project, but for now we chose source external code dir, then I would go with path starting from the root of the project.

Martinsos (approved):
LGTM, approved! One comment is only left, the one with which StrongPath to use for the custom vite config, but I am ok with both options, so give it a read, do as you like, and feel free to merge.

waspc/data/Cli/templates/basic/src/client/vite.config.ts
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from 'vite'
+
+export default defineConfig({
+  server: {
+    open: true,
```
Martinsos, line 5: Ok, nice!

web/docs/project/custom-vite-config.md
```diff
@@ -0,0 +1,115 @@
...
+
+```ts title="src/client/vite.config.ts"
+import { defineConfig } from 'vite'
+
+export default defineConfig({
+  base: '/my-app/',
+})
+```
+</TabItem>
+</Tabs>
```
Martinsos, line 115: You should mention that they might break stuff if they over-customize it, because we expect some stuff in vite config to be configured in a specific way, so they should be careful.

Martinsos (approved):
@infomiho LGTM, but the main thing missing I think is to add a caution message that they can mess up things by doing this.

web/docs/project/custom-vite-config.md
infomiho: Makes sense, I'll add that part as well 👍

PR #1469, opened 2023-10-04
[MAGE] show modal based on prev state & small fixes

vincanger:
addressed all the changes @Martinsos mentioned in: https://github.com/wasp-lang/wasp/pull/1463


wasp-ai/src/client/pages/MainPage.jsx
Martinsos, line 1: As mentioned in old PR, it would be a good move to extract logic related to saving and loading the last generation into two separate functions: `locallySaveLastGenerationInput` and `locallyLoadLastGenerationInput`. It would make it more obvious, at the place of usage, what is happening.

wasp-ai/src/client/pages/MainPage.jsx
```diff
@@ -78,7 +78,7 @@ const MainPage = () => {
         localStorage.setItem("appNum", 0);
       }
       if (appNum === 2) {
```
Martinsos, line 80: Aha, we pop this up only on second generation -> makes sense!

Martinsos, line 80: `appNum` is not the best name -> maybe `numGenerationsSoFar` instead? Or `numGenerationsStarted`?

wasp-ai/src/client/pages/MainPage.jsx
```diff
@@ -78,7 +78,7 @@ const MainPage = () => {
         localStorage.setItem("appNum", 0);
       }
```
Martinsos, line 79: The logic above, where appNum is obtained from local storage and set to 0 if not defined, could be encapsulated into a function `getAppNum()` that would hide the details of how it is done.

Btw the logic is a bit weird here -> you fetch and parse appNum, but if it is 0, you set it to 0 in local storage -> so you are not just fetching, but also setting here, but only in a special case, a bit weird.
I think it is better to get it and that is it. If it is unset, you return 0 while getting it, but don't try to set it to 0 -> no need, if it is undefined we treat it as 0 anyway.

wasp-ai/src/client/pages/ResultPage.jsx
```diff
@@ -73,10 +74,16 @@ export const ResultPage = () => {
...
+    if (currentAppStatus === "in-progress") {
+      setIsRunning(true)
+    } else if (currentAppStatus === "success" && isRunning) {
+      setIsSuccessModalOpen(true)
+      setIsRunning(false)
+    } else if (isRunning) {
+      setIsRunning(false)
     }
-  }, [currentStatus])
+  }, [appGenerationResult?.project?.status])
```
Martinsos, line 86: While this probably does the trick, it feels a bit peculiar when reading it -> why are we setting is running together with modal? Why do we do it at all, when we have `appGenerationResult.project.status`?
We know the answer, but it is not obvious and this knowledge will get lost.

To be clear, this whole file is quite messy and not written with clean code (we were in hurry so we let it slide), but we have a bit of time now so it is good opportunity to at least try to get these new modifications as clean as possible.

Quick google search tells me that best thing to track and obtain old state is via useRef: https://stackoverflow.com/questions/53446020/how-to-compare-oldvalues-and-newvalues-on-react-hooks-useeffect -> so you introduce custom hook `usePrevious` it seems. Maybe that is the best way to go?
What you could possibly do then is have `prevGenerationResultStatus`, and then in `useEffect` check that to decide what you do with `setIsSuccessModalOpen`.

So instead of tracking `isRunning`, we track `prevGenerationResultStatus`. One might wonder why we track that, but they can easily figure that out, and they don't have to wonder what is correct or more correct: `isRunning` vs `appGeneration.project.status === 'running'`, and if they are in sync.
Also, our logic for dealing with `setIsSuccessModalOpen` will be much shorter and standalone, decoupled from logic of updating `prevGenerationResultStatus`.

wasp-ai/src/client/pages/ResultPage.jsx
```diff
@@ -406,59 +413,71 @@ function getCardinalNumber(number) {
...
+export function OnSuccessModal({ isOpen, setIsOpen, appGenerationResult }) {
+  const [numTokensSpent, setNumTokensSpent] = useState(0)
+  const { data: numTotalProjects } = useQuery(getNumProjects, {}, { enabled: isOpen })
 
   useEffect(() => {
-    if (!!tokenNumberStr) {
-      const num = tokenNumberStr.slice(1, -1) * 1000
-      setTokens(num)
+    const logText = appGenerationResult?.project?.logs?.find((log) => /tokens usage/i.test(log.content))?.content;
+    const tokenNumberStr = logText ? logText.split(":")[1]?.trim() : null;
```
Martinsos, line 422: Can't you combine these two into one regex?
If I am correct, you are looking for something like `tokens usage: 25k`. You can go after that with one regex: `tokens\s+usage\s*:\s*([0-9]+)\s*k\b`. This should allow for more freedom with whitespace, so we are more robust, and it also extracts the number right away (puts it into the match group), I am not sure how exactly it works in JS but should be a relatively straightforward way to extract that match group, getting exactly the number out of it.

Btw I often use this page to test out regexes: https://regex101.com/ .

wasp-ai/src/client/pages/ResultPage.jsx
```diff
@@ -406,59 +413,71 @@ function getCardinalNumber(number) {
...
+    if (tokenNumberStr) {
+      const num =  parseInt(tokenNumberStr.replace(/[\s~k]/gi, '')) * 1000;
+      if (num.toString().match(/^[0-9]+$/)) {
```
Martinsos, line 425: Hm unusual that this is how you would need to check if `num` is actual int, I would expect there to be a better way. But if you do change from comment above, this should also simplify / dissapear.

Martinsos (commented):
@vincanger looking good, we are getting close! Left a bit more comments, on making code a bit cleaner.

wasp-ai/src/client/pages/ResultPage.jsx
infomiho: Here's a piece of code as well

```js
const string = "📊 Total tokens usage: ~52.4k ";
const regex = /Total tokens usage: ~(\d+\.\d+?)k/;
const match = string.match(regex);

if (match) {
    const num = parseFloat(match[1]);
    console.log(num);
} else {
    console.log("No match found");
}
```

Martinsos: Nice! This won't work for numbers that don't have `.` in them, like `52k` though.

Here is another iteration on the regex:
```
Total\s+tokens\s+usage\s*:\s*~\s*(\d+(?:\.\d+){0,1})\s*k\b
```
1. I added `\s+` or `\s*` instead of actual space, to make it more robust to changes in whitespace.
2. I enclosed `\.\d+` in `(?: ){0,1}` which says this sequence can occur zero or one time.
3. I added `\b` at the very end, so `k` won't match the case where you have `~52.4kilos`, instead `k` has to be end of the word, no other letters or digits or something after it -> but spaces, newlines, punctuation and whatever are ok. `\b` stands for boundary.

vincanger:
whoops. this got neglecgted, but these issues were all addressed in other PRs

PR #1487, opened 2023-10-10
Added the commander in the npm package

TheOjasSingh:
Signed-off-by: Ojas Singh <ojas856@gmail.com>


infomiho:
Adding `node_modules` to the version control is not considered the best practice and we avoid it.

Your PR is missing a description where you:
- describe the issue you had
- how to reproduce it
- how you fixed it

Please add those details, so somebody from the team can recreate the issue and test your fix :)

PR #1504, opened 2023-10-13
Tutorial Fixes for v0.11.6

sodic:
What I did in this PR:
- [x] Went through the tutorial using JS and verified it works in development
- [x] Went through the tutorial using TS and verified it works development
- [x] Matched the app in `wasp/examples/tutorials/TodoApp` with the current version of the tutorial (read: removed clocks and other old stuff).
- [x] Created a TS variant of the tutorial app in `wasp/examples/tutorials/TodoApp`.
- [x] Fixed some errors in tutorial docs.
- [x] Point users to different repos depending on whether they're looking at JS or TS tutorial.
- [x] Successfully built both app for production


web/docs/tutorial/07-auth.md
```diff
@@ -505,7 +498,16 @@ This is it, we have a working authentication system, and our Todo app is multi-u
...
+<ShowForJs>
+
+You can find the complete code for the JS version of the tutorial [here](https://github.com/wasp-lang/wasp/tree/release/examples/tutorials/TodoApp).
+
+</ShowForJs>
+<ShowForTs>
+
+You can find the complete code for the TS version of the tutorial [here](https://github.com/wasp-lang/wasp/tree/release/examples/tutorials/TodoAppTs).
+
+</ShowForTs>
```
sodic, line 510: You might have noticed we now have two todo ts apps:
- This one that mimics the tutorial.
- Vince's todo-typescript app.
- 
They're slightly different. Should we keep both or remove one?

web/docs/tutorial/07-auth.md
```diff
@@ -122,18 +122,17 @@ Great, Wasp now knows these pages exist! Now, the React code for the pages:
 
 ```jsx title="src/client/LoginPage.jsx"
 import { Link } from 'react-router-dom'
-
 import { LoginForm } from '@wasp/auth/forms/Login'
 
 const LoginPage = () => {
   return (
-    <>
+    <div style={{ maxWidth: '400px', margin: '0 auto' }}>
```
sodic, line 129: The login and signup screens were too wide.

web/docs/tutorial/07-auth.md
Martinsos: Let's remove @vincanger 's one? Is there anything we loose by that? If not, let's remove it.

Martinsos (approved):
@sodic looking good!

PR #1512, opened 2023-10-16
Make example app names more descriptive

sodic:


Martinsos:
@sodic oh wow I saw this change only now, when I was trying to get into `thoughts` example and I couldn't find it!
Not a big fan of this change, thoughts is not just about database-seeding, nor do I think changing a name here helps much.

PR #1516, opened 2023-10-17
Adds support for client subdir

### LINKED ISSUE #8: Add support in Wasp for hosting the frontend in non-root directory
Right now, we successfully hosted our todoApp example, generated in Wasp, at examples.wasp-lang.dev.

However, if we would want to host it at examples.wasp-lang.dev/todoApp (for example), we could not, because with current configuration it can only be served from the root directory of the domain.

We need to figure out what is needed to get this working, and add support for this in Wasp!

infomiho:
Closes #8

You can now serve the client from a subdirectory. This is useful if you want to serve the client from a subdirectory of your domain, e.g. `https://example.com/my-app/`.

To do this, you need to add the `client.baseDir` property to your `.wasp` file:

```wasp
app todoApp {
  // ...
  client: {
    baseDir: "/my-app",
  },
}
```


waspc/examples/subdir-example/main.wasp
```diff
@@ -0,0 +1,14 @@
+app subdirExample {
+  wasp: {
+    version: "^0.11.6"
+  },
+  title: "subdir-example",
+  client: {
+    baseDir: "/bla"
```
infomiho, line 7: Example of usage

waspc/src/Wasp/AppSpec/App/Client.hs
```diff
@@ -10,6 +10,7 @@ import Wasp.AppSpec.ExtImport (ExtImport)
 
 data Client = Client
   { setupFn :: Maybe ExtImport,
-    rootComponent :: Maybe ExtImport
+    rootComponent :: Maybe ExtImport,
+    baseDir :: Maybe String
```
Martinsos, line 14: It would be nice if this was something like "Url" instead of "String", but probably best not to get into it now :D.
Although hm, what do we know about this String, about what it should look like? I guess it needs to be absolute path and that is it? We don't allow stuff like `bla` (relative) or `/bla?x=1` (query params) or `/bla#summary`, right? If that is so, maybe it might make sense to actually roll with StrongPath here that is Abs Dir! I can't remember if we can embed it directly into our AppSpec though, will our TemplateHaskell magic know how to translate this into Wasp DSL.

What I would do is maybe give it a try -> try to make it StrongPath Abs Dir -> if that works out of the box, great, we got some extra safety here. If not, forget it, but you can maybe put a comment next to it explaining what we expect it to look like.

If this works, you can also remove that validation you have later.

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -364,6 +366,18 @@ validatePrismaOptions spec =
...
+  Just baseDir | startsWithSlash baseDir -> []
+  Nothing -> []
+  _anyOtherCase -> [GenericValidationError "The app.client.baseDir should start with a slash e.g. \"/test\""]
+  where
+    maybeBaseDir = Client.baseDir =<< AS.App.client (snd $ getApp spec)
+
+    startsWithSlash :: String -> Bool
+    startsWithSlash (x : _) = x == '/'
+    startsWithSlash _ = False
+
```
Martinsos, line 380:
```suggestion
validateWebAppBaseDir :: AppSpec -> [ValidationError]
validateWebAppBaseDir spec = case maybeBaseDir of
  Just baseDir | not (startsWithSlash baseDir) ->
    [GenericValidationError "The app.client.baseDir should start with a slash e.g. \"/test\""]
  _anyOtherCase -> []
  where
    maybeBaseDir = Client.baseDir =<< AS.App.client (snd $ getApp spec)

    startsWithSlash :: String -> Bool
    startsWithSlash ('/' : _) = True
    startsWithSlash _ = False
```

waspc/src/Wasp/Project/WebApp.hs
```diff
@@ -1,9 +1,19 @@
...
+import qualified Wasp.AppSpec.App.Client as AS.App.Client
 import Wasp.AppSpec.ExternalCode (SourceExternalCodeDir)
+import Wasp.AppSpec.Valid (getApp)
 
 data StaticAssetsDir
 
 staticAssetsDirInExtClientCodeDir :: Path' (Rel SourceExternalCodeDir) (Dir StaticAssetsDir)
 staticAssetsDirInExtClientCodeDir = [reldir|public|]
+
+getBaseDir :: AppSpec -> String
```
Martinsos, line 16: This is a bit peculiar, I would off the bat say that this doesn't belong here.
`Project` is about the user's source project, their code files and everything, then that project is analyzed and AppSpec is produced, and then Generator uses AppSpec to produce final code.

So this is now Project/WebApp answering questions about where the baseDir is, and you also use that in the Generator -> yeah, a bit weird. ALso a bit weird that Generator needs to import Project. I would rather stick to AppSpec here I think. So if you really want this helper function, I would put it in AppSpec somewhere, or maybe even in Generator, but I don't think I would put it into Project.

Martinsos (approved):
Reviewed, looks good!

I left some smaller comments, feel free to fix them as you wish and you can merge when you are satisfied.

Some additional questions, but also not critical:
1. How did you choose `baseDir` for a name? I see Vite uses `base`, react-router uses `basename`. I like `baseDir` the most probably, I think it is the most precise, I am just curious.
2. Idea: what if they need to use the value of "baseDir" in frontend/backend? I can see that being possible? Should we expose it for them, so they can just import it and use it, since Wasp already knows it? Instead of them duplicating it in `shared/`?

infomiho:
Note: check out the email sending code that relies on the frontend URL for email verification and password reset. URL in the config probably needs updating. Also the docs need to be updated.

Martinsos:
> Note: check out the email sending code that relies on the frontend URL for email verification and password reset. URL in the config probably needs updating. Also the docs need to be updated.

note: docs should be updated to mention that frontend URL needs to contain that same base dir in it. I wonder if we could even do a check for that in Wasp.

waspc/src/Wasp/AppSpec/App/Client.hs
infomiho: > will our TemplateHaskell magic know how to translate this into Wasp DSL.
Sadly, it doesn't seem to be the case 🤷 
```
No translation to wasp type for type AppT (AppT (AppT (ConT StrongPath.Internal.Path) (ConT StrongPath.Internal.Posix)) (ConT StrongPath.Internal.Abs)) (ConT StrongPath.Internal.Dir')
```

> If this works, you can also remove that validation you have later.

If the validation fails, the error will not be something specific, but rather a generic StrongPath Abs Dir parsing error which will confuse users IMHO

Martinsos: Ah ok, so we don't have a translation for that in our Evaluator.

I think if we had that working properly for StrongPath, that error toward user could be nice, it would basically tell them something like "this should be an absolute path to a dir and it is not", so I don't think it would confuse users.

But ok, since we don't have it right now, let's roll with validation check!

waspc/src/Wasp/Project/WebApp.hs
Martinsos: @infomiho bump on this one

waspc/data/Generator/templates/react-app/vite.config.ts
```diff
@@ -15,7 +15,7 @@ const defaultViteConfig = {
   base: "{= baseDir =}",
   plugins: [react()],
   server: {
-    port: 3000,
+    port: {= defaultClientPort =},
```
Martinsos, line 18: Interesting, why did we extract this?

waspc/src/Wasp/Project/WebApp.hs
infomiho: Moved it to the `Generator.WebAppGenerator.Common` 👍 

I misunderstood the `Project` module

waspc/src/Wasp/Generator/WebAppGenerator.hs
```diff
@@ -342,7 +342,8 @@ genViteConfig spec = return $ C.mkTmplFdWithData tmplFile tmplData
     tmplData =
       object
         [ "customViteConfig" .= jsImportToImportJson (makeCustomViteConfigJsImport <$> AS.customViteConfigPath spec),
-          "baseDir" .= WebApp.getBaseDir spec
+          "baseDir" .= WebApp.getBaseDir spec,
```
Martinsos, line 345: I commented on this before, but same as for baseDir, I don't think deafultClientPort should be in `WebApp`.

waspc/src/Wasp/Project/WebApp.hs
```diff
@@ -17,3 +17,9 @@ getBaseDir :: AppSpec -> String
 getBaseDir spec = fromMaybe "/" maybeBaseDir
   where
     maybeBaseDir = AS.App.Client.baseDir =<< AS.App.client (snd $ getApp spec)
+
+defaultClientPort :: Int
+defaultClientPort = 3000
+
+getDefaultClientUrl :: AppSpec -> String
+getDefaultClientUrl spec = "http://localhost:" ++ show defaultClientPort ++ getBaseDir spec
```
Martinsos, line 25: Why extract this? Are there other places that assume this and we should also maybe extract it?

web/docs/project/_baseDirEnvNote.md
```diff
@@ -0,0 +1,6 @@
+:::caution Setting the correct env variable
+
+When you set the `baseDir` option, make sure that the `WASP_WEB_CLIENT_URL` env variable also includes the base directory.
+
+For example, if you are serving your app from `https://example.com/my-app`, the `WASP_WEB_CLIENT_URL` should be set to `https://example.com/my-app`.
```
Martinsos, line 5: in first line: "When you set" -> "If you set"
in second line: "the base" -> "that base"
third line: I added a counter example.

```
If you set the `baseDir` option, make sure that the `WASP_WEB_CLIENT_URL` env variable also includes that base directory.

For example, if you are serving your app from `https://example.com/my-app`, the `WASP_WEB_CLIENT_URL` should be also set to `https://example.com/my-app`, and not just `https://example.com`.
```

waspc/src/Wasp/Project/WebApp.hs
Martinsos: You haven't pushed this changes yet though, have you? Because I don't see them here on the PR.

Martinsos:
> Note: check out the email sending code that relies on the frontend URL for email verification and password reset. URL in the config probably needs updating. Also the docs need to be updated.

How are with the email then, all good?

Martinsos:
> Reviewed, looks good!
> 
> I left some smaller comments, feel free to fix them as you wish and you can merge when you are satisfied.
> 
> Some additional questions, but also not critical:
> 
> 1. How did you choose `baseDir` for a name? I see Vite uses `base`, react-router uses `basename`. I like `baseDir` the most probably, I think it is the most precise, I am just curious.
> 2. Idea: what if they need to use the value of "baseDir" in frontend/backend? I can see that being possible? Should we expose it for them, so they can just import it and use it, since Wasp already knows it? Instead of them duplicating it in `shared/`?

@infomiho bump on this one, some questions here!

infomiho:
1. The `baseDir` name is arbitrary. It felt right given the `base` and `basename` nomenclature.
2. I wouldn't add it unless we get explicit requests for it, keeping it simple for now. `baseDir` is already an edge case, so exposing it an even more of an edge case IMHO
3. Email links now work as expected 👍 I've tested it out locally

waspc/src/Wasp/Project/WebApp.hs
infomiho: I wanted to construct the `defaultClientUrl` in Haskell and not piece it together in the template. 

That required the extraction of the default port as well.

web/docs/project/_baseDirEnvNote.md
infomiho: @Martinsos 
Screenshot 2023-10-25 at 11 49 42

Martinsos: I am aware of this feature :D, but it isn't always offered -> sometimes I can add a suggestions, sometimes I can't. This time I couldn't, so therefore I opted for this.

Reminder: don't forget to do our github review thing -> leave an emoji or comment so other person knows that you saw / acknowledge / solved the comment.

Martinsos:
> 1. The `baseDir` name is arbitrary. It felt right given the `base` and `basename` nomenclature.
> 2. I wouldn't add it unless we get explicit requests for it, keeping it simple for now. `baseDir` is already an edge case, so exposing it an even more of an edge case IMHO
> 3. Email links now work as expected 👍 I've tested it out locally

Adding `baseDir` in SDK -> I don't see why that is more of an edge case, it seemed logical to me to give them a way to access that value they defined in .wasp file. Otherwise they will have to duplicate it. They might need it to construct some kind of URL maybe?

waspc/data/Generator/templates/react-app/vite.config.ts
infomiho: Answered here: https://github.com/wasp-lang/wasp/pull/1516#discussion_r1371466384

infomiho:
I don't feel we have a strong argument for adding it. 

> Otherwise they will have to duplicate it. They might need it to construct some kind of URL maybe?

When users start mentioning they need it -> I would then add it. Otherwise, we are add things we _assume_ could be useful .

waspc/src/Wasp/Generator/WebAppGenerator/Common.hs
```diff
@@ -123,3 +130,14 @@ toViteImportPath :: Path Posix (Rel r) (File f) -> Path Posix (Rel r) (File f)
 toViteImportPath = fromJust . SP.parseRelFileP . dropExtension . SP.fromRelFileP
   where
     dropExtension = fst . splitExtension
+
+getBaseDir :: AppSpec -> String
```
Martinsos, line 134: I would make this one return Strong Path, (Posix) Abs Dir.

Martinsos (approved):
@infomiho approved! There might be some comments that are still somewhat unresolved, pls make sure those are all in good shape, and then you can merge!

waspc/src/Wasp/Generator/ServerGenerator/ConfigG.hs
```diff
@@ -22,8 +22,9 @@ genConfigFile spec = return $ C.mkTmplFdWithDstAndData tmplFile dstFile (Just tm
...
       object
         [ "isAuthEnabled" .= isAuthEnabled spec,
-          "databaseUrlEnvVarName" .= databaseUrlEnvVarName
+          "databaseUrlEnvVarName" .= databaseUrlEnvVarName,
+          "defaultClientUrl" .= WebApp.getDefaultClientUrl spec
         ]
 
-configFileInSrcDir :: Path' (Rel C.ServerSrcDir) File'
-configFileInSrcDir = [relfile|config.js|]
+    configFileInSrcDir :: Path' (Rel C.ServerSrcDir) File'
```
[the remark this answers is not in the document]
infomiho: I saw that it was used only in one place, therefore I moved it inside of the `genConfigFile`

waspc/src/Wasp/Generator/WebAppGenerator/Common.hs
infomiho: Good idea 👍

waspc/src/Wasp/Generator/ServerGenerator/ConfigG.hs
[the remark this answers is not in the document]
Martinsos: I would say it is not about how it is used currently, it is about how it is intended to be used. We keep all of such paths at top level (if I am correct), because they are often by other modules for imports and similar. You putting into `where` indicates you mean it to be private/local for this function, so if I want to use it later, I will have to wonder if I can use that one -> it seems like the author didn't intend so.

So I would leave where it was.

Martinsos:
> I don't feel we have a strong argument for adding it.
> 
> > Otherwise they will have to duplicate it. They might need it to construct some kind of URL maybe?
> 
> When users start mentioning they need it -> I would then add it. Otherwise, we are adding things we _assume_ could be useful .

Fair enough -> personally, to me it just makes a lot of sense to have it, and at the end we are always assuming that somebody will use something :D, the question is just how certain are we of it. What is tricky here is that people might not ask for it but just duplicate it. But ok, let's not get stuck on it, we can certainly add it later as you said, hopefully we will be the ones to stumble upon needing it first, if it is indeed useful.

waspc/src/Wasp/Generator/ServerGenerator/ConfigG.hs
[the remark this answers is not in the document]
infomiho: Makes sense 👍

Issue #1342, opened 2023-07-24
Fly by default creates 2 machines for server - document or change that?


Martinsos:

Martinsos:
They create 2 machines for server, which is their recommended minimum, so you have availability while deploying new version of server and similar.
But, I do wonder if this is what people expect -> it certainly surprised me quite a bit.
Suggestions:
1. Mention this in docs, to warn people.
2. Consider having `wasp` by default create only one server machine (but then again we should inform users about this choice).

Martinsos:
What does it even mean on Fly that 2 machines are used for server -> are they on at the same time? They have this thing where they shut down machines when they are not used -> so I guess only one of them is on normally? What if both are on? Is there some load balancing happening? Are they both used sometimes? In that case that might be an issue, if Wasp server is written in such a way that doesn't assume multiple server instances to be running (e.g. has some state).

infomiho:
Here's Fly's bit of docs on the topic https://fly.io/docs/reference/app-availability/#redundancy-by-default-on-first-deploy

Martinsos:
I discussed this a bit more with @infomiho .
What we will probably want to do is make it so by default 1 instance/machine of everything is created, and then inform user how they can bump it to 2 if they want.

PR #1535, opened 2023-10-25
Sets the minimum number of Fly machines to 1

### LINKED ISSUE #1466: Update our fly.toml files to have correct minimum number of instances running

infomiho:
Closes #1466

It updates the `fly-server.toml` file by replacing a line. The technique of replacing the line has already been used to setup some other things in the client TOML file.


waspc/packages/deploy/src/providers/fly/setup/setup.ts
```diff
@@ -61,6 +61,8 @@ async function setupServer(deploymentInfo: DeploymentInfo<SetupOptions>) {
 	// This creates the fly.toml file, but does not attempt to deploy.
 	await $`flyctl launch --no-deploy ${launchArgs}`;
 
+	replaceLineInLocalToml(/min_machines_running = 0/g, 'min_machines_running = 1');
```
Martinsos, line 64: Might be a bit fragile in case that line `min_machines_running = 0` changes, for example gets some extra white spaces, or that `0` becomes `2` or something else?

In that case, we will just silently fail here, right?

How do we know it will be 0 initially btw, and that this line will even be defined in the file, what if the line is missing?

If solving any of these concerns is too hard and they are not likely, ok, we can leave it as it is, but I would like to understand the situation.

Martinsos:
All right, nice!

Client -> it will have 0 minimum instances -> yeah, that makes sense to me.

Server -> we set it to 1 -> that makes sense, that way we are sure jobs and other stuff works.
What if they want to set it to 0 though? Would it be hard? Do we have any instructions for that? Should we explain this to them?

Db -> it will have 0 minimum instances -> I guess that is ok, right? Could that cause any issues? Connection dropping between server and db maybe?

Btw, while this is separate issue, I thought this might be a good moment to also give a bit of thought to this one: https://github.com/wasp-lang/wasp/issues/1342 .

infomiho:
### This PR

> What if they want to set it to 0 though?

They can edit the `fly-server.toml` and redeploy. These are just the initial values which we set once and then they can do whatever they want to the Fly's TOML file.

> Do we have any instructions for that?

We do not have explicit instructions for editing the TOML files. We only mention that they should include them their version control: https://wasp-lang.dev/docs/advanced/deployment/cli#setup

> Should we explain this to them?

We should 👍 

> Db -> it will have 0 minimum instances -> I guess that is ok, right? Could that cause any issues? Connection dropping between server and db maybe?

In our setup, the minimum for the DB is 1 and we don't use their "Scale to zero" feature for Postgres. https://fly.io/docs/postgres/managing/scale-to-zero/

### The two machines issue

There is an option to disable HA -> https://fly.io/docs/reference/app-availability/#turn-off-redundancy-on-deploy 
We should address it in a separate PR.

waspc/packages/deploy/src/providers/fly/setup/setup.ts
infomiho: This will silently fail if the line changes or it is missing.

I think this is an okay solution for now 🤷‍♂️

We would ideally upgrade our deployment script here to parse the TOML files and editing the parsed object and writing it back to TOML. But this only protects us from content changes, not from Fly API changes -> if they rename the field completely then we are still screwed :D

Martinsos: I really don't like silent fails, not when we are trying to make something work a certain way.

So I would suggest that we instead fail loudly here :D. At least we will know it is not behaving as we want it to, and we can then fix it, instead of finding out some time later that things are not happening as we expect them to happen (e.g. that people have servers shutting down when they should be running jobs).

infomiho: > So I would suggest that we instead fail loudly here :D

Imagine the following:
- we implement this so it fails loudly, I guess meaning "stop the deployment" when this line doesn't exist
- everything is good
- imagine then that Fly releases a new version of the TOML file (which they will do at some point)
- all of Wasp users from that point on **can't deploy stuff to Fly via CLI** until we release a new Wasp version

That's why I vote for a silent failure. If the TOML file changes, oh well, users will figure it out at some point. We are just using Fly CLI to deploy stuff, we can't guarantee the setup since we depend on them and their changes.

Martinsos:
So regarding instructions to users about how we set min number of machines and that they can set it to 0 if they don't have any Jobs / long running stuff -> you said above that we should provide them with those instructions -> let's do it in this PR, while we are at it? I guess we should put this info into docs? But I think it is also interesting to put this info into output of `wasp deploy fly`.

Database scaling down, two machines -> ok yeah, we can tackle those separately.

Anything else needed to finalize this PR?

waspc/packages/deploy/src/providers/fly/helpers/helpers.ts
```diff
@@ -38,8 +42,11 @@ function getWaspBuildDir(waspProjectDir: string) {
 	return path.join(waspProjectDir, '.wasp', 'build');
 }
 
-export function ensureDirsInCmdAreAbsoluteAndPresent(thisCommand: Command): void {
-	const waspProjectDirPath: string | undefined = thisCommand.opts().waspProjectDir;
+export function ensureDirsInCmdAreAbsoluteAndPresent(
+	thisCommand: Command,
```
Martinsos, line 46: Oh wow we are using tabs here for indentation? I see now they were also used before? Uff.

waspc/packages/deploy/src/providers/fly/setup/setup.ts
```diff
@@ -61,7 +84,23 @@ async function setupServer(deploymentInfo: DeploymentInfo<SetupOptions>) {
...
+
+	if (!doesLocalTomlContainLine(minMachinesOptionRegex)) {
+		await question(`⚠️  There was an issue setting up your server app.
+We tried modifying your server fly.toml to set ${boldText(
+		'min_machines_running = 1',
+	)}, but couldn't find the option ${boldText(
+		'min_machines_running',
+	)} in the fly.toml.
+
+This might mean your server app might not be always running,
```
Martinsos, line 97:
```suggestion
We advise that you additionaly check what is the value for "minimal number of machines running" on Fly for this server app and confirm that it is set to the value you are ok with.

Be aware that if it is set to 0, your server will shut down when there are no requests from the client, which might be an issue for you if you have recurring Jobs or some other processes that need to keep running on the server even without external input, in which case we advise keeping "minimal number of machines running" setting at a number larger than zero.

Contact Wasp Team for any help!
```

waspc/packages/deploy/src/providers/fly/setup/setup.ts
```diff
@@ -61,7 +84,23 @@ async function setupServer(deploymentInfo: DeploymentInfo<SetupOptions>) {
 	// This creates the fly.toml file, but does not attempt to deploy.
 	await $`flyctl launch --no-deploy ${launchArgs}`;
 
-	replaceLineInLocalToml(/min_machines_running = 0/g, 'min_machines_running = 1');
+	const minMachinesOptionRegex = /min_machines_running = 0/g;
+
+	if (!doesLocalTomlContainLine(minMachinesOptionRegex)) {
+		await question(`⚠️  There was an issue setting up your server app.
```
Martinsos, line 90:
```suggestion
		await question(`⚠️  There was a possible issue setting up your server app.
```

waspc/packages/deploy/src/providers/fly/setup/setup.ts
```diff
@@ -105,9 +146,24 @@ async function setupClient(deploymentInfo: DeploymentInfo<SetupOptions>) {
...
-	// assumes port 8080 (or 3000, depending on flyctl version).
-	replaceLineInLocalToml(/internal_port = \d+/g, 'internal_port = 8043');
+	const internalPortOptionRegex = /internal_port = \d+/g;
+
+	if (!doesLocalTomlContainLine(internalPortOptionRegex)) {
+		await question(`⚠️  There was an issue setting up your client app.
+We tried modifying your client fly.toml to set ${boldText(
+		'internal_port = 8043',
+	)}, but couldn't find the option ${boldText(
+		'modifying',
```
Martinsos, line 156:
```suggestion
		'internal_port',
```

waspc/packages/deploy/src/providers/fly/setup/setup.ts
```diff
@@ -105,9 +146,24 @@ async function setupClient(deploymentInfo: DeploymentInfo<SetupOptions>) {
...
+	)} in the fly.toml.
+
+This might mean your client app might not be accessible,
```
Martinsos, line 159:
```suggestion
This means your client app might not be accessible,
```

Martinsos: And maybe also mention here that we advise they reach out to us.

Martinsos: I mean ideally we would be notified in case any of these issues happens, automatically. To achieve something like that, we could possibly send some message to our analytics, although they are not really meant for that, what would probably make more sense is to use something like Sentry or similar. But maybe telling them to let us know is good enough for now.

waspc/packages/deploy/src/providers/fly/helpers/tomlFileHelpers.ts
```diff
@@ -56,18 +56,30 @@ export function getAppNameFromToml(path: string): string {
...
-export function replaceLineInLocalToml(searchValue: string | RegExp, replaceValue: string): void {
+export function replaceLineInLocalToml(
+	searchValue: string | RegExp,
+	replaceValue: string,
+): void {
 	const content = fs.readFileSync('fly.toml', 'utf8');
 	const updatedContent = content.replace(searchValue, replaceValue);
 	fs.writeFileSync('fly.toml', updatedContent);
 }
+
```
Martinsos, line 81: Instead of implementing a separate function, I would probably instead have `replaceLineInLocalToml` return or throw an error, and would then handle that and ask a question based on that. But ok, this also works (we just read the file twice and have a bit of duplication here).

Martinsos (changes requested):
@infomiho nice! Some changes are required: the main thing is I suggested some wording changes -> it is better to be redundant in this case and offer a bit longer explanation, then leave them wondering what this is about. I also wanted to let them know to reach out to us, so we learn about these issues.

infomiho:
I've adjusted the message 👍

PR #1541, opened 2023-10-26
fix: small typo and home button behavior

### LINKED ISSUE #1537: Change websockets example app home button behavior (Release branch)
Right now, clicking the logo of the [websockets example app](https://websockets-client-production.up.railway.app/login ) redirects to the flowbite react site. Not sure if this is intended but it seems strange when it's supposed to be a wasp example. I think it should be changed to regular behavior of redirecting to the webpage itself or even to the wasp site. There is also a small typo in the description of the app in the docs. If this fix sounds appropriate, I can create a pull request to address it.

gmirzayev:
### Description

Fixed a small typo in description of example app.

Changed the behavior of the home button in the realtime voting websockets example app to not go to external website. Other example apps don't do anything on click of home button so I did the same. Just changed cursor on hover to make it look like a button.

Fixes #1537

### DIFF
--- a/examples/websockets-realtime-voting/src/client/Layout.jsx
+++ b/examples/websockets-realtime-voting/src/client/Layout.jsx
@@ -11,7 +11,7 @@ const customTheme = {
     color: {
       primary: "bg-red-500 hover:bg-red-600",
     },
-  },
+  }
 };
 
 export const Layout = ({ children }) => {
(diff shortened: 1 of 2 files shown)


Martinsos (approved):
Nice @gmirzayev , thanks for the fix!

PR #1544, opened 2023-10-27
Fixes a Prisma SSL bug by bumping Prisma

### LINKED ISSUE #1543: Prisma crashes on Railway when SSL is used
I've had to debug a user's application that used Google login. When the Prisma code wanted to save the newly logged in user to the database the server would crash with **seg fault**.

image

Adding `?sslmode=disable`  to the `DATABASE_URL` fixed the situation. This looked like a classic Prisma SSL issue.

### The fix we should implement

This fixed the issue when I debugged the problem:
- Bumping the Prisma from `4.12.0` (`4.13.0` includes a fix for SSL behaviour https://github.com/prisma/prisma/issues/17946#issuecomment-1511416351) to `4.16.2`
- Removing line from `Dockerfile` installing OpenSSL 1.1

infomiho:
Closes #1543

Prisma had an issue with SSL in the `4.12.0`, they fixed it in `4.13.0`. I've bumped the version to the latest `4.X.X` and removed the `OpenSSL 1.1.` installation line from the `Dockerfile` so we would use `OpenSSL 3.0`.


Martinsos (approved):
@infomiho nice job!

I would, in case you haven't, to check all the changes that happened with Prisma from 4.12 to 4.16, in case there is something we should adapt to when doing this update.

Issue #1515, opened 2023-10-17
Code blocks in blog re-render to be single line when hovering logo


Sleavely:
I was reading your recent blog post and had trouble following along until I realized code blocks are shrunk because I accidentally hovered my mouse over the logo at some point:

![](https://i.imgur.com/fxfFnks.gif)

...this is probably the weirdest bug I've encountered today. 😂

Martinsos:
> Oh wow ok this is crazy @Sleavely :D. I agree, this is the craziest bug I have seen for some time :D!
> 
> Thanks for reporting this -> I am though having a bit of problems replicating it. I just tried it on Chrome on Linux (Gnome on Xorg) and this problem doesn't seem to happen. Could you please share your setup, and any additional details you think might help with replicating it?

@Sleavely I take that back, I just managed to replicate it! I am not sure why it wouldn't happen the first time, but now I got it happening. 
Seems that newlines get ignored. We are using Docusaurus, I wonder if this is somehow connected to them, or if it some of our custom code misbehaving. This will be interesting to debug!

Martinsos:
To make it even weirder, I figured out why I couldn't replicate it at first.
So, if you come to the blog post by clicking on it on the https://wasp-lang.dev/blog page, then this bug doesn't manifest.
If you refresh the blog post, or if you come directly to it via its direct URL, then the bug manifests itself!

khanjanSanghavi:
Hey, @Martinsos  I want to give it a shot!! 
You learn something new about this bug?
or
Can you say what might be creating this kind of crazy bug?

Martinsos:
> Hey, @Martinsos I want to give it a shot!! You learn something new about this bug? or Can you say what might be creating this kind of crazy bug?

Yes sure go for it, but I have to be honest I have no idea! That is what makes it quite fun :D.

Our docs are implemented via Docusaurus -> so maybe it is something Docusaurus related? But I somehow doubt it. So after mouse goes over that logo, something changes permanently. But why would something change permanently just upon hovering over something? Maybe if it changed temporarily, but permanently? Super weird. Almost sounds like we are triggering some kind of bug not caused by us?
Also, why would newlines get messed up? Should we look into the component we have for rendering the code -> I am not sure what we are using there -> and see if whatever we are using has similar bug documented on their github?

Tricky!

Martinsos:
Fixed by #1552

PR #1552, opened 2023-10-31
fix: stop code blocks shrinking on logo hover

gmirzayev:
### Description

Fixes issue #1515 
Not really sure why the bug happens exactly but it's something related to docusaurus and local href for logo button in config. The docusaurus config determines the href and it was not set so the default is siteConfig.baseUrl. Changing it to be an "external" url and then setting target to "_self" so it does not open in new tab fixes the bug and keeps the same behavior. 

Initially for this pull request, I removed the Link tag around the Logo element in Nav/index.js. However, I don't think you need the Logo element because when I removed it, nothing changed. However, since I may be missing something, I decided to add the Link tag back and just commit the bug fix itself. 

### DIFF
--- a/web/docusaurus.config.js
+++ b/web/docusaurus.config.js
@@ -51,6 +51,8 @@ module.exports = {
       logo: {
         alt: "Wasp logo",
         src: "img/wasp-logo-eqpar-circle.png",
+        href: "https://wasp-lang.dev/",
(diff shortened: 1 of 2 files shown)


Martinsos:
Awesome @gmirzayev , that is a nice catch! Merging it, thanks!

Issue #1489, opened 2023-10-10
Prettier formatting, .prettierignore, import orders


martinovicdev:
A suggestion for improving the code readability in the web page/docs section of the project: 

1. Create a .prettierignore file, ignoring files such as package.json, .MD files, etc. so they don't get formatted automatically.
2. Add plugin for ordering imports, so they are consistent (for example https://github.com/IanVS/prettier-plugin-sort-imports)
3. Add automatic Tailwind class sorting - https://tailwindcss.com/blog/automatic-class-sorting-with-prettier
4. _Optional:_ Add additional parameters to prettier.config.js, such as `endOfLine: 'lf'`, `tabWidth: 2,` to keep the code more consistent between different OS's and different configurations
5. Run the `prettier --write .` so the whole codebase is evenly formatted

Current situation is that we have the prettier.config.js file set up, but whole project isn't formatted yet. This could cause a bit of confusion for new contributors + we would want the code to be evenly formatted anyways.

I'd be happy to do it myself.

heysagnik:
I want to work on this issue

Martinsos:
@martinovicdev that sounds good, and if you want to go for it that would be amazing! Should we also make prettier a part of CI, or have it as a git commit hook, or something like that, to ensure it is always applied?

martinovicdev:
Thanks @Martinsos I'll do it 😄. 

I would suggest putting code formatting (Prettier) in a pre-commit hook and/or inside a check for PR, where it would fail if docs aren't formatted as they should be.
I think that having it inside CI wouldn't be ideal, as code changes could potentially happen and you have no control over them.

Martinsos:
> Thanks @Martinsos I'll do it 😄.
> 
> I would suggest putting code formatting (Prettier) in a pre-commit hook and/or inside a check for PR, where it would fail if docs aren't formatted as they should be. I think that having it inside CI wouldn't be ideal, as code changes could potentially happen and you have no control over them.

That makes sense to me, sounds good!

PR #1557, opened 2023-11-02
Prettier config

martinovicdev:
### Description

PR in relation to issue #1489. I have introduced .prettierrc file instead of prettier.config.js due to extension not working properly with the .js file. .prettierignore file has the exceptions for things such as Markdown files, static folder and .docusaurus folder. 


Martinsos:
Nice one @martinovicdev ! Sorry for taking a bit to take a look at this one, I will not be more responsive.

Nice that we are going to have prettier now, that is great!
I guess you had to make it work with docusaurus so that is that plugin for -> makes sense.

What could be cool is if we also added prettier to pre-commit hooks. I did it for another project couple of days ago, by following instructions in prettier's docs, it was quite easy -> but maybe it would be harder here due to docusaurus? It used stuff like `husky` and `lint-stage`. If this is interesting, maybe we could add it to this PR? Or we can leave it for later.
Oh and if we are doing this I would also add a command for formatting everything, something like `npm run format` that will really just do `prettier . --write`, so that even folks who don't have prettier set up in their editor can contribute.

martinovicdev:
I suggest rebasing and adding `npm run format` to the code for now. I am open to making precommit hooks in a separate PR, as they are a new thing that we're adding in the workflow.

Martinsos:
I see you did some changes but it has redundant commits and is still in conflict -> let me know if any help is needed!

Martinsos:
Awesome @martinovicdev , this is looking good, merging! Thanks!

PR #1558, opened 2023-11-03
Adjust the DB check error message

### LINKED ISSUE #1435: Fix database initialization
Wasp no longer knows how to initialize (create) the database for a project.

Before:
1. Set the database URL
2. Run `wasp db migrate-dev` or `wasp start`
3. Wasp create the database and start successfully.

Now:
1. Set the database URL
2. Run `wasp db migrate-dev` or `wasp start`
3. Wasp tells you that the database doesn't exist and does not create it.

image

This is a recent regression. @Martinsos  knows (or he had a feeling) which feature caused it (something Craig implemented IIRC).

infomiho:
Closes #1435

We had a regression with the way we handle **DB server** being available but the **DB** not existing.

1. **Before** the DB check was introduced
    - we wouldn't check **DB server** connectivity
    - Prisma would create the **DB** if missing
2. **After** the DB check was introduced
    - we check for **DB server** connectivity by trying to execute a dummy SQL command on the **DB**
    - Wasp fails with an error "DB server is not running"
3. After the fix
    - we check for **DB server** connectivity by trying to execute a dummy SQL command on the **DB**
    - if the error is "DB doesn't exist" -> we tolerate it, _since we know Prisma will create the DB if missing_
    - in all other cases -> Wasp fails with an error "DB server is not running"


waspc/src/Wasp/Generator/DbGenerator/Operations.hs
```diff
@@ -141,9 +146,18 @@ isDbRunning ::
...
+    ExitSuccess -> return True
+    ExitFailure _ -> do
+      textOutput <- collectJobTextOutputUntilExitReceived chan
+
+      -- "Database not created" error is fine since Prisma will create it for us.
+      let isErrorTolerated = any containsDatabaseNotCreatedError textOutput
+      return isErrorTolerated
+  where
+    -- Prisma error code for "Database not created" is P1003.
+    containsDatabaseNotCreatedError = T.isInfixOf "P1003"
```
Martinsos, line 160: So, while this will work in 99% cases, what if the prisma error text contains `PIP1003` or `P10030` or anything else that is not `\bP1003\b`? Then it will give false positive.

So to be 100% safe, I would do a bit fancier check here. Probably use regex `\bP1003\b`.

waspc/src/Wasp/Generator/Job/IO.hs
```diff
@@ -29,9 +33,26 @@ readJobMessagesAndPrintThemPrefixed chan = runPrefixedWriter go
...
   hFlush outHandle
+
+collectJobMessages :: Chan J.JobMessage -> IO [J.JobMessageData]
+collectJobMessages = collect []
+  where
+    collect messages chan = do
+      jobMsg <- readChan chan
+      case J._data jobMsg of
+        J.JobExit {} -> return messages
+        message -> collect (message : messages) chan
```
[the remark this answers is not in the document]
Martinsos: Btw I can't remember what is the difference between JobMessage and JobMessageData, but this function says it collects JobMessages but it returns JobMessageData, is that an issue? Should the name also have `Data` in it?

Martinsos, line 58:
```suggestion
collectJobMessages :: Chan J.JobMessage -> IO [J.JobMessageData]
collectJobMessages = collect []
  where
    collect jobMsgsData chan = do
      jobMsg <- readChan chan
      case J._data jobMsg of
        J.JobExit {} -> return messages
        jobMsgData -> collect (jobMsgData : jobMsgsData) chan
```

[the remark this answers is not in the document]
Martinsos: Aha ok, I realized it now from the casing in another function above. So JobMessageData is either Text or Exit? In that case, since you know you are returning only Text's here, why not return `IO [Text]`? It is more precise!

waspc/src/Wasp/Generator/Job/IO.hs
```diff
@@ -29,9 +33,26 @@ readJobMessagesAndPrintThemPrefixed chan = runPrefixedWriter go
         J.JobOutput {} -> printJobMessagePrefixed jobMsg >> go
         J.JobExit {} -> return ()
 
+collectJobTextOutputUntilExitReceived :: Chan J.JobMessage -> IO [Text]
+collectJobTextOutputUntilExitReceived chan = collectJobMessages chan <&> mapMaybe jobMessageToText
+  where
+    jobMessageToText :: J.JobMessageData -> Maybe Text
+    jobMessageToText = \case
+      J.JobOutput text _ -> Just text
+      J.JobExit _ -> Nothing
```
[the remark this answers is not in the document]
Martinsos: Ok now looking at the function below, `collectJobMessages`, it looks to me like you really should just merge the two functions into one, into `collectJobTextOutputUntilExitReceived`.

Martinsos (changes requested):
@infomiho I like the solution, and very nice job on writing the main description, it gave me all the context I needed!

I left comments mostly on naming / clean code and one very unlikely bug.

waspc/src/Wasp/Generator/DbGenerator/Operations.hs
```diff
@@ -38,13 +42,23 @@ import Wasp.Generator.DbGenerator.Common
...
 import qualified Wasp.Generator.WriteFileDrafts as Generator.WriteFileDrafts
 import Wasp.Project.Db.Migrations (DbMigrationsDir)
 import Wasp.Util (checksumFromFilePath, hexToString)
 import Wasp.Util.IO (deleteFileIfExists, doesFileExist)
 import qualified Wasp.Util.IO as IOUtil
 
+data DbConnectionTestResult
+  = DbConnectionSuccess
+  | DbNotCreated
+  | DbConnectionFailure
```
Martinsos, line 59: I would be tempted to add `String` to `DbConnectionFailure` that contains actual error message, is there a reason why you decided to drop that information?

waspc/src/Wasp/Generator/DbGenerator/Operations.hs
```diff
@@ -135,15 +149,32 @@ dbSeed genProjectDir seedName = do
...
+      let databaseNotCreated = any containsDatabaseNotCreatedError outputLines
+
+      return $
+        if databaseNotCreated
+          then DbNotCreated
+          else DbConnectionFailure
+
+-- Prisma error code for "Database not created" is P1003.
+containsDatabaseNotCreatedError :: T.Text -> Bool
+containsDatabaseNotCreatedError text = text TR.=~ ("\\bP1003\\b" :: String)
```
Martinsos, line 172: I would probably put this in `where` under the `testDbConnection`. But maybe you took it out so you can write tests for it? If so, I understand. But then I would give it a bit more descriptive name: e.g. `prismaOutputContainsDbNotCreatedError`, because now it is missing the context it would have if it was in the `where`.

waspc/src/Wasp/Generator/DbGenerator/Operations.hs
```diff
@@ -135,15 +149,32 @@ dbSeed genProjectDir seedName = do
...
+
+isDbConnectionEstablished :: DbConnectionTestResult -> Bool
```
Martinsos, line 174: I get what you do with this function, makes sense! I wonder a bit about name -> so we call that "connection established"? Even if database does not exist? It is not a bad name, but I wonder if it is precise enough. I guess the problem is I am not sure what "db connection established" really means.
Is there a better matching name? Maybe `isDbInValidState`? or `isDbReady`? `isDbInUsableState`? `isDbReadyForPrisma`? Something in that direction? If we can't find anything better, fine, let's leave it as it is, but I do feel that `isDbConnectionEstablished` could maybe be better.
If nothing, you could go with `didDbConnectionTestSucceed`.

waspc/src/Wasp/Generator/Job/IO.hs
```diff
@@ -29,6 +31,15 @@ readJobMessagesAndPrintThemPrefixed chan = runPrefixedWriter go
         J.JobOutput {} -> printJobMessagePrefixed jobMsg >> go
         J.JobExit {} -> return ()
 
+collectJobTextOutputUntilExitReceived :: Chan J.JobMessage -> IO [Text]
+collectJobTextOutputUntilExitReceived = collect []
```
Martinsos, line 35: In haskell, usual idiom is to go with `go`, for functions like `collect`, as a name.

Martinsos (approved):
Nice, looking good! I left a couple more comments but I think it is in good shape so I approved -> to with these comments as you wish and you can merge!

waspc/src/Wasp/Generator/DbGenerator/Operations.hs
infomiho: Because we don't get an error message per se, but we get some command output that might be useless unless parsed.

waspc/src/Wasp/Generator/DbGenerator/Operations.hs
infomiho: I've went with `isDbConnectionPossible` since that's the check I was going for. The DB server can be connected to.

PR #1561, opened 2023-11-07
Escape Prisma command when running with `script` on Linux

### LINKED ISSUE #1486: wasp-cli command not handling white-space in path while running app
**Description**
So basically, what happens, is that when we run `wasp db migrate-dev`, and our app is in a directory which has white-space in its path, database migration fails because it breaks on that white-space.

**To Reproduce**
Steps to reproduce the behaviour:
1. Following the setup guide from the readme.md , *todoApp* project under run example app.

2. Run `cabal run wasp-cli db migrate-dev`
output of the above command:
```─$ cabal run wasp-cli db migrate-dev
Up to date

[Db!]     Error: P1001
[Db!]     
[Db!]     Can't reach database server at `localhost`:`5432`
[Db!]     
[Db!]     Please make sure your database server is running at `localhost`:`5432`.
[Db!]     
                                                                                                                                     

✅ --- Your wasp project has successfully compiled. -------------------------------                                                  
                                                                                                                                     

                                                                                                                                     
❌ --- [Error] Can not connect to database: ---------------------------------------                                                  
                                                                                                                                     
The database needs to be running in order to execute this command. You can easily start a managed dev database with `wasp start db`. ```

3. Run `cabal run wasp-cli start db`
output of the above command:
```└─$ cabal run wasp-cli start db      
Up to date
                                                                                                                                     
Additional info:                                                                                                                     
     postgresql://postgresWaspDevUser:postgresWaspDevPass@localhost:5432/todoApp-29a2c8c08a                                          
     wasp-dev-db-todoApp-29a2c8c08a                                                                                                  
                                                                                                                                     
...

PostgreSQL Database directory appears to contain a database; Skipping initialization

2023-10-09 18:33:27.246 UTC [1] LOG:  starting PostgreSQL 16.0 (Debian 16.0-1.pgdg120+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 12.2.0-14) 12.2.0, 64-bit
2023-10-09 18:33:27.257 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
2023-10-09 18:33:27.257 UTC [1] LOG:  listening on IPv6 address "::", port 5432
2023-10-09 18:33:27.316 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
2023-10-09 18:33:27.380 UTC [28] LOG:  database system was shut down at 2023-10-08 17:45:39 UTC
2023-10-09 18:33:27.642 UTC [1] LOG:  database system is ready to accept connections

 ```
 Keeping this terminal running.
4. Open another terminal to same example app path and Check working directory 
``` 
└─$ pwd
/media/naruto/New Volume/Open Source Contribution/wasp/waspc/examples/todoApp

``` 
5.  Run again `cabal run wasp-cli db migrate-dev` 
It will give an error with incorrect file path
``` 
└─$ cabal run wasp-cli db migrate-dev
Up to date

[Db]      [-] Removed tables
[Db]        - User
[Db]        - SocialLogin
[Db]        - Task
[Db]      
[Db]      [*] Changed the `SocialLogin` table
[Db]        [-] Removed foreign key on columns (userId)
[Db]      
[Db]      [*] Changed the `Task` table
[Db]        [-] Removed foreign key on columns (userId)
                                                                                                                                     

✅ --- Your wasp project has successfully compiled. -------------------------------                                                  
                                                                                                                                     

🐝 --- Starting database migration... ---------------------------------------------                                                  
                                                                                                                                     
zsh:1: no such file or directory: /media/naruto/New

                                                                                                                                     
❌ --- [Error] Migrate dev failed: ------------------------------------------------                                                  
                                                                                                                                     
Migrate (dev) failed with exit code: 127 
```

**Expected behaviour**
CLI should consider the full path of the project directory to run the command
So ask you can see the project path is 
"*/media/naruto/New Volume/Open Source Contribution/wasp/waspc/examples/todoApp*"
But it is considering only till the first white space 
```zsh:1: no such file or directory:  /media/naruto/New```
 


 - OS: Linux (not tested on other OS)
 - Browser: N/A
 - Version: waspc-0.11.5

### PR DESCRIPTION
**Pull request by infomiho**
Closes #1486

### DIFF
--- a/waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
+++ b/waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
@@ -28,7 +28,6 @@ import qualified Wasp.Generator.Job as J
 import Wasp.Generator.Job.Process (runNodeCommandAsJob, runNodeCommandAsJobWithExtraEnv)
 import Wasp.Generator.ServerGenerator.Common (serverRootDirInProjectRootDir)
 import Wasp.Generator.ServerGenerator.Db.Seed (dbSeedNameEnvVarName)
-import Wasp.Util.ShellCommand (ShellCommandArg)
 import qualified Wasp.Util.ShellCommand as Shell
 
 migrateDev :: Path' Abs (Dir ProjectRootDir) -> MigrateArgs -> J.Job
@@ -57,22 +56,28 @@ migrateDev projectDir migrateArgs =
     -- NOTE(martin): We do "--skip-seed" here because I just think seeding happening automatically
     --   in some situations is too aggressive / confusing.
     prismaMigrateCmd =
-      [ Shell.QuotedArgument $ absPrismaExecutableFp projectDir,
-        Shell.RawArgument "migrate",
-        Shell.RawArgument "dev",
-        Shell.RawArgument "--schema",
-        Shell.QuotedArgument $ SP.fromAbsFile schemaFile,
-        Shell.RawArgument "--skip-generate",
-        Shell.RawArgument "--skip-seed"
+      [ Shell.Unsafe $ absPrismaExecutableFp projectDir,
+        Shell.Safe "migrate",
+        Shell.Safe "dev",
+        Shell.Safe "--schema",
+        Shell.Unsafe $ SP.fromAbsFile schemaFile,
+        Shell.Safe "--skip-generate",
+        Shell.Safe "--skip-seed"
       ]
         ++ asPrismaCliArgs migrateArgs
 
-asPrismaCliArgs :: MigrateArgs -> [ShellCommandArg]
+asPrismaCliArgs :: MigrateArgs -> [Shell.ShellCommandArg]
 asPrismaCliArgs migrateArgs = do
   concat . concat $ [createOnlyArg, nameArg]
   where
-    createOnlyArg = [[Shell.RawArgument "--create-only"] | _isCreateOnlyMigration migrateArgs]
-    nameArg = [[Shell.RawArgument "--name", Shell.RawArgument name] | Just name <- [_migrationName migrateArgs]]
+    createOnlyArg =
+      [[Shell.Safe "--create-only"] | _isCreateOnlyMigration migrateArgs]
+    nameArg =
+      [ [ Shell.Safe "--name",
+          Shell.Safe name
+        ]
+        | Just name <- [_migrationName migrateArgs]
+      ]
 
 -- | Diffs the Prisma schema file against the db.
 -- Because of the --exit-code flag, it changes the exit code behavior
@@ -95,7 +100,11 @@ migrateDiff projectDir = runPrismaCommandAsDbJob projectDir $ \schema ->
 -- Therefore, this should be checked **after** a command that ensures connectivity.
 migrateStatus :: Path' Abs (Dir ProjectRootDir) -> J.Job
 migrateStatus projectDir = runPrismaCommandAsDbJob projectDir $ \schema ->
-  ["migrate", "status", "--schema", SP.fromAbsFile schema]
+  [ "migrate",
+    "status",
+    "--schema",
+    SP.fromAbsFile schema
+  ]
 
 -- | Runs `prisma migrate reset`, which drops the tables (so schemas and data is lost) and then
 -- reapplies all the migrations.
@@ -103,7 +112,13 @@ reset :: Path' Abs (Dir ProjectRootDir) -> J.Job
 reset projectDir = runPrismaCommandAsDbJob projectDir $ \schema ->
   -- NOTE(martin): We do "--skip-seed" here because I just think seeding happening automatically on
   --   reset is too aggressive / confusing.
-  ["migrate", "reset", "--schema", SP.fromAbsFile schema, "--skip-generate", "--skip-seed"]
+  [ "migrate",
+    "reset",
+    "--schema",
+    SP.fromAbsFile schema,
+    "--skip-generate",
+    "--skip-seed"
+  ]
 
 -- | Runs `prisma db seed`, which executes the seeding script specified in package.json in
 --   prisma.seed field.
@@ -127,7 +142,14 @@ dbExecuteTest projectDir =
   let absSchemaPath = projectDir </> dbSchemaFileInProjectRootDir
    in runPrismaCommandAsDbJob
         projectDir
-        (const ["db", "execute", "--stdin", "--schema", SP.fromAbsFile absSchemaPath])
+        ( const
+            [ "db",
+              "execute",
+              "--stdin",
+              "--schema",
+              SP.fromAbsFile absSchemaPath
+            ]
+        )
 
 -- | Runs `prisma studio` - Prisma's db inspector.
 runStudio :: Path' Abs (Dir ProjectRootDir) -> J.Job
@@ -137,7 +159,10 @@ runStudio projectDir = runPrismaCommandAsDbJob projectDir $ \schema ->
 generatePrismaClient :: Path' Abs (Dir ProjectRootDir) -> (String, String) -> JobType -> J.Job
 generatePrismaClient projectDir prismaClientOutputDirEnv jobType =
   runPrismaCommandAsJobWithExtraEnv jobType envVars projectDir $ \schema ->
-    ["generate", "--schema", SP.fromAbsFile schema]
+    [ "generate",
+      "--schema",
+      SP.fromAbsFile schema
+    ]
   where
     envVars = [prismaClientOutputDirEnv]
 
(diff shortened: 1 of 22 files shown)

### CODE REVIEW

File: waspc/waspc.cabal [hunk 1]
```diff
@@ -65,7 +65,6 @@ data-files:
   packages/studio/dist/**/*.js
   packages/studio/dist/**/*.html
   packages/studio/dist/**/*.css
-  packages/studio/dist/**/*.png
```
Martinsos, line 68: Hmmm how is this connected with this PR?

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
```diff
@@ -46,31 +47,38 @@ migrateDev projectDir migrateArgs =
...
-      [ absPrismaExecutableFp projectDir,
+      [ Shell.Quoted $ absPrismaExecutableFp projectDir,
         "migrate",
         "dev",
         "--schema",
-        SP.fromAbsFile schemaFile,
+        Shell.Quoted $ SP.fromAbsFile schemaFile,
         "--skip-generate",
         "--skip-seed"
       ]
```
Martinsos, line 71: So these are all ShellCommandArg types, but the one that are string literals are being cast into Raw due to extension OverloadedStringLiterals + ShellCommandArg being isString?

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
```diff
@@ -46,31 +47,38 @@ migrateDev projectDir migrateArgs =
...
         then -- NOTE(martin): On MacOS, command that `script` should execute is treated as multiple arguments.
-          ["-Fq", "/dev/null"] ++ prismaMigrateCmd
+          ["-Fq", "/dev/null"] ++ Shell.getShellArgValues prismaMigrateCmd
         else -- NOTE(martin): On Linux, command that `script` should execute is treated as one argument.
-          ["-feqc", unwords prismaMigrateCmd, "/dev/null"]
+        -- NOTE(miho): Since we are passing the arguments as a single string, we need to wrap the arguments
+        --  which might contain spaces in them. We are using Shell.Quoted to know which
+        --   arguments should be wrapped in quotes and which shouldn't.
+        --   Specifically, we are using `showShellArgs` to wrap file paths in quotes. Other arguments
+        --   are not wrapped in quotes since we know their values and they don't contain spaces.
```
Martinsos, line 56: Where are we passing arguments as a single string?
Aha, now I realized there is a comment above saying that on Linux, command that script should execute is treated as one argument.

waspc/src/Wasp/Util/ShellCommand.hs
```diff
@@ -0,0 +1,22 @@
...
+
+instance Show ShellCommandArg where
+  show (Raw arg) = arg
+  show (Quoted arg) = "\"" ++ arg ++ "\""
+
+instance IsString ShellCommandArg where
+  fromString = Raw
+
+showShellArgs :: [ShellCommandArg] -> String
+showShellArgs = unwords . map show
```
Martinsos, line 16: I would drop this one, it doesn't do much, I can just call `unwords . map show` myself if I want and at least I know what it is doing then. Beacuse this not only shows the args but also concatenates them and that is not clear from the name.
So better not to have this function.

waspc/src/Wasp/Util/ShellCommand.hs
```diff
@@ -0,0 +1,22 @@
...
+
+getShellArgValues :: [ShellCommandArg] -> [String]
```
Martinsos, line 18: This one is a bit hard to understand. It basically unpacks the values. But I found it hard to understand from the name.

Maybe you could rather have `toRaw :: ShellCommandArg -> ShellCommandArg` and then you can do `(show . toRaw) <$> args` if you want to show them all without quotes -> because that is what you are using this function for. And then it is at least very clear this is what you are doing.

waspc/src/Wasp/Util/ShellCommand.hs
```diff
@@ -0,0 +1,22 @@
+module Wasp.Util.ShellCommand where
```
Martinsos, line 1: This is quite cool, I never used `IsString`!

I like the concept you used here, it is quite neat. So basically we have this notion of Shell command argument, and we can say if it is quoted or not, and we carry this knowledge around, and then can act on it.

What I found tricky is:
1. There are no comments in this file explaining the reasoning behind it and what it is for. It is somewhat obvious, but I don't think it is obvious enough.
2. Continuing on above, I think the reason it is not obvious enough is that usecase feels quite trivial. If I want args quoted, why don't I continue using String and just add quotes to it? Ok, I guess this gives more knowledge, so you can avoid quoting already quoted arg, or you can unquote already quoted arg (although that feels a bit weird, why was it quoted then, but ok, maybe we know somebody else will quote it). Still, I think it is a bit confusing so explanation would help.
3. Using it above in the Jobs.hs feels a bit like overkill. Basically we need to quote these two args that are paths. And in one situation we need to quote them, in other we don't. But now we have to introduce this new type in all of the args there. But, not all of them, just some of them -> its not like we are using it across the whole codebase, just here. Feels weird that we are using it just in the small part of the codebase, how do we choose? When should I use this? Btw, isn't there a simpler solution -> can't we just quote those paths and that is it? Is it a problem if they are also quoted in that situation where they shouldn't be quoted, or will it just work?

infomiho: So, the problem we have is the following: we need the same command in a slightly different ways for Linux and for MacOS.

### MacOS

- The arguments should be a list of arguments
- The paths passed into the prisma command **MUST NOT** be quoted

### Linux

- The arguments are concatenated into a single string
- The paths **MUST** be quoted (since they can have spaces in them)

I'll try to simplify the solution and make it very local (it will only affect the array of prisma db migrate args and nothing else).

infomiho:
@Martinsos I've simplified the impl, tested it 

Screenshot 2023-11-13 at 12 13 28
Screenshot 2023-11-13 at 12 12 37

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
```diff
@@ -93,15 +99,25 @@ migrateDiff projectDir = runPrismaCommandAsDbJob projectDir $ \schema ->
 -- Therefore, this should be checked **after** a command that ensures connectivity.
 migrateStatus :: Path' Abs (Dir ProjectRootDir) -> J.Job
 migrateStatus projectDir = runPrismaCommandAsDbJob projectDir $ \schema ->
-  ["migrate", "status", "--schema", SP.fromAbsFile schema]
+  [ "migrate",
+    "status",
+    "--schema",
+    SP.fromAbsFile schema
+  ]
```
Martinsos, line 106: If you really like it better in multiple lines fine, but I liked it better in one line, found it easier to read. And since it has nothing to do with this PR maybe we could revert it? Same goes for other couple of functions in this file.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
```diff
@@ -46,31 +47,36 @@ migrateDev projectDir migrateArgs =
...
-      [ absPrismaExecutableFp projectDir,
+          ["-feqc", unwords $ buildPrismaMigrateCmd Shell.quoteArg, "/dev/null"]
+
+    -- NOTE(miho): Since we are running the Prisma command using `script` and we are doing it
+    --  in two different ways (MacOS and Linux), we have to take care of quoting the paths
+    --  differently.
+    --  * MacOS - we are passing the command as multiple arguments, so we MUST NOT quote the paths.
+    --  * Linux - we are passing the command as one argument, so we MUST quote the paths.
+    buildPrismaMigrateCmd quoteArg =
+      [ quoteArg $ absPrismaExecutableFp projectDir,
```
Martinsos, line 60:
```suggestion
    -- NOTE(miho): Since we are running the Prisma command using `script` and we are doing it
    --  in two different ways (MacOS and Linux), we have to take care of quoting the path arguments
    --  differently.
    --  * MacOS - we are passing the command as multiple arguments, so we MUST NOT quote the path arguments.
    --  * Linux - we are passing the command as one argument, so we MUST quote the path arguments.
    buildPrismaMigrateCmd :: (String -> String) -> [String]
    buildPrismaMigrateCmd argQuoter =
      [ argQuoter $ absPrismaExecutableFp projectDir,
```

waspc/src/Wasp/Util/ShellCommand.hs
Martinsos: Nice, this is much simpler now!

I left some small comments but otherwise it is good.

I would possibly consider skipping this file here, ShellCommand.hs, and just inlining this function `quoteArg` where it is used, since it is so niche / local, I don't think we benefit much from having it here in a separate file.

Martinsos (approved):
Approved! Left some comments, but all is good already, so handle them as you like and all good from my side!

PR #1562, opened 2023-11-08
Add check if migrations dir exists before copying

### LINKED ISSUE #1352: Running `wasp db migrate-dev` without any migrations causes a weird error
Running `wasp db migrate-dev` when there are no entities defined results in this weird error which is shown to the user:
```
wasp-bin: /private/tmp/auth-ui-docs/.wasp/out/db/migrations/: getFileStatus: does not exist (No such file or directory)
```
We should just not output the error since it's okay for the `migrations` dir not to exist in this case.

infomiho:
Closes #1352

### DIFF
--- a/waspc/src/Wasp/Generator/DbGenerator/Operations.hs
+++ b/waspc/src/Wasp/Generator/DbGenerator/Operations.hs
@@ -36,7 +36,7 @@ import Wasp.Generator.DbGenerator.Common
     webAppPrismaClientOutputDirEnv,
   )
 import qualified Wasp.Generator.DbGenerator.Jobs as DbJobs
-import Wasp.Generator.FileDraft.WriteableMonad (WriteableMonad (copyDirectoryRecursive))
+import Wasp.Generator.FileDraft.WriteableMonad (WriteableMonad (copyDirectoryRecursive, doesDirectoryExist))
 import qualified Wasp.Generator.Job as J
 import Wasp.Generator.Job.IO (printJobMsgsUntilExitReceived, readJobMessagesAndPrintThemPrefixed)
 import qualified Wasp.Generator.WriteFileDrafts as Generator.WriteFileDrafts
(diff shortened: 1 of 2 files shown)


waspc/src/Wasp/Generator/DbGenerator/Operations.hs
```diff
@@ -77,11 +77,17 @@ finalizeMigration genProjectRootDirAbs dbMigrationsDirInWaspProjectDirAbs onLast
...
+  doesMigrationsDirExist <- doesDirectoryExist $ SP.fromAbsDir genProjectMigrationsDir
+
+  if doesMigrationsDirExist
+    then copyMigrationsDir
+    else return $ Right ()
   where
+    copyMigrationsDir =
+      copyDirectoryRecursive genProjectMigrationsDir waspMigrationsDir >> return (Right ())
+        `catch` (\e -> return $ Left $ show (e :: P.PathException))
+        `catch` (\e -> return $ Left $ show (e :: IOError))
```
Martinsos, line 90:
```suggestion
copyMigrationsBackToSource genProjectRootDirAbs dbMigrationsDirInWaspProjectDirAbs = do
  doesDirectoryExist (SP.fromAbsDir genProjectMigrationsDir) >>= \case ->
    False -> return $ Right ()
    True -> copyDirectoryRecursive genProjectMigrationsDir waspMigrationsDir >> return (Right ())
        `catch` (\e -> return $ Left $ show (e :: P.PathException))
        `catch` (\e -> return $ Left $ show (e :: IOError))
```

Martinsos: We could also consider renaming whole functoin to `copyMigrationsBackToSourceIfTheyExist`, but I am not sure, we can also skip it maybe.

Martinsos (approved):
Nice @infomiho ! I approved, but also left one/two comments that possibly improve code a bit, take a look and do as you like with them.

PR #1568, opened 2023-11-10
Add fallback sans-serif font

infomiho:
There is a brief flash of fallback font being used until `Inter` is loaded. This change changes that fallback to be a `sans-serif` font instead of an ugly `serif` font.

### Before
https://github.com/wasp-lang/wasp/assets/2223680/21cd074a-92a1-4a4b-b132-d15a7e07fabc

### After
https://github.com/wasp-lang/wasp/assets/2223680/4c1f3ae2-2519-4f77-9520-6c77e2cc2d96

### DIFF
--- a/web/src/css/custom.css
+++ b/web/src/css/custom.css
@@ -58,7 +58,7 @@
 
   /* Infima overrides */
   --ifm-container-width-xl: 1280px;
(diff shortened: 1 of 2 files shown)


matijaSos (approved):
awesome change!

PR #1577, opened 2023-11-17
[MAGE] add delete user option 

vincanger:
### Description

Adds delete account functionality to the UserPage, and another couple small improvements

### DIFF
--- a/wasp-ai/.prettierrc
+++ b/wasp-ai/.prettierrc
@@ -1,0 +1,6 @@
+{
+  "semi": true,
+  "singleQuote": false,
+  "trailingComma": "es5",
+  "printWidth": 120
+}
--- a/wasp-ai/main.wasp
+++ b/wasp-ai/main.wasp
@@ -106,6 +106,11 @@ action createFeedback {
   entities: [Feedback]
 }
 
+action deleteUser {
+  fn: import { deleteUser } from "@server/operations.js",
+  entities: [User]
+}
+
 query getFeedback {
   fn: import { getFeedback } from "@server/operations.js",
   entities: [Feedback]
@@ -154,7 +159,7 @@ entity SocialLogin {=psl
   providerId String
 
   userId Int
-  user User @relation(fields: [userId], references: [id])
+  user User @relation(fields: [userId], references: [id], onDelete: Cascade)
   createdAt DateTime @default(now())
 psl=}
 
--- a/wasp-ai/src/client/pages/ResultPage.jsx
+++ b/wasp-ai/src/client/pages/ResultPage.jsx
@@ -392,14 +392,19 @@ export function OnSuccessModal({ isOpen, setIsOpen, appGenerationResult }) {
     const logText = appGenerationResult?.project?.logs?.find((log) =>
       log.content.includes("tokens usage")
     )?.content;
-    const regex = /total tokens usage: ~(\d+(\.\d+)?)/i;
-    const match = logText?.match(regex);
-    if (match && match[1]) {
-      const num = Number(match[1]) * 1000;
-      if (Number.isInteger(num)) {
-        setNumTokensSpent(num);
+    
+    if (logText) {
+      const regex = /Total\s+tokens\s+usage\s*:\s*~\s*(\d+(?:\.\d+){0,1})\s*k\b/;
+      const match = logText.match(regex);
+
+      if (match) {
+        const num = parseFloat(match[1]);
+        setNumTokensSpent(num * 1000);
+      } else {
+        console.log('No match found');
       }
     }
+
   }, [appGenerationResult]);
 
   useEffect(() => {
@@ -638,7 +643,7 @@ function Feedback({ projectId }) {
             <RadioGroup value={score} onChange={setScore}>
               <div className="flex space-x-2">
                 {scoreOptions.map((option) => (
-                  <RadioGroup.Option value={option}>
+                  <RadioGroup.Option key={option} value={option}>
                     {({ active, checked }) => (
                       <div
                         className={`
--- a/wasp-ai/src/client/pages/UserPage.jsx
+++ b/wasp-ai/src/client/pages/UserPage.jsx
@@ -1,20 +1,27 @@
+import { useState } from "react";
 import { useQuery } from "@wasp/queries";
 import getProjectsByUser from "@wasp/queries/getProjectsByUser";
 import { Link } from "@wasp/router";
 import { Color } from "../components/Color";
 import { Header } from "../components/Header";
-import { PiDownloadDuotone } from "react-icons/pi";
+import { PiDownloadDuotone, PiSealWarningDuotone } from "react-icons/pi";
 import logout from "@wasp/auth/logout";
 import { FiLogOut } from "react-icons/fi";
 import { format } from "timeago.js";
 import { StatusPill } from "../components/StatusPill";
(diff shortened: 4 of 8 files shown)


wasp-ai/.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "semi": true,
+  "singleQuote": false,
+  "trailingComma": "es5",
+  "printWidth": 120
```
Martinsos, line 5: Ha, I read in their docs that this is not the same as actual width of how long the lines should be and that they advise leaving this at 80, even if we are going with lines of 100 or 120. I was just reading this yesterday. But I do have to admit I find it quite annoying, it is overly aggressive with breaking those lines.
I would however advise going with 100, I found that to be best number so far -> it is short enough to still mostly fit into smaller editors like this one on github, but is still not as short as 80 which is just annoying.

wasp-ai/main.wasp
```diff
@@ -154,7 +159,7 @@ entity SocialLogin {=psl
   providerId String
 
   userId Int
-  user User @relation(fields: [userId], references: [id])
+  user User @relation(fields: [userId], references: [id], onDelete: Cascade)
```
Martinsos, line 162: Thought: once projects are deleted, we will never know they existed, so our number of created projects will become smaller. Would be cool if they only thing that is left is this idea of "there was a project that was deleted". But to be honest I am not sure myself how to implement that simply, so probably best to just ignore it, doesn't matter.

wasp-ai/src/client/pages/ResultPage.jsx
```diff
@@ -392,14 +392,19 @@ export function OnSuccessModal({ isOpen, setIsOpen, appGenerationResult }) {
...
+    
+    if (logText) {
+      const regex = /Total\s+tokens\s+usage\s*:\s*~\s*(\d+(?:\.\d+){0,1})\s*k\b/;
+      const match = logText.match(regex);
+
+      if (match) {
+        const num = parseFloat(match[1]);
+        setNumTokensSpent(num * 1000);
+      } else {
+        console.log('No match found');
```
Martinsos, line 404: You could make this more descriptive: "Failed to parse total number of tokens used: no regex match."

wasp-ai/src/client/pages/UserPage.jsx
```diff
@@ -100,8 +124,54 @@ function UserProjectsTable({ projects }) {
...
     </table>
   );
-}
\ No newline at end of file
+}
+
+function DeleteUserModal({ isOpen, setIsOpen, deleteUser, user }) {
+  async function deleteUserHandler() {
+    await deleteUser({ userId: user.id });
+    logout();
```
Martinsos, line 143: Nice job remembering to log out the user!

wasp-ai/src/client/pages/UserPage.jsx
```diff
@@ -100,8 +124,54 @@ function UserProjectsTable({ projects }) {
...
+      onClose={() => setIsOpen(false)}
+      title={
+        <div className="flex items-center gap-2">
+          <PiSealWarningDuotone arie-hidden="true" /> Are You Sure You Want to Delete Your Account?
+        </div>
+      }
+    >
+      <div className="mt-10 space-y-10">
+        <p className="px-8 text-base leading-relaxed text-center text-gray-500">
+          You will lose access to your current projects and data.
```
Martinsos, line 158: Are they just loosing access, or are we actually deleting their data? If I am correct, the latter? If so, I would communicate that clearly for two reasons:
1. So it is clear to them they can't later restore this access.
2. So it is clear to them that we will not keep their data on our side.

So we could say smth like "We will delete all your current projects and data and you won't be able to restore it."

wasp-ai/src/server/operations.ts
```diff
@@ -237,3 +237,15 @@ export const getProjectsByUser: GetProjectsByUser<void, Project[]> = async (_arg
...
+  if (!context.user || context.user.id !== args.userId) {
+    throw new HttpError(401, "Not authorized");
+  }
+
+  return await context.entities.User.delete({
+    where: {
+      id: args.userId,
+    },
+  });
+};
```
Martinsos, line 251: I think you can simplify this in following way:
1. Rename to `DeleteMyself`.
2. Don't accept `userId` as an argument.
3. Use only `context.user.id` to know whom to delete.

Effectively it will work the same, but I think it reduces the space for us to make a mistake here in the code, be it now or in the future when changing something.

Martinsos (commented):
@vincanger nice! Looks great, couple of small comments and we can merge!

wasp-ai/main.wasp
vincanger: this isnt for projects. when the user is deleted it deletes the social login entity, but leaves their projects in the DB.

wasp-ai/src/client/pages/UserPage.jsx
vincanger: No, I cascaded the SocialLogin entity, not their projects

wasp-ai/main.wasp
vincanger: I think it's fine to leave the generated apps in the database in case they've been shared. there is no connection whatsoever to the user that created it.

vincanger: ok I've gone ahead and deleted all user-relevant info leaving just the project shell :)

Martinsos (approved):
@vincanger looking good! I approved, but just wanted to check, what does it look like if somebody tries to open a project of deleted user via its URL? Is it broken, or does it look somewhat reasonable? Would be good if it isn't a total mess, but instead it said something like `This project was deleted`.
Also,

Martinsos:
> @vincanger looking good! I approved, but just wanted to check, what does it look like if somebody tries to open a project of deleted user via its URL? Is it broken, or does it look somewhat reasonable? Would be good if it isn't a total mess, but instead it said something like `This project was deleted`. Also,

Aha, ok it will have `Deleted project` title and description.

Might be better if we set `title` and `description` to NULL or some value like that, so it is clear they are empty, and then in that UI we detect they are set to NULL and then show one big "Deleted project". But we don't have to bother with this if current solution looks relatively reasonable, your call.

vincanger:
> > @vincanger looking good! I approved, but just wanted to check, what does it look like if somebody tries to open a project of deleted user via its URL? Is it broken, or does it look somewhat reasonable? Would be good if it isn't a total mess, but instead it said something like `This project was deleted`. Also,
> 
> Aha, ok it will have `Deleted project` title and description.
> 
> Might be better if we set `title` and `description` to NULL or some value like that, so it is clear they are empty, and then in that UI we detect they are set to NULL and then show one big "Deleted project". But we don't have to bother with this if current solution looks relatively reasonable, your call.

Good catch!

I didn't want to change  the name and description properties to optional. so I just made the delete function update the status to "deleted", and then check for that on the client and show:
image

Martinsos (approved):
LGTM: I left a small note regarding how "delete" could be more "clean cody", but since we dropped the ball anyway for statuses regarding that, then we don't need to fix it now, we can fix it all later, so all good.

NOTE: It is a bit tricky when formatting is all mixed up with changes, it makes it really hard for me to figure out which are actual changes.
Ways to go about this, when we have a codebase like this where files are not yet formatted, is to e.g. turn of formatting in your editor and do it without, and then do separate PR which does the formatting. Or, maybe at least a separate commit (although that is also tricky to review if there is more than one non-formatting commit).
Or you could maybe do formatting, but then comment in the PR on the changes that are relevant, so I know what to focus on.

PR #1586, opened 2023-11-29
Made CI checks during telemetry a bit more thorough.

Martinsos:


waspc/cli/src/Wasp/Cli/Command/Telemetry/Project.hs
```diff
@@ -76,17 +77,33 @@ considerSendingData telemetryCacheDirPath userSignature projectHash cmdCall = do
...
 
 getTelemetryContext :: IO String
-getTelemetryContext =
+getTelemetryContext = do
   unwords . filter (not . null)
     <$> sequence
       [ fromMaybe "" <$> ENV.lookupEnv "WASP_TELEMETRY_CONTEXT",
-        checkIfOnCi <&> \case True -> "CI"; False -> ""
+        checkIfOnCi <&> \case True -> "CI"; False -> "",
+        -- NOTE(martin): We are not 100% sure how correctly does `hIsTerminalDevice` detect
```
infomiho, line 85: @Martinsos can this call fail for what ever reason? Do we need exception handling or smth here

Martinsos: Good point -> so I didn't think this one is particularly likely to throw an error (but every IO operation can throw an error!), so I haven't taken special measures here.
We do have one big catch already for telemetry, so any fail in it can't crash the CLI but will just not send telemetry.

That said, this function is a bit of a mystery to us, and it would be a shame if it started failing our telemetry silently, so better to be safe than sorry I think, so now prompted by your comment I added an additional catch to it that will interpret any error as "True".

infomiho (approved):

PR #1598, opened 2023-12-11
Enhance OpenAI models used for MAGE

jankovidakovic:
This PR enables MAGE to be used with all non-legacy OpenAI chat models.

Features:
 - added all OpenAI models that are compatible with the new "Chat-Completion" API
 - remove logic for checking if GPT-4 is available
 - refactor functions regarding initialization and modifications of ChatGPT params
 - enable configuration of both "planner" model (`planGptModel`), and "coder" model (`baseGptModel`) through the CLI
 - freeze model versions for both base and plan model

4. [x] **Breaking change** (fix or feature that would cause existing functionality to not work as expected).

### Models
The list of models is taken from the OpenAI website, available here: https://platform.openai.com/docs/models. The list is likely to change (e.g. models become deprecated, new models are added), so updates to the list of models that MAGE supports are probably expected in the future.

Right now, the codebase supports all OpenAI models that are compatible with "chat-completion" API. In practice, this means that all reasonable GPT-3.5 and GPT-4 models are supported.

### Plan generation vs the fixing step
Generation of Wasp app happens in two stages -- planning and coding. Each stage generates its respective output, and also subsequently fixes the output for "common mistakes".

The first stage generates the plan, based on which the second stage will create the actual code.
Since the first stage is deemed more important (for now), default model version is fixed to `gpt-4-0613`.

The second stage generates the actual code (both Wasp and Javascript). This is an easier task, so the default model version is fixed to `gpt-3.5-turbo-0613` (to save costs).

This PR enables configuring both of these models via the CLI-supplied JSON. An example configuration is given below:

```javascript
{
  "planGptModel": "gpt-4-0613", // used for the first stage (planning)
  "baseGptModel": "gpt-4-1106-preview",  // used for the second stage (generating code)
}
```

## Due to the minor refactoring, this PR (probably) introduces breaking changes.

The minor refactoring was done in order to increase clarity for future usage.

The breaking change probably affects the `usemage.ai` web-app in some way, since the naming of the CLI-supplied args changed. 
I'm not sure how exactly are "breaking" changes handled on wasp, so I'm happy to help with looking more deeply into this.


Martinsos (changes requested):
Nice work, I love the summary of the PR! Did GPT help with that one :D?

I left some comments, but we are 90% there.

Btw I noticed how you used "planning model" and "coding model" terms. They make sense and describe what they do well. On the other hand if I say "planning model" and "base model", then extra information is needed: "yeah, base model is the one that does coding" :). So, "coding" is really a better adjective than "base". Base is more robust to changes in what the model does, sure, but precision is often better.
So I am a bit tempted to go for "planningGptModel" and "codingGtpModel" instead of "planGtpModel" and "baseGptModel"! But, I am also not 100% sure, which means I am also ok with leaving it as it is (which is often the best choice when not sure). To conclude: if you want to do final renaming into "planning/coding" instead of "plan/base", I am ok with that, if you prefer keeping current state, I am also ok with that, your call.

Regarding breaking changes -> so that actually refers more to the API of Wasp Framework, not really to Mage because Mage doesn't have an official API yet, so we don't have to call this a breaking change, it's ok. If at one point API of Mage becomes official then we will behave a bit differently.

Also, you said you are not sure how this affects Mage web app -> yeah, we should make sure it still works ok, regarding how it passes parameters to the CLI. You can see how it calls CLI here: https://github.com/wasp-lang/wasp/blob/main/wasp-ai/src/server/jobs/generateApp.ts#L57-L79 . As you can notice, it doesn't pass any GPT model information, so we are good, nothing to update here.

jankovidakovic:
Everything resolved.

waspc/src/Wasp/AI/GenerateNewProject/Common.hs
```diff
@@ -7,9 +7,10 @@ module Wasp.AI.GenerateNewProject.Common
     getProjectPrimaryColor,
     emptyNewProjectConfig,
     queryChatGPTForJSON,
-    defaultChatGPTParams,
-    defaultChatGPTParamsForFixing,
     writeToWaspFileEnd,
+    planningChatGPTParams,
+    codingChatGPTParams,
```
Martinsos, line 12: Aha so you did switch from "base" to "coding" -> ok, sounds good!

Martinsos (approved):
LGTM, great job!

PR #1601, opened 2023-12-14
Wasp ai cli mage

### LINKED ISSUE #1575: Merge wasp-ai into main and improve Mage's CLI DX
- [x] Merge into `main`.
      What about the fact that `main` is going toward 0.12 version -> we should build in those change into Mage, right? Yeah, we should, we should update it.
- [x] Update Mage Wasp app to use newer version of Wasp?
- [x] Make CLI nice -> have it ask questions interactively, print stuff nicely, throw in some emojis ... . Remove debugging output.
- [x] Document CLI option better -> mention it in docs, maybe on the wasp landing page, certainly on the usemage.ai landing page, update FAQ there, add it to CLI usage docs, ... .
- [x] Document processes regarding deployment and hotfixes. If we want to make fixes to currently deployed version, we have to do it on `release`, don't we? What about `wasp-ai`?
- [x] Make it easy to provide an OpenAI key -> we can ask for it if you don't have it.
- [x] Possibly provide specifying any gpt model, not just the ones we have enumerated?

Martinsos:
This PR makes Wasp AI a proper, polished, well-documented feature of `wasp` CLI.

Resolves #1575 .

High-level list of changes:
1. If you choose `AI` in `wasp new`, you get asked a set of interactive questions about which GPT models and which temperature you want to use (besides app name and description).
2. `wasp new:ai` commands allows programmatically running Wasp AI. Its interface changed a bit, and it is now documented in the CLI usage output.
3. Output of Wasp AI from `wasp` CLI when writing to disk is now enriched with newlines, colors, and some extra information. In the meantime, output towards Mage hasn't changed (due to some cool logic we implemented in Haskell).
4. `wasp-ai/` dir got renamed to `mage/`.
5. Added README.md to `mage/`.
6. Updated all READMEs (/README.md, /waspc/README.md) with relevant info about mage/. Also updated docs.
7. Froze `wasp-ai` branch, from now on Mage will be deployed via `release` branch same as docs are deployed. Documented this in waspc/README.md.
8. Updated Mage to work with these newest changes.

Some screenshots of using Wasp AI via `wasp new`, interactively:

ai1
ai2
ai3


mage/README.md
```diff
@@ -0,0 +1,44 @@
...
+So if you want to deploy new version of Mage, you should get it in wanted state on `release` branch, and then deploy from there.
+
+Also, before deploying, check that version of `wasp` in `Dockerfile` makes sense.
+
+To deploy it, just run `wasp deploy fly deploy`. You might want to add `--org wasp` if needed.
+
+## FAQ
+
+Q: What is the difference between Wasp AI and Mage? Are those the same thing?
+A: We say "Wasp AI" we refer to logic implemented in `wasp` CLI, while when we say "Mage" we refer to the Mage web app that really serves as a client for "Wasp AI" (calls it in the background). That said, we sometimes use these interchangeably.
```
Martinsos, line 44:
```suggestion
A: When we say "Wasp AI" we refer to logic implemented in `wasp` CLI, while when we say "Mage" we refer to the Mage web app that really serves as a client for "Wasp AI" (calls it in the background). That said, we sometimes use these interchangeably.
```

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
```diff
@@ -80,7 +81,7 @@ defaultStarterTemplate =
       }
 
 findTemplateByString :: [StarterTemplate] -> String -> Maybe StarterTemplate
-findTemplateByString templates query = find ((== query) . show) templates
+findTemplateByString templates query = find ((query `isPrefixOf`) . show) templates
```
Martinsos, line 84: @infomiho I made the change both here, and in Interactive, so that matches happen on prefix, not on the exact match, because I thought that will be more practical for users + it allowed to putting stuff like emojis at the end of the template title (one for AI).

waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -72,7 +87,7 @@ askToChoose question options = do
         _invalidIndex -> Nothing
 
     getOptionByName :: String -> Maybe o
-    getOptionByName name = find ((== name) . showOption) options
+    getOptionByName name = find ((name `isPrefixOf`) . showOption) options
```
Martinsos, line 90: I made it match by prefix, not just exact match @infomiho

waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -83,18 +98,30 @@ askToChoose question options = do
...
-            indexPrefix = showIndex idx <> " "
+            indexPrefix =
+              Term.applyStyles
+                ([Term.Yellow] <> whenDefault [Term.Bold, Term.Underline])
+                (showIndex idx)
+                <> " "
+            optionName = Term.applyStyles [Term.Bold] (showOption option)
+            tags = whenDefault (Term.applyStyles [Term.Yellow] " (default)")
+            optionDescription = showDescription option
+            whenDefault xs = if isDefaultOption option then xs else mempty
```
Martinsos, line 116: I enriched the styling of options a bit, so now `(default)` tag is also yellow, and the index `[i]` of the default option is bold and underlined.

waspc/src/Wasp/AI/GenerateNewProject.hs
```diff
@@ -28,65 +40,79 @@ generateNewProject ::
...
-  CodeAgent ()
+  CodeAgent LogMsg ()
 generateNewProject newProjectDetails waspProjectSkeletonFiles = do
-  writeToLog . T.pack $
-    "Generating a new Wasp project named " <> _projectAppName newProjectDetails <> "!"
+  writeToLog $
+    "\nGenerating a new Wasp project named " <> L.styled L.Important (fromString $ _projectAppName newProjectDetails) <> "!"
 
-  writeToLog "Generating project skeleton..."
+  let showParams chatGPTParams =
```
Martinsos, line 48: Here I added some extra info: I show GPT model and temeprature

waspc/src/Wasp/AI/GenerateNewProject/LogMsg.hs
```diff
@@ -0,0 +1,55 @@
...
+    fromText,
+    Style (..),
+  )
+where
+
+import Data.String (IsString (fromString))
+import qualified Data.Text as T
+import qualified Wasp.Util.Terminal as Term
+
+data LogMsg = Plain String | Concat [LogMsg] | Styled Style LogMsg
```
Martinsos, line 15: This is our "Decorated String" type.

waspc/src/Wasp/AI/OpenAI/ChatGPT.hs
```diff
@@ -53,18 +53,17 @@ queryChatGPT apiKey params requestMessages = do
...
-          <> "\n===== GPT PARAMS ======\n"
-          <> show params
-          <> "\n====== REQUEST =======\n"
-          <> show requestMessages
-          <> "\n====== RESPONSE ======\n"
-          <> show chatResponse
-          <> "\n==================================\n\n\n\n"
-      )
-      $ return ()
+  debugTrace
```
Martinsos, line 56: We had this printing all the time before. Now, I made it so that it prints only when DEBUG env vars is set.

waspc/src/Wasp/Util/Debug.hs
```diff
@@ -0,0 +1,23 @@
+{-# OPTIONS_GHC -Wno-deprecations #-}
```
Martinsos, line 1: OK I went a bit wild here, I added this "fake" pure function that checks if we are in debug mode.

I spent some time investigating other ways to accomplish running code conditionally depending on DEBUG env var, which included stuff like cabal flags and compiler pragmas, but all the solutions ended up boing overly complex, and this just seems in practice and at the moment to be the easiest. We can revisit it in the future but I would waste time trying to find something "cleaner". I did read online that doing unsafe lookup of env var like this shouldn't be able to cause any problems, so we should be ok.

waspc/src/Wasp/Util/Terminal.hs
```diff
@@ -18,13 +20,15 @@ data Style
...
+  | Blink
+  deriving (Show, Eq)
 
 -- | Given a string, returns decorated string that when printed in terminal
 -- will have same content as original string but will also exibit specified styles.
 applyStyles :: [Style] -> String -> String
 applyStyles [] str = str
 applyStyles _ "" = ""
-applyStyles styles str = foldl applyStyle str styles ++ escapeCode ++ resetCode
+applyStyles styles str = foldl' applyStyle str styles ++ escapeCode ++ resetCode
```
Martinsos, line 31: Recent lesson about `fold`s educated me enough to fix this :D.

waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -37,22 +39,35 @@ import qualified Wasp.Util.Terminal as Term
...
+class IsOption o where
   showOption :: o -> String
   showOptionDescription :: o -> Maybe String
 
-instance Option [Char] where
+instance IsOption [Char] where
   showOption = id
   showOptionDescription = const Nothing
 
+data Option o = Option
```
Martinsos, line 54: To learn about the motivation for this change, check comment https://github.com/wasp-lang/wasp/pull/1601/files#r1426920624 .

waspc/cli/src/Wasp/Cli/Interactive.hs
infomiho: Screenshot 2023-12-19 at 12 40 28

I'd remove the Bold and Underline for the index, it confused me for a second that I can move the options with my keyboard which I couldn't 😢  Until we have that, let's maybe not do anything different for the index. 

I like the yellow `(default)` 👍

waspc/cli/src/Wasp/Cli/Interactive.hs
infomiho: Aren't we afraid of users selecting the wrong thing sometimes e.g. 

- `What do you want to buy?`
- `car` vs. `carrot`

Type type in `car` but maybe get a `carrot`?

waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -83,18 +98,30 @@ askToChoose question options = do
...
+            whenDefault xs = if isDefaultOption option then xs else mempty
 
         showIndex i = "[" ++ show (i :: Int) ++ "]"
 
-        showDescription option indentLength = case showOptionDescription option of
+        showDescription option = case showOptionDescription option of
           Just description -> "\n" <> replicate indentLength ' ' <> description
           Nothing -> ""
+          where
+            indentLength = 6
```
infomiho, line 124: I preferred the look of the description being aligned below the title 🤷 

Now it's a bit to the right 
Screenshot 2023-12-19 at 12 40 28

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
infomiho: What if we have some sort of overlap with the names like `saas-minimal` and `saas-stripe`?

waspc/src/Wasp/Util/Terminal.hs
```diff
@@ -39,6 +43,7 @@ styleCode Cyan = "[36m"
 styleCode White = "[37m"
 styleCode Bold = "[1m"
 styleCode Underline = "[4m"
+styleCode Blink = "[5m"
```
infomiho, line 46: This might not work in some terminal emulators 😢

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
Martinsos: Then it will take the first one, which I think is ok?

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/AI.hs
```diff
@@ -6,32 +6,110 @@ module Wasp.Cli.Command.CreateNewProject.AI
...
+import qualified Wasp.Util.Terminal as T
 
 createNewProjectInteractiveOnDisk :: Path' Abs (Dir WaspProjectDir) -> NewProjectAppName -> Command ()
 createNewProjectInteractiveOnDisk waspProjectDir appName = do
   openAIApiKey <- getOpenAIApiKey
   appDescription <- liftIO $ Interactive.askForRequiredInput "Describe your app in a couple of sentences"
-  liftIO $ createNewProjectOnDisk openAIApiKey waspProjectDir appName appDescription emptyNewProjectConfig
+  (planningGptModel, codingGptModel) <-
+    liftIO $
+      Interactive.askToChoose'
```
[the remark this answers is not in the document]
infomiho: Makes sense 👍 
`
One weird thing for me is the use of `askToChoose'` vs. `askToChoose` as the exposed function. Can we change that?

waspc/src/Wasp/AI/GenerateNewProject/LogMsg.hs
```diff
@@ -0,0 +1,55 @@
...
+toTermString (Styled s m) =
+  Term.applyStyles (styleToTermStyles s) $ toTermString m
+  where
+    styleToTermStyles = \case
+      Important -> [Term.Bold, Term.Magenta]
+      Generating -> [Term.Blue]
+      Fixing -> [Term.Green]
+      Error -> [Term.Red]
+      Custom styles -> styles
+toTermString (Concat ms) = concat $ toTermString <$> ms
```
infomiho, line 55: `map` would make this more readable on first read 👍

waspc/src/Wasp/AI/CodeAgent.hs
```diff
@@ -35,33 +35,33 @@ import qualified Wasp.Util.IO.Retry as R
 import Wasp.Util.Network.HTTP (catchRetryableHttpException)
 import qualified Wasp.Util.Network.HTTP as Utils.HTTP
 
-newtype CodeAgent a = CodeAgent {_unCodeAgent :: ReaderT CodeAgentConfig (StateT CodeAgentState IO) a}
-  deriving (Monad, Applicative, Functor, MonadIO, MonadReader CodeAgentConfig, MonadState CodeAgentState)
+newtype CodeAgent logMsg a = CodeAgent {_unCodeAgent :: ReaderT (CodeAgentConfig logMsg) (StateT CodeAgentState IO) a}
```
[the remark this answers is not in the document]
infomiho: I don't understand the `logMsg` parameter, it has no constraints on it?

waspc/cli/src/Wasp/Cli/Interactive.hs
Martinsos: Hm good point. They will get the first one that matches the prefix. But I haven't thought of this case, this is a problem! What should we do? Revert back? Or I can make it so it matches on exact match first, but if none, it matches by prefix?

Martinsos: Ok, I reverted it to as it was before.

waspc/cli/src/Wasp/Cli/Interactive.hs
Martinsos: Good point, done!

waspc/src/Wasp/AI/CodeAgent.hs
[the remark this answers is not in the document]
Martinsos: So `CodeAgent` here is a standalone module, it could be in theory extracted in to a library -> it doesn't know anything about how we use it in Mage.

I could have made it directly use LogMsg data type which comes with it, but I don't really need to be that specific, instead I use anything that implements IsString -> that is enough for CogeAgent to know what to do with it. This in theory allows using CodeAgent with one type of log messages in one place, and with other kind of log message in some other use case.

Now, why there are no restrictions on the data type -> that is a bit non-instinctive in Haskell, but actually it is much better to put type constraints on functions that use the data type, not on data type itself. SO you will see I have `IsString` constraint for `logMsg` on some functions in this file. This has some benefits to it and is how it is done. If I am correct, putting constraints on the data type actually won't even have the wanted effect. We can discuss this more if you are interested into this topic.

waspc/src/Wasp/Util/Terminal.hs
Martinsos: I added a comment about that.

waspc/src/Wasp/AI/GenerateNewProject/LogMsg.hs
Martinsos: I find this more readable! Experience. But I suggest trying to get into it, I think once you get used to it, code gets easier to read. `<$>` is very common in Haskell, and you should get used to being able to read it quickly.

One thing I want to emphasize is that in Haskell, I most often don't bother understanding what every symbol does. I know compiler has my back and that stuff is "connected" correctly. What I care about is the general flow of data and what is happening.

So for example, when reading this code, I read it like this:
> ms is taken out of Concat, then toTermString is applied to it, and then concat is applied to that.

Notice that I don't actually care too much, at least in the first read, if it is `<$>` or `$`, I see it all as application. It doesn't really matter to me, I know that the thing on the right is now "term stringified" and that at the end all stuff is concatenated.

And when reading code like that, it is easier to read it when using `$` and `<$>` then when using `fmap` or `map` because they break my reading flow and force me to think about them.

Same is if I encounter an expression with a bunch of `>>=`, `<$>`, `$`, `>>`, `&`, ... .
Most of the time, I just need a general idea of what it does, and for that I don't need to understand exactly why each of those operators was used in that specific place. If I need to modify it, then I will dig a bit deeper, and again compiler will be helping me out.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/AI.hs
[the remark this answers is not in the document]
Martinsos: Makes sense, but both are valid choices to use -> one is for when dealing with Option, another is for more general case.

WHy not use `askToChoose` for all of them? Beacuse it would return `Option ...` and when you are using `Option`, you really want to get the value, not the whole option.

Since they are so similar, I thought `'` works fine ok, it is a specialized version of `askToChoose`.

Martinsos:
@infomiho I responded to all comments and fixed stuff, pls give it another look!

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/AI.hs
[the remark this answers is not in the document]
infomiho: Okay, makes sense 👍

waspc/cli/src/Wasp/Cli/Interactive.hs
infomiho: Okay, it's obviously subjective, so I want insist on this 👍

infomiho:
Looking good 👍

The indent change is a bit subjective, but I prefer the old style.

The `fmap` vs `map` is a bit subjective as well, so I won't press on that matter too hard.

waspc/cli/src/Wasp/Cli/Interactive.hs
infomiho: I'd maybe look at other CLI tools how they indent their titles and descriptions. 

I found this in Astro's CLI 
Screenshot 2023-12-21 at 15 18 14

Martinsos: Just a reminder, this is what it looks like with old indent:
image

This does look harder to read to me, because all text is so crammed up. In the example above from Astro, they have newlines, so that solves the issue, detaches title from the text. We don't have that here, nor would it be a good fit.

That said, since it is not obvious that change I made is better (both to you and @sodic ), then I will revert it to previous state.

Martinsos: Done!

Martinsos:
> Looking good 👍
> 
> The indent change is a bit subjective, I prefer the old style 🤷
> 
> The `fmap` vs `map` is a bit subjective as well, so I won't press on that matter too hard.

Thanks! I reverted the indent change, and we agreed we will let people use what they want for map/fmap/<$>, so this should be it! If anything else let me know, otherwise I will merge once you approve.

infomiho (approved):
🚢  it!

waspc/src/Wasp/AI/GenerateNewProject/LogMsg.hs
infomiho: I still think that `map` is more easier to read for developers that have < 2 years of Haskell experience. But as we talked about it IRL, this seems a matter of personal preference and as I said already, I won't press this matter any more.

PR #753, opened 2022-10-14
Separate user code into client and server directories

sodic:
Separate client code and server code.

Fixes #522
Fixes #594

Takes a step towards implementing #710.

@Martinsos and @shayneczyzewski, feel free to start reviewing this. I've not yet updated e2e tests, example projects, or the docs. I'll do this once we settle on a structure. The current approach is:
```
ProjectRoot
├── client
│   └── .waspignore (and client source files)
├── .gitignore
├── main.wasp
├── .env.server
├── .env.client
├── server
│   └── .waspignore (and server source files)
└── .wasproot
```
I'd probabaly like to move the env files inside `client` and `server`

PR #1602, opened 2023-12-18
Implement the project install step

### LINKED ISSUE #1597: Support project dependencies in Wasp
As described in #1595, Wasp projects how have a top-level _package.json_ (and *node_modules*).

Modify the generator to account for that:
- Add an extra install step during `wasp start`
- Decide what to do with the `dependencies` field in `.wasp` files
- Update the docs.

sodic:
Fixes #1597.

Tasks:
- [x] Add an extra install step during `wasp start`
- [x] Decide what to do with the `dependencies` field in `.wasp` files - I decided to remove the field and only keep a single source of truth, the `_package.json_ file.
- [x] Update the docs.
- [x] Fix wasp not installing npm deps that are in package.json but are not installed. -> fixed in https://github.com/wasp-lang/wasp/pull/1652 .
- [ ] TODO: https://github.com/wasp-lang/wasp/pull/1602/files#diff-aa79f536fd3bb9203f9d976c63594a52a9f3533759e2b754e2a6baec6403a7dbR95-R96 -> this is the one that the only unresolved comment is about.

How to run and test stuff:
- Go to `waspc/examples/todo-typescript` and work inside that project.
- Running `./cleanstart` will clean everything, fix known issues (e.g., duplicate react) and start the app with all the necessary migrations applied.
- The rest of the commands (e.g,. `cabal run wasp-cli start`, `cabal run wasp-cli clean`) will do what you expect.
- After each change, either let the application pick it up, rerun `cabal run wasp-cli start`, or run `./cleanstart` (depending on what you're changing).
- You can run `./fix` to fix the double react installation issue (but often won't need to, as `./cleanstart` takes care of it). 
- Please don't commit the migrations or `package-lock.json`.

> [!IMPORTANT]  
> **PR 1603** (`filip-wasp-reset`) branches out of **PR 1604** (`filip-restructuring-prisma`), which branches out of **PR 1602** (`filip-project-install-step`).
> The PR numbers are unfortunately out of order.
> Make sure you keep the branches up to date with their bases as you change stuff.


waspc/src/Wasp/Generator/NpmDependencies.hs
```diff
@@ -141,9 +141,11 @@ combineNpmDepsForPackage npmDepsForWasp npmDepsForUser =
   if null conflictErrors && null devConflictErrors
     then
       Right $
+        -- todo(filip): check whether dependency updates and npm install work properly
+        -- todo(filip): reconsider whether we want to change the {sever,web-app}/package.json dynamically
         NpmDepsForPackage
-          { dependencies = waspDependencies npmDepsForWasp ++ remainingUserDeps,
-            devDependencies = waspDevDependencies npmDepsForWasp ++ remainingUserDevDeps
+          { dependencies = Map.elems remainingWapsDeps,
+            devDependencies = Map.elems remainingWaspDevDeps
```
sodic, line 148: Let's discuss what we should do here. As a first step, of course. Option 1 is the better long-term solution.  
1. Always specify the same dependencies in `{sever,web-app}/package.json` and alias them. Wasp code will use our version of the dependencies, while the user's code will use their versions.
2. For each dependency, decide which version we allow.
   - If the user doesn't specify that dependency in their `package.json`, Wasp lists it in `{sever,web-app}/package.json`.
   - If the user specifies that dependency with the correct version in their `package.json`, Wasp installs it in the top-level `node_modules` and doesn't list it in `{sever,web-app}/package.json` (this is what this PR implements).
   - If the user specifies that dependency with the wrong version in their `package.json`, Wasp reports a conflict and refuses to build (this is what we do in the old structure).

Martinsos: Hm I feel like I am missing information to make a decision here.

If option 1 is better long term, why not go with it now? I guess there are is a cost to implementation that solution vs what you implemented so far? What is that cost? What are pros/cons of each option?

In option 2, you said this PR doesn't list their dep in server|web-app/package.json since it will pull it from their node_modules instead. But you don't say what happens if they specify wrong version -> you said what happened before, not what happens now.

waspc/src/Wasp/AppSpec/PackageJson.hs
```diff
@@ -0,0 +1,31 @@
...
+
+$(deriveJSON defaultOptions {fieldLabelModifier = drop 1} ''PackageJson)
+
+dependencies :: PackageJson -> [Dependency]
+dependencies packageJson = D.fromList $ M.toList $ _dependencies packageJson
+
+devDependencies :: PackageJson -> [Dependency]
+devDependencies packageJson = D.fromList $ M.toList $ _devDependencies packageJson
+
+allDependencies :: PackageJson -> [Dependency]
```
Martinsos, line 30: Why would somebody want to import all deps and lose info about which ones are dev and which ones are non-dev?

waspc/src/Wasp/AppSpec/PackageJson.hs
```diff
@@ -0,0 +1,31 @@
...
+
+data PackageJson = PackageJson
+  { _name :: !String,
+    -- todo(filip): do this properly once you merge martin's PR
+    _dependencies :: !(Map String String),
+    _devDependencies :: !(Map String String)
+  }
+  deriving (Show, Generic)
+
+$(deriveJSON defaultOptions {fieldLabelModifier = drop 1} ''PackageJson)
```
Martinsos, line 22: Oho this is cool, wasn't aware of this!

What is not great here, is that if somebody renames those files to have different prefix (maybe `_` gets removed, or it now has `__`), this will stop working correctly but we will not know. Tests would help, but I don't think you added any (yet)?

Maybe just do it the manual way, so we have type safety? Or you could at least do a bit more involved logic, that will map from "_name" to "name" and so on so it is more visible, or you could check for `_` upfront and drop it only if it exists, ... .

But in the rest of AppSpec data we don't use `_` as prefix, do we? So how about we also not use it here and you rename those additional getters below?

Actually, why are you even storing these as `Map`? Why not as `[Dependency]` if that is what we had before? And if you really do want to store it as you do now, why do you bother with transforming it into `[Dependency]` here, why not do that in the place where `[Dependency]` is needed? Because at this point it seems like AppSpec kind of supports both formats, but that doesn't make much sense to me. I would either go with what you have and remove notion of `Dependency` fromAppSpec, or I would use `Dependency` entirely.

waspc/src/Wasp/Generator/Job/IO/PrefixedWriter.hs
```diff
@@ -172,8 +172,8 @@ makeJobMessagePrefix jobMsg =
   (T.pack . buildPrefix . concat)
     [ [("[", jobStyles)],
       [(jobName, jobStyles)],
-      styledFlags,
-      [("]", jobStyles)]
+      [("]", jobStyles)],
+      styledFlags
```
Martinsos, line 176: What happened here, why this change? If I am correct, this results in having `[Wasp]!` instead of `[Wasp!]`?

Also, I see you added padding below to names.

I am not sure if I like these changes or not -> previous choices for not having padding and having this exclamation mark inside were not made randomly, so I would like to discuss this. But let's not do this inside of this PR then, you can do this when you will go about improving CLI as you said you would like to do.

waspc/src/Wasp/Generator/NpmDependencies.hs
```diff
@@ -108,9 +107,10 @@ buildNpmDepsForFullStack spec forServer forWebApp =
 getUserNpmDepsForPackage :: AppSpec -> NpmDepsForUser
 getUserNpmDepsForPackage spec =
   NpmDepsForUser
-    { userDependencies = fromMaybe [] $ AS.App.dependencies $ snd $ ASV.getApp spec,
+    { -- todo(filip): what if package.json has no dependencies field?
```
Martinsos, line 110: Hm good question. Well that is same as [] right?

waspc/src/Wasp/Generator/NpmDependencies.hs
Martinsos, line 1: I remember this file being quite involved, so I didn't dive into it now, which means I also miss the context to evaluate the changes done here. Is there anything here you want me to look at specifically, or do you think it is all very clear and no need for me to look at it?

waspc/src/Wasp/Generator/NpmInstall.hs
```diff
@@ -100,34 +109,81 @@ loadInstalledFullStackNpmDependencies dstDir = do
...
+        "Installation going great - we'll get there soon!",
+        "The installation is taking a while, but we'll get there!",
+        "Yup, still not done installing.",
+        "We're getting closer and closer, everything will be installed soon!",
+        "Still waiting for the installation to finish? You should! We got too far to give up now!",
+        "You've been waiting so patiently, just wait a little longer (for the installation to finish)..."
+      ]
+
+installNpmDependenciesAndReport :: Job -> Chan JobMessage -> JobType -> IO ExitCode
+installNpmDependenciesAndReport installF chan jobType = do
```
Martinsos, line 133:
```suggestion
installNpmDependenciesAndReport installJob chan jobType = do
```

waspc/src/Wasp/Generator/NpmInstall.hs
```diff
@@ -100,34 +109,81 @@ loadInstalledFullStackNpmDependencies dstDir = do
...
+  writeChan chan $ J.JobMessage {J._data = J.JobOutput "Starting npm install\n" J.Stdout, J._jobType = jobType}
+  result <- installF chan `race` reportInstallationProgress chan jobType
+  case result of
+    Left exitCode -> return exitCode
+    Right _ -> error "This should be impossible"
```
Martinsos, line 138:
```suggestion
    Right _ -> error "This should never happen, reporting installation progress should run forever."
```

waspc/src/Wasp/Generator/NpmInstall.hs
```diff
@@ -100,34 +109,81 @@ loadInstalledFullStackNpmDependencies dstDir = do
...
+    _ -> do
+      let handleMessagesJob = handleJobMessages messagesChan
+      let runSetupJobs =
+            concurrently
+              (installNpmDependenciesAndReport (ServerSetup.installNpmDependencies dstDir) messagesChan J.Server)
+              (installNpmDependenciesAndReport (WebAppSetup.installNpmDependencies dstDir) messagesChan J.WebApp)
+      (_, results) <- concurrently handleMessagesJob runSetupJobs
+      case results of
+        (ExitSuccess, ExitSuccess) -> return $ Right ()
+        exitCodes -> return $ Left $ setupFailedMessage exitCodes
```
Martinsos, line 161: Possibly nicer:

```suggestion
      (handleMessagesJob `concurrently` runSetupJobs) >>= \case
        (ExitSuccess, ExitSuccess) -> return $ Right ()
        exitCodes -> return $ Left $ setupFailedMessage exitCodes
```

You could also maybe extract that return in front of case: `return . \case`, I think that should compile.

waspc/src/Wasp/Generator/NpmInstall.hs
```diff
@@ -100,34 +109,81 @@ loadInstalledFullStackNpmDependencies dstDir = do
...
   where
-    handleJobMessages = runPrefixedWriter . go (False, False)
+    handleProjectInstallMessage :: Chan J.JobMessage -> IO ()
```
Martinsos, line 163:
```suggestion
    handleProjectInstallMessages :: Chan J.JobMessage -> IO ()
```

waspc/src/Wasp/Generator/NpmInstall.hs
```diff
@@ -100,34 +109,81 @@ loadInstalledFullStackNpmDependencies dstDir = do
...
+    handleProjectInstallMessage = runPrefixedWriter . processMessages
+      where
+        processMessages :: Chan J.JobMessage -> PrefixedWriter ()
+        processMessages chan = do
+          jobMsg <- liftIO $ readChan chan
+          case J._data jobMsg of
+            J.JobOutput {} -> printJobMessagePrefixed jobMsg >> processMessages chan
+            J.JobExit {} -> return ()
+    handleJobMessages = runPrefixedWriter . processMessages (False, False)
```
Martinsos, line 172: We have `handleProjectInstallMessage` which is specific, but then `handleJobMessages` name is generic. It should also be specific instead, something like `handleSetupJobsMessages` or whatever you think describes well the two jobs that install server and web app.

waspc/src/Wasp/Generator/NpmInstall.hs
Martinsos, line 161: I would consider organizing this into two functions, one that does the whole thing for first step, and the other that does the whole thing for second step, so you can then make this much easier to read.

waspc/src/Wasp/Generator/NpmInstall.hs
```diff
@@ -100,34 +109,81 @@ loadInstalledFullStackNpmDependencies dstDir = do
...
               printJobMessagePrefixed jobMsg
-                >> go (isWebAppDone, isServerDone) chan
+                >> processMessages (isWebAppDone, isServerDone) chan
             J.JobExit {} -> case J._jobType jobMsg of
-              J.WebApp -> go (True, isServerDone) chan
-              J.Server -> go (isWebAppDone, True) chan
+              J.WebApp -> processMessages (True, isServerDone) chan
+              J.Server -> processMessages (isWebAppDone, True) chan
               J.Db -> error "This should never happen. No db job should be active."
+              J.Wasp -> error "This should never happen. No db job should be active."
```
Martinsos, line 186:
```suggestion
              J.Wasp -> error "This should never happen. No Wasp job should be active."
```

waspc/src/Wasp/Generator/SdkGenerator.hs
```diff
@@ -87,3 +90,9 @@ sdkRootDirInProjectRootDir = [reldir|sdk/wasp|]
 
 sdkTemplatesDirInTemplatesDir :: Path' (Rel TemplatesDir) (Dir SdkTemplatesDir)
 sdkTemplatesDirInTemplatesDir = [reldir|sdk|]
+
+-- todo(filip): figure out where this belongs
```
Martinsos, line 94: Can I help with this somehow?

waspc/src/Wasp/Project/Analyze.hs
```diff
@@ -86,12 +97,26 @@ constructAppSpec waspDir options decls = do
     [] -> Right appSpec
     validationErrors -> Left $ map show validationErrors
 
-findWaspFile :: Path' Abs (Dir WaspProjectDir) -> IO (Either String (Path' Abs File'))
+findWaspFile :: Path' Abs (Dir WaspProjectDir) -> IO (Maybe (Path' Abs File'))
 findWaspFile waspDir = do
   files <- fst <$> IOUtil.listDirectory waspDir
-  return $ maybeToEither "Couldn't find a single *.wasp file." $ (waspDir </>) <$> find isWaspFile files
+  return $ (waspDir </>) <$> find isWaspFile files
```
Martinsos, line 103: Why did you decide to take this error message out and then apply it out there? It is not terrible, but it seemed to me that it makes more sense for it to be generated here, where there is more context on what happened.

waspc/src/Wasp/Project/Analyze.hs
```diff
@@ -86,12 +97,26 @@ constructAppSpec waspDir options decls = do
...
         && (length (toFilePath path) > length (".wasp" :: String))
+
+analyzePackageJsonContent :: Path' Abs (Dir WaspProjectDir) -> IO (Either [CompileError] PackageJson)
+analyzePackageJsonContent waspProjectDir =
+  findPackageJsonFile >>= \case
+    Just packageJsonFile -> readPackageJsonFile packageJsonFile
+    Nothing -> return $ Left [fileNotFoundMessage]
+  where
+    fileNotFoundMessage = "couldn't find package.json file in the " ++ toFilePath waspProjectDir ++ " directory"
+    findPackageJsonFile = findFileInWaspProjectDir waspProjectDir [relfile|package.json|]
```
Martinsos, line 117: So this path to package.json we have here, how about we take that a bit higher up somewhere, so others can also access it if needed? Feels like it should be global knowledge, same like position of env files or dockerfile, but it is embedded here into `where`.

waspc/src/Wasp/Project/Analyze.hs
```diff
@@ -86,12 +97,26 @@ constructAppSpec waspDir options decls = do
...
+
+readPackageJsonFile :: Path' Abs File' -> IO (Either [CompileError] PackageJson)
+readPackageJsonFile packageJsonFile = do
+  byteString <- IOUtil.readFileBytes packageJsonFile
+  return $ maybeToEither ["Error reading the package.json file"] $ Aeson.decode byteString
```
Martinsos, line 122:
```suggestion
  return $ maybeToEither ["Error decoding the package.json file"] $ Aeson.decode byteString
```

waspc/src/Wasp/AppSpec/PackageJson.hs
sodic: I added this to make it backwards compatible with our current system that doesn't support dev deps (i.e., I added a layer that ignores the distinction on purpose).

However, I have since fully abandoned the legacy system of not differentiating between dependencies, so I'll remove this function.

waspc/src/Wasp/Generator/Job/IO/PrefixedWriter.hs
sodic: Makes sense, I'll move it.

waspc/src/Wasp/Project/Analyze.hs
sodic: Hm, not sure about this one. "Decoding" sounds very technical, while "Reading" implies that the file wasn't found. Perhaps "parsing" is the best choice?

waspc/src/Wasp/Project/Analyze.hs
sodic: I agree, I'll move it.

waspc/src/Wasp/Generator/NpmInstall.hs
sodic: Yes, I'll do that.

I kinda forgot to do my final self-review pass on this PR, as you probably noticed.

waspc/src/Wasp/Generator/NpmInstall.hs
sodic: Sure, makes sense!

waspc/src/Wasp/Generator/NpmDependencies.hs
sodic: Yes, but I'm wondering if I should model it with a `Maybe` inside AppSpec (spoiler: yes, probably, but I'll try to break it first).

waspc/src/Wasp/Generator/SdkGenerator.hs
sodic: Yes, of course.

Where would you put it?

waspc/src/Wasp/Project/Analyze.hs
sodic: I wanted to do it for a long time now, when you suggested this: https://github.com/wasp-lang/wasp/pull/753#discussion_r1004229754.

You might have meant something else, but `findWaspFile :: Path' Abs (Dir WaspProjectDir) -> IO (Maybe (Path' Abs File'))` seems like the cleanest name/type combo.

waspc/src/Wasp/Generator/NpmDependencies.hs
Martinsos: Bump on this one!

waspc/data/Cli/templates/basic/package.json
```diff
@@ -2,9 +2,10 @@
   "name": "prototype",
   "dependencies": {
     "wasp": "file:.wasp/out/sdk/wasp",
-    "react": "18.2.0"
+    "react": "^18.2.0"
   },
   "devDependencies": {
-    "@types/react": "^18.0.37"
+    "@types/react": "^18.0.37",
+    "prisma": "4.16.2"
```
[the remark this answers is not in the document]
Martinsos: I did completely forget about that. I think we should add a comment! Would you be up for adding it somewhere? I am not sure what is the best place though, but I am sure you will figure it out.

waspc/src/Wasp/AppSpec.hs
```diff
@@ -56,6 +57,8 @@ import Wasp.Project.Db.Migrations (DbMigrationsDir)
 data AppSpec = AppSpec
   { -- | List of declarations like App, Page, Route, ... that describe the web app.
     decls :: [Decl],
+    -- | The contents of the package.json file found in the root directory of the wasp project.
+    packageJson :: PackageJson,
```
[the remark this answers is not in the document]
Martinsos: Why do you want to change the type name? It sounds like a good type name to me.

waspc/src/Wasp/AppSpec/PackageJson.hs
```diff
@@ -0,0 +1,31 @@
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE TemplateHaskell #-}
+
+module Wasp.AppSpec.PackageJson where
+
+import Control.Applicative (liftA2)
+import Data.Aeson.TH
```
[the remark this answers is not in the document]
Martinsos: Reminder to put :rocket: once you are done, :+1: means "Got it", but not "Done" -> you are likely aware of this and just didn't get to having it done yet, but I am still writing this just to be safe.

waspc/src/Wasp/AppSpec/PackageJson.hs
Martinsos: Got it! Feel free then to resolve this comment once you don't need it any more.

waspc/src/Wasp/AppSpec/PackageJson.hs
Martinsos: Oh ok, sounds good! A TODO here would have helped, as I would be aware of your intentions then + it would increase the likelihood of you not forgetting to do it.

waspc/src/Wasp/Generator/NpmDependencies.hs
Martinsos: Ah got it! Yeah, you probably should, as that provides a bit more info, but, in practice, I don't think it will ever matter.

waspc/src/Wasp/Generator/NpmInstall.hs
Martinsos: No worries, haven't really noticed!

waspc/src/Wasp/Project/Analyze.hs
Martinsos: Yeah that is what I meant, damn me from past :D. But I was right I guess. Both approaches make sense in a way to me now. Ok, all good.

waspc/src/Wasp/Project/Analyze.hs
Martinsos: I agree!

waspc/src/Wasp/Util/IO.hs
```diff
@@ -95,6 +97,9 @@ doesFileExist = SD.doesFileExist . SP.fromAbsFile
 readFile :: Path' Abs (File f) -> IO String
 readFile = P.readFile . SP.fromAbsFile
 
+readFileBytes :: Path' Abs (File f) -> IO B.ByteString
+readFileBytes = B.readFile . SP.fromAbsFile
```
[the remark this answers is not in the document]
Martinsos: Hm, I am not sure. It is not really under the hood, you are returning a type that is Lazy ByteString, that type is not the same as Strict ByteString, so there won't be confusion later. But in general, lazy IO is source of errors often as it is a bit weird to grasp. So I would potentially go with strict as default. At the end, depends where it is used. I wonder now about that readFile above we have -> that is also probably lazy? Ha I am not sure. Actually yeah it likely is, because below we have readFileStrict.
Ok, actually I think what you did is fine. Usually IO is assumed to be lazy unless you add Strict or `'` to it, so this sounds good. As I said, type system should catch this I think, to some degree, I am not sure to which degree but I know haskell complains when I try to mix lazy and strict ByteString (or did I remember that wrong? I don't think so).

waspc/src/Wasp/Generator/SdkGenerator.hs
Martinsos: Hm I am kind of not sure what is this -> so we are installing deps for user's code, right? In the top-level node modules? Hmmmmm. This sounds like it is not even related to Generator maybe? Oh yeah this is a bit tricky to figure out where to put. I would say either in Wasp.Project somewhere, but you will need stuff from Generator right, so that makes it a no-no mostly, we don't really to import from Generator there mostly.
Although we could argue that `runNodeCommandAsJob` maybe should go outside of Generator, if it is used by other parts of Wasp also, and doesn't depend on Generator! So in that case it makes sense.
Or, we just stick to Generator for now, that is not terrible, but we put this somewhere at its high level. Maybe create a new module in it called WaspProject or maybe put it in Common or some other module that might make sense. This feels a bit like hacking it, but is survivable I think.

Martinsos (commented):
Did another round!

waspc/src/Wasp/Generator/NpmDependencies.hs
sodic: > If option 1 is better long term, why not go with it now? I guess there are is a cost to implementation that solution vs what you implemented so far?

Yes - I've never worked with package aliasing and would have to look into it. Also, we'd have to change more code than if we didn't go with it.

The implementation cost is the only con of this approach I currently know of, but there could be some caveats with aliased pacakges I'm not aware of.

> In option 2, you said this PR doesn't list their dep in server|web-app/package.json since it will pull it from their node_modules instead. But you don't say what happens if they specify wrong version -> you said what happened before, not what happens now.

I've edited my original comment to make the wording more clear. Sorry :upside_down_face:

waspc/data/Cli/templates/basic/package.json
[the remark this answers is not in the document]
sodic: Sure, I'll add it in.

waspc/src/Wasp/AppSpec.hs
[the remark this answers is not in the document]
sodic: I guess it's fine after all :smile: 

I just found it weird that it has "Json" in its name while not being a JSON string.

waspc/src/Wasp/AppSpec/PackageJson.hs
sodic: True... I guess I was just so happy I finally got it working, I just immediately closed my computer and left :smile:

waspc/src/Wasp/Generator/NpmDependencies.hs
Martinsos: So you have option 2 working now, right? How much time do you estimate to get option 1 working? Option 1 sounds like it works better + is quite simpler, so it is clear winner, as you already said. I would say go for it, but obviously you would go for it already if you thought it is worth the time. So I am assuming instead you would like to do it a bit later? If so, I am also ok with that.
To conclude, I vote for option 1 but obviously you want to do it later so if so, sure.

waspc/src/Wasp/AppSpec.hs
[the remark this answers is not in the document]
Martinsos: Ah good point, but in this case I think the name is so popular I think it is ok.

waspc/src/Wasp/Generator/SdkGenerator.hs
sodic: To whoever's doing this...

Please don't touch this for now. I'll handle it on a different branch. Leave the comment unresolved.

waspc/src/Wasp/Project/Analyze.hs
sodic: Verdict: we'll change it to `Reading`.

waspc/src/Wasp/Generator/NpmDependencies.hs
Martinsos: Ok, create an issue here: https://github.com/wasp-lang/wasp/issues/1644 .

waspc/data/Cli/templates/basic/package.json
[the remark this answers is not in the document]
Martinsos: Ok, I added the comment like this:
```
  "//": [
    "COMMENTS:",
    {
      "devDependencies.prisma": [
        "We on purpose specify exact version for prisma.",
        "Check this GH comment for the reasoning behind it:",
        "https://github.com/wasp-lang/wasp/pull/634#issuecomment-1158802302 ."
      ]
    }
  ],
```

Terrible, but works, and is a recommended way to do it online.

waspc/src/Wasp/Generator/SdkGenerator.hs
Martinsos: Ok, leaving it open. I linked to this comment inside of the code.

waspc/src/Wasp/Project/Analyze.hs
Martinsos: I changed it to `parsing`.

Martinsos:
Done!

Martinsos:
TODO: Update the docs.

Martinsos:
I fixed all the comments except for that one you said @sodic not to touch.
I also ran the new code, all seems fine.

Martinsos:
Ok I actually revived you change to how output in terminal is prefixed, I saw after running it more that it just doesn't look good.

Martinsos:
I also did the docs.

Martinsos:
I tried running `npm install -S <some_dep>` and Wasp picked it up and reacted to it, so it seems to me it works fine!

sodic:
> Or should this be working?

Everything you mentioned here should be working.

waspc/src/Wasp/Generator/NpmDependencies.hs
```diff
@@ -141,9 +140,11 @@ combineNpmDepsForPackage npmDepsForWasp npmDepsForUser =
   if null conflictErrors && null devConflictErrors
     then
       Right $
+        -- todo(filip): check whether dependency updates and npm install work properly
+        -- todo(filip): reconsider whether we want to change the {sever,web-app}/package.json dynamically
```
Martinsos, line 144:
```suggestion
```

waspc/src/Wasp/Generator/SdkGenerator.hs
sodic: Covered in https://github.com/wasp-lang/wasp/issues/1846

waspc/src/Wasp/Generator/NpmDependencies.hs
sodic: Covered in https://github.com/wasp-lang/wasp/issues/1846

PR #1604, opened 2023-12-19
Adapting Prisma for the new structure

### LINKED ISSUE #1596: Move all Prisma-related operations and packages to the project's top level
In Wasp's new structure (prototyped in #1584), `@prisma/client` is either an SDK dependency or a project dependency, possibly both (see #1595 for details).
It therefore makes most sense to make the `prisma` package a project dependency.

Anticipated tasks:
- Modify the generator to list `prisma` as a dependency in the project's `package.json` file. 
- Decide where to list the `@prisma/client` dependency.
- Verify all Prisma operations successfully execute at the project's top level.
- Verify everything's OK with the types and the IDE support.

sodic:
Fixes #1596.

Tasks:
- [x] Modify the generator to list `prisma` as a dependency in the project's `package.json` file. 
- [x] Decide where to list the `@prisma/client` dependency. - I put it in SDK and everything seems to work OK.
- [x] Verify all Prisma operations successfully execute at the project's top level.
- [x] Verify everything's OK with the types and the IDE support. - Check this again when generating the SDK code using the generator.
- [ ] Verify everything builds properly

How to run and test stuff:
- Go to `waspc/examples/todo-typescript` and work inside that project.
- Running `./cleanstart` will clean everything, fix known issues (e.g., duplicate react) and start the app with all the necessary migrations applied.
- The rest of the commands (e.g,. `cabal run wasp-cli start`, `cabal run wasp-cli clean`) will do what you expect.
- After each change, either let the application pick it up, rerun `cabal run wasp-cli start`, or run `./cleanstart` (depending on what you're changing).
- You can run `./fix` to fix the double react installation issue (but often won't need to, as `./cleanstart` takes care of it). 
- Please don't commit the migrations or `package-lock.json`.

> [!IMPORTANT]  
> **PR 1603** (`filip-wasp-reset`) branches out of **PR 1604** (`filip-restructuring-prisma`), which branches out of **PR 1602** (`filip-project-install-step`).
> The PR numbers are unfortunately out of order.
> Make sure you keep the branches up to date with their bases as you change stuff.


waspc/data/Generator/templates/Dockerfile
```diff
@@ -26,7 +26,7 @@ COPY server/ ./server/
 RUN cd server && npm install
 {=# usingPrisma =}
 COPY db/schema.prisma ./db/
-RUN cd server && {= serverPrismaClientOutputDirEnv =} npx prisma generate --schema='{= dbSchemaFileFromServerDir =}'
+RUN cd server && npx prisma generate --schema='{= dbSchemaFileFromServerDir =}'
```
sodic, line 29: Since we again rely on the implicit/default client generation location, the environment variable is unnecessary. I've removed all mentions of it.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
```diff
@@ -130,56 +141,48 @@ seed projectDir seedName =
...
+  [String] ->
   J.Job
-runPrismaCommandAsJobWithExtraEnv jobType envVars projectDir makeCmdArgs =
-  runNodeCommandAsJobWithExtraEnv envVars serverDir (absPrismaExecutableFp projectDir) (makeCmdArgs schemaFile) jobType
+runPrismaCommandAsJobWithExtraEnv envVars projectRootDir args =
+  runNodeCommandAsJobWithExtraEnv envVars waspProjectDir (absPrismaExecutableFp waspProjectDir) args J.Db
   where
-    serverDir = projectDir </> serverRootDirInProjectRootDir
-    schemaFile = projectDir </> dbSchemaFileInProjectRootDir
+    waspProjectDir = projectRootDir </> waspProjectDirFromProjectRootDir
```
sodic, line 179: This doesn't seem to be the cleanest way of specifying the Wasp project root as a working directory. Still, I choose to go with it because:
- It keeps the directory resolution in a single location.
- It ensures the simplest possible type signatures (i.e., no need of dragging two path arguments all the way down to the command)
- It requires the fewest changes.
- All functions above `runPrismaCommandAsJob` and `runPrismaCommandAsJobWithExtraEnv` really do need the path `ProjectRootDir`, so the "dirty part" is very localized.
- The signatures of `runPrismaCommandAsJobWithExtraEnv` and `runPrismaCommandAsJob` should specify the same `StrongPath` directory in their arguments, so it's either `WaspProjectDir` or `ProjectRootDir`. Going with the first one would require many changes and a lot of duplication.

waspc/src/Wasp/Generator/DbGenerator/Operations.hs
```diff
@@ -184,31 +179,21 @@ isDbConnectionPossible DbConnectionSuccess = True
...
-      writeDbSchemaChecksumToFile projectRootDir dbSchemaChecksumOnLastGenerateFileProjectRootDir
-
-generatePrismaClient ::
-  (String, String) ->
-  J.JobType ->
-  Path' Abs (Dir ProjectRootDir) ->
-  IO (Either String ())
-generatePrismaClient prismaClientOutputDirEnv jobType projectRootDir = do
+generatePrismaClient :: Path' Abs (Dir ProjectRootDir) -> IO (Either String ())
+generatePrismaClient projectRootDir = do
```
sodic, line 183: Since we're only generating a single client, this function is now much shorter.

waspc/src/Wasp/Generator/NpmDependencies.hs
```diff
@@ -142,9 +141,11 @@ combineNpmDepsForPackage npmDepsForWasp npmDepsForUser =
   if null conflictErrors && null devConflictErrors
     then
       Right $
+        -- todo(filip): check whether dependency updates and npm install work properly
+        -- todo(filip): reconsider whether we want to change the {sever,web-app}/package.json dynamically
         NpmDepsForPackage
-          { dependencies = waspDependencies npmDepsForWasp ++ remainingUserDeps,
-            devDependencies = waspDevDependencies npmDepsForWasp ++ remainingUserDevDeps
+          { dependencies = Map.elems remainingWapsDeps,
+            devDependencies = Map.elems remainingWaspDevDeps
```
sodic, line 148: Let's discuss what we should do here. As a first step, of course. Option 1 is the better long-term solution.  
1. Always specify the same dependencies in `{sever,web-app}/package.json` and alias them. Wasp code will use our version of the dependencies, while the user's code will use their versions.
2. For each dependency, decide which version we allow.
   - If the user doesn't specify that dependency in their `package.json`, Wasp lists it in `{sever,web-app}/package.json`.
   - If the user specifies that dependency with the correct version in their `package.json`, Wasp installs it in the top-level `node_modules` and doesn't list it in `{sever,web-app}/package.json` (this is what this PR implements).
   - If the user specifies that dependency with the wrong version in their `package.json`, Wasp reports a conflict and refuses to build (this is what we do in the old structure).

waspc/src/Wasp/Generator/DbGenerator/Common.hs
```diff
@@ -93,21 +90,8 @@ dbSchemaChecksumOnLastGenerateFileInDbRootDir = [relfile|schema.prisma.wasp-gene
...
-serverPrismaClientOutputDirEnv = appComponentPrismaClientOutputDirEnv serverRootDirFromDbRootDir
-
-webAppPrismaClientOutputDirEnv :: (String, String)
-webAppPrismaClientOutputDirEnv = appComponentPrismaClientOutputDirEnv webAppRootDirFromDbRootDir
-
-appComponentPrismaClientOutputDirEnv :: AppComponentRootDir d => Path' (Rel DbRootDir) (Dir d) -> (String, String)
-appComponentPrismaClientOutputDirEnv appComponentDirFromDbRootDir =
-  (prismaClientOutputDirEnvVar, SP.fromRelDir $ appComponentDirFromDbRootDir </> prismaClientOutputDirInAppComponentDir)
+waspProjectDirFromProjectRootDir :: Path' (Rel ProjectRootDir) (Dir WaspProjectDir)
+waspProjectDirFromProjectRootDir = [reldir|../../|]
```
Martinsos, line 94: Ok this is maybe a bit weird hm. You are going in the opposite direction, trying to get to user's code from the generated code dir, right? (terrible names I know: waspProjectDir and projectRootDir).

Since we certainly have information about path from waspProjectDir to projectRootDir, can't you use that info to construct this path? You should be able to "reverse" that path. And then this one gets defined via that one and won't become incorrect.

We have function `reversePosixPath`, look it up -> we could use something like that? I guess the problem is it works only for Posix.

Poor man's solution is always to put  a comment here that this depends on this and that variable, and then also go to those variables and put a comment those variables that they are connected with this one, so when somebody changes on of these, they also have to change the others.

waspc/src/Wasp/Generator/DbGenerator/Common.hs
```diff
@@ -93,21 +90,8 @@ dbSchemaChecksumOnLastGenerateFileInDbRootDir = [relfile|schema.prisma.wasp-gene
...
-serverPrismaClientOutputDirEnv :: (String, String)
-serverPrismaClientOutputDirEnv = appComponentPrismaClientOutputDirEnv serverRootDirFromDbRootDir
-
-webAppPrismaClientOutputDirEnv :: (String, String)
-webAppPrismaClientOutputDirEnv = appComponentPrismaClientOutputDirEnv webAppRootDirFromDbRootDir
-
-appComponentPrismaClientOutputDirEnv :: AppComponentRootDir d => Path' (Rel DbRootDir) (Dir d) -> (String, String)
-appComponentPrismaClientOutputDirEnv appComponentDirFromDbRootDir =
-  (prismaClientOutputDirEnvVar, SP.fromRelDir $ appComponentDirFromDbRootDir </> prismaClientOutputDirInAppComponentDir)
+waspProjectDirFromProjectRootDir :: Path' (Rel ProjectRootDir) (Dir WaspProjectDir)
```
Martinsos, line 93: Oh and as you said this should indeed go somewhere else, something like Generator/Common.hs , or even higher maybe.

waspc/src/Wasp/Generator/DbGenerator/Common.hs
Martinsos: I think I now remembered -> thi information is not hardcoded in a variable but is actually being passed as an argument? In that case, we either need to hardcode it as top level variable, or we will want to actually pass it around properly.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
```diff
@@ -85,43 +85,54 @@ asPrismaCliArgs migrateArgs = do
...
 -- Therefore, this should be checked **after** a command that ensures connectivity.
 migrateStatus :: Path' Abs (Dir ProjectRootDir) -> J.Job
-migrateStatus projectDir = runPrismaCommandAsDbJob projectDir $ \schema ->
-  ["migrate", "status", "--schema", SP.fromAbsFile schema]
+migrateStatus projectRootDir =
+  runPrismaCommandAsJob
+    projectRootDir
+    ["migrate", "status", "--schema", SP.fromAbsFile schema]
+  where
+    schema = projectRootDir </> dbSchemaFileInProjectRootDir
```
Martinsos, line 113: This is releating itself quite a bit, what about having a function `getSchemaPath :: Path' Abs (Dir ProjectRootDir) -> ...`, would that be a bit nicer? Or not much of an improvement?

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
```diff
@@ -130,56 +141,48 @@ seed projectDir seedName =
...
+runPrismaCommandAsJob :: Path' Abs (Dir ProjectRootDir) -> [String] -> J.Job
+runPrismaCommandAsJob projectRootDir args =
+  runPrismaCommandAsJobWithExtraEnv [] projectRootDir args
 
 runPrismaCommandAsJobWithExtraEnv ::
-  JobType ->
   [(String, String)] ->
   Path' Abs (Dir ProjectRootDir) ->
-  (Path' Abs (File PrismaDbSchema) -> [String]) ->
+  [String] ->
```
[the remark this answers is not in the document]
Martinsos: What was lost with not doing the callback? I believe injection of Schema, right? Meaning that each of those functions now has to provide schema on its own, while before schema was injected?

Purely from the standpoint of having less chance of introducing a mistake due to centralized logic (of schema path), to me what we had before looks superior. It is not a terrible difference, I am sure we can survive with redefining schema again and again, but sounds better not to have to do it. I don't understand what you gained by this change where you removed this callback, you said code is less "jumpy"? What does that mean? Why was code confusing?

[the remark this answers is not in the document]
Martinsos: Why wouldn't such function be concerned with locating schema file? Sounds to me better to have that function figure it out, then every single Prisma Command having to redefine it? But this is very hard thing to discuss, isn't it -> it is up to us to decide how "managed" experience this function provides.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
Martinsos: Looking at it, doesn't look like there is something very wrong with it. If you have these two pieces of information, `projectRootDir` and `waspProjectDirFromProjectRootDir`, why wouldn't you use them here?

I guess the real problem lies in the "dirtiness" of `waspProjectDirFromProjectRootDir`?

Ok, just to make sure I got this right: you were considering passing `ProjectRootDir` (what we have now) vs passing `WaspProjectDir` (what we could have in theory) instead?
And it feels better to pass the path that is higher above? Why would one be dirtier than the another?

To me, dirty thing seems to be `waspProjectDirFromProjectRootDir` at the moment, if that one is made clean, then this is also clean. The question is, can it be made clean in the form it is in now.

Btw I am trying to guess a bit here, because you seem to have an alternative approach in mind but don't describe it explicitly. If I got it wrong, please correct me.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
```diff
@@ -130,56 +141,48 @@ seed projectDir seedName =
...
-    serverDir = projectDir </> serverRootDirInProjectRootDir
-    schemaFile = projectDir </> dbSchemaFileInProjectRootDir
+    waspProjectDir = projectRootDir </> waspProjectDirFromProjectRootDir
 
 -- | NOTE: The expectation is that `npm install` was already executed
 -- such that we can use the locally installed package.
 -- This assumption is ok since it happens during compilation now.
-absPrismaExecutableFp :: Path' Abs (Dir ProjectRootDir) -> FilePath
-absPrismaExecutableFp projectDir = SP.fromAbsFile prismaExecutableAbs
+absPrismaExecutableFp :: Path' Abs (Dir WaspProjectDir) -> FilePath
```
Martinsos, line 184: Aha, so this basically says that Prisma is now living in user node_modules, right?

waspc/src/Wasp/Generator/DbGenerator/Operations.hs
Martinsos: Sweet

Martinsos (commented):
@sodic did the first round -> it is great to see things simplifying!

waspc/src/Wasp/Generator/DbGenerator.hs
```diff
@@ -173,23 +171,23 @@ warnProjectDiffersFromDb projectRootDir = do
         "Wasp was unable to verify your database is up to date."
           <> " Running `wasp db migrate-dev` may fix this and will provide more info."
 
-genPrismaClients :: AppSpec -> Path' Abs (Dir ProjectRootDir) -> IO (Maybe GeneratorError)
-genPrismaClients spec projectRootDir =
+generatePrismaClient :: AppSpec -> Path' Abs (Dir ProjectRootDir) -> IO (Maybe GeneratorError)
+generatePrismaClient spec projectRootDir =
```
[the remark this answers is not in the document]
sodic: Because it used to be "clients" (plural) and would generate two of them, the `gen` -> `generate` part overshadowed the more important change :D

> So why did you comment this here? I also wonder, should we have this comment somewhere in the actual code in the relevant place?

I don't think so. Once we merge this, the fact that we once had two clients as a workaround becomes unimportant.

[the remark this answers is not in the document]
Martinsos: Aha I didn't notice the change in plurality!

I was referring to the comment that it generates a single Prisma client in `node_modules/.prisma`, I thought that was worth documenting somewhere, not the fact that there were 2 clients before.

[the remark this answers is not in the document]
sodic: Hm, I'd skip documenting that one as well.
Generating the client in `node_modules` is the default and expected behavior from Prisma. Our previous approach was also a very custom workaround.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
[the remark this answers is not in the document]
sodic: By the "jumpy" code, I meant:
I was confused by the callback. Sending a callback as an argument separates the logic into two different places (the callback body and its call site), making the reader have to "jump" between two code locations to realize what's going on.

Sending a value, on the other hand, is simple - the code has a value and uses it.

In short, a higher-order function is a an abstraction that adds complexity and reduces duplication (as most abstractions do). In this case, I don't think the duplication justifies it.

> Purely from the standpoint of having less chance of introducing a mistake due to centralized logic (of schema path), to me what we had before looks superior.

This is true. I agree. But I'd argue we should handle it elsewhere (see below).

> Why wouldn't such function be concerned with locating schema file?

Such a function should, in my opinion, be as dumb as possible. It should take a command, env variables, arguments, and execute the command.

It should'b be constructing command line arguments through a provided callback.

I'd expect to use `runPrismaCommandAsJobWithExtraEnv` for running a command that doesn't require a `schema` argument without having to go through the ceremony of sending a schema to a callback and then ignoring it.

Also, what happens when different commands need different files for arguments (e.g., one needs a schema, the other needs a target directory, etc.). We'd have to update the HOF in the signature to expect all of those. In an extreme case, each passed callback would have to take multiple arguments and ignore most of them.

Furthermore, `runPrismaCommandAsJobWithExtraEnv` might not know how to get all the arguments, which means we'd be constructing one part of the argument array in the caller, and the other part by evaluating the callback (making the code even more "jumpy").

It seems cleaner to simply construct the arguments on the outside and pass them into `runPrismaCommandAsJobWithExtraEnv`. Yes, there's duplication, but only because our functions are simple and all of them only require a schema.

I don't think the callback abstraction occurs naturally. Still, if you have a strong preference for it, I'll change it back.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
sodic: > Ok, just to make sure I got this right: you were considering passing ProjectRootDir (what we have now) vs passing WaspProjectDir (what we could have in theory) instead?

If we wanted to avoid getting one from the other using relative paths, we'd need to pass both `ProjectRootDir` and `WaspProjectDir` almost all the way down. If we don't care about using relative paths for getting one from the other, then I think this is the cleanest solution.

The part I found dirty was defining the `WaspProjectDir` in terms of `ProjectRootDir` (as opposed to sending it down). So, yes, defining `waspProjectDirFromProjectRootDir` felt like a hack.

I think the textbook approach is sending both directories all the way down, but it just seems unnecessarily cumbersome.

Why did you find `waspProjectDirFromProjectRootDir` dirty, if not for this reason?

waspc/data/Generator/templates/Dockerfile
sodic: As mentioned in the other comment, I wouldn't document this. It's Prisma's default and expected behavior.

I does feel like we should write it somewhere, but I think we're just biased due to having a custom setup for all this time. If we had it working like this from the beginning, I don't think we would have thought twice about it.

waspc/examples/todo-typescript/package.json
```diff
@@ -2,7 +2,6 @@
   "name": "prototype",
   "dependencies": {
     "wasp": "file:.wasp/out/sdk/wasp",
-    "@prisma/client": "4.16.2",
```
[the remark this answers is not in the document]
sodic: > So it is not in their package.json because they are not using it directly?

Yes, exactly.

> What if they want to run npx prisma -> they can't right?

I think they can. They do have the Prisma CLI, they just don't have the client. Why do you think it wouldn't work?

> I mean even if they can it won't use our version of Prisma. Is that bad or good?

I'm not following this one. I think there's only a single version of Prisma. Which versions are you considering?

waspc/src/Wasp/Generator/DbGenerator/Common.hs
sodic: This is the dirty part I was referring to earlier. Let's discuss in in the other thread.

Just to repeat some of the points explained in the other thread:

> or we will want to actually pass it around properly.

This is what I've tried to avoid. Relevant comments: https://github.com/wasp-lang/wasp/pull/1604#discussion_r1431304174 and https://github.com/wasp-lang/wasp/pull/1604#discussion_r1437745002.

> we either need to hardcode it as top level variable

Unfortunately, this isn't possible because `ProjectRootDir` can be either `out` or `build`.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
sodic: Sure, I'll add it in.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
sodic: Exactly!

waspc/src/Wasp/Generator/DbGenerator.hs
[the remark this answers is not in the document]
Martinsos: It is, but still it doesn't mean it is very intuitive. I would mention it shortly in the place that actually performs that install.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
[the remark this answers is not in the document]
Martinsos: The idea was just that that command manages prisma schema argument, because that was a bit PITA and is almost always needed. Maybe I forgot how this part of code works, but I think you can still pass other arguments, even though the function was taking care of schema? So there is no need to beef the callback up with other stuff if needed, as you described?
In any case, not a biggie, I am ok with this solution also, I don't think it matters much, so all good from my side.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
Martinsos: I don't think both pieces of information have to be passed down, I think it is ok to pass one piece and have another hardcoded if that makes sense for us (and it could/does). I don't see why that would be dirty.

What I didn't like is how `waspProjectDirFromProjectRootDir` is defined, and I commented on that in another comment. So we can close this one and continue there.

waspc/examples/todo-typescript/package.json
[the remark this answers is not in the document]
Martinsos: Sorry, I mixed prisma cli and prisma client, all good.

waspc/src/Wasp/Generator/DbGenerator/Common.hs
Martinsos: And I just suggested in that comment thread that we continue the discussion here :D.

Even though it can be either `out` or `build`, we can still hardcode it -> it just means that the relative path is the same regardless of which one of these is it.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
Martinsos: Should we mention here why we expect that prisma is installed or user's node_modules, or that we expect it? I am maybe a bit out of context here so disregard this if it doesn't make sense.

Martinsos (commented):
Reviewed a bit more.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
sodic: We're specifying it as a dependency, and the comment above says it relies on "npm install" being done, so I think it's fine. I can remove the "This assumption is ok..." part.

waspc/src/Wasp/Generator/DbGenerator.hs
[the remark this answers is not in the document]
Martinsos: Just to check, since there is no :rocket: -> is this done?

waspc/src/Wasp/Generator/DbGenerator/Common.hs
```diff
@@ -19,15 +16,15 @@ module Wasp.Generator.DbGenerator.Common
...
+    waspProjectDirFromProjectRootDir,
     DbSchemaChecksumFile,
   )
 where
 
 import StrongPath (Dir, File, File', Path', Rel, reldir, relfile, (</>))
-import qualified StrongPath as SP
 import Wasp.Generator.Common (AppComponentRootDir, DbRootDir, ProjectRootDir, ServerRootDir)
 import Wasp.Generator.Templates (TemplatesDir)
+import Wasp.Project.Common (WaspProjectDir)
```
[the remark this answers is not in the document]
Martinsos: Feel free to resolve this one if you are all good with it.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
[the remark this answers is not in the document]
Martinsos: Feel free to resolve this comment if you are all good with it.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
Martinsos: We discussed live how to solve this, and agreed to move the `../../` next to definition of `out` and `build` dirs, as the fastest okish solution.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
Martinsos: I am out of context here, so whatever you think is best. For stuff like this, extra comments don't hurt.

sodic: We can comment something like "This command assumes the project's top level `npm install` was already executed and that `prisma` exists inside the top-level `node_modules` .

Martinsos:
Fixed the comments, didn't run anything yet.

PR #1607, opened 2023-12-19
Fix import/export JS errors in Mage

jankovidakovic:
### Fix import/export JS errors in Mage

> This PR fixes the common mistakes regarding the importing and exporting of JS files. 

Prompts are updated to increase consistency in instructions. Model is now instructed to always generate:
 - default exports in React components (when generating pages)
 - default imports in Wasp `page` declarations
 
 Model is also instructed to edit the named imports/exports in pages (during the fixing stage).
 
 Some prompt elements are extracted away to new `Text` variables for easier readability. 
 
A future option would be to further work on refactoring the prompts, but I don't think that this is now a priority.

### DIFF
--- a/waspc/src/Wasp/AI/GenerateNewProject/Common/Prompts.hs
+++ b/waspc/src/Wasp/AI/GenerateNewProject/Common/Prompts.hs
@@ -4,6 +4,8 @@ module Wasp.AI.GenerateNewProject.Common.Prompts
     systemPrompt,
     appDescriptionStartMarkerLine,
     appDescriptionBlock,
+    waspPlanExample,
+    waspPlanSchema,
   )
 where
 
@@ -135,3 +137,46 @@ waspFileExample =
         }
         ```
       |]
+
+waspPlanExample :: Text
+waspPlanExample =
+  [trimming|
+      Here is an example of a plan (a bit simplified, as we didn't list all of the entities/actions/queries/pages):
+
+        {
+          "entities": [{
+            "entityName": "User",
+            "entityBodyPsl": "  id Int @id @default(autoincrement())\n  username String @unique\n  password String\n  tasks Task[]"
+          }],
+          "actions": [{
+            "opName": "createTask",
+            "opFnPath": "@server/actions.js",
+            "opDesc": "Checks that user is authenticated and if so, creates new Task belonging to them. Takes description as an argument and by default sets isDone to false. Returns created Task."
+          }],
+          "queries": [{
+            "opName": "getTask",
+            "opFnPath": "@server/queries.js",
+            "opDesc": "Takes task id as an argument. Checks that user is authenticated, and if so, fetches and returns their task that has specified task id. Throws HttpError(400) if tasks exists but does not belong to them."
+          }],
+          "pages": [{
+            "pageName": "TaskPage",
+            "componentPath": "@client/pages/Task.jsx",
+            "routeName: "TaskRoute",
+            "routePath": "/task/:taskId",
+            "pageDesc": "Diplays a Task with the specified taskId. Allows editing of the Task. Uses getTask query and createTask action.",
+          }]
+        }
+|]
+
+waspPlanSchema :: Text
+waspPlanSchema =
+  [trimming|
+        Plan is represented as JSON with the following schema:
+
+        {
+          "entities": [{ "entityName": string, "entityBodyPsl": string }],
+          "actions": [{ "opName": string, "opFnPath": string, "opDesc": string }],
+          "queries": [{ "opName": string, "opFnPath": string, "opDesc": string }],
+          "pages": [{ "pageName": string, "componentPath": string, "routeName": string, "routePath": string, "pageDesc": string }]
+        }
+  |]
--- a/waspc/src/Wasp/AI/GenerateNewProject/Common/Prompts.hs
+++ b/waspc/src/Wasp/AI/GenerateNewProject/Common/Prompts.hs
@@ -93,7 +93,7 @@ waspFileExample =
           component: import Login from "@client/pages/auth/Login.jsx"
         }
 
-        route DashboardRoute { path: "/", to: Dashboard }
+        route DashboardRoute { path: "/", to: DashboardPage }
         page DashboardPage {
           authRequired: true,
           component: import Dashboard from "@client/pages/Dashboard.jsx"
--- a/waspc/src/Wasp/AI/GenerateNewProject/Common/Prompts.hs
+++ b/waspc/src/Wasp/AI/GenerateNewProject/Common/Prompts.hs
@@ -4,8 +4,6 @@ module Wasp.AI.GenerateNewProject.Common.Prompts
     systemPrompt,
     appDescriptionStartMarkerLine,
     appDescriptionBlock,
-    waspPlanExample,
-    waspPlanSchema,
   )
 where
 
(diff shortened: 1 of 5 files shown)


waspc/src/Wasp/AI/GenerateNewProject/Common/Prompts.hs
```diff
@@ -93,7 +95,7 @@ waspFileExample =
           component: import Login from "@client/pages/auth/Login.jsx"
         }
 
-        route DashboardRoute { path: "/", to: Dashboard }
+        route DashboardRoute { path: "/", to: DashboardPage }
```
Martinsos, line 98: Ouf, we had a mistake here, did we? Nice catch! This must have confused it to some degree.

waspc/src/Wasp/AI/GenerateNewProject/Page.hs
```diff
@@ -107,11 +107,18 @@ generatePage newProjectDetails entityPlans queries actions pPlan = do
...
+          "pageWaspDecl": "route ExampleRoute { path: \"/\", to: ExamplePage }\npage ExamplePage {\n  component: import ExamplePage from \"@client/ExamplePage.jsx\",\n  authRequired: true\n}",
           "pageJsImpl": "JS imports + React component implementing the page.",
         }
         There should be no other text in the response.
 
+        When writing the javascript implementation, make sure to always use the default export. 
+        Concretely, define the main component with `const ${pageName} = () => {...}`,
+        and then at the end export it with `export default ${pageName}`.
+        It is also really important that the ${pageName} is then imported as a default import. 
+        Concretely, in "pageWaspDecl", use `import ${pageName}` instead of `import { ${pageName} }`.
```
Martinsos, line 119:
```suggestion
        It is also really important that the ${pageName} is then imported as a default import in "pageWaspDecl": use `import ${pageName}` instead of `import { ${pageName} }`.
```

waspc/src/Wasp/AI/GenerateNewProject/WaspFile.hs
```diff
@@ -89,12 +89,20 @@ fixWaspFile newProjectDetails waspFilePath plan = do
             ${compileErrorsText}
 
             Some common mistakes to look for:
+              - using non-default imports in page components
```
Martinsos, line 92:
```suggestion
              - Using non-default imports in page components. 
```

waspc/src/Wasp/AI/GenerateNewProject/WaspFile.hs
```diff
@@ -89,12 +89,20 @@ fixWaspFile newProjectDetails waspFilePath plan = do
             ${compileErrorsText}
 
             Some common mistakes to look for:
+              - using non-default imports in page components 
+                  In a Wasp page, the component should always use the "default import" JS syntax.
+                  Instead of `component: import { PageName } from ...`, it should 
+                  always be `component: import PageName from ...`.
+                  Fix these by identifying named imports in wasp pages, and replacing 
+                  them with default imports. This only relates to wasp pages, other parts 
+                  of a Wasp file do not have to use the default imports.
```
Martinsos, line 98:
```suggestion
                  In a Wasp `page` declaration, the `component` should always use the "default import" JS syntax.
                  Instead of `component: import { PageName } from ...`, it should 
                  always be `component: import PageName from ...`.
                  Fix these by identifying named imports in wasp pages, and replacing 
                  them with default imports. This only relates to wasp pages, other parts 
                  of a Wasp file do not have to use the default imports.
```

waspc/src/Wasp/AI/GenerateNewProject/Plan.hs
```diff
@@ -56,6 +56,8 @@ generatePlan newProjectDetails planRules = do
     appDescriptionBlockText = appDescriptionBlock newProjectDetails
     basicWaspLangInfoPrompt = Prompts.basicWaspLangInfo
     waspFileExamplePrompt = Prompts.waspFileExample
+    waspPlanExample = Prompts.waspPlanExample
+    waspPlanSchema = Prompts.waspPlanSchema
```
Martinsos, line 60: While I don't terribly mind you extracting these, I am not sure what is the purpose of it? In Common/Prompts.hs we keep common prompts, that are used in multiple places -> these prompts here are not used in multiple places. I get that it might feel attractive to have all prompts in one place, but then we lose the structure we had with all the Haskell files, and all of them are in one file and are harder in my opinion to work with. We could have alternative hierary inside Prompts/, but I wouldn't do that.

Martinsos (commented):
@jankovidakovic Looking quite good, and I think this will work nicely! Left some comments.

waspc/src/Wasp/AI/GenerateNewProject/Page.hs
jankovidakovic: My understanding is that LLMs are extremely insensitive when it comes to small syntactic things (such as here -- using `:` instead of `. Concretely, `. Why "extremely"? See for example the following paper: https://arxiv.org/abs/2311.18805

Having said that, there's no good reason to keep it confusing for humans. I'll implement the suggested change.

waspc/src/Wasp/AI/GenerateNewProject/WaspFile.hs
jankovidakovic: This makes sense to keep it consistent with the rest of the bullet points.

waspc/src/Wasp/AI/GenerateNewProject/WaspFile.hs
jankovidakovic: Again, this change probably does not matter from the perspective of the LLMs, but it makes sense from a human perspective.

waspc/src/Wasp/AI/GenerateNewProject/Plan.hs
jankovidakovic: Extracting these seemed convenient because then the structure of `planPrompt` was easier for me to understand. However, I missed the part where fields in `Prompts` are used in multiple places. 

As you suggested, we could probably talk about extracting `waspPlanExample` and `waspPlanSchema` into other files, but I think we agree that that's not a priority right now.

I'll go ahead and un-extract these.

waspc/src/Wasp/AI/GenerateNewProject/WaspFile.hs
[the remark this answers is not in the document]
jankovidakovic: This is explicitly instructed when the React pages are first generated: https://github.com/jankovidakovic/wasp/blob/36f43fbc55070b8ce88935bffe1200c6fa88f3d0/waspc/src/Wasp/AI/GenerateNewProject/Page.hs#L115-L117

I'll also add it to the list of guidelines for fixing React pages: 
https://github.com/jankovidakovic/wasp/blob/36f43fbc55070b8ce88935bffe1200c6fa88f3d0/waspc/src/Wasp/AI/GenerateNewProject/PageComponentFile.hs#L160

waspc/src/Wasp/AI/GenerateNewProject/Page.hs
Martinsos: Makes sense, I would also expect GPT to not care much, but while at it why not make it as clear as possible :), at least for humans as you said!

waspc/src/Wasp/AI/GenerateNewProject/Plan.hs
Martinsos: Makes sense!

Tip again regarding commenting in PRs: what would make this the easiest for me here is if you put the :rocket: on intiail comment and didn't resolve this.

Then I see the :rocket: and I know change was done (right now I am not sure, I will have to go check it in the code, because you said you will do it but I am not sure if that happened yet).
Also, I read you comment and then I resolve, while if you resolved it, I would normally miss it completely.

Martinsos (approved):
Great @jankovidakovic , I think this is it, I approved!

Issue #1406, opened 2023-08-23
Pin down the version of starter templates?


Martinsos:
Right now, Wasp CLI inserts its version number in the starter templates when it creates them on the disks -> but that makes no sense, they were made for a specific Wasp version! Worst thing, you won't get a nice error message that you have wrong version, just some errors possibly.

Solution is to pin down the Wasp version in them, so we specify which version of CLI they work with. True, if you have older version of Wasp CLI and try to run it with templates, that will mean you will install template that needs newer version, but you will get nice error message which is good.
We could also have it so that we have github tags for starter templates, that say for which wasp version they are made, and then each version of CLI could specify, using tag, which version of starter templates is it using.

There are some small decisions to be made here, but I am confident this direction makes sense.

PR #1619, opened 2024-01-05
Added open-saas starter template to CLI & fixed all starter templates (instructions, versioning)

Martinsos:
I replaced the old "saas" starter template (from options shown on `wasp new`) with Open Saas.

While doing this, I removed the mechanism of looking up templates.json file in "starters" gh repo, and instead each template is now defined individually in StarterTemplates.hs, by specifying its gh repo and some more details. That means that our code doesn't give special treatment to "starters" gh repo any more, and that templates can be in any repo. The fact that some of them are in the same repo is not an issue though, but they are now treated separately.
This enabled me to use same mechanism for open-saas and for templates in "starters" repo.
This also brings the benefit of `wasp new` not crashing on missing internet connection until you actually try to pull a specific remote template.

While doing all this, I realized that us always having the same instructions after generation of every template (which are `cd <projectDir` and then `wasp start`) is not enough, and that we need specific instructions per starter template, so I also added that and made them correct for each template.

Finally, I added the versioning of starter templates that are in gh repos via git tags, as per reasoning here https://github.com/wasp-lang/wasp/issues/1406 . This ensures CLI will always download templates that work for it, which was not the case so far. This resolves https://github.com/wasp-lang/wasp/issues/1406 . I created tags for v0.12 of Wasp on needed gh repos of starter templates.

image


waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
```diff
@@ -106,55 +108,62 @@ openSaasStarterTemplate =
...
+          [ styleText $ "To run your new app, do:",
+            styleCode $ "    cd " ++ projectDirName,
+            styleCode $ "    wasp db migrate-dev",
+            styleCode $ "    wasp start",
+            styleText $ "",
+            styleText $ "Check the README for additional guidance!"
+          ]
     )
 
+{- Functions for styling instructions. They are on purpose of same length, for nicer formatting. -}
```
infomiho, line 145: What is the same length, the function name or the function output?

waspc/README.md
```diff
@@ -362,7 +362,12 @@ NOTE: If building of your commit is suddenly taking much longer time, it might b
 If it happens just once every so it is probably nothing to worry about. If it happens consistently, we should look into it.
 
 ### Typical Release Process
-- Ensure that all starter templates in `starter` repo are working with the version of Wasp we are about to release and upgrade their version of Wasp to the new one.
+- Starter templates
+  - Context: they are used by used by `wasp new`, you can find reference to them in `Wasp.Cli. ... .StarterTemplates`.
```
infomiho, line 366: `Wasp.Cli. ... .StarterTemplates` is a placeholder you wanted to replace with a proper module path or are you keeping vague intentionally?

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
```diff
@@ -7,78 +7,190 @@ module Wasp.Cli.Command.CreateNewProject.StarterTemplates
...
+        unlines
+          [ styleText $ "To run your new app, do:",
+            styleCode $ "    cd " <> projectDirName FP.</> "app",
+            styleCode $ "    wasp db start",
+            styleText $ "Then open new terminal window/tab in that same dir and do:",
+            styleCode $ "    wasp db migrate-dev",
+            styleCode $ "    cp .env.server.example .env.server",
+            styleCode $ "    wasp start",
+            styleText $ "",
+            styleText $ "Check the README for additional guidance and link to docs!"
```
infomiho, line 121:
```suggestion
            styleText $ "Check the README for additional guidance and the link to docs!"
```

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/GhRepo.hs
```diff
@@ -0,0 +1,31 @@
...
+createProjectOnDiskFromGhRepoTemplate ::
+  Path' Abs (Dir WaspProjectDir) ->
+  NewProjectName ->
+  NewProjectAppName ->
+  GithubRepoRef ->
+  Path' Rel' Dir' ->
+  Command ()
+createProjectOnDiskFromGhRepoTemplate absWaspProjectDir projectName appName ghRepoRef templatePathInRepo = do
+  fetchTheTemplateFromGhToDisk >>= either throwProjectCreationError pure
+  -- NOTE: Placeholders in main.wasp files of templates were probably not the best idea,
```
infomiho: Placeholders for the project and app name were used even before the starter template, I think it's nice to get your app name in the created project, all other frameworks do that.

As far the version is concerned, it wasn't the best idea to keep that dynamic 👍

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
```diff
@@ -7,78 +7,190 @@ module Wasp.Cli.Command.CreateNewProject.StarterTemplates
...
+          unlines
+            [ styleText $ "To run your new app, do:",
+              styleCode $ "    cd " <> projectDirName,
+              styleCode $ "    wasp db start"
+            ]
       }
 
+{- HLINT ignore openSaasStarterTemplate "Redundant $" -}
+
+openSaasStarterTemplate :: StarterTemplate
```
infomiho, line 103: Hard coding the templates here makes it impossible to add new templates without releasing a new Wasp version. I just wanted to point that out so we are aware.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
```diff
@@ -7,78 +7,190 @@ module Wasp.Cli.Command.CreateNewProject.StarterTemplates
...
+    )
+
+waspGhOrgName :: String
+waspGhOrgName = "wasp-lang"
+
+-- NOTE: As version of Wasp CLI changes, so we should update this tag name here,
+--   and also create it on gh repos of templates.
+--   By tagging templates for each version of Wasp CLI, we ensure that each release of
+--   Wasp CLI uses correct version of templates, that work with it.
+waspVersionTemplateGitTag :: String
```
infomiho, line 191: If the tag is not present on the repo, how would that look like for the user in the CLI?

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/GhRepo.hs
```diff
@@ -0,0 +1,31 @@
...
+  --   and we should likely drop the concept in the future, but we still do this replacement here
+  --   in order to keep it all working for old templates. Once we stop using any templates that
```
infomiho: We just need to replace placeholders with plain text and it will work. If placeholders are not found, the CLI won't complain.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
Martinsos: Function names! I improved the comment.
I played with with telling Ormolu to ignore these parts, but yo can tell it that only for the whole function, and then you loose formatting for the whole function which was annoying. So I went with this at the end.

waspc/README.md
Martinsos: I wanted to keep it vague intentionally, to make it a bit more robust to changes in the future. I could make it more concrete if needed.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
Martinsos: Good one!

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/GhRepo.hs
Martinsos: You are right, I updated the comment:
```
  -- NOTE: Placeholders in main.wasp files of templates were probably not the best idea
  --   because Wasp version should be fixed, not dynamically determined, and while injecting
  --   app/project name is nice, we can't do it for templates that are also to be used standalone,
  --   like open-saas, so that also isn't that useful then (and is not a big help anyway).
  --   We should likely drop the concept in the future, but we still do this replacement here
  --   in order to keep it all working for old templates. Once we stop using any templates that
  --   have placeholders in their main.wasp files, we can remove this line below.

```

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
Martinsos: I know, but concluded that is probably all right and not that important, and it keeps things simpler for us.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
Martinsos: It would return HTTP error where it failed to fetch from github URL that contains that tag. I don't know if it is worth spending time on improving that error message -> it shouldn't happen anyway.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/GhRepo.hs
Martinsos: Hm you are right, I had another error at that time that confused me thinking it will be an issue, but it won't.
Would you suggest we do it then? The thing is, we need to put a specific Wasp version in them then. So for some older Wasp CLI's, it might be too new version. But well, that is the state of things really. Yeah, I think we should do it: in starters/repo on `main` branch I will set Wasp version to ^0.11 and remove placeholders for remove this templating code completely. Just let me know if you agree with this first.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
```diff
@@ -45,9 +48,17 @@ createProjectOnDisk
...
         liftIO $ createProjectOnDiskFromLocalTemplate absWaspProjectDir projectName appName $ _path metadata
       AiGeneratedStarterTemplate ->
         AI.createNewProjectInteractiveOnDisk absWaspProjectDir appName
+
+-- | This function assumes that the project dir was created inside the current working directory.
+printGettingStartedInstructionsForProject :: NewProjectDescription -> IO ()
+printGettingStartedInstructionsForProject projectDescription = do
+  let projectDirName = init . SP.toFilePath . SP.basename $ _absWaspProjectDir projectDescription
+  let instructions = getTemplateStartingInstructions projectDirName $ _template projectDescription
+  putStrLn $ Term.applyStyles [Term.Green] $ "Created new Wasp app in ./" ++ projectDirName ++ " directory!"
```
infomiho, line 63: Screenshot 2024-01-09 at 11 29 57

I'd recommended introducing new lines, so the output looks like this:

```
Created new Wasp app in ./xoxo directory!

To run your new app, do:
    cd xoxo/app
    wasp db start

Then open new terminal window/tab in that same dir and do:
    wasp db migrate-dev
    cp .env.server.example .env.server
    wasp start
```

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
```diff
@@ -7,78 +7,190 @@ module Wasp.Cli.Command.CreateNewProject.StarterTemplates
...
+          [ styleText $ "To run your new app, do:",
+            styleCode $ "    cd " <> projectDirName FP.</> "app",
+            styleCode $ "    wasp db start",
+            styleText $ "Then open new terminal window/tab in that same dir and do:",
+            styleCode $ "    wasp db migrate-dev",
+            styleCode $ "    cp .env.server.example .env.server",
+            styleCode $ "    wasp start",
+            styleText $ "",
+            styleText $ "Check the README for additional guidance and the link to docs!"
+          ]
```
infomiho, line 122:
```suggestion
          [ styleText $ "To run your new app, do:",
            styleCode $ "    cd " <> projectDirName FP.</> "app",
            styleCode $ "    wasp db start",
            styleText $ "",
            styleText $ "Then open new terminal window/tab in that same dir and do:",
            styleCode $ "    wasp db migrate-dev",
            styleCode $ "    cp .env.server.example .env.server",
            styleCode $ "    wasp start",
            styleText $ "",
            styleText $ "Check the README for additional guidance and the link to docs!"
          ]
```

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
```diff
@@ -7,78 +7,190 @@ module Wasp.Cli.Command.CreateNewProject.StarterTemplates
...
+embeddingsStarterTemplate =
+  simpleGhRepoTemplate
+    ("starters", [reldir|embeddings|])
+    ( "embeddings",
+      "Comes with code for generating vector embeddings and performing vector similarity search."
+    )
+    ( \projectDirName ->
+        "To learn how to set up and run your new app, check instructions in "
+          <> projectDirName FP.</> "README.md"
+    )
```
infomiho, line 165: I think we can use the same instructions as for OpenSaas template by looking at the README.md

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/GhRepo.hs
infomiho: I think that's a good plan 👍 the `main` branch templates are locked forever to the `^0.11.0` version

Martinsos: Ok, so what I did is following:
1. On `starters` repo I locked `main` branch (made it read-only), made new `dev` branch from it, made it default branch, and on that `dev` branch I removed templates.json, removed saas template, and pinned versions to 0.12. I also updated README in `main` to explain why it is locked, and I updated README in `dev` to be in sync with the new situation there.
2. Here in this repo I removed this code comment we commented about, and I also removed `version` from the placehodlers -> I left app name and title though, why not, if some template wants to use it it can.

So that is it -> old Wasp versions will work as they worked so far, while new ones (starting with 0.12) will be following the new approach.

At one point we might want to delete the `main` branch from `starters`, but it doesn't really matter even if we don't.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
Martinsos: I turned it into this, for saas, because it is more complex than other examples (due to having to open another terminal window, which makes explaining harder):

image

Also added some new lines for other stuff.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
Martinsos: I did this in a different way, commented above.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
Martinsos: Yup, I enriched them in the same way!

I didn't want to do it initially because I thought `embeddings` are less interesting + thought instructions are too complex to show them here, but actually it works, and I now had a "framework" to follow, so I did it.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/GhRepo.hs
infomiho: Perfect 👍

infomiho:
Great improvement with this PR 👍

PR #1620, opened 2024-01-06
New auth model docs

infomiho:
Docs changes required for the auth model changes done in #1583 

- Removed the old auth model mentions
- Described the new auth model in a separate page
- Updated the texts for auth methods
- Added mentions of the new auth model in relevant places

### Left to do

- [x] Go though the tutorial
- [x] Go through each auth provider one more time
- [x] Add a section on the way to use the `user`
  - [x] Server
  - [x] Client
  - [x] New helpers to get the e.g. `username`


web/docs/auth/email.md
```diff
@@ -595,7 +584,234 @@ Users can enter their new password there.
 
 The content of the e-mail can be customized, read more about it [here](#passwordreset-passwordresetconfig-).
 
-## Using The Auth
+## Creating a Custom Sign-up Action
+
+> TODO: test out the code below and make sure it works
```
infomiho, line 589: I'm not sure if we should encourage custom signup actions for the Email auth since it's really complex and they have to take care of:
- checking for existing users
- the timing attack prevention (faking the work)
- sending the verification email
... 

Maybe the better workflow could be if we provided `preSignup` and `postSignup` hooks. But this is obviously out of scope here. We already talked about such hooks and how they could be useful.

web/docs/auth/overview.md
```diff
@@ -382,45 +397,52 @@ export const updatePassword = async (args, context) => {
...
-#### Username & password
+#### Username & Password
 
 If you use [Username & password](/docs/auth/username-and-pass) authentication, the default validations are:
 
 - The `username` must not be empty
 - The `password` must not be empty, have at least 8 characters, and contain a number
 
-Note that `username`s are stored in a **case-sensitive** manner.
+Note that `username`s are stored in a **case-insensitive** manner.
```
infomiho, line 445: Since we are normalising the provider user ID equally, the behaviour of the user case sensitivity changed (for the better iMHO)

web/docs/auth/username-and-pass.md
```diff
@@ -482,19 +484,29 @@ import {
...
-      data: {
-        username: args.username,
-        password: args.password, // Password is hashed automatically by Wasp
-      },
+    const providerId = createProviderId('username', args.username)
+    const providerData = await sanitizeAndSerializeProviderData({
+      hashedPassword: args.password,
     })
+
+    await createUser(
```
infomiho, line 504: You'll notice I've opted to use our helper functions since it's easier to implement the signup flow vs. using plain Prisma.

web/docs/tutorial/07-auth.md
```diff
@@ -271,33 +269,41 @@ wasp db studio
...
 style={{ border: "1px solid black" }}
 />
 
-We see there is a user and that its password is already hashed 🤯
+
+You'll notice that we now have a `User` entity in the database alongside the `Task` entity.
 
 However, you will notice that if you try logging in as different users and creating some tasks, all users share the same tasks. That's because we haven't yet updated the queries and actions to have per-user tasks. Let's do that next.
 
+<small>
```
infomiho, line 281: The auth entities are not relevant for the tutorial, but they will see them, so I thought it's best to address their existence.

web/docs/auth/overview.md
```diff
@@ -295,7 +296,7 @@ Since the `user` prop is only available in a page's React component: use the `us
 
 #### Using the `context.user` object
 
-When authentication is enabled, all [queries and actions](/docs/data-model/operations/overview) have access to the `user` object through the `context` argument. `context.user` contains all User entity's fields, except for the password.
+When authentication is enabled, all [queries and actions](/docs/data-model/operations/overview) have access to the `user` object through the `context` argument. `context.user` contains all User entity's fields and the auth identities connected to the user. We strip out the `hashedPassword` field from the identities for security reasons.
```
Martinsos, line 299: This idea that we have `user.auth.identities` -> we should make that very clear and obvious in docs. I am not saying we need to say it first, it is advanced stuff, but when the moment for advanced stuff comes (for example here in this overview), then I would take some space and their focus to introduce just this specific concept very clearly.

Maybe a very simple diagram, or maybe `user.auth.identities` code block in the middle of the page, with very clear explanation -> your user object now has auth field, added by Wasp. That field has `identities` field, which is an array contains info about all the auth methods (username / pass, email / pass, github, google, ...) that your user has signed up with so far. Via this array you can get detailed info about each of the identities -> which email they use for email / pass, what is their github id, ... .

And maybe some more diagrams or something.

Because this is a bit weird and out of blue, and is really the most important thing to remember about the whole Auth / AuthIdentity business -> if they remember this, they will know how to look up the rest.

web/docs/auth/overview.md
Martinsos, line 1: Title "Using Auth" here always confused me. Maybe we should just call it Overview? We dont' have do change this now, but I wanted to at least mention it.

web/docs/auth/overview.md
Martinsos: Yes, this is better!

web/docs/auth/social-auth/github.md
```diff
@@ -503,18 +464,19 @@ app myApp {
...
   wasp: {
     version: "^0.11.0"
   },
   title: "My App",
   auth: {
     userEntity: User,
-    externalAuthEntity: SocialLogin,
     methods: {
       gitHub: {
+        // highlight-next-line
```
Martinsos, line 477: Oho cool I didn't know this exists!

web/docs/auth/social-auth/overview.md
```diff
@@ -122,23 +91,23 @@ To learn more about what the fields on these entities represent, look at the [AP
...
 ## Default Behavior
 
 <DefaultBehaviour />
 
 ## Overrides
 
-Wasp lets you override the default behavior. You can create custom setups, such as allowing users to define a custom username rather instead of getting a randomly generated one.
+By default, Wasp doesn't store any information it receives from the social login provider. It only stores the user's ID specific to the provider.
+
+If you wish to store more information about the user, you can override the default behavior. You can do this by defining the `getUserFieldsFn` and `configFn` functions.
```
Martinsos, line 102: Where? In main.wasp?

web/docs/auth/username-and-pass.md
Martinsos: Sure, well that makes sense!

Here also I think we could benefit from higher-level functions, and of course hooks.

web/docs/tutorial/07-auth.md
```diff
@@ -19,24 +19,16 @@ First, we'll create a Todo list for what needs to be done (luckily we have an ap
 
 ## Creating a User Entity
 
-Since Wasp manages authentication, it expects certain fields to exist on the `User` entity. Specifically, it expects a unique `username` field and a `password` field, both of which should be strings.
+Since Wasp manages authentication, it will create [the auth related entities](/docs/auth/entities) for us. However, we still need to add the `User` entity that will help us keep track of which user owns which tasks.
```
Martinsos, line 22:
```suggestion
Since Wasp manages authentication, it will create [the auth related entities](/docs/auth/entities) for us in the background, that we don't have to worry about. However, we still need to add the `User` entity that will help us keep track of which user owns which tasks.
```

web/docs/tutorial/07-auth.md
```diff
@@ -271,33 +269,41 @@ wasp db studio
...
 
-We see there is a user and that its password is already hashed 🤯
+
+You'll notice that we now have a `User` entity in the database alongside the `Task` entity.
 
 However, you will notice that if you try logging in as different users and creating some tasks, all users share the same tasks. That's because we haven't yet updated the queries and actions to have per-user tasks. Let's do that next.
 
+<small>
+
+You'll notice some extra entities like `Auth` and `AuthIdentity` that Wasp created for us. You don't need to care about these right now, but if you are curious, you can read more about them [here](/docs/auth/entities).
```
Martinsos, line 283:
```suggestion
You might notice some extra Prisma models like `Auth` and `AuthIdentity` that Wasp created for us. You don't need to care about these right now, but if you are curious, you can read more about them [here](/docs/auth/entities).
```

Martinsos: I wrote this comment because these aren't entities really -> they are prisma models, but they are not "Wasp entities". Right? Or wait, are they, we did do some entities injecting in the generator hm? In that case disregard this, but you can still put "might".
Btw how would they notice these? Only if they inspected the DB, right?

web/src/components/Internal.tsx
```diff
@@ -0,0 +1,8 @@
+import * as React from 'react'
+import { Tag } from './Tag'
+
+const color = '#0b62f5'
+
+export function Internal() {
```
Martinsos, line 6: Maybe just write a small comment what is this for, standalone like this it is very hard to say what it is.
You could also name it "InternalTag".

web/src/components/Required.tsx
```diff
@@ -1,22 +1,8 @@
 import * as React from 'react'
+import { Tag } from './Tag'
 
 const color = '#f59e0b'
 
 export function Required() {
```
Martinsos, line 6: Yeah you could probably organize these components under `Tag` and also rename them to InternalTag, RequiredTag, ... .

Martinsos (commented):
@infomiho nice job all together, I think this is in good shape, I left some comments, some are on docs, some on Auth in general due to me understanding some stuff better, check it out.

web/docs/auth/email.md
```diff
@@ -595,7 +584,234 @@ Users can enter their new password there.
...
+  deserializeAndSanitizeProviderData,
+  findAuthIdentity,
+  createUser,
+} from '@wasp/auth/utils.js'
+import {
+  createEmailVerificationLink,
+  sendEmailVerificationEmail,
+} from '@wasp/auth/providers/email/utils.js'
+
+export const signup = async (args, { entities: { User } }) => {
```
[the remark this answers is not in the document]
infomiho: I'm not sure, maybe I mistyped the title and put "login" in it. 

We should probably wait for users to need the custom login action IMHO since I haven't come up with a use-case myself.

web/docs/auth/entities.md
```diff
@@ -0,0 +1,508 @@
...
+```
+
+</TabItem>
+</Tabs>
+
+### `getFirstProviderUserId`
+
+[As mentioned before](#authidentity-entity), the `providerUserId` field is the user's ID in the authentication provider. For example, the user's `username` in the case of the username-based or the user's `email` in the case of the email-based auth. This can be useful if you support multiple authentication methods and you need *any* ID that identifies the user in your app.
+
+This helper function returns the user's first auth identity's provider user ID or `null` if the user doesn't have any auth identities.
```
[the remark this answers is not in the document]
infomiho: I agree it's too technical, I'll simplify 👍

web/docs/auth/entities.md
```diff
@@ -0,0 +1,508 @@
...
+  // ...
+}
+```
+
+</TabItem>
+</Tabs>
+
+## Custom Login and Signup Actions
+
+Let's take a look how you can use the `Auth` and `AuthIdentity` entities to create custom login and signup actions. For example, you might want to create a custom signup action that creates a user in your app and also creates a user in a third-party service.
```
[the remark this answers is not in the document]
infomiho: Makes sense, I'll just link to the example we have for specific providers 👍

web/docs/auth/overview.md
```diff
@@ -200,7 +201,7 @@ page AccountPage {
 ```
 
 ```tsx title="client/pages/Account.tsx"
-import type { User } from '@wasp/entities'
+import type { User } from '@wasp/auth/types'
```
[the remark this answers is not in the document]
infomiho: We can rename the `User` to `AuthenticatedUser` to keep the distinction between the Prisma user with `providerData` as a `string` and the `User` with deserialized `providerData` 🤷

web/docs/auth/overview.md
infomiho: Makes sense 👍

web/docs/tutorial/07-auth.md
infomiho: Yep, they'll notice them when inspecting the DB with Prisma studio :D

web/docs/auth/email.md
infomiho: I've created an issue for this piece, I'll work on it in the near future: https://github.com/wasp-lang/wasp/issues/1631

web/docs/auth/username-and-pass.md
infomiho: Yep, I agree, here's an issue: https://github.com/wasp-lang/wasp/issues/1631

web/docs/auth/email.md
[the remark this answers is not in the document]
infomiho: I've removed the "Login" mention.

web/docs/auth/entities.md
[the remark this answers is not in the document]
infomiho: I've linked to specific providers, but I've kept this example as well since it talks about using the `createUser` helper or just using Prisma directly if they want.

web/docs/auth/overview.md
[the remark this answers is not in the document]
infomiho: I did this for now:
```ts
import { User as AuthenticatedUser } from '@wasp/auth/types'
```

web/docs/auth/overview.md
infomiho: I've added an intro with a code block here 👍

web/docs/auth/overview.md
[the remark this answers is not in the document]
Martinsos: Ha ok, I would prefer the opposite, because it would inform them about what that User really is, but this does help a bit I guess.
I understand of course that it requires bigger refactoring.

Martinsos (approved):
Approved! There are some conflicts, due to me merging `release` with the updated docs into the `main`. I had to fix some links, so that is probably the cause of the conflicts. If you are not sure how to solve them, let me know, I can help, although it should be relatively simple.

Issue #573, opened 2022-04-19
Should we use sessions instead of JWT in Wasp for authentication? What about cookies vs local storage?


sodic:
We currently use token-based authentication (JWT and the `Authorization` header) and persist it inside `localStorage`. We might want to consider switching:
1. Change token-based authentication to session-based authentication
2. Change the persistence mechanism (cookies instead of local storage)

Sessions persisted in cookies seem to be more appropriate for web applications for a multitude of reasons, all exceptionally well covered by [this article](http://cryto.net/~joepie91/blog/2016/06/13/stop-using-jwt-for-sessions/) (as well as [its follow-up](http://cryto.net/~joepie91/blog/2016/06/19/stop-using-jwt-for-sessions-part-2-why-your-solution-doesnt-work/)). I won't repeat the points here, but we can discuss them in the comments.

Martinsos:
Ha interesting! I thought JWT are relatively well accepted solution and I generally liked them because they don't require us storing anything on the backend and are pretty simple. That said, I certainly haven't investigated other alternatives thoroughly, so thanks for sharing this, we should certainly take this into consideration and see what is the best solution.

sodic:
> **they don't require us storing anything on the backend** and are pretty simple

I don't think this is actually true in most cases (perhaps the article mentions it). For example, what should the backend do when the user logs out?
- If it does nothing, an attacker can still use the "invalidated" token to impersonate a logged-out user.
- If it keeps a list of invalidated tokens, we introduce server state. So instead of sessions, we have "negative sessions".

The problem becomes even more apparent when you start scaling the backend servers horizontally. Now all instances need to know about invalidated tokens. In other words, we would need to share the state between the servers (for example, using a database) - all in all, no state-related benefits compared to sessions.

Martinsos:
When logging out with client that uses JWTs, usually that means deleting the JWT from the client side (local storage, cookie, or whatever mechanism is used), and that is it.

You are talking about a specific concern of somebody stealing JWT (at some point when that is possible) and then logging in as user? Sure, that is a possibility, but I don't normally see that connected with the logout in any way. When you use JWTs, normally you won't do any token invalidation at this step - I believe there are other mechanisms to prevent stealing of JWTs, and logout is not considered a mechanism to deal with that.

I guess if done in another way (using sessions and not JWTs), you would expect logout to require next auth attempt to be made exclusively with credentials, therefore eliminating the possibility of smth being stolen during previous communication and being used later?

[...]

PR #1625, opened 2024-01-10
Working auth with Lucia

infomiho:
]


waspc/src/Wasp/Generator/ServerGenerator/AuthG.hs
```diff
@@ -158,3 +154,38 @@ genProvidersTypes auth = return $ C.mkTmplFdWithData [relfile|src/auth/providers
...
+
+    userEntityName = AS.refName $ AS.Auth.userEntity auth
+
+depsRequiredByAuth :: AppSpec -> [AS.Dependency.Dependency]
+depsRequiredByAuth spec = maybe [] (const authDeps) maybeAuth
+  where
+    maybeAuth = AS.App.auth $ snd $ getApp spec
+    authDeps =
+      AS.Dependency.fromList
+        [ ("lucia", "^3.0.0-beta.14"),
```
infomiho, line 189: Lucia v3 is still now officially out, we are using the beta version here. The API shouldn't change with the release as per the author, so I think this should be fine to release with.

waspc/src/Wasp/Generator/DbGenerator/Auth.hs
```diff
@@ -104,17 +111,41 @@ makeAuthEntity userEntityIdField (userEntityName, _) = case parsePslBody authEnt
...
     userFieldOnAuthEntityNameText = T.pack userFieldOnAuthEntityName
     authIdentityEntityNameText = T.pack authIdentityEntityName
     identitiesFieldOnAuthEntityNameText = T.pack identitiesFieldOnAuthEntityName
+    sessionsFieldOnAuthEntityNameText = T.pack sessionsFieldOnAuthEntityName
+    sessionEntityNameText = T.pack sessionEntityName
 
     userEntityIdTypeText = T.pack $ show . Psl.Model.Field._type $ userEntityIdField
     userEntityIdFieldName = T.pack $ Psl.Model.Field._name userEntityIdField
 
+makeSessionEntity :: Generator (String, AS.Entity.Entity)
```
infomiho, line 128: Adding the `Session` entity behind the scenes

waspc/data/Generator/templates/server/src/core/auth.js
```diff
@@ -139,8 +75,4 @@ async function findAvailableUsername(potentialUsernames) {
   return availableUsernames[0]
 }
 
-export function throwInvalidCredentialsError(message) {
```
infomiho, line 142: Moved to `auth/utils.ts`

waspc/data/Generator/templates/server/src/core/auth.js
```diff
@@ -28,68 +15,17 @@ const auth = handleRejection(async (req, res, next) => {
...
-    return identity
-  });
-  return sanitizedUser
-}
-
-const SP = new SecurePassword()
+  req.sessionId = session.id
+  req.user = user
 
-export const hashPassword = async (password) => {
```
infomiho, line 82: Moved to `auth/password.ts`

waspc/data/Generator/templates/server/src/core/auth.js
```diff
@@ -28,68 +15,17 @@ const auth = handleRejection(async (req, res, next) => {
...
-  } catch (error) {
-    if (['TokenExpiredError', 'JsonWebTokenError', 'NotBeforeError'].includes(error.name)) {
-      throwInvalidCredentialsError()
-    } else {
-      throw error
-    }
-  }
+  const { session, user } = await getSessionAndUserFromBearerToken(req);
 
-  const user = await prisma.{= userEntityLower =}
```
infomiho, line 53: Moved to `auth/session.ts`

waspc/data/Generator/templates/server/src/core/auth.js
```diff
@@ -1,23 +1,10 @@
...
-import util from 'util'
 import { randomInt } from 'node:crypto'
 
 import prisma from '../dbClient.js'
 import { handleRejection } from '../utils.js'
-import HttpError from '../core/HttpError.js'
-import config from '../config.js'
-import { deserializeAndSanitizeProviderData } from '../auth/utils.js'
-
-const jwtSign = util.promisify(jwt.sign)
```
infomiho, line 13: Moved to `auth/jwt.ts`

waspc/data/Generator/templates/server/src/auth/utils.ts
```diff
@@ -154,7 +157,10 @@ export async function createUser(
           },
         }
       },
-    }
+    },
+    include: {
```
infomiho, line 161: We need the `Auth.id` to create a session

waspc/data/Generator/templates/server/src/auth/session.ts
```diff
@@ -0,0 +1,102 @@
...
+    };
+  }
+
+  return {
+    session,
+    user: await getUser(authEntity.userId)
+  }
+}
+
+async function getUser(userId: {= userEntityUpper =}['id']): Promise<SanitizedUser> {
```
infomiho, line 58: Moved from `core/auth.js` and added types. 

Now the `SanitizedUser` type and the `getUser` function are finally connected and changes in one place will cause type errors in the other place.

waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
```diff
@@ -116,7 +116,9 @@ export function getSignupRoute({
             await createUser(
                 providerId,
                 newUserProviderData,
-                userFields,
+                // Using any here because we want to avoid TypeScript errors and
+                // rely on Prisma to validate the data.
+                userFields as any
```
infomiho, line 121: Added this `as any` since we had the same coerce in the username auth impl. Added the same explainer comment as there.

waspc/data/Generator/templates/server/src/auth/lucia.ts
```diff
@@ -0,0 +1,43 @@
...
+}>(prismaAdapter, {
+  sessionCookie: {
+    name: "session",
+    expires: true,
+    attributes: {
+      secure: !config.isDevelopment,
+      sameSite: "lax",
+    },
+  },
+  getUserAttributes({ userId }) {
```
infomiho, line 28: We are getting the `userId` from `Auth` so we can fetch the `User` more efficiently.

waspc/data/Generator/templates/server/src/auth/jwt.ts
```diff
@@ -0,0 +1,12 @@
+import jwt from 'jsonwebtoken'
```
infomiho, line 1: Moved from `core/auth.js`

waspc/data/Generator/templates/react-app/src/api.ts
```diff
@@ -8,35 +8,35 @@ const api = axios.create({
   baseURL: config.apiUrl,
```
infomiho, line 8: Rename from token to session ID based name.

pilcrowOnPaper:
Good to see that it's using the v3 beta version. As for a basic timeline, we're planning to ship v3 stable in March or April.

waspc/data/Generator/templates/server/src/core/auth.js
```diff
@@ -28,68 +15,17 @@ const auth = handleRejection(async (req, res, next) => {
...
-      include: {
-        {= authFieldOnUserEntityName =}: {
-          include: {
-            {= identitiesFieldOnAuthEntityName =}: true
-          }
-        }
-      }
-    })
-  if (!user) {
+  if (!session || !user) {
```
Martinsos, line 20: Curious, is it possible that one of these things is truthy and another is not?

waspc/data/Generator/templates/server/src/core/auth.js
```diff
@@ -1,23 +1,10 @@
...
-
-const JWT_SECRET = config.auth.jwtSecret
-
-export const signData = (data, options) => jwtSign(data, JWT_SECRET, options)
-export const sign = (id, options) => signData({ id }, options)
-export const verify = (token) => jwtVerify(token, JWT_SECRET)
+import { getSessionAndUserFromBearerToken } from '../auth/session.js'
+import { throwInvalidCredentialsError } from '../auth/utils.js'
 
 const auth = handleRejection(async (req, res, next) => {
```
Martinsos, line 9: I know we previously had no docs, but it might be nice to put a comment now that says that auth attaches stuff to `req`, and which stuff.

waspc/data/Generator/templates/server/src/routes/auth/logout.ts
```diff
@@ -0,0 +1,14 @@
+import { handleRejection } from '../../utils.js'
+import { throwInvalidCredentialsError } from '../../auth/utils.js'
+import { invalidateSession } from '../../auth/session.js'
+
+export default handleRejection(async (req, res) => {
+  if (req.sessionId) {
+    await invalidateSession(req.sessionId)
+    delete req.sessionId
+    delete req.user
```
Martinsos, line 9: I am curious, why is this important?

waspc/src/Wasp/Generator/DbGenerator/Auth.hs
```diff
@@ -104,17 +111,41 @@ makeAuthEntity userEntityIdField (userEntityName, _) = case parsePslBody authEnt
...
+  where
+    sessionEntityPslBody =
+      T.unpack
+        [trimming|
+          id        String   @id @unique
+          expiresAt DateTime
+
+          // Needs to be called `userId` for Lucia to be able to create sessions
+          userId String
+          // Needs to be called `auth` for Lucia to be able to get sessions
```
Martinsos, line 141: Aha, so this `auth` here has nothing to do with our `Auth`, name-wise? If we renamed `Auth` to `Foo`, this would still be `auth`?

waspc/src/Wasp/Generator/WebAppGenerator/AuthG.hs
```diff
@@ -57,3 +57,8 @@ genUseAuth auth = return $ C.mkTmplFdWithData [relfile|src/auth/useAuth.ts|] tmp
   where
     tmplData = object ["entitiesGetMeDependsOn" .= makeJsArrayFromHaskellList [userEntityName]]
     userEntityName = AS.refName $ AS.Auth.userEntity auth
+
+genLogoutAction :: Generator FileDraft
+genLogoutAction = return $ C.mkTmplFdWithData (C.asTmplFile [relfile|src/auth/logout.ts|]) tmplData
+  where
+    tmplData = object ["logoutPath" .= ("/auth/logout" :: String)]
```
Martinsos, line 64: How do you know this path? It is hardcoded only here. So you are basing it on the knowledge that is embedded in the rest of the Genreator/templates but is not hardcoded explicitly like this. What I guess I am trying to say is that I am not sure if you did much with this move here, it seems more reasonable to me to embed this directly into template instead of passing it via Haskell. But maybe I am missing something?

Martinsos (commented):
@infomiho first review done! Nice comments, helped with guiding me through PR. Everything looks reasonable, I left mostly small comments, check them out.

waspc/src/Wasp/Generator/ServerGenerator/AuthG.hs
infomiho: Info from Lucia's author:

> What I can guarantee
> 1. No more API changes where you'd have to replace the overall code logic (you may need to replace e.g. a() with b())
> 2. No more changes to the database schema
>
> That said, I don't have any more planned changes to Lucia and Arctic

This puts in a good spot to carry on with this 👍

waspc/data/Generator/templates/react-app/src/auth/logout.ts
```diff
@@ -1,7 +1,10 @@
-import { removeLocalUserData } from '../api'
+{{={= =}=}}
+import api, { removeLocalUserData } from '../api'
 import { invalidateAndRemoveQueries } from '../operations/resources'
 
 export default async function logout(): Promise<void> {
+  await api.post('{= logoutPath =}')
```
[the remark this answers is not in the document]
infomiho: Great catch 👍  I'll update the code to clear the local storage even if the request fails

waspc/data/Generator/templates/react-app/src/api.ts
```diff
@@ -8,35 +8,35 @@ const api = axios.create({
...
 })
 
-const WASP_APP_AUTH_TOKEN_NAME = 'authToken'
+const WASP_APP_AUTH_SESSION_ID_NAME = 'sessionId'
 
-let authToken = storage.get(WASP_APP_AUTH_TOKEN_NAME) as string | undefined
+let appSessionId = storage.get(WASP_APP_AUTH_SESSION_ID_NAME) as string | undefined
```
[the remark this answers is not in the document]
infomiho: I've now renamed it to be `waspAppAuthSessionId`, I wanted to avoid using `sessionId` since in the functions below e.g. `setSessionId` the input argument is named `sessionId` and it would shadow this variable.

Before we had `authToken` and `token`, now we have `waspAppAuthSessionId` and `sessionId`

waspc/data/Generator/templates/server/src/auth/session.ts
```diff
@@ -0,0 +1,102 @@
...
+  deserializeAndSanitizeProviderData,
+} from "./utils.js";
+
+import prisma from '../dbClient.js';
+
+// Lucia's Session is tied to the Auth model,
+// so we keep the Auth ID in the session.
+export async function createSession(authId: string) {
+  return auth.createSession(authId, {});
+}
```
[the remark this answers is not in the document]
infomiho: I prefer not to type wrapper functions since most of the time we don't have the 3rd party lib type available to us and typing the return value here would be just `ReturnType<typeof auth.createSession>` which feels pointless since of course it's that since we are returning the `auth.createSession` call.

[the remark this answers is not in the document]
infomiho: About the comment, it is a bit redundant in its current form, I'll update it a bit.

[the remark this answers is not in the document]
infomiho: But here it's relatively easy to import `Session` from `lucia` and type it with `Promise<Session>`, so we can do it 👍

waspc/data/Generator/templates/server/src/auth/session.ts
```diff
@@ -0,0 +1,102 @@
...
+  const authorizationHeader = req.headers["authorization"];
+
+  if (typeof authorizationHeader !== "string") {
+    return {
+      user: null,
+      session: null,
+    };
+  }
+
+  const sessionId = auth.readBearerToken(authorizationHeader ?? "");
```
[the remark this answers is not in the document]
infomiho: I like the way it's written now, it's easy to see the conditions 🤷

waspc/data/Generator/templates/server/src/auth/utils.ts
infomiho: I think I'd like to keep the `createUser` name since this is the purpose of this function. The return type encodes the info that they also receive the `auth` relation.

waspc/data/Generator/templates/server/src/core/auth.js
infomiho: By the current `getSessionAndUserFromBearerToken` implementation it is not possible, but it felt weird only checking one of those things. 

Logically, it okay, both of those things need to be defined and that's why I'm checking it even if it's okay to check only one of the two.

waspc/data/Generator/templates/server/src/routes/auth/logout.ts
infomiho: Good question, I wanted to do it as "better safe than sorry" move. But `req` is valid for only one request, so there is not real need for this **unless** some other middleware is executed after this handler.

We don't currently have anything happening after the request handler, so I'll simplify the code.

waspc/src/Wasp/Generator/DbGenerator/Auth.hs
infomiho: Lucia uses the model name to find the relation name, if we named it `Foo` it would need to be `foo`.

waspc/src/Wasp/Generator/WebAppGenerator/AuthG.hs
infomiho: Makes sense, if I connected the backend and frontend routes via Haskell, it would have made more sense. But I'll simplify 👍

waspc/data/Generator/templates/server/src/auth/lucia.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const prismaAdapter = new PrismaAdapter(
+  // Using `as any` here since Lucia's model types are not compatible with Prisma 4
+  // model types. This is a temporary workaround until we migrate to Prisma 5.
+  // This **works** in runtime, but Typescript complains about it.
+  prisma.{= sessionEntityLower =} as any,
+  prisma.{= authEntityLower =} as any
+);
+
+// TODO: figure out CSRF protection in Lucia - https://v3.lucia-auth.com/guides/validate-session-cookies/express/
```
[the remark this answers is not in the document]
infomiho: I've commented on this issue since it already dealt with the issue of cookies and CSRF protection. https://github.com/wasp-lang/wasp/issues/573

waspc/data/Generator/templates/react-app/src/api.ts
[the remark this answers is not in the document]
Martinsos: Ok I like it!

waspc/data/Generator/templates/server/src/auth/session.ts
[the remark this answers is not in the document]
Martinsos: Ok! I will let you resolve this comment, just in case you wanted to do this but forgot (because I see no :rocket: ). I would vote for typing it, but am ok with skipping.

waspc/data/Generator/templates/server/src/core/auth.js
Martinsos: Ok, makes sense.

waspc/data/Generator/templates/server/src/routes/auth/logout.ts
Martinsos: No worries, you can leave it if you like it, just to be safe as you said.

waspc/src/Wasp/Generator/DbGenerator/Auth.hs
Martinsos: Aha got it! I didn't find that claer from the comment, I would update it.

Also, if I see that correctly, `auth` you hardcoded, but entity name you didn't -> you should also have `auth` dynamically determined then, based on or connected to `authEntityName`

Martinsos (approved):
@infomiho I approved the PR, but there are some minor changes yet to handle, however they should all be straightforward. So please take care of them, until all the comments are resolved, then feel free to merge. You also have some comments where you said you will do something but there is no :rocket:, just :+1: , so keep an eye on them, maybe that is work you have yet to do.

infomiho:
I've created a new issue to update the docs 👍 

https://github.com/wasp-lang/wasp/issues/1637

PR #1628, opened 2024-01-10
Versioned docs

Martinsos:
We now have multiple versions of documentation! I used Docusaurus' feature for this, and combined it the best I could with our `main` / `release` approach. I think it actually makes a lot of sense.

Screenshot from 2024-01-10 21-00-37

The idea is that you can take a snapshot of the current version of docs and number them with a version, and they become a separate, "duplicated" version of docs. You can edit them and all, as normal docs, but you try to touch them as little as can and work mostly on new docs. That is the gist of the story.

Check changes in web/README.md and in waspc/README.md for more details on how this will work.

This PR contains all the changes except for actually versioned docs for 0.11.8, because those are not really something you want to review, so I made a separate PR (https://github.com/wasp-lang/wasp/pull/1629) that merges those into this PR.

Plan for merging this is to, once this PR gets approved, first merge PR with 0.11.8 docs into this one, then merge this PR into `release`, and then later to also merge `release` into `main`.

I should also check if there were any updates to docs on `release` since the creation of this PR -> if there were, I will need to add those to 0.11.8 version of docs.


waspc/README.md
```diff
@@ -366,6 +366,7 @@ If it happens just once every so it is probably nothing to worry about. If it ha
 - If you modified ChangeLog.md or waspc.cabal, create a PR, wait for approval and all the checks (CI) to pass, then squash and merge mentioned PR into `main`.
 - Update your local repository state to have all remote changes (`git fetch`).
 - Update `main` to contain changes from `release` by running `git merge release` while on the `main` branch. Resolve any conflicts.
+- "Freeze"/"Version" the docs by running `npm run docusaurus docs:version {version}` in the [web](/web) dir. Check the README in the `web` dir for more details. Commit this change to `main`.
```
Martinsos, line 369:
```suggestion
- Take a versioned "snapshot" of the current docs by running `npm run docusaurus docs:version {version}` in the [web](/web) dir. Check the README in the `web` dir for more details. Commit this change to `main`.
```

web/docs/contributing.md
```diff
@@ -4,7 +4,7 @@ sidebar_label: Contributing
 slug: /contributing
 ---
 
-import DiscordLink from '../blog/components/DiscordLink';
+import DiscordLink from '@site/blog/components/DiscordLink';
```
Martinsos, line 7: Relative links don't work with versioned docs, so it is recommended to use `@site`, which we use in most places anyway, but we had a couple of places where we didn't use it, which I fixed.

web/docusaurus.config.js
```diff
@@ -56,12 +56,16 @@ module.exports = {
...
       items: [
         {
-          to: "docs/",
-          activeBasePath: "docs",
-          label: "Docs",
-          position: "left",
+          type: 'docsVersion',
+          position: 'left',
+          label: 'Docs',
           className: "navbar-item-docs navbar-item-outside",
```
Martinsos, line 62: I replaced the old "Docs" link in Navbar with this new one, which is made by Docusaurus, and is smarter with regards to versions, it respects current version of docs. So it looks and the same but is better.

web/src/css/custom.css
```diff
@@ -129,6 +129,14 @@
   font-size: var(--ifm-h6-font-size);
 }
 
+/******** DOCS NAVBAR *********/
+
+.navbar__item:has(.navbar-item-docs-version-dropdown:not(.active)) {
```
Martinsos, line 134: This hides Docs version dropdown when not on docs (i.e. when we are on blog).

web/README.md
```diff
@@ -36,4 +43,52 @@ First, ensure you are on the `release` branch. Next, run:
...
+By default, "current" docs are served under URL {baseUrl}/docs/next,
+each versioned doc is served under URL {baseUrl}/docs/{version},
+and last versioned docs (first version in versions.json)
+are served under URL {baseUrl}/docs/, as the default/latest docs.
+
+Since we don't want our users to read `docs/next` ("current" docs), we don't publish these when we deploy docs, instead we build them only during development.
+
+#### When/how do we create new version of docs from "current" docs?
+
+When releasing new version of Wasp, what we do is run `npm run docusaurus docs:version {version}` to create new versioned docs from the current docs. We do this on every new Wasp release.
```
infomiho, line 75: @Martinsos Maybe we want to do this for each **major** version to keep the number of published docs lower?

So the version is `0.11.X` and not `0.11.8` and `0.12.0` and `0.12.1`

infomiho (approved):

web/README.md
Martinsos: I explained live why doing it in such way is not so easy. Approach I proposed is the best fit for our current dynamics. However, what we can do is delete older versions, for example we can delete 0.12.1 if we published 0.12.2. We can even play with labels/paths to achieve having 0.12 all the time, as you suggested, that is also possible, but it would really just be us creating new version of docs, removing previous one, and naming this new one the same way. So the approach wouldn't change, but we can maneuver around it to achieve this if needed.

matijaSos (commented):
LGTM!

PR #1630, opened 2024-01-11
Migration instructions for new auth

infomiho:
Added migration docs for the new auth:
- steps to migrate
- migration scripts for all auth methods
- dev vs. production migration


infomiho:
I'll create a new PR

PR #1634, opened 2024-01-14
Enable easier email dev setup

infomiho:
1. Added a new `Dummy` email sender that developers can set when developing locally and use for the email auth.
2. Removed the `SEND_EMAILS_IN_DEVELOPMENT` env variable since it was mostly confusing people (emails wouldn't be sent in dev even though an email sender was defined).
3. Updated the dummy email sender output.
4. Updated the docs to include the `Dummy` email sender


waspc/src/Wasp/Generator/ServerGenerator/EmailSenderG.hs
```diff
@@ -26,7 +28,12 @@ genEmailSender spec = case maybeEmailSender of
       <++> genCore emailSender
   Nothing -> return []
   where
-    maybeEmailSender = AS.App.emailSender $ snd $ getApp spec
+    maybeEmailSender = AS.App.emailSender (snd $ getApp spec) <|> maybeDummyEmailSender
+    maybeDummyEmailSender =
```
infomiho, line 32: **Only in development** we provide the default dummy email sender.

waspc/src/Wasp/AppSpec/App/EmailSender.hs
Martinsos, line 1: So users need to specify `emailSender: Dummy` in their main.wasp file, right?

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -177,11 +177,11 @@ validateEmailSenderIsDefinedIfEmailAuthIsUsed :: AppSpec -> [ValidationError]
...
   Just auth ->
-    if not $ Auth.isEmailAuthEnabled auth
-      then []
-      else case App.emailSender app of
-        Nothing -> [GenericValidationError "app.emailSender must be specified when using email auth."]
+    if AS.isBuild spec && Auth.isEmailAuthEnabled auth
+      then case App.emailSender app of
+        Nothing -> [GenericValidationError "app.emailSender must be specified when using email auth in production."]
         Just _ -> []
+      else []
```
Martinsos, line 184:
```suggestion
    if AS.isBuild spec && Auth.isEmailAuthEnabled auth && isNothing (App.emailSender app)
      then [GenericValidationError "app.emailSender must be specified when using email auth in production."]
      else []
```

waspc/src/Wasp/Generator/ServerGenerator/EmailSenderG.hs
Martinsos: Ok, so if they haven't specified any email sender, and we are in development, then we behave as if they specified the "dummy email sender".

But they can also specify it manually, right? They can write `emailSender: Dummy` in main.wasp?

When I think about us assuming they want to use dummy when they don't specify anything, I see a potential problem -> maybe they are doing some actual email sending, and even though they haven't specified any real email sender, they are using Wasp's dummy email sender, not maybe even knowing for real that they are using it -> stuff just works, and emails are written in in the console, ok cool. Then they go to deploy, and it will get deployed indeed, but it will crash in production when email is to be sent, because there is no email sender. SOmething like that? Or will it fail during `wasp build` because some code is missing? How will that go?

I wonder if we should require them to always be explicit and do `emailSender: Dummy`. That way we know they want to use email sender, and we can tell them "use the real one now please, because you want to go to production", and we can crash `wasp build` with such nice message.
Yes, this means that when using email auth they have to add this one extra line, but actually that might be good, to make them aware that this is happening.

waspc/src/Wasp/Generator/ServerGenerator/EmailSenderG.hs
```diff
@@ -94,7 +100,7 @@ depsRequiredByEmail spec = maybeToList maybeNpmDepedency
   where
     maybeProvider :: Maybe Providers.EmailSenderProvider
     maybeProvider = getEmailSenderProvider <$> (AS.App.emailSender . snd . getApp $ spec)
-    maybeNpmDepedency = Providers.npmDependency <$> maybeProvider
+    maybeNpmDepedency = maybeProvider >>= Providers.npmDependency
```
Martinsos, line 103: This is not just a simple rewrite, something changed types here, right?

Martinsos (commented):
This is going to be nice -> left some comments!

waspc/src/Wasp/AppSpec/App/EmailSender.hs
```diff
@@ -15,11 +16,23 @@ data EmailSender = EmailSender
...
+data EmailProvider = SMTP | SendGrid | Mailgun | Dummy
   deriving (Eq, Data, Show)
 
 data EmailFromField = EmailFromField
   { name :: Maybe String,
     email :: String
   }
   deriving (Show, Eq, Data)
+
+defaultDummyEmailSender :: EmailSender
```
Martinsos, line 28: Usually we prefer putting default values into Generator, to keep AppSpec simple. Maybe that is the right move here also.

waspc/src/Wasp/Generator/ServerGenerator/EmailSenderG.hs
infomiho: > Then they go to deploy, and it will get deployed indeed, but it will crash in production when email is to be sent, because there is no email sender. SOmething like that? Or will it fail during wasp build because some code is missing? How will that go?

We only provide the dummy email sender implicitly when it's **not a build** therefore, as soon as they try building for deployment, the Wasp compiler would crash with an error that they need to set up an email sender.

infomiho: Hmmmm I'm not sure about what is the better option, if the email sender is explicit but dummy, it works in dev and they are aware they need to change something when they want to get serious. That makes sense to me as well.

waspc/src/Wasp/Generator/ServerGenerator/EmailSenderG.hs
infomiho: The `npmDepedency` turned into a `Maybe` since the `Dummy` sender doesn't have any deps to install alongside it.

waspc/src/Wasp/AppSpec/App/EmailSender.hs
infomiho: Makes sense 👍

waspc/src/Wasp/Generator/ServerGenerator/EmailSenderG.hs
Martinsos: We concluded in live communication to go with explicit!

infomiho: I've now updated the implementation to require setting the `Dummy` email sender explicitly by the user. 

If they attempt to deploy an app with the `Dummy` sender set, developers will get an error.

Martinsos (approved):

PR #1635, opened 2024-01-15
Migration instructions for new auth

infomiho:
Added migration docs for the new auth:
- steps to migrate
- migration scripts for all auth methods
- dev vs. production migration


web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,318 @@
...
+To migrate a deployed app, you need to go through the migration steps while deploying the app in between the steps.
+
+Read more about migrating a deployed app in the [Migrating a Deployed App](#migrating-a-deployed-app) section.
+:::
+
+### 1. Migrate to the New Auth System
+
+If you have an existing app with authentication set up, you can follow these steps to migrate to the new auth system:
+
+1. Upgrade Wasp to the latest version
```
Martinsos, line 92:
```suggestion
1. Upgrade Wasp to the latest 0.12 version.
```

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,318 @@
...
+        import { migrateEmailAuth } from "@server/migrateToNewAuth.js",
+        import { migrateGoogleAuth } from "@server/migrateToNewAuth.js",
+      ]
+    },
+  }
+  ```
+1. Run the migration script by running:
+  ```bash
+  wasp db seed
+  ```
```
Martinsos, line 122: Just one? What if there are multiple? You listed multiple seeds above.

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,318 @@
...
+Read more about migrating a deployed app in the [Migrating a Deployed App](#migrating-a-deployed-app) section.
+:::
+
+### 1. Migrate to the New Auth System
+
+If you have an existing app with authentication set up, you can follow these steps to migrate to the new auth system:
+
+1. Upgrade Wasp to the latest version
+1. Bump the version to `^0.12.0` in `main.wasp`
+1. Create the new auth tables in the database by running:
```
Martinsos, line 94: Maybe tell them the expected results: this should create tables X and Y. So they can check that worked.

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,318 @@
...
+
+  ```bash
+  wasp db migrate-dev
+  ```
+
+1. Create a data migration script in `src/server/migrateToNewAuth.ts`
+
+    - Below you can find [examples of migration scripts](#example-migration-scripts) for each of the auth methods
```
Martinsos, line 102: Why "example" migrations scripts -> are they supposed to modify them? If not, then I would not call them "example", but "prepared" or something like that.

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,318 @@
...
+        import { migrateGoogleAuth } from "@server/migrateToNewAuth.js",
+      ]
+    },
+  }
+  ```
+1. Run the migration script by running:
+  ```bash
+  wasp db seed
+  ```
+1. Verify still works as expected
```
Martinsos, line 123: Could we provide a bit more information here? How should they verify it? What exactly, auth in general? I would give a bit more guidance. Also a dot at the end of the sentence.

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,318 @@
...
+1. Migrate to using the new auth helper functions for getting the `email` or `username` of the currently logged in user. Read more about the helpers in the [Auth Entities](./entities#accessing-the-auth-fields) section.
```
Martinsos, line 124: Makes sense -> but if they didn't migrate these yet, then when they try to verify all works as expected in the step above, they might fail, right? What if they fail there and didn't yet read this next step? Should we put this step above "verifying"? Or make it clear in "verify" step that this might fail?

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,318 @@
...
+
+To migrate a deployed app, you need to go through the migration steps while deploying the app in between the steps.
+
+Read more about migrating a deployed app in the [Migrating a Deployed App](#migrating-a-deployed-app) section.
+:::
+
+### 1. Migrate to the New Auth System
+
+If you have an existing app with authentication set up, you can follow these steps to migrate to the new auth system:
+
```
Martinsos, line 91: I would add step "zero" that tells them again to read the new docs about Auth.

Actually that might be a lot hm -> but I remember I read somewhere exact changes we did. Was that part of some PR you did? Could we use that same text to tell them what changed? It would be nice if they could see in one place, concisely, what changed, otherwise they have to read the whole docs which is PITA. This same text can also go into Changelog potentially.

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,318 @@
...
+1. Remove the `externalAuthEntity` from the `auth` config and the `SocialLogin` entity (if you used it)
+1. Run `wasp db migrate-dev` again to remove the redundant fields from the database
+1. You can now delete the migration script and the `db.seeds` config
+
+
+## Migrating a Deployed App
+
+To migrate a deployed app, you need to go through the migration steps while deploying the app in between the steps.
+
+1. Go through the steps of [Migrate to the New Auth System](#1-migrate-to-the-new-auth-system) on your local machine _until the step where you run the migration script_
```
Martinsos, line 140: "until" -> meaning I don't run the migration script yet, I stop before that?

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,318 @@
...
+2. Deploy the app (#1)
```
Martinsos, line 141: Why do we have #1 here?

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,318 @@
...
+3. Run the migration script on the production database
+
+    :::info Running the migration script on the production database
+
+    If you have an app that is already deployed, you need to run the migration script on the production database.
```
Martinsos, line 146: Seems redundant, this one line

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,318 @@
...
+
+    We wrote instructions on how to do it for **Fly.io** deployments here: https://github.com/wasp-lang/wasp/issues/1464 
+
+    The instructions should be similar for other deployment providers: setting up a tunnel from your local machine to the production database and running the migration script locally with `DATABASE_URL` set to the production database.
```
Martinsos, line 150: SSH tunnel?

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,318 @@
...
+
+Now we need to clean up the old auth system:
+
+1. Delete auth-related fields from `User` entity
+1. Remove the `externalAuthEntity` from the `auth` config and the `SocialLogin` entity (if you used it)
+1. Run `wasp db migrate-dev` again to remove the redundant fields from the database
+1. You can now delete the migration script and the `db.seeds` config
+
+
+## Migrating a Deployed App
```
Martinsos, line 136: I wonder if instead of having this separate section, we should instead add these steps at their right places in the whole flow, and we can say they have to do these if their app is deployed somewhere, if not, they skip it. That sounds more natural to me and easier to follow.

Martinsos, line 136: During this whole process, is there a moment when my app is not working? Or a moment when it is not working correctly? If so, we should inform them about it very clearly.

Also, if there is such moment, we should consider if we avoid it. There is often a way to make migration painless, by first adding new stuff to DB schema while keeping the old code, then updating the code to use new schema, and then also doing the data migration to fill in the new stuff in the db, and finally deleting the old columns. Although even here you have this moment between new schema and doing initial data migration where app might be broken if not implement in a way to support this transition.

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,318 @@
...
+    },
+  });
+
+  for (const user of users) {
+    if (user.auth) {
+      console.log("User was already migrated, skipping", user);
+      continue;
+    }
+
+    if (!user.email || !user.password) {
```
Martinsos, line 225: You don't have this same check for username in the script above, why not?

web/docs/auth/migrate-from-11.md
Martinsos: Also we don't want them to have somebody create new stuff in the old columns in the database after they did the data migration, we have to be wary of that. Better to shut down an app for a minute if needed than to lose some data.

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,318 @@
...
+  const users = await prismaClient.user.findMany({
+    include: {
+      auth: true,
+      externalAuthAssociations: true,
+    },
+  });
+
+  for (const user of users) {
+    if (user.auth) {
+      console.log("User was already migrated, skipping", user);
```
Martinsos, line 287: Is it possible that user was migrated but with another identity? No, right, because right now they can have only one type of identity? However what if they first ran "username" script and that one migrated the user but the user also has oauth creds and now it will skip it here? But that is not possible because we don't allow one user to have both user/pass and oauth, is that so?

Martinsos (commented):
Nice job @infomiho ! Looking quite good, I left some comments but nothing huge.

web/docs/auth/migrate-from-11.md
infomiho: They can different user fields that they might want to migrate in some way. I wrote "they are good as-is, but feel free to modify them" so I'll keep the example name 👍

web/docs/auth/migrate-from-11.md
infomiho: I've written an introduction on this page that says "we had these models before, now we don't have them anymore". I also them them "Read more here" if they are interested :)

web/docs/auth/migrate-from-11.md
infomiho: I've now added extra info for the deployed up inside of the regular migration instructions.

web/docs/auth/migrate-from-11.md
infomiho: Removed it 👍

web/docs/auth/migrate-from-11.md
infomiho: One auth identity per user. Which ever they migrate first will take the identity. This is explained in the Auth Entities docs a bit, but they shouldn't face these issues since mostly they used either Google or email or username.

web/docs/auth/migrate-from-11.md
infomiho: Good catch 👍

web/docs/auth/migrate-from-11.md
Martinsos: That is great, but there are also some other small details, like how can they obtain username now, right?  Some small things like that, would be good if they can quickly learn about that, instead of having to read the whole new docs.

web/docs/auth/migrate-from-11.md
Martinsos: But they could have used both, right? It was allowed. Should we warn them then about this, so they can choose how to handle this? This is connected with my comment above, where I said that since we have this regression, we should probably let them know about it, before they start doing migration, so they can prepare for it and decide how they are going to handle that.

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,338 @@
...
+Now we need to clean up the old auth system:
+
+1. Delete auth-related fields from `User` entity.
+1. Remove the `externalAuthEntity` from the `auth` config and the `SocialLogin` entity (if you used it).
+1. Run `wasp db migrate-dev` again to remove the redundant fields from the database.
+1. You can now delete the migration script and the `db.seeds` config.
+
+:::info Migrating a deployed app
+
+  One last time, _deploy the app_ with the old auth system cleaned up. Your app should now be migrated to the new auth system.
```
Martinsos, line 161: Mention that this deployment, if they are using our Dockerfile for server, will run the new db migrations for them and therefore delete the old data from the db (which is ok, because we migrated it before).

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,338 @@
...
+
+:::
+
+
+### Expect Some Downtime
```
Martinsos, line 166: You can remove this section once you mention this up there. ANd I think we should mention this up there, at the very step where doing this, because otherwise they will not read it on time.

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,338 @@
...
+
+Expect some downtime when migrating a deployed app. The downtime should be minimal, but the auth system won't work 100% while the migration is in progress.
+
+As soon as you deploy the new auth system, users will be able to sign up immediately, but existing users won't be able to log in until you run the migration script on the production database.
+
+## Example Migration Scripts
+
+The provided migration scripts should be fine to use as-is, but you can also modify them to your needs.
```
Martinsos, line 174:
```suggestion
The migration scripts provided below are written with the typical use case in mind, so if that fits your situation you can use them as they are, or if your situation requires additional logic, you can use them as a good starting point.
```

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,338 @@
...
+  ```
+1. Run the migration script(s) by running:
+  ```bash
+  wasp db seed
+  ```
+  If you have multiple migration scripts, you can pick which one to run by selecting it from the list.
+
+  :::info Migrating a deployed app
+
+    _First, deploy your app_ with the changes from previous steps. Then you'll need to run the migration script on the production database with `wasp db seed` command.
```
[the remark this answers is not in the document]
infomiho: I've updated the docs to reflect all these suggestions. 

> Btw, when dealing with local app, they could really just drop the whole database if they don't want to bother with it and don't care about it, right? Should we mention that anywhere?

If they do that, they'll have a mismatch between production and dev if they do the migration differently. And they can't confirm the migration script works.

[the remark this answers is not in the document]
infomiho: I've updated the text now.

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,338 @@
...
+
+    **Running the migration script on the production database**
+
+    We wrote instructions on how to do it for **Fly.io** deployments here: https://github.com/wasp-lang/wasp/issues/1464 
+
+    The instructions should be similar for other deployment providers: setting up some sort of an SSH tunnel from your local machine to the production database and running the migration script locally with `DATABASE_URL` pointing to the production database.
+  :::
+
+1. Migrate to using the new auth helper functions for getting the `email` or `username` of the currently logged-in user. Read more about the helpers in the [Auth Entities](./entities#accessing-the-auth-fields) section.
```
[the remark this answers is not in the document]
infomiho: I've rewrote the steps a bit now and updated the intro to mention backups and staging envs.

web/docs/auth/migrate-from-11.md
infomiho: > That is great, but there are also some other small details, like how can they obtain username now, right? Some small things like that, would be good if they can quickly learn about that, instead of having to read the whole new docs.

I've added this in the later parts of the docs.

> Oh btw what about the fact that they can't anymore have multiple ways of auth per user? That is a regression -> what can they do about that if they had a user with multiple ways to log in?

I've updated the docs 👍

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,349 @@
...
+    After you have run the migration script locally, updated your JS code and saw that it's okay, _deploy the app_ with the changes you have made so far.
+
+    Then you'll need to run the migration script on the production database with `wasp db seed` command.
+
+    **Running the migration script on the production database**
+
+    We wrote instructions on how to do it for **Fly.io** deployments here: https://github.com/wasp-lang/wasp/issues/1464 
+
+    The instructions should be similar for other deployment providers: setting up some sort of an SSH tunnel from your local machine to the production database and running the migration script locally with `DATABASE_URL` pointing to the production database.
+  :::
```
Martinsos, line 159:
```suggestion
  :::info Migrating a deployed app
  
    After successfully performing migration locally so far, and verifying the migrated app works as expected, it is time to also migrate our deployed app (if you have one).
    
    Before migrating your production (deployed) app, we advise you create a copy of your production database in case something goes wrong. Also, besides testing it in development, it might also be advisable to test the migration in the staging environment (if you have one).
    
    Once you are ready, we will perform the production migration in 2 steps:
    1. Deploying the new code to production (client and server).
    2. Migrating the production database.
    
    Between these two steps, so after deploying the new code to production and before you migrate the production database, your app will not be working completely: new users will be able to sign up, but existing users won't be able to log in, and already logged in users will be logged out. Once you do the second step, migrating the production database, it will all be back to normal.
    Therefore, you will likely want to keep the time between the two steps as short as you can. It is advisable to make sure you know exactly what each step entails before doing them for real and also giving them a dry run (or trying it in staging), to eliminate any surprises.
    
    First step: deploy the new code (client and server), either via `wasp deploy` or manually. Check our [Deployment docs](../advanced/deployment/overview.md) for more details. 

    Second step: run the migration script on the production database with `wasp db seed` command. We wrote instructions on how to do it for **Fly.io** deployments here: https://github.com/wasp-lang/wasp/issues/1464 . The instructions should be similar for other deployment providers: setting up some sort of an SSH tunnel from your local machine to the production database and running the migration script locally with `DATABASE_URL` pointing to the production database.
    
    This should be it for the moment! At this point, your deployed app should be working normally.
  :::
```

Martinsos: We should also provide a link to Discord at some place and tell them to ask us for help if not sure about something.

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,349 @@
...
+
+
+### 2. Cleanup the Old Auth System
+
+To finish the migration, we need to clean up the old auth system:
```
Martinsos, line 164:
```suggestion
Our app is now working correctly and using new auth, but to completely finish the migration, we need to clean up the old auth system:
```

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,349 @@
...
+1. Run `wasp db migrate-dev` again to remove the redundant fields from the database.
+1. You can now delete the migration script and the `db.seeds` config.
+
+:::info Migrating a deployed app
+
+  After you make sure things work locally, _deploy the app again_ with the old auth system cleaned up. The database migrations will run on deployment and delete the auth-related `User` columns from the database. 
+  
+  Your app should now be migrated to the new auth system.
+
+:::
```
Martinsos, line 180:
```suggestion
:::info Migrating a deployed app

  After doing the steps above successfully locally and making sure everything is all right, it is time to push these changes to the deployed app again (if you have one).
  
  _Deploy the app again_, either via `wasp deploy` or manually. Check our [Deployment docs](../advanced/deployment/overview.md) for more details. 
  
  The database migrations will automatically run on successful deployment of the server and delete the now redundant auth-related `User` columns from the database. 
  
  Your app is now be fully migrated to the new auth system!
:::
```

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,349 @@
...
+
+
+## Example Migration Scripts
+
+The migration scripts provided below are written with the typical use cases in mind and you can use them as-is. If your setup requires additional logic, you can use them as a good starting point and modify them to your needs.
```
Martinsos, line 185:
```suggestion
## Example Migration Functions

The migration functions provided below are written with the typical use cases in mind and you can use them as-is. If your setup requires additional logic, you can use them as a good starting point and modify them to your needs.
```

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,349 @@
...
+As soon as you deploy the new auth system, new users will be able to sign up, but existing users won't be able to log in until you run the migration script on the production database.
+
+You should probably back up your database before migrating a deployed app. Testing the changes locally or on a staging environment first is also a good idea.
+:::
+
+### 1. Migrate to the New Auth System
+
+If you have an existing app with authentication set up, you can follow these steps to migrate to the new auth system:
+
+1. Upgrade Wasp to the latest 0.12 version.
```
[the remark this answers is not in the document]
infomiho: This feels bit redundant to provide this long command to a specific version. These docs are valid for the latest version.

web/docs/auth/migrate-from-11.md
```diff
@@ -0,0 +1,349 @@
...
+    },
+  }
+  ```
+1. Run the migration script(s) by running:
+  ```bash
+  wasp db seed
+  ```
+  If you added multiple migration scripts, you can pick which one to run by selecting it from the list.
+
+1. Verify that the auth still works locally by logging in with each of the auth methods.
```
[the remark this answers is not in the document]
infomiho: I removed `locally` but kept `auth`

web/docs/auth/migrate-from-11.md
infomiho: We want to say "your" vs. "our" in our docs when we are referring to their app. I'll implement this suggestion 👍

web/docs/auth/migrate-from-11.md
infomiho: > giving them a dry run

How would they do a dry run of the migration steps? I've removed that bit, but also modified the text above to feel more natural to me.

web/docs/auth/migrate-from-11.md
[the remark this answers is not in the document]
Martinsos: My concern was that they won't know how to update Wasp potentially, at least not off the bat. Also, what if Wasp 0.13 or 0.14 is out at the moment when they decide to do this transition, I think we first want them to upgrade to 0.12, right? We can't really write these instructions so they work all the future versions of Wasp -> they will be correct only for 0.12.x . So it sounded good to pin that down here, so they don't make a mistake.

web/docs/auth/migrate-from-11.md
Martinsos: I would try connecting with an SSH tunnel first, without doing anything else, just to make sure that works. But you are right I left it kind of undefined, because I am not sure how much can they "dry run". Ok!

Martinsos (approved):

web/docs/auth/migrate-from-11.md
Martinsos: I checked the new text you applied, looking good.

PR #1636, opened 2024-01-15
Fix typo  css-frameworks.md

ylck:
Fix typo `typographjy` --> `typography`
### Description


infomiho:
Thank you for the fix! 

There is that same file in the `web/docs/css-frameworks.md`, if you would fix that one as well 😄

ylck:
> Thank you for the fix!
> 
> There is that same file in the `web/docs/css-frameworks.md`, if you would fix that one as well 😄

 I'm glad I could help! I've already applied the same fix to the `web/docs/css-frameworks.md` file as well.

Martinsos:
Awesome, thanks @ylck , merging!

PR #1638, opened 2024-01-15
Remove user-facing unverified email flow

infomiho:
We removed the option the have unverified email login to avoid insecure edge cases e.g. signup spamming or impersonating a user with an email that the attacker doesn't own.

We introduce an env variable that's only available when running `wasp start` in dev `IS_EMAIL_AUTO_VERIFIED_IN_DEV` that enables writing easier e2e tests.


waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
```diff
@@ -24,12 +24,12 @@ export function getSignupRoute({
...
     clientRoute,
     getVerificationEmailContent,
-    allowUnverifiedLogin,
+    isEmailAutoVerified,
 }: {
     fromField: EmailFromField;
     clientRoute: string;
     getVerificationEmailContent: GetVerificationEmailContentFn;
-    allowUnverifiedLogin: boolean;
+    isEmailAutoVerified: boolean;
```
Martinsos, line 32: Why is this a variable coming from outside? Why not directly have that logic here inside this function? Because you wanted to templatizing this file? In that case ok, otherwise it would make more sense to me to put it inside here.

Martinsos (approved):
Looking good! Left some comments, but also approved.

I don't see any docs for this new ENV var, should we add some?

Issue #1011, opened 2023-02-17
Use import aliases in Wasp generator templates


infomiho:
Instead of constructing relative imports for each file our templates neede to import, we could use import alises to enable absolute imports. We would need to explore the performance impact if there is any if we do this.

### Relative import
```js
import userFn from '../../ext-src/file.js';
...
userFn();
```

### Absolute import
```js
import userFn from '@ext-src/file.js';
...
userFn();
```

### Client setup

We should probably do something like this:
- setup Typescript paths (`@ext-src` -> `ext-src` dir)
- setup `craco` 

Source: https://devtails.xyz/how-to-use-path-aliases-with-create-react-app-webpack-and-typescript

### Server setup

For the server part we should:
- setup Typescript paths (`@ext-src` -> `ext-src` dir)
- use `tsconfig-paths`

Source: https://stackoverflow.com/a/58226283/1502357

Martinsos:
Very interesting, could simplify our code in the Generator!

PR #1639, opened 2024-01-15
Unify `userSignupFields` for all auth methods

infomiho:
We need a way for developers to specify their `User` model fields that need to be set on signup. 

### Existing impl
We had two different mechanisms: `additionalSignupFields` for `email` and `username` auth and `getUserFieldsFn` for `google` and `github` auth.

The two functions did the same thing: gave us values to set on the `User` during signup e.g. `address` which came from the Google data or the signup form. It makes no more sense to call them differently since they do the same thing. 

`additionalSignupFields` was defined in `auth.signup.addiitonalFields` while `getUserFieldsFn` was defined in both `auth.methods.google` and `auth.methods.github`

### New impl

We now named both of the functions `getUserFieldsFn` and put them in each auth method i.e. added it to the `email` and `usernameAndPassword` methods.

We updated the social login's `getUserFieldsFn` to have the same signature as the previously called `additionalSignupFields` so it has better type support, but we also simplified it (no more `context` argument).

This means that there is no longer a difference in how you provide user fields between auth methods.

### Left to do

- [x] Update the docs
- [x] Update all the example apps


waspc/data/Generator/templates/server/src/auth/providers/username/signup.ts
```diff
@@ -11,33 +11,43 @@ import {
...
+export function getSignupRoute({
+  getUserFieldsFn,
+}: {
+  getUserFieldsFn?: () => ReturnType<typeof defineUserFields>
+}) {
+  return handleRejection(async function signup(req, res) {
+    const fields = req.body ?? {}
+    ensureValidArgs(fields)
+  
+    const userFields = await validateAndGetUserFields(
```
infomiho, line 26: Just this line changed, no other logic changes

waspc/examples/crud-testing/main.wasp
```diff
@@ -9,12 +9,11 @@ app crudTesting {
   auth: {
     userEntity: User,
     methods: {
-      usernameAndPassword: {},
+      usernameAndPassword: {
+        getUserFieldsFn: import { getUserFields } from "@server/auth.js",
```
infomiho, line 13: Username and password finally gets some config :)

waspc/examples/crud-testing/src/server/auth.ts
```diff
@@ -9,21 +9,22 @@ import prisma from '@wasp/dbClient.js'
...
-      .min(10, 'Address must be at least 10 characters long')
-    const result = AddressSchema.safeParse(data.address)
-    if (result.success === false) {
-      throw new Error(result.error.issues[0].message)
-    }
-    return result.data
-  },
-})
+export const getUserFields = () =>
+  defineUserFields({
```
Martinsos, line 13: Not sure if I missed something: so this `defineUserFields` is a new thing?

waspc/examples/todoApp/src/server/auth/google.js
```diff
@@ -7,7 +7,7 @@ export function config() {
   }
 }
 
-export async function getUserFields(_context, args) {
+export function getUserFields({ profile }) {
```
Martinsos, line 10: How is it that context got removed?

waspc/src/Wasp/Generator/ServerGenerator/Auth/LocalAuthG.hs
```diff
@@ -20,33 +24,42 @@ import qualified Wasp.Generator.AuthProviders.Local as Local
...
+          "userFieldsFn" .= extImportToImportJson relPathToServerSrcDir maybeGetUserFieldsFn
         ]
 
+    maybeGetUserFieldsFn = AS.Auth.getUsernameAndPasswordUserFieldsFn usernameAndPasswordConfig
+
     authIndexFileInSrcDir :: Path' (Rel C.ServerSrcDir) File'
     authIndexFileInSrcDir = [relfile|auth/providers/config/username.ts|]
 
+    relPathToServerSrcDir :: Path Posix (Rel importLocation) (Dir C.ServerSrcDir)
+    relPathToServerSrcDir = [reldirP|../../../|]
```
Martinsos, line 61: How did you determine this relative path? Could you maybe explain that in the comment, would that be helpful?

web/docs/auth/social-auth/_getuserfields-type.md
```diff
@@ -1,3 +1,3 @@
-Wasp automatically generates the type `GetUserFieldsFn` to help you correctly type your `getUserFields` function.
+Wasp automatically generates the `defineUserFields` function to help you correctly type your `getUserFields` function.
```
Martinsos, line 1: Aha, that is what it does! Ok, so it types them. Anything else it does? How is it that you switched from type to "hook"?

Martinsos (commented):
Looking good!

Some general questions:
1. It stopped being async? Why is that?
2. No more `context`, why is that?
3. What is the relation between `getUserFieldsFn` and `afterSignup`, a hook that we might define in the future? `getUserFieldsFn` seems like a precursor to `afterSignup`, right? Would you imagine that once we add something like `afterSignup`, `getUserFieldsFn` will become obsolete? Or will it still serve some purpose?

Martinsos:
One more thought: `getUserFieldsFn` seems like a bit weird name. Confuses me as to what it does. Wouldn't something like `defineAdditionalUserFieldsFn` be better? Or `defineAdditionalSignupFieldsFn`, I think you had that before?
If we are going to have something like `afterSignup` hook that replaces the `getUserFieldsFn`, that would also be ok conclusion of this, since `afterSignup` is a quite clear name.

waspc/examples/crud-testing/src/server/auth.ts
infomiho: Yep, we renamed it from `defineAdditionalSignupFields` to `defineUserFields` and now we expect the developer to define a function that returns the `fields` instead of just `fields`

waspc/examples/todoApp/src/server/auth/google.js
infomiho: I wanted to unify the two APIs and decided to remove the `context` argument since I didn't find many uses for it personally. We can return it if people complain?

waspc/src/Wasp/Generator/ServerGenerator/Auth/LocalAuthG.hs
infomiho: This is established pattern in our codebase by now, this is the relative path from the import location to the server src dir. We have it in 3+ places and we didn't add any special comments above, so I would avoid putting it here as well.

web/docs/auth/social-auth/_getuserfields-type.md
infomiho: This is something I worked on when we aded the `additionalSignupFields` feature. It's hard to type the object and force some rules regarding to generics without using an [identity function](https://www.typescriptlang.org/docs/handbook/2/generics.html) like this.

waspc/src/Wasp/Generator/ServerGenerator/Auth/LocalAuthG.hs
Martinsos: Hm ok, although I don't really get what it is. Is this relative part to server src dir from the file that we are generating? What does importLocation mean?
My main question is: in which cases can this path become incorrect? And can we do something about it, be it a comment or code design?

Martinsos (approved):

waspc/src/Wasp/Generator/ServerGenerator/Auth/LocalAuthG.hs
infomiho: Let's deal with these relative import issues when we solve the general issue e.g. introducing aliases https://github.com/wasp-lang/wasp/issues/1011 or rewriting our ext import implementation.

PR #1643, opened 2024-01-18
Adds Session related docs

### LINKED ISSUE #1637: Update auth docs to include `Lucia` and `Session`entity
We need to update our auth docs to mention that Wasp uses Lucia for session management under the hood. Also, the `Session` entity that the Wasp injects automatically needs to be mentioned.

infomiho:
Closes #1637


Martinsos (approved):
Looking good!

PR #1648, opened 2024-01-19
Fixes `HttpError` and `AuthError` `instanceof` bug

infomiho:
We had two definitions of `HttpError` and `AuthError` both in `server` framework code, and in the `sdk`, the `instanceof` check in our error pipeline didn't work properly.

This would lead to unhandled errors in the server process. This PR changes the server code to use the same instance of `HttpError` as the user's code. That way if the user throws a `HttpError` in their operations, our server code can recognise that and handle it properly.


sodic (approved):
All good, we can merge this!

PR #1650, opened 2024-01-20
Updating LSP to work with changes restructuring introduced

Martinsos:
- [x] `waspc/data/Lsp/templates/ts/` -> update the files to work for new restructuring (imports, maybe something else).
- [x] `waspls/src/Wasp/LSP/ExtImport/Path.hs` -> `waspStylePathToCachePath` needs updating, also `tryGetTsconfigForAbsPath`.
- [x] What about that typescript package that we use for parsing js/ts files? Anything needs updating there? -> No, doesn't seem so.
- [x] Test LSP in VSCode to see if anything else needs fixing.


Martinsos:
For start, I quickly read all the code in `waspls/` and identified places which will need updating. I wrote them down in the initial comment in this PR. There aren't many which is great.

sodic:
I'm not sure whether this is already included in one of the bullet points, but quick suggestions should also be updated (I'm guessing that might come automatically when we fix the import paths).

PR #1652, opened 2024-01-21
Restructuring: fix npm deps installation + fix unit tests

Martinsos:
Tested, works well!


sodic (approved):
If it works, it's ok.

PR #1653, opened 2024-01-21
Implement npm workspaces 

infomiho:
We had some issues with duplicate package installations and resolution issues since we are effectively managing multiple packages e.g. sdk, web app and server.

`npm` has a feature called `workspaces` that solves the issue of multiple packages having deps that they need: https://docs.npmjs.com/cli/v8/using-npm/workspaces

Specifying our packages in a list, and running a single `npm install` in the root of the project achieves the following:
- there is only one `node_modules` in the root of the project
- deps are "deduped" automatically e.g. there isn't `react` in the root dir and in the web app dir
- we shouldn't have Prisma and React problems like we did before

> [!WARNING]
> I've tested out this locally and it worked for me with the prototype. This should be tested more with different fail cases we had. Also, `wasp build` should be tested as well.


waspc/src/Wasp/Generator/NpmInstall.hs
```diff
@@ -181,27 +166,3 @@ installProjectNpmDependencies messagesChan projectDir =
           case J._data jobMsg of
             J.JobOutput {} -> printJobMessagePrefixed jobMsg >> processMessages chan
             J.JobExit {} -> return ()
-
-installWebAppAndServerNpmDependencies ::
```
infomiho, line 185: Since the project root `npm install` handles all of the subpackages we define, no need to run separate `npm install` for the subpackages.

PR #1659, opened 2024-01-23
Fixes DB seeding to work with new structure

infomiho:
We had to run `prisma` from the `server` dir since we specified in the server's `package.json` the fields `prisma.seed`.


infomiho:
@sodic I've merged this PR since I needed the Prisma command changes to fix dev DB. @Martinsos and I cowrote the change

PR #1662, opened 2024-01-23
Fixes all Prisma commands that depend on env

infomiho:
We need to run all of our Prisma commands from the `server` dir for Prisma to pick up our `.env` file there like before.

In the future, we might want to reconsider how Prisma and the server env vars interact and change this.


infomiho:
Merging this since we talked about this change IRL and it's straightforward.

PR #1665, opened 2024-01-23
Moved middleware configuration to SDK + fixed sdk/server/config.

Martinsos:


waspc/data/Generator/templates/sdk/server/middleware/globalMiddleware.ts
```diff
@@ -0,0 +1,6 @@
+import { type RequestHandler } from 'express'
+
+export type MiddlewareConfig = Map<string, RequestHandler>
```
Martinsos, line 3: I only extracted these two types to SDK, that was the only thing needed.

waspc/data/Generator/templates/server/src/middleware/globalMiddleware.ts
```diff
@@ -6,6 +6,8 @@ import cors from 'cors'
 import helmet from 'helmet'
 
 import config from 'wasp/server/config'
+import type { MiddlewareConfig, MiddlewareConfigFn } from 'wasp/server/middleware'
```
Martinsos, line 9: These types now moved to SDK, so here I import them from SDK and reexport them. They are used here, but also needed by other parts of server, and it made sense to reexport them.

waspc/examples/todo-typescript/package.json
```diff
@@ -1,13 +1,16 @@
 {
   "name": "prototype",
   "dependencies": {
-    "wasp": "file:.wasp/out/sdk/wasp",
-    "react": "^18.2.0"
+    "cors": "^2.8.5",
```
Martinsos, line 4: Following up on our discussion today, I didn't end up adding `cors` to list of deps in SDK. I was considering it, but realized `cors` is not part of our API. Yes, we do use it in our middleware, but you can provide any middleware you want. By "accident", that can also be cors middleware, and in that case you will write over our cors middleware. But you don't have to use that `cors` library, you can use your own library for cors, or `cors-something` library, or whatever. So it is not part of our API, and this `cors` here is "random".
This makes the situation very clear, so this is the right choice.

PR #1667, opened 2024-01-24
Moved serverSetupFn to SDK.

Martinsos:
Quite simple and straightforward.
Moves some types needed for serverSetup from `server/` to `sdk/server`.

### DIFF
--- a/waspc/data/Generator/templates/sdk/server/types/index.ts
+++ b/waspc/data/Generator/templates/sdk/server/types/index.ts
@@ -1,0 +1,12 @@
+import { type Application } from 'express'
+import { Server } from 'http'
+
+export type ServerSetupFn = (context: ServerSetupFnContext) => Promise<void>
+
+export type ServerSetupFnContext = {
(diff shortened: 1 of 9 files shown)


waspc/src/Wasp/Generator/ServerGenerator.hs
```diff
@@ -323,10 +323,8 @@ genExportedTypesDir spec =
   where
     tmplData =
       object
-        [ "isExternalAuthEnabled" .= isExternalAuthEnabled,
```
Martinsos, line 326: This was some reminder from the past not used anymore so I removed this line.

waspc/data/Generator/templates/sdk/server/types/index.ts
```diff
@@ -0,0 +1,12 @@
+import { type Application } from 'express'
```
Martinsos, line 1: This got moved from server do sdk. That is it really. I also updated one or two paths in server that used it.

PR #1672, opened 2024-01-24
Migrating `vitest` to SDK

infomiho:
You can try out the feature by running `cabal run wasp-cli test client`.


waspc/data/Generator/templates/react-app/src/test/vitest/setup.ts
```diff
@@ -1,4 +1,8 @@
-import matchers from '@testing-library/jest-dom/matchers'
-import { expect } from 'vitest'
+import { afterEach } from 'vitest'
```
infomiho, line 1: I've updated the `vitest` and all the accompanying libraries to the latest versions. I did this because there was an error with using `jsx` files from deps (in our case our SDK) which would mean we would need to bundle our SDK for `vitest` to work again.

Updating to the latest version got rid of the `jsx` imports error, but, this setup still has import errors e.g. it can't resolve the `setupFiles` properly.

waspc/data/Generator/templates/sdk/test/vitest/setup.ts
```diff
@@ -0,0 +1,8 @@
+import { afterEach } from 'vitest'
```
infomiho, line 1: I have this setup file in both web-app and sdk so I could experiment with importing it from different locations e.g. relatively from `.wasp/out/web-app/...` and `wasp/...` and `.wasp/out/sdk./..`

waspc/data/Generator/templates/sdk/tsconfig.json
```diff
@@ -20,7 +20,8 @@
     "esModuleInterop": true,
     "moduleResolution": "node",
     "outDir": "dist",
-    "allowJs": true
+    "allowJs": true,
+    "types": ["@testing-library/jest-dom"],
```
infomiho, line 24: There were type errors related to the unit test matchers e.g. `toBeInTheDocument` was reported as undefined.

waspc/examples/todo-typescript/.gitignore
```diff
@@ -1,4 +1,4 @@
 /.wasp/
 /.env.server
 /.env.client
-/node_modules/
+node_modules/
```
infomiho, line 4: Vitest creates some cache files in the `src` dir.

waspc/examples/todo-typescript/src/MainPage.tsx
```diff
@@ -48,36 +49,6 @@ export const MainPage = ({ user }: { user: User }) => {
   )
 }
 
-function Todo({ id, isDone, description }: Task) {
```
infomiho, line 51: Extracted this component so I can write a test for it.

waspc/examples/todo-typescript/tsconfig.json
```diff
@@ -16,11 +16,12 @@
       "esnext"
     ],
     "allowJs": true,
+    "types": ["@testing-library/jest-dom"],
```
infomiho, line 19: TS complains about `toBeInTheDocument` and other matchers to be undefined if we don't include this. 

This means that the user `tsconfig.json` needs to include this for the type support to work properly.

infomiho: Can we do this differently somehow? @sodic

waspc/cli/src/Wasp/Cli/Command/Test.hs
```diff
@@ -39,7 +42,7 @@ watchAndTest testRunner = do
   watchOrStartResult <- liftIO $ do
     ongoingCompilationResultMVar <- newMVar (warnings, [])
     let watchWaspProjectSource = watch waspRoot outDir ongoingCompilationResultMVar
-    watchWaspProjectSource `race` testRunner outDir
+    watchWaspProjectSource `race` testRunner waspRoot
```
infomiho, line 45: Instead of running from the `.wasp/out` dir, it now runs from the project root dir.

waspc/examples/todo-typescript/tsconfig.json
sodic: Hmm, how did it work before restructuring?

I know [this](https://www.typescriptlang.org/tsconfig#typeRoots) works, but I remember reading a TS issue recently where people complained about TS not looking for types in directories outside of `@types`. I'll try to find it and see what they decided.

Edit: It was [this one](https://github.com/microsoft/TypeScript/issues/36042), kind of related.

waspc/src/Wasp/Generator/Test.hs
```diff
@@ -7,16 +7,18 @@ import Control.Concurrent (newChan)
...
+testWebApp :: [String] -> Path' Abs (Dir WaspProjectDir) -> IO (Either String ())
 testWebApp args projectDir = do
   chan <- newChan
   let testWebAppJob = WebAppTest.testWebApp args projectDir chan
   (testExitCode, _) <-
     testWebAppJob `concurrently` readJobMessagesAndPrintThemPrefixed chan
   case testExitCode of
     ExitSuccess -> return $ Right ()
+    -- Exit code 130 is thrown when user presses Ctrl+C.
+    ExitFailure 130 -> return $ Right ()
```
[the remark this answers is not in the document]
infomiho: I'm not sure, but when you CTRL + C to stop the watching of tests, it outputs an ugly message "It failed with code 130" so I wanted to avoid showing that error to the user.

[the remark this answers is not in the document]
infomiho: I've updated `vitest` so maybe their behaviour changed and now we had to handle this case.

waspc/data/Generator/templates/react-app/vite.config.ts
```diff
@@ -29,12 +30,13 @@ const defaultViteConfig = {
     outDir: "build",
   },
   test: {
+    globals: true,
     environment: "jsdom",
-    setupFiles: ["./src/test/vitest/setup.ts"],
+    // vitest is running from the root of the project, so we need
+    // to specify the path to the setup file relative to the root.
+    setupFiles: {=& vitestSetupFilesArray =},
+    exclude: [...defaultExclude, ".wasp/**/*"]
```
sodic, line 38: Same story as with setup files, `.wasp` is already defined somewhere in Haskell I think

infomiho: Good call, I've now changed the template 👍

sodic (approved):
Great! Merging this.

PR #1674, opened 2024-01-24
Moves email sender to SDK

infomiho:
Moved everything from server generator to SDK generator. Updated email auth imports.


sodic (approved):
LGTM

PR #1675, opened 2024-01-24
Moved GetVerificationEmailContentFn and GetPasswordResetEmailContentF…

Martinsos:
@wasp/types: import { GetVerificationEmailContentFn, GetPasswordResetEmailContentFn } from '@wasp/types’

Now it is imported from `wasp/server/auth/email`.

### DIFF
--- a/waspc/data/Generator/templates/server/src/types/index.ts
+++ b/waspc/data/Generator/templates/server/src/types/index.ts
@@ -1,5 +1,0 @@
-{{={= =}=}}
-
-{=# isEmailAuthEnabled =}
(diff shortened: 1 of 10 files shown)


waspc/data/Generator/templates/server/src/types/index.ts
```diff
@@ -1,5 +0,0 @@
-{{={= =}=}}
```
Martinsos, line 1: This now became redundant, this file.

sodic (approved):
LGTM

PR #1678, opened 2024-01-26
Update .gitignore

hos715:
Make sure the ".env.server.example" file is visible to people

### Description

### DIFF
--- a/mage/.gitignore
+++ b/mage/.gitignore
@@ -8,7 +8,7 @@
 # If this is too agressive for you, consider allowing specific files with `!` operator,
 # or modify/delete these two lines.
 *.env
-*.env.*
+
 
 # Emacs specific
 .projectile


Martinsos:
> Thanks for the PR @hos715 !
> 
> The reason we put `*.env.*` is to protect people from accidentally committing `.env.server` that they might put in wrong place by accident. Or maybe they by accident call it `.env.sevrer` and that gets committed. So we went for a bit aggressive approach, ignoring just to be safe. People can easily edit .gitignore though, to be more permissive.
> 
> .env.server.example is quite common though, so it might make sense to allow for that. How about we instead add explicit rule for that?
> 
> So instead of removing `*.env.*`, we could add:
> 
> ```
> !.env.server.example
> !.env.client.example
> ```
> 
> Or even more general,
> 
> ```
> !*.env.*.example
> !*.env.example
> ```
> 
> I am not 100% sure about those globs though, would be good to test them a bit.

I tested these out, they all work well, so let's go with the last ones.

Martinsos:
I should take #2078 into account while doing this, solve it at the same time possibly.

Interesting: dotenv adds `.env*` to gitignore file. That seems a bit inprecise to me though. But maybe indeed our `*.env` and `*.env.*` is a bit too much, and we could tighten it to `.env` and `.env.*`, so that `.env` must be a prefix.

Martinsos:
Closing as it is superseeded by #2083. Thanks @hos715 for initiating this!

PR #1679, opened 2024-01-26
[New SDK] 'wasp/server/api'

Martinsos:
I didn't just add index.ts because it would be a bit tricky for me to define what to export from the existing file, due to content being dynamically generated, and I didn't want to just export everything because that could easily become incorrect in the future, so I rather just moved the file for now, that seemed like the best ratio of time investment / robustness. We can later reorganize as needed.

I tested it, it works.

Also, it was missing entry in sdk/package.json but it worked because it was just types, which is why you will see only addition in package.json.

As for name of generator -> I named it ServerApiG. Maybe we should dir Server in SdkGenerator, maybe not, I am not sure -> I think we can now do whatever seems best and we will refactor that when doing reorganizing. We will also be smarter then.


waspc/data/Generator/templates/sdk/package.json
```diff
@@ -141,7 +141,11 @@
       {=!  Used by users, documented. =}
       "./webSocket": "./dist/webSocket/index.js",
       {=!  Used by our code, uncodumented (but accessible) for users. =}
-      "./webSocket/WebSocketProvider": "./dist/webSocket/WebSocketProvider.jsx"
+      "./webSocket/WebSocketProvider": "./dist/webSocket/WebSocketProvider.jsx",
+
+      {=! ================= NEW API HERE =================== =}
+      {=! Only public exports =}
+      "./server/api": "./dist/server/api/index.js"
```
Martinsos, line 148: Let's put all the new routes here, so it is clear which ones are public and which ones are not.

waspc/data/Generator/templates/sdk/server/api/index.ts
```diff
@@ -16,6 +14,8 @@ import {
   {=/ shouldImportAuthenticatedApi =}
 } from '../_types'
 
+// PUBLIC API
```
Martinsos, line 17: This seemed useful -> otherwise I have to keep cross-referencing it with docs. It might not be long term solution, because that we need to sync stuff with docs, but maybe it makes sense, not sure yet. We can remove it, not biggy.

waspc/data/Generator/templates/sdk/server/api/index.ts
Martinsos, line 1: This template file moved.

waspc/data/Generator/templates/sdk/server/api/index.ts
```diff
@@ -2,8 +2,6 @@
 
 import { type ParamsDictionary as ExpressParams, type Query as ExpressQuery } from 'express-serve-static-core'
 
-export type { ParamsDictionary as ExpressParams, Query as ExpressQuery } from 'express-serve-static-core'
```
Martinsos, line 5: I checked, nobody is importing this anywhere in our codebase + they are not part of our public docs + they can be imported from `express-serve-static-core` if needed, so I removed these exports.

infomiho:
LGTM, the idea about using a `Server` dir vs. keeping it in the name is appealing to me 👍 

But I agree we can refactor module names later on when we moved more of the APIs

PR #1681, opened 2024-01-27
[New SDK]: 'wasp/client/api'.

Martinsos:


waspc/data/Generator/templates/sdk/package.json
```diff
@@ -144,13 +142,29 @@
...
-      {=! Public: { type MyEntity1, type MyEntity2, ... }  =}
-      "./server/api": "./dist/server/api/index.js"
+      {=! Public: { type MyEntity1, type MyEntity2, ... } =}
+      {=! Private: [] =}
+      "./server/api": "./dist/server/api/index.js",
+      {=! Public: { api }  =}
+      {=! Private: [sdk] =}
+      "./client/api": "./dist/api/index.js"
+   },
+   {=!
```
infomiho, line 152: Good to know 👍

waspc/data/Generator/templates/sdk/package.json
```diff
@@ -144,13 +142,29 @@
       "./webSocket/WebSocketProvider": "./dist/webSocket/WebSocketProvider.jsx",
 
       {=! ================= NEW API HERE =================== =}
-      {=! Public: { type MyEntity1, type MyEntity2, ... }  =}
-      "./server/api": "./dist/server/api/index.js"
+      {=! Public: { type MyEntity1, type MyEntity2, ... } =}
+      {=! Private: [] =}
+      "./server/api": "./dist/server/api/index.js",
+      {=! Public: { api }  =}
+      {=! Private: [sdk] =}
```
infomiho, line 149: What does `Private: [sdk]` mean?

I'd maybe avoid listing out here the PUBLIC vs PRIVATE stuff since you already have comments for each export in the file which is an easy way to check, Keeping the info in two places will become annoying and it error-prone.

infomiho:
Left one comment, but otherwise LGTM

waspc/data/Generator/templates/sdk/package.json
Martinsos: @sodic helped me with that, otherwise it would probably take me quite long to figure this out!

waspc/data/Generator/templates/sdk/package.json
Martinsos: It means it has exports taht are not public but used in SDK.

I am also not sure if this is the best way to log things, but we can easily remove it once we are through with it, I wanted to keep it for now for a short while because it helped me a bit stay organized. Info is in each file and it is duplicated, but this way I can easily get an idea of the situation without entering each file.

PR #1682, opened 2024-01-27
[New SDK]: 'wasp/server'.

Martinsos:


infomiho:
Maybe we should consider exporting `prisma` instead of `prismaClient` since that seems to be the convention Prisma docs use https://www.prisma.io/client and it seems that most people use it like that? Even we did in our codebase.

Martinsos:
> Maybe we should consider exporting `prisma` instead of `prismaClient` since that seems to be the convention Prisma docs use https://www.prisma.io/client and it seems that most people use it like that? Even we did in our codebase.

You are right, I see they do it like this in their example:

```
import { PrismaClient } from '@prisma/client'

const prisma = new PrismaClient()
```

I will rename it!

PR #1693, opened 2024-01-29
Migrate wasp/client/auth API

infomiho:


waspc/data/Generator/templates/sdk/client/auth/index.ts
```diff
@@ -0,0 +1,20 @@
+{{={= =}=}}
+export * from './ui'
+{=# isEmailAuthEnabled =}
+export * from './email'
+{=/ isEmailAuthEnabled =}
+{=# isUsernameAndPasswordAuthEnabled =}
+export * from './username'
```
Martinsos, line 7: Both email and username can contain login and/or signup. So we would be exporting same symbols here. But that is not an issue because two are never enabled at the same time, is that it? Do we have a GH issue to handle this in the future, should we create one?

waspc/data/Generator/templates/sdk/package.json
```diff
@@ -145,7 +121,8 @@
       {=! Public: { api }  =}
       {=! Private: [sdk] =}
       "./client/api": "./dist/api/index.js",
-      "./auth": "./dist/auth/index.js"
+      "./auth": "./dist/auth/index.js",
+      "./client/auth": "./dist/client/auth/index.js"
```
Martinsos, line 125: Ok, so we are not giving them the ability to import from those more specific paths like `client/auth/username`, are we? What are our thoughts on that then, since we said we might want to do that also?

Martinsos (approved):
@infomiho approved, but pls check the comments.

waspc/data/Generator/templates/sdk/client/auth/index.ts
infomiho: >  But that is not an issue because two are never enabled at the same time, is that it?
 
That's correct 👍 

> Do we have a GH issue to handle this in the future, should we create one?

How do you think we should handle it? Drop username or something like that or unify the two login methods?

Martinsos: Not sure -> I don't think anymore we will be dropping the username&pass soon, so I would like into ways to have both.
Either we unify the functions, or we have multiple functions but either export them from different modules `/auth/email`, `auth/username` or modify names `loginWithUsername` `loginWithEmail` or something like that.

Pls create a GH issue, mark it with `auth` and `restructuring` labels and we can further discuss there.

waspc/data/Generator/templates/sdk/auth/forms/Auth.tsx
```diff
@@ -33,6 +33,7 @@ const HeaderText = styled('h2', {
 })
 
 
+// PRIVATE API
```
sodic, line 36: Nice work with this comments! I'm a big fan.

waspc/data/Generator/templates/sdk/auth/index.ts
```diff
@@ -1,7 +1,5 @@
 export { defineUserSignupFields } from './providers/types.js';
 
-// PUBLIC
```
sodic, line 3: What happened with these markers?

waspc/data/Generator/templates/sdk/client/auth/index.ts
sodic: I'm not a big fan of the long names.

Why do you think it's problematic if `login` keeps its name but changes its type signature depending on the auth method users have configured?

waspc/data/Generator/templates/sdk/package.json
sodic: > @sodic and I commented that might be confusing in the end to offer 3 different ways to import the same thing and agreed to reduce the api to this one export.

I seems we misunderstood each other.

I do agree it's confusing to offer the same symbol via multiple imports _**of the same depth**_ (e.g., `VerifyEmailForm` from both `wasp/client/auth/email` and `wasp/client/auth/ui`).

That said, I still think it makes sense to group related symbols together (to enable the `import * as X from 'wasp/client/auth/x` pattern) **and** re-export all of them from `wasp/client/auth`.
This seems OK to me because `wasp/client/auth` is not as deep as `wasp/client/auth/email`.  I've seen this done in other JS libs (e.g., `crypto-js`) and don't find it confusing enough to justify sacrificing the namespacing capabilities it provides.

waspc/data/Generator/templates/sdk/client/auth/index.ts
Martinsos: This convo is related to https://github.com/wasp-lang/wasp/issues/1697 so I linked it there also.

waspc/data/Generator/templates/sdk/package.json
Martinsos: @sodic why do you think namespacing is important here, what does it offer? For me, it gives some context to the person using our SDK and it kind of makes it clear how are different symbols related to each other, which I think has some value, but I am not sure really how important is it.

I wonder if your motivation is similar or it you are driven by something else.

I do however also think that having multiple places to import something from, even if different depth levels, is not ideal, because as a user of the SDK, you might wonder what is the right choice, and there is no very good way to make that decision on what to import, and you might even not be quite sure if those are the same imports or not.

I am not sure what is the solution though. Not having multiple ways to import stuff + having levels of depth? I kind of like that actually, I don't think it all has to be flattened together at the "top level".

Martinsos: This convo is related to https://github.com/wasp-lang/wasp/issues/1697 so I linked it there also.

PR #1694, opened 2024-01-29
Adds wasp/server/auth API

infomiho:


waspc/data/Generator/templates/sdk/package.json
Martinsos, line 1: @infomiho where is the redirect in package.json for new `./server/auth`? I don't see it being added and I don't see it among the existing ones, maybe by accident you forgot to add it? Although I wonder how it works then hm!

Martinsos (changes requested):
I am requesting changes because it seems to me that entry in package.json is missing, I left a comment on that!

Also, the same question as for `wasp/client/auth` PR -> what about the subpaths like `wasp/server/auth/email` and `was/server/auth/username`, we are not exposing those?
What about the fact that some of the identifiers in `wasp/server/auth` are kind of missing context without that extra part in the path?

waspc/data/Generator/templates/sdk/package.json
infomiho: It's weird that it worked, but I've now added the missing export. Good catch! 👍

infomiho:
> Also, the same question as for wasp/client/auth PR -> what about the subpaths like wasp/server/auth/email and was/server/auth/username, we are not exposing those?

As I mentioned in the client API PR, @sodic and I found the argument that the users will be confused that they can import the same thing from multiple places persuasive enough to not export the public API from multiple different places.

> What about the fact that some of the identifiers in wasp/server/auth are kind of missing context without that extra part in the path?

Which of the identifiers do you mean? We talked about not wanting to have `loginWithEmailAndPassword` kind of names for our API if I recall correctly.

Martinsos (approved):

PR #1709, opened 2024-01-31
Update imports for `wasp/client/operations` (ex `wasp/rpc`)

sodic:


waspc/data/Generator/templates/sdk/client/operations/index.ts
```diff
@@ -0,0 +1,22 @@
...
+    type OptimisticUpdateDefinition,
+} from './core'
+
+export {
+    // PUBLIC API
+    configureQueryClient,
+    // PRIVATE API (framework code)
+    initializeQueryClient,
+    // PRIVATE API (framework code)
+    queryClientInitialized
```
Martinsos, line 21: I would maybe then not export these two here from index.ts and let our code obtain that code directly from queryClient.

Martinsos (approved):
LGTM! But check two small comments.

waspc/data/Generator/templates/sdk/client/operations/index.ts
sodic: This wouldn't quite work because of `configureQueryClient`. We agreed to take care of it later.

PR #1713, opened 2024-02-01
Restructuring fixes

sodic:
image


waspc/src/Wasp/Generator/SdkGenerator.hs
```diff
@@ -80,9 +80,7 @@ genSdkReal spec =
     [ genFileCopy [relfile|api/index.ts|],
       genFileCopy [relfile|api/events.ts|],
       genFileCopy [relfile|core/config.ts|],
-      genFileCopy [relfile|core/auth.ts|],
```
sodic, line 83: This file was generated (but weren't used) even when the user didn't define auth. However, the `auth` folder isn't generated under the same conditions.

Therefore, imports in `core/auth.ts` that refer to something from `wasp/auth` would cause build problems in apps that don't use auth.

waspc/src/Wasp/Generator/SdkGenerator.hs
```diff
@@ -80,9 +80,7 @@ genSdkReal spec =
...
       genFileCopy [relfile|core/storage.ts|],
-      genFileCopy [relfile|core/stitches.config.ts|],
```
sodic, line 85: This file needs access to `stitches/react` which  we only install for apps that use auth.

In other words, this file failed the build for apps that don't use auth because of a missing dependency.

waspc/src/Wasp/Generator/SdkGenerator/Auth/AuthFormsG.hs
```diff
@@ -20,6 +20,8 @@ genAuthForms auth =
   sequence
     [ genAuthComponent auth,
       genTypes auth,
+      -- todo (move this to somewhere more meaningful)
+      genFileCopy [relfile|core/stitches.config.ts|],
```
sodic, line 24: I've left this comment here because we probably want to move this file outside of `core`.

waspc/data/Generator/templates/sdk/vite-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="vite/client" />
```
sodic, line 1: Adding this file fixed one of the build errors from the description (the one with `import.meta.env`).

waspc/data/Generator/templates/server/tsconfig.json
```diff
@@ -13,6 +13,23 @@
     ],
     // Filip: end client file hacks
     "strict": false,
+    // When left unspecified, the 'rootDir' defaults to the longest common path of
```
infomiho, line 16: Nice catch

waspc/src/Wasp/Generator/SdkGenerator.hs
infomiho: Stiches is a way to style React component in general, but used only for our Auth UI components. It makes sense to put it somewhere related to auth for now, but in the future we might want to move it 👍

infomiho (approved):
LGTM

Martinsos (approved):

PR #1714, opened 2024-02-01
Prototype of server bundling

infomiho:
We wanted to tackle the issue of requiring our users to always use `.js` extension in the Wasp external imports.

One of the directions we can take is to bundle the server with e.g. Rollup and delegate the module resolution to the bundler.

Note: I've removed `tsc` in the server since we are using `esbuild` to compile TS. We can return `tsc --noEmit` if needed to also perform type checks.


waspc/data/Generator/templates/sdk/server/utils.ts
```diff
@@ -40,33 +40,3 @@ async (req: RequestWithExtraFields, res: Response, next: NextFunction) => {
 }
 
 export const sleep = (ms: number): Promise<unknown> => new Promise((r) => setTimeout(r, ms))
-
-export function getDirPathFromFileUrl(fileUrl: string): string {
```
infomiho, line 44: Since this code relied on importing from a specific directory e.g. `config` and bundling gets rid of the folder structure and bundles everything in one file that we can't use this method anymore.

waspc/data/Generator/templates/server/rollup.config.js
```diff
@@ -0,0 +1,37 @@
+import esbuild from 'rollup-plugin-esbuild'
```
infomiho, line 1: Uses Rollup and the `esbuild` plugin (which is fast and that's want we want in dev)

waspc/data/Generator/templates/server/src/auth/providers/index.ts
```diff
@@ -1,18 +1,15 @@
...
-const whitelistedProviderConfigFileNames = [
-  {=# enabledProviderIds =}
-  "{= . =}.js",
-  {=/ enabledProviderIds =}
+const providers = [
+  {=# providers =}
+  {= importIdentifier =},
+  {=/ providers =}
 ];
-const providers = await importProviders(whitelistedProviderConfigFileNames);
```
infomiho, line 15: Same reason for removing as https://github.com/wasp-lang/wasp/pull/1714/files#r1474701889

waspc/examples/todo-typescript/main.wasp
```diff
@@ -41,7 +41,7 @@ app TodoTypescript {
     system: PostgreSQL
   },
   webSocket: {
-    fn: import { webSocketFn } from "@src/websocket/index.js",
```
infomiho, line 44: Removing extensions and it keeps working 🎉

waspc/src/Wasp/Generator/ServerGenerator/AuthG.hs
```diff
@@ -61,16 +65,25 @@ genAuthRoutesIndex auth = return $ C.mkTmplFdWithDstAndData tmplFile dstFile (Ju
 genProvidersIndex :: AS.Auth.Auth -> Generator FileDraft
 genProvidersIndex auth = return $ C.mkTmplFdWithData [relfile|src/auth/providers/index.ts|] (Just tmplData)
   where
-    tmplData = object ["enabledProviderIds" .= (enabledProviderIds :: [String])]
+    tmplData = object ["providers" .= providers]
 
-    enabledProviderIds =
+    providers =
```
infomiho, line 70: Quick and dirty change to get the whole server working. Maybe this code will need refactoring later.

waspc/data/Generator/templates/server/package.json
```diff
@@ -6,12 +6,12 @@
   "type": "module",
   "comment-filip": "The server.js location changed because we have now included client source files above .wasp/out/server/src.",
   "scripts": {
-    "build": "npx tsc",
-    "start": "npm run validate-env && NODE_PATH=dist node -r dotenv/config dist/.wasp/out/server/src/server.js",
-    "build-and-start": "npm run build && npm run start",
-    "watch": "nodemon --exec 'npm run build-and-start || exit 1'",
+    "bundle": "rollup --config --silent",
+    "start": "npm run validate-env && NODE_PATH=dist node --enable-source-maps -r dotenv/config bundle/server.js",
```
infomiho, line 10: Using source maps for better error reporting 👍

waspc/data/Generator/templates/server/rollup.config.js
```diff
@@ -0,0 +1,37 @@
...
+      }),
+    ],
+    output: [
+      {
+        file: `bundle/server.js`,
+        format: 'es',
+        sourcemap: true,
+      },
+    ],
+    external: (id) => !/^[./]/.test(id),
```
infomiho, line 19: I saw this pattern in the wild, it basically means: anything that's not a relative import, it's a dependency and don't try to bundle it together. https://rollupjs.org/configuration-options/#external

Martinsos: Interesting! We should certainly provide this comment inside this file, since it won't be obvious.

waspc/src/Wasp/Generator/ServerGenerator/AuthG.hs
Martinsos: In which sense, could you elaborate a bit?

Martinsos:
> We wanted to tackle the issue of requiring our users to always use `.js` extension in the Wasp external imports.
> 
> One of the directions we can take is to bundle the server with e.g. Rollup and delegate the module resolution to the bundler.
> 
> Note: I've removed `tsc` in the server since we are using `esbuild` to compile TS. We can return `tsc --noEmit` if needed to also perform type checks.

So it doesn't do typechecking in the terminal any more? That sucks a bit. I know Vite does same thing, but I also think that sucks. I know IDE does the typechecking, but I love (and some other people in Discord said the same, or at least one person) when the actual tool that is running my development also let's me know about errors, so it works even if I don't open it in editor and search for type errors. I won't to know in terminal all is fine, from the "credible" source. But you said we can return `tsc --noEmit` then?

waspc/examples/todo-typescript/main.wasp
infomiho: We have an issue with a different behaviour in our LSP and what our code will tolerate.

| Screenshot 2024-02-05 at 17 06 53 |
| --- |
| LSP doesn't recognise imports from folders - this will run time though |

| Screenshot 2024-02-05 at 17 08 04 |
| --- |
| Adding `index` helps - this will also run fine |

Martinsos: Yeah this is tricky. DO you mind opening an issue for it? You can mark it with `restructuring`. I think we can handle this one after 0.12 is out.

Martinsos (commented):
@infomiho left a bit of comments, but I will let @sodic do the real review and approve this, since he has more knowledge than me. But I like the change!

waspc/examples/todo-typescript/main.wasp
infomiho: I've created an issue for us to tackle this later https://github.com/wasp-lang/wasp/issues/1717

waspc/src/Wasp/Generator/ServerGenerator/AuthG.hs
infomiho: Refactored it a bit in https://github.com/wasp-lang/wasp/pull/1714/commits/d2d1b4dd6122405e22b01ad98cba41fbbbe90327 

Now it feels fine 👍

waspc/data/Generator/templates/sdk/server/utils.ts
infomiho: Yes, we used the dynamic imports based on folder structure, but now there is not folder structure we can rely on. All of the code is bundled together. Not a big deal, we used it only for auth provider imports.

infomiho:
Here's an issue to introduce type checking to the dev setup: https://github.com/wasp-lang/wasp/issues/1718

Issue #398, opened 2021-12-13
Server dies with Prisma error when there are no entities defined


shayneczyzewski:
**Describe the bug**
Server dies with Prisma error when the `main.wasp` file doesn't specify any entities.

**To Reproduce**
1. `wasp new prismaTest`
2. Add ext/static.js file:
```
export const getStaticData = async (args, context) => {
  return { foo: 123 }
}
```
3. Update main.wasp to use it:
```
query getStaticData {
  fn: import { getStaticData } from "@ext/static.js"
}
```
4. `wasp start` produces the following error:
```
Server: /private/tmp/prismatest/.wasp/out/server/node_modules/.prisma/client/index.js:3
Server (stderr):     throw new Error(
Server (stderr):           ^
Server (stderr):
Server (stderr): Error: @prisma/client did not initialize yet. Please run "prisma generate" and try to import it again.
Server (stderr): In case this error is unexpected for you, please report it in https://github.com/prisma/prisma/issues
Server (stderr):     at new PrismaClient (/private/tmp/prismatest/.wasp/out/server/node_modules/.prisma/client/index.js:3:11)
Server (stderr):     at createDbClient (file:///private/tmp/prismatest/.wasp/out/server/src/dbClient.js:4:24)
Server (stderr):     at file:///private/tmp/prismatest/.wasp/out/server/src/dbClient.js:9:18
Server (stderr):     at ModuleJob.run (internal/modules/esm/module_job.js:152:23)
Server (stderr):     at async Loader.import (internal/modules/esm/loader.js:166:24)
Server (stderr):     at async Object.loadESM (internal/process/esm_loader.js:68:5)
Server (stderr): [nodemon] app crashed - waiting for file changes before starting...
Server:
```

<img width="1002" alt="Screen Shot 2021-12-13 at 10 54 44 AM" src="https://user-images.githubusercontent.com/523636/145844724-623f2e0c-252a-4dd9-bc6d-8ef51aa5f2e4.png">

**Expected behavior**
Allow users to use create apps without any Entities defined (e.g., in-memory/static server apps).

**Desktop (please complete the following information):**
 - OS: macOS 12.0.1
 - Wasp: 0.2.2.2

**Additional context**
The fix for now is to define a dummy Entity in main.wasp and run `wasp db migrate-dev` on it:
```
entity Dummy {=psl
    id          Int     @id @default(autoincrement())
psl=}
```

sodic:
Make sure the following features work with this as well:
[...]

PR #1716, opened 2024-02-06
Init PrismaClient only when there are entities

infomiho:
When we have no entities defined in Wasp, and we generate the `PrismaClient`, Prisma generates a dummy client which throws an error when we initialise it.

We now only init the client if there are entities defined.

Closes https://github.com/wasp-lang/wasp/issues/398

### DIFF
--- a/waspc/data/Generator/templates/sdk/server/dbClient.ts
+++ b/waspc/data/Generator/templates/sdk/server/dbClient.ts
@@ -1,11 +1,21 @@
+{{={= =}=}}
+{=# areThereAnyEntitiesDefined =}
 import Prisma from '@prisma/client'
 
-
-const createDbClient = () => {
-  const prisma = new Prisma.PrismaClient()
-
-  return prisma
+function createDbClient(): Prisma.PrismaClient {
+  return new Prisma.PrismaClient()
 }
+{=/ areThereAnyEntitiesDefined =}
+{=^ areThereAnyEntitiesDefined =}
+// * Prisma will not generate a PrismaClient if there no
+//   entities in the schema. Trying to init the PrismaClient
+//   will throw an error.
+// * To avoid throwing an error, we return null if there are no
+//   entities in the schema.
+function createDbClient(): null {
+  return null
+}
+{=/ areThereAnyEntitiesDefined =}
 
 const dbClient = createDbClient()
 
--- a/waspc/src/Wasp/Generator/DbGenerator.hs
+++ b/waspc/src/Wasp/Generator/DbGenerator.hs
@@ -2,6 +2,7 @@ module Wasp.Generator.DbGenerator
   ( genDb,
     warnIfDbNeedsMigration,
     postWriteDbGeneratorActions,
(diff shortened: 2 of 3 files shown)


waspc/data/Generator/templates/sdk/server/dbClient.ts
```diff
@@ -1,11 +1,21 @@
...
+{=/ areThereAnyEntitiesDefined =}
+{=^ areThereAnyEntitiesDefined =}
+// * Prisma will not generate a PrismaClient if there no
+//   entities in the schema. Trying to init the PrismaClient
+//   will throw an error.
+// * To avoid throwing an error, we return null if there are no
+//   entities in the schema.
+function createDbClient(): null {
+  return null
 }
```
Martinsos, line 17: This is all great, the PR, I am just wondering what is the best DX.

So they will get `dbClient` which is null. If they try to use it, they will get TS errors. Will they understand why they are getting them? I don't think so, right? They will just be confused that it is `null`, when it shouldn't be.

I wonder if we can somehow communicate to them what is the issue?

Btw, if they are not using TS, but JS, they will just get runtime error, right?

I guess the improvement is that they will be hitting these issues only if they try to use it, instead of Wasp app dying in any case, but still, would be nice to handle this the best we can.

I wonder if we should just not export `dbClient` for them at all. So they can't import it. At least they will be like "ok, for some reason it doesn't exist", which is semantically correct.

Not sure. Leaving it as you did here is also not terrible, I am just trying to think of some potentially nicer ideas DX-wise. Let me know what you think.

Martinsos (approved):
@infomiho approved, but please check that one comment I left!

waspc/data/Generator/templates/sdk/server/dbClient.ts
infomiho: > So they will get dbClient which is null. If they try to use it, they will get TS errors. Will they understand why they are getting them? I don't think so, right? They will just be confused that it is null, when it shouldn't be.

I guess they won't try to use it if they don't have entities defined? e.g. they would need to do `prisma.something` where `something` is one of entities they defined. 

> I wonder if we should just not export dbClient for them at all. So they can't import it. At least they will be like "ok, for some reason it doesn't exist", which is semantically correct.

We are reexporting the `prisma` symbol from `wasp/server`, so we would need to conditionally render that file now, which is okay. That way we wouldn't import stuff from `dbClient`. We would should conditionally generate the `dbClient.ts`, which is also okay. 

So the option is between:
1. export `null` if not defined _(currently in this PR)_
2. conditionally export from`wasp/server` and generate `dbClient.ts`

The second option is a bit more complex, but not by that much. What do you prefer?

Martinsos: Hm you are right, they are not likely to hit it. Unless they want to do some raw queries with prisma or something, but that is not very likely.

If we had infinite time I would probably vote for option (2), but I think it is just not worth it at the moment, the benefits are not super clear -> let's then stick with what we have in this PR!

PR #1723, opened 2024-02-07
Auth docs rework

infomiho:
Rewrite imports & text for the new API.

### DIFF
--- a/web/docs/auth/email.md
+++ b/web/docs/auth/email.md
@@ -997,7 +997,7 @@ It has the following fields:
 
 - `getEmailContentFn: ServerImport`: a function that returns the content of the e-mail that is sent to the user.
 
-  Defining `getEmailContentFn` can be done by defining a file in the `server` directory.
+  Defining `getEmailContentFn` can be done by defining a file in the `src` directory.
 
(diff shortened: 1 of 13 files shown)


web/docs/auth/email.md
```diff
@@ -167,27 +167,27 @@ Add the following to the `main.wasp` file:
 // 6. Define the routes
 route LoginRoute { path: "/login", to: LoginPage }
 page LoginPage {
-  component: import { Login } from "@client/pages/auth.jsx"
+  component: import { Login } from "@src/pages/auth.jsx"
```
Martinsos, line 170: You can also remove `.jsx`, right? I started removing the extensions in the ext imports in wasp files. Not sure how we want to do it, we don't have to be uniform, but I found that to be the simplest.

Martinsos (approved):
I did a quick scan, but it is looking good, approved!

I left one comment only, regarding which extensions do we use for the ext mports in wasp files.

web/docs/auth/email.md
infomiho: I didn't want to do that now as well, I guess it's fine if we keep some of the extensions?

PR #1726, opened 2024-02-08
Wasp ai 0.12

Martinsos:
Updated Wasp AI core logic to work for Wasp 0.12 .

All seems to work fine, even better than before! Seems that GPT finds new imports less confusing, so I was able to remove the heuristics we had for fixing the imports.


waspc/src/Wasp/AI/GenerateNewProject/PageComponentFile.hs
```diff
@@ -33,93 +21,8 @@ import Wasp.AI.GenerateNewProject.Common
...
 import Wasp.AI.GenerateNewProject.Common.Prompts (appDescriptionBlock)
 import qualified Wasp.AI.GenerateNewProject.Common.Prompts as Prompts
-import Wasp.AI.GenerateNewProject.Operation (Operation)
-import qualified Wasp.AI.GenerateNewProject.Operation as Operation
 import Wasp.AI.GenerateNewProject.Page (makePageDocPrompt)
-import qualified Wasp.AI.GenerateNewProject.Plan as Plan
 import Wasp.AI.OpenAI.ChatGPT (ChatMessage (..), ChatRole (..))
-import Wasp.Util (trim)
-
-fixImportsInPageComponentFile :: FilePath -> [Operation] -> [Operation] -> CodeAgent ()
```
Martinsos, line 43: We don't need this heuristic for fixing import paths any more.

infomiho: Why don't we need it anymore?

infomiho (approved):
LGTM, if you could only explain in the comment why we don't need the import fixing anymore :D

waspc/src/Wasp/AI/GenerateNewProject/PageComponentFile.hs
Martinsos: Before, ChatGPT was struggling with doing correct Wasp imports because it found the un-intuitive -> the fact that import path contains name of operation + the fact operation is default. It would instead often do a mistake and do something much more similar to what we have now.
So I tried removing our "heuristic fix", and it turned out it has no problems with imports any more! Because now imports fit its idea of what they should look like.

PR #1731, opened 2024-02-08
Implement a Tailwind content path wrapper

infomiho:
### Background 

We had an issue with the Tailwind content glob patterns. Since with restructuring we are now directly import user files and no longer copying them to the `ext-src` folder in the framework `web-app` folder, the patterns to find project files were no longer valid. 

For example, we have a pattern `./src/**/*.{js,ts,jsx,tsx}` which would look for files in the`web-app` framework folder. But ... what we really want to look for the in the user defined `src` folder that's three levels above.

### Implementation

Instead of asking users to define their pattern as `'../../..//src/**/*.{js,ts,jsx,tsx}'`, we provided them with a helper i.e. `makeProjectGlobPattern('./src/**/*.{js,ts,jsx,tsx}')` so they can define their content glob patterns with their project root in mind and not know about impl details of Wasp.

### Tailwind `content` option research

> [!NOTE]  
> This is a possible alternative solution, but I don't think it's worth doing now because of the size of the change it requires.

Reading about the Tailwind `content` option here: https://tailwindcss.com/docs/content-configuration says that the glob patterns are relative to the project root instead of the `tailwind.config.cjs` file. 

In the future versions of Tailwind, they'll probably change that so the patterns are relative to the `tailwind.config.cjs` file. You can enable this behaviour _now_ by setting `relative: true` in the config.

Why is this interesting? If we rewrote the way we do Tailwind we could maybe:
1. Generate the `postcss.config.cjs` ourselves (instead of users providing one) and reference the `tailwind.config.cjs` from user dir directly in the `web-app`
2. Users could then put `relative: true` in their `tailwind.config.cjs` and there wouldn't be a need for the `makeProjectGlobPattern` helper, since the glob patterns would be relative to the `tailwind.config.cjs` file, not the `web-app` dir


waspc/data/Generator/templates/sdk/client/index.ts
```diff
@@ -0,0 +1,9 @@
+import * as path from 'path';
+
+/**
+ * Make a pattern relative from Wasp's web-app directory to the project's root.
+ * This enables users to define a glob pattern that is relative to the project's root.
+ */
+export function makeProjectGlobPattern(globPattern: string): string {
```
infomiho, line 7: Users can use this helper when they are defining their glob patterns in the Tailwind config.

Alternatively, we could make a `createTailwindConfig` helper that could go through the `content` array and do this for them? I'm not sure which API has the better DX.

Martinsos: I think `createTailwindConfig` would be a bit much.

I think we might even go in the opposite direction and just offer them `waspProjectRootPath: string`? Then they have even more freedom when using that (it is not specific to globs).

So they would do `waspProjectRootPath + 'src/...'`.

Yeah, this sounds potentially the best to me, although if you don't like it, I am ok with sticking with current solution.

Martinsos (approved):
@infomiho Nice, I approved it!

I suggested one change -> considering going a bit more general with the API we offer, you will see a comment about it. If you like that, I would say great let's do that, if not, ok, let's merge this.

As for that other approach, with relative path -> well that sounds quite great, like a proper solution, but I understand if that is a bit too much for now. Let's at least create an issue for that one?

waspc/data/Generator/templates/sdk/client/index.ts
infomiho: We could rename the helper to say `resolveProjectPath` or similar that would make it more general.

In the Tailwind specific docs we show them how to use our project path helper. And users can use that helper where ever else which we will find out when people start asking questions 😄

Martinsos: Hah yes, why doesn't it work :D.
What I meant is that if we give them string instead of function, and we remove the `glob` word from it, then this thisn we exposed as an API will be more robust to not changing so quickly, which I thought could be beneficial. Yes, there is a risk of them using it incorrectly though, but for some reason that doesn't sound so scary to me.

OK, with this I will leave you to make the final decision on your own.
I would probably go with just a string that has no `glob` in the name, if not then the second option I would go with is function as you have without the `glob` in the name, and the last (but acceptable) option for me is what we currently have which is a function with `glob` in the name.

PR #1742, opened 2024-02-09
Updated examples/thoughts to 0.12.

Martinsos:


examples/thoughts/migrations/20240209145604_init/migration.sql
sodic, line 1: We didn't have migrations in the repo. Just making sure you know you're adding them in.

examples/thoughts/src/client/TopNavbar.jsx
```diff
@@ -1,10 +1,10 @@
-import React from "react";
+import { getUsername } from "wasp/auth";
```
sodic, line 1: Should we reorder the imports?

sodic (approved):
Left two small comments

infomiho (approved):
Tested locally

Martinsos:
> Tested locally

Hah tnx, although no need I also tested it locally of course!

examples/thoughts/migrations/20240209145604_init/migration.sql
Martinsos: I am looking at the code, and we had them already in the repo. I added a new one, along with existing migration(s), and this new one adds new Auth changes (Session), as you can see here.
I called the migration "init" by accident, my fingers were faster than my brain.

PR #1755, opened 2024-02-12
Exports HttpMethod and Route types from wasp/client

infomiho:
We were using `@wasp/types` before in the client tests to import `HttpMethod`, I've now exported it from `wasp/client` because the types were general enough and useful across different parts of Wasp.

### DIFF
--- a/waspc/data/Generator/templates/sdk/auth/useAuth.ts
+++ b/waspc/data/Generator/templates/sdk/auth/useAuth.ts
@@ -2,7 +2,7 @@
 import { deserialize as superjsonDeserialize } from 'superjson'
 import { useQuery, addMetadataToQuery } from 'wasp/client/operations'
 import { api, handleApiError } from 'wasp/client/api'
-import { HttpMethod } from 'wasp/types'
+import { HttpMethod } from 'wasp/client'
 import type { AuthUser } from './types'
 
 // PUBLIC API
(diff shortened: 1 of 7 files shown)


waspc/data/Generator/templates/sdk/auth/useAuth.ts
```diff
@@ -2,7 +2,7 @@
 import { deserialize as superjsonDeserialize } from 'superjson'
 import { useQuery, addMetadataToQuery } from 'wasp/client/operations'
 import { api, handleApiError } from 'wasp/client/api'
-import { HttpMethod } from 'wasp/types'
+import { HttpMethod } from 'wasp/client'
```
sodic, line 5: Maybe this is even general enough to export from `wasp`? Probably too much though.

sodic (approved):

Martinsos (approved):

waspc/data/Generator/templates/sdk/auth/useAuth.ts
Martinsos: Was thinking the same, but probably it doesn't matter much now.

infomiho: It has been used on the client so for, so I felt this made sense. But I get the general feel that `wasp` would make sense.

PR #1756, opened 2024-02-12
Updated docs/introduction/* to 0.12.

Martinsos:
`quick-start.md` and `editor-setup.md` didn't need any changes, while `introduction.md` I did update in this PR, but updates are pretty trivial. The least non-mechanical change is that I reorganized `src/server/recipe.js` into `src/recipe/operations.js`.

### DIFF
--- a/web/docs/introduction/introduction.md
+++ b/web/docs/introduction/introduction.md
@@ -56,7 +56,7 @@ Let's give our app a title and let's immediately turn on the full-stack authenti
 ```wasp title="main.wasp"
 app RecipeApp {
   title: "My Recipes",
-  wasp: { version: "^0.11.0" },
+  wasp: { version: "^0.12.0" },
   auth: {
     methods: { usernameAndPassword: {} },
     onAuthFailedRedirectTo: "/login",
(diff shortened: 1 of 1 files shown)


web/docs/introduction/introduction.md
```diff
@@ -193,7 +192,7 @@ Wasp is addressing the same core problems that typical web app frameworks are ad
 You don't need to know what a DSL is to use Wasp, but if you are curious, you can read more about it below.
 :::
 
-Wasp does not match typical expectations of a web app framework: it is not a set of libraries, it is instead a programming language that understands your code and can do a lot of things for you.
+Wasp does not match typical expectations of a web app framework: it is not a set of libraries, it is instead a simple programming language that understands your code and can do a lot of things for you.
```
infomiho, line 195: Do we maybe want to delang this?

Martinsos: I think so, but let's do it later properly, I did a tiny change here :D.

I actually think it is ok to leave this here for now, until we potentially introduce TS SDK, then we can rephrase it (although I would love mention of DSL, but we can talk a bit less about Wasp being a prog lang).

infomiho (approved):

PR #1759, opened 2024-02-13
Moves `resolveProjectPath` to `wasp/dev`

infomiho:
Since it uses the `path` Node.js module, it should be in `wasp/dev` (or similar) and not in the `wasp/client`.


Martinsos (changes requested):
Great job figuring this out!

I think this is going in right direction, but at the moment I think it shouldn't be `wasp/server`, which is why I am rejecting the PR.

Here is my reasoning, and suggestion on which path to use instead:

The trick is this helper function is actually used in tooling, during development, which is run as node code.

It is client-side tooling (vite) but it is node-js code.

I think if that is so, it doesn't make sense to have in either wasp/client nor wasp/server -> those are SDK that you are supposed to use on client or server. This helper is not supposed to be used in client or server -> it is supposed to be used in tooling, kind of like devDependency.

I would consider having some other path then, something like wasp/utils or wasp/paths as you said, or wasp/dev or something like that.

Same goes for the helper we are using for tailwind config js, right? Are those the same function?

infomiho:
Good line of reasoning.

- `wasp/utils` feels too generic and we might be tempted to make this a catch all module for all the code we don't know where to put
- `wasp/dev` feels off because that helper is useful for building production builds as well
- `wasp/paths` feels a bit too specific for paths, I'd maybe have a bit more generic place (in between paths and utils)

What about `wasp/tooling` or `wasp/project`, something similar that says: "hey this is used with Wasp project tooling"?

infomiho:
After talking offline, we agreed that `wasp/dev` sounds good following the same logic as for `devDependencies` (they are used for both prod and dev stuff)

Martinsos (approved):
Ok great, approved! `wasp/dev` is good enough for now I think.

PR #1765, opened 2024-02-14
Uses string for glob pattern

infomiho:


waspc/src/Wasp/Generator/WebAppGenerator.hs
```diff
@@ -275,7 +275,7 @@ genViteConfig spec = return $ C.mkTmplFdWithData tmplFile tmplData
           "vitest"
             .= object
               [ "setupFilesArray" .= makeJsArrayFromHaskellList vitestSetupFiles,
-                "excludeWaspArtefactsPattern" .= SP.fromRelFile (dotWaspDirInWaspProjectDir </> [relfile|**/*|])
+                "excludeWaspArtefactsPattern" .= (SP.fromRelDir dotWaspDirInWaspProjectDir ++ "**/*")
```
Martinsos, line 278: It seems you are assuming this is a Posix path here, since you are using `"**/*"` and not `"**\*"`, right? If you indeed want to go with Posix path, then you should make sure that `dotWaspDirInWaspProjectDir` is posix, which you can achieve with `relDirToPosix` function. Also, you should import `</>` from `System.FilePath.Posix` and use it to connect the two strings, so that you don't have to rely on StrongPath string always ending with `/` (although it does, but why depend on it if we don't have to).
So, new code could be something like:
```hs
import qualified System.FilePath.Posix as FP.Posix

-- ...


  "excludeWaspArtefactsPattern" .= (SP.fromRelDirP (SP.relDirToPosix dotWaspDirInWaspProjectDir) FP.Posix.</> "**" FP.Posix.</> "*")
```

If Vite doesn't expect Posix path, but instead expects the Path to be equivalent to the standard of the system it is running on (so Win on Windows, Posix on Linux/Mac), then you instead want to make the whole path depend on the system, so it adapts depending on where we run. In that case, we want to go with:

```hs
import qualified System.FilePath as FP

-- ...


  "excludeWaspArtefactsPattern" .= (SP.fromRelDir dotWaspDirInWaspProjectDir FP.</> "**" FP.</> "*")
```

By doing it this way (be it first approach or second, depends on what you need), we ensure that this code works well on Linux, Mac and Windows.
Btw this is why I love StrongPath, it forces me to consider this stuff, which I would never normally consider.

Martinsos (changes requested):

waspc/src/Wasp/Generator/WebAppGenerator.hs
Martinsos: Just to be clear, the current code you have will not work on Windows: first part of path will be using Windows separators, which are `\`, while in the second part you have hardcoded `/`. Nobody uses Wasp yet on Windows directly, true, but we want it to work nevertheless, so we can easily switch once we are ready.

infomiho: It seems that Vite normalizes paths to POSIX separators: https://vitejs.dev/guide/api-plugin.html#path-normalization

Thanks for the heads up, I really didn't think this through it seems. I'll make a new PR with the proposed changes.

infomiho: Vitest default exclude paths are in POSIX as well, so I'll go with using POSIX separators for paths https://vitest.dev/config/#exclude

PR #1766, opened 2024-02-14
Fixes Tailwind setup in client tests

infomiho:
1. When running `wasp test client` Vitest is running from the project root dir
   - **Note:** that's how we must run Vitest in order to detect test files.
2. This causes issues with Tailwind and PostCSS dependencies.
    - PostCSS tries to find Tailwind in the project root dir's `node_modules`, but they are installed inside of `.wasp/out/web-app` and this results in an error:
    
     ```
     Loading PostCSS Plugin failed: Cannot find module 'tailwindcss'
     ```
3. The fix is to move the install of the PostCSS and TailwindCSS deps into the SDK away from WebApp

### DIFF
--- a/waspc/examples/todo-typescript/package-lock.json
+++ b/waspc/examples/todo-typescript/package-lock.json
@@ -32,6 +32,7 @@
         "@testing-library/react": "^14.1.2",
         "@types/express-serve-static-core": "^4.17.13",
         "@vitest/ui": "^1.2.1",
+        "autoprefixer": "^10.4.13",
         "axios": "^1.4.0",
         "express": "~4.18.1",
         "jsdom": "^21.1.1",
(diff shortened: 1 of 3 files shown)


Martinsos (approved):
Makes sense to me, although I didn't understand it 100%.

How are you feeling about this, is this a good solution? Do you feel this is hacky or should be better in the future? If this is a bit hacky, is it a consequence of how we do/run tests at the moment and so it might improve once we improve that?
Any cons to this fix you did?

infomiho:
> How are you feeling about this, is this a good solution? Do you feel this is hacky or should be better in the future? If this is a bit hacky, is it a consequence of how we do/run tests at the moment and so it might improve once we improve that?
Any cons to this fix you did?

I think it's an okay solution since SDK deps get installed in the project root dir `node_modules`. When the `web-app` code wants to resolve Tailwind it goes recursively (first, its `node_modules`, then each parent folder until it gets to project root dir).

If you think about it, the user's code uses Tailwind and now it has the deps it needs in the project root dir `node_modules` which feels okay to me. 

In the future, if we wanted to do the npm workspaces change, then all the deps would be installed in the project root dir anyways. So, I think this solution feels okay. I'm not sure would be the alternative that would work with our current setup.

PR #1767, opened 2024-02-14
Uses POSIX path separators for Vitest exclude path

infomiho:
Vitest's defalut exclude patterns are using POSIX separators, so we'll do the same: https://vitest.dev/config/#exclude


Martinsos:
Ha ok! What would be perfect is to have a piece of Vite docs that says "We always expect Posix separators", just to confirm it, but ok, I think it is likely this is true.
For example, for JS imports, officially they always use posix standard, even if running on Windows, which is great.

Martinsos (approved):

PR #1768, opened 2024-02-14
Updates depedencies docs

infomiho:


web/docs/project/dependencies.md
```diff
@@ -2,12 +2,44 @@
...
+    "react": "^18.2.0",
+    "wasp": "file:.wasp/out/sdk/wasp"
+  },
+  "devDependencies": {
+    "@types/react": "^18.0.37",
+    "prisma": "4.16.2",
+    "typescript": "^5.1.0",
+    "vite": "^4.3.9"
+  }
+}
```
Martinsos, line 31: Suggestion: I wouldn't maybe explain in details how `npm install` works, they ought to know that or check npm docs. Instead, I would say that they can normally use `npm install --save` and similar, or they can manually add deps into package.json and then `wasp` will install them. So either manually or via `npm install`. I wouldn't show them this code example that you have, where dep is added, I think that is out of scope.

What I would do, is warn them that there are already some deps in their package.json that should be there. That is a URL to wasp SDK, and then also react, prisma, typescript, ... . We should tell them to not touch those, right? So I would put focus on that.

Btw, how are they later going to remember which deps in their package.json are theirs, and which are required by Wasp? That will be tricky hm. I wonder if we should have a small comment next to deps required by Wasp, we could just have
```
  "devDependencies": {
    "@types/react": "^18.0.37",
    "prisma": "4.16.2",  // Required by Wasp
    "typescript": "^5.1.0",
    "vite": "^4.3.9"  // Required by Wasp
  }
```

infomiho: > Suggestion: I wouldn't maybe explain in details how npm install works, they ought to know that or check npm docs. Instead, I would say that they can normally use npm install --save and similar, or they can manually add deps into package.json and then wasp will install them. So either manually or via npm install. I wouldn't show them this code example that you have, where dep is added, I think that is out of scope.

I wanted to make it super obvious that this is just `npm` installing stuff and working.

FYI `npm install` since newer versions saves the dep into the `package.json`, no need to use `--save` anymore.

> What I would do, is warn them that there are already some deps in their package.json that should be there. That is a URL to wasp SDK, and then also react, prisma, typescript, ... . We should tell them to not touch those, right? So I would put focus on that.

Good idea 👍 

> Btw, how are they later going to remember which deps in their package.json are theirs, and which are required by Wasp? That will be tricky hm. I wonder if we should have a small comment next to deps required by Wasp, we could just have

You can't put comments in the `package.json` since it's plain JSON unfortunately. I'm not sure how we can say "these are our deps". @sodic might have some ideas

Martinsos: oh wow no need for `--save` anymore? That is awesome, that was never intuitive.

Ah crap this is package.json, yes. There are some techniques to do it, I was doing it recently, and they can even work reliably, but they are not very nice. Ok, we will figure that out later.

Martinsos (approved):

PR #1770, opened 2024-02-14
Implemented a JS script that does part of migration from 0.11 to 0.12.

Martinsos:
It should do all the steps before the auth migration, except for moving deps from main.wasp to package.json, for now that is left to the user since it is a bit complex.

image


waspc/tools/migrate-0-11-to-0-12.js
```diff
@@ -0,0 +1,209 @@
...
+  const projectDirName = process.argv[2].replace(/\/$/, '');
+  if (projectDirName.split(path.sep).length > 1) {
+    console.error('Error: Please provide the name of the project, not the path.');
+    process.exit(1);
+  }
+  return projectDirName;
+}
+
+async function printWarning() {
+ console.log(
```
Martinsos, line 189: NodeJS is a bit of PITA when you have to get user input via the command line, so I found this to be much simpler: do a countdown.

web/docs/introduction/migrate-from-11.md
```diff
@@ -158,25 +157,40 @@ You can easily migrate your old Wasp project to the new structure by following a
...
-3. Create a new project by running `wasp new foo`.
-4. Delete all files of `foo/src` except `vite-env.d.ts`.
-5. If `foo_old/src/client/public` exists and contains any files, copy those files into
+2. Position yourself in the terminal in the directory that is a parent of your wasp project directory (so one level above: if you do `ls`, you should see your wasp project dir listed).
+3. Run the migration script (replace `foo` at the end with the name of your Wasp project directory) and follow the instructions:
+  ```
+    curl -sSL https://raw.githubusercontent.com/wasp-lang/wasp/main/waspc/tools/migrate-0-11-to-0-12.js | node - foo
+  ```
+
+<details>
```
Martinsos, line 172: I moved the manual steps into this thing that can be expanded. They are the same otherwise, except for one small update in the one of the latest steps (I specified some files a bit better after I learned from Filip what exactly is required there).

infomiho:
Ideally, we'd go the extra mile with this, but I'm not too bent on us doing all of the suggestions below.

I'd suggest making this a proper package instead of a single file:
- make a dir
- introduce `package.json`
- use ESM module instead of CommonJS / or introduce Typescript
- introduce a bundling step, so you can use 3rd party deps freely but still link to a single file (if we push the built code to Github)
  - 3rd party deps that might be useful: https://www.npmjs.com/package/commander and https://github.com/google/zx
- make it a executable package (useful for use with `npx`)
- publish it on `npm` under `@wasp/migrate` or something so it can be executed with `npx @wasp/migrate`

Martinsos:
> Ideally, we'd go the extra mile with this, but I'm not too bent on us doing all of the suggestions below.
> 
> I'd suggest making this a proper package instead of a single file:
> 
> * make a dir
> * introduce `package.json`
> * use ESM module instead of CommonJS / or introduce Typescript
> * introduce a bundling step, so you can use 3rd party deps freely but still link to a single file if you want (if we push the built code to Github)
>   
>   * 3rd party deps that might be useful: https://www.npmjs.com/package/commander and https://github.com/google/zx
> * or make it a executable package (useful for use with `npx`) and publish it on `npm` under `@wasp/migrate` or something so it can be executed with `npx @wasp/migrate`

Sounds good! I didn't think it is worth our time putting extra effort into this, but I see you got quite a bit of this working so cool, now we have it!

I merged this one for now, just so we have something working (and so Vinny can test), and then I will go take a look at your PR so we can merge the improvements you did.

PR #1771, opened 2024-02-15
Migrated examples/websockets-realtime-voting to 0.12.

Martinsos:
Went very smoothly!

I used the automated migration script for this, had no issues.

True, auth was already migrated before.


examples/websockets-realtime-voting/src/client/MainPage.tsx
```diff
@@ -20,6 +17,8 @@ const MainPage = () => {
 
   const { socket } = useSocket();
 
+  const username = user ? getUsername(user) : null;
```
infomiho, line 20: @Martinsos If the user is falsy doesn't exist, `getUsername` will return `null`

infomiho (approved):
Looks good, left one comment

PR #1773, opened 2024-02-15
Make build work for 0.12.0

sodic:
## Instructions for testing the build locally
1. Run `wasp build` command in the app's root directory
2. Build and start server by positioning yourself in the app's root directory and running the following commands.

    The commands assume you have a postgres database running in the background.
    Make sure you substitute `<app-name>` and `<database-name>` with real values.
    ```bash
    docker build -f .wasp/build/Dockerfile -t <app-name> .
    docker run \
      --env DATABASE_URL='postgresql://postgres:devpass1234@localhost:5432/<database-name>' \
      --env WASP_WEB_CLIENT_URL='http://localhost:3000' \
      --env JWT_SECRET='593b471137e8fca2db0693966f5db5140363fe6d' \
      --network host \
      <app-name>
    ```
3. Build and start client by positioning yourself in the app's root directory and running:
    ```bash
    cd .wasp/build/web-app
    npm install && REACT_APP_API_URL='http://localhost:3001' npm run build
    cd build
    python3 -m http.server 3000 # or any other way you want to host static files
    ```
4. Go to `localhost:3000` and verify everything's working.

### DIFF
--- a/waspc/cli/src/Wasp/Cli/Command/Build.hs
+++ b/waspc/cli/src/Wasp/Cli/Command/Build.hs
@@ -44,9 +44,9 @@ build = do
   let sdkDir = waspProjectDir </> dotWaspDirInWaspProjectDir </> generatedCodeDirInDotWaspDir </> sdkRootDirInProjectRootDir
   doesSdkDirExist <- liftIO $ doesDirectoryExist sdkDir
   when doesSdkDirExist $ do
-    cliSendMessageC $ Msg.Start "Clearing the content of the .wasp/build directory..."
+    cliSendMessageC $ Msg.Start "Clearing the content of the .wasp/out/sdk directory..."
     liftIO $ removeDirectory sdkDir
-    cliSendMessageC $ Msg.Success "Successfully cleared the contents of the .wasp/build directory."
+    cliSendMessageC $ Msg.Success "Successfully cleared the contents of the .wasp/out/sdk directory."
 
   cliSendMessageC $ Msg.Start "Building wasp project..."
   (warnings, errors) <- liftIO $ buildIO waspProjectDir buildDir
--- a/waspc/cli/src/Wasp/Cli/Command/Build.hs
+++ b/waspc/cli/src/Wasp/Cli/Command/Build.hs
@@ -41,6 +41,8 @@ build = do
     liftIO $ removeDirectory buildDir
     cliSendMessageC $ Msg.Success "Successfully cleared the contents of the .wasp/build directory."
 
+  -- We are using the same SDK location for both build and start. Read this issue
+  -- for the full story: https://github.com/wasp-lang/wasp/issues/1769
   let sdkDir = waspProjectDir </> dotWaspDirInWaspProjectDir </> generatedCodeDirInDotWaspDir </> sdkRootDirInProjectRootDir
   doesSdkDirExist <- liftIO $ doesDirectoryExist sdkDir
   when doesSdkDirExist $ do
--- a/waspc/cli/src/Wasp/Cli/Command/Build.hs
+++ b/waspc/cli/src/Wasp/Cli/Command/Build.hs
@@ -7,11 +7,6 @@ import Control.Monad (when)
 import Control.Monad.Except (throwError)
 import Control.Monad.IO.Class (liftIO)
 import StrongPath (Abs, Dir, Path', (</>))
-import qualified StrongPath as SP
-import System.Directory
-  ( doesDirectoryExist,
-    removeDirectoryRecursive,
-  )
 import Wasp.Cli.Command (Command, CommandError (..))
 import Wasp.Cli.Command.Compile (compileIOWithOptions, printCompilationResult)
 import Wasp.Cli.Command.Message (cliSendMessageC)
@@ -20,9 +15,11 @@ import Wasp.Cli.Message (cliSendMessage)
 import Wasp.CompileOptions (CompileOptions (..))
 import qualified Wasp.Generator
 import Wasp.Generator.Monad (GeneratorWarning (GeneratorNeedsMigrationWarning))
+import Wasp.Generator.SdkGenerator.Common (sdkRootDirInProjectRootDir)
 import qualified Wasp.Message as Msg
 import Wasp.Project (CompileError, CompileWarning, WaspProjectDir)
-import Wasp.Project.Common (buildDirInDotWaspDir, dotWaspDirInWaspProjectDir)
+import Wasp.Project.Common (buildDirInDotWaspDir, dotWaspDirInWaspProjectDir, generatedCodeDirInDotWaspDir)
+import Wasp.Util.IO (doesDirectoryExist, removeDirectory)
 
 -- | Builds Wasp project that the current working directory is part of.
 -- Does all the steps, from analysis to generation, and at the end writes generated code
@@ -37,12 +34,18 @@ build = do
   let buildDir =
         waspProjectDir </> dotWaspDirInWaspProjectDir
           </> buildDirInDotWaspDir
-      buildDirFilePath = SP.fromAbsDir buildDir
 
-  doesBuildDirExist <- liftIO $ doesDirectoryExist buildDirFilePath
+  doesBuildDirExist <- liftIO $ doesDirectoryExist buildDir
   when doesBuildDirExist $ do
     cliSendMessageC $ Msg.Start "Clearing the content of the .wasp/build directory..."
-    liftIO $ removeDirectoryRecursive buildDirFilePath
+    liftIO $ removeDirectory buildDir
+    cliSendMessageC $ Msg.Success "Successfully cleared the contents of the .wasp/build directory."
+
+  let sdkDir = waspProjectDir </> dotWaspDirInWaspProjectDir </> generatedCodeDirInDotWaspDir </> sdkRootDirInProjectRootDir
+  doesSdkDirExist <- liftIO $ doesDirectoryExist sdkDir
+  when doesSdkDirExist $ do
+    cliSendMessageC $ Msg.Start "Clearing the content of the .wasp/build directory..."
+    liftIO $ removeDirectory sdkDir
     cliSendMessageC $ Msg.Success "Successfully cleared the contents of the .wasp/build directory."
 
   cliSendMessageC $ Msg.Start "Building wasp project..."
--- a/waspc/data/Generator/templates/Dockerfile
+++ b/waspc/data/Generator/templates/Dockerfile
@@ -3,7 +3,8 @@
 #   Because if not, we had situations where it would use the different version
 #   locally and on Github CI. This way we ensure exact version is used,
 #   and also have control over updating it (instead of update surprising us).
-FROM node:{= nodeMajorVersion =}-alpine3.17 AS node
+# todo: issue update alpine na 4.18, nema verzije na dockeru. I sto cemo s node verzijom, mozda na 20? 
+FROM node:{= nodeVersion =}-alpine3.17 AS node
 
 
 # We split Dockerfile into base, server-builder and server-production.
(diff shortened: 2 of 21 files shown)


infomiho:
I've tested this locally with:
- username & password auth
- email auth
- Google & Github auth
- email sending work, jobs work, actions and queries work

Building the Docker image on my Apple Silicon Mac doesn't work unless I install the `python3` dependency in an earlier step of the build process

```diff
diff --git a/waspc/data/Generator/templates/Dockerfile b/waspc/data/Generator/templates/Dockerfile
index d5449d201..c1f3437c4 100644
--- a/waspc/data/Generator/templates/Dockerfile
+++ b/waspc/data/Generator/templates/Dockerfile
@@ -21,7 +21,7 @@ RUN apk --no-cache -U upgrade # To ensure any potential security patches are app

 # Todo this image stays on disk under <none>:<none>, should we remove it? Or is it useful for future builds?
 FROM base AS server-builder
-RUN apk add --no-cache build-base libtool autoconf automake
+RUN apk add --no-cache python3 build-base libtool autoconf automake
 WORKDIR /app
 COPY src ./src
 COPY package.json .
```

waspc/cli/src/Wasp/Cli/Command/Build.hs
```diff
@@ -37,12 +34,18 @@ build = do
...
   when doesBuildDirExist $ do
     cliSendMessageC $ Msg.Start "Clearing the content of the .wasp/build directory..."
-    liftIO $ removeDirectoryRecursive buildDirFilePath
+    liftIO $ removeDirectory buildDir
+    cliSendMessageC $ Msg.Success "Successfully cleared the contents of the .wasp/build directory."
+
+  let sdkDir = waspProjectDir </> dotWaspDirInWaspProjectDir </> generatedCodeDirInDotWaspDir </> sdkRootDirInProjectRootDir
+  doesSdkDirExist <- liftIO $ doesDirectoryExist sdkDir
+  when doesSdkDirExist $ do
+    cliSendMessageC $ Msg.Start "Clearing the content of the .wasp/build directory..."
```
infomiho, line 47: Is this message correct since you are deleting the `sdk` dir, and not the whole `build` dir?

waspc/data/Generator/templates/Dockerfile
```diff
@@ -18,18 +19,23 @@ FROM node AS base
...
+COPY package-lock.json .
+COPY .wasp/build/server .wasp/build/server
+COPY .wasp/out/sdk .wasp/out/sdk
 # Install npm packages, resulting in node_modules/.
-RUN cd server && npm install
+RUN npm install && cd .wasp/build/server && npm install
 {=# usingPrisma =}
-COPY db/schema.prisma ./db/
-RUN cd server && npx prisma generate --schema='{= dbSchemaFileFromServerDir =}'
+COPY .wasp/build/db/schema.prisma .wasp/build/db/
```
infomiho, line 34: Why did you have to do this?

waspc/data/Generator/templates/react-app/vite.config.ts
```diff
@@ -29,6 +29,12 @@ const defaultViteConfig = {
   build: {
     outDir: "build",
   },
+  resolve: {
+    // These packages rely on a single instance per page. Not dedpuing them
+    // causes runtime errors (e.g., hook rule violation in react, QueryClient
+    // instance error in react-query, Invariant Error in react-router-dom).
+    dedupe: ["react", "react-dom", "@tanstack/react-query", "react-router-dom"]
```
infomiho, line 36: As I commented in the DMs, I think we should pick the plugin that's the most used/supported. Then we should list these dedups explicitly like you just did.

waspc/data/Generator/templates/sdk/client/webSocket/WebSocketProvider.tsx
```diff
@@ -31,7 +36,7 @@ apiEventsEmitter.on('sessionId.set', refreshAuthToken)
 apiEventsEmitter.on('sessionId.clear', refreshAuthToken)
 
 // PRIVATE API
-export const WebSocketContext = createContext({
+export const WebSocketContext: Context<WebSocketContextValue> = createContext({
```
infomiho, line 39: Interesting that it couldn't infer this type 🤨

waspc/src/Wasp/Generator/SdkGenerator/Common.hs
```diff
@@ -46,7 +47,10 @@ mkTmplFd :: Path' (Rel SdkTemplatesDir) File' -> FileDraft
 mkTmplFd path = mkTmplFdWithDst path (SP.castRel path)
 
 sdkRootDirInProjectRootDir :: Path' (Rel ProjectRootDir) (Dir SdkRootDir)
-sdkRootDirInProjectRootDir = [reldir|sdk/wasp|]
+sdkRootDirInProjectRootDir =
```
infomiho, line 50: This seems very peculiar, could you give more context here?

infomiho (commented):
I know you said you'll add extra context, but I commented a bit

waspc/data/Generator/templates/react-app/vite.config.ts
sodic: OK, leaving this one then (it's much more popular).

waspc/data/Generator/templates/sdk/tsconfig.json
```diff
@@ -8,6 +8,7 @@
       "dom",
       "DOM.Iterable"
     ],
+    "declaration": true,
```
sodic, line 11: Here's the full story for this fix.

When trying the build the web-app, I got:
```
src/router.tsx:28:23 - error TS2339: Property 'to' does not exist on type 'unknown'.

28           path={route.to}
                         ~~

Found 1 error in src/router.tsx:28
```
To inspect further, I opened the generated `web-app` project in VS code and hovered over the import. TypeScript LSP reported:
```
Could not find a declaration file for module 'wasp/client/router'.
'.../todo-typescript/.wasp/out/sdk/wasp/dist/client/router/index.js' implicitly has an 'any' type.

There are types at '../todo-typescript/node_modules/wasp/client/router/index.ts',
but this result could not be resolved when respecting package.json "exports".
The 'wasp' library may need to update its package.json or typings.
```
By reading through https://github.com/microsoft/TypeScript/issues/52363 and [the TSconfig reference](https://www.typescriptlang.org/tsconfig#declaration), I concluded we needed to emit declarations.

Adding the `declaration: true` compile option fixed the issue, but caused different type-related problems (more on this in other comments).

waspc/data/Generator/templates/sdk/server/_types/serialization.ts
```diff
@@ -18,7 +18,7 @@ export interface JSONObject {
 
 type PrimitiveJSONValue = string | number | boolean | undefined | null
 
-interface JSONArray extends Array<JSONValue> {}
+export interface JSONArray extends Array<JSONValue> {}
```
sodic, line 21: After adding the `declaration: true` field to `sdk/tsconfig.json` (see https://github.com/wasp-lang/wasp/pull/1773/files#r1492341617), job definition started throwing errors such as:
```
Exported variable 'printTimeAndNumberOfTasks' has or is using name 'JSONArray' 
from external module ".../todo-typescript/.wasp/out/sdk/wasp/server/_types/serialization"
but cannot be named.
```
By reading https://github.com/microsoft/TypeScript/issues/5711 and [this StackOverflow question](https://stackoverflow.com/questions/43900035/ts4023-exported-variable-x-has-or-is-using-name-y-from-external-module-but), I tried to export the `JSONArray` interface, which solved the problem.

waspc/data/Generator/templates/sdk/client/webSocket/WebSocketProvider.tsx
sodic: It could infer the type, but couldn't construct a valid declaration file without the type's name in scope:
```
Exported variable 'WebSocketContext' has or is using name 'ClientToServerEvents' 
from external module ".../todo-typescript/.wasp/out/sdk/wasp/ext-src/websocket/index" 
but cannot be named.
```
 See [this comment](https://github.com/wasp-lang/wasp/pull/1773/files#r1492398790) for more details.

Also, this export should be explicitly typed anyway according to the [module boundaries rule](https://github.com/wasp-lang/vscode-wasp/pull/6#discussion_r970774084).

I've created an issue for explicitly typing our public API in SDK: https://github.com/wasp-lang/wasp/issues/1776

waspc/src/Wasp/Generator/SdkGenerator/Common.hs
sodic: Added a comment above the function.

waspc/data/Generator/templates/sdk/package.json
```diff
@@ -97,6 +97,7 @@
       {=! Public: { type WebSocketDefinition, type WaspSocketData } =}
       {=! Private: [server, sdk] =}
       "./server/webSocket": "./dist/server/webSocket/index.js",
+      "./entities": "./dist/entities/index.js",
```
sodic, line 100: When building the prototype without this export, I got the following error:
```
../../../src/MainPage.tsx:11:27 - error TS2307: Cannot find module 'wasp/entities' or its corresponding type declarations.

11 import type { Task } from 'wasp/entities'
                             ~~~~~~~~~~~~~~~

../../../src/Todo.tsx:2:22 - error TS2307: Cannot find module 'wasp/entities' or its corresponding type declarations.

2 import { Task } from 'wasp/entities'
                       ~~~~~~~~~~~~~~~


Found 2 errors in 2 files.

Errors  Files
     1  ../../../src/MainPage.tsx:11
     1  ../../../src/Todo.tsx:2
```
It seems that TS LSP and VS Code know how to find the correct types and provide IDE support without needing an explicit `exports` definition.
TSC, on the other hand, requires the package to export all symbols the user code references (types or otherwise).

waspc/src/Wasp/Generator/DockerGenerator.hs
```diff
@@ -46,7 +45,7 @@ genDockerfile spec = do
           object
             [ "usingPrisma" .= not (null $ AS.getDecls @AS.Entity.Entity spec),
               "dbSchemaFileFromServerDir" .= SP.fromRelFile dbSchemaFileFromServerDir,
-              "nodeMajorVersion" .= show (SV.major $ getLowestNodeVersionUserAllows spec),
+              "nodeVersion" .= show (getLowestNodeVersionUserAllows spec),
```
sodic, line 48: Using a major can create problems for users. For example:
- `getLowestNodeVersionUserAllows spec` returns `18.6.0`, which means "You need at least Node 18.6.0 to run my app"
- We only specify the major version for the Docker image (in this case, `18`). The base image can therefore include any version that's backwards compatible with `18.0.0`
- Let's say that we end up building an image with Node `18.3.0`. If the user's code relies on changes introduced after `18.3.0` and before `18.6.0`, the build can fail or behave unexpectedly.

Therefore, it's always best to stick with the user's exact desired version.

The only question is - how can we be sure of this specific image's availability in DockerHub.

infomiho: I guess, users will report if something breaks and then we'll have to think about it? 🤷 

I have a feeling that the Node.js images are built automatically on release of new Node.js images: https://github.com/nodejs/docker-node/blob/main/.github/workflows/automatic-updates.yml

waspc/src/Wasp/Generator/SdkGenerator/Common.hs
infomiho: Understood 👍

waspc/src/Wasp/Generator/SdkGenerator.hs
```diff
@@ -212,6 +212,20 @@ npmDepsForSdk spec =
...
+            -- Before 0.12.0 (i.e., restructuring), we made Wasp use
+            -- sodium-native 3.3.0. with package.json overrides in our web-app/package.json:
+            -- https://github.com/wasp-lang/wasp/pull/729
+            --
+            -- Because our code that uses secure-password now lives in the SDK,
+            -- and NPM apparently ignores package.json overrides for
+            -- dependencies (no reference, found out by trying it out), the only
+            -- way to force NPM to install sodium-native 3.3.0 is by listing it
+            -- as a direct dependency.
+            ("sodium-native", "3.3.0"),
```
sodic, line 228: Tracked here: https://github.com/wasp-lang/wasp/issues/1779

waspc/src/Wasp/Generator/DockerGenerator.hs
sodic: Had the same feeling but was too lazy to verify :upside_down_face:

That's great news if true.

I linked this conversation from the issue.

PR #1780, opened 2024-02-17
Enables e2e tests and build test

infomiho:
- Enables e2e tests
- Enables build test
  - This is **blocked 🚫** until the internal Todo app builds successfully: https://github.com/wasp-lang/wasp/pull/1781
- Moves e2e tests after headless tests
  - **Reasoning:**  usually when we develop, we don't bother with e2e tests being correct because after we are done with a feature, we fix them. e2e tests take a long time to run and they just fail while we still develop. Because of that and because they run first we don't run the "build" or "headless" tests which might be useful to run while developing.

### DIFF
--- a/waspc/examples/todo-typescript/src/MainPage.tsx
+++ b/waspc/examples/todo-typescript/src/MainPage.tsx
@@ -46,7 +46,7 @@ export const MainPage = ({ user }: { user: AuthUser }) => {
       )}
       <NewTaskForm />
       {tasks && <TasksList tasks={tasks} />}
-      <h2>All</h2>
+      <h2>Everything</h2>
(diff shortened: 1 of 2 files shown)


waspc/examples/todo-typescript/src/MainPage.tsx
```diff
@@ -46,7 +46,7 @@ export const MainPage = ({ user }: { user: AuthUser }) => {
       )}
       <NewTaskForm />
       {tasks && <TasksList tasks={tasks} />}
-      <h2>All</h2>
+      <h2>Everything</h2>
```
Martinsos, line 49: Lol

Martinsos (approved):

PR #1782, opened 2024-02-19
Uses Oslo for password hashing

infomiho:
Replaces `secure-password` with Oslo's password utility. Oslo could be useful for us in the future e.g. as a dependency for Arctic. 

### Docs update
One of the issues we wanted to address: users losing access to their account due to the password hashing algorithm changing. 

We wrote an extra piece of docs on how devs might solve this, a password migrating page where they check the password using the old algorithm and then save the password using the new algorithm.


waspc/src/Wasp/Generator/ServerGenerator/AuthG.hs
```diff
@@ -91,6 +91,7 @@ depsRequiredByAuth spec = maybe [] (const authDeps) maybeAuth
     maybeAuth = AS.App.auth $ snd $ getApp spec
     authDeps =
       AS.Dependency.fromList
-        [ ("lucia", "^3.0.0-beta.14"),
-          ("@lucia-auth/adapter-prisma", "^4.0.0-beta.9")
+        [ ("lucia", "^3.0.1"),
```
infomiho, line 94: I've also bumped Lucia from beta to proper version (there were not API changes)

web/docs/introduction/migrate-from-11.md
```diff
@@ -56,7 +56,7 @@ The main differences are:
 
 - The server/client code separation is no longer necessary. You can now organize
   your code however you want, as long as it's inside the `src` directory.
-- All external imports in your Wasp file must have paths starting with `@src` (e.g., `import foo from '@src/bar.js')
+- All external imports in your Wasp file must have paths starting with `@src` (e.g., `import foo from '@src/bar.js`)
```
Martinsos, line 59: Better, but still not completely correct :D. Here it is:

```suggestion
- All external imports in your Wasp file must have paths starting with `@src` (e.g., `import foo from '@src/bar.js'`)
```

web/docs/introduction/migrate-from-11.md
```diff
@@ -438,6 +438,322 @@ The migration functions provided below are written with the typical use cases in
...
+
+<details>
+  <summary>
+    Steps to create a custom page for migrating the password
+  </summary>
+
+  1. You will need to install the `secure-password` and `sodium-native` packages to use the old hashing algorithm:
+
+  ```bash
+  npm install secure-password sodium-native@3.3.0 --save-exact
```
Martinsos, line 455: Maybe we should also provide version of secure-password here. I know it hasn't been updated in 4 years, but in theory it could be updated tomorrow and then we might have problems.

web/docs/introduction/migrate-from-11.md
```diff
@@ -488,6 +804,14 @@ export async function migrateUsernameAuth(prismaClient: PrismaClient) {
 
 ##### Email
 
+:::caution Users will need to reset their password
+
+There is a breaking change between the old and the new auth in the way the password is hashed. This means that users will need to reset their password after the migration, as the old password will no longer work.
+
+It would be best to notify your users about this change and put a notice on your login page to **request a password reset**.
```
Martinsos, line 811: Can they do this on wrong password? So if users input wrong password, they show them the message? ANy way to do that easily?

web/docs/introduction/migrate-from-11.md
```diff
@@ -438,6 +438,322 @@ The migration functions provided below are written with the typical use cases in
 
 ##### Username & Password
 
+:::caution Users will need to migrate their password
+There is a breaking change between the old and the new auth in the way the password is hashed. This means that users will need to migrate their password after the migration, as the old password will no longer work.
+
+Since users can't reset their password by receiving an email when using the Username & Password auth method, you will need to provide a custom page to migrate the password.
```
Martinsos, line 444:
```suggestion
Since users can't reset their password by receiving an email when using the Username & Password auth method. One way to handle this is to inform them about the need to migrate their password (on the login page) and provide a custom page to migrate the password. 
```

Martinsos (approved):
Approved, I left some comments regarding ideas on how we can help them more in figuring out how to do this in the best way, but in general it is looking good and we can merge!

PR #1788, opened 2024-02-21
Minor improvements in migration instructions.

Martinsos:
I was going through migrating the examples/thoughts app, and while doing it, I was fixing migration instructions as I found them unclear/confusing.
I commented on the main changes, to make it easier to review.

### DIFF
--- a/web/docs/migrate-from-0-11-to-0-12.md
+++ b/web/docs/migrate-from-0-11-to-0-12.md
@@ -71,7 +71,7 @@ projects.
 
 In Wasp 0.11.X, authentication was based on the `User` model which the developer needed to set up properly and take care of the auth fields like `email` or `password`.
 
-```wasp title="main.wasp"
+```wasp title="main.wasp" {18-31}
 app myApp {
   wasp: {
     version: "^0.11.0"
@@ -149,7 +149,7 @@ The guide consists of two big steps:
 1. Migrating your Wasp project to the new structure.
 2. Migrating Wasp Auth.
 
-If you get stuck at any point, don't hesitate to ask us on [our Discord server](https://discord.gg/rzdnErX).
+If you get stuck at any point, don't hesitate to ask for help on [our Discord server](https://discord.gg/rzdnErX).
 
 ### Migrating your project to the new structure
 
@@ -157,21 +157,21 @@ You can easily migrate your old Wasp project to the new structure by following a
 series of steps. Assuming you have a project called `foo` inside the
 directory `foo`, you should:
 
-0. Install the latest `0.12.x` version of Wasp.
+0. **Install the latest `0.12.x` version** of Wasp.
   ```bash
   curl -sSL https://get.wasp-lang.dev/installer.sh | sh -s
   ```
-1. Make sure to backup or save your project before starting the procedure (e.g.,
+1. Make sure to **backup or save your project** before starting the procedure (e.g.,
    by committing it to source control or creating a copy).
-2. Position yourself in the terminal in the directory that is a parent of your wasp project directory (so one level above: if you do `ls`, you should see your wasp project dir listed).
-3. Run the migration script (replace `foo` at the end with the name of your Wasp project directory) and follow the instructions:
+2. **Position yourself in the terminal** in the directory that is a parent of your wasp project directory (so one level above: if you do `ls`, you should see your wasp project dir listed).
+3. **Run the migration script** (replace `foo` at the end with the name of your Wasp project directory) and follow the instructions:
   ```
     npx wasp-migrate foo
   ```
 
 <details>
   <summary>
-    In case migration script doesn't work well for you, you can do the same steps manually, as described below:
+    In case the migration script doesn't work well for you, you can do the same steps manually, as described here:
   </summary>
 <div>
 
@@ -302,16 +302,6 @@ directory `foo`, you should:
 That's it! You now have a properly structured Wasp 0.12.0 project in the `foo` directory.
 Your app probably doesn't quite work yet due to the breaking changes in Auth, which we will migrate next.
 
-:::tip
-
-Wasp no longer enforces separation between server-side and client-side source code.
-
-You can organize source files inside the `src` directory however you want! We
-recommend a feature-based organization. Read more about this in [the last section](#next-steps).
-
-:::
-
-
 
 ### Migrating auth
 As shown in [the previous section](#new-auth), Wasp significantly changed how authentication works in version 0.12.0.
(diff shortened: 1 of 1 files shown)


web/docs/migrate-from-0-11-to-0-12.md
```diff
@@ -71,7 +71,7 @@ projects.
 
 In Wasp 0.11.X, authentication was based on the `User` model which the developer needed to set up properly and take care of the auth fields like `email` or `password`.
 
-```wasp title="main.wasp"
+```wasp title="main.wasp" {18-31}
```
Martinsos, line 74: I highlighted the fields that get removed with new auth.

web/docs/migrate-from-0-11-to-0-12.md
```diff
@@ -149,29 +149,29 @@ The guide consists of two big steps:
...
+If you get stuck at any point, don't hesitate to ask for help on [our Discord server](https://discord.gg/rzdnErX).
 
 ### Migrating your project to the new structure
 
 You can easily migrate your old Wasp project to the new structure by following a
 series of steps. Assuming you have a project called `foo` inside the
 directory `foo`, you should:
 
-0. Install the latest `0.12.x` version of Wasp.
+0. **Install the latest `0.12.x` version** of Wasp.
```
Martinsos, line 160: In most of the important steps, I bolded a single crucial sequence of words, to make it easier to quickly read / visually anchor.

web/docs/migrate-from-0-11-to-0-12.md
```diff
@@ -302,16 +302,6 @@ directory `foo`, you should:
 That's it! You now have a properly structured Wasp 0.12.0 project in the `foo` directory.
 Your app probably doesn't quite work yet due to the breaking changes in Auth, which we will migrate next.
 
-:::tip
```
Martinsos, line 305: I removed this tip because we talk about this in Next Steps anyway, and we don't really want them to start changing the structure yet, because they need to yet migrate the Auth and will just make a situation more complex if they do this now, so better not to mention it, both to avoid additional cognitive load and to avoid them possibly doing it now.

web/docs/migrate-from-0-11-to-0-12.md
```diff
@@ -324,119 +314,182 @@ Migrating your existing app to the new auth system is a two-step process:
...
 :::
 
 #### 1. Migrate to the New Auth System
 
-You can follow these steps to migrate to the new auth system:
-
-1. Upgrade Wasp to the latest 0.12.X version (as described [above](https://discord.gg/rzdnErX)).
-
-
-1. Bump the version to `^0.12.0` in `main.wasp` (if you haven't already done this when [migrating the app to the new structure](#migrating-your-project-to-the-new-structure)).
```
Martinsos, line 341: We expect them to read this doc in sequence, so I removed this and just mentioned above that we expect them to have done this already, this will cover the case well enough if they are jumping through the docs.

web/docs/migrate-from-0-11-to-0-12.md
```diff
@@ -324,119 +314,182 @@ Migrating your existing app to the new auth system is a two-step process:
...
+1. Ensure your **local development database is running**.
+1. **Do the schema migration** (create the new auth tables in the database) by running:
    ```bash
    wasp db migrate-dev
    ```
+You should see the new `Auth`, `AuthIdentity` and `Session` tables in your database. You can use the `wasp db studio` command to open the database in a GUI and verify the tables are there. At the moment, they will be empty (no data).
+1. **Do the data migration** (move existing users from the old auth system to the new one by filling the new auth tables in the database with their data):
 
-You should see the new `Auth`, `AuthIdentity` and `Session` tables in your database. You can use the `wasp db studio` command to open the database in a GUI and verify the tables are there.
+   1. **Implement your data migration function(s)** in e.g. `src/migrateToNewAuth.ts`.
```
Martinsos, line 336: Main thing I did is indented data migration functions stuff under `Do the data migration`, before some of it was indented on one level, some was on the top level, this way it is clear when we start and when we are done with the data migration.

web/docs/migrate-from-0-11-to-0-12.md
```diff
@@ -324,119 +314,182 @@ Migrating your existing app to the new auth system is a two-step process:
...
+│   ├── Login.tsx
+│   ├── queries.ts    -- former userQueries.ts
+│   ├── Register.tsx
+│   └── User.tsx
+├── MainPage.tsx
+└── utils.ts
+```
+
+
+## Addendum
```
Martinsos, line 486: This is new -> so what I did is added `Addendum` chapter at the top level, and put `Example Data Migration Functions` down there.
This way, once you are done with the Auth migration, you don't hit the `Example Data Migration Functions` again and then have to scroll through all of them till you find Next Steps, instead you are immediately at the Next Steps.
Also, this makes it clear that this chapter is not part of the normal flow but is an additional information on this page.

web/docs/migrate-from-0-11-to-0-12.md
```diff
@@ -324,119 +314,182 @@ Migrating your existing app to the new auth system is a two-step process:
...
+
     Between these two steps, so after deploying the new code to production and before migrating the production database, your app will not be working completely: new users will be able to sign up, but existing users won't be able to log in, and already logged in users will be logged out. Once you do the second step, migrating the production database, it will all be back to normal.
 
-    You will likely want to keep the time between the two steps as short as you can. Make sure you know exactly what each step means before doing them for real to eliminate any surprises.
+    You will likely want to keep the time between the two steps as short as you can. Make sure you know exactly what each step involves before doing them for real to eliminate any surprises. Especially the second step, which is a bit more complex.
 
     ---
-    
+
     - **First step:** deploy the new code (client and server), either via `wasp deploy` or manually.
```
Martinsos, line 391:
```suggestion
    - **First step:** deploy the new code (client and server), either via `wasp deploy` (i.e. `wasp deploy fly deploy`) or manually.
```

web/docs/migrate-from-0-11-to-0-12.md
infomiho: Might I suggest using `// highlight-next-line` and `// highlight-start` and `// highlight-end` comments. They are more resistant to code changes in the future :) 

https://docusaurus.io/docs/markdown-features/code-blocks#highlighting-with-comments

web/docs/migrate-from-0-11-to-0-12.md
infomiho: I'd advise against using a complex word like `Addendum` and call the section something like `Extras` or `Additional info`

infomiho:
Just a quick Grammarly check says:

- occurences -> occurrences
- commiting -> committing

While you are editing the file you change those as well :)

web/docs/migrate-from-0-11-to-0-12.md
Martinsos: You might indeed :D! Oh yes this is great, I learned about highlight-next-line, but I forgot again about start/end, that is awesome.

web/docs/migrate-from-0-11-to-0-12.md
Martinsos: Sounds good, I changed it to Appendix!

Martinsos:
@infomiho fixed, pls check again.

web/docs/migrate-from-0-11-to-0-12.md
infomiho: Appendix is better than Addendum but it still feels a bit as an advanced vocabulary work to me 😄 but no need to change

infomiho (approved):

Issue #1764, opened 2024-02-14
Using AuthUI + Tailwind results in broken CSS


infomiho:
⚠️  Using Tailwind + out Auth UI results in the default button to be transparent since Tailwind bg color is loaded after but since it's at the same specificity, it overrides our Stitches defined bg color.

![image](https://github.com/wasp-lang/wasp/assets/2223680/643be0dd-32ab-4866-9b5a-c1b6ed53a3f7)

![image](https://github.com/wasp-lang/wasp/assets/2223680/998776a6-6d34-4a57-aa12-803b0c51d2a7)

### Research

It seems that this is a regression that happened with changes related to restructuring. I believe we changed order of some imports which resulted in Tailwind CSS being loaded after our Stitches CSS (Related issue: https://github.com/tailwindlabs/tailwindcss/issues/8670)

### Possible fix

- Change the way we style Auth UI
- Reorder imports of the CSS that's loaded (seems fragile)
- Hotfix it with `!important` on the all overridden properties e.g. `backgroundColor` for the `SubmitButton`

sodic:
I could only reproduce this with `wasp start`. It doesn't happen when I build the app and host static files

cprecioso:
This still happens after #3034, maybe it's a Tailwind thing in general?

PR #1790, opened 2024-02-21
Making Stitches CSS more important than Tailwind

infomiho:
A simple solution to the following issue:  #1764

I'd keep the issue open since I believe there must a more suitable solution than using `!important`, but that issue might not be relevant in the future if we decide to use something else instead of Stitches which would work better with TailwindCSS (maybe we decide to make TailwindCSS a requirement for Auth UI?)


Martinsos (approved):

Martinsos:
You might though want to add a comment in each of those files with `!important` linking to relevant issue on GH?

PR #1793, opened 2024-02-21
Uses `extends` in the user land tsconfig.json

infomiho:
The user's `tsconfig.json` for IDE support now extends a `tsconfig.json` in the SDK which gives us the ability to update it over time while keep their setup cleaner.

### DIFF
--- a/waspc/data/Cli/templates/skeleton/tsconfig.json
+++ b/waspc/data/Cli/templates/skeleton/tsconfig.json
@@ -4,31 +4,5 @@
 // your IDE checks, but none of these options will affect the TypeScript
 // compiler. Proper TS compiler configuration in Wasp is coming soon :)
 {
-  "compilerOptions": {
-    // JSX support
-    "jsx": "preserve",
-    "strict": true,
-    // Allow default imports.
-    "esModuleInterop": true,
-    "lib": [
-      "dom",
-      "dom.iterable",
-      "esnext"
-    ],
-    "allowJs": true,
-    "types": [
-      // This is needed to properly support Vitest testing with jest-dom matchers.
-      // Types for jest-dom are not recognized automatically and Typescript complains
-      // about missing types e.g. when using `toBeInTheDocument` and other matchers.
-      "@testing-library/jest-dom"
-    ],
-    // Since this TS config is used only for IDE support and not for
-    // compilation, the following directory doesn't exist. We need to specify
-    // it to prevent this error:
-    // https://stackoverflow.com/questions/42609768/typescript-error-cannot-write-file-because-it-would-overwrite-input-file
-    "outDir": "phantom",
-  },
-  "exclude": [
-    "phantom"
-  ],
+  "extends": "wasp/dev/tsconfig.json",
 }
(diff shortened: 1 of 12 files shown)


waspc/data/Cli/templates/skeleton/tsconfig.json
```diff
@@ -4,31 +4,5 @@
...
-    // Since this TS config is used only for IDE support and not for
-    // compilation, the following directory doesn't exist. We need to specify
-    // it to prevent this error:
-    // https://stackoverflow.com/questions/42609768/typescript-error-cannot-write-file-because-it-would-overwrite-input-file
-    "outDir": "phantom",
-  },
-  "exclude": [
-    "phantom"
-  ],
+  "extends": "wasp/dev/tsconfig.json",
```
sodic, line 7: Are we sure we want to extend _everything_?

I have a feeling most frameworks leave some of the things explicit on purpose (e.g., `noEmit`). Can you take a look at how next does it (perhaps they don't even use extend, I'm just guessing)?

My intuition says that we should only hide the weird stuff that they aren't allowed to change.

infomiho: > I have a feeling most frameworks leave some of the things explicit on purpose (e.g., noEmit). Can you take a look at how next does it (perhaps they don't even use extend, I'm just guessing)?

Unlike the other frameworks, we aren't actually using this `tsconfig.json` for anything (other then the IDE support). Showing them options might even prompt them to tweak them and expect to be used when building their app.

> My intuition says that we should only hide the weird stuff that they aren't allowed to change.

The whole config is a weird thing 😂 Yeah, I'm not sure. I don't see the value in keeping some options visible and some "hidden" even though then can just CTRL/CMD click the `wasp/dev/tsconfig.json` and see everything.

sodic: I think Next is also using it for the IDE support only.

Keep in mind, the plan (for the future) is to use their config when building  their code (so the options will definitely come into play).

> Showing them options might even prompt them to tweak them and expect to be used when building their app.

They might still want to configure how their IDE handles certain cases.

IDEs are doing all the TS heavy lifting these days anyway, and most user probably verify whether their tsconfig is working only by looking at what the IDE says (not manually compiling the code).

In other words, I think the expectation for what the tsconfig.json does is shifting towards "Primarily configures your IDE <sub><sup>and does the same thing later during the build I guess</sup></sub>."

> I don't see the value in keeping some options visible and some "hidden" even though then can just CTRL/CMD click the wasp/dev/tsconfig.json and see everything.

I get what you mean, but most tools I've used either don't extend anything (and they could, like Next and Vite) or combine the `extends` field with explicit field definitions (and I sort of have a feeling the division follows the rule I presented above).

FWIW, I don't think I've ever seen a `tsconfig.json` containing only a single `extends` field.

All that said, I don't have time to research this right now. I'll trust you'll make the right call on what to do.

sodic: I don't have particular examples in mind - something just feels off but I can't put my finger on it.

That's why I'm OK with whatever you decide is best.

infomiho: Okay, we can change this in the future if needed 👍

PR #1796, opened 2024-02-21
Replaces ServerImport with ExtImport

infomiho:


sodic (approved):
@Martinsos @infomiho Should we maybe use this chance and start calling these things `SrcImport`s?

In any case, we can merge this (I'm guessing you guys already used `ExtImports` in other parts of the docs) and I can then do a search-and-replace across the codebase: ExtImport -> SrcImport.

Martinsos:
@sodic @infomiho the problem here is that we just lost information. For some of these imports, it is not very clear any more if it is nodejs code or browser code.
ClientImport and ServerImport made it clear -> but not we don't have that info anymore, it is just SrcImport.
Which is why I would go through each of these replacements you made and if not clear, add additional piece of text making it clear if we are importing nodejs code or what.

PR #1797, opened 2024-02-22
Uses `noEmit` in tsconfig.json

infomiho:
- We experienced a hard to reproduce behaviour, sometimes there is a `phantom` dir created in the root project dir.

- Since we use the `tsconfig.json` for IDE support only, we had a hack with an `outDir` and `exclude` to stop Typescript from complaining about overwriting source files. 

- A similar option exists called `noEmit` which does what we need: saying to the Typescript compiler, we don't want to output any files.


infomiho:
No longer valid PR since the latest changes to the `tsconfig.json`

PR #1798, opened 2024-02-22
😃🚀📝

Martinsos:


Martinsos:
One quick example I encountered, actually the same as the usage I tried to do in the migration docs:
- https://tanstack.com/query/latest/docs/framework/react/guides/migrating-to-v5 -> :tada: , :rocket: .

PR #1802, opened 2024-02-22
Fix defineUserSignupFields example in migration instructions

Martinsos:
As discussed here https://discord.com/channels/686873244791210014/1205479316226445372/1210232860779479152 .

### DIFF
--- a/web/docs/migrate-from-0-11-to-0-12.md
+++ b/web/docs/migrate-from-0-11-to-0-12.md
@@ -512,7 +512,9 @@ You can follow these steps to migrate to the new auth system (assuming you alrea
     // highlight-end
     ```
 
-    Read more about the `userSignupFields` function [here](/auth/overview.md#1-defining-extra-fields).
+    If you want to properly type the `profile` object, we recommend you use a validation library like Zod to define the shape of the `profile` object.
+    
+    Read more about this and the `defineUserSignupFields` function in the [Auth Overview - Defining Extra Fields](./auth/overview.md#1-defining-extra-fields) section.
 
     </TabItem>
     </Tabs>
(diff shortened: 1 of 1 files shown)


web/docs/migrate-from-0-11-to-0-12.md
```diff
@@ -504,10 +504,9 @@ You can follow these steps to migrate to the new auth system (assuming you alrea
     // highlight-start
     export const fields = defineUserSignupFields({
       displayName: async (data) => {
-        if (!data.profile || !data.profile.displayName) {
-          throw new Error('Display name is not available')
-        }
-        return data.profile.displayName
```
Martinsos, line 510: This was throwing TS error, so I refactored it to have casting to `any`.

web/docs/migrate-from-0-11-to-0-12.md
```diff
@@ -418,14 +418,14 @@ You can follow these steps to migrate to the new auth system (assuming you alrea
     // highlight-next-line
     export const fields = defineUserSignupFields({
       address: async (data) => {
-        const address = data.address
+        const address = data.address;
```
infomiho, line 421: I'd maybe not include the semicolons if we didn't include it in other examples on this page.

web/docs/migrate-from-0-11-to-0-12.md
infomiho: I have to admit, I wrote this out of my head and didn't test it. Good catch 👍

infomiho (approved):

web/docs/migrate-from-0-11-to-0-12.md
infomiho: Should we maybe mentioned Zod here ... or more ideally in the `userSignupFields` docs?

Martinsos:
@infomiho feel free to maybe take this one over? YOu will do it better, I just did this quick PR so we have something and we don't forget.

PR #1803, opened 2024-02-22
Revert "Uses `extends` in the user land tsconfig.json"

sodic:
Reverts wasp-lang/wasp#1793


infomiho (approved):
I was fun while it lasted 😢

PR #1805, opened 2024-02-22
Mage 0.12 producing 0.11

Martinsos:
Next step will be Mage 0.12 producing 0.12 -> but that is tiny change, and I will do it once we release wasp 0.12.0.
NOTE: This version of the code, from this branch, is currently released on usemage.ai .

This PR in itself doesn't contain something super interesting, mostly automatic migrations, updated custom Dockerfile, and some small manual interventions.


mage/src/client/pages/MainPage.jsx
```diff
@@ -1,5 +1,6 @@
+import { useAuth, GitHubSignInButton } from "wasp/client/auth";
+import { startGeneratingNewApp, useQuery, getProjectsByUser } from "wasp/client/operations";
```
infomiho, line 2: Let's reorder imports (put React on top for example)

mage/src/client/pages/ResultPage.jsx
```diff
@@ -1,9 +1,13 @@
+import {
+  startGeneratingNewApp,
+  registerZipDownload,
+  createFeedback,
+  useQuery,
+  getAppGenerationResult,
+  getNumProjects,
+} from "wasp/client/operations";
+
```
infomiho, line 9: Needles space

infomiho (approved):
Seems good! I'd only advise import reordering to make it nicer for reading later on

PR #1806, opened 2024-02-23
Migrating Tailwind setup

infomiho:


web/docs/migrate-from-0-11-to-0-12.md
```diff
@@ -301,13 +301,14 @@ directory `foo`, you should:
     `foo/.gitignore`
 13. Copy the rest of the top-level files and folders (all of them except for `.gitignore`, `main.wasp` and `src/`)
     in `foo_old/` into `foo/` (overwrite the existing files in `foo`).
-14. Run `wasp clean`.
+14. Run `wasp clean` in `foo`.
+15. Delete the `foo_old` directory.
```
Martinsos, line 305: Oh did I drop this while adding info about migration script? Good catch!

web/docs/migrate-from-0-11-to-0-12.md
```diff
@@ -624,6 +625,57 @@ Your app should be working correctly and using new auth, but to finish the migra
 
 :::
 
+### Migrating the Tailwind Setup
```
Martinsos, line 628: Looking good! The only thing is that for me, I hit this error once I finally ran `wasp start`. That happened at line 563: `Verify that the basic auth functionality works ...`, when it tells you to run `wasp start` to test that. I got a blank white screen in the browser, and in the terminal it wrote to me about the issue with tailwind config.

So if this section about Tailwind is this low, after all the auth migration stuff, then I won't see it on time. So, what is better place to put it? I think good place could be to put it the "restructuring" part, at the very end of it. That is anyway the moment when we are updating the code to work with new structure (and this change in tailwind is the consequence of new structure).

infomiho: What about we move the Tailwind migration right below the Migrating to new structure?

Martinsos: Yeah I get it, if you want to put it right after that, that is also fine, I think it is just important we put it before starting with the "Migrate to new auth section".

We could even possibly make it automated hm :D, it is a bit tempting :D, although it is probably not worth the effort, ok true, we will end up with regexes that are not 100% sure to work, better for them to do it manually. Ah temptation.

Martinsos (approved):
Ok, approved!

PR #1807, opened 2024-02-23
Adds instructions for custom Dockerfile migration

infomiho:


web/docs/migrate-from-0-11-to-0-12.md
```diff
@@ -676,6 +676,18 @@ module.exports = {
...
+:::note
+If you didn't customize your Dockerfile, you can skip this section.
+:::
+
+Between Wasp 0.11.X and 0.12.X, the Dockerfile that Wasp generates for you has changed. If you defined a custom Dockerfile in your project root dir, you'll need to update it to incorporate the changes that Wasp 0.12.X made.
+
+We suggest that you temporarily move your custom Dockerfile to a different location, then run `wasp start` to generate the new Dockerfile.
+
+Check out the `.wasp/out/Dockerfile` to see the new Dockerfile and what changes you need to make. You'll probably need to copy some of the changes from the new Dockerfile to your custom one to make your app work with Wasp 0.12.X.
+
```
Martinsos, line 690: I just realized, these changes affect them even if they don't use custom Dockerfile, but have maybe looked at the command in our Dockerfile and did some custom build thing on their own. We had a person doing this -> they copied the commands from our Dockerfile to their bare metal machine, and were building the server that way.
I tried doing minimal adjustments to make the instructions a bit more general. I think they are still quite specific for Dockerfile, but are general enough to also get the needed attention for anybody who went even more custom.

```suggestion
### Default Server Dockerfile Changed

:::note
If you didn't customize your Dockerfile or had custom build process for the Wasp server, you can skip this section.
:::

Between Wasp 0.11.X and 0.12.X, the Dockerfile that Wasp generates for you for deploying the server has changed. If you defined a custom Dockerfile in your project root dir (or in any other way relayed on its contents), you'll need to update it to incorporate the changes that Wasp 0.12.X made.

We suggest that you temporarily move your custom Dockerfile to a different location, then run `wasp start` to generate the new Dockerfile.

Check out the `.wasp/out/Dockerfile` to see the new Dockerfile and what changes you need to make. You'll probably need to copy some of the changes from the new Dockerfile to your custom one to make your app work with Wasp 0.12.X.

```

Is it in `.wasp/build/Dockerfile`, or one level deeper, in `server/`? If the latter, I wrote it incorrectly.

Martinsos:
@infomiho I forgot what happened with this one, why did we close it?

infomiho:
Rebasing issues, I made a new PR right after this one.

PR #1813, opened 2024-02-26
Improved stdout of data migration functions.

Martinsos:
When I was doing data migration for Mage, I realized I am missing info on how the migration went. I would get a lot of output to stdout, but no summary that would tell me if all the users were migrated, how many were migrated, ... . Something that would give me an idea if things went right, or if something is suspicious (e.g. all users skipped, no migrated).

Another thing which I noticed is that all user info is being printed, including hashed passwords, which I didn't like, on one hand it was cluttering my stdout, on the other hand it felt even a bit insecure potentially.

So what I did here is:
1. Added counters for skipped and migrated users which at the end get printed.
2. Now when user is skipped, we print only their id, not the whole user data.

Oh and I also added a short note on functions being idempotent.

TODO: I haven't yet tested this properly.

### DIFF
--- a/web/docs/migrate-from-0-11-to-0-12.md
+++ b/web/docs/migrate-from-0-11-to-0-12.md
@@ -751,6 +751,8 @@ src
 
 The migration functions provided below are written with the typical use cases in mind and you can use them as-is. If your setup requires additional logic, you can use them as a good starting point and modify them to your needs.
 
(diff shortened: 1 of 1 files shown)


Martinsos:
Done via https://github.com/wasp-lang/wasp/pull/1812 instead.

PR #1814, opened 2024-02-26
Move phantom dir to .wasp

sodic:
During the 0.12.0 release, we noticed that Wasp sometimes creates the `phantom` directory in the project root. We aren't yet sure how and why this happens

This is a fake directory we set as the tsconfig's `outDir` option back when introducing TS support to Wasp (I guess we didn't know about the  `noEmit` option).

The long-term plan is still:
- Figuring out why `phantom` appears.
- Using `noEmit` instead of `phantom`.

Until then, I've moved `phantom` to the hidden `.wasp` dir.


Martinsos (approved):
Sounds good!

Suggestions:
1. Make sure we have a GH issue to handle this later properly.
2. Test this after `wasp clean`, when there is no `.wasp` dir.

PR #1818, opened 2024-02-26
Enforce casing in operations and entities

sodic:
Here's how it looks:
image

### DIFF
--- a/waspc/ChangeLog.md
+++ b/waspc/ChangeLog.md
@@ -53,10 +53,11 @@ The output of `wasp new myProject` with 0.12.0:
 ```
 
 The main differences are:
- - Your project now has a `package.json` file.
- - `tsconfig.json`, `vite.config.ts`, and `public/` moved to the top dir.
- - The server/client code separation is no longer necessary. You can now organize your code however you want, as long as it's inside the `src/` directory.
- - All external imports in your Wasp file now must have paths starting with `@src` (e.g., `import foo from '@src/MainPage.jsx'`). The paths can no longer start with `@server` or `@client`.
+
+- Your project now has a `package.json` file.
+- `tsconfig.json`, `vite.config.ts`, and `public/` moved to the top dir.
+- The server/client code separation is no longer necessary. You can now organize your code however you want, as long as it's inside the `src/` directory.
+- All external imports in your Wasp file now must have paths starting with `@src` (e.g., `import foo from '@src/MainPage.jsx'`). The paths can no longer start with `@server` or `@client`.
 
 #### New Auth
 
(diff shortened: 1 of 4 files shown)


infomiho (approved):
LGTM, unit tests maybe be useful to add as well 👍

sodic:
> LGTM, unit tests maybe be useful to add as well 👍

https://github.com/wasp-lang/wasp/issues/1819 :upside_down_face:

waspc/ChangeLog.md
```diff
@@ -68,31 +69,82 @@ The `User` model is now just a business logic model and you use it for storing t
...
+
+action UpdateFoo {
+  // ...
+}
+
+job PrintFoos {
+  // ...
+}
+
+```
```
Martinsos, line 117:
```suggestion
Operation (i.e., Queries and Actions) and Job names in `.wasp` files must now begin with a lowercase letter: `query getTasks {...}`, `job sendReport {...}`.

Entity names in `.wasp` files must now begin with an uppercase letter: `entity Foo {...}`.
```

web/docs/migrate-from-0-11-to-0-12.md
```diff
@@ -367,7 +367,7 @@ module.exports = {
 </TabItem>
 </Tabs>
 
-### Default Server Dockerfile Changed
+### Migrating a custom Dockerfile
```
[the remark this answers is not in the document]
Martinsos: @sodic there was a reason why it was named that way -> because you might not need to migrate a custom Dockerfile, but update your deployment mechanism.
What really changed is not default dockerfile, but how server is being built / prepared for deployment. We had some people asking how to get it working on bare metal, and what they did was copy the instructions from the Dockerfile. So for them, they don't need to migate a custom dockerfile, they need toupdate their command on their server or wherever.

We went with "Default Server Dockerfile Changed" as a compromise between correctnes and recognizability. If you based your custom instructions based on the default dockerfile, which is really the only source of truth on how server should be deployed, then this should pick your interest. Migrating a custom Dockerfile on the other hand sounds like I don't have to take a look at it if I don't have a custom dockerfile, which is not correct.

I don't mind "sneaking" in changes like this in the PR that has nothing to do with it, although in theory they should be a separate PR -> it can be too much bearocracy, I get that. But if doing so, it could be useful to mark it with a comment for reviewers, so we don't miss it, because something like this is super easy to miss.

Also, Chesterton's fence would be a good rule that could have helped here.

web/docs/migrate-from-0-11-to-0-12.md
```diff
@@ -377,6 +377,63 @@ Between Wasp 0.11.X and 0.12.X, the Dockerfile that Wasp generates for you for d
 We suggest that you temporarily move your custom Dockerfile to a different location, then run `wasp start` to generate the new Dockerfile.
 Check out the `.wasp/out/Dockerfile` to see the new Dockerfile and what changes you need to make. You'll probably need to copy some of the changes from the new Dockerfile to your custom one to make your app work with Wasp 0.12.X.
 
+### Migrating declaration names
```
Martinsos, line 380: Probably would be better if this went above the Dockerfile section above, since Dockerfile section tells user to use `wasp start` but `wasp start` will fail if names of declarations are incorrectly capitalized, so you really want to take care of that first.

waspc/ChangeLog.md
Martinsos, line 1: What happened here with all the small changes? Is that some kind of auto-formatting?

web/docs/migrate-from-0-11-to-0-12.md
Martinsos: Ok, I moved it above dockerfile and also tailwind.
I also shortened the examples, it was redundant. And added mention of `main.wasp` to put it into context.

Martinsos (approved):
@sodic Nice job!

I reverted your renaming of the 'Migrating Dockerfile' section in the migration instructions, check the details in the comment.

I also moved the declaration migration instructions above 'Migrationg Dockerfile' instructions because you can't run `wasp start` in the dockerfile section if the names are not fixed. Check the comment for that for more details.

I also shortened examples as they were redundantly long, and mentioned all this is happening in `main.wasp` file.

PR #1825, opened 2024-02-27
Migrating Waspello to 0.12

infomiho:


sodic (approved):
LGTM.

What do we want to do about the folder structure? Leave it as is or migrate it to feature-based?

This goes for all examlpes.

PR #1826, opened 2024-02-27
Updated Wasp AI to work with newest GPT 3.5 Turbo (0125).

Martinsos:
We had issues with 0125 dropping newlines from files and returning as one big line, and also with putting all the declarations in main.wasp under `app {...}`.

I did some prompt engineering here :D to try to combat that, and it seems it works. I didn't have enough time to run it many times, but I did manage to get it to produce working apps without any of those mistakes couple of times in a row.

We should do some more testing, to confirm it indeed works and doesn't make those mistakes, and then we can merge.


sodic (approved):
I have no idea how to review these things :man_shrugging: 

LGTM I guess :laughing:

Martinsos:
Tested more, it consistently works well now, no issues with newlines or with messed up main.wasp file. Also, the quality of apps seems to be similar as before.

PR #1834, opened 2024-02-29
Conditionally apply Web Crypto API polyfill

### LINKED ISSUE #1833: Web Crypto API polyfill breaks Wasp's server when using Node.js 20
> [!WARNING]
> Wasp doesn't work correctly with Node.js 20 because we try to load a polyfill which is only needed for Node.js 18 and this breaks Wasp's server when using Node.js 20

This was reported on Discord when one of our users tried to use Wasp 0.12 with Node.js 20:

```
File: /app/.wasp/out/server/src/polyfill.ts:7
 globalThis.crypto = webcrypto as Crypto;
 TypeError: Cannot set property crypto of #<Object> which has only a getter
     at <anonymous> (/Users/Dealwap/Desktop/resume-sass/app/.wasp/out/server/src/polyfill.ts:7:21)
```
https://ptb.discord.com/channels/686873244791210014/1212533517360177253/1212803377918902382

## Explanation 

We used Node.js 18 before and we ALWAYS needed the polyfill the Web Crypto API as per Lucia's docs recommendation: https://lucia-auth.com/getting-started/

### What can we do to fix it

Since the polyfill just defines a property on `globalThis` and the error message is about Node.js 20 complaining that we are setting a property which is only defined as a getter (which means _the property exists_) we could fix this by applying the polyfill _only_ if the `globalThis.crypto` property doesn't exist.

### Alternative solutions

We could detect the user's Node.js version and either import the polyfill or not?

## Future work

We need to introduce testing with Node.js 20 (or any supported versions) in our CI. https://github.com/wasp-lang/wasp/issues/1835

infomiho:
Closes #1833

### DIFF
--- a/waspc/data/Generator/templates/server/src/polyfill.ts
+++ b/waspc/data/Generator/templates/server/src/polyfill.ts
@@ -3,5 +3,8 @@
 
 import { webcrypto } from "node:crypto";
 
-// @ts-ignore
-globalThis.crypto = webcrypto as Crypto;
+if (typeof globalThis.crypto === "undefined") {
+  // @ts-ignore
+  globalThis.crypto = webcrypto as Crypto;
+}
+


waspc/data/Generator/templates/server/src/polyfill.ts
```diff
@@ -3,5 +3,8 @@
 
 import { webcrypto } from "node:crypto";
 
-// @ts-ignore
-globalThis.crypto = webcrypto as Crypto;
+if (typeof globalThis.crypto === "undefined") {
```
infomiho, line 6: The error message we see in `Node.js 20` is due to the `crypto` property already exists. This check ensures we don't try to set it if it exists.

infomiho: Something else I saw in the wild, check the Node.js version: https://github.com/mitschabaude/montgomery/blob/1b7c1d3f1adf7bdf39812fd23a5bdd5f95496171/src/extra/fix-webcrypto.ts#L3

Zeko369, line 6: Maybe add a comment about this that this only targets Node20+

Zeko369 (approved):

waspc/data/Generator/templates/server/src/polyfill.ts
Martinsos: Checking node version is more indirect, because there is additional assumption that certain versions of don't have or don't have this functionality, while he could have just checked directly. Unless there is some other reasoning here, I prefer our solution.

Martinsos: Yeah, might be useful to have that comment, mostly as an explanation of why this is even needed.

sodic (commented):
Add a comment like @Zeko369's suggested and link to an URL where the issue is described (perhaps the one you sent to Discord), and we're good to go!

waspc/data/Generator/templates/server/src/polyfill.ts
Martinsos, line 6:
```suggestion
// NOTE: node < 19 doesn't have Crypto API, which we need for Lucia, so we apply the polyfill if Crypto API is not defined.
if (typeof globalThis.crypto === "undefined") {
```

Martinsos:
Since Miho is on the move/vacation I added the comment and merged into main, I will go make a release now.

Issue #1835, opened 2024-02-29
Improve our CI to include testing with supported Node.js


infomiho:
Before Wasp 0.12 we only supported Node.js 18 and we based our tests in the CI on using the Node.js 18.

From Wasp 0.12 we support Node.js >= 18 and our tests should check if our code works for at least one other version > 18 to ensure things like this: https://github.com/wasp-lang/wasp/issues/1833 don't slip by.

infomiho:
Completed by #1836

PR #1836, opened 2024-02-29
Adding Node.js 18 and 20 CI tests

infomiho:
This is a test PR for now for https://github.com/wasp-lang/wasp/issues/1835

### DIFF
--- a/.github/workflows/waspc-ci.yaml
+++ b/.github/workflows/waspc-ci.yaml
@@ -39,10 +39,6 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        ghc:
-          - '8.10.7'
-        cabal:
-          - '3.6.2.0'
         os:
           # Using Ubuntu 20.04 due to recent GLIBC_2.34 version mismatch errors for
           # users and other services. The strategy here is to pick a slightly older LTS
@@ -54,6 +50,25 @@ jobs:
           - ubuntu-20.04
           - macos-latest
           - windows-latest
+        node-version:
+          - 'latest'
+        ghc:
+          - '8.10.7'
+        cabal:
+          - '3.6.2.0'
+        # In addition to the default matrix, we also want to run the build job for
+        # additional Node.js versions, to make sure that Wasp works with them.
+        # To reduce the number of jobs, we only test the Node.js versions on
+        # Ubuntu 20.04.
+        include:
+          - os: ubuntu-20.04
+            node-version: 18
+            ghc: '8.10.7'
+            cabal: '3.6.2.0'
+          - os: ubuntu-20.04
+            node-version: 20
+            ghc: '8.10.7'
+            cabal: '3.6.2.0'
 
     steps:
       - name: Configure git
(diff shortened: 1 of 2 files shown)


.github/workflows/waspc-ci.yaml
```diff
@@ -90,7 +93,7 @@ jobs:
       - name: Set up Node
         uses: actions/setup-node@v3
         with:
-          node-version: '18'
+          node-version: ${{ matrix.node-version }}
 
       - name: Cache
         uses: actions/cache@v3
```
Martinsos, line 99: One thing to also consider is cache -> if we were caching any node stuff, like node_modules or something, we would also want to add node version to cache key, but from what I see we are caching only cabal stuff, so no need to add anything, all good!

waspc/examples/todo-typescript/src/ChatPage.tsx
```diff
@@ -45,7 +45,7 @@ export const ChatPage = () => {
 
   return (
     <>
-      <h2>Chat</h2>
+      <h2 className="text-2xl font-bold mb-4">Chat</h2>
```
Martinsos, line 48: What happened here :D?

Martinsos (commented):
This is looking quite good!

What I would consider is, to reduce a bit a number of jobs we are running, is to test different types of node only on ubuntu. I don't think there is any need to test those also on mac and windows.
Because, so far we had 3 jobs. With this, we now have 3*2 = 6. Add one more node version, and that is 9. If we limit it to just Ubuntu, we can then have 3 node versions we are checking (18, 20, latest) and have 3 + 2 = 5 jobs, which is feasible.

waspc/examples/todo-typescript/src/ChatPage.tsx
infomiho: We need to have some change in `waspc` folder to trigger the CI

infomiho:
> What I would consider is, to reduce a bit a number of jobs we are running, is to test different types of node only on ubuntu. I don't think there is any need to test those also on mac and windows.

I'm not sure if that's possible with the way Github Actions work, I'll have to research what we can do. Now we are using their `matrix` option that does cross-product of all options e.g. `OS` x `Node Version` x `Haskell Version`

.github/workflows/waspc-ci.yaml
```diff
@@ -54,6 +54,21 @@ jobs:
           - ubuntu-20.04
           - macos-latest
           - windows-latest
+        node-version:
+          - 'latest'
```
infomiho, line 58: @Martinsos Should we run our "usual" tests on Node.js 18 or latest Node.js?

We now have the following matrix:
```
(Win, latest)
(MacOS, latest)
(Ubuntu, latest)
(Ubuntu, 18)
(Ubuntu, 20)
```

Martinsos: What do you mean by usual tests?

I would say this matrix is good! It covers everything.

Aha I see what you mean: should we have MacOS + latest, or MacOS + 18?

Ha good question. I don't think it is super important. I would probably go with MacOS + 18/lowest, because if it works for lowest, it should also work for the higher versions (although we have seen that is not always true, with the polifyll thing we had).
But combo of Ubuntu + node versions should really cover it all, which is why I think this choice is not so crucial.

infomiho: Okay, let's keep it like this then :)

Martinsos (approved):
@infomiho this is looking good, approving! There is still left to figure out what is the "default" node version for our jobs, so check that comment, but whatever you do, I am ok with it.

PR #1851, opened 2024-03-04
Use Arctic for OAuth providers

### LINKED ISSUE #1817: Auth UI and OAuth: show auth errors on screen
> [!NOTE]
> This will be closed by #1851 

We had the case (in wasp 0.12) where:
1. `User` has `email` field with `unique` constraint. It is a field we set with email via `userSignupFields` on successful signup.
2. I sign up with Google, with email address E.
3. I try to sign up with Github, with same email address E, but it just returns me to `login`.
  I get following errors in the console: 
image

Instead, I would expect that some kind of nice error is shown on the screen. I am guessing error gets thrown in `userSignupFields` -> well, actually as a result of executing the setters/gettters defined there -> but it doesn't get displayed by Auth UI for some reason.

infomiho:
### Status

**Needs to be implemented**
- [x] We need the `serverUrl` in our server environment (⚠️ breaking change)
   https://github.com/wasp-lang/wasp/pull/1856
- [x] We need to implement support for providing `configFn` again
- [x] We need to implement env vars validation again (e.g. `GOOGLE_CLIENT_ID`)
- [x] Fetch full user info for Github like the Passport libs did (mostly to get the `email` which is important to Wasp users)
- [x] Improve error handling in case of an error with OAuth providers
    - Redirect to the client with the error message instead of showing JSON.
- [x] Fix the type error in [OAuthCallback.tsx](https://github.com/wasp-lang/wasp/pull/1851/files#diff-5815b3745fac3ac14fb42c64efd4c65c702d5309986bab51a9aaf43e409da4f0)
- [x] Write migration docs https://github.com/wasp-lang/wasp/pull/1890
- [x] Redirect with **Prisma errors in the query params** raises an error due to Unicode chars in the query
- [ ] Write a migration script (bonus)
- [ ] Write a specific error handling if users used "scope" in their `configFn` before to say "Hey, we changed this, you have to use "scopes" now" to make the migration easier (bonus)

**Improvements**

- The client page now receives a one-time-code and exchanges it for a session
  - [x] Ideally, we don't redirect to "on login failed" path, but show a message #1817 
- [x] Move some of the paths (login path, oauth callback paths) in Haskell
- [x] `createToken` and `verifyToken` functions need to be refactored (tech debt from old auth)
    https://github.com/wasp-lang/wasp/pull/1852
- [x] DRY up the code for `google.ts` and `github.ts` (to reduce the effort of adding next providers)

**Changes for user**
- They need to change the redirect URL from client to server since we now do the OAuth redirect from backend to Google to backend (⚠️ breaking change)

**Open questions**
- The property in the user config (when users use `configFn` property in Wasp file) to set scopes changed from `scope` to `scopes` (⚠️ breaking change) -> we can maybe make it backwards compatible? 
- Passport strategies that we used mapped the [Google](https://github.com/jaredhanson/passport-google-oauth2/blob/master/lib/profile/openid.js#L20) data into a specific shape  -> do we want to do it as well to be backwards compatible? (Same things goes for [Github](https://github.com/cfsghost/passport-github/blob/master/lib/profile.js#L8))

Closes #1817


waspc/data/Generator/templates/server/src/auth/providers/config/github.ts
```diff
@@ -0,0 +1,149 @@
...
+                return res.status(400).json({
+                    success: false,
+                    message: "Invalid state",
+                });
+            }
+    
+            try {
+                const { accessToken } = await github.validateAuthorizationCode(code);
+    
+                // TODO: maybe an additional request to get the user's email?
```
infomiho, line 94: The passport strategy did it by default if the scopes included `user` or `user:email` https://github.com/cfsghost/passport-github/blob/master/lib/strategy.js#L119

waspc/data/Generator/templates/react-app/src/router.tsx
```diff
@@ -42,8 +42,8 @@ const router = (
       {=# isExternalAuthEnabled =}
       {=# externalAuthProviders =}
       {=# authProviderEnabled =}
-      <Route exact path="{= authFrontendUrl =}">
-        <OAuthCodeExchange pathToApiServerRouteHandlingOauthRedirect="{= authServerOauthRedirectUrl =}" />
+      <Route exact path="/oauth/callback">
```
infomiho, line 45: Hard coded path

waspc/data/Generator/templates/sdk/wasp/auth/providers/types.ts
```diff
@@ -16,19 +16,11 @@ export type ProviderConfig = {
     // Unique provider identifier, used as part of URL paths
     id: ProviderName;
     displayName: string;
-    // Each provider config can have an init method which is ran on setup time
-    // e.g. for oAuth providers this is the time when the Passport strategy is registered.
-    init?(provider: ProviderConfig): Promise<InitData>;
```
infomiho, line 21: Not needed anymore

waspc/data/Generator/templates/server/src/auth/providers/config/github.ts
```diff
@@ -0,0 +1,129 @@
...
+            ["GITHUB_CLIENT_ID", "GITHUB_CLIENT_SECRET"],
+            provider
+        );
+
+        const github = new GitHub(
+            env.GITHUB_CLIENT_ID,
+            env.GITHUB_CLIENT_SECRET,
+        );
+
+        router.get('/login', async (_req, res) => {
```
infomiho, line 44: Hard coded path

waspc/data/Generator/templates/server/src/auth/providers/oauth/user.ts
```diff
@@ -0,0 +1,79 @@
...
+  const oneTimeCode = await tokenStore.createToken(authId);
+
+  return {
+      redirectUri: getRedirectUriForOneTimeCode(oneTimeCode),
+  };
+}
+
+// We need a user id to create the auth token, so we either find an existing user
+// or create a new one if none exists for this provider.
+async function getAuthIdFromProviderDetails(
```
infomiho, line 40: Existing function

waspc/examples/todoApp/src/auth/github.js
```diff
@@ -1,9 +1,7 @@
 export function config() {
   console.log('Inside user-supplied GitHub config')
   return {
-    clientID: process.env['GITHUB_CLIENT_ID'],
```
infomiho, line 4: No need for client ID or client secret in the user config any more.

waspc/src/Wasp/Generator/ServerGenerator/Auth/OAuthAuthG.hs
```diff
@@ -119,28 +118,8 @@ genOAuthConfig provider maybeUserConfig pathToConfigDst = return $ C.mkTmplFdWit
     relPathFromAuthConfigToServerSrcDir :: Path Posix (Rel importLocation) (Dir C.ServerSrcDir)
     relPathFromAuthConfigToServerSrcDir = [reldirP|../../../|]
 
-getJsonForOAuthConfigProps :: OAuthAuthProvider -> [Aeson.Value]
```
infomiho, line 122: Moved the env variables knowledge in the template

waspc/src/Wasp/Generator/ServerGenerator/Auth/OAuthAuthG.hs
```diff
@@ -98,18 +99,16 @@ genOAuthProvider provider maybeUserConfig
 genOAuthConfig ::
   OAuthAuthProvider ->
   Maybe AS.Auth.ExternalAuthConfig ->
-  Path' (Rel ServerSrcDir) File' ->
+  Path' (Rel ServerTemplatesSrcDir) File' ->
   Generator FileDraft
-genOAuthConfig provider maybeUserConfig pathToConfigDst = return $ C.mkTmplFdWithDstAndData tmplFile dstFile (Just tmplData)
+genOAuthConfig provider maybeUserConfig pathToConfigTmpl = return $ C.mkTmplFdWithData tmplFile (Just tmplData)
```
infomiho, line 104: We used `_oauth.ts` as a common template before, but now each provider has their own template file because their code now differs (the way we fetch the user info, some providers require extra steps, some not)

waspc/data/Generator/templates/react-app/src/auth/pages/OAuthCallback.tsx
```diff
@@ -0,0 +1,72 @@
...
+      isFirstRender.current = false;
+      handleCallback();
+    }
+  }, []);
+  
+  if (user) {
+    return <Redirect to="{= onAuthSucceededRedirectTo =}" />;
+  }
+
+  /* TODO: Decide if we want to redirect to a page that shows the error message */
```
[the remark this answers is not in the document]
Martinsos: yeah that sounds about right! I guess "onAuthFailedRedirectTo" serves multiple things -> where to send user if they try to open a page that needs auth but they are not authenticated, but then also where to send user on failed login. First scenario is ok, but we don't want to use it upon failed login / signup -> better to just show an error message.

waspc/data/Generator/templates/server/src/auth/providers/config/github.ts
Martinsos: What are the options? Either we get it for them, or they get it on their own.

Maybe let's do the same as Passport strategy did, it sounds useful in practice, and will be a less of a breaking change.

Martinsos (commented):
Nice job @infomiho , I like the analytical and detailed approach!

I answered some top level stuff, and will leave the details to @sodic . Let me know if I missed something that you would like to get my opinion on.

scope -> scopes: why did that change? Because underlying thing names it that way? What is better, sounds like there is some semantic involved with those names? Is it actually scope or scopes, what makes more sense? I wouldn't try to avoid breaking change, I would instead go for the name that we think works the best, be it a breaking change or not, doesn't matter.

Passport strategies mapped the provider data into a specific shape -> how breaking is it? Will they need to migrate the database, because we stored that stuff in the DB? If so, then maybe it might make sense to do that mapping. If we are doing the mapping, is it complex, does it feel silly, or does it make sense, do we lose some info, ... -> how do we feel about it?
If there is no DB migration involved, then I would certainly not bother with any mapping (unless you think they way data is returned by Artcic is ugly and could benefit from mapping).

waspc/data/Generator/templates/server/src/auth/providers/config/github.ts
infomiho: Okay, this comment got outdated with latest changes. I replicated the logic that Passport had to keep the same information available. I think that makes the most sense!

waspc/data/Generator/templates/react-app/src/auth/pages/OAuthCallback.tsx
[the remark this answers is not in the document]
infomiho: > better to just show an error message.
I agree. Let's do that.

sodic:
> The property in the user config (when users use configFn property in Wasp file) to set scopes changed from scope to scopes (⚠️ breaking change) -> we can maybe make it backwards compatible?

I don't think we should - we're not stable enough for such sacrifices in consistency. What we can do is throw a nice error. Instead of "Unknown property: scope". We can say "Please change 'scope' to 'scopes' in this file".

> Passport strategies that we used mapped the [Google](https://github.com/jaredhanson/passport-google-oauth2/blob/master/lib/profile/openid.js#L20) data into a specific shape -> do we want to do it as well to be backwards compatible? (Same things goes for [Github](https://github.com/cfsghost/passport-github/blob/master/lib/profile.js#L8))

I think the same principle applies here. We're still immature enough to break stuff on account of creating something of high quality. Of course, change purely for the sake of change is also something we should avoid (especially when it impacts users). In short:
 - If the current format is a random thing people at passport) came up with and you see something obvious you want to change, I say "go for it".
 - If the current format is a well-documented or standard way of storing this particular provider's data (which could be the case if _everyone_ uses passport), I say we should leave it as is.

sodic (commented):
A couple of old comments

waspc/data/Generator/templates/react-app/src/auth/pages/OAuthCallback.tsx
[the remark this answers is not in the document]
infomiho: The OAuth login and signup, do in fact redirect. We are taking about this line: https://github.com/wasp-lang/wasp/blob/main/waspc/data/Generator/templates/react-app/src/auth/pages/OAuthCodeExchange.jsx#L58

waspc/data/Generator/templates/react-app/src/router.tsx
infomiho: I believe we can and we should!

waspc/examples/todoApp/src/user.ts
```diff
@@ -16,6 +16,11 @@ export function getName(user?: User) {
     return `Google user ${googleIdentity.providerUserId}`
   }
 
+  const githubIdentity = findUserIdentity(user, 'github')
+  if (githubIdentity) {
```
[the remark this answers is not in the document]
infomiho: > Still, since we're talking about a public API, we should probably do the ceremony and return null. Not now, of course, it's a breaking change that has nothing to do with this PR. But, if you agree, we can create an issue for polishing the public API.

Could you be more specific or better even, create the issue and link it here?

> NITPICK: In any case, I recommend using explicit comparisons instead of relying on coercion. I know typescript takes care of it, but I find it easier to verify the correctness of something like if (foo !== undefined) than if (foo).

Makes sense, I'll modify the helper.

waspc/examples/todoApp/src/auth/github.js
```diff
@@ -1,9 +1,7 @@
...
-    clientSecret: process.env['GITHUB_CLIENT_SECRET'],
-    scope: [],
+    scopes: ['user'],
```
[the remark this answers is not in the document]
infomiho: I'll create an issue to improve the docs 👍

waspc/data/Generator/templates/react-app/src/auth/pages/OAuthCallback.tsx
```diff
@@ -0,0 +1,72 @@
...
+    }}>
+      {error && <div style={{
+        color: "rgb(239 68 68)"
+      }}>{error}</div>}
+      {isLoading && <div>Please wait a moment while we log you in.</div>}
+    </div>
+  );
+}
+
+// TODO: don't hard code the URL
```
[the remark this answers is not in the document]
infomiho: I'd like to handle it now 👍

infomiho:
Of course! I've been waiting to get input from both Filip and you and I'll give some sort of a conclusion after we agree on what is the best here 👍

infomiho:
@Martinsos 

> scope -> scopes: why did that change? Because underlying thing names it that way? What is better, sounds like there is some semantic involved with those names? Is it actually scope or scopes, what makes more sense? I wouldn't try to avoid breaking change, I would instead go for the name that we think works the best, be it a breaking change or not, doesn't matter.

We just passed through the options to Passport before. Passport used `scope`, and that we pass through the options to Arctic, Arctic uses `scopes`. If we don't care about the backwards compatibility, I think it's okay to change the naming.

_IMHO We'll need good migration docs from 0.12 to 0.13 (Maybe even a script is possible?)_


> Passport strategies mapped the provider data into a specific shape -> how breaking is it? Will they need to migrate the database, because we stored that stuff in the DB? If so, then maybe it might make sense to do that mapping. If we are doing the mapping, is it complex, does it feel silly, or does it make sense, do we lose some info, ... -> how do we feel about it?
If there is no DB migration involved, then I would certainly not bother with any mapping (unless you think they way data is returned by Artcic is ugly and could benefit from mapping).

1. No database migrations will be necessary. What changes: the shape of the data they receive in `userSignupFields` handlers. So, if they used `data.profile.emails[0]` they'll need to use ... `data.profile.email` or smth along those lines to get the email.
2. Passport has their own format they tried to keep across their strategies: https://www.passportjs.org/reference/normalized-profile/ We could try to replicate their approach. Alternatively, we can just return what we receive from the provider. This will make our maintenance easier since if something changes, we are just passing it along anyways and no changes from our side our needed.

@sodic 

> I don't think we should focus on the breaking change part - we're not stable enough for such sacrifices in consistency (we shouldn't support both names).. Pick the more appropriate name, regardless of whether it causes a breaking change or not. 

Makes sense 👍 

> What we can do is throw a nice error. Instead of "Unknown property: scope". We can say "Please change 'scope' to 'scopes' in this file".

This could be a nice little touch. I've done stuff like this e.g. "Hey this error might mean you didn't run migrations" and it seemed to help the user.

> If the current format is a random thing people at passport) came up with and you see something obvious you want to change, I say "go for it".

As I said above, the format is something Passport invented and tried to stick to. But they are not the "one true solution" for OAuth, so I'd say, maybe stick to just passing through the info we get from providers. At least that way, it matches what the providers' might have in their docs?

> By the way, can we create migration scripts for these things?

+1 from me. I'd like to have at least good 0.12 to 0.13 migration docs.

waspc/examples/todoApp/src/auth/github.js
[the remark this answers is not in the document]
infomiho: https://github.com/wasp-lang/wasp/issues/1888

waspc/data/Generator/templates/server/src/auth/providers/oauth/cookies.ts
sodic, line 1: The first two functions (`setValueInCookie` and `getValueFromCookie`) are always composed with the other two functions (`getStateCookieName` and `getVerifierCookieName`).

I understand the idea (composability, smart constructors, etc), but would argue that it's a little too complex for its own good. Since it's a specialized API for OAuth cookies, I think we should make it as ergonomic as possible.

How do you feel about this?
```typescript
type OauthCookieKind = 'code_verifier' | 'state' 

function getOAuthCookieValue(provider: ProviderConfig, req: ExpressRequest, kind: OAuthCookieKind);
function setOAuthCookieValue(provider: ProviderConfig, res: ExpressResponse, kind: OAuthCookieKind, value: string);

// usage
setOAuthCookieValue(provider, res, 'state')
getOAuthCookieValue(provider, req, 'code_verifier')
```

Also, I like to follow [this rule](https://wiki.haskell.org/Parameter_order) when choosing the order of arguments. There's no currying in typescript, but the rule makes sense and is intuitive (IMO).

Another idea, Use your current API but have the cookie names as computed properties on provider configs:
```typescript
setValueInCookie(provider.stateCookieName, state, res)
```
I'd still reorder the arguments though.

waspc/data/Generator/templates/server/src/auth/providers/oauth/cookies.ts
```diff
@@ -0,0 +1,43 @@
+import {
+  Request as ExpressRequest,
+  Response as ExpressResponse,
+} from "express";
+import { parseCookies } from "oslo/cookie";
+
+type CookieName = {
+  cookieName: string;
+};
```
sodic, line 9: If you want a smart constructor, you can throw a symbol into the mix. That's gonna make it impossible for anyone to construct the value `CookieName` without using your functions. I can send a playground link if you don't know what I mean.

Although I'd prefer if we simplified it, as the other comment suggests.

waspc/data/Generator/templates/server/src/auth/providers/oauth/env.ts
```diff
@@ -0,0 +1,21 @@
+import { type ProviderConfig } from "wasp/auth/providers/types";
+
+export function ensureEnvVarsForProvider<EnvVars extends string>(
+  envVarNames: EnvVars[],
+  provider: ProviderConfig,
+): {
+  [name in EnvVars]: string;
```
sodic, line 7: Why not `Record<EnvVars, string>`?

waspc/data/Generator/templates/server/src/auth/providers/oauth/env.ts
```diff
@@ -0,0 +1,21 @@
+import { type ProviderConfig } from "wasp/auth/providers/types";
+
+export function ensureEnvVarsForProvider<EnvVars extends string>(
+  envVarNames: EnvVars[],
```
sodic, line 4: The `EnvVars` should be a singular name (i.e., `EnvVar`). Also, a better name is `EnvVarName`.

waspc/data/Generator/templates/server/src/auth/providers/oauth/env.ts
```diff
@@ -0,0 +1,21 @@
...
+  provider: ProviderConfig,
+): {
+  [name in EnvVars]: string;
+} {
+  const result: {
+    [name: string]: string;
+  } = {};
```
sodic, line 11: I see that sometimes you use index signatures, while other times you use Records.

If you're going to mix and match, I recommend:
- Records for bounded types and/or when you want to document the key's name
- Index signatures for unbounded types and when you don't care about documenting the key's name

See here for details: https://github.com/wasp-lang/wasp/pull/1235#discussion_r1300022520

waspc/data/Generator/templates/server/src/auth/providers/oauth/env.ts
```diff
@@ -0,0 +1,21 @@
...
+  for (const envVarName of envVarNames) {
+    if (!process.env[envVarName]) {
+      throw new Error(`${envVarName} env variable is required when using the ${provider.displayName} auth provider.`);
+    }
+    result[envVarName] = process.env[envVarName];
```
sodic, line 16: `process.env[envVarName]` can be extracted into a variable.

waspc/data/Generator/templates/server/src/auth/providers/oauth/oneTimeCode.ts
```diff
@@ -0,0 +1,77 @@
...
+      return validateJWT<{ id: string }>(token);
+    },
+    isUsed(token: string) {
+      return usedTokens.has(token);
+    },
+    markUsed(token: string) {
+      usedTokens.set(token, Date.now());
+      cleanUp();
+    },
+  };
```
sodic, line 76: You might want to define all functions in the body of `createTokenStore`, and then just reference them when returning the object.

I'm fine with it either way, decide what you prefer.

waspc/data/Generator/templates/server/src/auth/providers/oauth/redirect.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const loginPath = '{= serverOAuthLoginHandlerPath =}'
+export const callbackPath = '{= serverOAuthCallbackHandlerPath =}'
+export const exchangeCodeForTokenPath = '{= serverExchangeCodeForTokenHandlerPath =}'
+
+export function getRedirectUriForCallback(providerName: string) {
+  return `${config.serverUrl}/auth/${providerName}/${callbackPath}`;
+}
+
+const clientOAuthCallbackPath = '{= clientOAuthCallbackPath =}'
```
sodic, line 12: I'd move this to the end of the file (or at least add a space), or even the beginning next to other constants.

waspc/data/Generator/templates/server/src/auth/providers/oauth/user.ts
```diff
@@ -0,0 +1,85 @@
...
+  return {
+      redirectUri: getRedirectUriForOneTimeCode(oneTimeCode),
+  };
+}
+
+export function handleOAuthErrorAndGetRedirectUri(error: unknown): { redirectUri: string } {
+  const errorMessage = (error as any).message ?? "An unknown error occurred while trying to log in with the OAuth provider.";
+  return {
+    redirectUri: getRedirectUriForError(errorMessage),
+  };
```
sodic, line 41: How come you are wrapping this inside an object.

If it's because you want a smart constructor, it makes sense to extract the type (and employ a symbol, as mentioned in the other comments).

Still, I think it's OK if you just return a string.

sodic (commented):
Left improvement suggestions, but no major problems.

Here's the only major thing I'd like us to discuss: https://github.com/wasp-lang/wasp/pull/1851#discussion_r1526228469

Nice work!

waspc/data/Generator/templates/react-app/src/auth/pages/OAuthCallback.tsx
```diff
@@ -0,0 +1,82 @@
...
+  return (
+    <div style={{
+      display: "flex",
+      alignItems: "center",
+      justifyContent: "center",
+      padding: "4rem",
+    }}>
+      {error && <div style={{
+        color: "rgb(239 68 68)"
+      }}>{error}</div>}
```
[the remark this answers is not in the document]
infomiho: Inline CSS is the most vanilla way of styling React components (next to using CSS modules, but I wanted to keep everything in one file), so that's perfectly fine.

The `div`s are also okay IMHO since there isn't really any other semantic elements that fit for this use case. 

I've moved the CSS into separate objects to make it a bit more readable :)

waspc/data/Generator/templates/react-app/src/auth/pages/OAuthCallback.tsx
```diff
@@ -0,0 +1,82 @@
...
+    }
+  }
+
+  const isFirstRender = useRef(true);
+  useEffect(() => {
+    if (isFirstRender.current) {
+      isFirstRender.current = false;
+      handleCallback();
+    }
+  }, []);
```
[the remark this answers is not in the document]
infomiho: This is a unique hack that we'll only use here. In React 18 with strict mode, each component renders twice. This makes this page problematic since it will send the OAuth code twice (once successfully, once with an error since the OAuth code can be used only once).

waspc/src/Wasp/Generator/ServerGenerator/Auth/OAuthAuthG.hs
infomiho: I've given it a shot to minimise the repeated code. We could probably go further, but I didn't to be too smart about inventing new abstractions, this now feels okay, check it out and tell me what do you think. Should we go further?

waspc/data/Generator/templates/server/src/auth/providers/oauth/cookies.ts
infomiho: Nice simplification of the API! 

I've now adjusted the code to be more like
```
setOAuthCookieValue(provider, res, 'state', state)
getOAuthCookieValue(provider, req, 'code_verifier')
```

waspc/data/Generator/templates/server/src/auth/providers/oauth/user.ts
infomiho: I've decided to return URL since Arctic's method for generating the login URL returns an actual `URL` object. It keeps things consistent and also validates the URL string which is a nice side-effect.

waspc/src/Wasp/Generator/ServerGenerator/Auth/OAuthAuthG.hs
infomiho: I believe we prematurely abstracted stuff here. This code assumes that there are only client ID and client secret env vars, but in reality, there can be 2, 3 or 4 even env vars needed to set up a provider. We don't use the knowledge about the env vars in Haskell so it wasn't really useful.

Also, I wanted to keep things as simple as possible to make the rewrite easier :)

waspc/data/Generator/templates/server/src/auth/providers/oauth/user.ts
sodic: Even better

waspc/data/Generator/templates/server/src/auth/providers/oauth/oneTimeCode.ts
```diff
@@ -0,0 +1,85 @@
...
+      const { code } = req.body;
+
+      if (tokenStore.isUsed(code)) {
+        throw new HttpError(400, "Unable to login with the OAuth provider. The code has already been used.");
+      }
+
+      const { id: authId } = await tokenStore.verifyToken(code);
+      const auth = await findAuthWithUserBy({ id: authId })
+
+      if (!auth) {
```
sodic, line 25: Explicit check (if you want)

waspc/data/Generator/templates/server/src/auth/providers/oauth/oneTimeCode.ts
```diff
@@ -0,0 +1,85 @@
...
+      },
+      {
+        expiresIn: validFor,
+      }
+    );
+  }
+
+  function verifyToken(token: string) {
+    return validateJWT<{ id: string }>(token);
+  }
```
sodic, line 59: Why not:
```typescript
function getVerifiedToken(token: string) Promise<id: string> {
  return validateJWT(token);
}
```
It's a part of the public API through the returned object, so it makes sense to type it explicitly. Also, I'd say it's always better to type function signatures and leave the type inference for the body than the other way around.

waspc/data/Generator/templates/server/src/auth/providers/oauth/state.ts
```diff
@@ -0,0 +1,70 @@
...
+    result.state = state;
+  }
+
+  if (stateTypes.includes('codeVerifier' as ST)) {
+    const codeVerifier = generateCodeVerifier();
+    setOAuthCookieValue(provider, res, 'codeVerifier', codeVerifier);
+    result.codeVerifier = codeVerifier;
+  }
+
+  return result as { [name in ST]: string };
```
sodic, line 32: A less error prone way to implement this is by generating the resulting state first, and then using that to set the value. In pseudocode:
```typescript
// construct the result
setOAuthCookieValue(provider, res, result); // this function would iterate the object and set all cookies
return result
```
I probably wouldn't do this now though and would just merge this. Not a big deal.

waspc/data/Generator/templates/server/src/auth/providers/oauth/state.ts
```diff
@@ -0,0 +1,70 @@
...
+    result.codeVerifier = storedCodeVerifier;
+  }
+
+  const code = req.query.code;
+  if (typeof code !== "string") {
+    throw new Error("Invalid code");
+  }
+  result.code = code;
+
+  return result as { [name in ST]: string } & { code: string };
```
sodic, line 69: I don't think you need this cast as the function is currently written.

waspc/data/Generator/templates/server/src/auth/providers/oauth/user.ts
```diff
@@ -0,0 +1,90 @@
...
+  provider: ProviderConfig,
+  providerProfile: unknown,
+  providerUserId: string,
+  userSignupFields: UserSignupFields | undefined,
+): Promise<URL> {
+  const providerId = createProviderId(provider.id, providerUserId);
+
+  const authId = await getAuthIdFromProviderDetails(providerId, providerProfile, userSignupFields);
+
+  if (!authId) {
```
sodic, line 26: Explicit check if you want

waspc/src/Wasp/Generator/ServerGenerator/Auth/OAuthAuthG.hs
sodic: I think that's enough abstraction for now. We'll come back to it when we add more providers and are smarter.

sodic (approved):

waspc/data/Generator/templates/server/src/auth/providers/oauth/user.ts
infomiho: This function always returns a value or it throws an error, so I've removed this check.

PR #1852, opened 2024-03-05
Replaces `jsonwebtoken` with `oslo/jwt`

infomiho:
Reasons for doing this:
- cleaning up old code a bit
- solidify our investment in `oslo`
- since already have `oslo` for password hashing, it makes sense to drop the extra depedency on `jsonwebtoken`

Testing:
- I've compared the output and decoding the token. `jsonwebtoken` and `oslo/jwt` are compatible, users will not be affected by this change.
- I've tested out the email verification and password reset flows, they work.
- If the token is expired or invalid we get an error message.


waspc/data/Generator/templates/sdk/wasp/auth/jwt.ts
```diff
@@ -1,14 +1,22 @@
-import jwt from 'jsonwebtoken'
-import util from 'util'
-
+import * as jwt from 'oslo/jwt'
 import { config } from 'wasp/server'
 
-const jwtSign = util.promisify(jwt.sign)
-const jwtVerify = util.promisify(jwt.verify)
+const JWT_SECRET = new TextEncoder().encode(config.auth.jwtSecret)
+const JWT_ALGORITHM = 'HS256'
```
infomiho, line 5: We used the default algorithm with `jsonwebtoken` which is `HS256`

waspc/data/Generator/templates/sdk/wasp/auth/jwt.ts
```diff
@@ -1,14 +1,22 @@
...
 // PRIVATE API
-export const signData = (data, options) => jwtSign(data, JWT_SECRET, options)
+export async function validateJWT<Payload>(token: string): Promise<Payload> {
+  const { payload } = await jwt.validateJWT(JWT_ALGORITHM, JWT_SECRET, token)
+  return payload as Payload
+}
+
 // PRIVATE API
-export const verify = (token) => jwtVerify(token, JWT_SECRET)
+export { TimeSpan } from 'oslo'
```
infomiho, line 22: `TimeSpan` class is used to describe the expiration e.g. `TimeSpan(30, "m")`

waspc/data/Generator/templates/sdk/wasp/auth/utils.ts
```diff
@@ -174,11 +173,6 @@ export async function deleteUserByAuthId(authId: string): Promise<{ count: numbe
   } } })
 }
 
-// PRIVATE API
-export async function verifyToken<T = unknown>(token: string): Promise<T> {
```
infomiho, line 178: Not needed any more

waspc/data/Generator/templates/server/src/auth/providers/email/resetPassword.ts
```diff
@@ -37,10 +36,7 @@ export async function resetPassword(
             hashedPassword: password,
         });
     } catch (e) {
-        const reason = e.name === tokenVerificationErrors.TokenExpiredError
-            ? 'expired'
-            : 'invalid';
-        throw new HttpError(400, `Password reset failed, ${reason} token`);
+        throw new HttpError(400, `Password reset failed, invalid token`);
```
infomiho, line 39: Simplifying the error messages (in part to avoid parsing `oslo/jwt` messages since they are just `Error`s)

waspc/data/Generator/templates/server/src/auth/utils.ts
```diff
@@ -160,10 +159,6 @@ export async function deleteUserByAuthId(authId: string): Promise<{ count: numbe
   } } })
 }
 
-export async function verifyToken<T = unknown>(token: string): Promise<T> {
```
infomiho, line 163: Not needed anymore

waspc/src/Wasp/Generator/ServerGenerator/Auth/EmailAuthG.hs
```diff
@@ -100,9 +99,3 @@ genRoutes =
     ]
   where
     genFileCopy = return . C.mkSrcTmplFd
-
-genTypes :: AS.Auth.EmailAuthConfig -> Generator FileDraft
```
infomiho, line 104: Not needed anymore

waspc/data/Generator/templates/server/src/auth/providers/email/types.ts
```diff
@@ -1,3 +0,0 @@
-export const tokenVerificationErrors = {
```
infomiho, line 1: Not needed anymore

waspc/data/Generator/templates/sdk/wasp/auth/jwt.ts
```diff
@@ -1,14 +1,22 @@
...
 
-const jwtSign = util.promisify(jwt.sign)
-const jwtVerify = util.promisify(jwt.verify)
+const JWT_SECRET = new TextEncoder().encode(config.auth.jwtSecret)
+const JWT_ALGORITHM = 'HS256'
 
-const JWT_SECRET = config.auth.jwtSecret
+// PRIVATE API
+export function createJWT(
+  data: Parameters<typeof jwt.createJWT>[2],
```
Martinsos, line 9: Ok this `Parameters<type of function>[i]` is nice, good to know!

waspc/data/Generator/templates/server/src/auth/providers/email/resetPassword.ts
Martinsos: Aha, but now we can't say if it is expired or something else, right? At least we are not doing it any more. Is that bad?

Can you at least check here if `e` was indeed thrown by `oslo/jwt`?
It is generally a bit scary to catch just about any error and then ignore its value, as it could be anything really -> maybe `createProviderId` failed, or `findAuthIdentity` failed -> yeah, this is too wide a net we are casting here.

I would both make the `try` capture less of the code (just `validateJWT` statement) and also try to look a bit at the error, and throw 400 only if it is what we expected, otherwise rethrow it.

Martinsos (changes requested):
Looks great, a nice change, but I am rejecting changes because of how an error is handled -> once we figure that out, I think that is it.

waspc/data/Generator/templates/server/src/auth/providers/email/resetPassword.ts
infomiho: > It is generally a bit scary to catch just about any error and then ignore its value, as it could be anything really -> maybe createProviderId failed, or findAuthIdentity failed -> yeah, this is too wide a net we are casting here.

I wouldn't go any more specific than "something failed" or "the token is invalid" because that gives information to the attacker e.g. if we said "that user cannot be found" that would reveal extra info.

I proposed that we:
- say "the token is invalid" in the case of an invalid token (code: 400)
- say "something went wrong" in any other case (code: 500)

infomiho: Updated error handling here: https://github.com/wasp-lang/wasp/pull/1852/commits/fa0ad66d4b789df5f3c053933a1271bc7aaac58e

Martinsos: Yeah exactly -> this is default behaviour. So if you throw a HttpError, it will go out with that status, but if you just rethrow, our default error middleware we wrap it into 500, and hide any info. And it will show more info in development (I think, if not we should add that, but that is nice property of having it centralized like that).

infomiho: Maybe then best to remove the try catch if we handled some of the specific cases by throwing `HttpError` 😄 

Okay, I'll try to write it like that and report back.

Martinsos (approved):

PR #1856, opened 2024-03-05
Define the `WASP_SERVER_URL` env var

infomiho:
Before we did the OAuth redirect like this:
```
[server] ---> [Google] ---> [client] --> (client sends the Google code to server)
```

We need the `serverUrl` variable for the new OAuth setup since we are doing the redirect like this:

```
[server] ---> [Google] --> [server (gets the code directly from the redirect)] ---> [client]
```


Martinsos (commented):
Looking good from a quick check!

The only thing I thought about is that the name doesn't fit the name we use for the client. So on client, to specify the server's URL, we use `REACT_APP_API_URL`, where `API_URL` is really the important part, REACT_APP is just prefix. And that prefix is also silly, it is just a reminder from using CRA.
So, if we call it "API_URL" there, it is a bit weird that we call it "SERVER_URL" here. Would be nicer if they are the same. If we don't want to do it now, maybe we can open an issue for it, or maybe it is better to handle now. I don't think "API_URL" is a good name really though. "SERVER_URL" sounds like a good name. So it would be breaking change. Maybe we can do it at the same time when we drop the REACT_APP prefix?

Btw pls also add to Changelog immediately what you did, once PR has taken final shape.

waspc/data/Generator/templates/sdk/wasp/server/config.ts
```diff
@@ -28,6 +28,7 @@ type CommonConfig = BaseConfig & {
 
 type EnvConfig = BaseConfig & {
   frontendUrl: string;
+  serverUrl: string;
```
sodic, line 31: Does this include the port and the protocol? If so, we have a coupling between two config fields (`port` and `serverUrl`) and should modify our code to account for it.

waspc/src/Wasp/Generator/ServerGenerator/Common.hs
```diff
@@ -124,3 +125,6 @@ toESModulesImportPath :: FilePath -> FilePath
 toESModulesImportPath = changeExtensionTo "js"
   where
     changeExtensionTo ext = (++ '.' : ext) . fst . splitExtension
+
+defaultDevServerUrl :: String
+defaultDevServerUrl = "http://localhost:3001"
```
sodic, line 130: Do we want to support subpaths (like we do for the client)?

waspc/data/Generator/templates/sdk/wasp/server/config.ts
infomiho: - So we need the server URL to be able to implement OAuth successfully. In dev, we need the default local server URL to be set automatically e.g. `http://localhost:3001`. If the user can change the port to `6000` then it would be good if we set the `serverUrl` to `http://localhost:6000`.

- In the production use case, the port start being meaningless since they use a domain name e.g. if their port is `7000` for some reason, the server URL will still be the same e.g. `https://my-server.com`

- This means, we need to make sure that. the default dev server is set according to the port, but in the production case we should just use the user env var.

waspc/data/Generator/templates/react-app/src/config.js
```diff
@@ -1,9 +0,0 @@
-import { stripTrailingSlash } from "./universal/url";
```
infomiho, line 1: This template file wasn't used anywhere, so I removed it. It's a leftover from migrating to the SDK.

waspc/data/Generator/templates/sdk/wasp/server/config.ts
```diff
@@ -40,7 +41,7 @@ const config: {
   all: {
     env,
     isDevelopment: env === 'development',
-    port: parseInt(process.env.PORT) || 3001,
+    port: parseInt(process.env.PORT) || {= defaultServerPort =},
```
infomiho, line 44: @sodic Now the default server port and default server URL are both defined in Haskell and the URL is constructed using the default port variable.

infomiho:
@Martinsos Based on your comment, you are fine with the `SERVER_URL` name and would like to rename our client facing env variable to `REACT_APP_SERVER_URL` or similar. So, let's move this forward like this and worry about the client env in the future.

We would still need to keep some sort of prefix. `VITE_` is the default when using Vite (vs. `REACT_APP_` that we kept from CRA)

Martinsos:
> @Martinsos Based on your comment, you are fine with the `SERVER_URL` name and would like to rename our client facing env variable to `REACT_APP_SERVER_URL` or similar. So, let's move this forward like this and worry about the client env in the future.
> 
> FYI We would still need to keep some sort of prefix. `VITE_` is the default when using Vite (vs. `REACT_APP_` that we kept from CRA).

Ideal would probably be something like WASP_, or CLIENT_ -> whatever makes most sense, but I would try to not have it specific to underlying technology.

I understand if we are not syncing the names now, but in that case pls create a GH issue to do this.

PR #1862, opened 2024-03-07
Remove client file hacks from `server/tsconfig.json`

### LINKED ISSUE #1855: Investigate whether the server can work without client file hacks
Context: https://github.com/wasp-lang/wasp/pull/1584#discussion_r1404020428

> [!NOTE]
> It makes sense to tackle this first: #1854

Can we remove the client-specific options in our `server/tsconfig.json`?

With the current setup, it depends on how typescript (or rollup) chooses which files to process:
 - If it's "by demand" (i.e., only process what you need), then we probably don't need the hacks.
 - If it's "by includes directories" (i.e., process all files in directories that are part of the project), then we probably need them.
 
I don't know which one is it, so we should investigate.

sodic:
Closes #1855.


infomiho (approved):
LGTM

Tested locally as well to be extra safe

PR #1865, opened 2024-03-07
User defined imports name mangling

### LINKED ISSUE #284: In Generator, ensure user defined identifiers don't get into conflict with rest of the template as they are inserted
Code is generated from templates, where we insert pieces of code that user defined.

Sometimes, that means we insert imports that import user functions from user (@ext) files.

Or, we address some identifier that user defined somewhere.

This can cause problems if for example they want to import function `debug` and we add that import into template that already imports our own `debug` -> we have a conflict!

While I expect Generator to go through significant changes in the future, where we should end up with a more sophisticated model of describing specific templates, I believe this problem will remain.

The only solution I see is prefixing (or suffixing) all user-made identifiers with something like `wasp123`, where the number would be unique per identifier (we can just assign those numbers incrementally).

Good example of problematic situation is when setup server function is inserted into server.js. What is that function was named `config` by user? It would conflict with the `config` import that we already have in that file.

Bad side of this is that names will be somewhat obfuscated, but on the other hand, if we just prefix or suffix them, that will be relatively straight forward to remove in case devs want to eject. When I say straight forward, I mean so for the devs themselves. If we want to remove prefix/suffix, it will be tricky since we need to detect conflicts and then ask dev to resolve them. So probably best to let them do it.

### LINKED ISSUE #1062: Allow naming jobs the same as the JS implementation function
If the user names a job and the JS function that implements it, the jobs feature breaks. This error class is probably lurking in other Wasp features as well. 

### Example
```js
job generateResult {
  executor: PgBoss,
  perform: {
    fn:  import { generateResult } from "@server/generation.js",
    executorOptions: {
      pgBoss: {=json { "retryLimit": 1 } json=}
    }
  },
  entities: [Generation],
}
```
### Error
```
Server(stderr): file:///Users/ilakovac/dev/socialpostgpt/.wasp/out/server/dist/jobs/generateResult.js:4
Server(stderr): export const generateResult = createJob({
Server(stderr):              ^
Server(stderr):
Server(stderr): SyntaxError: Identifier 'generateResult' has already been declared
Server(stderr):     at ESMLoader.moduleStrategy (node:internal/modules/esm/translators:119:18)
Server(stderr):     at ESMLoader.moduleProvider (node:internal/modules/esm/loader:468:14)
Server(stderr):
Server(stderr): Node.js v18.15.0
```

### Cause
It caused because we just use the user defined names verbatim, maybe we should consider doing name mangling.

Generated code:
```js
import prisma from '../dbClient.js'
import { createJob } from './core/pgBoss/pgBossJob.js'
// Notice the import
import { generateResult } from '../ext-src/generation.js'

// And the variable name
export const generateResult = createJob({
  jobName: "generateResult",
  jobFn: generateResult,
  defaultJobOptions: {"retryLimit":1},
  jobSchedule: null,
  entities: {
    Generation: prisma.generation,
  },
})
```

infomiho:
I've had some issues with conflicting imports when integrating Arctic https://github.com/wasp-lang/wasp/pull/1851 and implemented this mangling to help.

Closes #1062
Closes #284

### DIFF
--- a/waspc/src/Wasp/Generator/WebAppGenerator/JsImport.hs
+++ b/waspc/src/Wasp/Generator/WebAppGenerator/JsImport.hs
@@ -6,11 +6,11 @@ import StrongPath.TH (reldirP)
 import Wasp.AppSpec.ExtImport (ExtImport (..))
 import qualified Wasp.AppSpec.ExtImport as EI
 import qualified Wasp.Generator.JsImport as GJI
+import Wasp.Generator.Monad (Generator)
 import Wasp.Generator.WebAppGenerator.Common (WebAppSrcDir)
 import Wasp.JsImport
   ( JsImport,
   )
-import Wasp.Generator.Monad (Generator)
 
 extImportToImportJson ::
   Path Posix (Rel importLocation) (Dir WebAppSrcDir) ->


waspc/src/Wasp/Generator/JsImport.hs
```diff
@@ -43,9 +46,16 @@ jsImportToImportJson maybeJsImport = maybe notDefinedValue mkTmplData maybeJsImp
...
               "importStatement" .= jsImportStmt,
               "importIdentifier" .= jsImportIdentifier
             ]
+      where
+        mangleImportIdentifier :: JsImport -> JsImport
+        mangleImportIdentifier JI.JsImport {JI._name = JsImportModule name} = prefixTheImportName name jsImport
+        mangleImportIdentifier JI.JsImport {JI._name = JsImportField name} = prefixTheImportName name jsImport
+
+        prefixTheImportName :: String -> JsImport -> JsImport
+        prefixTheImportName originalName = applyJsImportAlias (Just ("__userDefined" ++ toUpperFirst originalName))
```
sodic, line 61: I wanted to suggest mangling our imports instead of user imports (to keep the path between user symbols and our code as clear as possible).

However,  it seems that it's not our imports conflicting with user imports - it's the thing we're defining in the same file.

Anyway, I don't how to make this without changing user symbol names, but am leaving this comment in case you have an idea.

sodic (commented):
We already do the same thing in other places.

For example, these two
https://github.com/wasp-lang/wasp/blob/8d7c64a2967a1d9987876c70822e815de68f7418/waspc/src/Wasp/Generator/SdkGenerator/Server/OperationsGenerator.hs#L146

https://github.com/wasp-lang/wasp/blob/8d7c64a2967a1d9987876c70822e815de68f7418/waspc/src/Wasp/Generator/ServerGenerator/ApiRoutesG.hs#L56

And possibly others.

I suggest a regex text search in the generated project). I used this command to find the ones I listed here (but didn't do a fully thorough check):
   ```bash
   rg 'import.*\{.*as.*\}' -g '**/src/**/*' -g '**/sdk/**/*' -g '!**/dist/**/*'
   ```

sodic:
Btw, your PR description (as its written now) won't trigger automatic closing of both issues. I got burned by this several times.

Instead of:
```
Closes #123 and #234
```
Do
```
Closes #123
Closes #234 
```

Context: https://github.com/nus-cs2103-AY2021S2/forum/issues/293

infomiho:
Okay, so we need to unify the way we mangle the imports. I found the following using your regex:

- `server/src/routes/apis/index.ts`-> `ApiRoutesG` (`"wasp_"` prefix)
- `sdk/wasp/server/operations/actions/index.ts` and `sdk/wasp/server/operations/queries/index.ts` -> `OperationsGenerator` -> (`"_ext"` suffix)
- `server/src/middleware/globalMiddleware.ts` -> `ServerGenerator` (custom `"_waspGlobalMiddlewareConfigFn`")
- `server/src/jobs/*.ts` -> `JobGenerator` (custom `"_waspJobDefinition"`)
- `sdk/wasp/server/webSocket/index.ts` -> `WebSocketGenerator` using the same fn as `OperationsGenerator`


Okay, that seems manageable. Great call to ripgrep!

infomiho:
The biggest issue I'd like to resolve to make this "proper name mangling" is to somehow keep track of aliases that were used.

### Basic case

Ideally, we would keep record of assigned aliases within a single template e.g. 
- `import { foo as foo__userDefined } from  ...`
- `import { bar as bar__userDefined from ...`
 
From this we would keep track that we had aliases `foo__userDefined` and `bar__userDefined`. This is all fine and the basic case of nothing going wrong.

### Duplicate alias

What if we had a situation like this in a single template file:
- `import { foo as foo__userDefined } from  ...`
- `import { foo as foo__userDefined } from  ...`
 
From this we would keep track that we had aliases `foo__userDefined` and `foo__userDefined`. This would cause our code to break ⚠️ 

### Ideal solution (keep track at template level)

Ideally, we would do it like e.g. Rollup does it and keep a global store of aliases and then on duplicate, we would just add a number at the end. We would then end up with something like this:
- `import { foo as foo__userDefined$0 } from  ...`
- `import { foo as foo__userDefined$1 } from  ...`

### Okay solution IMHO (keep track at global level)

Same like the previous solution, but we could have a "global state" of previously used aliases and then apply the same suffix logic like in the ideal solution. It might end up like this (if we had many `foo__userDefined` aliseses):
- `import { foo as foo__userDefined$24 } from  ...`
- `import { foo as foo__userDefined$25 } from  ...`

waspc/data/Generator/templates/server/src/dbSeed.ts
```diff
@@ -9,12 +9,12 @@
...
 
 {=# dbSeeds =}
-{=& importStatement =}
+{=& seedFn.importStatement =}
 {=/ dbSeeds =}
 
 const seeds = {
   {=# dbSeeds =}
-  {= importIdentifier =},
+  {= name =}: {= seedFn.importIdentifier =},
```
infomiho, line 17: Since we reference the import name in the CLI, we had to adjust the code a bit.

waspc/data/Generator/templates/server/src/middleware/globalMiddleware.ts
```diff
@@ -11,9 +11,10 @@ export type { MiddlewareConfig, MiddlewareConfigFn } from 'wasp/server/middlewar
 
 {=# globalMiddlewareConfigFn.isDefined =}
 {=& globalMiddlewareConfigFn.importStatement =}
+const globalMiddlewareConfigFn = {= globalMiddlewareConfigFn.importIdentifier =}
 {=/ globalMiddlewareConfigFn.isDefined =}
 {=^ globalMiddlewareConfigFn.isDefined =}
-const {=& globalMiddlewareConfigFn.importAlias =} = (mc: MiddlewareConfig) => mc
```
infomiho, line 16: No need for a custom alias anymore 😄

waspc/data/Generator/templates/server/src/routes/apis/index.ts
```diff
@@ -9,32 +9,32 @@ import { type AuthUser } from 'wasp/auth'
...
 {=# routeMiddlewareConfigFn.isDefined =}
 {=& routeMiddlewareConfigFn.importStatement =}
 {=/ routeMiddlewareConfigFn.isDefined =}
 {=/ apiRoutes =}
 
 const idFn: MiddlewareConfigFn = x => x
 
-{=# apiRoutes =}
-{=^ routeMiddlewareConfigFn.isDefined =}
-const {=& routeMiddlewareConfigFn.importAlias =} = idFn
```
infomiho, line 26: No need for custom aliases anymore, rewrote this code to work with the new mangling.

Martinsos:
@infomiho I am curious, when does this situation where two user-defined symbols come into conflict happen? Does it happen if they have a query and action with the same name, or something like that? I am asking, because I wonder if we are even ok with allowing them to do something like that. But I am guessing there is some other use case you were considering here.

Because for me, the main thing I wanted us to fix with name mangling was ensuring the user-defined symbols don't come into conflict with the symbols that we defined.

waspc/src/Wasp/Generator/Monad.hs
```diff
@@ -61,24 +64,44 @@ runGenerator generator =
...
+mangleName originalName = do
+  mangledNamesMap <- gets mangledNames
+
+  let (mangledName, newMangledNamesMap) = mangleName' originalName mangledNamesMap
+
+  modify $ \s -> s {mangledNames = newMangledNamesMap}
+
+  return mangledName
+  where
+    mangleName' :: String -> Data.Map.Map String Int -> (String, Data.Map.Map String Int)
```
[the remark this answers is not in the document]
Martinsos: Ok there is one thing I thought of only now -> and that is that we might want to minimize the impact of them adding a new user defined name to the codebase. If such user defined name shifts all other names for a +1, then we have a lot of files updating in generated code, while we could have easily avoided that. Is that why you used the approach you used? If so, that is a quite good reason, and it might be worthing mentioning it in the comments somewhere, in case somebody in the future gets same idea like me to simplify the code.

waspc/src/Wasp/Generator/SdkGenerator/Server/JobGenerator.hs
```diff
@@ -91,15 +91,14 @@ genJob (jobName, job) =
...
 
-getImportJsonForJobDefinition :: String -> Aeson.Value
+getImportJsonForJobDefinition :: String -> Generator Aeson.Value
 getImportJsonForJobDefinition jobName =
   GJI.jsImportToImportJson $
     Just $
       JI.JsImport
         { JI._path = JI.ModuleImportPath $ makeSdkImportPath [relfileP|server/jobs|],
           JI._name = JI.JsImportField jobName,
-          -- NOTE: We are using alias to avoid name conflicts with user defined imports.
```
Martinsos, line 101: Ok, so we are not aliasing our names anymore, because we are aliasing their names.

I wonder how this will affect the error messages? Will they now be seeing stuff like `getTasks__userDefined` in the error messages? If so, it would be better that we mangle our names, and not theirs, when possible.

Also, I remember at some point in the past, I was worrying about the situation where we embed their code directly into the template (so not just importing it, but actually embedding it) -> then there was a lot of opportunity for conflicts with our names in that file. I think we don't have such situations any more though, since we don't embed their code directly -> we had that as a feature in Wasp in very early days and that has been removed.

Martinsos (commented):
Awesome initiative @infomiho !
I like the general idea, although it seems to me like we should still figure out some details a bit: I left comments.

I didn't review all the single files in the Generator where names are defined, but I reviewed the files with the core name mangling logic.

infomiho:
This PR is stale and I think we'd probably want to redo name mangling from scratch when we decide to continue this. This will serve as a reference for any future work 👍

PR #1887, opened 2024-03-13
Update the release process guide

sodic:
I did the 80/20 on this one by just adding disclaimers and markers to steps that don't always apply, just enough to reassure the reader and tell them "You don't have to necessarily do this, but think about whether you do."


waspc/README.md
sodic, line 1: There is some autoformatting.

Martinsos (commented):
@sodic approved, seems practical!

I think "bold" is maybe not the best choice, as it can be a bit hard ot figure out sometimes what is bold and what is not with all the links and other styling, and also you just made it impossible to use bold normally in this piece of text, where it might be useful to emphasize some stuff since there is a lot of text.

I would maybe suggest a different measure, something like putting `[!]` at the start of that bullet point. What would probably work the best is an emoji, because it is visually easy to detect and comprehend. You could do something like ⭐ or ❗ or ❕ or 👉 .

Martinsos (approved):

PR #1893, opened 2024-03-14
Add client config to SDK API

### LINKED ISSUE #1735: Export client config from SDK
We should give our users access to the client config object (the same way we do for the server config).

I ran into this issue because one of our example apps (streaming) used an undocumented import of `@wasp/config` to get the `apiUrl`.

sodic:
Closes #1735


waspc/data/Generator/templates/sdk/wasp/client/index.ts
```diff
@@ -9,3 +9,5 @@ export enum HttpMethod {
 
 // PUBLIC API
 export type Route = { method: HttpMethod; path: string }
+
+export { default as config } from './config'
```
infomiho, line 13: PUBLIC API comment

waspc/data/Generator/templates/sdk/wasp/client/test/vitest/helpers.tsx
```diff
@@ -7,7 +7,7 @@ import { render, RenderResult, cleanup } from '@testing-library/react'
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
 import { beforeAll, afterEach, afterAll } from 'vitest'
 import { Query } from 'wasp/client/operations/core'
-import config from 'wasp/core/config'
+import { config } from 'wasp/client'
```
infomiho, line 10: Maybe add this rewrote to the code mod as well? Probably some users used it.

sodic: Nice thinking, but it was never a part of the public API (and we had the same path for the server config, which was).

Therefore, the code mod is up to date.

infomiho: What's the name of the principle you like to mention? The public API is what users think there public API is 😂

sodic: Hyrum's law, but I don't think we understood each other.

In 11.8, the import `wasp/config` resolved to two symbols (depending on which folder it was in). One of those symbols was part of the public API (the server config), the other wasn't (the client config).

Codemod does not support mapping a single import (`wasp/config`) to two different new imports (`wasp/client` and `wasp/server`) depending on where it finds the file that uses them.
It currently maps all config imports from `wasp/config` imports to `wasp/server` imports (which was a documented feature) regardless of where it finds them.

I could therefore either:
1. Change codemod to map `wasp/config` imports to `wasp/client` imports, which would be correct for the undocumented client config imports, but incorrect for the documented server config imports (which, of course, doesn't make much sense :smile:).
2. Update codemod to be context-aware and map `wasp/config` to the correct path depending on whether it's inside the `src/client` or `src/server` directory. This is obviously the best solution, but would take a lot of work for a feature we technically never had, so I haven't considered it.

web/docs/advanced/accessing-app-config.md
```diff
@@ -0,0 +1,59 @@
...
+ 
+  During development, this is an Express server controlled by a
+  [`nodemon`](https://www.npmjs.com/package/nodemon) process that takes care of
+  hot reloading and restarts. In production, it's a regular Express server run
+  using Node.
+
+Check [the introduction](/docs) for a more in-depth explanation of Wasp's runtime architecture.
+
+You can configure both processes through environment variables. See [the
+deployment instructions](/docs/advanced/deployment/manually#environment-variables) for a full list
```
sodic, line 22: We should make a better list. Here's an issue that tracks it: https://github.com/wasp-lang/wasp/issues/1894

web/docs/advanced/accessing-app-config.md
```diff
@@ -0,0 +1,59 @@
...
+of supported variables.
+
+Wasp gives you runtime access to the processes' configurations through **configuration objects**.
+
+## Server configuration object
+
+The server configuration object contains these fields:
+
+- `frontendUrl: String` - Set it with env var `WASP_WEB_CLIENT_URL`.
```
sodic, line 31: I've only documented the stuff we mention in other parts of our docs, which is only `frontendUrl`.

We should discuss which env variables we want to expose and can then make an issue for adding that to the docs.

waspc/data/Generator/templates/sdk/wasp/client/test/vitest/helpers.tsx
infomiho: Oh okay, I was thinking along the lines of supporting users to migrate from `'wasp/core/config'` to `wasp/client`. But I guess that would be an exception for our codemod, no? It upgraded users from 0.11 to 0.12 not 0.12 to 0.13!

I agree with what you wrote, no need to go on a complicated refactor to support this one import.

web/docs/advanced/accessing-app-config.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Accessing the configuration 
+---
+
+Whenever you start a Waps app, you are starting two processes.
```
infomiho, line 5:
```suggestion
Whenever you start a Wasp app, you are starting two processes.
```

web/docs/advanced/accessing-app-config.md
```diff
@@ -0,0 +1,59 @@
...
+ - **The client process** - A React app that implements your app's frontend.
```
infomiho, line 6:
```suggestion
 - **The client process** - A React app that implements your app's client.
```

web/docs/advanced/accessing-app-config.md
```diff
@@ -0,0 +1,59 @@
...
+
+- `frontendUrl: String` - Set it with env var `WASP_WEB_CLIENT_URL`.
+
+  The URL of your client (the app's frontend).<br/>
+  Wasp automatically sets it during development when you run `wasp start`.<br/>
+  In production, you should set it to your client's URL as the server sees it
+  (i.e., with the DNS and proxies considered).
+
+You can access it like this:
+```javascript
```
infomiho, line 39: Might I suggest using `js` instead of `javascript` for the language of the code block since we used it everywhere. It might help with future ripgreps :)

web/docs/advanced/accessing-app-config.md
infomiho, line 5: Might I suggest using `app` instead of `process` to keep it a bit less technical i.e. " you are starting two apps"

web/docs/advanced/accessing-app-config.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Accessing the configuration 
+---
+
+Whenever you start a Waps app, you are starting two processes.
+ - **The client process** - A React app that implements your app's frontend.
+
+    During development, this is a dev server with hot reloading. In production,
```
infomiho, line 8: > In production, it's a Node process that serves pre-built static files with environment variables embedded during the build.

This is technically incorrect since for automatic Fly deployments we are using a Docker image with a Go server that serves static files. But for other deployments, the user is free to serve the client in which-ever way they like.

Maybe say: "the static files are serve via some sort of a server" and "this is something you can choose, but usually you'd use something like Netlify or Cloudflare Pages or a simple static files server". I'm just giving you some random ideas here to help with writing.

web/docs/advanced/accessing-app-config.md
sodic: This was intentional.

The bullet starts with "The client process" and then explains it by using the term "frontend", making sure that:
- Everyone finds the terminology they're familiar with.
- We don't self-reference terms in their explanation.
- We connect the frontend terminology with the client before we mention our env var `frontendUrl` (this is why we mention the offical name first and emphasize it with bold text).

web/docs/advanced/accessing-app-config.md
```diff
@@ -0,0 +1,59 @@
...
+
+Wasp gives you runtime access to the processes' configurations through **configuration objects**.
+
+## Server configuration object
+
+The server configuration object contains these fields:
+
+- `frontendUrl: String` - Set it with env var `WASP_WEB_CLIENT_URL`.
+
+  The URL of your client (the app's frontend).<br/>
```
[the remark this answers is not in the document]
sodic: Also intentional, see [here](https://github.com/wasp-lang/wasp/pull/1893#discussion_r1525199716).

web/docs/advanced/accessing-app-config.md
sodic: Are you sure?

The article mainly talks about environment variables. I think processes make more sense at this level of abstraction

web/docs/advanced/accessing-app-config.md
sodic: I can change it, sure, but we definitely shouldn't rely on that when grepping (unless we make a build rule that enforces it).

Btw, I just grepped it, and there are already exceptions to this rule :smile:

web/docs/advanced/accessing-app-config.md
sodic: Thanks, I'll figure something out.

Had no idea about the go server.

infomiho (approved):
LGTM

web/docs/advanced/accessing-app-config.md
infomiho: It sounded very technical for me "we are starting two processes", I was expected to hear the word multi-threading next 😄 That's why I wanted to tone it down a bit to used "app", at least in the intro section. But maybe I'm overreacting, decide what feels the best, I'm okay with either option.

sodic: I decided to go with process because "app" would mean two different things in that sentence: "When you start a Wasp *app*, you are starting two _apps_."

PR #1902, opened 2024-03-18
Bumps Wasp version in docs and example apps

infomiho:
- Bumps Wasp version in examples in docs
- Bumps Wasp version in example apps + fixes issues


web/docs/migrate-from-0-11-to-0-12.md
```diff
@@ -178,9 +178,9 @@ You can easily migrate your old Wasp project to the new structure by following a
 series of steps. Assuming you have a project called `foo` inside the
 directory `foo`, you should:
 
-0. **Install the latest `0.12.x` version** of Wasp.
+0. **Install the `0.12.x` version** of Wasp.
   ```bash
-  curl -sSL https://get.wasp-lang.dev/installer.sh | sh -s
+  curl -sSL https://get.wasp-lang.dev/installer.sh | sh -s -- -v 0.12.4
```
infomiho, line 183: When 0.13 goes out, the command without a specific version would install 0.13.X instead of 0.12.X

Issue #1894, opened 2024-03-14
Document our environment variables


sodic:
We should have a table documenting the names, scopes, and uses of all our environment variables.

A couple of sections only mention them in passing (manual deployment instructions and configuration object docs).

infomiho:
Related to https://github.com/wasp-lang/wasp/issues/1067#issuecomment-1552971969

sodic:
Related to #2053

sodic:
Closed by #2375

PR #1913, opened 2024-03-21
Improved env var docs a bit.

Martinsos:
Inspired by this Discord question: https://discord.com/channels/686873244791210014/1220070515276775516/1220070515276775516 .

### DIFF
--- a/web/docs/advanced/deployment/manually.md
+++ b/web/docs/advanced/deployment/manually.md
@@ -96,6 +96,8 @@ While these are the general instructions on deploying the server anywhere, we al
 
 The command above will build the web client and put it in the `build/` directory in the `.wasp/build/web-app/`.
 
+This is also the moment to provide any additional env vars for the client code, next to `REACT_APP_API_URL`. Check the [env vars docs](../../project/env-vars#client-env-vars-1) for more details.
+
 Since the result of building is just a bunch of static files, you can now deploy your web client to any static hosting provider (e.g. Netlify, Cloudflare, ...) by deploying the contents of `.wasp/build/web-app/build/`.
 
 ### 4. Deploying the Database
@@ -178,7 +180,7 @@ Next, let's copy the `fly.toml` file up to our Wasp project dir for safekeeping.
 cp fly.toml ../../
 ```
 
-Next, let's add a few more environment variables:
+Next, add a few more environment variables for the server code.
 
 ```bash
 flyctl secrets set PORT=8080
--- a/web/docs/project/env-vars.md
+++ b/web/docs/project/env-vars.md
@@ -6,14 +6,14 @@ title: Env Variables
 
 For instance, _during development_, you may want your project to connect to a local development database running on your machine, but _in production_, you may prefer it to connect to the production database. Similarly, in development, you may want to use a test Stripe account, while in production, your app should use a real Stripe account.
 
-While some env vars are required by Wasp, such as the database connection or secrets for social auth, you can also define your env vars for any other useful purposes.
+While some env vars are required by Wasp, such as the database connection or secrets for social auth, you can also define your env vars for any other useful purposes, and then access them in the code.
 
 In Wasp, you can use environment variables in both the client and the server code.
 ## Client Env Vars
 
-Client environment variables are embedded into the client code during the build and shipping process, making them public and readable by anyone. Therefore, you should **never store secrets in them** (such as secret API keys).
+Client environment variables are embedded into the client code during the build and shipping process, making them public and readable by anyone. Therefore, you should **never store secrets in them** (such as secret API keys -> you can provide those to server instead).
 
-To enable Wasp to pick them up, client environment variables must be prefixed with `REACT_APP_`, for example: `REACT_APP_SOME_VAR_NAME=...`.
+To enable Wasp to pick them up, client env vars must be prefixed with `REACT_APP_`, for example: `REACT_APP_SOME_VAR_NAME=...`.
 
 You can read them from the client code like this:
 
@@ -35,11 +35,12 @@ console.log(import.meta.env.REACT_APP_SOME_VAR_NAME)
 
 Check below on how to define them.
 
+
 ## Server Env Vars
 
-In server environment variables, you can store secret values (e.g. secret API keys) since are not publicly readable. You can define them without any special prefix, such as `SOME_VAR_NAME=...`.
+In server environment variables, you can store secret values (e.g. secret API keys) since they are not publicly readable. You can define them without any special prefix, such as `SOME_VAR_NAME=...`.
 
-You can read them in the server code like this:
+You can read the env vars from server code like this:
 <Tabs groupId="js-ts">
 <TabItem value="js" label="JavaScript">
 
@@ -59,7 +60,7 @@ Check below on how to define them.
 
 ## Defining Env Vars in Development
 
-During development, there are two ways to provide env vars to your Wasp project:
+During development (`wasp start`), there are two ways to provide env vars to your Wasp project:
 1. Using `.env` files. **(recommended)**
 2. Using shell. (useful for overrides)
 
@@ -95,41 +96,45 @@ By default, in the `.gitignore` file that comes with a new Wasp app, we ignore a
 ### 2. Using Shell
 If you set environment variables in the shell where you run your Wasp commands (e.g., `wasp start`), Wasp will recognize them.
 
-You can set environment variables in the `.profile` or a similar file, or by defining them at the start of a command:
+You can set environment variables in the `.profile` or a similar file, which will set them permanently, or you can set them temporarily by defining them at the start of a command (`SOME_VAR_NAME=SOMEVALUE wasp start`).
 
-```shell
-SOME_VAR_NAME=SOMEVALUE wasp start
-```
+This is not specific to Wasp and is simply how environment variables can be set in the shell.
 
- This is not specific to Wasp and is simply how environment variables can be set in the shell.
-
-Defining environment variables in this way can be cumbersome even for a single project and even more challenging to manage if you have multiple Wasp projects. Therefore, we do not recommend this as a default method for providing environment variables to Wasp projects. However, it can be useful for occasionally **overriding** specific environment variables because environment variables set this way **take precedence over those defined in `.env` files**.
+Defining environment variables in this way can be cumbersome even for a single project and even more challenging to manage if you have multiple Wasp projects. Therefore, we do not recommend this as a default method for providing environment variables to Wasp projects during development, you should use .env files instead. However, it can be useful for occasionally **overriding** specific environment variables because environment variables set this way **take precedence over those defined in `.env` files**.
 
 ## Defining Env Vars in Production
-
-![Env vars usage in development and production](/img/env/prod_dev_fade_2.svg)
 
 While in development, we had the option of using `.env.client` and `.env.server` files which made it easy to define and manage env vars.
 However, for production, `.env.client` and `.env.server` files will be ignored, and we need to provide env vars differently.
 
+![Env vars usage in development and production](/img/env/prod_dev_fade_2.svg)
+
 ### Client Env Vars
 
-To set client env vars for production, you need to ensure they are set for the terminal session in which you are running the build command, for example:
+Client env vars are embedded into the client code during the build process, making them public and readable by anyone. Therefore, you should **never store secrets in them** (such as secret API keys).
+
+You should provide them to the build command that turns client code into static files:
 ```shell
-REACT_APP_SOME_VAR_NAME=somevalue npm run build
+REACT_APP_SOME_VAR_NAME=somevalue REACT_APP_SOME_OTHER_VAR_NAME=someothervalue npm run build
 ```
 
-These client env vars are then embedded into the client code during the build and shipping process, making them public and readable by anyone. Therefore, you should **never store secrets in them** (such as secret API keys).
+Check the [deployment docs](../advanced/deployment/manually#3-deploying-the-web-client-frontend) for more details.
+
+Also, notice that you can't and shouldn't provide env vars to the client code by setting them on the hosting provider where you deployed them (unlike server env vars, where this is how you should do it). Your client code will ignore those, as at that point client code is just static files.
 
 :::info How it works
-What happens behind the scenes is that Wasp will replace all occurrences of `import.meta.env.REACT_APP_SOME_VAR_NAME` with the value you provided. This is done during the build process, so the value is embedded into the client code.
+What happens behind the scenes is that Wasp will replace all occurrences of `import.meta.env.REACT_APP_SOME_VAR_NAME` in your client code with the env var value you provided. This is done during the build process, so the value is embedded into the static files produced from the client code.
 
 Read more about it in Vite's [docs](https://vitejs.dev/guide/env-and-mode.html#production-replacement).
 :::
 
 ### Server Env Vars
 
-The way you provide env vars to your Wasp project in production depends on where you deploy it. For example, if you deploy your project to [Fly](https://fly.io), you can define them using the `flyctl` CLI tool:
+You can provide env vars to your server code in production by defining them and making them available on the server where your server code is running.
+
+Setting this up will highly depend on where you are deploying your Wasp project, but in general it comes down to defining the env vars via mechanisms that your hosting provider provides.
+
+For example, if you deploy your project to [Fly](https://fly.io), you can define them using the `flyctl` CLI tool:
 
 ```shell
 flyctl secrets set SOME_VAR_NAME=somevalue


web/docs/project/env-vars.md
```diff
@@ -6,14 +6,14 @@ title: Env Variables
...
+While some env vars are required by Wasp, such as the database connection or secrets for social auth, you can also define your env vars for any other useful purposes, and then access them in the code.
 
 In Wasp, you can use environment variables in both the client and the server code.
 ## Client Env Vars
 
-Client environment variables are embedded into the client code during the build and shipping process, making them public and readable by anyone. Therefore, you should **never store secrets in them** (such as secret API keys).
+Client environment variables are embedded into the client code during the build and shipping process, making them public and readable by anyone. Therefore, you should **never store secrets in them** (such as secret API keys -> you can provide those to server instead).
 
-To enable Wasp to pick them up, client environment variables must be prefixed with `REACT_APP_`, for example: `REACT_APP_SOME_VAR_NAME=...`.
+To enable Wasp to pick them up, client env vars must be prefixed with `REACT_APP_`, for example: `REACT_APP_SOME_VAR_NAME=...`.
```
sodic, line 16: Why tho? :smile: 

You probably did it because you feel that `env vars` has become such a universal acronym that it's reached a point where it's more recognizable than the full name. I disagree, but fair enough.

I don't have a strong preference either way.

Still, we should then probably update all occurrences in the text to be consistent.

web/docs/advanced/deployment/manually.md
```diff
@@ -96,6 +96,8 @@ While these are the general instructions on deploying the server anywhere, we al
 
 The command above will build the web client and put it in the `build/` directory in the `web-app` directory.
 
+This is also the moment to provide any additional env vars for the client code, next to `REACT_APP_API_URL`. Check [env vars docs](../../project/env-vars#client-env-vars-1) for more details.
```
sodic, line 99:
```suggestion
Use the build command to provide any additional env vars for the client code (other than `REACT_APP_API_URL`). Check the [env vars docs](../../project/env-vars#client-env-vars-1) for more details.
```
Reasonings:
- "This is" -> "Use the build command" - Better to call things by their name than relying on previous context.
- "The" - I guess this one is pretty clear on its own. :smile: 
- The parentheses I can't explain but looks nicer to me.

web/docs/advanced/deployment/manually.md
```diff
@@ -178,7 +180,7 @@ Next, let's copy the `fly.toml` file up to our Wasp project dir for safekeeping.
 cp fly.toml ../../
 ```
 
-Next, let's add a few more environment variables:
+Next, let's add a few more environment variables that will be available to the server code:
```
sodic, line 183: It's better to stick to the present tense and addressing the user directly:
```suggestion
Next, add a few more environment variables for the server code.
```

web/docs/project/env-vars.md
```diff
@@ -6,14 +6,14 @@ title: Env Variables
 
 For instance, _during development_, you may want your project to connect to a local development database running on your machine, but _in production_, you may prefer it to connect to the production database. Similarly, in development, you may want to use a test Stripe account, while in production, your app should use a real Stripe account.
 
-While some env vars are required by Wasp, such as the database connection or secrets for social auth, you can also define your env vars for any other useful purposes.
+While some env vars are required by Wasp, such as the database connection or secrets for social auth, you can also define your env vars for any other useful purposes, and then access them in the code.
```
sodic, line 9: Let's use this chance to get rid of the passive form.
Also, the sentence is a little long and can be broken down into simpler sentences:
```suggestion
Wasp requires you to set certain env vars (e.g., the database connection string, the secrets for social auth).
You can also set your env vars with anything you might need to access from the code (e.g., list some examples for this)
```

web/docs/project/env-vars.md
```diff
@@ -6,14 +6,14 @@ title: Env Variables
...
 
 In Wasp, you can use environment variables in both the client and the server code.
 ## Client Env Vars
 
-Client environment variables are embedded into the client code during the build and shipping process, making them public and readable by anyone. Therefore, you should **never store secrets in them** (such as secret API keys).
+Client environment variables are embedded into the client code during the build and shipping process, making them public and readable by anyone. Therefore, you should **never store secrets in them** (such as secret API keys -> you can provide those to server instead).
```
sodic, line 14: Let's also get rid of the passive form (are embedded -> Wasp embeds).

I'm not a fan of using parentheses and an arrow here - it's a little weird, and I feel it could be it's own sentence:
```
For example, you shouldn't provide your API keys as client env vars. Instead, set them on the server.
```
Btw, I'm not sure this makes sense. The API keys will still be visible by anyone who accesses the web page (not storing them in the env file will just prevent them from getting into version control).

web/docs/project/env-vars.md
```diff
@@ -35,11 +35,12 @@ console.log(import.meta.env.REACT_APP_SOME_VAR_NAME)
...
 Check below on how to define them.
 
+
 ## Server Env Vars
 
-In server environment variables, you can store secret values (e.g. secret API keys) since are not publicly readable. You can define them without any special prefix, such as `SOME_VAR_NAME=...`.
+In server environment variables, you can store secret values (e.g. secret API keys) since they are not publicly readable. You can define them without any special prefix, such as `SOME_VAR_NAME=...`.
 
-You can read them in the server code like this:
+You can read them from the server code like this:
```
sodic, line 43:
```suggestion
You can read the env vars from server code like this:
```

web/docs/project/env-vars.md
```diff
@@ -92,40 +93,44 @@ These files should not be committed to version control, and they are already ign
 ### 2. Using Shell
 If you set environment variables in the shell where you run your Wasp commands (e.g., `wasp start`), Wasp will recognize them.
 
-You can set environment variables in the `.profile` or a similar file, or by defining them at the start of a command:
-
-```shell
-SOME_VAR_NAME=SOMEVALUE wasp start
-```
+You can set environment variables in the `.profile` or a similar file, which will set them permanently, or you can set them temporarily by defining them at the start of a command (`SOME_VAR_NAME=SOMEVALUE wasp start`).
```
sodic, line 96:
```suggestion
One option is setting your environment variables in `.profile` or a similar file, which will set them permanently,.
Another option is setting your environment variables temporarily by defining them at the start of a command (e.g., `SOME_VAR_NAME=SOMEVALUE wasp start`).
```

web/docs/project/env-vars.md
```diff
@@ -92,40 +93,44 @@ These files should not be committed to version control, and they are already ign
...
 
- This is not specific to Wasp and is simply how environment variables can be set in the shell.
+This is not specific to Wasp and is simply how environment variables can be set in the shell.
 
-Defining environment variables in this way can be cumbersome even for a single project and even more challenging to manage if you have multiple Wasp projects. Therefore, we do not recommend this as a default method for providing environment variables to Wasp projects. However, it can be useful for occasionally **overriding** specific environment variables because environment variables set this way **take precedence over those defined in `.env` files**.
+Defining environment variables in this way can be cumbersome even for a single project and even more challenging to manage if you have multiple Wasp projects. Therefore, we do not recommend this as a default method for providing environment variables to Wasp projects during development, you should use .env files instead. However, it can be useful for occasionally **overriding** specific environment variables because environment variables set this way **take precedence over those defined in `.env` files**.
```
sodic, line 100:
```suggestion
Defining environment variables inline can be cumbersome even for a single project and even more challenging to manage if you have multiple Wasp projects.
Therefore, we do not recommend this as a default method for providing environment variables to Wasp projects during development. Instead, use the `.env` files instead.
Inline definitions are useful for occasionally **overriding** specific environment variables because environment variables set this way **take precedence over those defined in `.env` files** (list example perhaps?)
```

web/docs/project/env-vars.md
```diff
@@ -92,40 +93,44 @@ These files should not be committed to version control, and they are already ign
...
 
 ## Defining Env Vars in Production
 
-While in development, we had the option of using `.env` files which made it easy to define and manage env vars. However, in production, we need to provide env vars differently.
+While in development, we had the option of using `.env` files which made it easy to define and manage env vars. However, in production, when our code is built and deployed, we need to provide env vars differently.
```
sodic, line 104: We should change "we" to "you" to address the user directly.

web/docs/project/env-vars.md
```diff
@@ -92,40 +93,44 @@ These files should not be committed to version control, and they are already ign
...
 
 ![Env vars usage in development and production](/img/env/prod_dev_fade_2.svg)
 
 ### Client Env Vars
 
-Client env vars are embedded into the client code during the build and shipping process, making them public and readable by anyone. Therefore, you should **never store secrets in them** (such as secret API keys).
+Client env vars are embedded into the client code during the build process, making them public and readable by anyone. Therefore, you should **never store secrets in them** (such as secret API keys).
```
sodic, line 110: Again, the API key is I believe an incorrect example (as it's visible by anyone once the site is deployed).

web/docs/project/env-vars.md
```diff
@@ -92,40 +93,44 @@ These files should not be committed to version control, and they are already ign
...
 
-You should provide them to the build command, for example:
+You should provide them to the build command that turns client code into static files:
 ```shell
-REACT_APP_SOME_VAR_NAME=somevalue npm run build
+REACT_APP_SOME_VAR_NAME=somevalue REACT_APP_SOME_OTHER_VAR_NAME=someothervalue npm run build
 ```
 
+Notice that the exact moment when you should provide these env vars is during the `npm run build` command that you execute from `.wasp/build/web-app/` after you have ran `wasp build`. Check [deployment docs](../advanced/deployment/manually#3-deploying-the-web-client-frontend) for more details.
```
sodic, line 117: I find the first sentence a little too long (many filler words). Can you simplify it somehow? 

Also, the deployment docs are missing a "the." I see that Grammarly is bad at catching missing articles before links because of the bracket - something to keep in mind.

web/docs/project/env-vars.md
```diff
@@ -92,40 +93,44 @@ These files should not be committed to version control, and they are already ign
...
+
+Also, notice that you can't and shouldn't provide env vars to the client code by setting them on the hosting provider where you deployed them (unlike server env vars, where this is how we do it). Your client code will ignore those, as at that point client code is just static files.
```
sodic, line 119: All these explanations are great and understandable, but I'm wondering if maybe we're explaining too much.

People should already probably know this stuff if they're deploying a web app.

web/docs/project/env-vars.md
```diff
@@ -92,40 +93,44 @@ These files should not be committed to version control, and they are already ign
...
-What happens behind the scenes is that Wasp will replace all occurrences of `import.meta.env.REACT_APP_SOME_VAR_NAME` with the value you provided. This is done during the build process, so the value is embedded into the client code.
+What happens behind the scenes is that Wasp will replace all occurrences of `import.meta.env.REACT_APP_SOME_VAR_NAME` in your client code with the env var value you provided. This is done during the build process, so the value is embedded into the static files produced from the client code.
 
 Read more about it in Vite's [docs](https://vitejs.dev/guide/env-and-mode.html#production-replacement).
 :::
 
 ### Server Env Vars
 
-The way you provide env vars to your Wasp project in production depends on where you deploy it. For example, if you deploy your project to [Fly](https://fly.io), you can define them using the `flyctl` CLI tool:
+To provide env vars to your server code in production you will need to have them defined and available on the server where your server code is running.
```
sodic, line 129: Can we change this to the present tense?

web/docs/project/env-vars.md
```diff
@@ -92,40 +93,44 @@ These files should not be committed to version control, and they are already ign
...
+
+Setting this up will highly depend on where you are deploying your Wasp project, but in general it comes down to defining the env vars via mechanisms that your hosting provider provides.
```
sodic, line 131: I recommend splitting into two sentences.

sodic (commented):
Nice work, I'm glad we're improving this!

How does the PR relate to these two issues:
 - https://github.com/wasp-lang/wasp/issues/1894
 - https://github.com/wasp-lang/wasp/issues/2053

Can we take care of those as well?

Martinsos:
> Nice work, I'm glad we're improving this!
> 
> How does the PR relate to these two issues:
> 
> * [Document our environment variables #1894](https://github.com/wasp-lang/wasp/issues/1894)
> * [Make it clear in docs which env vars and how users need to set when deploying to production #2053](https://github.com/wasp-lang/wasp/issues/2053)
> 
> Can we take care of those as well?

For the context, this is 3 months old PR what was adding quite specific bits to the docs regarding env vars that were missing, which was made obviuos to me from the Discord convo I linked. Somehow we forgot about this PR, and I realized now it is still open, so thought let's merge it.

That's a good point regardin g the issues you linked, and there is a general initiative to improve the env vars situation, but from taking a quick look at those issues, it is clear to me they are going quite wider in their scope. So I wouldn't tackle those now as they are not trivial, this PR is a useful incremental change on its own, and this bigger issues we can independently do later.

web/docs/project/env-vars.md
Martinsos: Can't remember exactly, I think I felt "environemnt variables" was a mouthful here and this was in the part of docs where the naming "env vars" was already established, so kind of as you said.

We could update all the occurences, but they will again diverge at some point, since we don't have an automatic mechanism to check it or enforce it. So I wouldn't bother with it.

web/docs/advanced/deployment/manually.md
Martinsos: - I missed the "the", thanks.
- "This is" -> "Use the build command": not exactly the same, I was referring to the moment at which we are currently ("this is also the moment"), not the build command, so you somewhat changed the semantics of the sentence.

I don't really mind small changes like this, why not, but this one kind of breaks the flow for me -> if you start reading a bit up and read from there, before it was reading ok to me: "Also, this is the moment to provide any additional env vars" -> exactly what I wanted to say. But now, it is "Use the build command" -> a bit weird, starts out of nowhere, also "use" the "command" is slightly weird -> ok, not terribly, but it doesn't emphasize that at this very moment they need to specify client env vars, this is it.

I can add "the", sure, but the rest seems to not be communicating what I wanted to commnicate.

web/docs/project/env-vars.md
Martinsos: I get that we don't like passive, but you also removed the "while", therefore changing the flow of the sentence for the worse in my opinion. Also, there might be some other purpose for providing env vars than for accessing them in the code, so you narrowed that down too much.

I wouldn't try to fix the current documentation at this moment, let's do that in another PR if you wish. This is 3 months old PR with relatively small additions, that was created to quickly address the issue user -> we can merge that and get that value, while avoiding the discussion that will now arise from additional trying to improve the docs near by it.

web/docs/project/env-vars.md
Martinsos: Same as I said above, I would not try to improve the whole pieces of docs here in this PR, but would just focus on adding neccessary bits that I thoguht are important based on the user's experience. We can make a separate effort of improving the docs flow and sentencing, we will have better results if we do it as a whole, then if we attack only random bits that were touched in this PR, missing context.

I don't see much improvement with the sentence you provided. Also, you dropped word "secret", which is crucial -> they can provide public API keys.

I don't get what doesn't make sense to you? We are not mentioning env files at all here, how did you get to those?

web/docs/project/env-vars.md
Martinsos: How is that better? Didn't you just go passive, while it is better to speak directly to the user?

web/docs/project/env-vars.md
Martinsos: While "in this way" refers to the context above, which is not ideal, what you did with "inline" is likely worse, as it is a new concept we haven't mentioned so far and you haven't explained it here, so it is not clear to what you are referring.

But similar like above, I wouldn't go into refactoring the whole paragraphs here when we just added one small piece of the sentence, let's leave that for another PR.

web/docs/project/env-vars.md
Martinsos: Let's do it for the whole docuemnt then in another PR.

web/docs/project/env-vars.md
Martinsos: I don't get what you mean. We are telling them they shouldn't provide secrets as clienv env vars, specifically taking an example of secret API keys. I am not sure how that refers to what you said.

web/docs/project/env-vars.md
Martinsos: It's what confused multiple people in Discord, and I myself wasn't 100% sure about the process, so I would love if I have had this in the docs. There is a big difference between writing an app and deploying it, those are different knowledge sets although related, plus Wasp makes a lot of things easier, allowing even beginners to implement their apps, so we can expect some of them don't have a lot of experience in deploying them. Also, Fly for example offers place to set env vars for the client server, so that allows for extra confusion.

web/docs/project/env-vars.md
Martinsos: I dropeed the first part, might not be that important! Can't remember exactly since it was some time since I wrote this, but if it will be missing we can add it again. I added missing "the".

web/docs/project/env-vars.md
Martinsos: Why is that? Reads better to me like this.

Martinsos (commented):
Did another iteration!

I accepted some of the changes, but some of them, that are about generally improving the pieces of docs we touched, I would propose not doing in this PR. It is 3 months old PR with specific small changes, I don't have all the context I had then and I would rather not get entangled into the rest of the docs here. We can have a separate effort where we rehaul the whole file to make it be closer to our writing guide. I did though make sure that additions do stick to writing guide.

web/docs/project/env-vars.md
Martinsos: Done!

web/docs/project/env-vars.md
sodic: > We can make a separate effort of improving the docs flow and sentencing, we will have better results if we do it as a whole, then if we attack only random bits that were touched in this PR, missing context.

Ok, makes sense.

> I don't get what doesn't make sense to you?

I get it know, you're talking about some data that the client doesn't need at all (it's only required by the server). But why would they then pass it to the client in the first place if the client doesn't need them?

web/docs/project/env-vars.md
sodic: "One option is setting" isn't passive form. It's an active sentence with a different subject.

I believe the whole thing is better because:
- The two sentences are shorter and simpler than the original longer sentence.
- The two sentences follow the same structure, contrasting the differences.
- The subjects of the two sentences also emphasize the two possibilities ("one option is", "another option is"). It's always best to use the most relevant actor in the sentence as the subject. In this case, that's "the way of setting environment variables"

Btw, I'm not hung up on every word I suggest in these changes. I just add them as examples (maybe I shouldn't be using the suggestion box). You can pick and apply what you think makes sense.

The core idea I care about here is shorter and simpler sentences. The rest of the stuff is just how I'd approach the core idea, but is not set in stone.

P.S. whatever you choose to do with this, there's an extra "the" before ".profile" you should remove. It may have slipped unnoticed.

web/docs/project/env-vars.md
sodic: I'm resolving this because the same discussion is happening somewhere else.

web/docs/project/env-vars.md
sodic: Makes sense, I'm probably forgetting my past struggles with this stuff.

You can resolve when you see this.

web/docs/project/env-vars.md
sodic: I had to read it multiple times to realize what's going on.

I don't know what reads better (because I'm not sure what exactly you're comparing it to), I just know I found this particular form to complex for my taste.

But, even for the simplest change (change "but" to a full stop"), I think it's an improvement:
```
Setting this up will highly depend on where you are deploying your Wasp project.
In general, it comes down to defining the env vars via mechanisms that your hosting provider provides.
```
I believe simpler and shorter sentences are almost always easier to read (since they give you a "checkpoint"). This is especially true for non-native speakers.

sodic (approved):
I'm approving. You can merge this when you're satisfied.

Do what you want with all remaining unresolved comments (and resolve them when you're happy). None of them are important enough to take up more time.

Btw, I forgot this is three months old. If I had remembered, I wouldn't leave that many comments (since the changes have been out of your context for a while, and aren't that important)

web/docs/project/env-vars.md
Martinsos: So it is actually a quite common question/mistake we get in Discord: I have let's say Sendgrid API secret, can I put it in the client? So I don't have to bother with the server. I think you know quite a bit about it so it seems this is obvious, but you would be surprised.

Aha you say that client doesn't need them -> if they decide to use some third party service, let's say S3, it is their choice how to impelemnt it in Wasp, and they might easily try to do it all in client. With our features it is clear where env var goes, but if they are doing it alone like this, then they can more easily make a mistake.

web/docs/project/env-vars.md
Martinsos: This is what I find tricky with these docs discussion: you say it sounds better to you, but to me original sounds better. I can read it in one piece with clear contract, versus (to me) unnatural break you went for in order to emphasize it is two options. And again it goes against the recommendation to talk to them directly, which you use as a quite strong argument in different places, but now you drop that.

Seems to me like for most of the way docs are written one can go and make a counter arugment: either splitting something, or combining something, or swtiching to "you", or switching away form it, all for the argument of "it is simpler / reads better this way", while it is really quite subjective and probalby has 0 impact on the total mass of readers who are also all unique people.

I don't think we shouldn't comment on docs in PRs at all, we should, and there are things in writing guide we should be looking to enforce indeed, but I think there is a line here, in the sense that if a suggestion is not clearly an objective improvement, it is probably better not make one. Or while making it, say very clearly it is just an alternative idea and author can do whatever they like, so they can take a look at it, take what they lilke, and move on, wihtout having to discuss it.

Martinsos:
> I'm approving. You can merge this when you're satisfied.
> 
> Do what you want with all remaining unresolved comments (and resolve them when you're happy). None of them are important enough to take up more time.
> 
> Btw, I forgot this is three months old. If I had remembered, I wouldn't leave that many comments (since the changes have been out of your context for a while, and aren't that important)

I forgot to mention it is that old! That would have helped.
Ok, I took care of the obvious improvements and resolved the rest as I saw fit.
I will be merging it once I resolve the conflicts.

web/docs/project/env-vars.md
sodic: I agree with the core idea behind your comment: _Everyone can choose their style, and we should emphasize when which suggestions are subjective alternatives._

So I'll keep that in mind for the future.

I do have one thing to clear up.

This part here:

> And again it goes against the recommendation to talk to them directly, which you use as a quite strong argument in different places, but now you drop that.

Isn't right. 

To avoid future confusion, I'll elaborate on "addressing the user directly" (and we can change the style guide to communicate it more clearly).

_Addressing the user directly_ means _write as if you were speaking to someone_, for example:
- "You can" instead of "We can."
- "Do this" instead of "Let's do this."
- "You can" instead of "The user can."
- "You should add a query" instead of "A query should be added." 

_Addressing the user directly_ doesn't mean _You must write every sentence in the second person, and every subject must be "you" (the user)._
That would be very impractical. Imagine if you were talking to me with the constraint that every sentence must start with "You." It's perfectly normal to use subjects that aren't the person you're speaking to (and we do it all the time, as we should).

So, the change I suggested here doesn't at all contradict the recommendation of addressing the user directly.

Anyway, here's what the writing guide currently says:
> **Address the reader directly.** Instead of "We can implement a Query..." or "A Query can be implemented..." or "The user can implement a Query...", say "You can implement a Query..." The docs should speak to the reader and address them directly to avoid possible confusion (e.g., Who are we? The Wasp team? The reader together with the Wasp team? etc.). Use the word user only to refer to the user of the software that your reader is developing. You can sometimes use first-person plural pronouns to refer to the organization (Wasp). For example, "We support both TypeScript and JavaScript" is OK, but "Wasp supports both TypeScript and JavaScript" is usually better.

How do you think we can improve it to make it less confusing?

Martinsos: I get the idea, but based on this, to me it sounds like we should do "You can do thing X" instead of "One option is to do thing X". Shorter, speaks direct to the user. As writing guide suggests as the best way to phrase stuff. Would you make it longer + loose the direct talking to the user?

Issue #1800, opened 2024-02-22
getUsername and getEmail methods are confusing a bit


Martinsos:
From their name, one would think they are able to obtain the username/email for user independent of which auth method they use, but the truth is, getUsername works only if username auth method is used, while getEmail works only if email&pass is used.

Both @vincanger and I got confused and thought at first it works for more cases, in the sense that it extract username from google auth information if needed, or githubs info, or whatever is user authenticating with. The problem is, there is nothing to stop you from thinking wrong, type system can't catch this, so this could be the source of errors later in the runtime.

We could either make these methods more general, so they actually do work in a way that they go through all the AuthIdentities user has an try to find username in each of them until it finds it, or we should somehow make it clear these work only for their respective methods. We could e.g. rename them to `getUsernameFromUsernameAndPassAuth` and `getEmailFromEmailAndPassAuth`.

sodic:
The final consensus was that it makes sense to turn these into methods (or even properties) on the `AuthUser` object. Instead of having to import `getUsername` and call it on a `user` object, we could just do:
```javascript
user.getUsername()

// or even

user.username // calls a getter
```

Martinsos:
Case of a user getting confused with this: https://discord.com/channels/686873244791210014/1262018188867010642/1262018190289141871 .

Martinsos:
Another one somewhat related: https://discord.com/channels/686873244791210014/1262012864999194635/1262012866995683431

Martinsos:
Also related: https://discord.com/channels/686873244791210014/1260975468979552340/1260975468979552340 .

PR #1915, opened 2024-03-21
Update AuthUser to offer a better DX

infomiho:
> [!NOTE] 
> This is probably need to be a breaking change to make the API clean.


waspc/data/Generator/templates/sdk/wasp/server/_types/index.ts
```diff
@@ -88,20 +87,5 @@ type Context<Entities extends _Entity[]> = Expand<{
 {=# isAuthEnabled =}
 type ContextWithUser<Entities extends _Entity[]> = Expand<Context<Entities> & { user?: AuthUser }>
 
-// TODO: This type must match the logic in auth/session.js (if we remove the
```
infomiho, line 91: This TODO is now resolved yay! Since the type and the runtime value are based on the same logic.

waspc/src/Wasp/Generator/AuthProviders.hs
```diff
@@ -43,3 +46,13 @@ emailAuthProvider =
     { E._providerId = fromJust $ makeProviderId "email",
       E._displayName = "Email and password"
     }
+
+getEnabledAuthProvidersJson :: AS.Auth.Auth -> Aeson.Value
```
infomiho, line 50: Generating this JSON in one place instead of three.

waspc/data/Generator/templates/sdk/wasp/auth/user.ts
```diff
@@ -20,8 +27,78 @@ export function getFirstProviderUserId(user?: AuthUser): string | null {
...
+// PUBLIC API
+export type AuthUser = ReturnType<typeof createAuthUser>
+
+// PRIVATE API
+export function createAuthUser(
+  user: UserEntityWithAuth
+) {
+  const { auth, ...rest } = user
+  const identities = {
+    {=# enabledProviders.isEmailAuthEnabled =}
```
infomiho, line 45: Generating fields based on the enabled auth methods.

waspc/examples/todoApp/src/apis.ts
```diff
@@ -1,10 +1,9 @@
-import { getFirstProviderUserId } from "wasp/auth";
-import { type MiddlewareConfigFn } from "wasp/server";
-import { type BarBaz, type FooBar, type WebhookCallback } from "wasp/server/api";
+import { type MiddlewareConfigFn } from 'wasp/server'
+import { type BarBaz, type FooBar, type WebhookCallback } from 'wasp/server/api'
 import express from 'express'
 
 export const fooBar: FooBar = (_req, res, context) => {
-  const username = getFirstProviderUserId(context?.user) ?? 'Anonymous'
+  const username = context.user?.getFirstProviderUserId()
```
Martinsos, line 6: Ok cool this is nicer now!
What does `getFirstProviderUserId` promise? If I have multiple providers, it will return id from one of them. And it will be consistent in the order, right? What if I add a new provider? Maybe this ID will become different, right? That might be ok,I am just curious.

waspc/examples/todoApp/src/pages/ProfilePage.tsx
```diff
@@ -41,15 +45,13 @@ export const ProfilePage = ({ user }: { user: User }) => {
...
-  const providerData = getProviderData(user)
-
   return (
     <>
       <h2>Profile page</h2>
       <div>
         Hello <strong>{getName(user)}</strong>! Your status is{' '}
         <strong>
-          {providerData && providerData.isEmailVerified
+          {user.identities.email && user.identities.email.data.isEmailVerified
```
Martinsos, line 54: I commented somewher else I remember, I think `user.identities.email` is a bit misleading because it reads as if that is email, but it is not.
I wonder if `user.identities.emailAndPass` would be nicer therefore, I think it would. But, I guess that would need data migrations? Maybe not? Or maybe it wouldn't be consistent with the rest of the naming we use, if we use just "email" everywhere for "email and pass"?

Martinsos (commented):
Took a quick look, looks good in general, left some comments!
But I will let @sodic do a proper review.

How would you describe the crux of this PR regarding its impact on the user? `getFirstProviderId` is now method on `user`. You also have `createAuthUser` which helps with something from what I saw. Something else? This is only part of the Auth DX improvements right?

waspc/examples/todoApp/src/apis.ts
infomiho: It will return the first available provider :D the order might change with if you add a new provider.

I predict that this method will be mostly used by people who only have one provider anyways or just want to display _something_ identifying the user. For proper "production usage" I suspect people will want to put some sort of an "email" or "username" on the `User` object anyways.

waspc/examples/todoApp/src/pages/ProfilePage.tsx
infomiho: We use `email` everywhere right now. We were pretty consistent with that. 

For all other providers we followed the same logic, put the provider name as the key in the `identities` object.

waspc/examples/todoApp/src/user.ts
```diff
@@ -1,42 +1,27 @@
...
-    return getEmail(user)
+  if (user.identities.email !== null) {
+    return user.identities.email.id
   }
 
-  const googleIdentity = findUserIdentity(user, 'google')
-  if (googleIdentity !== undefined) {
-    return `Google user ${googleIdentity.providerUserId}`
+  if (user.identities.google !== null) {
+    return `Google user ${user.identities.google.id}`
```
sodic, line 14: Much better!

sodic (commented):
I'm a big fan of this change!

Left a couple of comments to discuss.

waspc/examples/todoApp/src/apis.ts
Martinsos: Yeah, I agree! We should make sure to mention this in the docs of this function, to make it clear how it is intended to be used, and what the pitfalls might be, and why they should therefore potentially put stuff on `User` instead. I don't think this stuff will be clear to a new Wasp user.

waspc/examples/todoApp/src/pages/ProfilePage.tsx
Martinsos: @infomiho it is not about the position of `email`, it makes that it is in `identities`, it is that it would be clearer what it is if it was called `emailAndPass`.
Same probably stands for `username`, if there is also `identities.username`.

Maybe it is worth changing here, even if we `email` in imports? In imports i also guess context is a bit different, and they are more autmoatized, and nobody reads them directly.

waspc/examples/todoApp/src/dbSeeds.ts
```diff
@@ -34,14 +35,7 @@ async function createUser(prismaClient: PrismaClient, data: any) {
...
 
-  if (newUser.auth?.identities) {
-    newUser.auth.identities = newUser.auth.identities.map((identity) => {
-      identity.providerData = JSON.parse(identity.providerData)
-      return identity
-    })
-  }
-
-  return newUser
+  return createAuthUser(newUser as any)
```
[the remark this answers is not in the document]
infomiho: I wanted to get this code to work, I wanted to make it work without an assertion but it proved difficult. I'll give it an another try.

waspc/examples/todoApp/src/webSocket.ts
```diff
@@ -8,7 +7,7 @@ export const webSocketFn: WebSocketDefinition<
   InterServerEvents
 > = (io, context) => {
   io.on('connection', (socket) => {
-    const username = getFirstProviderUserId(socket.data.user) ?? 'Unknown'
+    const username = socket.data.user?.getFirstProviderUserId() ?? 'Unknown'
```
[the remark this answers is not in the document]
infomiho: I was using a rule of thumb: if it's a property that they would be able to get just by going through the raw DB entity, I put a getter. If it's something special like getting the first available provider, I used a function.

The options seem to be:
- `user.getIdentities().email`
- `user.identities.email`
- `user.firstProviderId`
- `user.getFirstProviderId()`

The combination felt okay to me, as `user.firstProviderId` feels off since the jobs of getting the first provider id feels dynamic.

waspc/examples/todoApp/src/pages/ProfilePage.tsx
Martinsos: Aha! Oh, ok if that is so, then indeed it would be nicest if we went with `emailAndPassword` everywhere. Maybe better to do that now than later.
But we don't have to do it in this PR if you think it is too much, we can do it in another PR? But let's try to do it soon after this one?
What is your feeling about this, does it make sense to do this change? Sounds nice to me, like we have nice / consistent naming, but I think you are a bit better suited to judge if it is worthy the effort.

infomiho:
Docs are ready for review. I've tried to rewrite them to fit the new API and explain some extra stuff.

web/docs/auth/entities/entities.md
```diff
@@ -1,22 +1,454 @@
...
+</TabItem>
+</Tabs>
+
+If you have some **piece of the auth data that you want to access frequently** (for example the `username`), it's best to store it at the top level of the `User` entity.
+
+For example, save the `username` or `email` as a property on the `User` and you'll be able to access it without fetching the auth-related fields. We show an example in the [Defining Extra Fields on the User Entity](../overview.md#1-defining-extra-fields) section of the docs.
+
+### Getting the User's Data
+
+If you fetch the user data like this, you'll lose the benefits of the `AuthUser` object. You'll need to access the auth data by going through the `auth` and `identities` relations:
```
Martinsos, line 261: What does "like this" stand for here? I would try to make it more precise.

Maybe "When you have `user` object with `user.auth.identities` field, it can be a bit tedious to obtain the auth data (like username or google id) from it.

web/docs/auth/entities/entities.md
```diff
@@ -1,22 +1,454 @@
...
+```
+
+</TabItem>
+</Tabs>
+
+If you have some **piece of the auth data that you want to access frequently** (for example the `username`), it's best to store it at the top level of the `User` entity.
+
+For example, save the `username` or `email` as a property on the `User` and you'll be able to access it without fetching the auth-related fields. We show an example in the [Defining Extra Fields on the User Entity](../overview.md#1-defining-extra-fields) section of the docs.
+
+### Getting the User's Data
```
Martinsos, line 259: Title seems a bit vague

web/docs/auth/entities/entities.md
```diff
@@ -1,22 +1,454 @@
...
+```
+
+</TabItem>
+</Tabs>
+
+Wasp offers a few helper methods to access the user's data when you retrieve the `user` like this. They are `getUsername`, `getEmail` and `getFirstProviderUserId`. They can be used both on the client and the server.
+
+#### `getUsername`
+
+It accepts the `user` object and returns the username of the user if it exists or `null` if it doesn't. The `user` object needs to have the `auth` and the `identities` relations included.
```
Martinsos, line 306: If I am correct this returns username ONLY as defined by `auth.identities.username`, right? It won't get username from `user.username` if user defined it there on their own, for sure. It also won't get username from their Google signing, or any other auth method.
While this might be clear to us, I don't think this would be clear to the first time reader, so I think we should make this very clear.
Although we do have a separate issue for this here https://github.com/wasp-lang/wasp/issues/1800, so maybe that is ok and we will just fix it there? Same goes for getEmail.

web/docs/auth/entities/entities.md
```diff
@@ -1,22 +1,454 @@
...
+      ))}
+    </div>
+  )
+}
+```
+
+</TabItem>
+</Tabs>
+
+#### `getFirstProviderUserId`
```
Martinsos, line 398: We already mentioned this method somewhere above in the docs, right? How is it we are mentioning it again? That can be ok, just curios.

web/docs/migrate-from-0-13-to-0-14.md
```diff
@@ -0,0 +1,260 @@
...
+
+This guide only covers the migration from **0.13.X to 0.14.X**. If you are migrating from 0.11.X or earlier, please read the [migration guide from 0.11.X to 0.12.X](./migrate-from-0-11-to-0-12.md) first.
+
+:::
+
+## What's new in 0.14.0?
+
+### Better auth user API
+
+We introduced a much simpler API for accessing user auth fields like `username`, `email` or `isEmailVerified` on the `user` object. You don't need to helper functions every time you want to access the user's `username` or extra steps to get proper typing.
```
Martinsos, line 15: "you don't need to helper ..." -> missing a word: maybe "need to use"?

web/docs/migrate-from-0-13-to-0-14.md
```diff
@@ -0,0 +1,260 @@
...
+
+## How to migrate?
+
+### Migrate how you access user auth fields
+
+We had to make a couple of breaking changes to reach the new simpler API. 
+
+Follow the steps below to migrate:
+
+1. **Replace `getUsername` use** with `user.identities.username.id`
```
Martinsos, line 25: use -> usage?

Martinsos: Btw why is this so? Don't we still have `getUsername` helper function?

web/docs/migrate-from-0-13-to-0-14.md
```diff
@@ -0,0 +1,260 @@
...
+    ```ts title=src/tasks.ts
+    export const createTask: CreateTask<...>  = async (args, context) => {
+        const email = context.user.identities.email?.id
+        // ...
+    }
+    ```
+    </TabItem>
+    </Tabs>
+
+1. **Replace any other auth field usage** with `user.identities.<provider>.<value>`
```
Martinsos, line 116: I dind't quite get this. What got deprecated? `findUserIdentity`? What else? Any way to phrase this in a bit simpler way? Maybe I am just missing context and would understand this if actually used this older code?

web/docs/migrate-from-0-13-to-0-14.md
```diff
@@ -0,0 +1,260 @@
...
+    const MainPage = ({ user }: { user: AuthUser }) => {
+        // The email object is properly typed, so we can access `isEmailVerified` directly
+        const isEmailVerified = user.identities.email?.isEmailVerified
+        // ...
+    }
+    ```
+    </TabItem>
+    </Tabs>
+
+1. **Use method access for `getFirstProviderUserId`**
```
Martinsos, line 157: A bit weird title, feels like it is missing something grammatically?

web/docs/migrate-from-0-13-to-0-14.md
```diff
@@ -0,0 +1,260 @@
...
+    ```ts title=src/tasks.ts
+    export const createTask: CreateTask<...>  = async (args, context) => {
+        const userId = user.getFirstProviderUserId()
+        // ...
+    }
+    ```
+    </TabItem>
+    </Tabs>
+
+1. **Replace `findUserIdentity`** with simple checks on `user.identities.<provider>`
```
Martinsos, line 205: Feels like this was already covered two steps above?

web/docs/migrate-from-0-13-to-0-14.md
Martinsos, line 1: I am still a bit like "hm, I wish we would not eery time create a migration doc but instead keep things in Changelog if possible".
But I guess you felt there is enough complexity here to go for a new migration page.
Still, can you also add something to the Changelog? A short description of what happened + link to the migration docs at least.

Martinsos (commented):
@infomiho I gave it a look, looks fine in general! Seems like there was no big structural changes, you mostly covered the change in the API + added the migration instructions?
It can be a bit hard to review docs sometimes, even harder than code, because in code I can really on Typescript / Haskell to know that stuff is vibing together, and I can somehow easier get a bigger picture, in docs it is a bit harder to figure out what was supposed to change, what is the context of the rest of the docs, ... . What might be helpful when doing docs change like this is if you wrote a tiny bit about what you changed, from the very high level, to give me an idea of what I am looking at, what should I focus on.

web/docs/auth/_accessing-user-data-note.md
```diff
@@ -0,0 +1,7 @@
+<small>
+
+Read more about accessing the user data in the [Accessing User Data](/docs/auth/entities#accessing-the-auth-fields) section of the docs.
```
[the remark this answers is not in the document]
infomiho: > But, what I noticed then, is that this relative path will be resolved relative to where the React component is called from, and not relative to where React component is defined! Give it a try to confirm this is an issue, but from what I saw, it indeed is.

Yes, that's indeed an issue so I had to go for absolute links to be able to use snippets like this.

web/docs/migrate-from-0-13-to-0-14.md
infomiho: Yep, we need to update the Changelog to mention `0.14.0` and I'll add a short description of the breaking changes + link to this page. 

I feel like these migration instructions are not a "change log" because they offer examples, link to other parts of the docs etc.

waspc/examples/todoApp/src/pages/ProfilePage.tsx
infomiho: Okay, I've created and issue and I'll make sure to tackle this soon after we merge this one: https://github.com/wasp-lang/wasp/issues/1947

I'd go for renaming stuff completely: `email` -> `emailAndPassword` and also making sure we use `usernameAndPassword` everywhere.

waspc/examples/todoApp/src/dbSeeds.ts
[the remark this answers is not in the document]
infomiho: I've now hidden this API since it would expose some server imports on the client. Now this test is written a bit differently with the `{} as AuthUser` method which works well for these kinds of tests.

web/docs/auth/_accessing-user-data-note.md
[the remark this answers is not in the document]
Martinsos: Right, but if I am correct, absolute links will not work across different doc versions! THey will work only for the "current" docs, but once you give them version number and they become a bit older, they won't work. That is what this link I provided speaks about.
So at the moment when these docs end up being at `/docs/v0.13` instead of just `/docs`, that link will stop working (I think). And that means absolute URLs also don't work.
Which is tricky, I don't know the solution to this.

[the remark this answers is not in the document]
infomiho: I think I found a solution for this problem: "root relative links" 

Instead of writing it like this: `/docs/auth/entities`, I've written it like this `/auth/entities/entities.md`.

It's still an absolute link, but it's absolute to the "files root", not the "website" and includes the extension.

It works for me:
1. Visit:          `/docs/0.13.5/auth/username-and-pass#using-auth` and click on the link
2. It leads to `/docs/0.13.5/auth/entities#accessing-the-auth-fields`
3. Visit           `/docs/auth/username-and-pass` and click on the link
4. It leads to `/docs/auth/entities#accessing-the-auth-fields`

The `0.13.5` version is something I created locally to test out this link. I found this in the Hasura writing guide 😮‍💨  https://hasura.io/docs/3.0/wiki/docusaurus-mdx-guide/links/

web/docs/auth/entities/entities.md
```diff
@@ -1,22 +1,454 @@
...
+```
+
+</TabItem>
+</Tabs>
+
+All the `User` fields you defined will be present at the top level of the `AuthUser` object. The auth-related fields will be on the `identities` object. For each auth method you enable, there will be a separate data object in the `identities` object.
+
+In the examples above, you can see the `identities` object contains the `email` and `google` objects. The `email` object contains the email-related data and the `google` object contains the Google-related data.
+
+### Auth Methods Data
```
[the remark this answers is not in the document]
infomiho: I've removed the title and now all the content is under "AuthUser Object Fields"

web/docs/auth/entities/entities.md
```diff
@@ -1,22 +1,454 @@
...
+  // ...
+} else if (user.identities.google !== null) {
+  const googleId = user.identities.google.id
+  // ...
+}
+```
+
+### `getFirstProviderUserId` Helper
+
+The `getFirstProviderUserId` method returns the first user ID (e.g. `username` or `email`) that it finds for the user or `null` if it doesn't find any.
```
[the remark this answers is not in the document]
infomiho: When I said `username` here I was thinking about `"fluffyllama"` a not the object. I'll make that clearer!

web/docs/auth/entities/entities.md
```diff
@@ -1,22 +1,454 @@
...
+
+[As you'll learn below](#authidentity-entity-), the `providerUserId` field is how providers identify our users. For example, the user's `username` in the case of the username auth or the user's `email` in the case of the email auth. This can be useful if you support multiple authentication methods and you need _any_ ID that identifies the user in your app.
```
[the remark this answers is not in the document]
infomiho: I've simplified this text a bit not to go into the `providerUserId` detail. We introduce it later and it's a nice to know thing, but not need to know thing. 

I've also added a note about multiple identities per user.

web/docs/auth/entities/entities.md
```diff
@@ -1,22 +1,454 @@
...
+</TabItem>
+</Tabs>
+
+## Including the User with Other Entities
+
+Sometimes, you might want to include the user's data when fetching other entities. For example, you might want to include the user's data with the tasks they have created.
+
+:::caution Be careful to exclude the password
+
+You'll need to include the `auth` and the `identities` relations to get the full data about the user. However, you should exclude the `providerData` field from the `AuthIdentity` entity because it contains the user's hashed password.
```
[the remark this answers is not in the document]
infomiho: The phrasing `might want to include to get the full auth data` sounds lean and strict at the same time to me 😄 

They _need_ to include it to get the auth data, so I've kept the `need` part, but I've used your version to modify the text.

web/docs/auth/entities/entities.md
infomiho: It works with the `AuthUser` and the "raw" user with relations object 😄

web/docs/auth/entities/entities.md
infomiho: Updated a bit

web/docs/migrate-from-0-13-to-0-14.md
infomiho: Because the `AuthUser` they receive _now_ is different from the `AuthUser` they received before.

Before it was `getUsername(user: AuthUser)` and now it's `getUsername(user: UserWithAuthAndIdentities)`, so the helper no longer works on the same type of user.

infomiho: I've updated the text to mention this 👍

web/docs/migrate-from-0-13-to-0-14.md
infomiho: I've updated the phrasing to be clear that it's related to `providerData`

web/docs/auth/entities/entities.md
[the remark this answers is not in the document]
Martinsos: I guess it wasn't so clear to me what auth data means. Or I wanted to make it clear they might not need the full auth data. Because often they won't, they will save some stuff on `user` directly instead.
Ok, good!

web/docs/auth/_accessing-user-data-note.md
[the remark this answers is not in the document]
Martinsos: Ok perfect! Nice job, Resolve this issue if you implemented it.

web/docs/migrate-from-0-13-to-0-14.md
Martinsos: Yeah I guess that makes sense. I would say Changelog should be first choice, and we should create migration script only if the thing is complex enough to need one.
Can you update the Changelog now in this PR? THe idea is we update Changelog in each PR.

Martinsos (commented):
@infomiho responded to comments!

web/docs/migrate-from-0-13-to-0-14.md
infomiho: Yep, I was meaning to do it 👍

Martinsos (approved):

waspc/examples/todoApp/src/apis.ts
infomiho: I mentioned putting stuff on `User` here: https://github.com/wasp-lang/wasp/pull/1915/files#diff-1da5418e699dcbebea8ab516c4e60235efd7191abad17057916ca5f9bba2ddb1R263

But I didn't comment on the pitfalls of the `getFirstProviderUserId` method.

PR #1918, opened 2024-03-23
AuthMethods + Template Haskell experiment

infomiho:
I was think how can we make adding a new auth method a more type safe process. Since we are kinda loosely defining them by adding a new key in the `AuthMethods` record, we don't really have a union data type over which we can force exhaustive pattern matches. 

I've played a bit with Template Haskell to try to change that.

### DIFF
--- a/waspc/src/Wasp/AppSpec/App/Auth/AuthMethods.hs
+++ b/waspc/src/Wasp/AppSpec/App/Auth/AuthMethods.hs
@@ -1,0 +1,105 @@
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE TemplateHaskell #-}
+
+module Wasp.AppSpec.App.Auth.AuthMethods
+  ( generateAuthMethods,
+    userSignupFieldsForEmailAuth,
+    userSignupFieldsForUsernameAuth,
+    userSignupFieldsForExternalAuth,
+    isAuthMethodExternal,
+    AuthMethod (..),
+    UsernameAndPasswordConfig (..),
+    ExternalAuthConfig (..),
+    EmailAuthConfig (..),
+  )
+where
+
+import Data.Data (Data)
+import Language.Haskell.TH
+import Language.Haskell.TH.Syntax (VarBangType)
+import Wasp.AppSpec.App.Auth.EmailVerification (EmailVerificationConfig)
+import Wasp.AppSpec.App.Auth.PasswordReset (PasswordResetConfig)
+import Wasp.AppSpec.App.EmailSender (EmailFromField)
+import Wasp.AppSpec.ExtImport (ExtImport)
+import Wasp.Util (toLowerFirst)
+
+data AuthMethod = UsernameAndPassword | Email | Google | Keycloak | GitHub deriving (Show, Enum, Bounded)
+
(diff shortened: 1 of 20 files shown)


waspc/src/Wasp/AppSpec/App/Auth/AuthMethods.hs
```diff
@@ -0,0 +1,105 @@
...
+    isAuthMethodExternal,
+    AuthMethod (..),
+    UsernameAndPasswordConfig (..),
+    ExternalAuthConfig (..),
+    EmailAuthConfig (..),
+  )
+where
+
+import Data.Data (Data)
+import Language.Haskell.TH
```
Martinsos, line 19: Unqalified, unlisted import -> not a good practice!
I would suggest importing it qualified as TH.

waspc/src/Wasp/AppSpec/App/Auth/AuthMethods.hs
```diff
@@ -0,0 +1,105 @@
...
+-- `userSignupFields` function (otherwise we need to use the DuplicateRecordFields
+-- extension in each module that uses them).
+userSignupFieldsForEmailAuth :: EmailAuthConfig -> Maybe ExtImport
+userSignupFieldsForEmailAuth = userSignupFields
+
+userSignupFieldsForUsernameAuth :: UsernameAndPasswordConfig -> Maybe ExtImport
+userSignupFieldsForUsernameAuth = userSignupFields
+
+userSignupFieldsForExternalAuth :: ExternalAuthConfig -> Maybe ExtImport
+userSignupFieldsForExternalAuth = userSignupFields
```
Martinsos, line 99: What you can do, if you want is, use typeclass for something like this, which has a method `userSignupFields`, and each of these types will implement it. That might be a bit more elegant, depending on how you use it though, what you really need.

waspc/src/Wasp/AppSpec/App/Auth/AuthMethods.hs
```diff
@@ -0,0 +1,105 @@
...
+
+configType :: AuthMethod -> Name
+configType UsernameAndPassword = ''UsernameAndPasswordConfig
+configType Email = ''EmailAuthConfig
+configType Google = ''ExternalAuthConfig
+configType Keycloak = ''ExternalAuthConfig
+configType GitHub = ''ExternalAuthConfig
+
+-- Generate the AuthMethods data type
+-- data AuthMethods = AuthMethods
```
Martinsos, line 56: Seems to me like you ended up defining this relationship anyway above! Becauwe you defined it via `configType` at the end, only to then generate AuthMethods from that. So did you really gain anything?

Martinsos (commented):
Interesting experiment @infomiho , and good job on learning more about TH in practice!

Looking at the PR, it seems to me like the conclusion is that TH brought more complexity then it took away. How are you thinking about it?

AUthMethods part of PR looks reasonable to me, although I would have to see it without TH to make proper review. But TH changes I think you can drop. It would be interesting to understand better what you tried to solve with them, and see if we can achieve that in some other way -> maybe by organizing some code differently, or by using typeclasses, or something like that. Maybe even some TH again, but I think it will be much simpler, if it makes sense at all.

infomiho:
Stale PR, closing this since we won't work on it anymore and we are moving towards using libs #3061

PR #1921, opened 2024-03-26
Fix autocomplete for `wasp` package

### LINKED ISSUE #1839: Investigate and Improve Auto-imports
The restructuring (#1688) we've done for Wasp 0.12.0 introduced Wasp's SDK inside `node_modules`. Imported Wasp functionalities have finally become a normal NPM package.

This opens up an opportunity to greatly improve our DX.
However, our IDE support is a little flaky.

Import autosuggestions should suggest everything a user might want to import (regardless of whether there's already an existing import from that path) and nothing else.

In theory, for each symbol, we should check:
  1. Does the IDE suggest the correct autoimport work (when this is the first symbol imported from that path)?
  2. If not, does the IDE suggest an correct autoimport (when there already is an import from that path)?
  4. Is the correct import suggestion the only import suggestion? We don't want to confuse users and accidentally expose our internals.

In reality, we won't have to do this for each of our public symbols. If we figure out what's going on and how we can best assist the IDE in figuring out what it should suggest, we'll be pretty sure it always works.

sodic:
Patrially fixes #1839.

This PR enables the IDE to properly autocomplete imports exposed from `node_modules/wasp`.

Here's how it looks:

https://github.com/wasp-lang/wasp/assets/22752307/ba4856c6-0914-4950-8af4-db496a5be239

After we merge this, all that's left is hiding the stuff we don't want the user to see: #1922.


waspc/data/Cli/templates/skeleton/tsconfig.json
```diff
@@ -5,6 +5,10 @@
 // compiler. Proper TS compiler configuration in Wasp is coming soon :)
 {
   "compilerOptions": {
+    "target": "esnext",
+    // We're bundling all code in the end so this is the most appropriate option,
+    // it's also important for autocomplete to work properly.
+    "moduleResolution": "bundler",
```
sodic, line 11: I got here by trial and error, but `bundler` seems the most (if not only) appropriate option for our setup.

By implicit defaults cascading, this got set `CommonJs`, which is why it failed to recognize ES exports from the `wasp` package.

The cascading was:
- `target` was unset and defaulted to `ES3` 
- `module` was unset and defaulted to `CommonJS` because the `target` was `ES3`
- `moduleResolution` was unset and defaulted to `Node` because `module` was `CommonJS`

Source for all rules: https://www.typescriptlang.org/tsconfig#module

Martinsos:
Awesome!

waspc/data/Cli/templates/skeleton/tsconfig.json
infomiho: Given that this is in the user's `tsconfig.json`, does this mean that only the new newly created projects will get the benefits of these changes?

Do we also need to update _all_ other templates?

sodic: > Given that this is in the user's tsconfig.json, does this mean that only the new newly created projects will get the benefits of these changes?

Yes, any ideas on what we can do here?

> Do we also need to update all other templates? Specifically the ones you get when running wasp new:

Apparently yes, nice catch! I thought they all used `skeleton`... We really do need to change those names :smile:

Martinsos: Yeah, those names are a bit tricky. But don't they use `skeleton`? Except for open-saas?

sodic: I tested it out, and unfortunately not.

sodic: Btw, what do I do with that?

Should I do anything now or do we take care of it in the release process: https://github.com/wasp-lang/wasp/tree/main/waspc#typical-release-process.

Martinsos: I think it makes sense to do now, when we still have context about it and it is fresh.

But you can do it another PR if you wish! Or maybe you even have to, for these starters.

Issue #1841, opened 2024-02-29
Refactor our generated codebase


sodic:
Wasp 0.12.0 split our generated code into Framework code (`web-app` and `server`) and SDK code (the `wasp` package). See #1429.

The split is suboptimal, mostly because the SDK contains a lot of the code it doesn't need (it was easier to implement the split without changing too much).

We should go through the modules we generate and reorganize them to minimize the dependencies between them and make them as modular as possible.

This will have an added bonus of speeding up our SDK build step: #1840).

PR #1922, opened 2024-03-26
[RFC & POC] Hiding Private API

sodic:
## Context
The plan for Wasp 0.12.0 was:
- Extract all the functionalities we want to expose to our users into the Wasp SDK (`node_modules/wasp`)
- Keep all under-the-hood stuff Wasp needs to run hidden from the user inside the framework code (`.wasp/out{server,web-app}`).

Unfortunately, because of coupling between functionalities, many of Wasp's internal functionalities had to be moved from the framework code into the SDK:

What we wanted:             |  What we currently have:
:-------------------------:|:-------------------------:
  |  

## The problem
This architecture would not be a problem in and of itself.

Unfortunately, since the framework code (i.e., the runtime) needs to import the internal functionalities from the SDK, the SDK must expose them as a part of its public API.

In other words, **Wasp's private functionalities are now visible to the user**.
Perhaps even worse, **Wasp's private functionalities are also visible to the IDE** which suggest them as autoimports, further pushing our users to rely on undocumented or incorrect API.

## The proposed solution
There are two ways to approach the problem:
1. **Refactor our codebase so that the SDK does not expose private API** (covered by [#1841](https://github.com/wasp-lang/wasp/issues/1841)) 
  This is the best possible solution which we plan to undertake at some point. The downside is that it's very time consuming and has considerable chances of breaking something in Wasp.
2. **Hide the private API from the user's code and IDE**
  If possible, doing so should give us an identical visible result with much less work.

This PR proposes a way to implement the second approach. Read the rest of the comments and the code to see how it's done.


waspc/data/Generator/templates/react-app/vite.config.ts
```diff
@@ -30,6 +30,7 @@ const defaultViteConfig = {
     outDir: "build",
   },
   resolve: {
+    conditions: ["client-runtime"],
```
sodic, line 33: This is where we set the framework runtime's condition string, thus gaining access to the SDK's private API.

waspc/data/Generator/templates/sdk/wasp/package.json
```diff
@@ -9,6 +9,9 @@
       "types": "tsc --declaration --emitDeclarationOnly --stripInternal --declarationDir dist"
     },
   "exports": {
+      "./client/operations/queryClient": {
+        "client-runtime": "./dist/client/operations/queryClient.js"
+      },
```
sodic, line 14: This how we expose the SDK's private API to the framework, while hiding it from the user.

waspc/data/Generator/templates/sdk/wasp/client/operations/index.ts
sodic, line 1: As explained in the description, I've reorganized this file to include export only the public API.

waspc/data/Generator/templates/sdk/wasp/auth/useAuth.ts
```diff
@@ -1,6 +1,7 @@
 {{={= =}=}}
 import { deserialize as superjsonDeserialize } from 'superjson'
-import { useQuery, addMetadataToQuery } from 'wasp/client/operations'
+import { useQuery } from '../client/operations'
+import { addMetadataToQuery } from '../client/operations/queries/core'
```
sodic, line 4: As explained in the issue's description, all the import paths inside the SDK have been converted from package imports to relative imports (also accounting for the updated paths).

waspc/data/Generator/templates/react-app/src/index.tsx
```diff
@@ -7,7 +7,7 @@ import router from './router'
 import {
   initializeQueryClient,
   queryClientInitialized,
-} from 'wasp/client/operations'
+} from 'wasp/client/operations/queryClient'
```
sodic, line 10: All SDK imports in the framework code have been updated according to the new code organization.

waspc/data/Generator/templates/sdk/wasp/package.json
infomiho: Great stuff! I didn't know that `exports` API worked this way. I always thought that the `import` and `require` were the only options.

Martinsos:
Nice job @sodic , nice analysis!

It all sounds reasonable, and it is cool that there is such thing as conditional exports. How much is our solution "hacky" here -> is this what conditional exports are supposed to be used for?

What about that potential bug with symlink you mentioned -> did you get any response regarding that?

Additional idea, I will shortly explain it and you tell me if it doesn't make any sense: what if we had another package, called "wasp-core", that would contain logic from SDK that is used both by SDK and the framework code? And then that package would be used by SDK, but not reexported, and it would also be used by framework code. We would need to do refactoring, extract code from SDK into it, but result would achieve what we wanted, right? What would the pros/cons be here (if it works)?

Btw I will let @infomiho do the details review of this PR, while I will stick to high level stuff.

sodic:
> How much is our solution "hacky" here -> is this what conditional exports are supposed to be used for?

Not very hacky.

Conditional exports are meant to be used with user-provided strings and arbitrary semantics (see How much is our solution "hacky" here -> is this what conditional exports are supposed to be used for? for details). Of course, I doubt anyone envisioned our case specifically (since we're so cutting edge), but it's not like we're abusing a completely unrelated mechanism.

> What would the pros/cons be [of a hidden wasp-core package] (if it works)?

The biggest pro is a very obvious separation between internal and external stuff.

As for the cons...
This is another way of refactoring our code, so carries most of the same cons (except for the duplication). It also comes with some additional cons that are a consequence of creating a different package.

Let's use this terminology when exploring the cons:
 - `wasp-core` - private SDK
 - `wasp` - public SDK

Anyway, the `wasp-core` approach: 
- **Takes a lot more time** - self-explanatory.
- **Complicates our setup** - One more build step, another `package.json` and `tsconfig.json` to maintain, one more folder added to the Docker's build context... 
- **Creates an extra layer of separation** - More precisely, our SDK is no longer organized by "features" (i.e., all auth stuff goes here). The first layer of separation now depends on whether something is internal (`wasp-core`) or external (`wasp`), which is not as nice. It creates yet another file tree of similarly named files and folders, which makes it easier to edit the wrong stuff during development
- **Makes future refactors and changes difficult** - With the conditional exports approach, refactors and visibility changes are  pretty straightforward. With the `wasp-core` approach, not so much. I'll give two examples:
  - **Example 1:** Imagine a very local function used only in `wasp` (it sits in `wasp/something` and is not exported). Now imagine we want to reuse this function in our framework code. We have to move it to `wasp-core` and change all the imports in `wasp`.
  - **Example 2:** Imagine we want to make a private function in `wasp-core` public. Let's assume the function is used by both `wasp` and the framework code. To make it public, we must: change the function's location (move from `wasp-core` to `wasp`), update all the imports in the SDK, update all the imports in the framework code, and possibly do some refactoring (if other functions in the `wasp-core` depended on it)
- **Doesn't hide anything from the user** - They can still import and get auto-suggestions from `wasp-core`. Sure, we could add a compile-time check that prohibits any `wasp-core` imports, but that's another complication. Or we could use conditional exports for the `wasp-core` stuff, but then we could have just used that in the first place.

> What about that potential bug with symlink you mentioned -> did you get any response regarding that?

I'll update the PR's description with the issue link once as soon as I report it.

Martinsos:
> ### Is this hacky?
> > How much is our solution "hacky" here -> is this what conditional exports are supposed to be used for?
> 
> Not very hacky.
> 
> Conditional exports are meant to be used with user-provided strings and arbitrary semantics (read [the docs](https://nodejs.org/api/packages.html#resolving-user-conditions) for details).
> 
> Of course, I doubt anyone envisioned our case specifically (since we're so cutting edge :)), but it's not like we're abusing a completely unrelated mechanism.
> 
> ### How does it compare to the `wasp-core` idea?
> > What would the pros/cons be [of a hidden wasp-core package] (if it works)?
> 
> The most obvious pro is a very obvious separation between internal and external stuff. Another pro is its simplicity - the mechanism it uses (two packages) is much simpler and more normal than conditional exports.
> 
> As for the cons...
> 
> This is another way of refactoring our code, so carries similar time implications. It also comes with additional cons that are a consequence of creating a different package.
> 
> I'll use this terminology when exploring the cons:
> 
> * `wasp-core` - private SDK
> * `wasp` - public SDK
> 
> Anyway, the `wasp-core` approach:
> 
> 1. **Takes a lot more time** - self-explanatory.
> 2. **Complicates our setup** - One more build step, another `package.json` and `tsconfig.json` to maintain, one more folder added to the Docker's build context...
> 3. **Creates an extra layer of separation** - More precisely, our SDK is no longer organized by "features" (i.e., all auth stuff goes here). The first layer of separation now depends on whether something is internal (`wasp-core`) or external (`wasp`), which is not as nice. It creates yet another file tree of similarly named files and folders, which makes it easier to edit the wrong stuff during development
> 4. **Makes future refactors and changes difficult** - With the conditional exports approach, refactors and visibility changes are  pretty straightforward. With the `wasp-core` approach, not so much. I'll give two examples:
>    
>    * **Example 1:** Imagine a very local function used only in `wasp` (it sits in `wasp/something` and is not exported). Now imagine we want to reuse this function in our framework code. We have to move it to `wasp-core` and change all the imports in `wasp`.
>    * **Example 2:** Imagine we want to make a private function in `wasp-core` public. Let's assume the function is used by both `wasp` and the framework code. To make it public, we must: change the function's location (move from `wasp-core` to `wasp`), update all the imports in the SDK, update all the imports in the framework code, and possibly do some refactoring (if other functions in the `wasp-core` depended on it)
> 5. **Doesn't hide anything from the user** - They can still import and get auto-suggestions from `wasp-core`. Sure, we could add a compile-time check that prohibits any `wasp-core` imports, but that's another complication. Or we could use conditional exports for the `wasp-core` stuff, but then we could have just used that in the first place.
> 
> ### Alternative to the alternative
> @infomiho suggested that, instead of the currently proposed separation between `wasp-core` and `wasp`, we have:
> 
> * `wasp-core` - All SDK functionality and implementations (both public and private).
> * `wasp` - Exclusively reexports stuff from `wasp-core` we want to make public (no implementations).
> 
> This approach is slightly better because it solves problem 4 (difficult to change visibility and refactor), but still suffers from the rest of the issues.
> 
> ### What about the symlink?
> > What about that potential bug with symlink you mentioned -> did you get any response regarding that?
> 
> I'll update the PR's description with the issue link once as soon as I report it. Just to be clear: this isn't related to conditional exports in any way, I've tested it with and without them.
> 
> Still, even if we doesn't get fixed immediately, implementing conditional exports would still be an improvement - that long import looks very out of place, and users will have an easier time identifying a mistake.

Ok! Convincing analysis, makes sense to me. I am ok with sticking with the approach you started in this PR, and it is nice to know we potentially also have some other options in the future also on the table.

sodic:
Closing - we've done a proof of concept, it was successful.

I've linked this POC from the issue so that someone who continues working on it knows to check it out.

PR #1930, opened 2024-04-02
Explicit public API types

### LINKED ISSUE #1776: Explicitly type our public API
According to the module boundaries rule explained in https://github.com/wasp-lang/vscode-wasp/pull/6#discussion_r970774084, we should explicitly type all module exports.

This is especially true for our public APIs (it also helps avoid issues such as [this one](https://github.com/wasp-lang/wasp/pull/1773/files#r1492398790)).

To close this issue:
- Add the [explicit-module-boundaries](https://typescript-eslint.io/rules/explicit-module-boundary-types/) rule to `sdk/tsconfig.json`
- Correctly type all module boundaries and public exports.

sodic:
Fixes #1776


waspc/data/Generator/templates/sdk/wasp/client/index.ts
```diff
@@ -11,4 +11,4 @@ export enum HttpMethod {
 export type Route = { method: HttpMethod; path: string }
 
 // PUBLIC API
-export { default as config } from './config'
+export { config, ClientConfig } from './config'
```
sodic, line 14: I've read in "Effective typescript" (or something like that):

> When you export a value X of type T, you should also expose T. In other words, if users can and see a type, it's already public, so make it easily accessible.

I've followed this principle for `emailSender` as well

waspc/src/Wasp/Generator/SdkGenerator/Auth/OAuthAuthG.hs
```diff
@@ -40,7 +40,7 @@ genHelpers auth =
     mkHelpersFd :: OAuthAuthProvider -> Path' Rel' File' -> FileDraft
     mkHelpersFd provider helpersFp =
       mkTmplFdWithDstAndData
-        [relfile|auth/helpers/Generic.tsx|]
+        [relfile|auth/helpers/_Provider.tsx|]
```
sodic, line 43: @infomiho This is the convention the rest of our code uses for template files that change their name and output multiple source files. Hope it's ok!

waspc/data/Generator/templates/sdk/wasp/client/router/Link.tsx
```diff
@@ -13,7 +13,7 @@ export function Link(
     hash?: string;
   }
   & Routes
-) {
+): React.JSX.Element {
```
sodic, line 16: I haven't done anything fancy with the types in this PR, only explicitly stated what the functions already returned.

That said, I'm still a little fuzzy on these React component return types. It might make sense to return to them one day.

waspc/data/Generator/templates/sdk/wasp/auth/useAuth.ts
```diff
@@ -4,12 +4,13 @@ import { useQuery, addMetadataToQuery } from 'wasp/client/operations'
...
 import type { AuthUser } from './types'
+import { UseQueryResult } from '@tanstack/react-query'
 
 // PUBLIC API
-export const getMe = createUserGetter()
+export const getMe: () => Promise<AuthUser | null> = createUserGetter()
 
 // PUBLIC API
-export default function useAuth(queryFnArgs?: unknown, config?: any) {
+export default function useAuth(queryFnArgs?: unknown, config?: any): UseQueryResult<AuthUser> {
```
sodic, line 13: This function has a strange signature. I'll create an issue for investigating (out of scope for this PR).

waspc/e2e-test/test-outputs/waspBuild-golden/waspBuild/.wasp/build/sdk/wasp/dist/client/router/Link.d.ts
```diff
@@ -4,5 +4,5 @@ type RouterLinkProps = Parameters<typeof RouterLink>[0];
 export declare function Link({ to, params, search, hash, ...restOfProps }: Omit<RouterLinkProps, "to"> & {
     search?: Record<string, string>;
     hash?: string;
-} & Routes): import("react").JSX.Element;
+} & Routes): React.JSX.Element;
```
sodic, line 7: Does anyone know why referencing React globally like this works?

waspc/src/Wasp/Generator/SdkGenerator/Auth/OAuthAuthG.hs
infomiho: Great change 👍

waspc/data/Generator/templates/sdk/wasp/client/operations/internal/index.ts
```diff
@@ -6,10 +6,10 @@ import {
  } from 'superjson'
 
 // PRIVATE API
-export type OperationRoute = { method: HttpMethod, path: string }
+export type OperationRoute = { method: HttpMethod.Post, path: string }
 
 // PRIVATE API
-export async function callOperation(operationRoute: OperationRoute & { method: HttpMethod.Post }, args: any) {
+export async function callOperation(operationRoute: OperationRoute, args: any) {
```
sodic, line 12: They all used `HttpMethod.Post` and this was causing type errors in `core.ts`, so I've changed it.

waspc/data/Generator/templates/sdk/wasp/client/operations/actions/core.ts
```diff
@@ -32,5 +37,12 @@ export function createAction(relativeActionRoute, entitiesUsed) {
   const action = (args) => internalAction(args, [])
   action.internal = internalAction
 
-  return action
+  return action as ActionFor<BackendAction>
```
sodic, line 40: Needed to add this to avoid type errors.

I don't think a comment is necessary, the file is not migrated fully to TS anyway. However does that one day will know what to do.

waspc/data/Generator/templates/sdk/wasp/client/operations/queries/core.ts
```diff
@@ -1,11 +1,16 @@
...
+import { type Query } from '../core.js'
 import { callOperation, makeOperationRoute } from '../internal/index.js'
 import {
   addResourcesUsedByQuery,
   getActiveOptimisticUpdates,
 } from '../internal/resources'
 
-// PRIVATE API
-export function createQuery(relativeQueryPath, entitiesUsed) {
+export function createQuery<BackendQuery extends GenericBackendQuery>(
```
sodic, line 10: Declaration merged from `core.d.ts`. Same goes for `queries/core.ts`.

I had to do this because our `core.d.ts` file would get overwritten during compilation - TypeScript would compile the `core.js` file and take the types from there. These types were useless - all of them were `any`.

Issue #1840, opened 2024-02-29
Optimize SDK build step


sodic:
Building the SDK (i.e., the `wasp` package) takes several seconds even on high-end machines.

Explore ways to optimize the SDK build step.

For example:
1. Don't build the SDK if not necessary (e.g., when there no changes or when the user user-code has errors).
2. Minimize how often it's necessary to build the SDK (e.g., by stripping the SDK of all unnecessary functionality).
3. Find other ways to speed up the build.

infomiho:
Relevant flame chart https://kitchen-sink-startup-profile.share.wasp.sh

Issue #1829, opened 2024-02-29
Reduce or hide Vite HMR messages


sodic:
After 0.12.0 (possibly sooner), Vite HMR messages began cluttering the output:
  1. Start a Wasp app
  2. Open it in a browser
  3. Change something inside one of the components

Here's the produced output for a single line change in one of the pages:
```
✅ --- Database successfully set up. ----------------------------------------------


🐝 --- Building SDK... ------------------------------------------------------------


[ Client ] 5:06:20 PM [vite] page reload /home/filip/Projects/wasp/wasp-repo/waspc/examples/todo-typescript/.wasp/out/sdk/wasp/dist/api/events.js
[ Client ] 5:06:20 PM [vite] page reload /home/filip/Projects/wasp/wasp-repo/waspc/examples/todo-typescript/.wasp/out/sdk/wasp/dist/universal/url.js
[ Client ] 5:06:20 PM [vite] page reload /home/filip/Projects/wasp/wasp-repo/waspc/examples/todo-typescript/.wasp/out/sdk/wasp/dist/core/config.js
[ Client ] 5:06:20 PM [vite] page reload /home/filip/Projects/wasp/wasp-repo/waspc/examples/todo-typescript/.wasp/out/sdk/wasp/dist/core/storage.js
[ Client ] 5:06:20 PM [vite] page reload /home/filip/Projects/wasp/wasp-repo/waspc/examples/todo-typescript/.wasp/out/sdk/wasp/dist/api/index.js
[ Client ] 5:06:20 PM [vite] hmr update /@fs/home/filip/Projects/wasp/wasp-repo/waspc/examples/todo-typescript/src/MainPage.tsx, /@fs/home/filip/Projects/wasp/wasp-repo/waspc/examples/todo-typescript/src/Main.css
[ Client ] 5:06:20 PM [vite] hmr update /@fs/home/filip/Projects/wasp/wasp-repo/waspc/examples/todo-typescript/src/MainPage.tsx, /@fs/home/filip/Projects/wasp/wasp-repo/waspc/examples/todo-typescript/src/Main.css
[ Client ] 5:06:20 PM [vite] page reload /home/filip/Projects/wasp/wasp-repo/waspc/examples/todo-typescript/.wasp/out/sdk/wasp/dist/client/operations/queryClient.js
[...]

PR #1931, opened 2024-04-02
Updates SDK tsconfig to enable the `incremental`

infomiho:
This helps with Vite HMR working properly since rebuilds of the SDK files would result in full reload instead of proper HMR. 

This also has some nice side-effects:
- faster rebuilds of the SDK (helpful for https://github.com/wasp-lang/wasp/issues/1840)
- less Vite messages (helpful for https://github.com/wasp-lang/wasp/issues/1829)

### DIFF
--- a/waspc/data/Generator/templates/sdk/wasp/tsconfig.json
+++ b/waspc/data/Generator/templates/sdk/wasp/tsconfig.json
@@ -28,6 +28,7 @@
       // about missing types e.g. when using `toBeInTheDocument` and other matchers.
       "@testing-library/jest-dom"
     ],
+    "incremental": true,
     // todo(filip): Only works with common js, see https://www.typescriptlang.org/tsconfig#paths and daily-article.
     // "paths": {
     //   "@wasp/*": [


waspc/e2e-test/GoldenTest.hs
```diff
@@ -96,7 +96,8 @@ runGoldenTest goldenTest = do
                     "dev.db-journal",
                     "package-lock.json",
                     ".gitignore",
-                    ".DS_Store"
+                    ".DS_Store",
+                    "tsconfig.tsbuildinfo"
```
infomiho, line 100: It's a debug file written by TSC when using the `incremental` option.

waspc/data/Generator/templates/sdk/wasp/tsconfig.json
```diff
@@ -28,6 +28,7 @@
       // about missing types e.g. when using `toBeInTheDocument` and other matchers.
       "@testing-library/jest-dom"
     ],
+    "incremental": true,
```
sodic, line 31: Great find!

Wondering why this isn't the default option :thinking:

sodic (approved):
As good looking to me as it gets :smile:

PR #1960, opened 2024-04-15
Testing types of `todoApp` in CI

infomiho:
Sometimes we have type errors in the IDE due to `tsconfig.json` not being properly set up. We want to raise those errors in the CI so we can avoid releasing with type issues.

This should help with this class of issues: #1989

### DIFF
--- a/.github/workflows/waspc-ci.yaml
+++ b/.github/workflows/waspc-ci.yaml
@@ -158,7 +158,7 @@ jobs:
       - name: Run unit tests
         run: cabal test cli-test waspc-test waspls-test
 
-      - name: Ensure todoApp builds
+      - name: Ensure todoApp works
         if: matrix.os == 'ubuntu-20.04' || matrix.os == 'macos-latest'
         run: ./tools/ensure_todoapp_builds.sh
 
--- a/.github/workflows/waspc-ci.yaml
+++ b/.github/workflows/waspc-ci.yaml
@@ -160,7 +160,7 @@ jobs:
 
       - name: Ensure todoApp works
         if: matrix.os == 'ubuntu-20.04' || matrix.os == 'macos-latest'
-        run: ./tools/ensure_todoapp_builds.sh
+        run: ./tools/ensure_todoapp_works.sh
 
       - name: Headless - Cache Node Modules
         id: headless-cache-node-modules
--- a/waspc/examples/todoApp/package.json
+++ b/waspc/examples/todoApp/package.json
@@ -9,6 +9,7 @@
   "devDependencies": {
     "@types/express": "^4.17.13",
     "@types/react": "^18.0.37",
+    "@types/uuid": "^9.0.8",
     "prisma": "4.16.2",
     "typescript": "^5.1.0",
     "vite": "^4.3.9"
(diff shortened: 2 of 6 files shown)


waspc/tools/ensure_todoapp_builds.sh
```diff
@@ -13,6 +13,9 @@ cd "$dir/../examples/todoApp"
 # Compile example app.
 cabal run wasp-cli build
 
+echo "Ensure the app has no IDE type errors"
+npx tsc --noEmit --skipLibCheck
```
infomiho, line 17: Building with `tsc` in the project root dir gives us the same IDE errors users would see.

waspc/examples/todoApp/src/App.tsx
```diff
@@ -13,11 +13,13 @@ export function App({ children }: any) {
 
   const connectionIcon = isConnected ? '🟢' : '🔴'
 
+  const appName = import.meta.env.REACT_APP_NAME ? import.meta.env.REACT_APP_NAME : 'TODO App'
```
infomiho, line 16: With the current `tsconfig.json`, `import.meta.env` causes errors.

waspc/examples/todoApp/package.json
```diff
@@ -9,6 +9,7 @@
   "devDependencies": {
     "@types/express": "^4.17.13",
     "@types/react": "^18.0.37",
+    "@types/uuid": "^9.0.8",
```
infomiho, line 12: Fixes a type error which we saw in the IDE due to missing types 👍

.github/workflows/waspc-ci.yaml
```diff
@@ -158,7 +158,7 @@ jobs:
       - name: Run unit tests
         run: cabal test cli-test waspc-test waspls-test
 
-      - name: Ensure todoApp builds
+      - name: Ensure todoApp works
         if: matrix.os == 'ubuntu-20.04' || matrix.os == 'macos-latest'
         run: ./tools/ensure_todoapp_builds.sh
```
sodic, line 163: Maybe change the script's name to say "works" instead of "builds" too

sodic (approved):

waspc/tools/ensure_todoapp_builds.sh
Martinsos: Just curios, why do we need to mention "IDE" here? I know they are visible in the IDE, but they are not really IDE specific, right? Aren't these TS errors and that is it -> the fact that they are shown by IDE sounds irrelevant. They could be shown by Wasp CLI also. I think we care about the fact that there are type errors from user's code in their Wasp project? So maybe let's call them that way then, if that is correct.

sodic: > Aren't these TS errors and that is it

Yes, but also no. :smile: 
Due to how our dev setup works, the only place they'll see them is inside the IDE.

> They could be shown by Wasp CLI also.

Yes, but they aren't. They're only shown in the IDE (by design).

All that said, I agre with @Martinsos , IDE type errors is a little misleading.

I'd still mention the IDE though, mostly because user will report these errors as "I have errors in my IDE, but everything works". They'll see it as IDE-related because they don't know about all the crazy stuff we do :smile:  

Anyway, I suggest:
```
Ensure the user's code has no TypeScript errors (shown in the IDE).
```
A little verbose, but this way it's both:
- Correct.
- Easy to figure out what you need to look at when a user reports their IDE throwing errors.

sodic:
@infomiho It's mergin' time? :D

infomiho:
@sodic Let's see if the CI passes then we'll merge 🙃

Issue #1556, opened 2023-11-02
Add hooks after login / sign-in, to execute custom server code


Martinsos:
> Be able to execute custom server actions after a successful auth

> Both after signup and login would be best case scenario. I would like to redirect user to a 'setup workspace' flow after registering, depending if they have created a workspace already (checking a field in DB). Currently I'm redirecting the user after page/app load using router from  react-router-dom.

Related convo: https://discord.com/channels/686873244791210014/867713251479519232/1168305292098936932 .

infomiho:
I quite liked working with hooks in Directus, they are using `eventemitter2` for that: https://github.com/infomiho/directus/blob/0230174b846c41b8030a14103a195c94e4d34e97/api/src/emitter.ts#L4

PR #1961, opened 2024-04-15
Adding `postLogin` and `postSignup` hooks

infomiho:
A lot of users asked for the ability to execute code after some user has signed up or logged in.

We decided in a internal RFC that we want to build these hooks like we did the client and server setup functions: by importing user defined function and executing it where it makes sense.

In this PR, I have now only create a sketch of the API I plan on implementing. There are some open questions:
- [ ] **Which params do we want to pass to the hook?**
  I went with the minimal `ProviderId` which is an object containing the provider name and the user ID for that provider. They can probably fetch everything else they need.

- [ ] **Do we want the hook to be async?**
  Probably yes. They'd made want to send some info to an external API.

- [ ] **Do we want the hook to _block_ while it executes?**
  Aka the login and signup process won't continue until the hook code finished executing.

- [ ] **Do we use the same hooks (postLogin and postSignup) for both email, username and _OAuth_ providers?**
  Maybe yes, since it's the simplest thing to do.

- [ ] **Do we have special hook for OAuth providers?** 
  I'm wondering about this since for OAuth providers there is a moment of signup and a moment of login, but _usually_ it's not that important of a distinction. 

- [ ] **Do we want some `preLogin` and `preSignup` hooks?**
  The users haven't requested this, so maybe not add it immedietly? 

- [ ] **Can the `pre` hooks modify data or reject signup or login?**
  If they can't, I don't see much point in having these hooks.
  
  Closes https://github.com/wasp-lang/wasp/issues/1556

### DIFF
--- a/waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
+++ b/waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
@@ -1,0 +1,29 @@
+import { type ProviderId } from '../../auth/utils.js'
+import { type PrismaClient } from '@prisma/client'
+
+export type AuthHookFn =
+  | OnUserSignUpHookFn
+  | OnUserLoginHookFn
+  | OnUserLogoutHookFn
+
+type OnUserLoginHookFn = (context: {
+  providerId: ProviderId
+  prisma: PrismaClient
+}) => Promise<void> | void
+type OnUserSignUpHookFn = (context: {
+  providerId: ProviderId
+  prisma: PrismaClient
+}) => Promise<void> | void
+type OnUserLogoutHookFn = (context: {
+  providerId: ProviderId
+  prisma: PrismaClient
+}) => Promise<void> | void
+
+type HookType = 'onUserSignUp' | 'onUserLogin' | 'onUserLogout'
+
+export function defineHook<HT extends HookType>(
+  hookType: HT,
+  hookFn: AuthHookFn,
+): AuthHookFn {
+  return hookFn
+}
--- a/waspc/data/Generator/templates/server/src/auth/hooks.ts
+++ b/waspc/data/Generator/templates/server/src/auth/hooks.ts
@@ -1,0 +1,39 @@
+{{={= =}=}}
+import { prisma } from 'wasp/server'
+import { type ProviderId } from 'wasp/auth/utils'
+import { type AuthHookFn } from 'wasp/auth/hooks'
+
+{=# onUserSignUpHook.isDefined =}
+{=& onUserSignUpHook.importStatement =}
+export const onUserSignUpHook = addExtraParamsToTheHookCall({= onUserSignUpHook.importIdentifier =})
+{=/ onUserSignUpHook.isDefined =}
+{=^ onUserSignUpHook.isDefined =}
+export const onUserSignUpHook = undefined
+{=/ onUserSignUpHook.isDefined =}
+
+{=# onUserLoginHook.isDefined =}
+{=& onUserLoginHook.importStatement =}
+export const onUserLoginHook = addExtraParamsToTheHookCall({= onUserLoginHook.importIdentifier =})
+{=/ onUserLoginHook.isDefined =}
+{=^ onUserLoginHook.isDefined =}
+export const onUserLoginHook = undefined
+{=/ onUserLoginHook.isDefined =}
+
+{=# onUserLogoutHook.isDefined =}
+{=& onUserLogoutHook.importStatement =}
+export const onUserLogoutHook = addExtraParamsToTheHookCall({= onUserLogoutHook.importIdentifier =})
+{=/ onUserLogoutHook.isDefined =}
+{=^ onUserLogoutHook.isDefined =}
+export const onUserLogout = undefined
+{=/ onUserLogout.isDefined =}
+
+type InternalAuthHookFn = ({ providerId }: { providerId: ProviderId }) => Promise<void> | void
+
+function addExtraParamsToTheHookCall(hookFn: AuthHookFn): InternalAuthHookFn {
+  return ({ providerId }) => {
+    return hookFn({
+      providerId,
+      prisma,
+    })
+  }
+}
--- a/waspc/data/Generator/templates/server/src/auth/providers/email/login.ts
+++ b/waspc/data/Generator/templates/server/src/auth/providers/email/login.ts
@@ -36,6 +36,8 @@ export function getLoginRoute() {
     
         const auth = await findAuthWithUserBy({ id: authIdentity.authId })
         const session = await createSession(auth.id)
+
+        // postLogin(providerId)
       
         return res.json({
             sessionId: session.id,
--- a/waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
+++ b/waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
@@ -137,6 +137,8 @@ export function getSignupRoute({
             console.error("Failed to send email verification email:", e);
             throw new HttpError(500, "Failed to send email verification email.");
         }
+
+        // postSignup(providerId)
       
         return res.json({ success: true });
     };
--- a/waspc/data/Generator/templates/server/src/auth/providers/oauth/user.ts
+++ b/waspc/data/Generator/templates/server/src/auth/providers/oauth/user.ts
@@ -63,6 +63,9 @@ async function getAuthIdFromProviderDetails(
   })
 
   if (existingAuthIdentity) {
+
+    // postLogin(providerId)
+
     return existingAuthIdentity.{= authFieldOnAuthIdentityEntityName =}.id
   } else {
     const userFields = await validateAndGetUserFields(
@@ -80,6 +83,8 @@ async function getAuthIdFromProviderDetails(
       // rely on Prisma to validate the data.
       userFields as any,
     )
+
+    // postSignup(providerId)
 
     return user.auth.id
   }
--- a/waspc/data/Generator/templates/server/src/auth/providers/username/login.ts
+++ b/waspc/data/Generator/templates/server/src/auth/providers/username/login.ts
@@ -32,9 +32,11 @@ export default handleRejection(async (req, res) => {
 
   const auth = await findAuthWithUserBy({
     id: authIdentity.authId
-  }) 
+  })
 
   const session = await createSession(auth.id)
+
+  // postLogin(providerId)
 
   return res.json({
       sessionId: session.id,
--- a/waspc/data/Generator/templates/server/src/auth/providers/username/signup.ts
+++ b/waspc/data/Generator/templates/server/src/auth/providers/username/signup.ts
@@ -44,6 +44,8 @@ export function getSignupRoute({
     } catch (e: unknown) {
       rethrowPossibleAuthError(e)
     }
+
+    // postSignup(providerId)
   
     return res.json({ success: true })
   })
--- a/waspc/examples/todoApp/main.wasp
+++ b/waspc/examples/todoApp/main.wasp
@@ -39,6 +39,10 @@ app todoApp {
         },
       },
     },
+    hooks: {
+      postSignup: import { postSingupHook } from "@src/auth/hooks.js",
+      postLogin: import { postLoginHook } from "@src/auth/hooks.js",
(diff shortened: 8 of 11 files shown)


waspc/data/Generator/templates/server/src/auth/providers/email/login.ts
```diff
@@ -36,6 +36,8 @@ export function getLoginRoute() {
     
         const auth = await findAuthWithUserBy({ id: authIdentity.authId })
         const session = await createSession(auth.id)
+
+        // postLogin(providerId)
```
infomiho, line 40: After we create the session, execute `postLogin`

waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
```diff
@@ -137,6 +137,8 @@ export function getSignupRoute({
             console.error("Failed to send email verification email:", e);
             throw new HttpError(500, "Failed to send email verification email.");
         }
+
+        // postSignup(providerId)
```
infomiho, line 141: After the user is successfully created, execute `postSignup`

waspc/data/Generator/templates/server/src/auth/providers/email/login.ts
infomiho: The param `providerId` is something like `{ providerName: "username", providerUserId: "miho" }`

waspc/data/Generator/templates/server/src/auth/providers/oauth/user.ts
```diff
@@ -63,6 +63,9 @@ async function getAuthIdFromProviderDetails(
   })
 
   if (existingAuthIdentity) {
+
+    // postLogin(providerId)
```
infomiho, line 67: This is an existing OAuth user, `postLogin` is executed.

waspc/data/Generator/templates/server/src/auth/providers/oauth/user.ts
```diff
@@ -81,6 +84,8 @@ async function getAuthIdFromProviderDetails(
       userFields as any,
     )
 
+    // postSignup(providerId)
```
infomiho, line 87: The first time the OAuth user is encountered, it's created, so the `postSignup` hook is executed.

waspc/data/Generator/templates/server/src/auth/providers/username/login.ts
```diff
@@ -32,10 +32,12 @@ export default handleRejection(async (req, res) => {
 
   const auth = await findAuthWithUserBy({
     id: authIdentity.authId
-  }) 
+  })
 
   const session = await createSession(auth.id)
 
+  // postLogin(providerId)
```
infomiho, line 39: Session is created, `postLogin` is executed.

waspc/data/Generator/templates/server/src/auth/providers/username/signup.ts
```diff
@@ -44,6 +44,8 @@ export function getSignupRoute({
     } catch (e: unknown) {
       rethrowPossibleAuthError(e)
     }
+
+    // postSignup(providerId)
```
infomiho, line 48: User is created, `postSignup` is executed.

waspc/examples/todoApp/main.wasp
```diff
@@ -39,6 +39,10 @@ app todoApp {
         },
       },
     },
+    hooks: {
```
infomiho, line 42: `auth.hooks.postSignup` and `auth.hooks.postLogin` is something users define in their Wasp file.

Do we want to put it somewhere else? `auth.postLogin`?

Martinsos:
> A lot of users asked for the ability to execute code after some user has signed up or logged in.
> 
> We decided in a internal RFC that we want to build these hooks like we did the client and server setup functions: by importing user defined function and executing it where it makes sense.
> 
> In this PR, I have now only create a sketch of the API I plan on implementing. There are some open questions:
> 
> * [ ]  **Which params do we want to pass to the hook?**
>   I went with the minimal `ProviderId` which is an object containing the provider name and the user ID for that provider. They can probably fetch everything else they need.

I imagine this could depend on the hook. In the "after signup hook", I imagine they would like to receive the newly create user object? And for pre-signup hook, if it executes before the actual singup but right after they provided their data, then they would probably like to receive that data and have the opportunity to modify it potentially? We did have specific people asking for these features because they had specific needs -> what were their use cases, what did they need from their hooks?

> * [ ]  **Do we want the hook to be async?**
>   Probably yes. They'd made want to send some info to an external API.

May also depend on hook, but I think we do.

> * [ ]  **Do we want the hook to _block_ while it executes?**
>   Aka the login and signup process won't continue until the hook code finished executing.

Yeah I think they should block, in the sense that `login` should `await` the hook.
Then they can choose in their hook if they also want to `await` stuff or not. Meaning that they can easily make the hook that blocks or not in that case, while if we don't wait for it, they can't make it blocking even if they want, which could be bad.

> * [ ]  **Do we use the same hooks (postLogin and postSignup) for both email, username and _OAuth_ providers?**
>   Maybe yes, since it's the simplest thing to do.

Hm good question. What are the use cases? Why would somebody even need these hooks, which info do they want / need? Maybe there is some difference in usage?

> * [ ]  **Do we have special hook for OAuth providers?**
>   I'm wondering about this since for OAuth providers there is a moment of signup and a moment of login, but _usually_ it's not that important of a distinction.

What do you mean by special hook?

> * [ ]  **Do we want some `preLogin` and `preSignup` hooks?**
>   The users haven't requested this, so maybe not add it immediately?

Interesting, I thought somebody did request something like this, but maybe I am remembering wrong. Pre login is probably not very interesting, I can't imagine a use case for that.
Pre signup I can possibly -> you want to do some additional check on the provided data, or maybe enrich it somehow? Although they can do validations already. But maybe some more complex stuff?

> * [ ]  **Can the `pre` hooks modify data or reject signup or login?**
>   If they can't, I don't see much point in having these hooks.

Yeah that is interesting, I think they should be able to.
So pre-login I can't imagine use case for and would probably not do.
pre-signup I think it could be useful, and it might be interesting to have it receive data and then return data, and it can also throw, therefore allowing it to modify data or to reject signup.

>   Closes [Add hooks after login / sign-in, to execute custom server code #1556](https://github.com/wasp-lang/wasp/issues/1556)

Martinsos:
@infomiho maybe this would be a good opportunity to do the "User Story" approach: list some use cases of how these would be used, so we can figure out the best API.

We have one user story in the https://github.com/wasp-lang/wasp/issues/1556, which also has link to discord convo. We have just one though, but we can reason about some other possible stories.
Once we have those written down, I think it will be easier to figure out the best API.

waspc/examples/todoApp/main.wasp
Martinsos: We didn't have `hooks` so far, so maybe I would skip that, just go with `auth.postLogin`, as you said. What do you like better?

Martinsos (commented):
Left first round of comments! I wonder if this should have still been an RFC in Notion, maybe it would have been easier to iterate on it? But we can also try to do it here no biggie.

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
```diff
@@ -0,0 +1,29 @@
...
+  | OnUserSignUpHookFn
+  | OnUserLoginHookFn
+  | OnUserLogoutHookFn
+
+type OnUserLoginHookFn = (context: {
+  providerId: ProviderId
+  prisma: PrismaClient
+}) => Promise<void> | void
+type OnUserSignUpHookFn = (context: {
+  providerId: ProviderId
```
Martinsos, line 14: Wondering, why not give them the whole User also? I guess they can fetch it on their own so there is no need?

Also, you are passing prisma. We normally either pass `entities`, or we let them import `prisma`. This is a third mechanism now -> why do this?

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
```diff
@@ -0,0 +1,29 @@
...
+  prisma: PrismaClient
+}) => Promise<void> | void
+type OnUserLogoutHookFn = (context: {
+  providerId: ProviderId
+  prisma: PrismaClient
+}) => Promise<void> | void
+
+type HookType = 'onUserSignUp' | 'onUserLogin' | 'onUserLogout'
+
+export function defineHook<HT extends HookType>(
```
Martinsos, line 24: What is the purpose of this function? Looks to me like it just returns the hook it takes, and it even stays the same type.

waspc/data/Generator/templates/server/src/auth/hooks.ts
```diff
@@ -0,0 +1,39 @@
...
+{=^ onUserLoginHook.isDefined =}
+export const onUserLoginHook = undefined
+{=/ onUserLoginHook.isDefined =}
+
+{=# onUserLogoutHook.isDefined =}
+{=& onUserLogoutHook.importStatement =}
+export const onUserLogoutHook = addExtraParamsToTheHookCall({= onUserLogoutHook.importIdentifier =})
+{=/ onUserLogoutHook.isDefined =}
+{=^ onUserLogoutHook.isDefined =}
+export const onUserLogout = undefined
```
Martinsos, line 27:
```suggestion
export const onUserLogoutHook = undefined
```

Is `Hook` missing here? I added it above.

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
```diff
@@ -0,0 +1,29 @@
+import { type ProviderId } from '../../auth/utils.js'
+import { type PrismaClient } from '@prisma/client'
+
+export type AuthHookFn =
```
Martinsos, line 4: I wonder what is the point of this type.
You say it is one of three things. But you have no way to distinguish between those three (no discriminant). So this makes sense only if you don't care which type it is, it is all the same. But then why do you even have 3 types below, why not collapse it all into a single type that is `function (context: { providerId, prisma }) ...` and that is it?
I still have to read more code, maybe it will make sense, but right now it looks like this split into three hook types that are then unified into one hook type would be the same if it was just one hook type from start.

waspc/data/Generator/templates/server/src/auth/hooks.ts
```diff
@@ -0,0 +1,39 @@
...
+{=/ onUserLogout.isDefined =}
+
+type InternalAuthHookFn = ({ providerId }: { providerId: ProviderId }) => Promise<void> | void
+
+function addExtraParamsToTheHookCall(hookFn: AuthHookFn): InternalAuthHookFn {
```
Martinsos, line 32: So you say this adds extra params, but it seems to me like it removes params -> it takes user's hook function, which takes providerId and prisma, and partially applies prisma to it, returning a function that takes only providerId, which is now internal auth hook fn.

Aha, maybe when you said "addExtraParams" you meant "applyExtraParams"? Btw we could go further and say "applyExtraArgs", since parameters are variables in the function declaration, and arguments are values passed to them.

Maybe `userAuthHookToInternalAuthHook` might also be a good name btw.

waspc/examples/todoApp/main.wasp
```diff
@@ -39,10 +39,10 @@ app todoApp {
         },
       },
     },
-    hooks: {
-      postSignup: import { postSingupHook } from "@src/auth/hooks.js",
-      postLogin: import { postLoginHook } from "@src/auth/hooks.js",
-    },
+    // hooks: {
```
Martinsos, line 42: What if we don't group them under `hooks`? We don't do it for other hooks in Wasp. Also, if they are using only one or two hooks, which is likely, they don't get much with `hooks` bring in front.

Martinsos (commented):
@infomiho left some initial comments! I just realized though it was draft PR, so maybe I was a bit early -> in that case no worries, disregard comments that don't make sense because it is still WIP.

infomiho:
@Martinsos I'll close this PR in the interest of starting fresh since we agreed on the 4 auth hooks we want to do in the Notion RFC (something like `beforeUserCreated`, `afterUserCreated`, `beforeOAuthRedirect`, `afterOAuthTokenReceived`)

I'll take your comments into consideration when I'm writing the implementation 👍

PR #1962, opened 2024-04-15
Reenable the reset command

### LINKED ISSUE #1787: `wasp db reset` command is not running -> it is missing from cli/exe/Main.hs
Either it was never added or it was dropped, although we have the implementation -> let's make sure to add it to cli/exe/Main.hs .
Also, let's add a test (e2e) to prevent this from happening again.

sodic:
Fixes #1787 

I'll handle CLI e2e tests in a different issue.


infomiho (approved):

PhreeMason:
This doesnt seem to work even with a fresh install of wasp. Do you think the older version of the cli is still being cached?

sodic:
Hi @PhreeMason, we haven't released this fix yet.

Until we do, you can execute this command from the root of your project:
```
npx prisma migrate reset --schema .wasp/out/db/schema.prisma --skip-generate --skip-seed
```
This is exactly what `wasp db reset` does under the hood.

Sorry for the inconvenience :)

Issue #1243, opened 2023-06-09
Add top level Error Boundary on frontend


Martinsos:
Suggestion by @Zeko369.
Error Boundary is a React concept - it is a component that catches errors from any child components and shows them on screen. We should add one top-level one on frontend (only during development), because although Vite mostly takes cares of errors, there are cases where Vite itself fails, and then we just get blank/white screen.

Martinsos:
This is not a component that comes with React, you have to do your own.

Here are React docs on how you can implement it: https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary .

There is also a nice library for this: https://github.com/bvaughn/react-error-boundary .

Martinsos:
Once we add it, we will likely want to use it in `createAuthRequiredPage.jsx`, in `data/Generator/templates/react-app/src/auth/pages/`. We return `<span> ... error ... </span>` there on error, but we should rather throw an error then and let it hit the ErrorBoundary.

Issue #1076, opened 2023-03-21
Get rid of an ugly loading message that appears when auth is being processed.


matijaSos:
Currently, when you use Wasp auth and a user logs in / signs up, the unstyled loading message appears on a blank white screen in the upper left corner.

Can we get rid of that and put something nicer by default, e.g. a spinner?

Martinsos:
"Offending" piece of code: https://github.com/wasp-lang/wasp/blob/20eb9e007a7afbf9a2edf5edde7e20b596f09be0/waspc/data/Generator/templates/react-app/src/auth/pages/createAuthRequiredPage.jsx#L21 .

We should either put something nicer, or a blank screen. Ideally, we would allow customizing it!

Martinsos:
Related discord convo: https://discord.com/channels/686873244791210014/1156236595842785321/1156236595842785321 .

Martinsos:
Dmitry also asked if we could fix this

PR #1968, opened 2024-04-16
Loading page

Martinsos:
This is a preliminary draft of a possible solution for https://github.com/wasp-lang/wasp/issues/1076 -> making our loading screen ("Loading...") customizable!

I looked first into Suspense, thinking that might be nice solution, and while I think it might be so ultimately, allowing us to not even have the ext import in main.wasp, but just having them define a Suspense in rootComponent, that is not a way to go for now since Suspense is still experimental in most libraries (including tanstack query).

So, for now I went with following solution:
You can define `app.pageLoader` in `main.wasp`, which is an ext import of a React component that takes props `status: 'loading' | 'error'` and `error: SomeErrorWhatever`, and then render loader component. So quite simple! And that loader is then used while loading the auth user for a page that requires them.

Btw I know Remix has some cool loader stuff, and that is something we should probably look into future, but I didn't want to get into a generalized solution for loading resources for pages now, just wanted to solve this with something tailored for the situation.

I got something working, I can confirm it works! What I would love is to get some feedback on the high-level idea and arch, does it make sense, is DX ok, did I put something in very wrong place, did I not use some technique we have that I didn't know about, ... .

EDIT: @infomiho informed me that I looked at v4 docs of tanstack query instead of v5 and Suspense is not experimental anymore :D. So, one direction could be to drop what I implemented here and go with Suspense + Error Boundary. That said we would need to have that newer version of tanstack query then, and we don't have it yet. We should also expand Wasp's react query support to provide stuff like useSuspenseQuery. We will have to upgrade anyway though. So, TLDR; we can drop this solution here now and use Suspense once we upgrate to tanstack v5. Or go with this and anyway upgrade to Suspense once we get to tanstack v5.

To give a bit more clarity on Suspense / Error Boundary approach, this is what I think is needed to get that done:
1. Upgrade React Query to v5. This is not a small move, as we will also probably want to update our wrapping of useQuery to support stuff like useSuspenseQuery and similar. Maybe that is just adding more API, or maybe we rethink how we do it, not sure yet.
2. At the very top level of our React tree, define our ErrorBoundary and Suspense components. This is something users can't change. They would have nice default error and loading components. They would be at the similar place as we have react query provider component. This would also take care of https://github.com/wasp-lang/wasp/issues/1243 .
3. Modify `useAuth` to use `useSuspenseQuery`, that would then trigger/use the ReactBoundary and Suspense.
4. Finally, document that if they want to override how loading is handled, they can define their own Suspense using rootComponent Wasp feature. They can also define their own ErrorBoundary there if they want.

This approach will provide the same functionality as the one I implemented in this PR, but it will be more React specific, less general. Still, it will maybe be a bit more natural for React, and we don't have to have add it to main.wasp, which might be ok since this is not such an interesting feature. The main problem is that we have to wait for React Query 5 upgrade on our side.

TLDR; I am a bit tempted to right now drop the changes I did in this PR, just remove Loading... and let it be a white screen, and then we use Suspense once we upgrade React Query to v5.

TODO:
 - [ ] Solve TODOs from the code.
 - [ ] Make DefaultAppLoader nicer. 
 - [ ] Write docs.
 - [ ] Update Changelog.
 - [ ] Make sure e2e tests and unit tests are passing.
 - [ ] Add this to some example app.


waspc/data/Generator/templates/react-app/src/auth/pages/DefaultPageLoader.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+// TODO: Should I take this out of auth/, into more general place?
```
Martinsos, line 3: The question is also, do we want to go with
`app.pageLoader` in `main.wasp`, or `app.auth.pageLoader`.
The thing is, concept of page loader could be generic. While resources are loading for a page that block its showing, we use page loader.
However, we currently use that only to fetch auth user.
So we could also go with it being more specialized.
I am not sure. I think I would immediately go general from the DX side, while internally we will apply it only in the auth user situation. Which means `app.pageLoader`, not `app.auth.pageLoader`. And also, this file here, would go some levels up.

waspc/data/Generator/templates/react-app/src/auth/pages/DefaultPageLoader.jsx
```diff
@@ -0,0 +1,13 @@
...
+export const DefaultPageLoader = ({status, error}) => {
```
Martinsos, line 4: I would love to throw some TS here, to specify status is supposed to be `'loading' | 'error'`. How should I go about that though? Provide them with a type they can use? Provide them with a hook they use to define this component? Any advice, something I should look at and copy?

waspc/data/Generator/templates/sdk/wasp/auth/useAuth.ts
```diff
@@ -20,6 +20,7 @@ function createUserGetter() {
   async function getMe(): Promise<AuthUser | null> {
     try {
       const response = await api.get(getMeRoute.path)
+      await new Promise((resolve) => setTimeout(resolve, 1000));
```
Martinsos, line 23: I got to remove this, this is just while testing so I can see the loader nicely.

sodic:
> TLDR; I am a bit tempted to right now drop the changes I did in this PR, just remove Loading... and let it be a white screen, and then we use Suspense once we upgrade React Query to v5.

@Martinsos I agree. I don't think we should introduce a new mechanism now since it seems the only think stopping us from using the obvious solution is our version of `tanstack-query` (which we should update as soon as possible anyway).

To bad, I think you envisioned a nice system :)

Martinsos:
Convo with Boris, on how he does it: https://discord.com/channels/686873244791210014/1080864617347162122/1230477928102498365 -> shows blurred overlay + spinner while also rendering a component which is implemented in such a way that it doesn't break in the background.

Not sure we can do this because we can't assume this about their components, but interesting to consider still.

waspc/data/Generator/templates/react-app/src/auth/pages/createAuthRequiredPage.jsx
```diff
@@ -4,25 +4,31 @@ import React from 'react'
 import { Redirect } from 'react-router-dom'
 import { useAuth } from 'wasp/client/auth'
 
+{=# pageLoader.isDefined =}
+{=& pageLoader.importStatement =}
+// TODO: Do we have a way here to simply provide an alias?
```
infomiho, line 9: That's why I should finish this thing, it's close to being completed! https://github.com/wasp-lang/wasp/pull/1865

waspc/data/Generator/templates/react-app/src/auth/pages/DefaultPageLoader.jsx
infomiho: > Provide them with a type they can use? 

You simply type the props of the function to be:
```ts
export function DefaultPageLoader(
  { status, error }: { status: 'loading' | 'error';  error: { message?: string } }
) { ... }
```

infomiho: Oh, you want to provide them a way to define a component that will fit in our place of use i.e. [createAuthRequiredPage.jsx](https://github.com/wasp-lang/wasp/pull/1968/files#diff-8cb8bc1efac0cbcc354c11abe7f36616f7096a8369e6282eecf51cb1077f08ee)

Using some function like `definePageLoaderComponent` could be a could idea!

Martinsos: Yup that is what I meant! But ok, not doing this now, still thanks for letting me know.

PR #1977, opened 2024-04-19
add codespaces template that points to todo tutorial app

vincanger:
### Description

By adding the `devcontainer.json` users will be able to try out our todo tutorial app by just clicking `Code > Codespaces > Create a codspace...`


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,27 @@
...
+  "customizations": {
+    "vscode": {
+      "extensions": ["wasp-lang.wasp", "PKief.material-icon-theme"],
+      "settings": {
+        "workbench.iconTheme": "material-icon-theme"
+      }
+    }
+  },
+  "containerEnv": {
+    "WASP_TELEMETRY_CONTEXT": "codespaces"
```
Martinsos, line 13: Good one, we actually use "codespaces" in our analytics! So this will pick it up correctly.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,27 @@
...
+  "portsAttributes": {
+    "3000": {
+      "label": "React-Vite Client"
+    },
+    "3001": {
+      "label": "NodeJS Server"
+    }
+  },
+  "postCreateCommand": "curl -sSL https://get.wasp-lang.dev/installer.sh | sh -s",
+  "postStartCommand": "printf 'export WASP_WEB_CLIENT_URL=https://$CODESPACE_NAME-3000.app.github.dev\n' >> $HOME/.bashrc && printf 'export REACT_APP_API_URL=https://$CODESPACE_NAME-3001.app.github.dev\n' >> $HOME/.bashrc && echo 'gh codespace ports visibility 3000:public -c $CODESPACE_NAME\n' >> $HOME/.bashrc && echo 'gh codespace ports visibility 3001:public -c $CODESPACE_NAME\n' >> $HOME/.bashrc",
```
Martinsos, line 25: Sounds like this one runs on every start of codespace? If so, we will be adding this stuff to .bashrc again and again on every start, won't we? Won't .bashrc keep growing and growing then, if its content is being persisted? I don't know how codespaces work nor what postStartCommand does so I might be wrong, but it seemed like this might be the case.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,27 @@
...
+  "workspaceFolder": "/workspaces/wasp/examples/tutorials/TodoApp"
```
Martinsos, line 26: How do you determine this folder, what does it mean?

Martinsos (approved):
Looking good @vincanger !

Left a couple of minor comments/questions.

What if we also wanted to offer more codespace for Wasp apps -> for example for other Wasp apps, or maybe even for just starting a completely fresh Wasp app, how would we do that? Would we just define more stuff here in .devcontainer? How does this work?

.devcontainer/devcontainer.json
vincanger: You're right. I switched the wasp install to `onCreateCommand` and the env var exporting to `postCreateCommand`

.devcontainer/devcontainer.json
vincanger: This means that the codespace editor will open up the directory at that path, which is the tutorial todo app, rather than the entire main wasp repo

PR #1984, opened 2024-04-22
Loading page

Martinsos:
This is a follow up on https://github.com/wasp-lang/wasp/pull/1968 .

We decided not to implement custom loading screen until we have tanstack-query v5. Once we do, we will implement it via Suspense.

In the meantime however, I wanted to at least replace the "Loading..." with something nicer, since people complained it looks weird. So even though they can't customize it easily, at least it wouldn't look like a page that somebody forgot to style.

I didn't do anything special, just replaced the "Loading..." with a nice spinner and put it into a DefaultLoader component.

loading-page

Once we move to Suspense, we can just continue using this DefaultLoader component as a default Suspense.fallback.

There are some things I am not sure about and wanted to ask for advice:
1. I used stitches for CSS in JS. However, I put DefaultLoader.tsx into client framework code (react-app/src). But I see that stitches is imported and configured in SDK. What should I do, is it ok that I am importing it from the SDK? I guess that is fine? Is it ok that DefaultLoader.tsx is in client framework code, or should I move it to SDK for some reason?
2. Following up on this, stitches is currently imported as dep for SDK only if Auth is used. But, this DefaultLoader, I put it outside of auth, as a standalone thing, that exists regardless of auth being used or not. Truth is, right now it is used only if Auth is used, but it could be used for something else also, and once we add Suspense, it will indeed be used for any loading that hits Suspense, not just for auth. So it made sense to me to already make it a general concept, and not auth-specific concept. So is it ok if I just say that stitches is always imported as a dep, regardless of auth being used or not?

TLDR; Where do I put DefaultLoader.tsx, in framework code or SDK code? If it is in framework code, is it ok that it imports stitches from SDK? Is it ok if I say stitches is always imported in the SDK, not just when auth is used (as is now)? Is it ok that I treat DefaultLoader.tsx as a general concept, not just auth-related concept? I think answer to all of these is yes, but I wanted to check if anybody has different opinion.

Btw I know that stitches are deprecated and we want to replace them with something else, but until that happens, I am just using stitches here.


waspc/data/Generator/templates/react-app/src/DefaultLoader.tsx
```diff
@@ -0,0 +1,37 @@
+import { styled, keyframes } from 'wasp/core/stitches.config'
+
+const fullRotationKeyframes = keyframes({
+  '0%': { transform: 'rotate(0deg)' },
+  '100%': { transform: 'rotate(360deg)' }
+});
+
+// TODO: Do I need to add !important on all of this stuff because of Tailwind css?
```
Martinsos, line 8: @infomiho any idea, do I need to add !important here?

waspc/data/Generator/templates/sdk/wasp/auth/useAuth.ts
```diff
@@ -20,6 +20,7 @@ function createUserGetter() {
   async function getMe(): Promise<AuthUser | null> {
     try {
       const response = await api.get(getMeRoute.path)
+      await new Promise((resolve) => setTimeout(resolve, 1000));
```
Martinsos, line 23: TODO: Remove this before merging.

waspc/src/Wasp/Generator/WebAppGenerator/AuthG.hs
Martinsos, line 1: Nothing really changed in this file. I was doing some changes, but reverted all of them, and was left only with these tiny (formatting) improvements.

waspc/data/Generator/templates/react-app/src/auth/pages/createAuthRequiredPage.jsx
```diff
@@ -4,28 +4,27 @@ import React from 'react'
...
+      case 'success':
+        if (user) {
+          return (
+            <Page {...props} user={user} />
+          );
+        } else {
+          return <Redirect to="{= onAuthFailedRedirectTo =}" />;
+        }
+      case 'loading':
+        return <DefaultLoader/>;
```
Martinsos, line 23: This is the main change here, using `DefaultLoader` here.

waspc/data/Generator/templates/react-app/src/DefaultLoader.tsx
infomiho: Yep, you should enable Tailwind for some test project and see what happens :) 

On the topic of which loader to pick, here are some example from ShadCN, NextUI, Chakra and DaisyUI

https://github.com/wasp-lang/wasp/assets/2223680/ac8d4f98-e952-4834-8207-23b9eec432da


https://github.com/wasp-lang/wasp/assets/2223680/b6f616aa-53c6-4db7-9703-e664a5bc11f0


https://github.com/wasp-lang/wasp/assets/2223680/108e1781-af0f-4e2b-ab21-56245b53605e


https://github.com/wasp-lang/wasp/assets/2223680/af78d614-fd42-4ea5-a23c-edd23a1b4636

They are all using a "single piece of a circle spinning" which I think we should also go for! So our aesthetic is as modern as possible.

sodic:
> Where do I put DefaultLoader.tsx, in framework code or SDK code?

If the user never needs it, it belongs in the framework code. The only reasons to put something in the SDK are:
1. It's public API - the user needs it directly.
2. The public API needs it - the user needs it indirectly.

>  If it is in framework code, is it ok that it imports stitches from SDK?

Not a problem, this happens all the time.

> Is it ok if I say stitches is always imported in the SDK, not just when auth is used (as is now)?

I think so. If we need it, we need it.

> Is it ok that I treat DefaultLoader.tsx as a general concept, not just auth-related concept?

Yes, I even think it's better to treat it as a general concept.

> I think answer to all of these is yes, but I wanted to check if anybody has different opinion.

Not me :smile:

waspc/data/Generator/templates/react-app/src/DefaultLoader.tsx
infomiho: Here's a PR with the Chakra UI https://github.com/wasp-lang/wasp/pull/1990

waspc/data/Generator/templates/react-app/src/DefaultLoader.tsx
```diff
@@ -0,0 +1,37 @@
+import { styled, keyframes } from 'wasp/core/stitches.config'
```
infomiho, line 1: I'd also maybe suggest putting this file in `components/` dir so it's clear it's a component a not some top level file like `router.tsx or `index.tsx`.

waspc/data/Generator/templates/react-app/src/auth/pages/createAuthRequiredPage.jsx
```diff
@@ -4,28 +4,27 @@ import React from 'react'
...
+    case 'error':
+      return <span>An error occurred. Please refresh the page. {error?.message}</span>
```
infomiho, line 25: I'd suggest using the same error compnent as we used in the `OAuthCallbackPage`.

This component is not extracted in a separate file, but I can do that now in a simple PR.

infomiho: Here's the PR extracting the component: https://github.com/wasp-lang/wasp/pull/1991

I'd recommend using `<MessageError>` instead of the `<span>` to keep the look consistent :)

Martinsos (commented):
A bit later (after merging), but here is my review on the changes you did @infomiho !

PR #1993, opened 2024-04-24
Auth Hooks

### LINKED ISSUE #1556: Add hooks after login / sign-in, to execute custom server code

infomiho:
Closes #1556 

Adding hooks:
- onBeforeSignup
- onAfterSignup
- onBeforeOAuthRedirect
- ~~onAfterOAuthTokenReceived~~

Left to do:
- [x] Changelog
- [x] Docs
- [x] Headless tests https://github.com/wasp-lang/wasp/pull/2022


waspc/examples/todoApp/src/auth/hooks.ts
```diff
@@ -0,0 +1,30 @@
+import type {
```
infomiho, line 1: Example usage of the hooks

waspc/examples/todoApp/main.wasp
```diff
@@ -40,7 +40,11 @@ app todoApp {
       },
     },
     onAuthFailedRedirectTo: "/login",
-    onAuthSucceededRedirectTo: "/profile"
+    onAuthSucceededRedirectTo: "/profile",
+    onBeforeSignup: import { onBeforeSignup } from "@src/auth/hooks.js",
```
infomiho, line 44: Declare the hooks in the `auth` dict

waspc/data/Generator/templates/server/src/auth/providers/config/github.ts
```diff
@@ -80,10 +80,8 @@ const _waspConfig: ProviderConfig = {
             stateTypes: ['state'],
             userSignupFields: _waspUserSignupFields,
             getAuthorizationUrl: ({ state }) => github.createAuthorizationURL(state, config),
-            getProviderInfo: async ({ code }) => {
-                const { accessToken } = await github.validateAuthorizationCode(code);
-                return getGithubProfile(accessToken);
-            },
+            getProviderTokens: ({ code }) => github.validateAuthorizationCode(code),
```
infomiho, line 83: Refactor OAuth a bit to split getting tokens and getting user profile into two steps so it's easier to write a hook `onAfterOAuthTokenReceived`

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
```diff
@@ -0,0 +1,46 @@
+import type { Request as ExpressRequest } from 'express'
+import type { ProviderId, createUser } from '../../auth/utils.js'
+import { prisma } from '../index.js'
+
+type CommonInput = {
```
infomiho, line 5: Each hook will receive `hookName`, `prisma` and Express's `req` object.

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
```diff
@@ -0,0 +1,46 @@
...
+  params: OnAfterSignupHookFnInput,
+) => Promise<void>
+
+type OnAfterSignupHookFnInput = {
+  providerId: ProviderId
+  user: Awaited<ReturnType<typeof createUser>>
+} & CommonInput
+
+/* On Before OAuth Redirect Hook */
+export type OnBeforeOAuthRedirectHookFn = (
```
infomiho, line 31: Users have the ability to modify the OAuth redirect URL

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
```diff
@@ -0,0 +1,46 @@
...
+import { prisma } from '../index.js'
+
+type CommonInput = {
+  hookName: string
+  prisma: typeof prisma
+  req: ExpressRequest
+}
+
+/* On Before Signup Hook */
+export type OnBeforeSignupHookFn = (
```
infomiho, line 12: I haven't added the ability for developers to modify the data that's used to create the `user`. I'm not sure if it's useful since we also have `userSignupFields` 🤷

waspc/data/Generator/templates/server/src/auth/hooks.ts
```diff
@@ -0,0 +1,97 @@
...
+{=& onAfterSignupHook.importStatement =}
+{=/ onAfterSignupHook.isDefined =}
+{=# onBeforeOAuthRedirectHook.isDefined =}
+{=& onBeforeOAuthRedirectHook.importStatement =}
+{=/ onBeforeOAuthRedirectHook.isDefined =}
+{=# onAfterOAuthTokenReceivedHook.isDefined =}
+{=& onAfterOAuthTokenReceivedHook.importStatement =}
+{=/ onAfterOAuthTokenReceivedHook.isDefined =}
+
+/*
```
infomiho, line 22: We wrap the user defined functions so we don't have to call them with common params like `prisma` or `hookName` when invoking the hooks in our server code.

waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
```diff
@@ -1,150 +1,167 @@
-import { Request, Response } from 'express';
```
infomiho, line 1: This file was auto formatted, but the only changes are calls to the `onBeforeSignup`
and `onAfterSignup` hooks.

waspc/data/Generator/templates/server/src/auth/providers/oauth/handler.ts
```diff
@@ -55,54 +68,82 @@ export function createOAuthProviderRouter<ST extends StateType>({
...
+  ) => Promise<URL>,
 ) {
-  return handleRejection(async (_req, res) => {
-    const oAuthState = generateAndStoreOAuthState(stateTypes, provider, res);
-    const url = await getAuthorizationUrl(oAuthState);
-    return redirect(res, url.toString());
+  return handleRejection(async (req, res) => {
+    const oAuthState = generateAndStoreOAuthState(stateTypes, provider, res)
+    let url = await getAuthorizationUrl(oAuthState)
+    if (onBeforeOAuthRedirectHook) {
```
infomiho, line 89: Hook

waspc/data/Generator/templates/server/src/auth/providers/oauth/handler.ts
```diff
@@ -55,54 +68,82 @@ export function createOAuthProviderRouter<ST extends StateType>({
...
   }>,
 ) {
   return handleRejection(async (req, res) => {
     try {
-      const oAuthState = validateAndGetOAuthState(stateTypes, provider, req);
-      const { providerProfile, providerUserId } = await getProviderInfo(oAuthState);
+      const oAuthState = validateAndGetOAuthState(stateTypes, provider, req)
+      const { accessToken } = await getProviderTokens(oAuthState)
+      if (onAfterOAuthTokenReceivedHook) {
+        await onAfterOAuthTokenReceivedHook({
```
infomiho, line 120: Hook

waspc/data/Generator/templates/server/src/auth/providers/oauth/user.ts
```diff
@@ -73,13 +77,19 @@ async function getAuthIdFromProviderDetails(
     // For now, we don't have any extra data for the oauth providers, so we just pass an empty object.
     const providerData = await sanitizeAndSerializeProviderData({})
   
+    if (onBeforeSignupHook) {
```
infomiho, line 80: Hook

waspc/data/Generator/templates/server/src/auth/providers/oauth/user.ts
```diff
@@ -73,13 +77,19 @@ async function getAuthIdFromProviderDetails(
...
+      await onBeforeSignupHook({ req, providerId })
+    }
     const user = await createUser(
       providerId,
       providerData,
       // Using any here because we want to avoid TypeScript errors and
       // rely on Prisma to validate the data.
       userFields as any,
     )
+    if (onAfterSignupHook) {
```
infomiho, line 90: Hook

waspc/data/Generator/templates/server/src/auth/providers/username/signup.ts
```diff
@@ -34,13 +35,19 @@ export function getSignupRoute({
     })
   
     try {
-      await createUser(
+      if (onBeforeSignupHook) {
```
infomiho, line 38: Hook

waspc/data/Generator/templates/server/src/auth/providers/username/signup.ts
```diff
@@ -34,13 +35,19 @@ export function getSignupRoute({
...
+        await onBeforeSignupHook({ req, providerId })
+      }
+      const user = await createUser(
         providerId,
         providerData,
         // Using any here because we want to avoid TypeScript errors and
         // rely on Prisma to validate the data.
         userFields as any
       )
+      if (onAfterSignupHook) {
```
infomiho, line 48: Hook

waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
```diff
@@ -1,150 +1,167 @@
...
+    )
 
-        const newUserProviderData = await sanitizeAndSerializeProviderData<'email'>({
-            hashedPassword: fields.password,
-            isEmailVerified: isEmailAutoVerified ? true : false,
-            emailVerificationSentAt: null,
-            passwordResetSentAt: null,
-        });
+    try {
+      if (onBeforeSignupHook) {
```
infomiho, line 121: Hook

waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
```diff
@@ -1,150 +1,167 @@
...
+        await onBeforeSignupHook({ req, providerId })
+      }
+      const user = await createUser(
+        providerId,
+        newUserProviderData,
+        // Using any here because we want to avoid TypeScript errors and
+        // rely on Prisma to validate the data.
+        userFields as any,
+      )
+      if (onAfterSignupHook) {
```
infomiho, line 131: Hook

waspc/e2e-test/Tests/WaspComplexTest.hs
```diff
@@ -166,47 +166,69 @@ addServerEnvFile = do
...
       unlines
         [ "  auth: {",
           "    userEntity: User,",
-          "    externalAuthEntity: SocialLogin,",
           "    methods: {",
           "      google: {}",
           "    },",
-          "    onAuthFailedRedirectTo: \"/login\"",
+          "    onAuthFailedRedirectTo: \"/login\",",
+          "    onBeforeSignup: import { onBeforeSignup } from \"@src/auth/hooks.js\",",
```
infomiho, line 186: I've included hooks in the e2e test

waspc/e2e-test/Tests/WaspComplexTest.hs
```diff
@@ -166,47 +166,69 @@ addServerEnvFile = do
...
-          "  username                  String        @unique",
-          "  password                  String",
-          "  externalAuthAssociations  SocialLogin[]",
           "psl=}"
         ]
 
-    socialLoginEntity =
+    hooksFile =
       unlines
-        [ "entity SocialLogin {=psl",
```
infomiho, line 201: I've cleaned up the e2e tests not use old auth concepts 👍

web/docs/auth/auth-hooks.md
```diff
@@ -0,0 +1,293 @@
...
+    ...
+    onBeforeSignup: import { onBeforeSignup } from "@src/auth/hooks",
+  },
+}
+```
+
+```ts title="src/auth/hooks.ts"
+import type { OnBeforeSignupHookFn } from 'wasp/server/auth'
+
+export const onBeforeSignup: OnBeforeSignupHookFn = async ({ providerId, prisma, req, hookName }) => {
```
infomiho, line 94: TODO: add better example e.g. throw a `HttpError` to show an error message on the client.

web/docs/auth/auth-hooks.md
```diff
@@ -0,0 +1,293 @@
...
+  auth: {
+    ...
+    onAfterSignup: import { onAfterSignup } from "@src/auth/hooks",
+  },
+}
+```
+
+```js title="src/auth/hooks.js"
+export const onAfterSignup = async ({ providerId, user, prisma, req, hookName }) => {
+  // Do something after signup
```
infomiho, line 139: TODO: add a better example e.g. set a field on the user entity or send data to an external service.

web/docs/auth/auth-hooks.md
```diff
@@ -0,0 +1,293 @@
...
+  auth: {
+    ...
+    onBeforeOAuthRedirect: import { onBeforeOAuthRedirect } from "@src/auth/hooks",
+  },
+}
+```
+
+```js title="src/auth/hooks.js"
+export const onBeforeOAuthRedirect = async ({ url, prisma, req, hookName }) => {
+  // Do something before OAuth redirect
```
infomiho, line 202: TODO: show a useful example of setting something in the query params and getting it back in the callback.

infomiho:
@sodic and I jumped on quick call to discuss some of the OAuth hooks since I felt I didn't design them to be _that useful_.

We concluded the following:
- we should remove the `onAfterOAuthTokenReceived` hook since the developer doesn't have access to `user` and the `accessToken` at the same time. This makes the `accessToken` useless if you want to save it on some user for later use.
  - **Solution:** We'll just forward the `accessToken` in the `onAfterSignup` hook.
- we should investigate how to have "state" when using the `onBeforeOAuthRedirect` hook. This is the only OAuth hook where the developer can get some user input from the client. This input gets lost in the follow up hooks e.g. `onAfterSignup` doesn't have access to it, so it never can be saved on the `user` entity.
  - **Idea:** It would be ideal to somehow **keep track of the user input across the OAuth hooks** so the it can be used again in the `onAfterSignup` hook.
  - I'll investigate how to best achieve this.

waspc/data/Generator/templates/server/src/auth/providers/config/github.ts
```diff
@@ -77,13 +77,11 @@ const _waspConfig: ProviderConfig = {
 
         return createOAuthProviderRouter({
             provider,
-            stateTypes: ['state'],
+            optionalStateTypes: [],
```
infomiho, line 80: Since the `state` is _always_ sent and we rely on that knowledge in the new hooks, I've refactored this a bit to reflect the fact that the `state` will be always present in the oAuthState object.

waspc/data/Generator/templates/server/src/auth/providers/oauth/state.ts
```diff
@@ -1,29 +1,31 @@
...
+  const result = {} as {
+    [name in OptionalStateType | RequiredStateType]: string;
+  };
 
-  if (stateTypes.includes('state' as ST)) {
-    const state = generateState();
-    setOAuthCookieValue(provider, res, 'state', state);
-    result.state = state;
-  }
+  const state = generateState();
```
infomiho, line 24: Always set the state since it's always used and will always be used for each future OAuth provider.

waspc/data/Generator/templates/server/src/auth/providers/oauth/cookies.ts
```diff
@@ -4,14 +4,14 @@ import {
...
+import type { ProviderConfig } from "wasp/auth/providers/types";
 
-import { type StateType } from './state';
+import type { OptionalStateType, RequiredStateType } from './state';
 
 export function setOAuthCookieValue(
   provider: ProviderConfig,
   res: ExpressResponse,
-  stateType: StateType,
+  stateType: RequiredStateType | OptionalStateType,
```
sodic, line 14: Seems like we could make a `type StateType = RequiredStateType | OptionalStateType`

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
sodic: Let's add it when it proves necessary.

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
```diff
@@ -0,0 +1,42 @@
...
```
sodic, line 12: I suggest dropping the `Fn` part, `OnBeforeSignupHook` should be enough.

I know we have `SeverSetupFn` and whatnot, but I'm not a fan of those either. I'd change them to `SetUpServer` or something.

Of course, changing this know makes our public API names inconsistent. If you think keeping the consistency is more important, I'm all for it.

I'd probably sacrifice consistency in this case, but it's your call.

sodic (commented):
Gotta run now, will finish the review tomorrow morning.

Nice work with this!

sodic (changes requested):
Ok, finally done with everything :smile: 

I didn't try the playing around with the feature, I trust it works. Let me know if you still want me to try it out.

All in all, this is looking pretty nicely. I left a lot of comments mostly because of:
 - The docs.
 - Refactoring suggestions.
 - The fact that I'm the only one reviewing this.

I left many comments that talk about the previous version of the code. Decide what you want to tackle and what not.

As for the stuff that must be fixed, it's just one type error (`ST`  instead of `OST`, you'll know when you find the comment). And please figure out why that's not failing anywhere.

The other things are all open for discussion, I can't remember I found any proper errors.

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
```diff
@@ -0,0 +1,42 @@
...
+) => Promise<void>
+
+type OnBeforeSignupHookFnInput = {
+  providerId: ProviderId
+} & CommonInput
+
+/* On After Signup Hook */
+export type OnAfterSignupHookFn = (
+  params: OnAfterSignupHookFnInput,
+) => Promise<void>
```
sodic, line 23: What happens when a user wants the hook to be a non-async function.

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
infomiho: Renamed the symbol and added the `Expand` helper so user would see full types
Screenshot 2024-06-05 at 16 51 33

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
```diff
@@ -0,0 +1,42 @@
...
+  prisma: typeof prisma
+  req: ExpressRequest
+}
+
+/* On Before Signup Hook */
+export type OnBeforeSignupHookFn = (
+  params: OnBeforeSignupHookFnInput,
+) => Promise<void>
+
+type OnBeforeSignupHookFnInput = {
```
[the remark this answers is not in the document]
infomiho: I've added a comment explaining why we didn't export the `*Input` types.

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
```diff
@@ -0,0 +1,42 @@
...
+} & CommonInput
+
+/* On After Signup Hook */
+export type OnAfterSignupHookFn = (
+  params: OnAfterSignupHookFnInput,
+) => Promise<void>
+
+type OnAfterSignupHookFnInput = {
+  providerId: ProviderId
+  user: Awaited<ReturnType<typeof createUser>>
```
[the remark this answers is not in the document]
infomiho: It seems to be working for this use-case 👍 I'll keep it vanilla

waspc/data/Generator/templates/server/src/auth/hooks.ts
```diff
@@ -0,0 +1,79 @@
...
+  We pass extra params to the user defined hook functions, but we don't want to
+  pass 'prisma' and 'hookName' to them when we call them in the server code.
+  So, we need to remove them from the params object which is used to define the
+  internal hook functions.
+*/
+type UserHookParamsToInternalHookParams<T> = T extends (
+  params: infer P,
+) => unknown
+  ? Omit<P, 'prisma' | 'hookName'>
```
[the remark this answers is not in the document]
infomiho: That would work we didn't have a param we want to keep from the `CommonInput` which is `req`.

Common input has `hookName`, `prisma` and `req`, we want to keep out the first two. If I used `keyof` to get the keys of the `CommonInput`, I'd still need to write down the literal `req` which would still be duplication.

Did I understand this comment correctly?

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
infomiho: Yep, we need to account for that 👍

waspc/data/Generator/templates/server/src/auth/hooks.ts
```diff
@@ -0,0 +1,79 @@
...
+  params: UserHookParamsToInternalHookParams<OnBeforeSignupHookFn>,
+) =>
+  {= onBeforeSignupHook.importIdentifier =}({
+    hookName: 'onBeforeSignup',
+    prisma,
+    ...params,
+  })
+{=/ onBeforeSignupHook.isDefined =}
+{=^ onBeforeSignupHook.isDefined =}
+export const onBeforeSignupHook = undefined
```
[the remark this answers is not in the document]
infomiho: I've went with no-op functions 👍

waspc/data/Generator/templates/server/src/auth/hooks.ts
```diff
@@ -0,0 +1,79 @@
...
+export const onBeforeOAuthRedirectHook = undefined
+{=/ onBeforeOAuthRedirectHook.isDefined =}
+
+/*
+  We pass extra params to the user defined hook functions, but we don't want to
+  pass 'prisma' and 'hookName' to them when we call them in the server code.
+  So, we need to remove them from the params object which is used to define the
+  internal hook functions.
+*/
+type UserHookParamsToInternalHookParams<T> = T extends (
```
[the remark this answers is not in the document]
infomiho: I've went with your suggestion 👍 it makes the code cleaner

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
infomiho: I've adjusted the return type to allow for async and sync functions.

web/docs/auth/auth-hooks.md
```diff
@@ -0,0 +1,359 @@
...
+    This is the same value that will be passed to `onAfterSignup` hook and can be used to save some user data temporarily that can be used later.
+- `prisma: PrismaClient`
+    
+    Prisma client instance. You can use it to query your database.
+- `req: Request`
+
+  Express request object. You can access the request headers, cookies, etc. from this object.
+- `hookName: string`
+    
+  The name of the hook that is being called (in this case, it will be `'onBeforeOAuthRedirect'`)
```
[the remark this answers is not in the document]
infomiho: I'll add an extra explanation: `This value can be useful when logging or debugging.`

web/docs/auth/auth-hooks.md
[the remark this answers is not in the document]
infomiho: You are correct, I'll split the page into more "guide"-like content and the more API reference content.

waspc/data/Generator/templates/server/src/auth/providers/oauth/state.ts
```diff
@@ -32,37 +34,43 @@ export function generateAndStoreOAuthState<ST extends StateType>(
   return result;
```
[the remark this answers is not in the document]
infomiho: I've adjusted the code to be simpler and generate the `result` all the once :D

waspc/data/Generator/templates/server/src/auth/providers/oauth/cookies.ts
infomiho: I've simplified the logic so this no longer applies.

web/static/img/auth-hooks/signup_flow_with_hooks.png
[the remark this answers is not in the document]
infomiho: I've simplified the diagrams and dropped all of the "data" rects.

web/static/img/auth-hooks/oauth_flow_with_hooks.png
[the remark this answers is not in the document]
infomiho: I've made it into a single action that reads okay. 

> Also, why do we need the rectangle with oauth: token below if we can't use it? Can we drop that?

I wanted to point out where the data that appears in the hooks comes from. But it was getting crowded so I dropped all the data rects.

web/docs/auth/auth-hooks.md
```diff
@@ -0,0 +1,359 @@
+---
+title: Auth Hooks
+---
+
+import { EmailPill, UsernameAndPasswordPill, GithubPill, GooglePill, KeycloakPill } from "./Pills";
+import ImgWithCaption from '@site/blog/components/ImgWithCaption'
+
+Sometimes you need to do some action _before_ the user signs up or _after_ they sign up. Or you need to access the OAuth redirect URL or the OAuth access token. For these cases, you can use the Auth Hooks.
```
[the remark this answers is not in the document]
sodic: I really like the change, short and clear. Nice work!

web/static/img/auth-hooks/signup_flow_with_hooks.png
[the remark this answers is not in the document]
sodic: Ok, I like the new version better than the old one.

It might be a little too clean though. Perhaps it makes sense keeping the data rectangles below the hook name (to say what the hook sees). I liked those.

Either way, your call. Resolve when happy.

web/docs/auth/auth-hooks.md
```diff
@@ -0,0 +1,359 @@
...
+    }
+    someKindOfStore.delete(id)
+  }
+}
+```
+
+</TabItem>
+</Tabs>
+
+### `onBeforeOAuthRedirect` hook
```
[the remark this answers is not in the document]
sodic: I like the changes. Here's an extra idea.

You have a chapter "When are the hooks called" where you show the graphs. This chapter is followed by "Available hooks" which lists all the available hooks. This feels a little backwards.

Also, we first have "Using hooks" and only then do we tell the reader what they can do.

I recommend swapping the order of the chapters or even having this structure:
```
What can I do with hooks (list of hooks followed by the graph)
Using hooks
  Executing code ...
  Executing code ...
  Executing code...
API Reference
```
Take what you like out of here and resolve when you are happy.

web/docs/auth/auth-hooks.md
```diff
@@ -0,0 +1,590 @@
...
+
+Let's go through each of them.
+
+### Executing code before the user signs up
+
+Wasp calls the `onBeforeSignup` hook before the user is created.
+
+The `onBeforeSignup` hook can be useful if you want to reject a user based on some criteria before they sign up.
+
+Works with <EmailPill /> <UsernameAndPasswordPill /> <GithubPill /> <GooglePill /> <KeycloakPill />
```
sodic, line 90:
```suggestion
This hook works with with <EmailPill /> <UsernameAndPasswordPill /> <GithubPill /> <GooglePill /> <KeycloakPill />
```
To avoid a sentence fragment.

web/docs/auth/auth-hooks.md
```diff
@@ -0,0 +1,359 @@
...
+
+    The URL object that will be used for the OAuth redirect.
+- `uniqueRequestId: string`
+
+    The unique request ID for the OAuth flow. (Some might know it as `state` parameter in OAuth.)
+
+    This is the same value that will be passed to `onAfterSignup` hook and can be used to save some user data temporarily that can be used later.
+- `prisma: PrismaClient`
+    
+    Prisma client instance. You can use it to query your database.
```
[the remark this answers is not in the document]
sodic: Some of these guys are still missing articles.

web/docs/auth/auth-hooks.md
```diff
@@ -0,0 +1,359 @@
...
+    onBeforeSignup: import { onBeforeSignup } from "@src/auth/hooks.js",
+    onAfterSignup: import { onAfterSignup } from "@src/auth/hooks.js",
+    onBeforeOAuthRedirect: import { onBeforeOAuthRedirect } from "@src/auth/hooks.js",
+  },
+}
+```
+
+Wasp _awaits_ the async hook functions in various stages of the auth process. The hooks are called with an `args` object that contains the necessary information for the hook to do its job. 
+
+The return value is ignored for all hooks except `onBeforeOAuthRedirect` which enables you to modify the OAuth redirect URL.
```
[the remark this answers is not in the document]
sodic: Seems you still have "the return values of the hooks", why not "the hooks' return values?" I believe it's a simpler sentence, sounds more natural, and carries the same information.

No need to reply to the comment. Feel free to resolve it yourself whether you accept the suggestion or not.

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
```diff
@@ -0,0 +1,79 @@
...
+   */
+  providerId: ProviderId
+  /**
+   * User object that was created during the signup process.
+   */
+  user: Awaited<ReturnType<typeof createUser>>
+  oauth?: {
+    /**
+     * Access token that was received during the OAuth flow.
+     */
```
sodic, line 46: My personal feeling is that most of these type field comments are redundant and just repeat the information visible from the code.

Examples include:
```
oauth.accessToken - access token received during the oauthFlow (redundant)
url - URL that the OAuth flow should redirect to (not redundant, but 
perhaps better to call the field `oauthRedirectUrl` and drop the comment)
hookName - Name of the hook that is being executed (redundant)
```
I'm getting Cupic JavaDoc vibes :smile:

waspc/data/Generator/templates/server/src/auth/hooks.ts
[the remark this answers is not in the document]
sodic: > Did I understand this comment correctly?

You sure did.

> Common input has hookName, prisma and req, we want to keep out the first two. If I used keyof to get the keys of the CommonInput, I'd still need to write down the literal req which would still be duplication.

To be honest, I didn't know how I'd do it. I just had an idea that it'd be good to do it and thought that you might come up with something :smile: 

But now that I think of it. I see two solutions:
 - Changing `CommonAuthHookParmas` to `InternalAuthHookParams` and repeating the request in all hook params types. Something being "Internal" is a more semantically meaningful reason for extracting a separate type than something being "Common" (which is mostly to get syntactic convenience).
 - If you really want to keep `Common`, you could do this:
   ```typescript
   type CommonInternalAuthHookParams = { ... }
   type CommonExtranalAuthHookParams = { ... }
   type CommonAuthHookParams = CommonInternalAuthHookParams | CommonExtranalAuthHookParams
   
   // Somewhere else
   Omit<P, keyof InternalAuthHookParams>
   ```
   
I personally prefer the first approach and would even recommend it regardless of whether we address the hardcoded field types (which isn't really a big deal)

waspc/data/Generator/templates/server/src/auth/providers/oauth/state.ts
```diff
@@ -1,70 +1,93 @@
...
+import { setOAuthCookieValue, getOAuthCookieValue } from './cookies.js';
 
-export type StateType = 'state' | 'codeVerifier';
+/**
+ * OAuth state shape depends on whether the provider uses PKCE.
+ * If a provider uses PKCE, the state will include a code verifier.
+ */
+export type OAuthState<UsesCodeVerifier extends boolean = false> = {
+  state: string;
+} & (UsesCodeVerifier extends true
```
sodic, line 17: Boolean type arguments are generally considered an antipattern (even more than boolean arguments in functions, who also get a bad rep). This is because they move the type system away from "declaring sets of values" and into "implementing decision algorithms":
 - `Foo<User>` - It's a `Foo` of Users (whatever `Foo` is).
 - `Foo<true>` - Hmm.
 
Of course, boolean type arguments are sometimes necessary, and we do use them throughout the codebase. But if we can help it, we should avoid them.

All in all, this is a pretty complex type, it includes: a type parameter (and a boolean one at that), a default value, a type constraint, an intersection type, and an inline conditional type.

I understand this code, but put yourself in someone else's shoes. Would you say the prevented duplication/extra keystrokes are worth all the additional complexity?

We do sometimes use all of these things in our generics, but we're

web/docs/auth/auth-hooks.md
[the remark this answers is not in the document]
infomiho: Gotcha 👍  I'll modify all the instances of using `of the` and be mindful of them in the future.

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
```diff
@@ -0,0 +1,79 @@
...
+  /**
+   * Provider ID object that contains the provider name and the provide user ID.
+   */
+  providerId: ProviderId
+} & CommonAuthHookParams
+
+type OnAfterSignupHookParams = {
+  /**
+   * Provider ID object that contains the provider name and the provide user ID.
+   */
```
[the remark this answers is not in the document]
infomiho: In VS Code it shows up for me like this:

Screenshot 2024-06-10 at 12 33 59

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
infomiho: These comments are for the end user like some sort of documentation they can see at a glance: 
Screenshot 2024-06-10 at 12 33 59

From a point of view of a developer developing Wasp (us) they are too much, but from the point of view of a Wasp user they seem okay to me. Giving a bit of extra "context". `Access token that was received during the OAuth flow.` mentions where the access token came from and might help them with understanding the flow better.

With that in mind, of course we can improve the comments to be less redundant. But do you agree with my intention here?

waspc/data/Generator/templates/server/src/auth/providers/oauth/state.ts
```diff
@@ -1,70 +1,93 @@
...
-): { [name in ST]: string } {
-  const result = {} as { [name in StateType]: string }
+  res: ExpressResponse
+}): OAuthState<IsCodeVerifierUsed> {
+  const state = generateState();
+  setOAuthCookieValue(provider, res, 'state', state);
+
+  return {
+    state,
+    ...(isCodeVerifierUsed && generateAndStoreCodeVerifier(provider, res)),
```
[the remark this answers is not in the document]
infomiho: Sure, let's make it uniform 👍

waspc/data/Generator/templates/server/src/auth/providers/oauth/state.ts
infomiho: Ok, I've decided to be explicit about the OAuth flow type we are using i.e. with or without PKCE and this is what will result in the state object type.

waspc/data/Generator/templates/server/src/auth/hooks.ts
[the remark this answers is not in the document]
infomiho: Makes sense 👍 I've updated the type to be `Internal*` and they used the `keyof` operator to get the list of fields that are omitted. I've also dropped the redundant `hookName` field as well.

waspc/data/Generator/templates/server/src/auth/providers/oauth/state.ts
infomiho: Something like
```ts
export type OAuthStateFor<
  OT extends OAuthType
> = OAuthStateForOAuthType[OT];

export type OAuthStateWithCodeFor<OT extends OAuthType> = OAuthStateFor<OT> & {
  code: string;
};

export type OAuthType = keyof OAuthStateForOAuthType;

export type OAuthStateFieldName = keyof OAuthState | keyof OAuthStateWithPKCE;

type OAuthStateForOAuthType = {
  OAuth2: OAuthState,
  OAuth2WithPKCE: OAuthStateWithPKCE,
};

type OAuthState = {
  state: string;
};

type OAuthStateWithPKCE = {
  state: string;
  codeVerifier: string;
};
```

waspc/data/Generator/templates/server/src/auth/providers/oauth/handler.ts
```diff
@@ -1,108 +1,143 @@
...
-  stateTypes: ST[],
-  userSignupFields: UserSignupFields | undefined,
+  optionalStateTypes: OST[]
+  userSignupFields: UserSignupFields | undefined
   /*
     The function that returns the URL to redirect the user to the
     provider's login page.
   */
-  getAuthorizationUrl: Parameters<typeof createOAuthLoginHandler<ST>>[2],
+  getAuthorizationUrl: Parameters<typeof createOAuthLoginHandler<OST>>[2]
```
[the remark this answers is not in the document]
infomiho: This is another comment that is outdated by Github's standards but I still think we should either address is properly or resolve it if no longer relevant. 

The current implementation goes like this:
1. We need a router for a certain OAuth provider
2. We use the `createOAuthProviderRouter` helper which accepts the `oAuthType` param which can be `OAuth2` or `OAuth2WithPKCE`
3. The `oAuthType` dictates which kind of OAuth state is generated and validated in the OAuth flow
4. To define the "callback" functions like `getAuthorizationUrl` and `getProviderTokens` we use the type of `oAuthType` which need different OAuth state based on the OAuth flow type (with or without PKCE)

waspc/data/Generator/templates/server/src/auth/providers/oauth/handler.ts
```diff
@@ -1,108 +1,143 @@
...
-        );
+        const redirectUri = await finishOAuthFlowAndGetRedirectUri({
+          provider,
+          providerProfile,
+          providerUserId,
+          userSignupFields,
+          req,
+          accessToken,
+          oAuthState,
+        })
```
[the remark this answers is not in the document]
infomiho: I've created an issue to handle this outside of this PR: https://github.com/wasp-lang/wasp/issues/2092

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
sodic: Ohh, so they do show up. It never occurred to me to check the individual properties, I was only hovering on the value.

Yes, I agree wholeheartedly.

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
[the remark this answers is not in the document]
sodic: My bad, was only hovering over the object, not the fields. This is great.

web/docs/auth/auth-hooks.md
sodic, line 1: I've read the docs and they read nicely.

There are still many passive sentences in there that could be improved, but that's not important enough to hold the PR.

sodic (approved):
Everything looks great, approved (finally, sorry for the delay :sweat_smile:)

I left some comments though, so check them out.

waspc/data/Generator/templates/server/src/auth/providers/oauth/state.ts
```diff
@@ -1,70 +1,121 @@
...
+  provider: ProviderConfig,
+  req: ExpressRequest
+): { codeVerifier: string } {
+  const codeVerifier = getOAuthCookieValue(
+    provider,
+    req,
+    'codeVerifier'
+  );
+  if (!codeVerifier) {
+    throw new Error('Missing code verifier');
```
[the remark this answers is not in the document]
infomiho: It looks like this:

| Screenshot 2024-06-20 at 13 18 13 |
| --- |
| On the client |

| Screenshot 2024-06-20 at 13 18 17 |
| --- |
| In the terminal |

waspc/data/Generator/templates/server/src/auth/providers/oauth/state.ts
```diff
@@ -1,70 +1,121 @@
...
-    setOAuthCookieValue(provider, res, 'state', state);
-    result.state = state;
-  }
+  res: ExpressResponse
+}): OAuthStateFor<OT> {
+  return {
+    ...generateAndStoreState(provider, res),
+    ...(oAuthType === 'OAuth2WithPKCE' && generateAndStoreCodeVerifier(provider, res)),
+  };
+}
```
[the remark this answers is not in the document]
infomiho: Makes sense 👍 

I'll split up the code to work in "stages". Thanks for the suggestion!

waspc/data/Generator/templates/server/src/auth/providers/oauth/state.ts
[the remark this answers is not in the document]
sodic: Perfect!

web/docs/auth/auth-hooks.md
sodic: No no, nothing serious. We can merge this as far as I'm concerned.

waspc/data/Generator/templates/server/src/auth/providers/oauth/state.ts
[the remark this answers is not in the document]
infomiho: I've now written the methods as:
```ts
export function generateAndStoreOAuthState(...) {
  const state: OAuthStateFor<OT> = {
    ...generateState(),
    ...(oAuthType === 'OAuth2WithPKCE' && generateCodeVerifier()),
  };

  storeOAuthStateInCookies(provider, res, state);

  return state;
}

export function validateAndGetOAuthState(...) {
  const state: OAuthStateWithCodeFor<OT> = {
    ...getCode(req),
    ...getState(req),
    ...(oAuthType === 'OAuth2WithPKCE' && getCodeVerifier(provider, req)),
  };

  validateOAuthState(provider, req, state);

  return state;
}
```


Doing the `for (key in result)` bit gives me these kinds of errors:
```
Argument of type 'string' is not assignable to parameter of type 'OAuthStateFieldName'.ts(2345)
```
and
```
Argument of type 'OAuthStateFor<OT>[keyof OAuthStateFor<OT>]' is not assignable to parameter of type 'string'.
  Type 'OAuthStateFor<OT>[string] | OAuthStateFor<OT>[number] | OAuthStateFor<OT>[symbol]' is not assignable to type 'string'.
    Type 'OAuthStateFor<OT>[string]' is not assignable to type 'string'.
```
so I went with a more basic approach of doing hardcoded calls.

waspc/data/Generator/templates/server/src/auth/providers/oauth/state.ts
```diff
@@ -64,54 +68,64 @@ export function validateAndGetOAuthState<OT extends OAuthType>({
...
-  const state = generateState();
-  setOAuthCookieValue(provider, res, 'state', state);
+  validateOAuthState(provider, req, state);
 
-  return { state };
+  return state;
 }
 
-function generateAndStoreCodeVerifier(
+function storeOAuthStateInCookies<OT extends OAuthType>(
```
sodic, line 82: I would leave the implementation details out of the function's name -> `storeOAuthState`.

Makes it easier to change storage mechanisms too, the beauty of encapsulation :)

waspc/data/Generator/templates/server/src/auth/providers/oauth/state.ts
```diff
@@ -64,54 +68,64 @@ export function validateAndGetOAuthState<OT extends OAuthType>({
...
-): { codeVerifier: string } {
-  const codeVerifier = generateCodeVerifier();
-  setOAuthCookieValue(provider, res, 'codeVerifier', codeVerifier);
-
-  return { codeVerifier };
+  res: ExpressResponse,
+  state: OAuthStateFor<OT>
+): void {
+  setOAuthCookieValue(provider, res, 'state', state.state);
+  if (isOAuthStateWithPKCE(state)) {
```
sodic, line 88: Why is this typeguard necessary?

We'll have to add a new "save" statement whenever we add a field, and the type system does not enforce it (i.e., it's possible to forget).

Why not simply loop over the state object and store everything that's inside? That way, if we add a new field, it will work automatically.

infomiho: Because state is a union of `OAuthState` and `OAuthStateWithPKCE`. 

As I mentioned here: https://github.com/wasp-lang/wasp/pull/1993#discussion_r1647687897

Iterating over the values requires from me to assert the types 😢 I can do that ofc

waspc/data/Generator/templates/server/src/auth/providers/oauth/state.ts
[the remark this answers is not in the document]
infomiho: We found the reason for these type errors is using generics e.g. `OAuthStateFor<OT>` vs. just using all the available values with `OAuthStateFor<OAuthType>`. Works now 👍

PR #1995, opened 2024-04-25
Serialize AuthUser properly

infomiho:
We were serialising the `AuthUser` object via JSON but since the `user` object included methods, those couldn't be sent over the wire. We used the same type on the server and the client (incorrectly) which made it seem like the methods could be used. 

Now, the user is split into `AuthUserData` which contains plain data and the superset `AuthUser` which contains additional methods.


waspc/data/Generator/templates/sdk/wasp/server/utils.ts
```diff
@@ -2,12 +2,12 @@
...
 
 {=# isAuthEnabled =}
-import { type AuthUser } from 'wasp/auth'
+import { type AuthUserData } from 'wasp/auth'
 {=/ isAuthEnabled =}
 
 type RequestWithExtraFields = Request & {
   {=# isAuthEnabled =}
-  user?: AuthUser;
+  user: AuthUserData | null;
```
infomiho, line 10: This is not really correct. The `user` can be `undefined` if there is no auth token present!

waspc/data/Generator/templates/server/src/middleware/operations.ts
```diff
@@ -4,13 +4,14 @@ import {
...
 import { handleRejection } from 'wasp/server/utils'
+import { enrichAuthUser } from 'wasp/auth/user'
 
 export function createOperation (handlerFn) {
     return handleRejection(async (req, res) => {
         const args = (req.body && superjsonDeserialize(req.body)) || {}
         const context = {
             {=# isAuthEnabled =}
-            user: req.user
+            user: req.user ? enrichAuthUser(req.user) : null,
```
infomiho, line 14: This piece of code also change the type from `undefined` to `null` when the user is not authenticated.

waspc/data/Generator/templates/sdk/wasp/server/utils.ts
infomiho: I've adjusted the code to now always be `null` for non-authenticated user.

waspc/data/Generator/templates/sdk/wasp/auth/session.ts
```diff
@@ -19,7 +19,7 @@ export async function createSession(authId: string): Promise<Session> {
 
 // PRIVATE API
 export async function getSessionAndUserFromBearerToken(req: ExpressRequest): Promise<{
-  user: AuthUser | null,
+  user: AuthUserData | null,
   session: Session | null,
 }> {
```
sodic, line 24: I know it's not related to this PR, but I was just wondering, why isn't the return type here:
```typescript
Promise<{ user: AuthUser, session: Session } | null>
```
This type would be more correct, as `{ user: AuthUser | null, session: Session | null }` implies you could have several states that in reality aren't possible (e.g., ` { user: <some user> , session: null }`).

Is it because of how the frontend code usually uses this response?

waspc/data/Generator/templates/sdk/wasp/auth/useAuth.ts
```diff
@@ -21,7 +22,8 @@ function createUserGetter() {
     try {
       const response = await api.get(getMeRoute.path)
   
-      return superjsonDeserialize(response.data)
+      const userData = superjsonDeserialize<AuthUserData | null>(response.data)
+      return userData ? enrichAuthUser(userData) : null;
```
sodic, line 26: Hm, I feel like there's a better name for this than `enrich`, especially considering this function maps one data type to another (i.e., it doesn't only enrich a value through side effects, it performs a transformation).

Even if there isn't a standard name for this operation (other than `serialize` which would obviously be confusing :)), I'd still prefer something simpler like `makeAuthUser` or `getAuthUser` of type `(data: AuthUserData) => AuthUser`.

Perhaps even better, an `AuthUser` could be a class that takes `AuthUserData` through its constructor. This way, we'd have `new AuthUser(userData)`.

waspc/data/Generator/templates/sdk/wasp/auth/user.ts
```diff
@@ -26,11 +30,22 @@ export function getFirstProviderUserId(user?: UserEntityWithAuth): string | null
   return user.auth.identities[0].providerUserId ?? null;
 }
 
+// PRIVATE API
+export function enrichAuthUser(data: AuthUserData): AuthUser {
+  return {
+    ...data,
+    getFirstProviderUserId: () => {
```
sodic, line 37: Again, not related to this PR, but I forgot...

How comes this is a method instead of a property: `user.firstProviderUserId`?

Especially considering this is something we can precompute for each because it never changes... Or does it?

waspc/data/Generator/templates/sdk/wasp/package.json
```diff
@@ -22,6 +22,8 @@
       {=!  Used by our code, uncodumented (but accessible) for users. =}
       "./auth/session": "./dist/auth/session.js",
       {=!  Used by our code, uncodumented (but accessible) for users. =}
+      "./auth/user": "./dist/auth/user.js",
+      {=!  Used by our code, uncodumented (but accessible) for users. =}
```
sodic, line 26: Don't forget to add to the table of visible API (I know it's a pain :grimacing:).

Also, please specify which part of our code needs it (i.e., framework or sdk). I know most exports don't specify this, but doing it for new stuff can't hurt (unless it's wrong, of course, but it won't be :smile:).

waspc/data/Generator/templates/sdk/wasp/server/auth/user.ts
```diff
@@ -61,7 +64,7 @@ export type AuthEntityWithIdentities = {= authEntityName =} & {
 }
 
 // PRIVATE API
-export function createAuthUser(user: UserEntityWithAuth): AuthUser {
+export function createAuthUser(user: UserEntityWithAuth): AuthUserData {
```
sodic, line 67: We should change the function's name to match the type it returns.

waspc/data/Generator/templates/server/src/webSocket/initialization.ts
```diff
@@ -43,7 +44,10 @@ async function addUserToSocketDataIfAuthenticated(socket: Socket, next: (err?: E
   if (sessionId) {
     try {
       const { user } =  await getSessionAndUserFromSessionId(sessionId)
-      socket.data = { ...socket.data, user }
+      socket.data = {
+        ...socket.data,
+        user: user ? enrichAuthUser(user) : null
```
sodic, line 49: This null thing repeats a couple of times. What do you think about extracting it into a function?

sodic (commented):
I have a couple of easy suggestions, no big issues.

Nice work!

waspc/data/Generator/templates/sdk/wasp/auth/session.ts
infomiho: I kinda went with Lucia's types: https://github.com/lucia-auth/lucia/blob/d285a08dc2eda428236a1b267b19c01886e786ad/packages/lucia/src/core.ts#L122

But ... I can adjust them to be more sane 👍

waspc/data/Generator/templates/sdk/wasp/auth/useAuth.ts
infomiho: Enrich felt okay since it turned one _plain thing_ into this new _cool thing_ 😄 

The suggestion to use a class sounds good to me, I'll try doing that and see where it leads me.

waspc/data/Generator/templates/sdk/wasp/auth/user.ts
infomiho: It never changes, but it felt like more logic heavy piece on the `user` so I wanted to use a method instead of a getter 🤷

waspc/data/Generator/templates/sdk/wasp/package.json
infomiho: Sure thing, I updated the Sheet a bit, but not all the way 👍  

I'll update the comments as well.

waspc/data/Generator/templates/server/src/webSocket/initialization.ts
infomiho: Hm, I wonder what to call it. `enrichAuthUserIfExists` or something like that?

waspc/data/Generator/templates/sdk/wasp/auth/useAuth.ts
infomiho: Using a class seemed complex to maintain e.g. all the props in the object would need to be maintained as class properties. I went with a simpler name i.e. `makeAuthUser` :)

waspc/data/Generator/templates/sdk/wasp/package.json
infomiho: Updated the comments to say private API exports are used in SDK and server

waspc/data/Generator/templates/sdk/wasp/auth/user.ts
sodic: Got it, but what about this?
> Especially considering this is something we can precompute for each because it never changes... Or does it?

It's logic heavy because we compute it every time. But if it never changes, why not compute it once (during the construction of AuthUser), and then just use a property?

I'm not sure I prefer one solution over the other, just putting it out there.

infomiho: It feels weird to me to call a property `firstProviderUserId` since it _feels_ dynamic to me. I could precompute it and store it some closure variable. But still, it's super cheap to compute it so I'd keep it simple for now.

waspc/data/Generator/templates/sdk/wasp/server/auth/user.ts
```diff
@@ -9,23 +9,27 @@ import {
...
+export type AuthUser = AuthUserData & {
+  getFirstProviderUserId: () => string | null,
+}
+
+// PRIVATE API
 /**
  * Ideally, we'd do something like this:
  * ```
- * export type AuthUser = ReturnType<typeof createAuthUser>
+ * export type AuthUserData = ReturnType<typeof createAuthUser>
```
sodic, line 23: Minor, but update the function name too.

waspc/data/Generator/templates/server/src/webSocket/initialization.ts
```diff
@@ -42,8 +43,12 @@ async function addUserToSocketDataIfAuthenticated(socket: Socket, next: (err?: E
   const sessionId = socket.handshake.auth.sessionId
   if (sessionId) {
     try {
-      const { user } =  await getSessionAndUserFromSessionId(sessionId)
-      socket.data = { ...socket.data, user }
+      const sessionAndUser =  await getSessionAndUserFromSessionId(sessionId)
+      const user = sessionAndUser ? makeAuthUserIfPossible(sessionAndUser.user) : null
```
sodic, line 47: Can you do `makeAuthUserIfPossible(sessionAndUser?.user)` here?

I forgot how that questionmark thing works.

sodic (approved):

waspc/data/Generator/templates/server/src/webSocket/initialization.ts
infomiho: Let me check it out and I'll update it if it makes sense 👍

infomiho: Given that the [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) short-circuits to `undefined` we should either need to update the `makeAuthUserIfPossible` definition to accept `undefined` or write it like this:

```ts
makeAuthUserIfPossible(sessionAndUser?.user) ?? null
```

We agreed offline that the ternary operator is more readable and we'll keep the existing logic.

PR #2012, opened 2024-05-06
Adds `"module"` to `tsconfig.json`

### LINKED ISSUE #1989: Current template `tsconfig.json` breaks `import.meta.env` in the client app
Users will get an error in their IDE when trying to use env variables defined in the `.env.client`. The way they use those variables is by writing `import.meta.env.REACT_APP_SOMETHING`. With our current `tsconfig.json` set up, this will be underlined as an error.

This was reported by Vinny here: https://ptb.discord.com/channels/686873244791210014/1229414936967381062

### Reproduction
Creating a new project:
```
.
├── main.wasp
├── package.json
├── public
├── src
│   ├── Main.css
│   ├── MainPage.jsx
│   ├── vite-env.d.ts
│   └── waspLogo.png
├── tsconfig.json
└── vite.config.ts
```
which has then has the following `tsconfig.json`:
```json
{
  "compilerOptions": {
    "target": "esnext",
    "moduleResolution": "bundler",
    "jsx": "preserve",
    "strict": true,
    "esModuleInterop": true,
    "lib": [
      "dom",
      "dom.iterable",
      "esnext"
    ],
    "allowJs": true,
    "typeRoots": [
      "node_modules/@testing-library",
      "node_modules/@types"
    ],
    "outDir": ".wasp/phantom"
  }
}
```
results in the following error when using `import.meta.env.*`:
image

### Possible fix

Vinny reported that adding this to the project root `tsconfig.json` helped:
```diff
{
  "compilerOptions": {
+  "module": "ES2022"
  }
}
```

infomiho:
Closes #1989 

**Investigation:** Default `tsconfig.json` that Vite uses includes a `"module"` option, while our setup _does not_. This has been the case since the Wasp version 0.9.0 (when we introduced Vite and `import.meta.env` replaced `process.env`)! 

If we don't put a `module` value, Typescript uses an implicit default:
```
Default:
CommonJS if "target" is ES3 or ES5; 
ES6/ES2015 otherwise.
```
Since our target is `esnext`, the implicit default is `ES6`. 
We can't use use `import.meta` with `module` set to ES6, since it's too low and Typescript complains. We need  to put some value explicitly which is _at least_ ES2020, so we put `esnext` to match our target.

### Left to do
- [x] Update Changelog
- [ ] Update all other starters
  - [x] OpenSaaS https://github.com/wasp-lang/open-saas/commit/8a4eb95ebeb58c8e7fac458ddf79a1da6336f801
  - [x] todo-ts https://github.com/wasp-lang/starters/commit/21eb0d0bfade8c6f72d5b410dc4060d1a96e493e
  - [x] embeddings https://github.com/wasp-lang/starters/commit/21eb0d0bfade8c6f72d5b410dc4060d1a96e493e
  - [ ] Mage (we'd need to update Mage to use latest Wasp compiler version anyways, so I won't do it now)

### DIFF
--- a/waspc/data/Cli/templates/skeleton/tsconfig.json
+++ b/waspc/data/Cli/templates/skeleton/tsconfig.json
@@ -5,6 +5,7 @@
 // compiler. Proper TS compiler configuration in Wasp is coming soon :)
 {
   "compilerOptions": {
+    "module": "esnext",
     "target": "esnext",
     // We're bundling all code in the end so this is the most appropriate option,
     // it's also important for autocomplete to work properly.
--- a/waspc/ChangeLog.md
+++ b/waspc/ChangeLog.md
@@ -42,6 +42,9 @@ These changes improve code readability and lower the complexity of accessing use
 
 These changes only apply to getting auth fields from the `user` object you receive from Wasp, for example in the `authRequired` enabled pages or `context.user` on the server. If you are fetching the user and auth fields with your own queries, you _can_ keep using most of the helpers. Read more [about using the auth helpers](https://wasp-lang.dev/docs/auth/entities#including-the-user-with-other-entities).
 
+### 🐞 Bug fixes
+- Update the `tsconfig.json` to make sure IDEs don't underline `import.meta.env` when users use client env vars.
+
 ### 🔧 Small improvements
 
 - Improved the default loading spinner while waiting for the user to be fetched.
--- a/waspc/e2e-test/test-outputs/waspBuild-golden/waspBuild/tsconfig.json
+++ b/waspc/e2e-test/test-outputs/waspBuild-golden/waspBuild/tsconfig.json
@@ -5,6 +5,7 @@
 // compiler. Proper TS compiler configuration in Wasp is coming soon :)
 {
   "compilerOptions": {
+    "module": "esnext",
     "target": "esnext",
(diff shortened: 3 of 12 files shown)


waspc/data/Cli/templates/skeleton/tsconfig.json
```diff
@@ -5,6 +5,7 @@
 // compiler. Proper TS compiler configuration in Wasp is coming soon :)
 {
   "compilerOptions": {
+    "module": "esnext",
```
sodic, line 8: I believe the latest version of Vite uses `ES2020`. If we're using Vite, should we use that one too?

The same question applies for the `target` field. I set that one to `esnext`, my bad.

Btw, nice work remembering to update OpenSaas and starters. Just link those commits/PRs in the description please :)

infomiho: Our web app uses `esnext` and IMHO it makes sense to keep them in sync: https://github.com/wasp-lang/wasp/blob/main/waspc/data/Generator/templates/react-app/tsconfig.node.json#L5

Vite uses `esnext` as well 🤷 

It makes sense TBH since users won't get IDE errors for new things and we can update our runtime over time without asking them to update their `tsconfig.json`.

I'll link to commits 👍

sodic (approved):

PR #1992, opened 2024-04-24
Fix directly calling operations on the frontend 

sodic:
Fixes #1923.

I'll do all the testing once this gets approved.
I've already tested it a couple of times and it all works, but I'll hold of the final just-in-case tests until all changes are final.

Tasks:
- [x] Fix query API.
- [ ] Figure out what to do with the type `Query`, `Action`, `Operation`, `QueryFor`, `OperationFor`. Should we expose some of them and remove duplication?
- [x] Test queries with no payload.
- [x] Test queries with payload.
- [x] Test optimistic updates on queries with no payload.
- [x] Test optimistic updates on queries with payload.
- [x] Test caching on queries with payload.
- [x] Test caching on queries without payload.
- [x] Fix query api types.
- [x] Test actions.
- [x] Test todoApp build.
- [x] Update docs.
- [ ] Test CRUD

PR #2021, opened 2024-05-08
Small cleanup after #1992

sodic:
Cleanup after #1992


waspc/ChangeLog.md
sodic, line 1: Autoformatter did some of the work here.

PR #2022, opened 2024-05-08
Adds auth hook headless tests

infomiho:
I've tested what I can (since I can't test the OAuth flow):
- onBeforeSignup: reject some invalid email
- onAfterSignup: set property on the `user` in the DB

I've also extracted some repeated logic into a few helper functions, minor stuff.


sodic (approved):
Looks good to me! Nothing to add here.

PR #2025, opened 2024-05-09
Reduce sleep for testing optimistic update

sodic:
Forgot to address this in #2021 

5 seconds is long. 2 seconds gets the same point across and is more reasonable.


infomiho (approved):
Seems simple enough

PR #2039, opened 2024-05-17
Fixes route ordering to prevent Wasp routes being overriden

### LINKED ISSUE #2029: User defined routes can override `/oauth/callback` and break OAuth flow
Context: https://ptb.discord.com/channels/686873244791210014/1238173111363305574/1239325290824732673

If users define a wildcard route (to serve as a error page for missing routes) they will override the OAuth callback because user routes come before Wasp defined routes.

We should make sure Wasp defined routes are before in the `router.tsx`.

infomiho:
Closes #2029

- Moved Wasp routes to the top of the `router.tsx`
- Added a headless test to cover the scenario

### DIFF
--- a/waspc/ChangeLog.md
+++ b/waspc/ChangeLog.md
@@ -64,8 +64,8 @@ These changes only apply to getting auth fields from the `user` object you recei
 ### 🐞 Bug fixes
 
 - Update the `tsconfig.json` to make sure IDEs don't underline `import.meta.env` when users use client env vars.
-- Fixes the `netlify.toml` to include the correct build path for the client app.
-- Moves Wasp defined routes above the user defined routes in the client router to ensure that user defined routes don't override Wasp defined routes.
+- Fix the `netlify.toml` to include the correct build path for the client app.
+- Fix the client router to ensure that user defined routes don't override Wasp defined routes by moving the user defined routes to the end of the route list.
 
 ### 🔧 Small improvements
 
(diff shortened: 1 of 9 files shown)


waspc/ChangeLog.md
```diff
@@ -62,7 +62,9 @@ These changes improve code readability and lower the complexity of accessing use
 These changes only apply to getting auth fields from the `user` object you receive from Wasp, for example in the `authRequired` enabled pages or `context.user` on the server. If you are fetching the user and auth fields with your own queries, you _can_ keep using most of the helpers. Read more [about using the auth helpers](https://wasp-lang.dev/docs/auth/entities#including-the-user-with-other-entities).
 
 ### 🐞 Bug fixes
+
 - Update the `tsconfig.json` to make sure IDEs don't underline `import.meta.env` when users use client env vars.
+- Moves Wasp defined routes above the user defined routes in the client router to ensure that user defined routes don't override Wasp defined routes.
```
sodic, line 67: Nit: let's keep the verb form consistent (Update vs Moves). "Moves" is a little weird. Even if inconsistent, "move" and "moved" would both be find, but "moves" just feels unsettling to me :smile:

sodic (approved):
Nice work adding the test. Left one minor comment

waspc/ChangeLog.md
infomiho: I've update the line to read better 👍 Thanks!

PR #2044, opened 2024-05-20
Improve server-side operations API

### LINKED ISSUE #1909: Rethink server-side Operations API
### Update by @sodic

When attempting to fix this, I realized our server-side Operations API is virtually unusable:
- Server Operation types are completely unspecified.
- The API isn't properly documented.
- The API expects users to construct Wasp's internal types such as `AuthUser` (which also isn't public or documented). Even if it was documented, it still wouldn't be a great API :)

This happened because we created this API as an afterthought and plugged users into an arbitrary point of our internal RPC architecture. Therefore, the API has never really been user friendly. To make things worse, since it's basically a hook into our internal implementation, it can easily change without us noticing (e.g., it happened with `AuthUser`).

This problem slipped under the radar for so long because calling operations from the server isn't that crucial of a feature, but we should still fix it.

Idea: create a separate wrapper for server side queries that's as clean as possible and does most of the ugly work under the hood.

Fixing this will also fix part of #1958.

### Original issue
Currently we say

> To use a Query, you can import it from `wasp/client/operations` and call it directly. As mentioned, the usage doesn't change depending on whether you're on the server or the client:

and 

> To use an Action, you can import it from `wasp/client/operations` and call it directly. As mentioned, the usage doesn't change depending on whether you're on the server or the client:

which is incorrect. The users shouldn't import the actions and queries on the server from `wasp/client/operation` but rather from `wasp/server/operations`.

Also, we need to explain that they need to provide the `context` as the first argument and then the arguments as the second argument.

Discord thread about this: https://ptb.discord.com/channels/686873244791210014/1218567315185209464/1219591538942804019

### LINKED ISSUE #2004: Improve type signature for `void` RPCs
For any `void` backend Operation, such as:
```typescript
export const getAnything: GetAnything = async () => {
  return 'anything'
}
```
The inferred payload type is `args: void`. Calling `Parameters<typeof getAnything>` returns `[args?: void | undefined]`:
```typescript
// What we have
type Params = Parameters<typeof getAnything> // Params = [args?: void | undefined]
getAnything()          // works
getAnything(undefined) // works

// What we want
type Params = Parameters<typeof getAnything> // Params = []
getAnything()          // works
getAnything(undefined) // doesn't work
```
It works well enough for most use cases, but is not 100% correct (i.e., it will allow a call without arguments, but will also allow a call with `undefined`).

sodic:
Closes #1909 using the approach described in the [last comment on the issue](https://github.com/wasp-lang/wasp/issues/1909#issuecomment-2104476140).
Closes #2004.

> [!NOTE]
> I've only implemented the API for `AuthUser` for now. I think I'll make another issue that deals with adding `User` to the mix because it's probably not important enough to justify the time it would take.

> [!TIP]
> I recommend generating the files by running `wasp start` and looking at the SDK. It's much easier to understand what's going on this way. Choose an app with many operations (e.g., `waspc/todoApp`)

> [!WARNING]
> I love these admonitions and have started overusing them.

Todos:
 - [x] Make [an issue](https://github.com/wasp-lang/wasp/issues/2066) for supporting `User`.
 - [x] Test thoroughly one more time.
 - [x] Update public API table.
 - [x] Update docs.
 - [x] Update changelog.


waspc/data/Generator/templates/sdk/wasp/package.json
```diff
@@ -105,6 +105,7 @@
       "./server/email": "./dist/server/email/index.js",
       "./server/jobs": "./dist/server/jobs/index.js",
       "./server/operations": "./dist/server/operations/index.js",
+      "./server/operations/actions/wrappers": "./dist/server/operations/actions/wrappers.js",
```
sodic, line 108: Todo: relict of testing remove this.

waspc/data/Generator/templates/sdk/wasp/server/_types/index.ts
```diff
@@ -1,41 +1,52 @@
...
-import { type Payload } from "./serialization";
+import { type _Entity } from './taggedEntities'
+import { type Payload } from './serialization'
 
 export * from "./taggedEntities"
 export * from "./serialization"
 
-export type Query<Entities extends _Entity[], Input extends Payload, Output extends Payload> = 
-  Operation<Entities, Input, Output>
+export type UnauthenticatedQueryDefinition<
```
sodic, line 18: The `Definition` suffix is appropriate here, as we're using these generics to construct the types users assign to their operation definitions.

We want to use `Query`, `Action`, etc. for typing the user-facing API for calling the operations (see `wrappers.ts`).

waspc/data/Generator/templates/sdk/wasp/server/operations/wrappers.ts
sodic, line 1: This is the machinery. I've added comments to help understanding.

Please comment if something is still unclear.

waspc/data/Generator/templates/sdk/wasp/server/operations/wrappers.ts
```diff
@@ -0,0 +1,223 @@
...
+/**
+ * Constructs the unauthenticated operation's server-side API type from its 
+ * definition.
+ *
+ * @template OperationDefinition The type of the unauthenticated operation's
+ * definition.
+ */
+export type UnauthenticatedOperationFor<
+  OperationDefinition extends GenericOperationDefinition
+> = Parameters<OperationDefinition> extends []
```
sodic, line 23: [TS Playground](https://www.typescriptlang.org/play/?noImplicitAny=false&strictNullChecks=false&strictFunctionTypes=false&strictPropertyInitialization=false&strictBindCallApply=false&noImplicitThis=false&noImplicitReturns=false&alwaysStrict=false&ts=4.9.5#code/PTAEBUAsEsGdTqAhqW0C2AHANtAZtAKYAmoAboQE5oD2AdqDXqAO6RIAurhoxNCDGgFdKoAMY1ihADSgARkI4AoEKkjDspQgA8cSaAw6QelQklj15hGHVJGeAAwAKSSknSEOVWAB4AQkhiANaEtgDymFSc0PQAfKA6XrbwANoAug4qYBK20BwxdAB0SlmgAILkNNCkEVH5llIEdCSseZCgDrDRsASEsA4JyaBCmMjwRogcAJ6RpQAUmJQ0kaJkVaR4QnRi9XQAlErTkaAAausAYltioAC8oHN7t-F0QuhyVCWqUIiTxqCLNDE2H0HlIsBGVCOPCYyGw2FAtTcu1gxR0mBolC4UNAAHFQlRoGIAsFQsREdFLHc5q4AOawABcoGaFEoshyXm0HEZSDoU0eN3iWyCdBoLDonzAAFEyEhsEJOH1QBx+AByDiUISEFVWMRIISwHjgADKStMnHgDjO1Uu2wG5lQQjk2KYpQceOalEJxJC4RWFLoAwAtKh+HhXCHQDSaJJitircQbWJJZzSbBvaTybtbkpQKcLlcEimhu6CUTAj6yX6swB+U2a0CMsPYA0S0AAdXoatYGKCOr1Brz1oL7HgIv+rncnm8sntaMIOxIsZmPBcbnQ8Duq8nXmoPihMPjidiKQADGlW35FAhmCweJgr-Zx2up6JEm4dgUBMrkJH8Z7rlCpRGJwSpICE8AoNitKvKEXA5LA6r6HQXh2Pw7zIKU4JOsujDMG6f5euWGZVgUDhzBM8DsJgkR0PAbChNwKqmPIQjQJoBg0owIilAASk4ADCSo4Y0Bh5AUsB7IUUmHDheIcOc0DUBwW7oD45yFkkxDwCW-7pr6dQFPENw5qAKkvr45zHmerZSYU7Y8IEYjVLBsrYFMpqEr2xpfvwEx0EEHEdIeVydE+24fKouAIQgsB0F2zRiH0XSem5CRYNMxQgKUCl0K5UwzgwhDpe5rhuO5UxEJo8BbCJzTELIbCEu0BgSJQphAtMoDAu8zZKn8mBIFM2A0EgdjLqUB75tsYwdDVhBNCQDhLscLhDUgkhhHgwXTXcckKUpKk+NtYixK2UBUPNGIyNwTKEC0KA0vKlB2OwcHGMEdHGPYoiPgNz47l1cBcDCKCbNsuylIgRWYBlMnHHtikISpABMakaamuIEWWJL6UihnZrmZk7hZ8SJBj6SgLWazVA2pkTh4xNqVZ56lGUth9T88DU-VrAwGIzWfSBt6sDyKHLSug3AhtW1TWIyO3Jj8mI8p9Oo8dp2szSSENTwfAhh4GluKgHBCHgzDvGgUjwG+gT5HQnGPgQSlhQzVDSKUl5YjA8C-DwrU8M5KDvDYL2OET3j+ERuP+qTRZaaA6SZKoOTEGJ9DFEoQA) explaining why this check is necessary.

waspc/src/Wasp/Generator/SdkGenerator/Server/OperationsGenerator.hs
```diff
@@ -129,14 +141,15 @@ serverOperationsDirInSdkRootDir =
     (AS.Operation.QueryOp _ _) -> [reldir|queries|]
     (AS.Operation.ActionOp _ _) -> [reldir|actions|]
 
-getOperationTmplData :: AS.Operation.Operation -> Aeson.Value
-getOperationTmplData operation =
+getOperationTmplData :: Bool -> AS.Operation.Operation -> Aeson.Value
+getOperationTmplData isAuthEnabledGlobally operation  =
```
sodic, line 145: Dragging `isAuthEnabledGlobally` all the way down seems a little silly.

Querying the app spec for an Operation would ideally tell us whether it's authenticated or not, but that's not how the AppSpec works.

@Martinsos am I missing something?

waspc/examples/todoApp/src/testing.ts
sodic, line 1: I'll find a better way to test/specify this behavior before merging.

waspc/data/Generator/templates/sdk/wasp/server/operations/wrappers.ts
Martinsos, line 1: Lot of stuff in this file! But very cool!

Would it benefit from splitting further? Multiple files in subdirs or something? It is a bit hard to follow who is adding on top of whom here. But if that would just be too complex, I get it, no worries. This also goes for some other files in this PR, feels a bit like it could all be a bit better split, but I do understand if it doesn't feel worth it at the moment.

Oh ok, and the name of this file is a bit meh, couldn't we have a better name? What does `wrappers` stand for really? I look at that name, have no idea what that is.

waspc/examples/todoApp/main.wasp
Martinsos, line 1: Yeah just throw it all into todoApp, who cares

Martinsos (approved):
Nice stuff! This will be great now. You get my approval, but you do need to finish stuff from the TODO list + also add to ChangeLog.

waspc/data/Generator/templates/sdk/wasp/server/operations/wrappers.ts
sodic: > Would it benefit from splitting further? Multiple files in subdirs or something?

I don't think it would—there are only four exported symbols from the file, and everything else is used to build those four public symbols.

Also, most of the private symbols are only referenced once, and only one or two are referenced twice.

> So splitting this machinery into multiple files would IMO create unnecessary exports and confusion.

You're probably talking about `server/_types/index.ts`, as other files are pretty small. That one has more exports and could possibly benefit from being split, but since:
- It's not a clear benefit.
- The file is not a core part of the PR (I only did renamed and exported some of its symbols).
- Due to our SDK TS/package.json shenanigans, moving and changing ts files always risks  breaking something (this will be improved soon)

I decided against messing with that one.

> Oh ok, and the name of this file is a bit meh, couldn't we have a better name?

This brings us back to the naming problem I mentioned in the comment where we discussed "server-side API." On the client, this file's equivalent is called `rpc.ts`. I don't have an idea what to call it here.

I'll describe its function, and you can help me out:

> It's a file with all the utilities necessary for building an API Wasp users will use to call their operations from the server's runtime.

I thought of:
- `serverSideAPIHelpers` - Too verbose and has the word "helpers" in it which is arguably even worse :smile: 
- `serverSideOperationWrappers` - Also too verbose, carries the same information as `wrappers` (since it's in the folder `server/operations`)
- `operationWrappers` - Same deal as `serverSideOperationWrappers`

web/docs/data-model/operations/queries.md
```diff
@@ -188,7 +188,11 @@ For a detailed explanation of the Query definition API (i.e., arguments and retu
 
 ### Using Queries
 
-To call a Query in the client code, you can import it from `wasp/client/operations` and call it directly.
+#### Using Queries on the client
```
sodic, line 191: I didn't update the API reference because:
 - It already points to this example of usage, which I think is clear enough
 - It's something they define themselves, not a core part of Wasp's API.
 - Documenting the exact type signature of the public API in the API reference would require shuffling stuff around, which is more work.
 
Let me know if you think I should update it regardless.

web/docs/data-model/operations/queries.md
```diff
@@ -217,6 +221,47 @@ const doneTasks = await getFilteredTasks({ isDone: true })
...
 
+#### Using Queries on the server
+Calling a Query on the server is similar to calling it on the client.
+
+Here's what you have to do differently:
+ - Import Queries from `wasp/server/operations` instead of `wasp/client/operations`.
+ - Make sure you pass in a context object with the user to authenticated Queries.
+
+<Tabs groupId="js-ts">
+<TabItem value="js" label="JavaScript">
```
infomiho, line 232: These examples are "floating" without a function in which they are called. Is it an another action or query? Is it API? Also, how do I acquire an `AuthUser`?

Making the example more practical will be more ... practical for the users since they will want to do this in operations anyways and for that use case we have a good answer for them how to get the `AuthUser`.

waspc/data/Generator/templates/sdk/wasp/server/operations/wrappers.ts
```diff
@@ -0,0 +1,234 @@
...
+>(
+  userOperation: OperationDefinition,
+  entities: EntityMapFor<OperationDefinition>
+): AuthenticatedOperationFor<OperationDefinition> {
+  async function operation(...args: AuthenticatedOperationArgsFor<OperationDefinition>) {
+    // To understand this function and its limitations, read
+    // https://github.com/wasp-lang/wasp/issues/2050
+    if (args.length < 1) {
+      // No arguments sent -> no user and no payload specified -> there's no way this was called correctly.
+      throw new Error('Invalid number of arguments')
```
infomiho, line 110: I'd maybe expand on the error and help the users to move forward. "You called an operation without arguments. Since this operation requires auth..."

waspc/data/Generator/templates/sdk/wasp/server/operations/wrappers.ts
```diff
@@ -0,0 +1,233 @@
...
+
+/**
+ * Constructs the type of the arguments array for an authenticated operation
+ * based on the type of its definition.
+ */
+type AuthenticatedOperationArgsFor<Op extends GenericOperationDefinition> =
+  Parameters<AuthenticatedOperationFor<Op>>
+
+/**
+ * Constructs the type for an authenticated operation's server-side API.
```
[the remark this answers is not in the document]
infomiho: Server-side API feels okay to me since there are the `wasp/server/operations` and `wasp/client/operations` modules which we are trying to get the users to get used to :D

waspc/examples/todoApp/src/testing.ts
```diff
@@ -0,0 +1,84 @@
...
+  voidToNumberNoAuth as waspVoidToNumberNoAuth,
+  boolToNumberNoAuth as waspBoolToNumberNoAuth,
+  boolToNumberAuth as waspBoolToNumberAuth,
+} from 'wasp/server/operations'
+
+export const testingAction: TestingAction = async (args, context) => {
+  const result1 = await waspVoidToNumberNoAuth()
+  const result2 = await waspBoolToNumberNoAuth(true)
+
+  const user = context.user!
```
infomiho, line 19: If we are keeping these tests, I'd recommend using a conditional statement here so the action can work even without the user 🤷

infomiho:
Good job! 

It seems like the `AuthUser` is causing some trouble for us here 😢 they can simply coerce some object like `{ id: 5 } as AuthUser` as an extreme measure?

waspc/data/Generator/templates/sdk/wasp/server/operations/wrappers.ts
```diff
@@ -0,0 +1,233 @@
...
+        entities,
+      })
+    }
+  }
+
+  return operation as AuthenticatedOperationFor<OperationDefinition>
+}
+
+/**
+ * Returns a boolean carrying compiler type information about whether the
```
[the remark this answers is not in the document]
Martinsos: This isn't an implementation detail, this is behaviour. Not mentioning implementation details in the top header means you are not explaining HOW it does what it does. But stuff like complexity (time, memory), assumptions it relies on, those are all part of the WHAT, now HOW, therefore part of API of this function. So not a really specific case, but an assumption, and those should be documented so the caller of the function can know what they can expect.

waspc/data/Generator/templates/sdk/wasp/server/operations/wrappers.ts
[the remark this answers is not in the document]
Martinsos: Actually that makes sense, I am not sure why it wasn't clear to me at the moment of the reading. Might be that I got involved in some other context so that threw me off. Sounds good!

waspc/data/Generator/templates/sdk/wasp/server/operations/wrappers.ts
```diff
@@ -0,0 +1,233 @@
...
+/**
+ * Constructs the type for an unauthenticated operation's server-side API.
+ *
+ * @template Input The type of the payload the operation expects (must be `void` if the operation doesn't expect a payload).
+ * @template Output The type of the operation's return value.
+ */
+type UnauthenticatedOperation<Input, Output> = Operation<Input, Output, false>
+
+// todo(filip): Should i define this independently of AuthenticatedOperationDefinition?
+// Using AuthenticatedOperationDefinition because it's
```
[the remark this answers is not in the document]
Martinsos: Probably wasn't out of date while I was commenting on it, but glad to see it is fixed.

waspc/data/Generator/templates/sdk/wasp/server/operations/wrappers.ts
```diff
@@ -0,0 +1,233 @@
...
+{=/ isAuthEnabled =}
+/**
+ * Constructs the type for an unauthenticated operation's server-side API.
+ *
+ * @template Input The type of the payload the operation expects (must be `void` if the operation doesn't expect a payload).
+ * @template Output The type of the operation's return value.
+ */
+type UnauthenticatedOperation<Input, Output> = Operation<Input, Output, false>
+
+// todo(filip): Should i define this independently of AuthenticatedOperationDefinition?
```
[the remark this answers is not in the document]
Martinsos: Yup, I thought that defining a generic operation via more specific authenticated operation is weird. And I still do.

Defining public API type via internal stuff -> why not?

waspc/data/Generator/templates/sdk/wasp/server/operations/wrappers.ts
Martinsos: 4 exported public symbols, the rest used only once by these -> ok, that doesn't necessarily mean it should be split, on the opposite, it might mean you might want to split this into ~4 modules, so it becomes clear who is calling whom and it is easier to find your way around. But it really depends on dependencies here (pun intended), so if you think it makes no sense to split it, I trust so.

Naming issue -> hm. Good attempt with describing it, that usually gives a good material to shape into a name. But if that doesn't work, then one can question the code design as the cause of not being able to find a name.
Ok at least now I understand what this file does. And it is hard to find a better name. Because the "material" is quite involved. `userCodeWrappers` potentially. But not much better, as that user code part is somewhat implicit here I guess. I also don't see clear way now to easily improve code design here, nothing seems obviously off.

Ok, let's stay with `wrappers` if we don't have anything better, but could you add a top level comment in this file explaining what this file is about, same like you described in the message above?

waspc/data/Generator/templates/sdk/wasp/server/operations/wrappers.ts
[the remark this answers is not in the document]
sodic: So... is that a "yes" or a "no" to my suggestion? :smile:

waspc/data/Generator/templates/sdk/wasp/server/operations/wrappers.ts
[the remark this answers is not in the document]
sodic: > Defining public API type via internal stuff -> why not?

Because the API can change unexpectedly when we change an internal definition, and the public API should arguably be its own source of truth.

This was one of the core reasons decisions behind structural typing in TS and Go: _define the structure you expect instead of importing names_.

It all depends on where it makes sense to have the source of truth for something. In this case, I was on the edge, but if you don't mind it, I'm keeping it as is.

> Yup, I thought that defining a generic operation via more specific authenticated operation is weird. And I still do.

Yes, it's a little unintuitive, but correct in terms of subtyping and the type system.

There's a comment above it that explains what's going on. You possibly haven't seen it since you reviewed an old version of the code (that's what [this](https://github.com/wasp-lang/wasp/pull/2044#discussion_r1613744651) is about).

So, If you haven't seen it, please take a look at the comment.

If you still don't like it, I'll split the definition into `GenericUnauthenticatedOperation` and `GenericAuthenticatedOperation`.

waspc/data/Generator/templates/sdk/wasp/server/operations/wrappers.ts
sodic: Sorry, but I didn't get the pun :sweat_smile: 

But ok, staying with wrappers.

Btw, before calling it `rpc`, we called the client version of this code `core`, so... yeah, wrappers are actually an improvement :smile:

web/docs/data-model/operations/queries.md
sodic: You got me.

I left it out on purpose because I didn't want to admit the API is still crippled.

Anyway, I'll improve what the comment says and make sure we tackle https://github.com/wasp-lang/wasp/issues/2066 as soon as possible.

waspc/data/Generator/templates/sdk/wasp/server/operations/wrappers.ts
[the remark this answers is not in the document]
Martinsos: That means YES to the fact that you should document assumption in the header of the function and NO to the argument you made that this is some kind of special case, at least not the way you put it there.

I took a look, looks fine, although it is a bit out of context.

>  * To understand why "two arguments" means "includes payload", read
>  * https://github.com/wasp-lang/wasp/issues/2050
 
-> sounds like I already have context on what this "two arguments" is, but I am hearing about it for the first time. You are assuming I read the implementation I believe, but Imight have not and maybe I am just looking at the docs of this function, not how it is implemented.

So better would be something like this:

```
NOTE: This function assumes that payload is included if and only if there are 2 arguments. To understand why is this so, check ... .
```

Martinsos (approved):
I think one or two comments are ongoing, but they should be clear enough to finish from here. So approval from me! But @infomiho  will be able to cover the TS ground more thoroughly.

waspc/data/Generator/templates/sdk/wasp/server/operations/wrappers.ts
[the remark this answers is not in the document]
sodic: > You are assuming I read the implementation I believe

Not the implementation - the type signature, which is a part of the function's docs (check the screenshot above).

There's a "best practice" in typescript to not repeat type information in documentation (well, any static language, I guess :smile:).

In this case, the type signature says:
>  _This function takes a single argument. This argument is either an array with the payload and the context, or an array with just the context. It returns a type guard that says which one it is._ 

Once you know this, the implementation is straightforward and unambiguous (not that you need to know it). Therefore, this assumption
> NOTE: This function assumes that payload is included if and only if there are 2 arguments.

is made somewhere else (this function doesn't assume that at all, the caller does).

This function scope is simply, "If you give me something that's either A or B, I'll tell you which is it." The reasons why we're assuming A and B are the only options is within the caller's scope, and is documented there.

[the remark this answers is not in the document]
Martinsos: I didn't maybe pay enough deep attention to that sentence, but neither will typical reader.

That sentence says it will receive an array with payload and context, or an array with just the context, and it will correctly determine which one it is. That doesn't so obviously translte into "it if it array with any two elems in it it will be one, or it is one elem it will be the other". And I remember you were saying before that we can't be smart enough here, that we are only checking if it has 1 or 2 elements.

Maybe I got something wrong here, but maybe it also doesn't matter -> what happened is that you had me read this, and I was confused. You possibly explained it to me here, but ideally your code would do that for you. So can you improve this to make it not confuse the next person, even though it is clear to you?

waspc/examples/todoApp/src/testTypes/operations/client.ts
sodic, line 1: I've added type tests for operations on both the client and the server, and confirmed they fail the build on incorrect types:
image

waspc/data/Generator/templates/server/src/actions/_action.ts
```diff
@@ -16,5 +16,3 @@ export default async function (args, context) {
     },
   })
 }
-
-export type {= operationTypeName =} = typeof {= jsFn.importIdentifier =}
```
sodic, line 20: This is unused. We have the same thing in the SDK, and that's where it's used.

waspc/data/Generator/templates/sdk/wasp/client/operations/actions/core.ts
```diff
@@ -37,7 +37,10 @@ export function createAction<BackendAction extends GenericBackendOperation>(
   const action = (args) => internalAction(args, [])
   action.internal = internalAction
 
-  return action as ActionFor<BackendAction>
+  // todo(filip): why is unkonwn necessary here?
+  // Maybe something to do with
+  // https://www.typescriptlang.org/play/?#code/MYewdgzgLgBCBGArGBeGBvGBDAXDA5AGYgj4wC+AUAogHTyoHxYBO+llA9JzIQK5hgUAJbhsAG3EgA7hGxgYAUwBuIccuFgA5jCgBPAA6KY8PrBqKhMACYhFcsCCiVQkWPwVoAFAEpUAPgJiUkoPekZ8ZjYgA
```
infomiho, line 42: I'd slightly give more info in the comment (for your future self's sake). What is the error you are seeing? Maybe the playground can reproduce it more faithfully.

waspc/data/Generator/templates/sdk/wasp/server/operations/wrappers.ts
```diff
@@ -0,0 +1,254 @@
...
+ * Constructs the type of the server-side API for an operation that
+ * doesn't expect a payload.
+ *
+ * @template Output The type of the operation's return value.
+ * @template IsAuthenticated true if the operation is authenticated, false otherwise
+ */
+type OperationWithoutPayload<
+  Output,
+  IsAuthenticated extends boolean
+> = IsAuthenticated extends true
```
infomiho, line 252: I guess this one of the cases where the boolean type param is necessary 😄  https://github.com/wasp-lang/wasp/pull/1993#discussion_r1632964113

waspc/examples/todoApp/src/testTypes/operations/helpers.ts
```diff
@@ -0,0 +1,12 @@
+// All generics in this file are taken from
```
infomiho, line 1: `helpers.ts` -> `testHelpers.ts` or `assertions.ts`? 🤷

waspc/examples/todoApp/src/testTypes/operations/server.ts
```diff
@@ -0,0 +1,92 @@
...
+  taskToTaskUnspecified as taskToTaskUnspecifiedDefinition,
+  taskToTaskSatisfies as taskToTaskSatisfiesDefinition,
+} from './definitions'
+
+import { Equal, Expect } from './helpers'
+import { AuthUser } from 'wasp/auth'
+import { Task } from 'wasp/entities'
+
+export const testingAction: TestingAction = async (args, context) => {
+  // todo: test whether the outputs are correct
```
infomiho, line 25: What do we do with this TODO?

waspc/src/Wasp/Generator/SdkGenerator/Client/OperationsGenerator.hs
```diff
@@ -14,7 +14,7 @@ import qualified Wasp.AppSpec.Query as AS.Query
 import Wasp.Generator.Common (makeJsArrayFromHaskellList)
 import Wasp.Generator.FileDraft (FileDraft)
 import Wasp.Generator.Monad (Generator)
-import Wasp.Generator.SdkGenerator.Common (SdkTemplatesDir, clientTemplatesDirInSdkTemplatesDir, makeSdkImportPath, relDirToRelFileP)
+import Wasp.Generator.SdkGenerator.Common (SdkTemplatesDir, clientTemplatesDirInSdkTemplatesDir, getOperationTypeName, makeSdkImportPath, relDirToRelFileP)
```
infomiho, line 17: Nit: put imported symbols on newlines

infomiho (approved):
Left some minor comments, but looks good to me

waspc/examples/todoApp/src/testTypes/operations/helpers.ts
sodic: They're in the wrong folder too (should be in in `testTypes`). I'll correct it.

waspc/data/Generator/templates/sdk/wasp/server/operations/wrappers.ts
sodic: Haha, I knew you'll notice this!

Yeah, in this case since:
- It's very local (only one generic calls the boolean)
- The alternative option is writing a Cartesian product with 4 simple generics with duplicate definitions of very subtle behaviors (e.g., optionality of arguments) whose logic is arguably simpler to follow with the boolean.

I've decided to use the boolean.

I might be biased for working inside this file for too long so, if you think the boolean is confusing, I'll gladly remove it.

waspc/data/Generator/templates/sdk/wasp/client/operations/actions/core.ts
sodic: I'll add the info somewhere (either here or in the playground).

waspc/examples/todoApp/src/testTypes/operations/server.ts
sodic: We remove it :upside_down_face:

waspc/data/Generator/templates/sdk/wasp/server/operations/wrappers.ts
infomiho: Sounds good, thanks for the detailed write-up 👍

waspc/examples/todoApp/src/testTypes/operations/helpers.ts
sodic: I moved the file to `testTypes/helpers`, but decided against renaming it:
- `testTypes/testHelpers` is redundant
- `assertions` would not be correct, since they aren't really assertion (type challenges also calls them `helpers`)

Issue #1938, opened 2024-04-04
Use `strict: true` for TS for the SDK package


infomiho:
The `d.ts` file are generated incorrectly if we let the function values to be inferred. 

If I write a complex function like `createAuthUser` that does some data transformation, I'd like to use `typeof` to infer the final return value. Something like this:

```ts
data AuthType = ReturnType<typeof createAuthUser>

function createAuthUser() { ... }
```

Now, let's check the generate `d.ts` files for this example (the strict one being the expected type):

| ![Screenshot 2024-04-04 at 16 21 04](https://github.com/wasp-lang/wasp/assets/2223680/4bdf8b6e-1c22-4dd0-b691-ec742564ac00) |
| --- |
| `strict: true` |

| ![Screenshot 2024-04-04 at 16 21 12](https://github.com/wasp-lang/wasp/assets/2223680/4d22b50f-ecf6-4527-845f-5987011d7fbe) |
| --- |
| `strict: false` |

sodic:
https://github.com/wasp-lang/wasp/pull/2054 took care of some of the flags, but not all of them (e.g., the one which motivated this issue, `strictNullChecks` is still off).

There's also the problem of ensuring that mismatches between the user's and the SDK's `tsconfig.json` don't cause problems. This is covered by: https://github.com/wasp-lang/wasp/issues/1827#issuecomment-2133815538

Issue #1827, opened 2024-02-29
Figure out Wasp's `tsconfig` story after Wasp 0.12.0


sodic:
During the restructuring in Wasp 0.12.0, we had plans to optimize how Wasp uses different `tsconfig.json` files.

Wasp 0.12.0  ended up being a compromise between ease of implementation and DX. Now that 0.12.0 is out, we can focus on optimizing how Wasp uses and interprets `tsconfig` files.

Here's the original issues for context:
- #1430

TSconfig-related issues that are still relevant after 0.12.0:
- https://github.com/wasp-lang/wasp/issues/1057 (no longer fully applicable, but [this comment](https://github.com/wasp-lang/wasp/issues/1057#issuecomment-1525892261) explains an approach to open questions).
- https://github.com/wasp-lang/wasp/issues/835
- https://github.com/wasp-lang/wasp/issues/939

sodic:
Project references may help us have users' code compiled with their own `tsconfig.json.`:
  - **SDK:** SDK includes a full copy of the user's code under `node_modules/wasp/ext-src`. We can also copy the user's `tsconfig.json` into `node_modules/wasp/ext-src` and make `mode_modules/wasp/tsconfig.json` point to it using a project reference.
   Possible problem: If parts of the config only work in the file's original location, we'll have a problem after copying it.
  - **Framework:**  Framework imports some components from `../../../src` directly. We can use a project reference to include the user's `tsconfig.json` (which should already be applied to everything in `src`).

Reference: Effective Typescript, 2nd edition, item 78.

sodic:
Users asking for this: https://discord.com/channels/686873244791210014/1267098295327653908/1267428319096344678

PR #2054, opened 2024-05-23
Enable a subset of `strict: true` when building SDK

### LINKED ISSUE #2010: Use a subset of `strict: true` in the SDK
Building the SDK generates two kinds of files:
 - JavaScript files (`.js`) used in the runtime.
 - Declaration files (`.d.ts`) used for type support by the IDE.
 
Declaration files (`d.ts`) are subject to our user's _tsconfig.json_ which is more strict than _sdk/wasp/tsconfig.json_ used to build the SDK. This results in the SDK's declaration files having type errors when looked at from the user's perspective (e.g., during go-to-definition jumps).

We've avoided using `strict: true` in the _tsconfig.json_ we use for building the SDK due to legacy reasons (old JS or improperly typed code that still works). 

This is not a big problem, but definitely isn't ideal:
- Our SDK is deemed "wrong" by our own system, which is a little unprofessional.
- These type errors truly are type errors (they're not false-positives). Not compiling the SDK in strict mode  prevents us from catching them. In other words, a small part of legacy code disqualifies the entire SDK from reaping the full benefits of TypeScript.

Since the `strict: true` compiler flag is a shorthand for multiple more specific flags, we should activate all the "subflags" that don't require a major refactor of the SDK. After that's done, we can tackle #1938 and change whatever we need to go full strict mode (this will probably take some time).

sodic:
Fixes #2010 

> [!CAUTION]
> **Possible problem** (and breaking change):
>
> If users change their `tsconfig.json` from `strict: true` to `strict: false` and break the rules enforced in this PR, our SDK build will fail.

This isn't a big deal since the recommendation is to start all new TS projects with `strict: true`.

Also, problems caused by mismatches between the user's `tsconfig.json` and the one used to compile the SDK aren't new - it's just that `strict` options specifically didn't have this problem because the SDK was using `strict: false` (the most permissive possible setting).

The good news is that I've stumbled upon a way to fix most of our SDK-related TS config problems (and possibly even framework-code-related problems). I already have a POC and will document it with an issue ASAP. This is all already covered by #1827.

Extra stuff to do before merging: 
- [x] Update tests.
- [x] Update changelog.
- [x] Update [the issue](https://github.com/wasp-lang/wasp/issues/1938) relevant for the remaining options.
- [x] Update the relevant issue with [a comment](https://github.com/wasp-lang/wasp/issues/1827) on project references.


waspc/data/Generator/templates/sdk/wasp/client/router/types.ts
```diff
@@ -32,7 +32,7 @@ type RoutesDefinition = {
   }
 }
 
-type BuildFn = (params: unknown) => string
+type BuildFn = (params: never) => string
```
sodic, line 35: This was throwing errors after enabling [strictFunctionTypes](https://www.typescriptlang.org/tsconfig/#strictFunctionTypes).

If we wanted to say "all `build` fields in route definitions must be a subtype of this function", then `never` is the correct argument type.

waspc/data/Generator/templates/sdk/wasp/client/test/vitest/helpers.tsx
```diff
@@ -31,7 +31,7 @@ export function renderInContext(ui: ReactElement): RenderResult {
   )
   return {
     ...result,
-    rerender: (rerenderUi: ReactElement) =>
+    rerender: (rerenderUi: ReactNode) =>
```
sodic, line 34: @infomiho Thanks for linking the original code.

They indeed use `ReactElement` and then assert the entire object `as any`. [Here](https://github.com/TanStack/query/blob/aadb83122b4dba8926c09e355ffbd994e8e699cf/packages/react-query/src/__tests__/utils.tsx#L18)'s the code.

The docs say:
> Where ReactElement only represents JSX, ReactNode represents everything that can be rendered.

Any idea why they use `ReactElement`. Is that a mistake?

waspc/data/Generator/templates/sdk/wasp/client/webSocket/WebSocketProvider.tsx
```diff
@@ -36,7 +36,7 @@ apiEventsEmitter.on('sessionId.set', refreshAuthToken)
 apiEventsEmitter.on('sessionId.clear', refreshAuthToken)
 
 // PRIVATE API
-export const WebSocketContext: Context<WebSocketContextValue> = createContext({
+export const WebSocketContext: Context<WebSocketContextValue> = createContext<WebSocketContextValue>({
```
sodic, line 39: Without explicitly setting the type argument to `WebSocketContextValue`, the function `createContext` infers it as the narrowest possible type of the passed object literal.

The inferred type of `createContext`'s return value is therefore:
```typescript
Context<{ 
  socket: Socket<ServerToClientEvents, ClientToServerEvents>, 
  isConnected: false
}>
```

Since `Context<T>` is contravariant for `T`<sup>1</sup>, the compiler complains when we try to assign `Context<{ socket: Socket<...>, isConnected: false}>` to `Context<{ socket: Socket<...>, isConnected: boolean}>`.

[1] This is thanks to a function call over `T` deep inside the `Context`. See https://github.com/wasp-lang/wasp/pull/989#discussion_r1101155001 for more on variance.

Martinsos:
LGTM from far above, I will let @infomiho do the proper review!

waspc/data/Generator/templates/sdk/wasp/client/test/vitest/helpers.tsx
infomiho: I don't have any extra insights other than it might have been an error in `react-query` :)

waspc/data/Generator/templates/sdk/wasp/client/router/types.ts
infomiho: This is fine 👍 Thanks for making it correct!

infomiho (commented):
Looking good 👍

infomiho (approved):

PR #2057, opened 2024-05-24
Fix the auth helpers to only require fields they need from `AuthIdentity`

infomiho:
The auth helpers like `getEmail` and `getUsername` didn't work unless you provided the full user object with full `AuthIdentity` objects in the `identities` array.

Given that we[ suggest to users to include only the `providerName` and `providerUserId`](https://github.com/wasp-lang/wasp/blob/4bb0f9e50d6028681e3d3c6e9ffd599edc455d72/web/docs/auth/entities/entities.md?plain=1#L212) when manually fetching the user + auth + identities, we should make sure our auth helpers work in that scenario. 

This PR:
- makes it so that they only need to pass in the `providerName` and `providerUserId` when using the helpers.
- includes a headless test to ensure this helper works in the future.


waspc/data/Generator/templates/sdk/wasp/server/auth/user.ts
```diff
@@ -63,6 +63,23 @@ export type AuthEntityWithIdentities = {= authEntityName =} & {
   {= identitiesFieldOnAuthEntityName =}: {= authIdentityEntityName =}[]
 }
 
+// PRIVATE API
```
infomiho, line 66: This is basically a copy of the types we used so far, but it reduces the needed fields from _all the AuthIdentity fields_ to only the ones we really need.

waspc/src/Wasp/Generator/SdkGenerator/AuthG.hs
```diff
@@ -30,7 +30,7 @@ genAuth spec =
     Just auth ->
       -- shared stuff
       sequence
-        [ genUserTs
+        [ genFileCopy [relfile|auth/user.ts|]
```
infomiho, line 33: No need for templating anymore for this file, I've used a copy file draft instead.

waspc/headless-test/examples/todoApp/src/server/queries.ts
```diff
@@ -1,29 +1,54 @@
...
+              },
+            },
+          },
+        },
+      },
+    },
+  })
   return tasks
-}
+}) satisfies GetTasks<void>
```
sodic, line 39: How come the frontend type tests aren't complaining about this?

infomiho:
Good point! The tricky part with the `UserEntityWithAuth` name is that it is correct in a sense that it represents a tree of full Prisma entities. The second type is derived from it by making it _as minimal as possible_ to make the function as flexible as possible. 

The problem is that the user will see this wacky (albeit correct) name. 

Some alternatives:
- internal: `UserEntityWithFullAuth`, public: `UserEnittyWithAuth`
- internal `AuthEntityTree`, public: `UserEnittyWithAuth`
- internal: `UserEntityWithAuthFields`, public: `UserWithRequiredAuthFields`

I maybe like the first option the best since it's just a `Full` vs nothing which gives us the relationship you suggested.

waspc/data/Generator/templates/sdk/wasp/auth/user.ts
```diff
@@ -49,7 +51,7 @@ function makeAuthUser(data: AuthUserData): AuthUser {
   };
 }
 
-function findUserIdentity(user: UserEntityWithAuth, providerName: ProviderName): {= authIdentityEntityName =} | null {
+function findUserIdentity(user: UserEntityWithAuth, providerName: ProviderName): UserEntityWithAuth['auth']['identities'][number] | null {
```
sodic, line 54: Nice!

waspc/headless-test/examples/todoApp/src/client/Todo.tsx
```diff
@@ -130,6 +132,8 @@ const TaskView = ({ task }: { task: Task }) => {
     }
   }
 
+  const email = getEmail(task.user)
```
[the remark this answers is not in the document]
sodic: What do you mean here?

sodic (approved):
Exactly what I had in mind, nice work!

waspc/headless-test/examples/todoApp/src/client/Todo.tsx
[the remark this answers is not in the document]
infomiho: Using the `getEmail` helper in headless tests app to test it in headless tests :)

PR #2068, opened 2024-05-27
Made sidebar in docs nicer while also adding link to examples.

### LINKED ISSUE #2009: Point to examples more prominently in the docs
I notice people often ask about more Wasp examples hm. We do have them and there is a link to them at our landing page https://github.com/wasp-lang/wasp/tree/release/examples, but it seems people don't see them. Probably they jump into docs, and it seems then they are not prominent enough in the docs!
We should make them more prominent in our docs. I suggest adding a link to examples in the docs at the end of the Tutorial, and to also have a special section just for them, so it appears on the left in the sidebar, probably pretty high, under "Getting Started" or "Tutorial".

Martinsos:
Main thing I wanted to do was add Examples link to the top level.
But to do that, I have to update the styling to be more uniform as it didn't account for having non-category stuff at top level.

This is what it looked like before this PR:

image

This is what it looks like now with this PR (notice also additional Examples link in the sidebar at the top level):

image

Closes #2009 .

### DIFF
--- a/web/sidebars.js
+++ b/web/sidebars.js
@@ -14,7 +14,7 @@ module.exports = {
     {
       type: 'category',
       label: 'Tutorial',
-      collapsed: true,
+      collapsed: false,
       collapsible: true,
       items: [
         'tutorial/create',
@@ -34,7 +34,7 @@ module.exports = {
     {
       type: 'category',
       label: 'Data Model',
-      collapsed: true,
+      collapsed: false,
       collapsible: true,
       items: [
         'data-model/entities',
@@ -55,7 +55,7 @@ module.exports = {
     {
       type: 'category',
       label: 'Authentication',
-      collapsed: true,
+      collapsed: false,
       collapsible: true,
       items: [
         'auth/overview',
@@ -79,7 +79,7 @@ module.exports = {
     {
       type: 'category',
       label: 'Project Setup',
-      collapsed: true,
+      collapsed: false,
       collapsible: true,
       items: [
         'project/starter-templates',
@@ -97,14 +97,14 @@ module.exports = {
     {
       type: 'category',
       label: 'Wasp AI',
-      collapsed: true,
+      collapsed: false,
       collapsible: true,
       items: ['wasp-ai/creating-new-app', 'wasp-ai/developing-existing-app'],
     },
     {
       type: 'category',
       label: 'Advanced Features',
-      collapsed: true,
+      collapsed: false,
       collapsible: true,
       items: [
         {
@@ -129,14 +129,14 @@ module.exports = {
     {
       type: 'category',
       label: 'General',
-      collapsed: true,
+      collapsed: false,
       collapsible: true,
       items: ['general/language', 'general/cli'],
     },
     {
       type: 'category',
       label: 'Miscellaneous',
-      collapsed: true,
+      collapsed: false,
       collapsible: true,
       items: [
         'contributing',
--- a/web/sidebars.js
+++ b/web/sidebars.js
@@ -4,7 +4,7 @@ module.exports = {
       type: 'category',
       label: 'Getting Started',
       collapsed: false,
-      collapsible: false,
+      collapsible: true,
       items: [
         'introduction/introduction',
         'introduction/quick-start',
@@ -14,8 +14,8 @@ module.exports = {
     {
       type: 'category',
       label: 'Tutorial',
-      collapsed: false,
-      collapsible: false,
+      collapsed: true,
+      collapsible: true,
       items: [
         'tutorial/create',
         'tutorial/project-structure',
@@ -27,10 +27,15 @@ module.exports = {
       ],
     },
     {
+      type: 'link',
+      label: 'Examples',
+      href: 'https://github.com/wasp-lang/wasp/tree/release/examples'
+    },
+    {
       type: 'category',
       label: 'Data Model',
-      collapsed: false,
-      collapsible: false,
+      collapsed: true,
+      collapsible: true,
       items: [
         'data-model/entities',
         {
@@ -50,8 +55,8 @@ module.exports = {
     {
       type: 'category',
       label: 'Authentication',
-      collapsed: false,
-      collapsible: false,
+      collapsed: true,
+      collapsible: true,
       items: [
         'auth/overview',
         'auth/ui',
@@ -74,8 +79,8 @@ module.exports = {
     {
       type: 'category',
       label: 'Project Setup',
-      collapsed: false,
-      collapsible: false,
+      collapsed: true,
+      collapsible: true,
       items: [
         'project/starter-templates',
         'project/customizing-app',
@@ -92,15 +97,15 @@ module.exports = {
     {
       type: 'category',
       label: 'Wasp AI',
-      collapsed: false,
-      collapsible: false,
+      collapsed: true,
+      collapsible: true,
       items: ['wasp-ai/creating-new-app', 'wasp-ai/developing-existing-app'],
     },
     {
       type: 'category',
       label: 'Advanced Features',
-      collapsed: false,
-      collapsible: false,
+      collapsed: true,
+      collapsible: true,
       items: [
         {
           type: 'category',
@@ -124,15 +129,15 @@ module.exports = {
     {
       type: 'category',
       label: 'General',
-      collapsed: false,
-      collapsible: false,
+      collapsed: true,
+      collapsible: true,
       items: ['general/language', 'general/cli'],
     },
     {
       type: 'category',
       label: 'Miscellaneous',
-      collapsed: false,
-      collapsible: false,
+      collapsed: true,
+      collapsible: true,
       items: [
         'contributing',
         'telemetry',
--- a/web/src/css/custom.css
+++ b/web/src/css/custom.css
@@ -86,8 +86,8 @@
 
   /* Docs */
   --ifm-menu-color: var(--ifm-color-gray-800);
-  --sidebar-category-color: var(--ifm-color-gray-600);
-  --sidebar-category-spacing: 2em;
+  --sidebar-item-level-1-color: var(--ifm-color-primary);
+  --sidebar-item-level-1-spacing: 1em;
 }
 
 :root[data-theme='dark'] {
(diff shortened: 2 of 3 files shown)


web/src/css/custom.css
```diff
@@ -116,21 +116,18 @@
...
+  font-weight: bold;
 }
 
-.theme-doc-sidebar-item-category-level-1
-  > .menu__list-item-collapsible
-  > .menu__link {
-  color: var(--sidebar-category-color);
-  font-weight: bold;
-  font-size: var(--ifm-h6-font-size);
+.theme-doc-sidebar-item-category > .menu__list-item-collapsible > .menu__link {
```
infomiho, line 129:
```suggestion
.theme-doc-sidebar-item-category-level-1 > .menu__list-item-collapsible > .menu__link {
```
Might I suggest this change, it feels cleaner since only the top level is yellow.

| Screenshot 2024-05-28 at 10 56 36 |  Screenshot 2024-05-28 at 11 00 38 |
| --- | --- |
| Before | After |

web/sidebars.js
```diff
@@ -14,8 +14,8 @@ module.exports = {
     {
       type: 'category',
       label: 'Tutorial',
-      collapsed: false,
-      collapsible: false,
+      collapsed: true,
```
infomiho, line 17: We kinda lost the _overview of all the topics_ with everything being collapsed. We had this in the old docs if I'm not mistaken and we decided to keep them expanded in this latest iteration.

For example, data model is not indicative of operations or advanced is not indicative of jobs. You have seen _jobs_ and now you want to know more, but you have to expand multiple sections to find them.

Screenshot 2024-05-28 at 11 01 53

web/src/css/custom.css
Martinsos: But I did it on purpose, to make it clear what is "openable". Otherwise, once you open those Operations, it is a "see" of white/black, it is hard to see who is parent a bit. Notice only top level is bold though.

web/sidebars.js
Martinsos: Yeah, but there was so much stuff you really couldn't see anything. So instead of having a top level overview, you had a glimpse into 20% of the docs and that is it.

This way you can scan and find your way around.

If we have important stuff we want to take up, we can do that by organizing a new category. For example, take Operations to the top level. Or, we can put data model and ops and more stuff under "Foundations", and then the rest under "Advanced".

So I am not saying we can't improve this, but I think the old solution stopped working at some point when the docs grew, and this is the "future".

infomiho: > Yeah, but there was so much stuff you really couldn't see anything. So instead of having a top level overview, you had a glimpse into 20% of the docs and that is it.

It was more like 50%, but that's fine since users know how to scroll. It's much easier to scroll than to open 5 sub menus, not to mention the feelings you get while looking for something you can't find or even worse not knowing where to start.

I'll let @sodic comment of this as well, but we kinda took this from Vue.js docs: https://vuejs.org/guide/introduction.html since they were the gold standard in our eyes.

Martinsos: We could in theory go somewhere in between: have some categories collapsed by default (like Tutorial and Auth, those don't benefit from being open), and some open by default (e.g. Advanced).
Although it seems to me like some kind of reorganization would be the best solution.

sodic (commented):
I gotta say I'm not a fan of the change, sorry :sweat_smile: 

I strongly prefer what we had before. So, if it's a vote, I vote against it.

I understand the motivation, but I don't think this is the right approach.

Intead, we should choose what to feature in the sidebar instead of featuring everything and collapsing it. We can maybe collapse the unimportant of the stuff on the bottom.

Perhaps even putting the important stuff first is enough, without collapsing anything

If we do go through it, then I suggest:
 - Dropping the yellow text, it looks weird.
 - Definitely dropping the red rectangle.

web/sidebars.js
Martinsos: @infomiho for me it was a huge mess on the left, very hard to get any kind of an overview, or what is important or what is not. I keeping having that issue very time I look at the docs. This finally enables me to navigate the top level and dive to the depth I need.

Martinsos:
@infomiho @sodic from what I get, you like the old layout because it allowed you to scroll through stuff and get a feel for all stuff that is part of the docs, right?
I can understand/see that.

For me, what wasn't good is that I had hard time navigating through that noise, it is too much at the same time, and instead of at one glimpse seeing what docs offer and then diving deeper into what I need, I have to scroll through a bunch of information that additionally all in the same color to make it harder to distinguish.
I believe this also sounds like a valid concern, and very likely others are experiencing this also. Do you think this point has some weight?

I would say both are valid requirements.

Why don't we do something in the middle? People certainly don't need to see all the sections of tutorial on first glance, just Tutorial is enough. Same goes for Auth. While Advanced we can leave open, so they can see all the features.

So what I am suggesting is we collapse the category that create more noise than bring information, and leave open those that do the opposite.

How does that sound?

infomiho:
> For me, what wasn't good is that I had hard time navigating through that noise, it is too much at the same time, and instead of at one glimpse seeing what docs offer and then diving deeper into what I need, I have to scroll through a bunch of information that additionally all in the same color to make it harder to distinguish.

I see that side of the argument and of course it will a trade off between visibility and cleanliness. I'd like to point out that the way you navigate or perceive docs as the framework author is different than a beginner or intermediate users. That's why I'm reiterating the source for the layout: Vue.js docs which have a bunch of _docs maintainers_ and iterated on them a lot.

Martinsos:
> > For me, what wasn't good is that I had hard time navigating through that noise, it is too much at the same time, and instead of at one glimpse seeing what docs offer and then diving deeper into what I need, I have to scroll through a bunch of information that additionally all in the same color to make it harder to distinguish.
> 
> I see that side of the argument and of course it will be a trade off between visibility and cleanliness. I'd like to point out that the way you navigate or perceive docs as the framework author is different than a beginner or intermediate users. That's why I'm reiterating the inspiration for the current layout: Vue.js docs which have a bunch of _docs maintainers_ and iterated on them a lot.

That's a fair point and has weight, but:
1. Me being who I am shouldn't completely devalue my opinion and ability to put myself in other's shoes.
2. We can't really say if that sidebar is the worst or the best piece of their docs, even if many people are working on that.

That said, what others are doing is a good enough way to resolve this situation where we don't have a better way to resolve it, I agree with that, I just don't want us to accept blindly that what others are doing must be the best way to do it.

Btw I just took a look at Vue docs and I see that they first have, as a dropdown, option to choose between Guide, TUtorial, Examples, and couple more things. Then, once you go into Guide, you get this expanded list on the left that is very simple. So they in a sense do have a step first where you get an ability to choose from high level. Although it is quite different also. And that sidebar in Guide doesn't look like much thought was put into it -> maybe it is and this minimalism is the perfection, but maybe there just wasn't.

Let's take a look at React: https://react.dev/learn and https://react.dev/reference/react . I would say their approach is closer to the change I proposed here? The whole Sidebar fits on the screen, and you can expand it as needed. There are some differences, but I would say it is obvious they are more on the side of high level overview that you expand, than everything listed out for you and you nee dot scroll through it.

Svelte: It is all expanded, but again, it fits on screen. Smaller font also, easier to have an overview.

Ok let's take a look at something with bigger API surface, like Wasp: Next. 
They have stuff on top level, but it almost fits on the screen, and each of those things opens to a bunch of stuff in it. So it does go quite more in depth than into surface. And, they are even able to fit API reference on the screen in the sidebar. So I would say closer to this PR than current state.

Laravel: high level overview, all collapsed, like this PR: https://laravel.com/docs/11.x/readme .

Remix: high level overview, all collapsed, like this PR:
https://remix.run/docs/en/main .

Tailwind: all expanded, doesn't fit at all into screen, so closest to Vue and us at the moment, before this PR: https://tailwindcss.com/docs/installation .


Looks quite split: some do one approach, some other, and some are somewhere in between. It doesn't give me grounds to say that what I did in this PR is the best approach for sure, but I think it does give ground to say that there is no reason why what Vue is doing is the best approach.

Martinsos:
I tried to do a couple of things at the same time here, and since some of them I believe are easy to agree on and others seem to not be so easy to agree on, I will reduce this PR to cover only the easier stuff, and I created this issue to capture the harder stuff: https://github.com/wasp-lang/wasp/issues/2071 .

web/src/css/custom.css
Martinsos: Ok, done! I don't want us to shy that much from design though, we can try, but your are right if docusaurus has it that way, probably be should also.

web/sidebars.js
Martinsos: I dropped this part our of this PR to be discussed separately!

Martinsos:
Ok, so I reverted the PR to how it was before, all expanded.
I also reverted the change I did where non-top-level categories were colored, now they are not colored any more.

So now sidebar looks like in the following image.

image

sodic:
As mentioned, I don't think subsections should be yellow. It feels like we shouldn't use our highlight color for something that isn't highlighted/selected. If you ask me, only the currently selected page should be yellow (as is the case now).

> I also reverted the change I did where non-top-level categories were colored, now they are not colored any more.

Hm, not sure what this is referring to by looking at the screenshots.

I'm also not a fan of collapsable sections (as explained above). I'd rather keep flat.

All in all, I still don't see these changes as a clear improvement and prefer the older look. If it were up to me, **I'd just add the Examples link and leave the design changes for another time.**

That said, if you guys want to change it and have done some research (I'm just looking at it and saying what I think), I won't object.

Martinsos:
@sodic I can reverse the yellow color, although I liked it, but I get the argument.

Some of the design changes (spacing and some small fixes) I will have to keep because adding Examples wasn't working otherwise (we never had such top level link so design for it was broken).

Collapsable stuff -> not it is all flat and open, so that is as before, then only thing is you can collapse them manually if you wish, I don't see how is that an issue.

Martinsos:
Did a little poll on Discord here, as an additional data point we might want to consider: https://discord.com/channels/686873244791210014/1080864617347162122/1245474468810199212

sodic:
> Collapsable stuff -> now it is all flat and open, so that is as before, then only thing is you can collapse them manually if you wish, I don't see how is that an issue.

Oh, ok then, didn't know that. I thought it was collapsed.

In that case, everything's fine with me. I prefer grey to yellow (for reasons already listed), but both are fine.

Please just test it on a light background if you haven't (and include the screenshot).

Martinsos:
> > Collapsable stuff -> now it is all flat and open, so that is as before, then only thing is you can collapse them manually if you wish, I don't see how is that an issue.
> 
> Oh, ok then, didn't know that. I thought it was collapsed.
> 
> In that case, everything's fine with me. I prefer grey to yellow (for reasons already listed), but both are fine.
> 
> Please just test it on a light background if you haven't (and include the screenshot).

Tested it already, it looks good! Here is screenshot:

image

PR #2069, opened 2024-05-27
Extracted deployment to top lvl of docs sidebar.

Martinsos:
Didn't make sense to have it under Advanced/, it is the only part that has another level and it deserves its own section. Also, people should be able to easily see where the deployment part of docs are not dig for it under Advanced. This will make even more sense / look better with changes from #2068 due to styling improvements.

### DIFF
--- a/web/sidebars.js
+++ b/web/sidebars.js
@@ -102,16 +102,6 @@ module.exports = {
       collapsed: false,
       collapsible: false,
       items: [
-        {
-          type: 'category',
-          label: 'Deployment',
-          collapsed: true,
-          items: [
-            'advanced/deployment/overview',
-            'advanced/deployment/cli',
-            'advanced/deployment/manually',
-          ],
-        },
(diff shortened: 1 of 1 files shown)


web/sidebars.js
```diff
@@ -121,6 +111,17 @@ module.exports = {
         'advanced/links',
       ],
     },
+    {
```
infomiho, line 114: What do you think about moving it above the Advanced section since it's important?

infomiho (approved):
Left one comment

web/sidebars.js
Martinsos: I think deployment comes some time after learning about features, so it made sense to me to first have features (data model, auth, advanced, ...), then comes deployment.

infomiho: People might finish their app without digging into the advanced features and deploy it :) 

But I don't have a strong preference here, I'm fine either way 👍

PR #2081, opened 2024-06-05
Adds `GPT4-o` model to Wasp AI

### LINKED ISSUE #2043: Add gpt4-o to Mage
### Discussed in https://github.com/wasp-lang/wasp/discussions/2042

<div type='discussions-op-text'>

<sup>Originally posted by **PhilBrk8** May 20, 2024</sup>
could you update the mage to utilize the gpt-4o model?
As of today, gpt-4-0613 is in use and gpt-4o would provide faster and cheaper API-use</div>

infomiho:
Adds GPT4-o to the list of supported list of Wasp AI models. 

Based on https://platform.openai.com/docs/models/gpt-4o

Closes #2043

### DIFF
--- a/waspc/src/Wasp/AI/OpenAI/ChatGPT.hs
+++ b/waspc/src/Wasp/AI/OpenAI/ChatGPT.hs
@@ -101,6 +101,9 @@ data Model
   | --
     GPT_4_32k -- Alias model
   | GPT_4_32k_0613
+  | --
+    GPT_4o
+  | GPT_4o_2024_05_13
   deriving (Eq, Bounded, Enum)
 
 instance Show Model where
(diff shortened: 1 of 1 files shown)


sodic (approved):

Martinsos:
Ok sweet! We can actually merge this already, since it allows them to pick this model now via CLI via that config object, but I was also hoping to do a bit more with this:
1. Test how well it works, if it works as good as GPT4, then make GPT4o new default.
2. In CLI, in wizard where we offer them 3 different options for which model they want to use, again, if GPT4o is a good replacement, replace GPT4 with GPT4o in those.

So that way GPT4o would replace GPT4 in current setup, and GPT4 would become something you can pick only via CLI via config object, so the opposite of what we have now.

If you want to do this in this PR (maybe easiest to do it now!), let's do it, if not, then let's create new issue for it. I would go for it now while it is fresh, especially if you did some testing already.

Martinsos (commented):
Commented above :)! We can merge this but suggest going step further and making GPT4o default instead of GPT4, if it works well.

PR #2083, opened 2024-06-06
Improved default gitignore regarding .env rules.

Martinsos:
Inspired by and solves this PR https://github.com/wasp-lang/wasp/pull/1678 and some dotenv reading lately, I improved our default Wasp gitignore to have better rules for dotenv files.

Now it doesn't ignore .env.client.example and .env.server.example and any other dotenv example files.
Also, it doesn't ignore .env.client any more, because there is no need -> no secrets are anyway allowed in that file, and it can be committed to version control.
Finally, I relaxed ignoring of all dotenv files a bit -> it is enough to ignore those with .env prefix, not all that contain .env. anywhere inside of them, that was a bit too aggressive I think.

I also updated the docs with info on .env.client, explaining it can be versioned. 

Pls let me know if this is ok, if so, I will proceed updating all the example apps with this and then merging.

TODO:
- [x] Update all the example apps! OpenSaas also.
- [x] Add to ChangeLog

### DIFF
--- a/examples/streaming/.gitignore
+++ b/examples/streaming/.gitignore
@@ -1,13 +1,14 @@
 .wasp/
+
 node_modules/
-**/.DS_Store
 
-# We ignore env files recognized and used by Wasp.
-.env.server
-.env.client
-
-# To be extra safe, we by default ignore any files with `.env` extension in them.
+# We by default ignore any dot-env files to avoid committing any secrets by accident.
 # If this is too agressive for you, consider allowing specific files with `!` operator,
 # or modify/delete these two lines.
-*.env
-*.env.*
+.env
+.env.*
+# Don't ignore .env.client as it can't contain any secrets.
+!.env.client
+# Don't ignore example env files.
+!.env.example
+!.env.*.example
--- a/examples/streaming/.gitignore
+++ b/examples/streaming/.gitignore
@@ -1,8 +1,7 @@
 .wasp/
-
 node_modules/
 
-# We by default ignore any dot-env files to avoid committing any secrets by accident.
+# We by default ignore any dotenv files to avoid committing any secrets by accident.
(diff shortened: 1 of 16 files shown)


examples/streaming/.gitignore
Martinsos, line 1: I started updating examples here but then stopped before I get a green light.

web/docs/project/env-vars.md
```diff
@@ -85,9 +85,10 @@ In the root of your Wasp project you can create two distinct files:
     REACT_APP_SOME_VAR_NAME=somevalue
     ```
 
-These files should not be committed to version control, and they are already ignored by default in the `.gitignore` file that comes with Wasp.
+`.env.server` should not be committed to version control as it can contain secrets and is therefore by default already ignored in the `.gitignore` file that comes with a new Wasp app.
+On the other hand, `.env.client` can be commited as it must not contain any secrets.
 
-<!-- `dotenv` files are a popular method for storing configuration: to learn more about them in general, check out the [README of the lib we use for them](https://github.com/stackbuilders/dotenv-hs). -->
```
infomiho, line 90: This text is inside of a HTML comment which means it won't be visible in the docs. Is this what we want?

infomiho (approved):
LGTM, left some minor comments

web/docs/project/env-vars.md
Martinsos: Good question I haven't noticed that! I guess somebody commented it out. I think I will comment it "in" back to life.

Martinsos: Ok, took care of this! I uncommented it + put it into "info" block.

Martinsos:
@infomiho I took care of all the comments + updated all the examples apps + open saas (in separate PR) + updated Changelog. Will be merging now!

PR #2090, opened 2024-06-08
Again updated default .gitignore: removed !.env.client.

Martinsos:
Follow up on https://github.com/wasp-lang/wasp/pull/2083 -> I realized after working on open-saas it is actually better to not by default include .env.client. We will instead in the future add a piece of the docs where we will tell them about the idea of versioning .env.client -> but that should be done as a separate thing, together with the idea of vault for storing .env.server and so on.

So, for now I removed `!.env.client` from default gitignore.

### DIFF
--- a/web/docs/project/env-vars.md
+++ b/web/docs/project/env-vars.md
@@ -85,7 +85,7 @@ In the root of your Wasp project you can create two distinct files:
     REACT_APP_SOME_VAR_NAME=somevalue
     ```
 
-`.env.server` should not be committed to version control as it can contain secrets, while `.env.server` can be versioned as it must not contain any secrets.
+`.env.server` should not be committed to version control as it can contain secrets, while `.env.client` can be versioned as it must not contain any secrets.
 By default, in the `.gitignore` file that comes with a new Wasp app, we ignore all dotenv files.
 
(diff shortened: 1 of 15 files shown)


web/docs/project/env-vars.md
```diff
@@ -85,8 +85,8 @@ In the root of your Wasp project you can create two distinct files:
     REACT_APP_SOME_VAR_NAME=somevalue
     ```
 
-`.env.server` should not be committed to version control as it can contain secrets and is therefore by default already ignored in the `.gitignore` file that comes with a new Wasp app.
-On the other hand, `.env.client` can be committed as it must not contain any secrets.
+`.env.server` should not be committed to version control as it can contain secrets, while `.env.server` can be versioned as it must not contain any secrets.
```
infomiho, line 88: Typo, did you mean to say .env.client?

Martinsos, line 88:
```suggestion
`.env.server` should not be committed to version control as it can contain secrets, while `.env.client` can be versioned as it must not contain any secrets.
```

Martinsos: Thanks, fixed!

infomiho (approved):

PR #2095, opened 2024-06-11
Update e2e tests to work with Prisma file

infomiho:
Update the e2e tests to use the `schema.prisma` file.


waspc/e2e-test/ShellCommands.hs
```diff
@@ -83,6 +88,14 @@ insertCodeIntoFileAtLineNumber fileName atLineNumber line =
         "mv " ++ fileName ++ ".tmp " ++ fileName
       ]
 
+replaceLineInFile :: FilePath -> Int -> String -> ShellCommandBuilder ShellCommand
```
Martinsos, line 91: Ha these e2e tests might be a good place to use .diff s! Just an idea for the future.

Martinsos (approved):

Martinsos:
@infomiho I see that hooks are being added here now? Isn't that beyond scope of this PR?

PR #2130, opened 2024-06-26
Update Entities docs page

### LINKED ISSUE #2125: Update data model pages
We want to update the **data model** pages to use Prisma file models instead of entities in the Wasp file.

We want do something like this:
- We want to describe how Wasp parses the Prisma file and extracts the entities. Then those entities can referenced in the Wasp file for the `userEntity` or as `entities` in operations. 
- Explain how to define entities in the `prisma.schema` file.
- Explain how to work with entities
- Update the operations, CRUD and databases pages to make sense

infomiho:
Closes #2125

I've didn't expand too much on the Wasp <> Prisma relationship because I plan on writing a new page titled "Prisma file" which will include diagrams and in-depth explanations.


infomiho:
I've added an extra toggle to explain the `Entity vs Model` bit 

https://github.com/wasp-lang/wasp/assets/2223680/24d6ae1c-7894-4b45-8a5c-e3be2cdf6ca5

PR #2131, opened 2024-06-26
Adds Prisma file page in the docs

### LINKED ISSUE #2126: Create a new Prisma configuration page
We want to create a new page under the **data model** category that describes how the Prisma file is used by Wasp and how the user can configure various Prisma properties.

We should explain that Wasp:
- requires the `datasource` with `url` set to `env("DATABASE_URL")`
- requires one generator with the `prisma-client-js` provider
- allows only `postgresql` and `sqlite` DB providers
- passes through all other properties

infomiho:
Closes #2126

### DIFF
--- a/web/docs/data-model/prisma-file.md
+++ b/web/docs/data-model/prisma-file.md
@@ -6,7 +6,7 @@ import ImgWithCaption from '@site/blog/components/ImgWithCaption'
 
 Wasp uses [Prisma](https://www.prisma.io/) to interact with the database. Prisma is a "Next-generation Node.js and TypeScript ORM" that provides a type-safe API for working with your database.
 
-With Prisma, you define your application's data model in a `schema.prisma` file. 
+With Prisma, you define your application's data model in a `schema.prisma` file. Read more about how are Wasp Entities and Prisma models related on the [Entities](./entities.md)](./entities.md) page.
 
 In Wasp, the `schema.prisma` file is located in your project's root directory:
 
@@ -21,7 +21,7 @@ In Wasp, the `schema.prisma` file is located in your project's root directory:
 └── vite.config.ts
 ```
 
- Wasp uses the `schema.prisma` file to understand your app's data model and generate the necessary code to interact with the database.
+Wasp uses the `schema.prisma` file to understand your app's data model and generate the necessary code to interact with the database.
 
 ## Wasp file and Prisma schema file
 
@@ -103,7 +103,7 @@ In the implementation of the `getTasks` query, `Task` is a Wasp Entity that corr
 
 The same goes for the `myJob` job and `fooBar` API, where `Task` is used as an Entity.
 
-To read more about the relationship between Wasp Entities and Prisma models, check out the [Entities](./entities.md) page.
+To learn more about the relationship between Wasp Entities and Prisma models, check out the [Entities](./entities.md) page.
 
 ## Wasp-specific Prisma configuration
 
--- a/web/docs/data-model/entities.md
+++ b/web/docs/data-model/entities.md
@@ -4,7 +4,7 @@ title: Entities
 
 Entities are the foundation of your app's data model. In short, an Entity defines a model in your database.
 
-Wasp uses the excellent [Prisma ORM](https://www.prisma.io/) to implement all database functionality and occasionally enhances it with a thin abstraction layer. This means that you use the `schema.prisma` file to define your database models and relationships. Wasp understands the Prisma schema file and picks up all the models you define there.
+Wasp uses the excellent [Prisma ORM](https://www.prisma.io/) to implement all database functionality and occasionally enhances it with a thin abstraction layer. This means that you use the `schema.prisma` file to define your database models and relationships. Wasp understands the Prisma schema file and picks up all the models you define there. You can read more about this in the [Prisma Schema File](./prisma-file.md) section of the docs.
 
 In your project, you'll find a `schema.prisma` file in the root directory:
 
@@ -18,7 +18,7 @@ In your project, you'll find a `schema.prisma` file in the root directory:
 └── vite.config.ts
 ```
 
-Prisma uses the **_Prisma Schema Language (PSL)_**, a simple definition language explicitly created for defining models.
+Prisma uses the _Prisma Schema Language_, a simple definition language explicitly created for defining models.
 The language is declarative and very intuitive. We'll also go through an example later in the text, so there's no need to go and thoroughly learn it right away. Still, if you're curious, look no further than Prisma's official documentation:
 
 - [Basic intro and examples](https://www.prisma.io/docs/orm/prisma-schema/overview)


web/docs/data-model/prisma-file.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Prisma File
```
Martinsos, line 2:
```suggestion
title: prisma.schema File
```
Maybe like this? Or you think Prisma File is understandeable?

web/docs/data-model/prisma-file.md
```diff
@@ -0,0 +1,11 @@
...
+---
+
+TODO: create a diagram showing how the Wasp file takes entities from the Prisma file
```
Martinsos, line 5: Interesting I am curious to see what it turns out like!

web/docs/data-model/prisma-file.md
```diff
@@ -0,0 +1,11 @@
...
+
+TODO: talk about how Wasp understands the Prisma file
```
Martinsos, line 7: I think we will have some repetition to what we said in entities.md. But we can see, some repetition is probably fine, and some stuff we can just link to.

Martinsos:
Looking forward to this one!

web/docs/data-model/prisma-file.md
infomiho: We could call it `Prisma Schema file` which feels more like a title. We also call the Wasp file, Wasp file vs. `main.wasp` in our docs.

web/docs/data-model/prisma-file.md
infomiho: Nothing too fancy since it’s a simple relationship :D

web/docs/data-model/prisma-file.md
```diff
@@ -0,0 +1,200 @@
...
+}
+```
+
+Wasp reads this `schema.prisma` file and extracts the info about your database models and database config:
+
+<ImgWithCaption alt="Relationship between Wasp file and Prisma file" source="img/data-model/prisma_in_wasp.png" caption="Relationship between Wasp file and Prisma file"/>
+
+Wasp uses the `datasource` block to know which database you want to use (PostgreSQL in this case) and some other options.
+
+Wasp uses the `generator` block to generate the Prisma Client code that you can use in your application to interact with the database.
```
Martinsos, line 62: Well not really Wasp, Prisma does this.

web/docs/data-model/prisma-file.md
```diff
@@ -0,0 +1,200 @@
...
+
+Finally, Prisma models represent Wasp Entities which can be then used in the `main.wasp` file:
```
Martinsos, line 64:
```suggestion
Finally, Prisma models become Wasp Entities which can be then used in the `main.wasp` file:
```
`become` is maybe better? Not sure.

web/docs/data-model/prisma-file.md
```diff
@@ -0,0 +1,200 @@
...
+datasource db {
+  provider = "postgresql"
+  url      = env("DATABASE_URL")
+}
+```
+
+Wasp will take the `datasource` you write and use it as-is.
+
+There are some rules you need to follow:
+- You can only use `"postgresql"` or `"sqlite"` as the `provider` because Wasp only supports PostgreSQL and SQLite databases.
```
Martinsos, line 122:
```suggestion
- You can only use `"postgresql"` or `"sqlite"` as the `provider` because Wasp only supports PostgreSQL and SQLite databases for now.
```

web/docs/data-model/prisma-file.md
```diff
@@ -0,0 +1,200 @@
...
+  user        User       @relation(fields: [userId], references: [id])
+  userId      Int
+}
+```
+
+You can define your models in any way you like, if it's valid Prisma schema code, it will work with Wasp.
+
+Some minor caveats to keep in mind:
+
+- Wasp doesn't support `/// comment` syntax in the `schema.prisma` file. We are tracking it [here](https://github.com/wasp-lang/wasp/issues/2132), let us know if this is something you need.
```
Martinsos, line 158:
```suggestion
- Wasp doesn't fully support `/// comment` syntax in the `schema.prisma` file yet. We are tracking it [here](https://github.com/wasp-lang/wasp/issues/2132), let us know if this is something you need.
```
I added "yet" because you can write those comments if I am correct, so we will not crash on them, but yeah, we will not forward them to genreated code.

web/docs/data-model/prisma-file.md
```diff
@@ -0,0 +1,200 @@
...
+- Wasp doesn't support MongoDB-specific [composite types](https://www.prisma.io/docs/orm/reference/prisma-schema-reference#type) in the `schema.prisma` file.
```
Martinsos, line 159:
```suggestion
- Wasp doesn't yet support MongoDB-specific [composite types](https://www.prisma.io/docs/orm/reference/prisma-schema-reference#type) in the `schema.prisma` file.
```

web/docs/data-model/prisma-file.md
```diff
@@ -0,0 +1,200 @@
...
+  provider        = "prisma-client-js"
+  // highlight-next-line
+  previewFeatures = ["postgresqlExtensions"]
+}
+```
+
+<small>
+
+Read more about Prisma preview features in the [Prisma docs](https://www.prisma.io/docs/concepts/components/preview-features/client-preview-features).
+</small>
```
Martinsos, line 178: This is Prisma sepcific stuff, should we be having this here at all?

web/docs/data-model/prisma-file.md
```diff
@@ -0,0 +1,200 @@
...
+  // highlight-next-line
+  extensions = [hstore(schema: "myHstoreSchema"), pg_trgm, postgis(version: "2.1")]
+}
+
+// ...
+```
+
+<small>
+
+Read more about PostgreSQL configuration in Wasp in the [API Reference](#the-appdb-field).
```
Martinsos, line 199: I thought `app.db` field doesn't exist anymore? Ok it does for seeds, but that is it, no configuration of db happens there any more? Is this line redundant?

web/docs/data-model/prisma-file.md
```diff
@@ -0,0 +1,200 @@
...
+```prisma title="schema.prisma"
+datasource db {
+  provider   = "postgresql"
+  url        = env("DATABASE_URL")
+  // highlight-next-line
+  extensions = [hstore(schema: "myHstoreSchema"), pg_trgm, postgis(version: "2.1")]
+}
+
+// ...
+```
```
Martinsos, line 195: This is also strictly Prisma related. Although I guess it might be interesting since it is PostrgreSQL. But still, I would either remove it, or shorten it to be one or two sentences with a URL to prisma docs and that is it, like a super short FYI: "For how to use Postgre extensions with Prisma, check URL".

web/docs/data-model/prisma-file.md
Martinsos: This is the first moment where you introduce entities, in this doc, but it doesn't feel like detailed enough introduction. I know thi sis not super simple, but I would enrich it a bit. Explain that Wasp has the context of Entity, and these are data models in the context of Wasp, and that the way to define Entities is currently by defining Prisma models. And then also link to the part of docs where we explain it better. You could even maybe start the whole doc with this hm -> but ok maybe not, it can be more about how we use Prisma and then this we touch when we talk about models, ok.

web/static/img/data-model/prisma_in_wasp.png
Martinsos, line 1: Love the style! But I found diagram a bit confusing.

What are we trying to say with this diagram?

I think we can't skip the entities here.

So I would maybe have models -> entities, instead of models -> jobs, apis, operations. Yeah, so you can add entities on the right side, and have models point to them. That makes it clear what they mean.

Although the whole diagram potentially feels a bit reundandnt, I think you wanted to make it cooler but realized there is not much to put into it? Maybe we can just drop it if you wish. ALthough I like diagrams, and this one could still explain a bit that message about entities <> models.

Btw I would consider putting main.wasp on the left, as it is the "main" character, and them schema on the right, to look a bit like it "works for" the main.wasp, not the opposite.

web/docs/data-model/prisma-file.md
infomiho: I'd rather link to the Entities page where we explained that bit 👍

web/docs/data-model/prisma-file.md
infomiho: I wanted to keep the Prisma extensions mention because of the `pg_vector` mention which is useful and popular these days. 

I've reduced the amount of info to "hey there are some preview features, one of them is postgresql extensions".

web/static/img/data-model/prisma_in_wasp.png
infomiho: I've updated the diagram 👍

web/docs/data-model/prisma-file.md
Martinsos: Ok let's link at least then -> I thought a bit of redunandcy might be a good fit here, but you do as you think is best!

Martinsos (approved):

PR #2139, opened 2024-07-01
Fixes missing `node_modules` causing errors

infomiho:
### Reproduce the bug

1. `wasp new todo`
2. `cd todo`
3. `wasp compile`
4. `rm -rf node_modules`
5. `wasp compile`

I get this error:
```

✅ --- Successfully completed npm install. ----------------------------------------


🐝 --- Building SDK... ------------------------------------------------------------


[  Wasp  ]
[  Wasp  ]
[  Wasp  ]                 This is not the tsc command you are looking for
[  Wasp  ]
[  Wasp  ]
[  Wasp  ] To get access to the TypeScript compiler, tsc, from the command line either:
[  Wasp  ]
[  Wasp  ] - Use npm install typescript to first add TypeScript to your project before using npx
[  Wasp  ] - Use yarn to avoid accidentally running code from un-installed packages

- SDK build failed with exit code: 1


❌ --- [Error] Compilation of wasp project failed: --------------------------------

1 errors found
```

Under certain circumstances (deleting the Prisma checksum file) I would get this as well:
```
wasp compile

✅ --- Successfully completed npm install. ----------------------------------------


wasp-bin: /private/tmp/xxx/node_modules/.bin/prisma: streamingProcess: exec: invalid argument (Bad file descriptor)
```

### Why it happens

If the npm deps cache file says we are up to date, we assume that the `node_modules` exists. This results in a crash. The `node_modules` dir might get deleted by accident by some users while debugging their app.

### DIFF
--- a/waspc/src/Wasp/Generator/NpmInstall.hs
+++ b/waspc/src/Wasp/Generator/NpmInstall.hs
@@ -49,7 +49,7 @@ installNpmDependenciesWithInstallRecord spec dstDir = runExceptT $ do
         <$> sequence
           [ -- Users might by accident delete node_modules dir, so we check if it exists
             -- before assuming that we don't need to install npm deps.
-            not <$> doesNodeModulesDirExistInWaspProjectDir waspProjectDirPath,
+            not <$> doesNodeModulesDirExist waspProjectDirPath,
             areThereNpmDepsToInstall allNpmDeps dstDir
           ]
 
@@ -179,7 +179,7 @@ areThereNpmDepsToInstall allNpmDeps dstDir = do
   installedNpmDeps <- loadInstalledNpmDepsLog dstDir
   return $ installedNpmDeps /= Just allNpmDeps
 
-doesNodeModulesDirExistInWaspProjectDir :: Path' Abs (Dir WaspProjectDir) -> IO Bool
-doesNodeModulesDirExistInWaspProjectDir waspProjectDirPath = IOUitl.doesDirectoryExist nodeModulesDirInWaspProjectDirAbs
+doesNodeModulesDirExist :: Path' Abs (Dir WaspProjectDir) -> IO Bool
+doesNodeModulesDirExist waspProjectDirPath = IOUitl.doesDirectoryExist nodeModulesDirInWaspProjectDirAbs
   where
     nodeModulesDirInWaspProjectDirAbs = waspProjectDirPath SP.</> nodeModulesDirInWaspProjectDir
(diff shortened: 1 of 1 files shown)


waspc/src/Wasp/Generator/NpmInstall.hs
```diff
@@ -42,26 +43,34 @@ installNpmDependenciesWithInstallRecord spec dstDir = runExceptT $ do
...
-      -- In case anything fails during installation that would leave node modules in
-      -- a broken state, we remove the log of installed npm deps before we start npm install.
-      liftIO $ forgetInstalledNpmDepsLog dstDir
+  shouldInstallNpmDeps <-
+    liftIO $
+      or
+        <$> sequence
+          [ -- Users might by accident delete node_modules dir, so we check if it exists
+            -- before assuming that we don't need to install npm deps.
+            not <$> doesNodeModulesDirExist waspProjectDirPath,
```
sodic, line 52: This can still cause problems if they do something more specific (i.e., delete some packages without deleting the entire directory).

@Martinsos What was the reason for not running NPM install every time and letting NPM see whether it needs to install anything? I know we had a reason (a bug, IIRC), but I can't remember.

sodic (approved):
I'm approving because this check probably handles 80% or more of possible errors, but I'm still curious about why we have an npm deps log so I left a question for @Martinsos.

Does this PR have an accompanying issue? If so, please link it.

waspc/src/Wasp/Generator/NpmInstall.hs
infomiho: If I understood this correctly, the optimisation was done to prevent `npm install` from running if the deps didn't change. I guess `npm` is smart and optimised enough to figure out quickly that it already has all the deps installed? 

If just run it every time without this optimization, it would be an overhead, maybe negligible but the system would be more robust.

Martinsos: Yeah hm I can't remember exactly why we have this check -> I believe it was as you both assumed, which was just to avoid unneccesary runs of `npm install`. Running `npm install` on every file save, it seemed very wasteful, and it potentially makes quite fast iteration cycle quite longer. That said, I can't remember if we tested/measuer these times or just assumed it.

Martinsos (approved):

waspc/src/Wasp/Generator/NpmInstall.hs
sodic: I think it was initially implemented by Martijn, but then (during restructuring) it either caused a bug or was necessary to fix it.

I can dig into it if it's important, but probably not, right?

PR #2140, opened 2024-07-01
Add instructions for accessing the dev database

medardm:
### Description

Add the instructions for accessing the dev database under the [Connecting to a Database](https://wasp-lang.dev/docs/data-model/backends#connecting-to-a-database) section. 

~~I think providing the default credentials for the dev database in the docs is important because Some developers might want to access the database using external tools like **pgAdmin** or **datagrip** like me. **Prisma db studio** is not enough.~~

~~Also, personally, I found it really hard to look for the credentials online, even AI doesn't know the right credentials yet.~~

There's a chance that some developers may overlook the additional info printed in the console that contains the credentials for accessing the database directly. Therefore, as suggested by @Martinsos, it would be better to add a note in the documentation about where to find the credentials instead of putting them directly in the docs, for easier maintainability.

### Type of change this PR introduces:
[x] **Just code/docs improvement** (no functional change).

### DIFF
--- a/web/docs/data-model/backends.md
+++ b/web/docs/data-model/backends.md
@@ -73,9 +73,9 @@ Also, make sure that:
 - You have [Docker installed](https://www.docker.com/get-started/) and in `PATH`.
 - The port `5432` isn't taken.
 
-### Accessing the Dev Database
-
-In case you might want to connect to the dev database through the CLI or using external tools like [pgAdmin](https://www.pgadmin.org/), the credentials are printed in the console when you run `wasp db start`.
+:::tip
+In case you might want to connect to the dev database through the external tool like `psql` or [pgAdmin](https://www.pgadmin.org/), the credentials are printed in the console when you run `wasp db start`, at the very beginning.
+:::
 
 ### Connecting to an existing database
 
--- a/web/docs/data-model/backends.md
+++ b/web/docs/data-model/backends.md
@@ -75,10 +75,7 @@ Also, make sure that:
 
 ### Accessing the Dev Database
 
-To connect to the dev database through the CLI or using tools like [pgAdmin](https://www.pgadmin.org/), use the following credentials:
-
-- User: `postgresWaspDevUser`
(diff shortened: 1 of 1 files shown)


Martinsos:
@medardm actually, when you run `wasp db start`, just take a look at the initial output in yellow color: you should see the instructions for connecting to the created database right there!

Martinsos:
So I wouldn't be adding the credentials direclty to the docs, as we might change them in the implementation and then we need to keep it in sync and there is no need. But, we could add a line in the docs that says that creds will be printed by `wasp db start` the first thing! How does that sound?

medardm:
I see my bad @Martinsos I overlooked that detail in the console. Yes, that is a better alternative. Which section should I add it? Should I just change the description of `Accessing the Dev Database` in my current PR?

medardm:
I have updated the PR @Martinsos, is this okay? 

It's an honor, by the way. Thank you for building Wasp!

web/docs/data-model/backends.md
```diff
@@ -73,6 +73,10 @@ Also, make sure that:
 - You have [Docker installed](https://www.docker.com/get-started/) and in `PATH`.
 - The port `5432` isn't taken.
 
+### Accessing the Dev Database
+
+In case you might want to connect to the dev database through the CLI or using external tools like [pgAdmin](https://www.pgadmin.org/), the credentials are printed in the console when you run `wasp db start`.
+
```
Martinsos, line 79:
```suggestion
:::tip
In case you might want to connect to the dev database through the external tool like `psql` or [pgAdmin](https://www.pgadmin.org/), the credentials are printed in the console when you run `wasp db start`, at the very start.
:::

```

Martinsos:
No worries @medardm , and thanks for adding this to the docs, I think it will be helpful, as it is relatively easy to miss that output at the start of `wasp db start`! I modified it slightly (instead of a new header I used "tip" admonition + I changed wording a bit), will be merging now. And thanks for the kind words, it means a lot :)!

PR #2145, opened 2024-07-02
Moves crud routes under `/crud`

### LINKED ISSUE #2121: Enabling CRUD feature sets the default middleware globally (breaking the API feature)
The way we define CRUD router and the apis router is conflicting. 
```ts
router.use('/', middleware, rootCrudRouter)
router.use(apis)
```

The `router.use('/', middleware, ...)` code sets the defaultmiddleware for all the routes under `/` which is ... all routes 😄 This breaks the custom middleware option of the custom API routes feature.

Related Discord thread: https://ptb.discord.com/channels/686873244791210014/1253382715135234211/1253707556614832160

### Possible fixes
- Rewrite CRUD routes to be `/crud/*` (breaking for existing users)
- Apply middleware for CRUD inside of the `rootCrudRouter` (more precise)

infomiho:
Closes #2121 

Moves CRUD routes under the `/crud` prefix so the global default middleware only gets applied to the CRUD routes instead of all routes (which includes custom API routes - this made it impossible to define custom middleware for API endpoints _and_ use the CRUD feature).

This is changing the route the client uses, but since it was never part of public API, it's not a breaking change.


sodic (approved):
Nice!

PR #2152, opened 2024-07-04
Make sure `entity` declaration is not used in the Wasp file

### LINKED ISSUE #2148: Disallow using entities in the Wasp file
Users with the current iteration of Wasp + Prisma file can still define their entities in the Wasp file.

Wasp compiler will allow it and include them in the Wasp AST. 

We didn't remove the `entity` declaration to make it easier to switch to using the `models` from the Prisma file without touching the compiler too much.

### What we need to decide

- Do we want to allow users to define entities in the Wasp file? 

  IMHO we should disallow it. It creates a cognitive dissonance for our users with two things being possible but Wasp now officially support only one of them. 

- How do we behave in that scenario? 
  
  We should make it a validation error: "Hey, entities are no longer allowed in the Wasp, please use the Prisma file at `schema.prisma` in your project's root directory ..."

infomiho:
Closes #2148 

Introduces AST validation:
- it runs after the parser 
- there is only one check: if `entity` declaration is used, throw an error

### How it looks like 

| Screenshot 2024-07-04 at 11 55 53 |
| --- |
| In the CLI |

LS in VS Code:

https://github.com/wasp-lang/wasp/assets/2223680/39143363-43de-423d-bee5-4496afb4ce27


waspc/src/Wasp/Analyzer/TypeChecker/TypeError.hs
```diff
@@ -27,15 +27,15 @@ data TypeError'
   -- We use "unify" in the TypeChecker when trying to infer the common type for
   -- typed expressions that we know should be of the same type (e.g. for
   -- elements in the list).
-  = UnificationError    TypeCoercionError
+  = UnificationError       TypeCoercionError
```
infomiho, line 30: Formatting

waspc/src/Wasp/Analyzer.hs
```diff
@@ -138,6 +139,7 @@ import qualified Wasp.Psl.Ast.Schema as Psl.Schema
 analyze :: Psl.Schema.Schema -> String -> Either [AnalyzeError] [Decl]
 analyze prismaSchemaAst =
   (left (map ParseError) . parseStatements)
+    >=> (left ((: []) . ValidationError) . validateAst)
```
sodic, line 142: Imagine opening up this PR after reviewing and writing TypeScript all day. And the first thing you see is this line here.

Oh well, time to gaslight myself into thinking Haskell is readable again :smile:

waspc/src/Wasp/Analyzer/Parser/Valid.hs
```diff
@@ -0,0 +1,27 @@
...
+where
+
+import Data.List (find)
+import qualified Wasp.Analyzer.Parser as P
+
+validateAst :: P.AST -> Either (String, P.Ctx) P.AST
+validateAst = validateNoEntityDeclInWaspFile
+
+validateNoEntityDeclInWaspFile :: P.AST -> Either (String, P.Ctx) P.AST
+validateNoEntityDeclInWaspFile ast@(P.AST stmts) = case findEntityDecl stmts of
```
sodic, line 13: Why not `statements`?

(yes, I proudly admit I hate non-standard abbreviations).

waspc/src/Wasp/Analyzer/Parser/Valid.hs
```diff
@@ -0,0 +1,27 @@
...
+  Just (P.WithCtx ctx _) -> Left (entitiesNoLongerSupportedError, ctx)
+  Nothing -> Right ast
+  where
+    findEntityDecl :: [P.WithCtx P.Stmt] -> Maybe (P.WithCtx P.Stmt)
+    findEntityDecl = find isEntityDecl
+
+    isEntityDecl :: P.WithCtx P.Stmt -> Bool
+    isEntityDecl = \case
+      P.WithCtx _ (P.Decl "entity" _ _) -> True
```
sodic, line 22: How come we need to hardcode "entity" here? I haven't looked at this code for a long time so it might be necessary, but it seems weird.

waspc/src/Wasp/Analyzer/Parser/Valid.hs
```diff
@@ -0,0 +1,27 @@
...
+validateAst :: P.AST -> Either (String, P.Ctx) P.AST
+validateAst = validateNoEntityDeclInWaspFile
+
+validateNoEntityDeclInWaspFile :: P.AST -> Either (String, P.Ctx) P.AST
+validateNoEntityDeclInWaspFile ast@(P.AST stmts) = case findEntityDecl stmts of
+  Just (P.WithCtx ctx _) -> Left (entitiesNoLongerSupportedError, ctx)
+  Nothing -> Right ast
+  where
+    findEntityDecl :: [P.WithCtx P.Stmt] -> Maybe (P.WithCtx P.Stmt)
+    findEntityDecl = find isEntityDecl
```
sodic, line 18: You can pattern-match the maybe here and return a boolean, no need to propagate it.

You can call the function `includesEntities :: [P.WithCtx P.Stmt] -> Bool`

sodic (approved):
All seems well, but let me just make sure I understand what's going on.

Can you explain why entities can't be a `ParseError` (as one would expect) and rather require their own phase in the compilation pipeline:
- Is it to provide a better error message?
- Is it because we hooked into a part of the pipeline in a way that prevents us from changing the parser?

waspc/src/Wasp/Analyzer/Parser/Valid.hs
infomiho: The data type is defined as:
```haskell
newtype AST = AST {astStmts :: [WithCtx Stmt]} deriving (Eq, Show)
```
so it seemed fitting to use `stmts` :D

waspc/src/Wasp/Analyzer/Parser/Valid.hs
infomiho: Well spotted! We should make sure that the entity declaration type name (`"entity"`) is the same in the:
- definition of the entity declaration type
- here.

I'll make it happen 👍

waspc/src/Wasp/Analyzer/Parser/Valid.hs
infomiho: I've adjusted the code a bit 🤷

infomiho:
@sodic I've added an explanation on why we don't just throw a `ParseError`: https://github.com/wasp-lang/wasp/pull/2152/files#diff-cc433ce1c818fd8c72ac77775fb42d6a276b6a0225a3e44d162e6a286ae5cd07R143

sodic (approved):
Nice explanation!

PR #2157, opened 2024-07-05
Migrate streaming example to 0.14.0

infomiho:
This app opens up an interesting question - do we require `schema.prisma` file even for apps that don't use entities?

### DIFF
--- a/examples/streaming/schema.prisma
+++ b/examples/streaming/schema.prisma
@@ -1,0 +1,8 @@
+datasource db {
+  provider = "sqlite"
+  url      = env("DATABASE_URL")
+}
+
+generator client {
+  provider = "prisma-client-js"
+}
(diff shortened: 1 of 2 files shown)


examples/streaming/schema.prisma
```diff
@@ -0,0 +1,8 @@
+datasource db {
+  provider = "sqlite"
+  url      = env("DATABASE_URL")
+}
+
+generator client {
+  provider = "prisma-client-js"
+}
```
sodic, line 8: Missing empty line.

Can we enforce this with a tool (does PRisma do it?)

sodic (approved):

sodic:
So, we currently throw an error if the `schema.prisma` is missing even if the app has no entities.

Is this something we want to reconsider before the release or not?

infomiho:
We agreed to keep requiring the `schema.prisma` file because it simplifies the implementation. We'll evolve this as people start giving us feedback :)

PR #2159, opened 2024-07-05
Migrate todo-typescript example app to 0.14.0

infomiho:


sodic (approved):
All good.

What about `todo-typescript` in `waspc`?

infomiho:
It was already updated 👍

PR #2161, opened 2024-07-05
Removes the Real World example app

infomiho:
Since the Real World app is stuck in `0.11.0` it seems the time has come to remove it. It's outdated and we have Open Saas now which is much more real world IMHO.


sodic (approved):
I think it can go.

Martinsos (approved):
Soudns good @infomiho ! Can you just check if maybe link to it from somewhere? I think landing page could be the only such place. If we don't, all good.

infomiho:
I've searched through the codebase and remove all references I found :)

PR #2177, opened 2024-07-12
Fix db reset failing after wasp clean

sodic:
This PR fixes `wasp db reset` failing after `wasp clean`:
```
$ wasp clean

🐝 --- Deleting the .wasp/ directory... -------------------------------------------


✅ --- Deleted the .wasp/ directory. ----------------------------------------------


🐝 --- Deleting the node_modules/ directory... ------------------------------------


✅ --- Deleted the node_modules/ directory. ---------------------------------------

$ wasp db reset

🐝 --- Resetting the database... --------------------------------------------------

wasp-bin: /home/filip/Projects/wasp/wasp-repo/waspc/examples/todoApp/node_modules/.bin/prisma: streamingProcess: chdir: invalid argument (Bad file descriptor)
$ 
```

### DIFF
--- a/waspc/cli/exe/Main.hs
+++ b/waspc/cli/exe/Main.hs
@@ -18,7 +18,7 @@ import Wasp.Cli.Command.Clean (clean)
 import Wasp.Cli.Command.Compile (compile)
 import Wasp.Cli.Command.CreateNewProject (createNewProject)
 import qualified Wasp.Cli.Command.CreateNewProject.AI as Command.CreateNewProject.AI
-import qualified Wasp.Cli.Command.Db as Command.Db
+import Wasp.Cli.Command.Db (runCommandThatRequiresDbRunning)
 import qualified Wasp.Cli.Command.Db.Migrate as Command.Db.Migrate
 import qualified Wasp.Cli.Command.Db.Reset as Command.Db.Reset
 import qualified Wasp.Cli.Command.Db.Seed as Command.Db.Seed
@@ -103,9 +103,7 @@ main = withUtf8 . (`E.catch` handleInternalErrors) $ do
     Command.Call.StartDb -> runCommand Command.Start.Db.start
     Command.Call.Clean -> runCommand clean
     Command.Call.Compile -> runCommand compile
-    Command.Call.Db dbArgs -> case dbArgs of
-      ["start"] -> runCommand Command.Start.Db.start
-      _commandOnDb -> runCommandOnRunningDb dbArgs
+    Command.Call.Db dbArgs -> dbCli dbArgs
     Command.Call.Version -> printVersion
     Command.Call.Studio -> runCommand studio
     Command.Call.Uninstall -> runCommand uninstall
@@ -218,19 +216,17 @@ printVersion = do
       ]
 
 -- TODO: maybe extract to a separate module, e.g. DbCli.hs?
-
--- | Execute a command that acts on a running database.
-runCommandOnRunningDb :: [String] -> IO ()
-runCommandOnRunningDb args =
-  maybe printDbUsage Command.Db.runCommandOnRunningDb parsedCommand
-  where
-    parsedCommand = case args of
-      ["reset"] -> Just Command.Db.Reset.reset
-      "migrate-dev" : optionalMigrateArgs -> Just $ Command.Db.Migrate.migrateDev optionalMigrateArgs
-      ["seed"] -> Just $ Command.Db.Seed.seed Nothing
-      ["seed", seedName] -> Just $ Command.Db.Seed.seed $ Just seedName
-      ["studio"] -> Just Command.Db.Studio.studio
-      _unknownCommand -> Nothing
+dbCli :: [String] -> IO ()
+dbCli args = case args of
+  -- These commands require an existing and running database.
+  ["reset"] -> runCommandThatRequiresDbRunning Command.Db.Reset.reset
+  "migrate-dev" : optionalMigrateArgs -> runCommandThatRequiresDbRunning $ Command.Db.Migrate.migrateDev optionalMigrateArgs
+  ["seed"] -> runCommandThatRequiresDbRunning $ Command.Db.Seed.seed Nothing
+  ["seed", seedName] -> runCommandThatRequiresDbRunning $ Command.Db.Seed.seed $ Just seedName
+  ["studio"] -> runCommandThatRequiresDbRunning Command.Db.Studio.studio
+  -- These commands don't require an existing and running database.
+  ["start"] -> runCommand Command.Start.Db.start
+  _unknownDbCommand -> printDbUsage
 
 {- ORMOLU_DISABLE -}
 printDbUsage :: IO ()
--- a/waspc/cli/exe/Main.hs
+++ b/waspc/cli/exe/Main.hs
@@ -67,7 +67,7 @@ main = withUtf8 . (`E.catch` handleInternalErrors) $ do
         ("waspls" : _) -> Command.Call.WaspLS
         ("deploy" : deployArgs) -> Command.Call.Deploy deployArgs
         ("test" : testArgs) -> Command.Call.Test testArgs
-        _ -> Command.Call.Unknown args
+        _unknownCommand -> Command.Call.Unknown args
 
   telemetryThread <- Async.async $ runCommand $ Telemetry.considerSendingData commandCall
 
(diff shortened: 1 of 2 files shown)


waspc/cli/exe/Main.hs
```diff
@@ -98,12 +98,14 @@ main = withUtf8 . (`E.catch` handleInternalErrors) $ do
...
-      _ -> printWaspNewAiUsage
+      _unknownCommand -> printWaspNewAiUsage
     Command.Call.Start -> runCommand start
     Command.Call.StartDb -> runCommand Command.Start.Db.start
     Command.Call.Clean -> runCommand clean
     Command.Call.Compile -> runCommand compile
-    Command.Call.Db dbArgs -> dbCli dbArgs
+    Command.Call.Db dbArgs -> case dbArgs of
+      ["start"] -> runCommand Command.Start.Db.start
+      _dbCommand -> dbCli dbArgs
```
sodic, line 108: The bug was introduced when I added the `reset` case into the `dbCli` by copying the line that handles `db start` and just changing the command's name.

As it turns out, all other commands inside `dbCli` perform their task on a database and rely on the same prerequisites. `wasp db start` is the only exception.

So, to prevent this bug from happening again, I pulled the `runDbCommand` call out of the case statement in `dbCli`.

The `dbCli` function is now only in charge of the commands that act on the database (I've added that to its documentation).

That said, the whole thing does look more complicated than it maybe needs to be, so I could be convinced to change it back.

@Martinsos @infomiho What do you guys think?

infomiho: +1 for using the `runDbCommand` for every db command, makes sense to me that we check prerequisites for db commands in the same way, if they don't need it, they are some other type of command.

One change I'd propose here is that this line stays like before:
```hs
Command.Call.Db dbArgs -> dbCli dbArgs
```
and you handle the `wasp db start` case above:
```hs
["start", "db"] -> Command.Call.StartDb
["db", "start"] -> Command.Call.StartDb
```

sodic: > Or, we could just group these commands into two groups, but under dbCli: first group that applies runCommand on all of them, the other that applies runDbCommand

Can you give a code example of what you mean? I'm not sure I got it.

Martinsos: You are right, it is a bit tricky: I likely didn't go enough into details when I was imagining it in my head.
You could do it I believe by doing what you did twice: having one case for normal commands, that returns Maybe, then another for those that require a DB, that also returns Maybe, and then connect those two together and if both are Nothing return `printDbUsage`. It has its drawbacks as you could by accident have same command string in both cases and now know it, although I don' think that is likely.
But since this is a bit more complicated than I imagined, it might be easier to just keep it as it was and add comments and all good.

sodic: Ok, here's what I did in the end.

Since Miho and I prefer to have all commands under `dbCli` act on a running database, and Martin is OK with both suggestions above, I kept my change.

I also took Martin's suggestion of renaming the `runDbCommand` to something more meaningful.

The main argument is:

_If the `dbCli` function contains commands that require a database and those that don't, then its role is purely syntactic convenience (to save us from inlining all cases in main)<sup>***</sup>.
If the `dbCli` function contains only the commands that need a database, it has a meaningful semantic connection, which makes it easier to give the function a meaningful name._

That's why I renamed it from `dbCli` to `runCommandOnRunningDb`. I also changed the name of `runDbCommand` to `runCommandOnRunningDb` and used a qualified import (as we usually do).

The new name should make it obvious that all commands it runs require a running database.

I've also decided not to do this:
```haskell
["start", "db"] -> Command.Call.StartDb
["db", "start"] -> Command.Call.StartDb
```
Precisely to emphasize that, under the "db" namespace, you can run both commands that act on the database, as well as commands that don't:
```haskell
Command.Call.Db dbArgs -> case dbArgs of
  ["start"] -> runCommand Command.Start.Db.start
  _commandOnDb -> runCommandOnRunningDb dbArgs
```

To address Martin's concern:

> Yes, it has some space for the mistake that happened, but if you take runDbCommand out, one can also easily make the opposite type of mistake...

With the new names, the haddock above the function, the explicit branching under "db" in the main function, and the code path in `runCommandOnRunningDb` always going through `runDbCommand` -- I believe that both the author and the reviewer will be forced to think about whether their command needs a database running or not.

<sub>
*** You could say that the semantic connection was that they were all under the "db" namespace, but that logic was fully handled in main, and the function's call site shouldn't affect its inner semantics.
</sub>

infomiho: The new structure feels okay to me. The name is a bit weird to me: `runCommandOnRunningDb`, maybe `runCommandRequiringDb` or `runCommandWithRunningDb` or full `runCommandThatRequiresDbRunning`.

But, I'm also okay with the situation as-is.

infomiho (approved):

waspc/cli/exe/Main.hs
sodic: `runCommandThatRequiresDbRunning` - Martin suggested this one too, so I'll rename it.

waspc/cli/exe/Main.hs
```diff
@@ -98,12 +98,14 @@ main = withUtf8 . (`E.catch` handleInternalErrors) $ do
...
-      _ -> printWaspNewAiUsage
+      _unknownCommand -> printWaspNewAiUsage
     Command.Call.Start -> runCommand start
     Command.Call.StartDb -> runCommand Command.Start.Db.start
     Command.Call.Clean -> runCommand clean
     Command.Call.Compile -> runCommand compile
-    Command.Call.Db dbArgs -> dbCli dbArgs
+    Command.Call.Db dbArgs -> case dbArgs of
+      ["start"] -> runCommand Command.Start.Db.start
+      _commandOnDb -> runCommandOnRunningDb dbArgs
```
Martinsos, line 108: Ok! What's a bit weird now is that if it is not a "normal" command then we assume it is command running on db, well thtat is ok, but then we also assume that `runCommandOnRunningDb` will print usage if no command is found. Which is weird a bit, that that logic got moved into `runCommandOnRunningDb`. Also, that logic will be printing info about `wasp db start`, which is not even one of the commands in the case inside that function. So this is looking weird.

One way to improve this is to inline it directly here. It will be a bit crowded though.

Another way is to go back to waht we had before and just add comments, maybe that is best.

Third way is to make "db commands that don't need db running" and "db commands that need db running" equivalent, something like this:

```hs
Command.Call.Db dbArgs -> dbCli dbArgs

dbCli :: [String] -> IO ()
dbCli args = fromMaybe printDbUsage $
  runCommand <$> parseRegularCmd args
  <|> runDbCommand <$> parseCmdThatRequiresRunningDb args
  where
    parseRegularCmd args' = case args' of
      ["start"] -> Just Command.Db.Reset.reset
      _ -> Nothing

    parseCmdThatRequiresRunningDb args' = case args' of
      ["reset"] -> Just Command.Db.Reset.reset
      "migrate-dev" : optionalMigrateArgs -> Just $ Command.Db.Migrate.migrateDev optionalMigrateArgs
      ["seed"] -> Just $ Command.Db.Seed.seed Nothing
      ["seed", seedName] -> Just $ Command.Db.Seed.seed $ Just seedName
      ["studio"] -> Just Command.Db.Studio.studio
      _ -> Nothing
```

I haven't tried compiling the code so maybe I wrote some nonsense but the idea is to have two parsers that return Maybe and connect them with Alternative (`<|>`) and if both fail go with `printDbUsage`.

sodic: > ... but then we also assume that runCommandOnRunningDb will print usage if no command is found. Which is weird a bit, that that logic got moved into runCommandOnRunningDb. Also, that logic will be printing info about wasp db start

Ahh, I did consider that. I thought it was a little messy (having a catch-all in a nested function), but I didn't notice that it will also print info about `wasp db start`. Yeah, that's bad.

We can also do this:
```haskell
Command.Call.Db dbArgs -> case dbArgs of
  ["start"] -> runCommand Command.Start.Db.start
  _commandOnDb -> fromMaybe printDbUsage (runCommandOnRunningDbIfValid dbArgs)

runCommandOnRunningDbIfValid :: [String] -> Maybe (IO ())
runCommandOnRunningDbIfValid args = Command.Db.runCommandOnRunningDb <$> parsedCommand
  ...
```
That solves all the problems I think, but is too weird for my taste.

About the parser suggestion - also too complicated IMO. Thanks for sharing it though, the mechanics are interesting.

> Another way is to go back to what we had before and just add comments, maybe that is best.

Yes, that's my favorite option now. I was opposed to it at first because I really wanted the express constraints through code rather than through comments, but in this case the complexity isn't worth it.

@Martinsos Let me know if you by any chance prefer the `runCommandOnRunningDbIfValid` approach. If not, I'm sticking with the comment.

Martinsos: Yeah, not worth it. What you presented now is better, but let's say I want to add a third class of db commands or something here, then this printUsage is again stuck under comandOnDb case, it is a bit weird, it is still not clear this is actually the third case next to normal command and db running command.
Which is why I liked a bit better the code I proposed because it is more explicit in the sense of "it is a normal command, or command on running db, or I will print usage", and it allows for an modification of that on the high level (like inserting a new case) or adding commands to each of the groups.

But yeah, if you think that is too complex, let's do the comments and we are good. Plus maybe longer name for runDbCommand, that is good change.

Martinsos (approved):

PR #2178, opened 2024-07-12
Uses CSS module instead of Stitches for Loader component

infomiho:
`Loader` shouldn't depend on `stitches` being available since it's installed only when auth is used. 

Moves `Loader` to use CSS Modules instead.


waspc/data/Generator/templates/react-app/src/components/Loader.module.css
```diff
@@ -0,0 +1,37 @@
+.loader {
+  color: var(--loader-color, #1a202c);
```
infomiho, line 2: Wrote it like this to enable users to adjust the loader color with CSS vars. It's really documented but we can tell users that if they ask.

Martinsos: So this is still local CSS, is it? Not global, right? Because of the ".module.css"? WHo is enabling this, Vite, or are we using something special?

waspc/data/Generator/templates/react-app/src/components/Loader.tsx
```diff
@@ -1,51 +1,12 @@
...
-const fullRotationKeyframes = keyframes({
-  '0%': { transform: 'rotate(0deg)' },
-  '100%': { transform: 'rotate(360deg)' },
-})
-
-// DefaultLoader is a React component that spans accross the whole screen and displays
-// a spinner in the very middle.
+// Spans a accross the whole screen and displays a spinner in the very middle.
 export function Loader() {
+  const loaderClassName = ['loader', styles.loader].join(' ')
```
Martinsos, line 5: I would maybe just inline this, but do as you like.

Martinsos: Btw why do you also need to set 'loader' class here? I would imagine only `styles.loader` is enough?

Martinsos (approved):
Left a couple of comments but approved! Btw i believe i left somewhere in the codebase a comment/TODO saying that we will have this issue with stitches and loader - can you please check it and remove it if you find it? If not then it was probably just in that post-PR review I did.

waspc/data/Generator/templates/react-app/src/components/Loader.module.css
infomiho: Vite supports CSS modules out of the box: https://vitejs.dev/config/shared-options.html#css-modules

infomiho:
@Martinsos the comment about Stitches and loader is in your post-PR review, not in an issue, so there was nothing to delete.

waspc/data/Generator/templates/react-app/src/components/Loader.tsx
Martinsos: In that case, shouldn't we name it sometihng like `wasp-loader`? Because this is now global CSS class if I am right -> calling it just `loader` seems problematic.

infomiho: Most people use Tailwind these days or one of the CSS-in-JS options, so I don't think there will be a lot of interference with built in classes. If it happens to happen, let's change it then.

Martinsos: So you are saying that our global CSS hopefully wont cause any issues because others might not use global CSS themselves? That is ok, but we can't count on that. Also, as a library/framework, we should at least try to not affect their global space if we can. I think we should do this now, it is an easy change and is a good practice and we know it has a potential to cause problems down the road: it would be downright sloppy to leave this as it is now.

infomiho: Sure, I can add the `wasp-` prefix in a new PR.

PR #2179, opened 2024-07-12
Adds support for view and type blocks

infomiho:
Adds support for `view` and `type` blocks by reusing the `model` parser.

### DIFF
--- a/waspc/src/Wasp/Psl/Ast/Type.hs
+++ b/waspc/src/Wasp/Psl/Ast/Type.hs
@@ -1,0 +1,13 @@
+module Wasp.Psl.Ast.Type
+  ( Type (..),
+  )
+where
+
+import Wasp.Psl.Ast.Common (Name)
+import Wasp.Psl.Ast.Model (Body)
+
+data Type
+  = Type
+      Name
+      Body
+  deriving (Show, Eq)
--- a/waspc/src/Wasp/Psl/Parser/Model.hs
+++ b/waspc/src/Wasp/Psl/Parser/Model.hs
@@ -1,6 +1,8 @@
 module Wasp.Psl.Parser.Model
   ( parseBody,
     model,
+    -- Used by the `view` and `type` parsers.
+    body,
   )
 where
 
--- a/waspc/src/Wasp/Psl/Parser/Model.hs
+++ b/waspc/src/Wasp/Psl/Parser/Model.hs
@@ -1,7 +1,6 @@
 module Wasp.Psl.Parser.Model
   ( parseBody,
     model,
-    -- Used by the `view` and `type` parsers.
     body,
   )
 where
--- a/waspc/src/Wasp/Psl/Parser/Schema.hs
+++ b/waspc/src/Wasp/Psl/Parser/Schema.hs
@@ -17,6 +17,8 @@ import Wasp.Psl.Parser.Common (SourceCode, whiteSpace)
 import Wasp.Psl.Parser.ConfigBlock (configBlock)
 import Wasp.Psl.Parser.Enum (enum)
 import Wasp.Psl.Parser.Model (model)
+import Wasp.Psl.Parser.Type (typeBlock)
+import Wasp.Psl.Parser.View (view)
 
 parsePrismaSchema :: SourceCode -> Either Parsec.ParseError Psl.Schema.Schema
 parsePrismaSchema = Parsec.parse schema ""
@@ -32,6 +34,8 @@ schema = do
     many $
       choice
         [ Psl.Schema.ModelBlock <$> model,
+          Psl.Schema.ViewBlock <$> view,
+          Psl.Schema.TypeBlock <$> typeBlock,
           Psl.Schema.EnumBlock <$> enum,
           Psl.Schema.ConfigBlock <$> configBlock
         ]
--- a/waspc/data/Generator/templates/db/schema.prisma
+++ b/waspc/data/Generator/templates/db/schema.prisma
@@ -14,3 +14,11 @@
 {=& . =}
 
 {=/ enumSchemas =}
+{=# viewSchemas =}
+{=& . =}
+
+{=/ viewSchemas =}
+{=# typeSchemas =}
+{=& . =}
+
+{=/ typeSchemas =}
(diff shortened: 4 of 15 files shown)


waspc/src/Wasp/Psl/Ast/Model.hs
```diff
@@ -1,64 +1,21 @@
-{-# LANGUAGE DeriveDataTypeable #-}
-
 module Wasp.Psl.Ast.Model
   ( Model (..),
-    Body (..),
```
Martinsos, line 5: I know you moved these to `Model.Common`, but now you have a lot of imports that have to reference `Model.Common` instead of `Model` which is a bit weird. What you could do instead is re-export this stuff from here again! So it behaves same like before, but it will now leave in Common.

waspc/src/Wasp/Psl/Ast/Model.hs
```diff
@@ -1,64 +1,21 @@
...
-  | Unsupported String
-  | UserType String
-  deriving (Show, Eq, Data)
-
-data FieldTypeModifier
-  = List
-  | Optional
-  deriving (Show, Eq, Data)
-
 getFields :: Model -> [Field]
```
Martinsos, line 20: Doesn't this function operate on Body really? It is not specific to model in any way, from what I see in its implementation. So why not move it to also to Common with the rest of the stuff?

waspc/src/Wasp/Psl/Ast/Model/Common.hs
Martinsos, line 1: So you put it under Model.Common, which sounds to me like it is some stuff that is needed by the rest of the Model but not big enough to stand alone so you put it here. And that would be ok. But that is not it -> this is instead functionality that is shared by Model, Type and View. So why would it be specifically in Model.Common, under Model?

So it needs to be taken out, to higher level. It could be `Body/` and all the stuff under it. Or maybe better would be to introduce a new name for this. You said `ModelLike` before, that was going in the right direction. You could have such module, `ModelLike`. Or maybe a better name if we can find one. 

But wait, isn't enum much simpler? It is just `enum Role { USER ADMIN }` (with newlines). So we are parsing much too much stuff for it now. It should be completely standalone and well it is super simple.

And `type` -> is that "composite type"? If I got it correctly from their docs, those currently work only for MongoDB. ANd we don't support MongoDB yet. Meaning we don't have to implement these at all for now.

So what I would do is revert most of this PR, therefore dropping `type` and `Common`, and implement just `enum` with pristine parser for it which will be super simple, and that is it.

Martinsos (changes requested):

waspc/src/Wasp/Psl/Ast/Model/Common.hs
infomiho: `ModelLike` sounds okay to me, I couldn't think of anything better to say "stuff that's like a model".

We are not dealing with enums in this PR at all, `enum` has already its own custom parser.

We are dealing with 2 new block types `view` and `type` that are 100% the same as `model` minus the keyword. So, I'll move things around a bit and we'll see how it look :)

infomiho:
@Martinsos Now I've implemented the `view` and `type` parser based on importing the `body` parser from `Wasp.Psl.Ast.Model` module. Looks cleaner and it's clear that the `view` and `type` blocks are "model-like" blocks.

waspc/src/Wasp/Psl/Parser/Schema.hs
```diff
@@ -32,6 +34,8 @@ schema = do
     many $
       choice
         [ Psl.Schema.ModelBlock <$> model,
+          Psl.Schema.ViewBlock <$> view,
+          Psl.Schema.TypeBlock <$> typeBlock,
```
Martinsos, line 38: Why is `typeBlock` only parser that has `Block` suffix here?

waspc/src/Wasp/Psl/Parser/Model.hs
```diff
@@ -1,6 +1,8 @@
 module Wasp.Psl.Parser.Model
   ( parseBody,
     model,
+    -- Used by the `view` and `type` parsers.
```
Martinsos, line 4: I would remove this maybe -> it will invetiable become outdated at some point, and if somebody wants to check for usage, they can use their editor.

waspc/src/Wasp/Psl/Ast/Type.hs
```diff
@@ -0,0 +1,13 @@
...
+  )
+where
+
+import Wasp.Psl.Ast.Common (Name)
+import Wasp.Psl.Ast.Model (Body)
+
+data Type
+  = Type
+      Name
+      Body
```
Martinsos, line 12: So does type actually support anything that can go into Model body, or is it a subset? That is relevant both here for AST and also for parser. If it is indeed anything that goes into model, then great. If not (and I would guess so, at least for view), then I would consider putting a comment here explaining that a bit.
Same goes for `view`.

Martinsos (approved):
Nice! Approved but also left a couple of smaller comments.

waspc/src/Wasp/Psl/Parser/Schema.hs
infomiho: I wouldn't define a function with a name `type` since it's a reserved Haskell keyword.

waspc/src/Wasp/Psl/Parser/Model.hs
infomiho: Ok, good point 👍

waspc/src/Wasp/Psl/Ast/Type.hs
infomiho: > So does type actually support anything that can go into Model body, or is it a subset? 

I'm only looking at this from the point of view of Prisma grammar and `type` and `view` are separate blocks. 

You can however, use composite types (`type` block) in models by using them as field types e.g.:
```prisma

model MyModel {
  id String @id
  someType SomeType
}

type SomeType {
  name String
  age Int
}
```

infomiho: Added some comments to both view and type.

Martinsos: i meant in the sense that their ast/parser is subset of model's

PR #2182, opened 2024-07-15
Add tsconfig migration instructions

sodic:
This change was actually introduced in 0.13.0, but we forgot to put it in that migration guide.

Is it ok if we put it here?


sodic:
I'm merging this.

If you don't like something, comment and I'll make a new PR.

PR #2192, opened 2024-07-17
Improve wrong db provider error message

sodic:
Prevents this from happening:

image


Martinsos (approved):

infomiho:
This points out to a code smell, the `postgresql` and `sqlite` should be some sort of constants we use in multiple places :D Mea culpa

Issue #2203, opened 2024-07-23
Allow users to configure the websocket client


infomiho:
Add support for users to specify the options for the init of the socket.io client. Something like `webSocket.clientConfigFn`.

Issue #1958, opened 2024-04-14
Improve type signatures for parts of our public API


sodic:
While implementing #1930, I ran across several symbols that don't yet have proper types or were more difficult to explicitly type. These include:
- `MyCrud` from `wasp/client/crud` - A Crud object. It should be possible to explicitly type, albeit a bit tedious.
- `myJob` from `wasp/server/jobs` - A job function. Typing it will require a little code shuffling.
- `myQuery` and `myAction` from `wasp/server/operations` - server-side wrappers for user-defined Queries and Actions. They're types are pretty lacking (almost all arguments are `any`).

FranjoMindek:
`MyCrud` being done by https://github.com/wasp-lang/wasp/pull/4049

PR #2204, opened 2024-07-23
Use `websocket` transport for the websocket client

### LINKED ISSUE #2040: Websocket throws "Unknown Session ID" when trying to connect
I'm not sure what could be the cause of the error: some re-render behaviour, our auth session logic or something misconfiguration of Socket.io, but this is how to looks like:

https://github.com/wasp-lang/wasp/assets/2223680/dec925d4-c3a6-46fa-9033-214c981d574e

It tries to connect multiple times, sometimes it succeeds, sometime it fails, but in the end it succeeds.

This happened after I deployed our `waspc/examples/todoApp` to Fly and opened the client.

infomiho:
Closes #2040

### Background

When users deploy a Wasp app with websockets and their deployment provider (e.g. Fly) uses a load balancer, there are some "socket.io session ID" errors. This happens because socket.io client is trying to first establish an HTTP connection to the server. It sends multiple HTTP requests and for each the "socket.io session ID" needs to match. This becomes a problem when the different HTTP requests go to different servers.

The docs suggest either:
- using **sticky load balancing** (for client X, only use server A and not servers B or C)
- or to switch to using only the **websocket transport** (eliminating the multiple HTTP requests and just having one TCP connection). 

More details: https://github.com/wasp-lang/wasp/issues/2040#issuecomment-2244663846

### Changes in this PR

Changes the connection mode for the socket.io client from the default `["polling", "websocket"]` to `["websocket"]` because `polling` option uses multiple HTTP requests to establish the Websocket connection.
### Future work

Of course, this might not be ideal for some people - that's why we have this issue: #2203 which would allow users to override the connection mode to whatever they want.

### DIFF
--- a/waspc/e2e-test/test-outputs/waspComplexTest-golden/waspComplexTest/.wasp/out/sdk/wasp/dist/server/jobs/mySpecialJob.d.ts
+++ b/waspc/e2e-test/test-outputs/waspComplexTest-golden/waspComplexTest/.wasp/out/sdk/wasp/dist/server/jobs/mySpecialJob.d.ts
@@ -16,7 +16,7 @@ export declare const mySpecialJob: {
         readonly pgBoss: {
             readonly cancel: () => ReturnType<import("pg-boss")["cancel"]>;
             readonly resume: () => ReturnType<import("pg-boss")["resume"]>;
-            readonly details: () => Promise<(Omit<import("pg-boss").JobWithMetadata<JSONObject>, "output" | "state"> & {
+            readonly details: () => Promise<Omit<import("pg-boss").JobWithMetadata<JSONObject>, "output" | "state"> & {
                 data: JSONObject;
             } & ({
                 state: "failed";
@@ -35,7 +35,7 @@ export declare const mySpecialJob: {
                 } | {
                     value: true;
                 };
-            })) | null>;
+            })>;
         };
         readonly job: import("./core/job").Job;
         readonly jobId: string;
--- a/waspc/data/Generator/templates/sdk/wasp/client/webSocket/WebSocketProvider.tsx
+++ b/waspc/data/Generator/templates/sdk/wasp/client/webSocket/WebSocketProvider.tsx
@@ -16,7 +16,13 @@ export type WebSocketContextValue = {
 // PRIVATE API
 // TODO: In the future, it would be nice if users could pass more
 // options to `io`, likely via some `configFn`.
-export const socket: Socket<ServerToClientEvents, ClientToServerEvents> = io(config.apiUrl, { autoConnect: {= autoConnect =} })
+export const socket: Socket<ServerToClientEvents, ClientToServerEvents> = io(
+  config.apiUrl,
+  {
+    transports: ['websocket'],
+    autoConnect: {= autoConnect =},
+  }
+)
 
 function refreshAuthToken() {
   // NOTE: When we figure out how `auth: true` works for Operations, we should
--- a/waspc/e2e-test/test-outputs/waspComplexTest-golden/waspComplexTest/.wasp/out/sdk/wasp/dist/server/jobs/returnHelloJob.d.ts
+++ b/waspc/e2e-test/test-outputs/waspComplexTest-golden/waspComplexTest/.wasp/out/sdk/wasp/dist/server/jobs/returnHelloJob.d.ts
@@ -20,7 +20,7 @@ export declare const returnHelloJob: {
         readonly pgBoss: {
             readonly cancel: () => ReturnType<import("pg-boss")["cancel"]>;
             readonly resume: () => ReturnType<import("pg-boss")["resume"]>;
-            readonly details: () => Promise<(Omit<import("pg-boss").JobWithMetadata<JSONObject>, "output" | "state"> & {
+            readonly details: () => Promise<Omit<import("pg-boss").JobWithMetadata<JSONObject>, "output" | "state"> & {
                 data: JSONObject;
             } & ({
                 state: "failed";
@@ -39,7 +39,7 @@ export declare const returnHelloJob: {
                 } | {
                     value: true;
                 };
-            })) | null>;
+            })>;
         };
         readonly job: import("./core/job").Job;
         readonly jobId: string;
--- a/waspc/e2e-test/test-outputs/waspJob-golden/waspJob/.wasp/out/sdk/wasp/dist/server/jobs/mySpecialJob.d.ts
+++ b/waspc/e2e-test/test-outputs/waspJob-golden/waspJob/.wasp/out/sdk/wasp/dist/server/jobs/mySpecialJob.d.ts
@@ -16,7 +16,7 @@ export declare const mySpecialJob: {
         readonly pgBoss: {
             readonly cancel: () => ReturnType<import("pg-boss")["cancel"]>;
             readonly resume: () => ReturnType<import("pg-boss")["resume"]>;
-            readonly details: () => Promise<(Omit<import("pg-boss").JobWithMetadata<JSONObject>, "output" | "state"> & {
+            readonly details: () => Promise<Omit<import("pg-boss").JobWithMetadata<JSONObject>, "output" | "state"> & {
                 data: JSONObject;
             } & ({
                 state: "failed";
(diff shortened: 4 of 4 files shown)


waspc/e2e-test/test-outputs/waspComplexTest-golden/waspComplexTest/.wasp/out/sdk/wasp/dist/server/jobs/mySpecialJob.d.ts
```diff
@@ -16,7 +16,7 @@ export declare const mySpecialJob: {
         readonly pgBoss: {
             readonly cancel: () => ReturnType<import("pg-boss")["cancel"]>;
             readonly resume: () => ReturnType<import("pg-boss")["resume"]>;
-            readonly details: () => Promise<(Omit<import("pg-boss").JobWithMetadata<JSONObject>, "output" | "state"> & {
+            readonly details: () => Promise<Omit<import("pg-boss").JobWithMetadata<JSONObject>, "output" | "state"> & {
```
infomiho, line 19: I'm not sure why these e2e test changes happened 🤷

infomiho: Probably due to some Typescript emit fixes: https://github.com/microsoft/TypeScript/releases

sodic (approved):
Looks good. If you made sure the error is fixed (i.e., tested it on fly), merge when ready.

waspc/e2e-test/test-outputs/waspComplexTest-golden/waspComplexTest/.wasp/out/sdk/wasp/dist/server/jobs/mySpecialJob.d.ts
sodic: What do you mean by "emit fixes?"

Does this change our public API?

infomiho: Typescript keeps refining and fixing what they emit in `d.ts` files from version to version. Since we type our public API explicitly, I don't think it affects our public API since it mostly relates to the types Typescript infers (99% sure).

sodic: Jobs are not yet explicitly typed (unfortunately, see https://github.com/wasp-lang/wasp/issues/1958), but I can't tell whether this affects the public API.

So, if you're relying on us explicitly typing the API, please double check.

sodic (commented):
Ok, I'm slowly changing my mind.

This is a behavior change that hopes to fix a specific production setup. It feels backward (reminds me of [this](https://github.com/wasp-lang/wasp/issues/1958)).

Anyway, we shouldn't change the application code to fit a certain infrastructure. It's the infrastructure's job to adapt to the app.

So, the question is: _Is `transports: ['websockets']` the right default, regardless of what load balancers do?_

If you think so, we can change it. Keep in mind that socket.io doesn't think so, and there's value in keeping us in line with their defaults.

If not, then the better option is leaving the default as is and letting users fix their deployments through the config function you implemented in #2205.

waspc/e2e-test/test-outputs/waspComplexTest-golden/waspComplexTest/.wasp/out/sdk/wasp/dist/server/jobs/mySpecialJob.d.ts
infomiho: In this change we are relying on the `pg-boss` library types, that's not something we can type explicitly IMHO. So, it will change from time to time in the `d.ts` files. Or should we copy their types and use them directly instead of importing?

sodic: So it does change the public API?

To avoid confusion, I don't know whether that's the case, I was just speculating.

infomiho:
> So, the question is: Is transports: ['websockets'] the right default, regardless of what load balancers do?

> If you think so, we can change it. Keep in mind that socket.io doesn't think so, and there's value in keeping us in line with their defaults.

I'll just quote the docs on this:
> the WebSocket transport does not have this limitation, since it relies on a single TCP connection for the whole session. Which means that if you disable the HTTP long-polling transport (which is a perfectly valid choice in 2021), you won't need sticky sessions

HTTP long-polling was a way to be backwards compatible with clients that don't support websockets. Last browser that didn't support it was released 10+ years ago if I'm reading this correctly: https://caniuse.com/?search=websocket 

So, in my opinion, if we set `transports: ['websockets']` as our default, we are making an opinionated choice that will give our users an easier setup and faster connection times (since it goes directly for the websockets instead of first doing the HTTP polling dance).

Again, most of our users now use Fly and probably will use modern deployment options that _most of the time_ come with load balancing. This means that they will need to customise the options _most of the time_, so why not give them better defaults? :D

waspc/e2e-test/test-outputs/waspComplexTest-golden/waspComplexTest/.wasp/out/sdk/wasp/dist/server/jobs/mySpecialJob.d.ts
infomiho: Ok, now that I look at the actual change: it went from `Promise<T | null>` to `Promise<T>` which means PgBoss types couldn't affected this. This means it's pure TS compiler related change. It looks like they were fixing a bug between 5.5.3 and 5.5.4.

```
npm install typescript@5.5.3
npx tsc --declaration --emitDeclarationOnly --stripInternal --declarationDir dist
```
diffed with
```
npm install typescript@5.5.4
npx tsc --declaration --emitDeclarationOnly --stripInternal --declarationDir dist
```
give me the diff we have here.

sodic:
@infomiho 

> So, in my opinion, if we set transports: ['websockets'] as our default, we are making an opinionated choice that will give our users an easier setup and faster connection times (since it goes directly for the websockets instead of first doing the HTTP polling dance).

Ok, makes sense.

One final question. Why not use `transports: ['websocket', 'polling']` which falls back to polling if websockets aren't available (suggested [here](https://socket.io/docs/v4/using-multiple-nodes/#why-is-sticky-session-required))?

Or even add a block that falls back on error (suggested [here](https://socket.io/docs/v4/client-options/#transports)):
```javascript
const socket = io("https://example.com", {
  transports: ["websocket", "polling"] // use WebSocket first, if available
});

socket.on("connect_error", () => {
  // revert to classic upgrade
  socket.io.opts.transports = ["polling", "websocket"];
});
```
Is it just to avoid complexity or are there other implications?

infomiho:
@sodic 

> One final question. Why not use transports: ['websocket', 'polling'] which falls back to polling if websockets aren't available (suggested [here](https://socket.io/docs/v4/using-multiple-nodes/#why-is-sticky-session-required))?

If _most of our users_ use load balancers by default, the fallback method will fail for them. If the websocket transport failed, the server is probably down and `polling` will then also fail. This way we keep the code simpler and if some minority of our users really want to have this fallback, that's something they can configure themselves.

I kept going back and forth with this comment because, I agree with your - it's nice having a fallback - but in this case it feels like it won't do much in most cases, so why even keep it since it causes cryptic session ID error messages when it fails.

waspc/e2e-test/test-outputs/waspComplexTest-golden/waspComplexTest/.wasp/out/sdk/wasp/dist/server/jobs/mySpecialJob.d.ts
sodic: I still don't know whether it changes our public API :smile: 

Does it?

infomiho: Yes, our public API changes between various TS versions due to their regressions and fixes. To avoid this, we should pin our TS version we use to build the SDK. But this is outside the scope of this PR.

PR #2205, opened 2024-07-23
Adds client config fn for websockets

### LINKED ISSUE #2203: Allow users to configure the websocket client

infomiho:
Allow users to specify all `io` client options. 

Closes #2203 

### Left to do

- [ ] Docs


waspc/src/Wasp/Generator/Common.hs
```diff
@@ -77,6 +78,10 @@ makeJsArrayFromHaskellList list = "[" ++ intercalate ", " listOfJsStrings ++ "]"
   where
     listOfJsStrings = map (\s -> "'" ++ s ++ "'") list
 
+makeJsBoolFromHaskellBool :: Bool -> String
```
infomiho, line 81: I felt this was cleaner than using `show` + `toLower`.

waspc/examples/todoApp/main.wasp
```diff
@@ -5,7 +5,8 @@ app todoApp {
   title: "ToDo App",
   // head: [],
   webSocket: {
-    fn: import { webSocketFn } from "@src/webSocket",
+    fn: import { webSocketFn } from "@src/webSocket/index.js",
+    clientConfigFn: import { getConfig } from "@src/webSocket/clientConfigFn.js",
```
sodic, line 9: You'll probably want to stick with this (since we already have a `webSocket` field), but just FYI.

Another possible approach is exposing a JS function that users could call in their client setup, like we do with `configureQueryClient`: https://wasp-lang.dev/docs/project/client-config#overriding-default-behaviour-for-queries.

Both options are both consistent and inconsistent with other parts of the API (it depends on what you compare them with).

Finally, does it make sense to take the middle ground?
For example, we expose a `transports` field in the Wasp file while also exposing a `configureWebSocket` hook users could use in the client setup function. This would make things easier for those who just want to change the `transport` option.

waspc/data/Generator/templates/sdk/wasp/client/webSocket/index.ts
```diff
@@ -5,6 +5,9 @@ import type {
   ServerToClientEvents,
 } from 'wasp/server/webSocket'
 
+// PUBLIC API
+export type { WebSocketClientConfigFn } from './WebSocketProvider'
```
sodic, line 9: Don't forget to update the import table after merge (I won't share the URL here since it's public).

sodic (commented):
Left some comments, only one major question (assuming we want to go with this).

Still, my biggest question is "Are we sure we want to go with this?"
I'm more on the side of "Let's wait for someone to request the feature before we include it."

If you strongly feel that people will need this and it's as clear as day to you that we should have it (or if they've been asking for something similar), then go for it,  If not, I'd wait.

waspc/examples/todoApp/main.wasp
infomiho: I kinda dislike what I have now tbh, it feels weird to have it defined like that for some reason. 

The `configureQueryClient` approach might make more sense since it's a lower-level configuration than some other configuration e.g. `setupClient`. I'm not sure what to go for here.

sodic: What about not doing it yet (see here: https://github.com/wasp-lang/wasp/pull/2205#pullrequestreview-2212096740)?

waspc/src/Wasp/Generator/SdkGenerator/WebSocketGenerator.hs
```diff
@@ -50,7 +49,12 @@ genWebSocketProvider spec = return $ C.mkTmplFdWithData [relfile|client/webSocke
   where
     maybeWebSocket = AS.App.webSocket $ snd $ getApp spec
     shouldAutoConnect = (AS.App.WS.autoConnect <$> maybeWebSocket) /= Just (Just False)
-    tmplData = object ["autoConnect" .= map toLower (show shouldAutoConnect)]
+    maybeClientConfigFn = extImportToJsImport <$> (maybeWebSocket >>= AS.App.WS.clientConfigFn)
+    tmplData =
+      object
+        [ "autoConnect" .= makeJsBoolFromHaskellBool shouldAutoConnect,
```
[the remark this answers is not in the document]
infomiho: Fair enough 👍 I like the reasoning behind it. Haskell declares what it wants and it's up to the template to make it happen.

waspc/examples/todoApp/main.wasp
infomiho: Now I'm no longer sure. This PR came out of the feeling: _We are setting the transports to an opinionated default, we should allow customisation_ rather than coming as a request from our users. Maybe the best thing for now is to close the PR and see if users really want this. I guess by then we'll know more if we want to design it like `configureQueryClient` or as the client setup fn.

sodic: > Maybe the best thing for now is to close the PR and see if users really want this.

I advise closing it, yes.

Another option is including this as a "shadow feature" (I made up the term, but it probably exists). We don't document the feature, we just put it in. Then, if someone requests it, we can "deploy" it right away. I'd say this makes sense for features that:
- Can have a high impact.
- We are certain someone will need.

But I don't think this one falls under those categories.

The developers (Wasp users) probably won't ever notice a problem. Their users might (if they're using ancient browsers), and will report it to the developers (who will probably tell them to update their browser, as they should anyway) :smile:

So yeah, I'd close it.

infomiho: Okay, let's keep it simple and close it. The impl is not that hard anyways, so when we see the need for this, we'll be able to quickly add it.

infomiho:
Closing this PR in the name of simplicity and not adding a new feature our users might never need. Details: https://github.com/wasp-lang/wasp/pull/2205#discussion_r1701784727

PR #2210, opened 2024-07-29
Replace `oslo/password` with directly using `@node-rs/argon2`

### LINKED ISSUE #2136: Use `@node-rs/argon2` directly instead of `oslo/password`
Based on this [tweet](https://x.com/pilcrowonpaper/status/1806539558290686237) from Lucia's author and official [Lucia docs](https://lucia-auth.com/guides/email-and-password/basics) we should move away from `oslo/password` to directly using `@node-rs/argon2`.

We have been using that lib anyways since we used `Argon2id` from `oslo/password` which [uses](https://github.com/pilcrowOnPaper/oslo/blob/main/src/password/argon2id.ts) `@node-rs/argon2`.

infomiho:
Closes #2136

This is a drop-in replacement with the same options as `oslo/password` used under the hood. Motivation is described in the linked issue.


waspc/data/Generator/templates/sdk/wasp/auth/password.ts
```diff
@@ -1,16 +1,36 @@
...
+  outputLen: 32,
+  parallelism: 1,
+  version: Version.V0x13,
+};
 
 // PRIVATE API
-export const hashPassword = async (password: string): Promise<string> => {
-  return argon2id.hash(password)
+export async function hashPassword(password: string): Promise<string> {
+  return hash(normalizePassword(password), hashingOptions);
```
sodic, line 14: How do we know this is equivalent to what we had?

I know it follows the guide and you most likely tested it, but is it possible to unit test it somehow? I know our current setup isn't test-friendly, but maybe we can think of something (e.g., building an app and testing it there, like we did for operation type tests).

If it's too much work and nothing comes to mind, we can postpone until the testing effort which I hope is happening soon.

waspc/data/Generator/templates/sdk/wasp/auth/password.ts
```diff
@@ -1,16 +1,36 @@
...
 }
 
 // PRIVATE API
-export const verifyPassword = async (hashedPassword: string, password: string): Promise<void> => {
-  const isValidPassword = await argon2id.verify(hashedPassword, password)
-  if (!isValidPassword) {
-    throw new Error('Invalid password.')
+export async function verifyPassword(
```
sodic, line 18: Don't forget to update the visible API table if applicable.

waspc/src/Wasp/Generator/ServerGenerator/AuthG.hs
```diff
@@ -127,5 +127,7 @@ depsRequiredByAuth spec = maybe [] (const authDeps) maybeAuth
       AS.Dependency.fromList
         [ ("lucia", "^3.0.1"),
           ("@lucia-auth/adapter-prisma", "^4.0.0"),
-          ("oslo", "^1.1.2")
+          ("oslo", "^1.1.2"),
+          -- Argon2 is used for hashing passwords.
+          ("@node-rs/argon2", "^1.8.3")
```
sodic, line 132: You're importing the dependency into the SDK, but listing the dependency under the server's package.json.

sodic (changes requested):
I think this is mistake: 

After fixing that, we're good to go. Tests would be nice, though (if we can figure something out).

waspc/data/Generator/templates/sdk/wasp/auth/password.ts
infomiho: Everything that we consider public API can be imported from userland and we could create a test app that unit tests all the APIs. Basically, write `.spec.ts` files for each piece of our public API and test it if possible. 

But this seems like a larger effort than these two functions and I'd do it when we do everything else.

waspc/src/Wasp/Generator/ServerGenerator/AuthG.hs
infomiho: Great catch, let me update it 👍

waspc/data/Generator/templates/sdk/wasp/auth/password.ts
infomiho: > How do we know this is equivalent to what we had?

I've tested it with an existing app (password hashed with oslo but checked by argon2 lib) and with a new app (password hashed and checkd by argon2 lib) and it worked. 

I've used what the oslo lib used internally, we are using the same code.

waspc/src/Wasp/Generator/ServerGenerator/AuthG.hs
infomiho: It's installed in both places because of this line: https://vscode.dev/github/wasp-lang/wasp/blob/miho-replace-oslo-password/waspc/src/Wasp/Generator/SdkGenerator.hs#L204

But, it's not imported in the `server` directly, so that's redundant. I'll move it to SDK.

waspc/data/Generator/templates/sdk/wasp/auth/password.ts
sodic: > But this seems like a larger effort than these two functions and I'd do it when we do everything else.

Makes sense! Do you mind creating an issue?

infomiho: Here it is, please add any additional thoughts you might have on the topic: https://github.com/wasp-lang/wasp/issues/2221

sodic (approved):

PR #2211, opened 2024-07-30
Implement login hooks

### LINKED ISSUE #2117: Login hooks
We want to add `onBeforeLogin` and `onAfterLogin` hooks analogously to our signup hooks. 

We would be able to reject logins based on some criteria in the `onBeforeLogin` hook.

Also, we would be able to get the access token for OAuth logins in the `onAfterLogin` hook. Also, we would get the `uniqueRequestId` in the `onAfterLogin` (same as we had for `onAfterSignup`).

infomiho:
Closes #2117 

### What did we do here

- Added two new auth hooks: `onBeforeLogin` and `onAfterLogin`

### Left to do

- [x] Update e2e tests
- [x] Add a headless test
- [x] Docs


waspc/examples/todoApp/src/auth/hooks.ts
```diff
@@ -47,6 +50,25 @@ export const onBeforeOAuthRedirect: OnBeforeOAuthRedirectHook = async (
...
 }
 
+export const onBeforeLogin: OnBeforeLoginHook = async (args) => {
+  const log = createLoggerForHook('onBeforeLogin')
+  log('providerId object', args.providerId)
+
+  // const now = new Date()
+  // if (now.getHours() < 14) {
+  //   throw new HttpError(403, 'Login is only possible after 2 PM')
+  // }
```
sodic, line 60: Should we remove this?

waspc/data/Generator/templates/server/src/auth/providers/oauth/user.ts
```diff
@@ -78,12 +84,36 @@ async function getAuthIdFromProviderDetails({
...
-    return existingAuthIdentity.{= authFieldOnAuthIdentityEntityName =}.id
+    const authId = existingAuthIdentity.{= authFieldOnAuthIdentityEntityName =}.id
+
+    // NOTE: We are calling login hooks here even though we didn't log in the user yet.
+    // We are doing it here because we have access to the OAuth tokens and we can pass them to the hooks.
+    // This isn't a big deal because the next step of the OAuth flow happens immediately after this function
+    // and the user is redirected to the client with the one-time code which is then used to create the session.
+    // The downside of this approach is that we can't provide the session to the login hooks, but this is
+    // an okay trade-off for now.
+    await onBeforeLoginHook({ req, providerId })
```
sodic, line 95: I'm not sure I understand this.

So, both hooks are called before the login. Shouldn't this comment then be above the `onAfterLooginHook` call (since this is the hook that's called when it's supposed to be called, while that one isn't)?

If that's the case, can you explain (in the comment), why having access to the OAuth token is more important than calling the hook at the right time.

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
```diff
@@ -21,6 +21,16 @@ export type OnBeforeOAuthRedirectHook = (
   params: Expand<OnBeforeOAuthRedirectHookParams>,
 ) => { url: URL } | Promise<{ url: URL }>
 
+// PUBLIC API
+export type OnBeforeLoginHook = (
```
sodic, line 25: Make sure to update the visible API table. I know this is annoying, I hope we can get rid of it soon.

sodic (commented):
The implementation looks good. Very straightforward, you've built a nice system.

I do have some concerns I left in the comments.

Most important question though: Do we need this? Did someone ask for it?

waspc/examples/todoApp/src/auth/hooks.ts
infomiho: Sure, it serves no purpose 👍

waspc/data/Generator/templates/server/src/auth/providers/oauth/user.ts
infomiho: Both of these hooks are called differently than for email and username providers, so the comment still makes sense above `onBeforeLoginHook`. For email and username, we call `onBeforeLoginHook` right before the `createSession` call and the `onAfterLoginHook` right after that. 

For OAuth providers, we call them both right before we redirect to the client and then the session is created when the client exchanges the one time code with the server for the session. 

The OAuth tokens are _only available here_ and never later, and these tokens are very valuable for our users because they enable various API integrations e.g. using Google Sheets with their users' credentials. We could temporarily store the token and then make it available in the one time code <-> session exchange handler, but that felt needlessly complicated. The session ID is not that important and this a good trade off. 

I'll write these details in the comment above 👍

web/docs/auth/auth-hooks.md
```diff
@@ -567,3 +722,132 @@ The hook receives an object as **input** with the following properties:
...
+
+- `accessToken: string`
+
+  The OAuth access token that can be used to make requests to the provider's API on the user's behalf.
+
+- `uniqueRequestId: string`
+
+  The unique request ID for the OAuth flow (you might know it as the `state` parameter in OAuth.)
+
+  You can use the unique request ID to get the data saved in the `onBeforeOAuthRedirect` hook.
```
sodic, line 853: Unnecessary passive form (to get the data saved -> to save the data). I'd expect Grammarly to flag this

web/docs/auth/auth-hooks.md
```diff
@@ -567,3 +722,132 @@ The hook receives an object as **input** with the following properties:
...
+
+### OAuth fields
+
+The `oauth` object is passed to the `onAfterSignup` and `onAfterLogin` hooks only when the user is authenticated with [Social Auth](./social-auth/overview.md).
+
+It has the following fields:
+
+- `accessToken: string`
+
+  The OAuth access token that can be used to make requests to the provider's API on the user's behalf.
```
sodic, line 847: Passive form.

web/docs/auth/auth-hooks.md
```diff
@@ -30,6 +32,17 @@ If you are using OAuth, the flow includes extra steps before the signup flow:
   caption="OAuth Flow with Hooks"
```
sodic, line 32: I can't reach line 27, but we should update that one too probably (signup flow -> auth flow or something).

web/docs/auth/auth-hooks.md
```diff
@@ -30,6 +32,17 @@ If you are using OAuth, the flow includes extra steps before the signup flow:
...
 
+<ImgWithCaption
+  source="/img/auth-hooks/login_flow_with_hooks.png"
+  alt="Login Flow with Hooks"
+  caption="Login Flow with Hooks"
+/>
+
+<small>
+
+\* When using the OAuth auth providers, the login hooks are both called before the session is created but the session is created quickly afterward, so it shouldn't make any difference in practice.
```
sodic, line 43: This astrix isn't attached to anything.

Maybe put it on "Login Flow with Hooks*" (image caption).

web/docs/auth/auth-hooks.md
```diff
@@ -268,7 +285,7 @@ Read more about the data the `onAfterSignup` hook receives in the [API Reference
 
 Wasp calls the `onBeforeOAuthRedirect` hook after the OAuth redirect URL is generated but before redirecting the user. This hook can access the request object sent from the client at the start of the OAuth process.
 
-The `onBeforeOAuthRedirect` hook can be useful if you want to save some data (e.g. request query parameters) that can be used later in the OAuth flow. You can use the `uniqueRequestId` parameter to reference this data later in the `onAfterSignup` hook.
+The `onBeforeOAuthRedirect` hook can be useful if you want to save some data (e.g. request query parameters) that can be used later in the OAuth flow. You can use the `uniqueRequestId` parameter to reference this data later in the `onAfterSignup` or `onAfterLogin` hooks.
```
sodic, line 288: Passive form: can be used (unless I'm misunderstanding who uses it).

web/docs/auth/auth-hooks.md
sodic, line 43: Hm, this _really_ seems to suggest that the comment discussed here should be on the `onAfterLoginHook`: https://github.com/wasp-lang/wasp/pull/2211#discussion_r1699849062

`onBeforeLogin` truly is called before login (session creation), while `onAfterLogin` isn't called after login (session creation), or am I missing something?

web/docs/auth/auth-hooks.md
```diff
@@ -567,3 +722,132 @@ The hook receives an object as **input** with the following properties:
...
+
+  The provider's name (e.g. `'email'`, `'google'`, `'github`)
+
+- `providerUserId: string`
+
+  The user's unique ID in the provider's system (e.g. email, Google ID, GitHub ID)
+
+### OAuth fields
+
+The `oauth` object is passed to the `onAfterSignup` and `onAfterLogin` hooks only when the user is authenticated with [Social Auth](./social-auth/overview.md).
```
sodic, line 841: Passive form again, maybe "Wasp passes the..."

web/docs/auth/auth-hooks.md
```diff
@@ -567,3 +722,132 @@ The hook receives an object as **input** with the following properties:
...
+
+- `req: Request`
+
+  The [Express request object](https://expressjs.com/en/api.html#req) from which you can access the request headers, cookies, etc.
+
+Wasp ignores this hook's **return value**.
+
+### ProviderId fields
+
+The `providerId` object is passed to the `onBeforeSignup`, `onAfterSignup`, `onBeforeLogin`, and `onAfterLogin` hooks.
```
sodic, line 827: Also passive form, but debatable whether "Wasp passes" is better.

web/docs/auth/auth-hooks.md
```diff
@@ -567,3 +722,132 @@ The hook receives an object as **input** with the following properties:
...
+```
+</TabItem>
+</Tabs>
+
+The hook receives an object as **input** with the following properties:
+- [`providerId: ProviderId`](#providerid-fields)
+
+- `prisma: PrismaClient`
+
+  The Prisma client instance which you can use to query your database.
```
sodic, line 762: Should we somehow address the duplication for all these hook API reference bits?

I hate dealing with duplication in docs. It's never obvious to me what's the right thing to do.

web/docs/auth/auth-hooks.md
```diff
@@ -567,3 +722,132 @@ The hook receives an object as **input** with the following properties:
...
+```
+</TabItem>
+</Tabs>
+
+The hook receives an object as **input** with the following properties:
+- [`providerId: ProviderId`](#providerid-fields)
+
+- `user: User`
+
+  The user object of the user who logged in.
```
sodic, line 811: Perhaps "The logged in user's object?"

web/docs/auth/auth-hooks.md
infomiho: I've worried about that myself but didn't pull the trigger on extracting them.

Now I've extracted them into `Common hook input` and link from each hook.

web/docs/auth/auth-hooks.md
infomiho: ### Email and username providers

`onBeforeLogin` is called _right before_ session creation.

`onAfterLogin` is called _right after_ session creation.

### OAuth is a bit different

`onBeforeLogin` is called _way before_ session creation i.e. before redirecting to the client with OTC (one time code) which will then exchange the OTC for session (harder to say it's wrong)

`onAfterLogin` is called at the same time (easy to say it's wrong)

### Placement of the comment

That's why I put the comment above the `onBeforeLogin` because it also differs from the email and username cases, so it felt right.

sodic: Ohh, I see the misunderstanding. You wanted to cover both hooks with the comment above `onBeforeLogin`, and since that one comes before `onAfterLogin`, you put it there.

Ok, makes sense, but I wonder if it's easily missed. Do you think it makes sense to include a note above `onAfterLogin` too? Something simple like: `Check the comment above`.

sodic (approved):

web/docs/auth/auth-hooks.md
infomiho: Makes sense 👍  I'll add an extra comment above the `onAfterLogin`.

Issue #175, opened 2021-02-04
Implement good support for environment variables


Martinsos:
Right now there are some env vars that server requires: PORT, DATABASE_URL, and sometimes JWT_SECRET.
We should implement it so in the generated JS code that it fails early if they are missing -> that should probably be done in server/config.js

Also, it would be cool if Wasp understood which ENV vars are required and it would be on the lookout for them in some way -> for example, it would list them in the generated README.

Finally, we should figure out a way for the users to provide their own env vars, and also our env vars, in a nice way -> via .env files. I am not yet sure what is the best way to go about this, but we should probably allow them to define files like .env and .env.development and similar next to the .wasp files, and they would be picked up and appropriately used / placed in the generated code. We would also certainly need to separate them in regard to client vs server code (since client ones are ok being public and server ones are usually private), and also should be careful which ones are ok being committed into git and which ones are not.
Create React App has system of env files well defined, so we could look into how they do it.
Some npm packagea we will probably want to use are dotenv, maybe cmdenv.

Martinsos:
We added .env which contains env vars that are passed to the server. It is not version controled.

Martinsos:
Related to #1067

infomiho:
An example way of validating env vars:

```ts
import * as z from "zod";

const envSchema = z.object({
  NODE_ENV: z.enum(["development", "production"]).default("development"),
  PORT: z.coerce.number().default(3000),
  GITHUB_CLIENT_ID: z.string({
    required_error: "GITHUB_CLIENT_ID is required",
  }),
  GITHUB_CLIENT_SECRET: z.string({
    required_error: "GITHUB_CLIENT_SECRET is required",
  }),
  GOOGLE_CLIENT_ID: z.string({
    required_error: "GOOGLE_CLIENT_ID is required",
  }),
  GOOGLE_CLIENT_SECRET: z.string({
    required_error: "GOOGLE_CLIENT_SECRET is required",
  }),
  DISCORD_CLIENT_ID: z.string({
    required_error: "DISCORD_CLIENT_ID is required",
  }),
  DISCORD_CLIENT_SECRET: z.string({
    required_error: "DISCORD_CLIENT_SECRET is required",
  }),
  SERVER_URL: z.string({
    required_error: "SERVER_URL is required",
  }),
  CLIENT_URL: z.string({
    required_error: "CLIENT_URL is required",
  }),
  JWT_SECRET: z.string({
    required_error: "JWT_SECRET is required",
[...]

PR #2212, opened 2024-07-30
Refresh token impl

### LINKED ISSUE #2124: Implement proper support for OAuth refresh tokens
We are only giving the users the access token in `onAfterSignup` auth hook. We plan on introducing login related hooks in #2117 

We should also give users the refresh token alongside the access token. We should also then investigate how the refresh token will be used to get the new access token over time (https://arctic.js.org/guides/oauth2, scroll to Refresh access token).

This would most likely mean that the OAuth client instance would need to be available to the users (e.g. `google` client we use internally should become available via some SDK import).

infomiho:
Closes #2124 

### What was done
- Moved init of OAuth clients (google, github, discord and keycloak) to the SDK so users can also import them and use them
  - These clients are needed to refresh auth tokens
- Move some of the OAuth helpers to SDK - because they are only used in the SDK or could be useful for users
  - The env helpers are only used in the SDK
  - The redirect and one-time token helpers can be useful if users implement some OAuth client themselves
- Updated the OAuth tokens that we provide in auth hooks: we now provide all the data we get and not just the access token
  - We forward the data we receive so users will get the refresh token if we receive it

### Left to do
- [x] Update docs
- [x] Fix provider name types


web/docs/auth/auth-hooks.md
```diff
@@ -191,7 +197,7 @@ Wasp calls the `onAfterSignup` hook after the user is created.
 
 The `onAfterSignup` hook can be useful if you want to send the user a welcome email or perform some other action after the user signs up like syncing the user with a third-party service.
 
-Since the `onAfterSignup` hook receives the OAuth access token, it can also be used to store the OAuth access token for the user in your database.
+Since the `onAfterSignup` hook receives the OAuth tokens, it can also be used to store the OAuth access token and/or [refresh token](#refreshing-the-oauth-access-token) in your database.
```
sodic, line 199: Passive voice. Please fix in other places as well.

web/docs/auth/auth-hooks.md
```diff
@@ -220,9 +226,9 @@ export const onAfterSignup = async ({
   console.log('number of users after', count)
   console.log('user object', user)
 
-  // If this is an OAuth signup, we have the access token and uniqueRequestId
+  // If this is an OAuth signup, we have the OAuth tokens and uniqueRequestId
```
sodic, line 228: Address the user directly plus "the" before `uniqueRequestId`. Same applies below.

web/docs/auth/auth-hooks.md
```diff
@@ -433,7 +441,7 @@ Wasp calls the `onAfterLogin` hook after the user logs in.
 
 The `onAfterLogin` hook can be useful if you want to perform some action after the user logs in, like syncing the user with a third-party service.
 
-Since the `onAfterLogin` hook receives the OAuth access token, it can also be used to update the OAuth access token for the user in your database.
+Since the `onAfterLogin` hook receives the OAuth tokens, it can also be used to update the OAuth access token for the user in your database. It can also be used to [refresh the OAuth access token](#refreshing-the-oauth-access-token) if the provider supports it.
```
sodic, line 443: Passive voice.

web/docs/auth/auth-hooks.md
```diff
@@ -520,6 +528,54 @@ export const onAfterLogin: OnAfterLoginHook = async ({
 
 Read more about the data the `onAfterLogin` hook receives in the [API Reference](#the-onafterlogin-hook).
 
+### Refreshing the OAuth access token
+
+If you are using OAuth, you can use the `onAfterSignup` or `onAfterLogin` hooks to refresh the OAuth access token if the provider supports it.
```
sodic, line 532: Which of these hooks should they use for this and why?

Can we recommend/explain it?

web/docs/auth/auth-hooks.md
```diff
@@ -520,6 +528,54 @@ export const onAfterLogin: OnAfterLoginHook = async ({
...
+
+You'll need to import the OAuth client from the `wasp/server/oauth` module for the provider you are using. The OAuth client has a `refreshAccessToken` method that you can use to refresh the access token.
```
sodic, line 534:
```suggestion
Import the OAuth client for the desired provider from `wasp/server/oauth` and use its `refreshAccessToken` method to refresh the access token.
```
Suggestions based on:
 - Addressing the user directly
 - Prefer present tense
 - Prefer shorter and simpler sentences

web/docs/auth/auth-hooks.md
```diff
@@ -520,6 +528,54 @@ export const onAfterLogin: OnAfterLoginHook = async ({
...
+
+Here is an example of how you can refresh the Google OAuth access token:
```
sodic, line 536:
```suggestion
Here's an example of how you can refresh the access token for Google OAuth:
```
Subjective, but since we're talking about "refreshing the access token" all the time, I recommend keeping that phrase together, and appending the specific provider at the end.

web/docs/auth/auth-hooks.md
```diff
@@ -520,6 +528,54 @@ export const onAfterLogin: OnAfterLoginHook = async ({
...
+      oauth.tokens.refreshToken
+    )
+    log('new tokens', newTokens)
+  }
+}
+```
+</TabItem>
+</Tabs>
+
+Google provides you with the `accessTokenExpiresAt` field in the `oauth.tokens` object. You can use this field to determine when the access token expires and refresh it before it expires.
```
sodic, line 576:
```suggestion
Google exposes the `accessTokenExpiresAt` field in the `oauth.tokens` object. You can use this field to determine when the access token expires.
```
Subjective, but I like it because it's shorter.

I removed the "and refresh it before it expires." part because it suggests that they can use the field to refresh the token, which I don't think is right.

web/docs/auth/auth-hooks.md
```diff
@@ -520,6 +528,54 @@ export const onAfterLogin: OnAfterLoginHook = async ({
...
+
+You can of course perform the token refresh in an [Wasp Job](../advanced/jobs.md) if you want to refresh the token periodically.
```
sodic, line 578:
```suggestion
If you want to refresh the token periodically, use a [Wasp Job](../advanced/jobs.md).
```
Suggestion based on:
 - Start with a problem, follow with a solution.
 - Shorter sentences.

web/docs/auth/auth-hooks.md
```diff
@@ -828,9 +888,31 @@ Wasp passes the `oauth` object to the `onAfterSignup` and `onAfterLogin` hooks o
 
 It has the following fields:
 
-- `accessToken: string`
+- `providerName: string`
+
+  The OAuth provider name (e.g. `'google'`, `'github'`) that the user authenticated with.
```
sodic, line 894:
```suggestion
  The name of the OAuth provider the user authenticated with (e.g. `'google'`, `'github'`).
```
This formulation is less confusing IMO. The original soulds like the user authenticated with an "OAuth provider name."
Plus, the examples are in the middle of the sentence, breaking the flow.

web/docs/auth/auth-hooks.md
```diff
@@ -828,9 +888,31 @@ Wasp passes the `oauth` object to the `onAfterSignup` and `onAfterLogin` hooks o
...
+
+- `tokens: Tokens`
+
+  You can use the OAuth tokens to make requests to the provider's API on the user's behalf.
+
+  Depending on the OAuth provider, the `tokens` object might have different fields. For example, for Google, it has the `accessToken`, `refreshToken`, `idToken`, and `accessTokenExpiresAt` fields.
```
sodic, line 900:
```suggestion
  Depending on the OAuth provider, the `tokens` object might have different fields. For example, Google has the fields `accessToken`, `refreshToken`, `idToken`, and `accessTokenExpiresAt`.
```
Suggestions based on:
 - Fewer commas make a simpler sentence.
 - Listing stuff at the end of the sentence improves the flow.

web/docs/auth/auth-hooks.md
```diff
@@ -828,9 +888,31 @@ Wasp passes the `oauth` object to the `onAfterSignup` and `onAfterLogin` hooks o
...
+
+  <ShowForTs>
+
+  You can narrow down the `oauth.tokens` object type to the specific OAuth provider type to access the provider-specific fields:
```
sodic, line 904:
```suggestion
To access the provider-specific fields, you must first narrow down the `oauth.tokens` object type to he specific OAuth provider type.
```
Suggestions based on:
- Start with a problem, follow with a solution.
- can -> must (because they must :))

sodic (commented):
Nice work, just took a look at the docs for now to familiarize myself with the feature.

I'm now going for the code.

waspc/data/Generator/templates/server/src/auth/providers/oauth/handler.ts
```diff
@@ -104,8 +104,7 @@ export function createOAuthProviderRouter<OT extends OAuthType>({
             providerUserId,
             userSignupFields,
             req,
-            accessToken,
-            oAuthState,
+            oauth: { uniqueRequestId: oAuthState.state, tokens, provider: provider.id },
```
infomiho, line 107: TODO: fix this bug, it should be `providerName`

web/docs/auth/auth-hooks.md
infomiho: I keep writing using passive voice!

web/docs/auth/auth-hooks.md
infomiho: I've rewritten this section to be more clear, thanks for the nudge 👍

waspc/src/Wasp/Generator/SdkGenerator/Server/OAuthG.hs
```diff
@@ -0,0 +1,97 @@
...
+import qualified Wasp.Generator.SdkGenerator.Common as C
+import Wasp.Util ((<++>))
+
+genOAuth :: AS.Auth.Auth -> Generator [FileDraft]
+genOAuth auth
+  | AS.Auth.isExternalAuthEnabled auth =
+      sequence
+        [ genIndexTs auth,
+          genRedirectHelper,
+          genFileCopy [relfile|server/oauth/env.ts|],
```
sodic, line 37: How come this is not under `auth`?

waspc/src/Wasp/Generator/SdkGenerator/Server/OAuthG.hs
```diff
@@ -0,0 +1,97 @@
...
+genOAuthProvider provider maybeUserConfig
+  | isJust maybeUserConfig = sequence [genOAuthConfig provider]
+  | otherwise = return []
+
+genOAuthConfig ::
+  OAuthAuthProvider ->
+  Generator FileDraft
+genOAuthConfig provider = return $ C.mkTmplFdWithData tmplFile tmplData
+  where
+    tmplFile = [reldir|server/oauth/providers|] </> providerTsFile
```
sodic, line 81: I suggest extracting the root folder to avoid repeating it in all generator functions.

web/docs/auth/auth-hooks.md
```diff
@@ -433,7 +440,7 @@ Wasp calls the `onAfterLogin` hook after the user logs in.
 
 The `onAfterLogin` hook can be useful if you want to perform some action after the user logs in, like syncing the user with a third-party service.
 
-Since the `onAfterLogin` hook receives the OAuth access token, it can also be used to update the OAuth access token for the user in your database.
+Since the `onAfterLogin` hook receives the OAuth tokens, you can also use it to update the OAuth access token for the user in your database. You can also use it to [refresh the OAuth access token](#refreshing-the-oauth-access-token) if the provider supports it.
```
sodic, line 443: Redundant "also" in the first sentence I think (makes the second sentence seeem repetitive).

web/docs/auth/auth-hooks.md
```diff
@@ -520,6 +527,58 @@ export const onAfterLogin: OnAfterLoginHook = async ({
...
+    )
+    log('new tokens', newTokens)
+  }
+}
+```
+
+</TabItem>
+</Tabs>
+
+Google exposes the `accessTokenExpiresAt` field in the `oauth.tokens` object. You can use this field to determine when will the access token expire.
```
sodic, line 578:
```suggestion
Google exposes the `accessTokenExpiresAt` field in the `oauth.tokens` object. You can use this field to determine when the access token expires.
```
Based on preferring present tense, and it's shorter.

sodic (commented):
Nice work!

Left some suggestions, don't think I found anything major.

web/docs/auth/auth-hooks.md
infomiho: I already have this version locally, this is some weird Github diff.

waspc/src/Wasp/Generator/SdkGenerator/Server/OAuthG.hs
infomiho: I've moved it now, good idea 👍

waspc/data/Generator/templates/sdk/wasp/package.json
```diff
@@ -109,6 +109,7 @@
       "./client/test": "./dist/client/test/index.js",
       "./client": "./dist/client/index.js",
       "./dev": "./dist/dev/index.js",
+      "./server/oauth": "./dist/server/oauth/index.js",
```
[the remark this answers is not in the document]
infomiho: Yep, you are right. I wanted to group files, but that doesn't mean I can't export it from `wasp/server/auth`.

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
```diff
@@ -48,86 +50,98 @@ export type InternalAuthHookParams = {
...
-    /**
-     * Unique request ID that was generated during the OAuth flow.
-     */
-    uniqueRequestId: string
-  },
+  /**
+   * OAuth flow data that was generated during the OAuth flow. This is only
+   * available if the user signed up using OAuth.
+  */
+  oauth?: OAuthParams
```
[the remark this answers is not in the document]
infomiho: I like the look now e.g. `args.oauth.tokens` vs. `args.oauthData.tokens`.

The type should be called `OAuthData` I agree 👍

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
```diff
@@ -48,86 +50,98 @@ export type InternalAuthHookParams = {
...
    * URL that the OAuth flow should redirect to.
-   */
+  */
   url: URL
   /**
    * Unique request ID that was generated during the OAuth flow.
-   */
-  uniqueRequestId: string
+  */
+  uniqueRequestId: OAuthParams['uniqueRequestId']
```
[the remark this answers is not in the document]
infomiho: Sounds like a good change, keeping things consistent 👍 

We can't really test this in a e2e test because it depends on a redirect to Google 😢

waspc/data/Generator/templates/sdk/wasp/server/oauth/env.ts
```diff
@@ -1,14 +1,13 @@
-import { type ProviderConfig } from "wasp/auth/providers/types";
-
+// PRIVATE API (SDK)
 export function ensureEnvVarsForProvider<EnvVarName extends string>(
```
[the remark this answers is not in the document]
infomiho: I'd solve it in https://github.com/wasp-lang/wasp/issues/175 where we enumerate all the variables and validate in one place.

waspc/data/Generator/templates/sdk/wasp/server/oauth/oneTimeCode.ts
```diff
@@ -0,0 +1,50 @@
+import { createJWT, validateJWT, TimeSpan } from '../../auth/jwt.js'
+
+export const tokenStore = createTokenStore();
```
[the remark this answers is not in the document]
infomiho: We need the one time tokens to allow the client to exchange one time codes for session IDs. If the server and client were on the same domain, it would be possible. This is not the case right now, so we need this kind of a singleton that keeps track of the tokens. This is not introduced in this PR btw.

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
```diff
@@ -48,86 +50,98 @@ export type InternalAuthHookParams = {
...
+  */
+  oauth?: OAuthParams
   /**
    * Request object that can be used to access the incoming request.
-   */
+  */
   req: ExpressRequest
 } & InternalAuthHookParams
+
+// PRIVATE API (server)
```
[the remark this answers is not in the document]
infomiho: Didn't we want to hide parts in the top level exports so the users don't get autocomplete suggestions that are not really used by them? Based on that I put "private API" since ideally, users wouldn't need to see this when using the SDK.

waspc/data/Generator/templates/sdk/wasp/server/oauth/providers/github.ts
```diff
@@ -0,0 +1,17 @@
+{{={= =}=}}
+import { GitHub  } from "arctic";
+
+import { ensureEnvVarsForProvider } from "../env.js";
+
+export const id = "{= providerId =}";
+export const displayName = "{= displayName =}";
+
+const env = ensureEnvVarsForProvider(
+  ["GITHUB_CLIENT_ID", "GITHUB_CLIENT_SECRET"],
```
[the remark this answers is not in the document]
infomiho: I don't want to abstract this code away since it's pretty simple and there are subtle differences between OAuth provider initialisation. This is easy to read and easy to maintain IMHO.

waspc/data/Generator/templates/sdk/wasp/server/oauth/providers/github.ts
```diff
@@ -0,0 +1,17 @@
+{{={= =}=}}
+import { GitHub  } from "arctic";
+
+import { ensureEnvVarsForProvider } from "../env.js";
+
+export const id = "{= providerId =}";
```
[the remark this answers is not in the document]
infomiho: This worked out great in the end, it forced consistency and it'll be easier for new provider authors to follow along.

waspc/src/Wasp/Generator/SdkGenerator/Server/OAuthG.hs
sodic: I now see that no other module did this, so thanks for taking a step beyond :)

waspc/data/Generator/templates/sdk/wasp/server/auth/oauth/providers/discord.ts
```diff
@@ -0,0 +1,28 @@
...
+  getRedirectUriForCallback(id).toString(),
+);
+
+// PUBLIC API
+export const discord = defineProvider({
+  id,
+  displayName,
+  env,
+  oAuthClient,
+})
```
sodic, line 28: Missing empty lines in some of these files.

Working without the linter on these templates really is a pain. It's all trivial stuff, but very annoying (for example, once you notice how inconsistent we are with the semicolons, you can't unsee it).

waspc/data/Generator/templates/sdk/wasp/server/auth/oauth/index.ts
```diff
@@ -0,0 +1,31 @@
...
+export { google } from './providers/google.js';
+{=/ enabledProviders.isGoogleAuthEnabled =}
+{=# enabledProviders.isDiscordAuthEnabled =}
+// PUBLIC API
+export { discord } from './providers/discord.js';
+{=/ enabledProviders.isDiscordAuthEnabled =}
+{=# enabledProviders.isGitHubAuthEnabled =}
+// PUBLIC API
+export { github } from './providers/github.js';
+{=/ enabledProviders.isGitHubAuthEnabled =}
```
sodic, line 13: Another great change with scoping the providers flags under `enabledProviders`

sodic (approved):
Ok, resolved most of the stuff.

I left several more comments with suggestions I'd like, but it can definitely go out as is!

waspc/data/Generator/templates/sdk/wasp/server/oauth/oneTimeCode.ts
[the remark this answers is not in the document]
infomiho: It's not a singleton in a sense that there can only exist only one instance of the object. It's true that we create a single instance that we _have to use_ in both places where it's needed. In the case of unit tests, we could make the token store a param of the which ever fn we wanted to unit test.

waspc/data/Generator/templates/sdk/wasp/server/oauth/providers/github.ts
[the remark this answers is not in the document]
infomiho: I'm against calling this function with a spread operator while mapping over env vars, it feels too complex even with just that. I've written enough of the auth related code to get bitten by not keeping things dead simple. This is very basic Lucia OAuth client invocation and I'd like to keep it like that.

waspc/data/Generator/templates/sdk/wasp/server/auth/hooks.ts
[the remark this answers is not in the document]
infomiho: This should be a public API since the users can access it anyways through other types, yes? I'll make it public then 👍 It doesn't make sense to hide it and your other arguments make sense as well.

Issue #2222, opened 2024-08-02
Make Ext import extensions consitent


sodic:
## Summary
Our allowed external import extensions are all over the place:
- LSP, Snippets, and `wasp start` don't agree on what's allowed.
- Different rules apply for different declarations (depending on where Wasp uses the user's code).

Issues that already talk about this problem

Let's fix this.

## Details
There are three implicit categories of external imports (that I've discovered so far, there might be more). Each import falls into one or more of the categories: 
1. Ext imports used by the **SDK** (e.g., Queries).
2. Ext imports used by the **server** (e.g., server setup function, Queries).
3. Ext imports used by the **web-app** (e.g., pages, client setup function).

Notice how Queries fall into two categories. That's because they're copied into the SDK and also directly imported on the server (we do this to give users better error messages).

Since different imports end up in different build contexts, they are subject to different constraints (one context allows certain extensions, the other allows other extensions), making the behavior look random and confusing to users.

We've made some attempts at fixing this, and we have issues that talk about the problem:
- https://github.com/wasp-lang/wasp/pull/1714
- https://github.com/wasp-lang/wasp/issues/1717
- https://github.com/wasp-lang/wasp/issues/1363

But we never fully committed to properly deciding what we allow and where. The situation got worse after 0.12.0, which introduced an additional context with its own set of import constraints (the SDK).

## Solution
Craig made some good points here: https://github.com/wasp-lang/wasp/issues/1363#issuecomment-1670035480. He talks about two distinct but related problems:
- Handling extensions in the Wasp file (this issue).
- Handling extensions in user files (another issue, todo create).

We must:
1. Decide which extensions we want to support ([discord discussion](https://discord.com/channels/686873244791210014/1080864617347162122/1268855113301102633)).
2. Implement proper mappings in the SDK generator.
3. Implement proper mappings in the server generator.
4. Implement proper mappings in the web app generator.

By "mappings" I mean "mappings in Haskell and resolve configurations in the project." 

### Extra problem with the SDK build:
It's possible for the SDK build to pass, leaving extensions to cause problems in runtime. It happens with relative import references to files without extensions, and it also happened here: https://github.com/wasp-lang/wasp/issues/2096#issuecomment-2263566708.

I'm not yet sure 100% how this is possible. I'm guessing:
- TypeScript can resolve the files during the "Build SDK" step and doesn't change the extension.
- The runtime can't resolve the same fails with the same extensions and reports an error.

[...]

Issue #2096, opened 2024-06-12
Resolving `ext-src` imports in the SDK if the user declares an import with the `.js` ext


infomiho:
**Users get a weird error saying they can't import `queries.js.js something something`.** 

I've reproduced with the latest Wasp version: `0.13.2`

As far as I can see, it seems related to the exports fields and the `ext-src/*` entry 😄 

Two instances on Discord:
- https://ptb.discord.com/channels/686873244791210014/1250145383322091561/1250145383322091561
- https://ptb.discord.com/channels/686873244791210014/1249857194925752401/1249857194925752401

Martinsos:
Sending this back to triage, multiple people had this and error message is quite confusing, we should take care of it.

Martinsos:
This is in a way related to #2087

Martinsos:
User said following:

> Was odd because the issue was nonexistent until one day it wasn't.

That is peculiar!

sodic:
> That is peculiar! Might it mean that one of the deps updated minor version and caused this that way?

The error is only triggered when you import an operation from `wasp/server/operations` into server code that gets used (e.g., the server setup function or an API call). Here's a minimal repro example.

You need the following Wasp file:
```c
app extension {
  wasp: {
    version: "^0.14.0"
  },
  title: "example",
  server: { setupFn: import { setup } from "@src/setup" }
}

// ...

action jsAction {
  fn: import { jsAction } from "@src/jsAction.js"
}
```
And the following code in `setup.js`:
```javascript
import { jsAction } from "wasp/server/operations";

export function setup() {
  console.log(jsAction);
}
```
Without the import in `setup.js`, the error doesn't happen. This is why it our users probably felt the issue appeared out of thin air - it was caused by something seemingly unrelated (and I had a hard time tracking it down).

Since **this specific was accidentally fixed with Wasp 0.14.0** thanks to [this change](https://github.com/wasp-lang/wasp/blob/ecf1f8228774aaa2b8e1ec5d8d398dbc03171b95/waspc/src/Wasp/Generator/SdkGenerator/Server/OperationsGenerator.hs#L176) introduced in commit https://github.com/wasp-lang/wasp/pull/1930/commits/c184a9bd23d808475c78fcc75c2f7150b22e8b13, I'm closing the issue.

We still have other extension-related problems which I'll cover in a different issue.

PR #2225, opened 2024-08-02
Start making ext-src SDK imports relative

sodic:
Originally motivated by #2096, but it turns out that was solved with 0.14.0 (see the issue for details)

We still have other extension-related problems (described in #2222), so this is still something we should do.

### DIFF
--- a/waspc/src/FilePath/Extra.hs
+++ b/waspc/src/FilePath/Extra.hs
@@ -15,8 +15,8 @@ reversePosixPath :: FilePath -> FilePath
 reversePosixPath path
   | null parts = "."
   | otherwise =
-      assert (".." `notElem` parts) $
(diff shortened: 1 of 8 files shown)


infomiho:
@sodic what did you try to do here? Could you give me a high level explanation so I can review with that context in mind?

sodic:
I tried to change the ext imports inside the SDK:
- From `import { getFoo } from 'wasp/ext-src/operations/queries`
- To `import { getFoo } from '../../wasp/ext-src/operations/queries/index.js`

Here's my attempt: https://github.com/wasp-lang/wasp/pull/2225

I stopped because it opened a can of worms I documented in #2222

sodic:
I mentioned the PR in its relevat issue, we'll pick it up in the future.

PR #2230, opened 2024-08-08
reduce img size (#2228)

infomiho:


infomiho:
We'll replace this with a new PR

PR #2231, opened 2024-08-08
Update Prisma to v5

### LINKED ISSUE #1281: Migrate to latest Prisma
Two biggest changes since 4.12 to take into account:

1. Prisma now has "client extensions" feature, that enables hooking into various parts of Prisma Client. It replaces Prisma client middleware, which therefore becomes deprecated. If I am correct, we use Prisma middleware at the moment, specifically for hashing the user password, possibly some other things. We should look into using the "client extensions" for this now.
2. Arguments can now be passed to prisma db seed: `prisma db seed -- --foo bar`. Might be interesting to see if we want to utilize that somehow in our db seeding feature.

### LINKED ISSUE #1570: Fix Full-stack type safety type names
#1544 bumped the version of Prisma from 4.12 to 4.16.

This change introduced a regression in our full-stack type safety DX. The functionality is still there, but type names are far less ergonomic.

We should do our best to find a way to again show meaningful type names again, and also address other potential changes between Prisma 4.12 and 4.16.

### Before
image

### After
image

### LINKED ISSUE #2099: Fix strange Prisma inferred type errors
Something weird happens when:
- An operation doesn't explicitly specify its return type (i.e., it relies on type inference)
- Its return type is a Prisma entity.

More precisely, at some point of our pipeline, the type inferred for the operations API (rpc for the client, wrappers for the server) no longer matches the type inferred at the definition site.

When you fix this, make sure to update the type tests in `waspc/todoApp`.

### Example
I'll explain what's going on on the `taskToTaskUnspecified` action. You can find it in `waspc/todoApp` (at the time of writing).

Consider the action defined as:
```typescript
export const taskToTaskUnspecified = async (args: Task) => args
```
This is the return type inferred at the definition site:
```typescript
Promise<GetResult<{
    id: number;
    description: string;
    isDone: boolean;
    userId: number;
}, unknown> & {}>
```
And this is the return type inferred when I import it from `wasp/server/operations` or `wasp/client/operations`:
```typescript
Promise<GetResult<{
    id: number;
    description: string;
    isDone: boolean;
    userId: number;
}, unknown, never>>
```
The two types look similar, but they differ in the last type argument.: the original type doesn't specify the last type argument, while the "public" type specifies `never`.

infomiho:
Closes #1281 
Closes #1570
Closes #2099

### Left to do
- [x] [Blocker 🔴] Make sure CRUD features works again (Prisma types are not compatible with SuperJSON types for some reason)
- [x] Update e2e tests
- [x] ~~Update docs~~ No updates needed
- [x] Update example apps
- [x] Update starters
  - https://github.com/wasp-lang/starters/pull/9
  - https://github.com/wasp-lang/open-saas/pull/255
- [x] Format the outputted `schema.prisma` to enable checking if users ran the `npx prisma` command manually
- [x] Test with the `waspc/examples/todoApp`
- [x] Test with a new basic app
- [x] Test with embeddings app (uses preview features)
  - Works well, still need to update the starters template to have Prisma version in `package.json` as `5.18.0`
- [x] Update to latest Prisma version


waspc/data/Generator/templates/sdk/wasp/server/crud/_operationTypes.ts
```diff
@@ -103,7 +103,12 @@ export type CreateActionResolved = typeof _waspCreateAction
 
 {=# crud.operations.Update =}
 {=^ overrides.Update.isDefined =}
-type UpdateInput = Prisma.{= crud.entityUpper =}UpdateInput & Prisma.{= crud.entityUpper =}WhereUniqueInput
+type UpdateInput = Prisma.XOR<
```
infomiho, line 106: Fixes `UpdateInput` to match what Prisma Client expects.

waspc/data/Generator/templates/sdk/wasp/prisma-runtime-library.d.ts
```diff
@@ -0,0 +1,14 @@
+// Without this import, Prisma types are resolved incorrectly.
```
infomiho, line 1: @sodic this is the fix I implemented in the end, it slightly nudges the `FieldRef` type to work with our SuperJSON serialization. 

I went down the rabbit hole of recursively converting all interfaces into types using `Pick<T, keyof T>` (as suggested by some in the Github issue) but this ended up being less problematic and simple enough.

waspc/data/Generator/templates/sdk/wasp/server/crud/_operationTypes.ts
```diff
@@ -91,7 +91,10 @@ export type GetQueryResolved = typeof _waspGetQuery
 
 {=# crud.operations.Create =}
 {=^ overrides.Create.isDefined =}
-type CreateInput = Prisma.{= crud.entityUpper =}CreateInput
+type CreateInput = Prisma.XOR<
```
infomiho, line 94: Fixes `CreateInput` to match what Prisma Client expects.

waspc/data/Generator/templates/server/src/crud/_operations.ts
```diff
@@ -39,9 +50,7 @@ const entities = {
 // 1. We either use the default implementation of the operation...
 =}
 {=^ overrides.GetAll.isDefined =}
-type GetAllInput = {}
```
infomiho, line 42: For some reason, we built these types in two places: in the server and the in the SDK, but there was no need since the server can import the built types from the SDK. There is only one place where CRUD operation types are now defined.

waspc/examples/crud-testing/schema.prisma
```diff
@@ -1,5 +1,5 @@
 datasource db {
-  provider = "sqlite"
+  provider = "postgresql"
```
infomiho, line 2: Mistake that was left after the migration to `schema.prisma` file. I'm not sure how that got through.

waspc/examples/todoApp/src/testTypes/operations/client.ts
```diff
@@ -32,17 +32,17 @@ import { QueryMetadata } from 'wasp/client/operations/rpc'
 
 type TestCases = [
   Expect<Equal<typeof taskToTaskSpecified, (args: Task) => Promise<Task>>>,
-  // todo(filip): Prisma errors casuing this test to fail, try to add Except
-  // after updating Prisma: https://github.com/wasp-lang/wasp/issues/2099
-  Equal<
-    typeof taskToTaskUnspecified,
-    (args: Task) => ReturnType<typeof taskToTaskUnspecifiedDefinition>
+  Expect<
```
infomiho, line 35: These tests now pass 👍

waspc/examples/todo-typescript/src/jobs/print.ts
```diff
@@ -5,6 +5,6 @@ export const printTimeAndNumberOfTasks: PrintTimeAndNumberOfTasks<
   {},
   void
 > = async (data, context) => {
-  const count = await getTotalNumberOfTasks(null, context)
+  const count = await getTotalNumberOfTasks()
```
infomiho, line 8: This broke after the last operations API update, quick fix.

examples/thoughts/package-lock.json
Martinsos, line 1: @infomiho you tested all these apps with the new Prisma? They all work fine?

waspc/cli/exe/Main.hs
```diff
@@ -143,11 +143,7 @@ setDefaultCliEnvVars = do
   where
     cliEnvVars :: [(String, String)]
     cliEnvVars =
-      [ ("PRISMA_HIDE_UPDATE_MESSAGE", "true"),
-        -- NOTE: We were getting errors from Prisma v4 related to their Checkpoint system
```
Martinsos, line 147: Ok sweet!

waspc/src/Wasp/Generator/DbGenerator.hs
```diff
@@ -143,6 +146,18 @@ postWriteDbGeneratorActions spec dstDir = do
   dbGeneratorErrors <- maybeToList <$> generatePrismaClient spec dstDir
   pure (dbGeneratorWarnings, dbGeneratorErrors)
 
+-- | One of the checks we perform is to compare the Wasp generated schema.prisma file
+-- and the schema.prisma file in the node_modules. Prisma formats the schema in node_modules
+-- automatically, so we have to do the same to be able to compare them.
+formatPrismaSchemaFileOnDisk :: Path' Abs (Dir ProjectRootDir) -> IO ()
```
Martinsos, line 152: How is it that this is new?

Martinsos (commented):
Cool stuff @infomiho ! I will leave it to @sodic to check the Typescript stuff. The rest is looking good to me.
I assume you did a "diff" between the Prisma version we were on and Prisma 5, by checking their Changelog -> is there anything new we should be adding support for in Wasp, or taht significantly changed? Seems from this PR that not much changed on the API level?

waspc/src/Wasp/Generator/DbGenerator.hs
infomiho: I wasn't precise enough

This is new:
> Prisma formats the schema in node_modules automatically

They didn't do that and we could do our diff check without formatting our output.

examples/thoughts/package-lock.json
infomiho: Yep, all the apps were tested and they worked fine. The v5 major version was more about the new JSON protocol than anything, so nothing broke in our apps really :)

infomiho:
> I assume you did a "diff" between the Prisma version we were on and Prisma 5, by checking their Changelog

According to their migration guide to v5: min. version of Node.js, Typescript and PostgreSQL changes, `rejectOnNotFound` was removed (we didn't use it) and some other stuff: https://www.prisma.io/docs/orm/more/upgrade-guides/upgrading-versions/upgrading-to-prisma-5#removal-of-array-shortcuts%23removal-of-runtimeindexjs-from-generated-client

But nothing that major. I'll go over the release notes to see if I missed something 👍 

>  is there anything new we should be adding support for in Wasp, or taht significantly changed?

There are new things, yes! I'm excited for their `join` support which would optimise some of our auth DB code - but I wouldn't really do it in this PR. I'll create an issue for that.

> Seems from this PR that not much changed on the API level?

Yep, they kept things pretty much the same - which is great for us and our users :)

waspc/src/Wasp/Generator/DbGenerator.hs
sodic: Makes sense. One question: How come we aren't comparing the ASTs? Seems more robust.

I'm guessing it was probably too much work for too litle gain.

Martinsos:
> > is there anything new we should be adding support for in Wasp, or taht significantly changed?


Awesome! Here I was asking more in the direction of, did they add something that we need to put extra effort to enable people to use it, becaues Wasp somehow doesn't allow it? But it sounds like there is nothing like that, at least nothing major.

sodic (approved):
Looks great!

Please just augment that comment and merge when ready.

waspc/data/Generator/templates/sdk/wasp/server/crud/_operationTypes.ts
infomiho: I copied the type the `create` method has defined for its `data` input. Before I only use the simpler type because I thought it was enough, but now I realise it's better to follow 1:1 what Prisma uses internally - users will get the same type as if they were using `create` directly themselves.

waspc/src/Wasp/Generator/DbGenerator.hs
infomiho: We are in the business of comparing checksums like in all other places in the Wasp compiler :) that's why the source files needed to be the same.

waspc/data/Generator/templates/sdk/wasp/prisma-runtime-library.d.ts
infomiho: I've updated the comment to mention the playground to make it more complete as an explanation.

> BTW, what happend with trying to fix the root of the problem, the definition of the JSONSerializable type (as attempted [here](https://hackernoon.com/mastering-type-safe-json-serialization-in-typescript)). Did Superjson make it too complex?

We concluded that `JSONSerializable` would stop users from using the Prisma field reference feature: https://www.prisma.io/docs/orm/reference/prisma-client-reference#compare-columns-in-the-same-table since the proposed approach drops fields.

Then we set on to recursively convert all interfaces to types - which in the end was a long way of doing this `d.ts` patch we have in this repo.

Issue #1446, opened 2023-09-14
Upgrade our Haskell toolchain


infomiho:
We are looking to upgrade our Haskell toolchain to the latest LTS/stable version. That would mean 9.X.X for GHC and we need to investigate which makes most sense for Wasp.

Possible steps:
1. Investigate which tools we need to upgrade (GHC, Cabal)
2. Decide on the versions we want to use
3. Upgrade
4. Test

infomiho:
When we upgrade the Haskell tool chain, make sure we update the Github actions used as well: https://github.com/wasp-lang/wasp/pull/2239

Martinsos:
We should also update the ormolu version we use. Would be also cool if there is a way to specify ormolu version per-project -> right now if I am working on two Haskell projects and one of them uses older ormolu and the other one uses the newer one, I can't easily switch because ormolu is on whatever version is installed via plugin for my IDE. Maybe we just need to tell editor to use the local version of ormolu when working on a specific project, the one we currently install as a `dev-tool`? If we can specify that in our plugins, to get it to not use the global ormolu installation but the local one per-project, and if that one is fast enough, then that is probably how it should be done (and that would reflect how it is done in other ecosystems). Yeah that must be the right way. Maybe there is a way to specify it in cabal file? We should read ormolu docs again, and also ask around how others do it.

Martinsos:
Ok yeah from vscode you can tell it to use ormolu from specific path, this is what @FranjoMindek does:
```
{
  "ormolu.path": "~/git/wasp/waspc/.bin/ormolu",
  "haskell.manageHLS": "GHCup",
  "haskell.toolchain": {
    "hls": "1.8.0.0"
  },
  "[haskell]": {
    "editor.defaultFormatter": "sjurmillidahl.ormolu-vscode",
    "editor.tabSize": 2
  },
}
```

cprecioso:
Whoever takes this, you might want to take a look at #2759 while you're here

FranjoMindek:
Note when updating `ormolu` take care to properly also update `run-ormolu` action inside of `check-formatting.yaml`.

https://github.com/wasp-lang/wasp/pull/3112/files#diff-b33df3bc1b0e0e2d5de6a1598d468a6fea6cf9dc2660c363ade3638ad011b91aR46-R47

We need to bump the version if we end up with `ormolu` version >= 0.7.5.0

PR #2239, opened 2024-08-14
Updates actions in the CI

infomiho:
The action we used to setup Haskell got deprecated: https://github.com/haskell/actions?tab=readme-ov-file

EDIT: this might be a no go, since MacOS builds fail. Related issue: https://github.com/haskell-actions/setup/issues/77 -> workaround is to use MacOS 13 instead of the latest MacOS 14.


waspc/README.md
```diff
@@ -11,15 +11,15 @@ If you would like to make your first contribution, here is a handy checklist we
 - [ ] Read [Quick overview](#quick-overview).
 - [ ] Compile the project successfully and get todoApp example running (follow [Basics](#basics)).
 - [ ] Join [Discord](https://discord.gg/rzdnErX) and say hi :)!
-- [ ] Pick an issue [labeled with "good first issue"](https://github.com/wasp-lang/wasp/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) and let us know you would like to work on it - ideally immediatelly propose a plan of action and ask questions.
+- [ ] Pick an issue [labeled with "good first issue"](https://github.com/wasp-lang/wasp/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) and let us know you would like to work on it - ideally immediately propose a plan of action and ask questions.
```
infomiho, line 14: Changes in `waspc` dir to trigger the CI

Martinsos:
> The action we used to setup Haskell got deprecated: https://github.com/haskell/actions?tab=readme-ov-file
> 
> EDIT: this might be a no go, since MacOS builds fail. Related issue: [haskell-actions/setup#77](https://github.com/haskell-actions/setup/issues/77) -> workaround is to use MacOS 13 instead of the latest MacOS 14.

What would also help is us upgrading to the newer GHC!

infomiho:
@Martinsos How big of an undertaking is to upgrade to GHC 9.X?

Martinsos:
> @Martinsos How big of an undertaking is to upgrade to GHC 9.X?

Ha not sure, if I had to estimate it now, I would say it would take me, who knows the Haskell setup we currently have the best, probably a couple of days. Depends if hit any dep that is problematic or not. I would say 2 days if we are lucky, 5 days if there are a lot of deps that need a major update (Tasty testing suite might be such one).

Martinsos:
Issue we have here: https://github.com/wasp-lang/wasp/issues/1446 , you opened it.

We could go after this sooner than later.

infomiho:
Closing this as stale and I'll add a comment in the #1446 issue to consider implementing the changes from here as well.

PR #2243, opened 2024-08-21
Use native DB type in injected auth models if user uses it

### LINKED ISSUE #2167: Using native DB fields for `User` id is not supported when using Wasp auth
If users want to use Wasp's auth system, we ask users to specify the `userEntity` e.g. `User`. We then in the background connect the `User` entity with our internal `Auth` entity via a one-to-one relationship. We take into account the type of the `User` entity ID field when creating the relation.

If it's a `String` or `Int` we'll use that in the `Auth` entity:

```diff
  authEntityPslBody =
    T.unpack
      [trimming|
        id ${authEntityIdTypeText}   @id @default(uuid())
+       userId    ${userEntityIdTypeText}? @unique
        ${userFieldOnAuthEntityNameText}      ${userEntityNameText}?    @relation(fields: [userId], references: [${userEntityIdFieldName}], onDelete: Cascade)
        ${identitiesFieldOnAuthEntityNameText} ${authIdentityEntityNameText}[]
        ${sessionsFieldOnAuthEntityNameText}   ${sessionEntityNameText}[]
      |]
```

What we don't take into account is that if the user uses native DB types e.g. `@db.Uuid`:
```diff
model User {
+ id                String       @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
  createdAt         DateTime     @default(now())
  updatedAt         DateTime     @default(now())
  email             String       @unique
  firstName         String
  lastName          String
  phone             String       @unique
}
```
The resulting `Auth` entity will have the `userId` type as `String` but it will be missing `@db.Uuid` annotation. 

### Solution

We should take into account the `@db.{type}` attributes and apply them alongside the `userEntityIdTypeText`, something like this:
```diff
  authEntityPslBody =
    T.unpack
      [trimming|
        id ${authEntityIdTypeText}   @id @default(uuid())
+       userId    ${userEntityIdTypeText}? @unique ${optionalNativeDbTypeText}
        ${userFieldOnAuthEntityNameText}      ${userEntityNameText}?    @relation(fields: [userId], references: [${userEntityIdFieldName}], onDelete: Cascade)
        ${identitiesFieldOnAuthEntityNameText} ${authIdentityEntityNameText}[]
        ${sessionsFieldOnAuthEntityNameText}   ${sessionEntityNameText}[]
      |]
```

Based on this comment: https://github.com/wasp-lang/wasp/issues/2137#issuecomment-2213965587

infomiho:
Closes #2167


waspc/src/Wasp/Psl/Ast/Attribute.hs
```diff
@@ -19,3 +21,7 @@ data Attribute = Attribute
     _attrArgs :: [Psl.Argument.Argument]
   }
   deriving (Show, Eq, Data)
+
+-- | @db.Uuid or @db.String or @db.VarChar are examples of native db types.
+isNativeDbTypeAttr :: Attribute -> Bool
+isNativeDbTypeAttr = isPrefixOf "db." . _attrName
```
Martinsos, line 27:
```suggestion
isNativeDbTypeAttr = ("db." `isPrefixOf`) . _attrName
```
reads nicer :D

waspc/test/Generator/AuthInjectionTest.hs
Martinsos, line 1: A lot of AST here which is a bit hard to read / maintain -> any way we might be able to replace it with actual PSL, here in the test? That is parsed either during test runtime, or even test compile time if that is doable (although I think I remember we had hard time with that).
If not, never mind, but if yes, I think it could make the test easier to read and maintain.

Martinsos (approved):
@infomiho nice stuff! I approved it, but check the comment about possibly making the test more readable.

waspc/test/Generator/AuthInjectionTest.hs
infomiho: > That is parsed either during test runtime

I've done this bit since it was similar to something we already did with the full schema file. https://github.com/wasp-lang/wasp/blob/362ccff7181053a5c479fa72c38b92c8a20019ce/waspc/test/Util/Prisma.hs#L10

PR #2256, opened 2024-08-27
Bumps `react-router-dom` to v6

### LINKED ISSUE #2245: Update React router to the latest version
### Old comment
Some migration will be needed (some recommended styles of doing stuff have changed). Check https://reactrouter.com/en/main/upgrading/v5, and don't skip chapter with "5.1" instructions . We will need to move from `<Route component` to `<Route children` approach, that much I read, then didn't read further.

infomiho:
Closes #2245 

### Left to do
- [x] Update the type-safe `Link` component
  - [x] Optional static segment
  - [x] Splat routes
  - [x] `.build` method on the route
- [x] Test all example apps
- [x] Test all starters
- [x] Update docs
  - [x] Update API usage (e.g. `useHistory` is replaced with `useNavigate`)
  - [x] Update `<Link />` component docs
    - [x] Talk about the "splat" route 
    - [x] Talk about optional static segments syntax
  - [x] Update tutorial not to use `match.params`
  - [x] Update `rootComponent` to use `<Outlet />` instead of `children`
- [x] Explore using lazy loading by default for pages - https://github.com/wasp-lang/wasp/commit/f65f929cb1fe35784185e5ba45ee4a704a72355f
- [x] Explore using a non-default 404 error page (default image below)

| Screenshot 2024-09-09 at 15 26 37 |
| --- |
| Default 404 page in dev |
| Screenshot 2024-09-09 at 15 28 51 |
| Default 404 page in prod |

### Breaking changes

- `useHistory()` -> `useNavigate()`
- `<Redirect />` -> `<Navigate />`
- Wasp's root component feature requires rendering `<Outlet />` instead of rendering children
- `props.match.params` -> `useParams()`


waspc/data/Generator/templates/react-app/src/auth/pages/OAuthCallback.tsx
```diff
@@ -14,7 +14,7 @@ export function OAuthCallbackPage() {
   const { error, user } = useOAuthCallbackHandler();
   
   if (user !== undefined && user !== null) {
-    return <Redirect to="{= onAuthSucceededRedirectTo =}" />;
+    return <Navigate to="{= onAuthSucceededRedirectTo =}" replace />;
```
infomiho, line 17: `<Navigate />` replace `<Redirect />` and the default behaviour changed, we need to add `replace` explicitly.

waspc/data/Generator/templates/react-app/src/router.tsx
```diff
@@ -25,35 +25,31 @@ export const routeNameToRouteComponent = {
...
+  element: <{= rootComponent.importIdentifier =} />,
+  {=/ rootComponent.isDefined =}
+  children: [
+    ...waspDefinedRoutes,
+    ...userDefinedRoutes,
+  ],
+}])
+
+
+export const router = <RouterProvider router={browserRouter} />
```
infomiho, line 55: From the docs:
> You should only opt into the Component API for data routes via RouterProvider. Using this API on a <Route> inside <Routes> will de-optimize React's ability to reuse the created element across renders.

Since we are using `Component` and not element API, using the `RouterProvider` is optimal.

waspc/src/Wasp/Generator/WebAppGenerator.hs
```diff
@@ -135,7 +135,6 @@ npmDepsForWasp _spec =
             ("typescript", "^5.1.0"),
             ("@types/react", "^18.0.37"),
             ("@types/react-dom", "^18.0.11"),
-            ("@types/react-router-dom", "^5.3.3"),
```
infomiho, line 138: Types are not included in the package, there is no need for extra `@types` package.

examples/websockets-realtime-voting/src/Layout.tsx
```diff
@@ -40,7 +39,9 @@ export const Layout = ({ children }: { children: ReactNode }) => {
                 }
               >
                 <Dropdown.Header>
-                  <span className="block text-sm">{user.getFirstProviderUserId()}</span>
+                  <span className="block text-sm">
```
infomiho, line 42: Formatting

examples/websockets-realtime-voting/src/Layout.tsx
```diff
@@ -61,7 +62,9 @@ export const Layout = ({ children }: { children: ReactNode }) => {
             <Navbar.Link href="#">Contact</Navbar.Link>
           </Navbar.Collapse> */}
         </Navbar>
-        <div className="grid place-items-center mt-8">{children}</div>
+        <div className="grid place-items-center mt-8">
+          <Outlet />
```
infomiho, line 66: Breaking change: `children` -> `<Outlet />`

waspc/data/Generator/templates/react-app/src/components/DefaultRootErrorBoundary.tsx
```diff
@@ -0,0 +1,13 @@
+import { useRouteError } from 'react-router-dom'
+
+import { FullPageWrapper } from './FullPageWrapper'
+
+export function DefaultRootErrorBoundary() {
```
infomiho, line 5: We are using a default root error boundary to make it look nicer when rendering fails. In the future, we should allow customisation of the root error boundary: https://github.com/wasp-lang/wasp/issues/2279

waspc/data/Generator/templates/sdk/wasp/client/router/index.ts
```diff
@@ -4,25 +4,22 @@ import type {
...
-      } & OptionalRouteOptions,
-    ) => interpolatePath("{= urlPath =}", options.params, options?.search, options?.hash),
-    {=/ hasUrlParams =}
-    {=^ hasUrlParams =}
-    build: (
-      options?: OptionalRouteOptions,
-    ) => interpolatePath("{= urlPath =}", undefined, options?.search, options?.hash),
-    {=/ hasUrlParams =}
+      }{=/ hasUrlParams =}
+      {=# hasOptionalStaticSegments =}& { path: ExpandRouteOnOptionalStaticSegments<"{= urlPath =}"> }{=/ hasOptionalStaticSegments =}
```
infomiho, line 21: If a route has optional static segments e.g. `/my/:id/route?` which can be either `/my/:id` or `/my/:id/route`, we require the user to specify which variant they want to use by providing the `path` argument.

waspc/data/Generator/templates/sdk/wasp/client/router/linkHelpers.ts
```diff
@@ -17,6 +17,9 @@ export function interpolatePath(
 
 function interpolatePathParams(path: string, params: Params) {
   function mapPathPart(part: string) {
+    if (part === '*') {
```
infomiho, line 20: React Router has wild card routes that are called splats, so we need to support users specifying what the `*` part should be. We called that param `splat` to keep in line with the React Router naming.

waspc/data/Generator/templates/sdk/wasp/client/router/types.ts
```diff
@@ -21,7 +21,7 @@ export type Search =
 
 type RouteDefinitionsToRoutesObj<Routes extends RoutesDefinition> = {
   [K in keyof Routes]: {
-    to: Routes[K]['to']
+    to: ExpandRouteOnOptionalStaticSegments<Routes[K]['to']>
```
infomiho, line 24: This is a bit of Typescript type level programming magic. It converts string literals that have a route with optional static segments into a union of string literals of possible route variants. 

`ExpandRouteOnOptionalStaticSegments<'/my/:id/route?'>` will result in `'/my/:id' | '/my/:id/route'` type.

waspc/src/Wasp/Generator/SdkGenerator.hs
```diff
@@ -188,13 +188,9 @@ npmDepsForSdk spec =
...
             ("react", "^18.2.0"),
             ("lodash.merge", "^4.6.2"),
-            ("react-router-dom", "^5.3.3"),
+            ("react-router-dom", "^6.26.1"),
             ("react-hook-form", "^7.45.4"),
-            ("superjson", "^1.12.2"),
-            -- Todo: why is this in dependencies, should it be in dev dependencies?
-            -- Should it go into their package.json
-            ("@types/express-serve-static-core", "^4.17.13"),
-            ("@types/react-router-dom", "^5.3.3")
```
infomiho, line 197: React Router 6 comes with types, no need for the `@types/` package.

waspc/data/Generator/templates/react-app/src/auth/pages/OAuthCallback.tsx
sodic: If you got links to the changelog that prescribes this, you can include it into these comments.

I'm not sure is it worth spending the time, but it would help me better judge the change's validity.

waspc/src/Wasp/Generator/WebAppGenerator.hs
sodic: What do you mean by this? Where are they not included?

examples/websockets-realtime-voting/src/Layout.tsx
sodic: Yes, children are often used as an emotional outlet for their parents, checks out.

waspc/src/Wasp/Generator/WebAppGenerator.hs
infomiho: My bad, I miswrote: types _are_ included in the package.

waspc/data/Generator/templates/react-app/src/auth/pages/OAuthCallback.tsx
infomiho: https://reactrouter.com/en/main/upgrading/v5#use-usenavigate-instead-of-usehistory

> If you prefer to use a declarative API for navigation (ala v5's Redirect component), v6 provides a Navigate component.

web/docs/advanced/links.md
```diff
@@ -40,6 +40,44 @@ export const TaskList = () => {
 }
 ```
 
+### Catch-all Routes
+
+If a route path ends with a `/*` pattern (also known as [splat](https://reactrouter.com/en/main/route/route#splats)), you can use the `Link` component like this:
```
sodic, line 45:
```suggestion
If a route path ends with a `/*` pattern (also known as a [splat](https://reactrouter.com/en/main/route/route#splats)), you can use the `Link` component like this:
```
Also, is this only allowed at the route's end? Can't it be in the middle as well?

web/docs/advanced/links.md
```diff
@@ -69,18 +107,36 @@ const linkToTask = routes.TaskRoute.build({ params: { id: 1 } })
 
 This will result in a link like this: `/task/1`.
 
-You can also pass `search` and `hash` props to the `build` function. Check out the [API Reference](#routes-object) for more details.
+### Optional Static Segments
```
sodic, line 110: Why is this repeated twice?

web/docs/advanced/links.md
```diff
@@ -69,18 +107,36 @@ const linkToTask = routes.TaskRoute.build({ params: { id: 1 } })
...
 
 ### `Link` Component
 
 The `Link` component accepts the following props:
+
 - `to` <Required />
 
   - A valid Wasp Route path from your `main.wasp` file.
 
+    In the case of optional static segments, you must provide one of the possible paths which include or exclude the optional segment. For example, if the path is `/task/:id/details?`, you must provide either `/task/:id/details` or `/task/:id`.
```
sodic, line 138: Questionmarks at the end missing I think.

web/docs/advanced/links.md
```diff
@@ -109,16 +164,39 @@ export const routes = {
...
       }
     ) => // ...
-  }
+  },
+
+  // OptionalRoute has a path like "/task/:id/details?"
+  OptionalRoute: {
+    build: (
+      options: {
+        path: '/task/:id/details' | '/task/:id',
```
sodic, line 182: How come this is the only one that specifies `path`?

waspc/src/Wasp/Util/WebRouterPath.hs
```diff
@@ -1,18 +1,39 @@
...
-        if "?" `isSuffixOf` xs
-          then Optional (take (length xs - 1) xs)
-          else Required xs
-    parseParam _ = Nothing
+    parseSegment :: String -> Segment
+    parseSegment "*" = ParamSegment $ RequiredParamSegment "splat"
+    parseSegment (':' : xs) =
+      ParamSegment $
+        if isSegmentOptional xs
+          then OptionalParamSegment (take (length xs - 1) xs)
```
sodic, line 30: I think the `init` function from Prelude does exactly this.

waspc/src/Wasp/Util/WebRouterPath.hs
```diff
@@ -1,18 +1,39 @@
...
-    parseParam :: String -> Maybe Param
-    parseParam (':' : xs) =
-      Just $
-        if "?" `isSuffixOf` xs
-          then Optional (take (length xs - 1) xs)
-          else Required xs
-    parseParam _ = Nothing
+    parseSegment :: String -> Segment
+    parseSegment "*" = ParamSegment $ RequiredParamSegment "splat"
+    parseSegment (':' : xs) =
```
sodic, line 27: Can we use something other than `xs`.Even `segments` is fine by me :)

waspc/src/Wasp/Util/WebRouterPath.hs
```diff
@@ -1,18 +1,39 @@
...
+      ParamSegment $
+        if isSegmentOptional xs
+          then OptionalParamSegment (take (length xs - 1) xs)
+          else RequiredParamSegment xs
+    parseSegment x =
```
sodic, line 32: Can we also change the `x` to something more meaningful?

waspc/src/Wasp/Util/WebRouterPath.hs
```diff
@@ -1,18 +1,39 @@
...
-import Data.Maybe (mapMaybe)
 
-data Param = Optional String | Required String deriving (Show, Eq)
+data Segment = StaticSegment StaticSegment | ParamSegment ParamSegment deriving (Show, Eq)
 
-extractPathParams :: String -> [Param]
-extractPathParams = mapMaybe parseParam . splitOn "/"
+data StaticSegment = RequiredStaticSegment StaticSegmentValue | OptionalStaticSegment StaticSegmentValue deriving (Show, Eq)
+
+data ParamSegment = RequiredParamSegment ParamName | OptionalParamSegment ParamName deriving (Show, Eq)
```
sodic, line 16: I recommend breaking these up into multiple lines, more readable that way. Ormolu won't do it unless you nudge it (and it's unfortunatley not configurable).

waspc/src/Wasp/Generator/SdkGenerator.hs
sodic: It's weird that's part of the dependencies (and not dev dependencies), right? Or am I missing something?

waspc/src/Wasp/Generator/SdkGenerator.hs
```diff
@@ -215,7 +211,9 @@ npmDepsForSdk spec =
           ++ depsRequiredByTailwind spec,
       N.devDependencies =
         AS.Dependency.fromList
-          [ ("@tsconfig/node" <> majorNodeVersionStr, "latest")
+          [ ("@tsconfig/node" <> majorNodeVersionStr, "latest"),
+            -- Should @types/* go into their package.json?
+            ("@types/express-serve-static-core", "^4.17.13")
```
sodic, line 216: Where did this go?

waspc/data/Generator/templates/react-app/src/router.tsx
```diff
@@ -25,35 +27,31 @@ export const routeNameToRouteComponent = {
...
+  return {
+    path: route.to,
+    Component: routeNameToRouteComponent[routeKey],
+  }
+})
+
+const browserRouter = createBrowserRouter([{
+  path: '/',
+  {=# rootComponent.isDefined =}
+  element: <{= rootComponent.importIdentifier =} />,
```
sodic, line 48: Pleas make sure the root component still works as expected. I'm guessing you did, but mentioning just in case.

waspc/data/Generator/templates/sdk/wasp/client/router/types.ts
```diff
@@ -39,3 +39,56 @@ type ParamsFromBuildFn<BF extends BuildFn> = Parameters<BF>[0] extends {
...
   ? { params: Params }
   : { params?: never }
+
+// PRIVATE API (sdk)
+/**
+ * Optional static segments handling: expands routes with optional segments
+ * into multiple routes, one for each possible combination of optional segments.
+ *
+ * For example: /users/tasks?/:id? will be expanded into two routes:
+ * - /users/:id
```
sodic, line 49: Querstionmarks missing in the comment.

sodic (commented):
Nice work!

I'm still looking at the types file, but more out of curiosity than necessity.

If we fix everything else, it can go as is.

web/docs/advanced/links.md
sodic: Also, what about this:
```typescript
type Test = ExpandRouteOnOptionalStaticSegments<'/a?'> 
```
Should it be `'/a'` or `'/' | '/a'`?

examples/thoughts/src/client/MainPage.jsx
```diff
@@ -61,7 +61,7 @@ const Thought = (props) => {
       }
       try {
         await createThought({ textMarkdown: textMd.trim(), tagNames })
-        history.push('/thoughts') // TODO: Would be cool if this was type checked somehow or if string was coming from the Wasp API.
+        navigate('/thoughts') // TODO: Would be cool if this was type checked somehow or if string was coming from the Wasp API.
```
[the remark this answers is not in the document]
infomiho: We could be using the `routes` helper - true - I'll do that.

web/docs/advanced/links.md
infomiho: > Questionmarks at the end missing I think.

What do you mean exactly? 

> type Test = ExpandRouteOnOptionalStaticSegments<'/a?'> 

I believe this is an edge case I didn't handle, I'll explore the React Router behaviour and adjust the code.

web/docs/advanced/links.md
infomiho: For others, we know the exact string on which we will operate. When optional static segments are involved - we need the user to tell us which variant to use.

waspc/src/Wasp/Util/WebRouterPath.hs
infomiho: Nice, I'll upgrade my code 👍

waspc/src/Wasp/Generator/SdkGenerator.hs
infomiho: I'm not sure what you mean?

waspc/data/Generator/templates/react-app/src/router.tsx
infomiho: Yes, I have tested having the root component, not having it at all. It works as expected in all the example apps.

waspc/data/Generator/templates/sdk/wasp/client/router/index.ts
infomiho: I agree 100% 👍 

I realised we have 2 degrees of freedom (has params, has optional static segments) and didn't want to do 4 variants. But 2 variants are enough, makes it a lot cleaner!

waspc/data/Generator/templates/sdk/wasp/client/router/linkHelpers.ts
infomiho: Good catch, `splat` is their secondary name 👍

web/docs/advanced/links.md
sodic: Questionmarks -> same story as here: https://github.com/wasp-lang/wasp/pull/2256#discussion_r1773519169

infomiho: Aw yes, this is written as intended. `/task/:id/details?` expands into `/task/:id/details` or `/task/:id`.

waspc/src/Wasp/Generator/SdkGenerator.hs
sodic: My bad, brain fart.

web/docs/advanced/links.md
sodic: It seems they both ended up in the guide:
image

I don't a chapter dedicated to it in the API reference. Maybe you mispositioned it?

waspc/src/Wasp/Util/WebRouterPath.hs
sodic: The `Segment` is still in one line. Not sure if that's on purpose.

web/docs/advanced/links.md
infomiho: Oh my bad, I had a brain fart it seems. The reason this is repeated is to explain the same concept in two different contexts: `Link` component and the `routes` object.

It's a bit repetitive, I agree, but it's easier to follow if the structure is the same in both places so users can better pattern match "ah yes, that's the same thing". At least that's my goal here.

waspc/src/Wasp/Util/WebRouterPath.hs
infomiho: Ah yes, my bad! Changed now :)

waspc/src/Wasp/Util/WebRouterPath.hs
```diff
@@ -1,18 +1,45 @@
...
+          then OptionalParamSegment $ init paramName
+          else RequiredParamSegment paramName
+    parseSegment segmentValue =
+      StaticSegment $
+        if isSegmentOptional segmentValue
+          then OptionalStaticSegment segmentValue
+          else RequiredStaticSegment segmentValue
+
+isSegmentOptional :: String -> Bool
+isSegmentOptional = isSuffixOf "?"
```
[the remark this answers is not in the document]
infomiho: Good call, I agree 👍

sodic (approved):

Issue #2202, opened 2024-07-23
On wasp db migrate-dev, if TS compilation failed, we don't do actual migration


Martinsos:
Here is related discord convo, this caused quite some confusion: https://discord.com/channels/686873244791210014/1263910702117163119/1263910702117163119 .

> Here are the steps to reproduce -- wasp v0.13.2 via [Open SaaS](https://opensaas.sh/)
> - create a saas app with `wasp new` (select 3 for `saas`)
> - `cd` into `app` and start the db. Run the initial migration with `wasp db migrate-dev`
> - in `main.wasp` define a property on `User` entity such as `myField    String?`
> - run `wasp db migrate-dev`
> - observe the TS failure stemming from `/server/scripts/usersSeed.ts` prevents you from being able to name/save the migration.

What happens is that if prisma schema is changed in such way that the new Prisma SDK is causing TS errors in the project (which is likely), we don't do actual migration.
Ok, so we probably first do prisma generate , then try to compile the Typescript project, and only then prisma migrate. I wonder if we should change this: we could run prisma migrate even if TS compilation fails. That sounds like it might be somewhat more correct / expected, and is in spirit with TS compilation philosophy. Seems that is what you also expected in this situation.

Even if not, we should make it clear in the error message what happened.

But I would suggest we revisit the `wasp db migrate-dev` and make sure we do perform actual migration even if TS compilation fails.

sodic:
## The problem
Here's what happens with `wasp db migrate-dev`:
1. The user changes the `prisma.schema` file.
2. The user runs `wasp db migrate-dev`.
3. Wasp generates the new Prisma client.
4. Wasp builds the SDK, which also builds the user code that's copied into its `ext-src` folder.
5. The SDK build fails because the old user code in `ext-src` doesn't fit the new Prisma client.
6. Wasp exists before actually migrating the database.

## Proposed solutions
@martinsos identified two possible approaches:
 1. Tolerating the TypeScript errors and doing the migration anyway.
 2. Migrating the database before generating the client.

### Tolerating TypeScript errors
I decided against this approach because:
 - Changing how we treat TypeScript errors during the SDK build would change it for every command, not just `wasp db migrate-dev`, which might have unintended consequences. I could have added a flag that tolerates TS errors only for `wasp db migrate-dev`, but it would have been too hacky IMO.
[...]

PR #2269, opened 2024-09-06
Small fixes in the generator

sodic:
Small stuff I noticed while working on #2202.


waspc/src/Wasp/AppSpec/Job.hs
```diff
@@ -49,13 +45,10 @@ data Schedule = Schedule
   }
   deriving (Show, Eq, Data)
 
-instance IsDecl Schedule
-
 -- These are optional executor-specific JSON options we pass
 -- directly through to the executor when submitting jobs.
 data ExecutorOptions = ExecutorOptions
   { pgBoss :: Maybe JSON,
-    simple :: Maybe JSON
```
sodic, line 58: We don't use this as far as I can see, and it's not documented.

waspc/src/Wasp/Project/Analyze.hs
```diff
@@ -69,8 +84,57 @@ analyzeWaspProject waspDir options = do
...
+    ExitSuccess -> do
+      otherChan <- newChan
+      (_, runExitCode) <-
+        concurrently
+          (handleRunJsMessages otherChan)
+          ( runNodeCommandAsJob
+              workingDir
+              "node"
+              [ SP.fromAbsDir workingDir ++ "node_modules/wasp-ts-sdk/dist/run.js",
+                SP.fromAbsDir workingDir ++ ".wasp/config/main.wasp.mjs"
```
Martinsos, line 112: This won't work on Win, because workingDir will have Win separators and you have posix separators on the right.
In general I recommend converting from StrongPath at the last moment possible, that will save you from mistakes like this.
So in this case:
```hs
SP.fromAbsFile $ workingDir </> [relfile|node_modules/wasp-ts-sdk/dist/run.js|],
...
```

waspc/src/Wasp/Project/Analyze.hs
```diff
@@ -69,8 +84,57 @@ analyzeWaspProject waspDir options = do
...
+  chan <- newChan
+  (_, tscExitCode) <-
+    concurrently
+      (readJobMessagesAndPrintThemPrefixed chan)
+      (runNodeCommandAsJob workingDir "npx" ["tsc", "-p", "tsconfig.node.json"] J.Wasp chan)
+
+  case tscExitCode of
+    ExitFailure _status -> return $ Left ["Error while running TypeScript compiler on the *.wasp.mts file."]
+    ExitSuccess -> do
+      otherChan <- newChan
```
Martinsos, line 104: wow `otherChan`, what happened here :D

waspc/src/Wasp/Project/Analyze.hs
```diff
@@ -69,8 +84,57 @@ analyzeWaspProject waspDir options = do
...
+              workingDir
+              "node"
+              [ SP.fromAbsDir workingDir ++ "node_modules/wasp-ts-sdk/dist/run.js",
+                SP.fromAbsDir workingDir ++ ".wasp/config/main.wasp.mjs"
+              ]
+              J.Wasp
+              otherChan
+          )
+      case runExitCode of
+        ExitFailure _status -> return $ Left ["Error while running the compiled *.wasp.mts file."]
```
Martinsos, line 118: Why not use actual file name here. You have it available. And you avoid a bit of hardcoding.

waspc/src/Wasp/Project/Analyze.hs
```diff
@@ -69,8 +84,57 @@ analyzeWaspProject waspDir options = do
...
+        ExitSuccess -> return $ Right []
```
Martinsos, line 119: No decls? Aha I guess we don't yet have that?

waspc/src/Wasp/Project/Analyze.hs
```diff
@@ -69,8 +84,57 @@ analyzeWaspProject waspDir options = do
...
+    handleRunJsMessages :: Chan J.JobMessage -> IO ()
+    handleRunJsMessages = runPrefixedWriter . processMessages
+    processMessages chan = do
+      jobMsg <- liftIO $ readChan chan
+      case J._data jobMsg of
+        J.JobOutput payload J.Stdout -> do
+          -- let payload:: String = read $ T.unpack text
+          liftIO $ putStrLn "Ovo smo dobili na stdout"
+          -- parse payload as json
+          let json = Aeson.toJSON payload
```
Martinsos, line 130: Aha this is still WIP! Sorry I thought this PR is ready for review since it wasn't draft PR.

waspc/src/Wasp/Project/Analyze.hs
```diff
@@ -118,16 +182,25 @@ constructAppSpec waspDir options packageJson parsedPrismaSchema decls = do
 
   return $ runValidation ASV.validateAppSpec appSpec
 
-findWaspFile :: Path' Abs (Dir WaspProjectDir) -> IO (Maybe (Path' Abs File'))
+data WaspFile
+  = WaspLangFile {_path :: !(Path' Abs File')}
+  | WaspTsFile {_path :: !(Path' Abs File')}
+  deriving (Show)
```
Martinsos, line 188: Interesting, feels a bit unusual. I get why you did it though, you want to be able to say it is one of them but you don't want to yet know which one. Records in sum types are usually an anti-pattern, but these records are identical except for the data constructor, so it works.

What if you want to specify somewhere that a function needs specifically path to WaspTS file? Or Sepcifically WaspLangFile? We have that knowledge but we didn't put it in StrongPath.

I guess soltuion would be
```hs
data WaspLangFile
data WaspTsFile
data WaspFile = WaspLang (Path' Abs WaspLangFile) | WaspTs (Path' Abs WaspTsFile)
```
And I wouldn't bother with `_path`, I probably won't be needing that anyway as I will most likely always want to deconstruct hm.
But ok, this is verbose a bit, and it is all quite localized so I guess it is not important. Although you do export `WaspFile (..)`, which makes it less localized hm. Ok, I am ok with this what you have, although I would be tempted to go this extra step.

Martinsos:
Ha small fixes you say, but then there is also the flow for main.wasp.ts vs main.wasp.lang!

waspc/src/Wasp/Project/Analyze.hs
sodic: Aw man, I made a git error.
The `Analyze.hs` file was never supposed to be here, must have committed it by accident.

Sorry for wasting your time, my bad!

What worries me is that you got through the entire review without noticing that the code is too horrible for something I'd send to code review.

Commenting with a straight face at all the atrocities and only realizing something was off when you encountered a string in Croatian.

Is that what you think of me, Martin? 

Anyway, I'll remove this file from the PR. Please take a look at the other files. :)

Martinsos: Hah, I believed in you!

All right, I will take another look. Maybe of some the change syou did in this file though, and that I commented on, are still present in the version of the code you are woroking on? In that case do take a look at my comments on Analyze.hs and see if you can apply them.

waspc/src/Wasp/Generator/Monad.hs
```diff
@@ -25,7 +25,7 @@ import Data.List.NonEmpty (NonEmpty, fromList)
 -- The mechanism to catch errors is only there to assist in collecting more errors, not recover.
 -- There may optionally be additional errors or non-fatal warnings logged in the State.
 newtype Generator a = Generator
-  { _runGenerator :: ExceptT GeneratorError (StateT GeneratorState Identity) a
+  { _runGenerator :: ExceptT GeneratorError (State GeneratorState) a
```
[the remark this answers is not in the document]
Martinsos: Hah, can't remember at the moment! I guess it is the same thing?

Martinsos (approved):

waspc/src/Wasp/Project/Analyze.hs
Martinsos: Or maybe this is just your sneaky way of giving up on a bad PR :D?

waspc/src/Wasp/AppSpec/Job.hs
infomiho: Origin story:
Screenshot 2024-09-16 at 15 09 21

waspc/src/Wasp/Project/Analyze.hs
infomiho: https://arewetssdkyet.com

waspc/src/Wasp/Generator/Monad.hs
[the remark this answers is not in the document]
sodic: From what I know, it is.

PR #2297, opened 2024-09-18
Userland files validation

### LINKED ISSUE #2169: Validate `tsconfig.json` and `package.json`
## Sub issues
Added by @sodic:
 - Validate `tsconfig.json` - covered by #939.
 - Validate `package.json` - no issue yet.

## Backstory

While developing Wasp's support for `schema.prisma` file we agreed that it's best if we let users write whatever they want inside and we take it as the source of truth (as opposed to injecting config values on the fly which users don't see). This made sense since the users saw the config Wasp _actually_ uses. Of course, we had to introduce validation for some fields Wasp depended on e.g. the `url` field _has to be_ `env("DATABASE_URL")` - the compiler throws an error if it isn't that value.

### Idea

Since Wasp doesn't generate the `tsconfig.json` or the `package.json` but instead it uses the files as-is, Wasp apps can break if users change the config in weird ways or the config file can become outdated with the latest versions of Wasp. For example, if we figure out the new best values for the `tsconfig.json` then the user should update that file manually. If they don't, they have a broken IDE experience _but Wasp compiler doesn't warn the user about it._

We should introduce some validation of the `tsconfig.json` and the `package.json` files:
- there should be `"wasp": "file:.wasp/out/sdk/wasp"` in the `package.json`
- `"module"`, `"target"`, `"moduleResolution"` fields in the `tsconfig.json` should have certain values
- etc.

This way we are ensuring prerequisites for the Wasp app to run successfully and making the whole dev experience more robust. Users will know if they mess something up since the compiler will tell them.

infomiho:
Closes #2169 

### Validates `package.json`

It loads the JSON and checks versions of deps: Wasp SDK, Prisma and React Router DOM.

### Validates `tsconfig.json`

It loads the `tsconfig.json` which is JSON with comments (uses Node.js's eval to get proper JSON). It checks all the fields.


waspc/src/Wasp/ExternalConfig.hs
```diff
@@ -0,0 +1,31 @@
...
+    WaspProjectDir,
+  )
+
+data ExternalConfigs = ExternalConfigs
+  { packageJson :: PackageJson,
+    tsConfig :: TsConfig
+  }
+  deriving (Show)
+
+analyzeExternalConfigs :: Path' Abs (Dir WaspProjectDir) -> IO (Either [CompileError] ExternalConfigs)
```
infomiho, line 22: I've introduced a concept of "external configs" which are userland files which Wasp depends on but doesn't control. I'm not sure if something else would go into this category?

waspc/src/Wasp/ExternalConfig/PackageJson.hs
```diff
@@ -0,0 +1,127 @@
...
+type DependenciesMap = Map PackageName PackageVersion
+
+type PackageName = String
+
+type PackageVersion = String
+
+instance FromJSON PackageJson
+
+analyzePackageJsonContent :: Path' Abs (Dir WaspProjectDir) -> IO (Either [CompileError] PackageJson)
+analyzePackageJsonContent waspProjectDir = runExceptT $ do
```
infomiho, line 56: I've used a GPT suggestion to simplify the existing code and use the [`Except` monad transformer](https://hackage.haskell.org/package/mtl-2.3.1/docs/Control-Monad-Except.html#g:2) to make the code cleaner. Let me know if this looks okay.

waspc/src/Wasp/Generator/SdkGenerator.hs
```diff
@@ -188,20 +188,21 @@ npmDepsForSdk spec =
...
             -- Todo: why is this in dependencies, should it be in dev dependencies?
             -- Should it go into their package.json
             ("@types/express-serve-static-core", "^4.17.13"),
-            ("@types/react-router-dom", "^5.3.3")
+            ("@types/react-router-dom", show reactRouterVersion)
           ]
           ++ depsRequiredForAuth spec
           ++ depsRequiredByOAuth spec
-          -- This must be installed in the SDK because it lists prisma/client as a dependency.
+          -- Server auth deps must be installed in the SDK because "@lucia-auth/adapter-prisma"
```
infomiho, line 201: Updated this comment since I talked with @sodic recently twice and always forgot what was it really talking about.

waspc/src/Wasp/Generator/Common.hs
```diff
@@ -57,6 +58,9 @@ prismaVersion :: SV.Version
 --       Then, make sure `data/Generator/templates/sdk/wasp/prisma-runtime-library.d.ts` is up to date.
 prismaVersion = SV.Version 5 19 1
 
+reactRouterVersion :: SV.ComparatorSet
+reactRouterVersion = SV.backwardsCompatibleWith $ SV.Version 5 3 3
```
infomiho, line 62: A single place to define the version which we use in the validation code.

waspc/src/Wasp/Util/Json.hs
```diff
@@ -0,0 +1,16 @@
...
+
+import Data.Aeson (FromJSON)
+import System.Exit (ExitCode (..))
+import qualified System.Process as P
+import Wasp.Util.Aeson (decodeFromString)
+
+-- | Uses Node.js to parse JSON with comments by treating it as a JavaScript object.
+parseJsonWithComments :: FromJSON a => String -> IO (Either String a)
+parseJsonWithComments jsonStr = do
+  let evalScript = "const v = " ++ jsonStr ++ ";console.log(JSON.stringify(v));"
```
infomiho, line 11: This is a trick we used to parse JSON with comment successfully. Aeson can't handle it. Other options included using the `tsc --showConfig` which returns something similar to the user's config, but it's still modified by `tsc` a bit (paths are absolute). I also considered implementing my own parser that would drop comments and just return a string we'll then parse - but the Node.js option worked so well and it's not that costly to invoke IMHO.

waspc/src/Wasp/ExternalConfig/TsConfig.hs
Martinsos, line 1: Feedback from PackageJson that is applicable here, should also be applied here.

waspc/src/Wasp/ExternalConfig/TsConfig.hs
```diff
@@ -0,0 +1,134 @@
...
+          validateRequiredFieldInCompilerOptions "typeRoots" ["node_modules/@testing-library", "node_modules/@types"] typeRoots,
+          validateRequiredFieldInCompilerOptions "outDir" ".wasp/phantom" outDir
+        ]
+
+    validateRequiredFieldInCompilerOptions fieldName expectedValue getField =
+      validateRequiredField ("compilerOptions." ++ fieldName) expectedValue $ getField compilerOptionsFields
+
+    compilerOptionsFields = compilerOptions tsConfig
+
+-- | Used to show expected values in error messages.
```
Martinsos, line 109:
```suggestion
-- | Haskell type that implements ShowJS is a type whose values can be mapped to Javascript values (their string representation).
```

waspc/src/Wasp/ExternalConfig/TsConfig.hs
```diff
@@ -0,0 +1,134 @@
...
+class ShowJs a where
```
Martinsos, line 110: Maybe you should rename this class to `IsJavascriptValue` and rename its method to `toJsValue` or something in that direction.

waspc/src/Wasp/ExternalConfig/TsConfig.hs
```diff
@@ -0,0 +1,134 @@
...
+instance ShowJs [String] where
+  showJs = show
+
+instance ShowJs Bool where
+  showJs True = "true"
+  showJs False = "false"
+
+type FieldName = String
+
+validateRequiredField :: (Eq value, ShowJs value) => FieldName -> value -> Maybe value -> [CompileError]
```
Martinsos, line 125: I mentioned this in previous PRs already, taking `Maybe` as an arugment is always a bit suspicious. You should probably rather instead handle that outside, and take just `value` here.

waspc/src/Wasp/ExternalConfig/PackageJson.hs
```diff
@@ -0,0 +1,127 @@
...
+  packageJsonFile <- ExceptT findPackageJsonFileOrError
+  packageJson <- ExceptT $ readPackageJsonFile packageJsonFile
+  ExceptT $ validatePackageJson packageJson
+  where
+    findPackageJsonFileOrError = maybeToEither [fileNotFoundMessage] <$> findPackageJsonFile waspProjectDir
+    fileNotFoundMessage = "Couldn't find the package.json file in the " ++ toFilePath waspProjectDir ++ " directory"
+
+validatePackageJson :: PackageJson -> IO (Either [CompileError] PackageJson)
+validatePackageJson packageJson =
```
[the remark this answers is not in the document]
Martinsos: Actually, it should certainly be in the Generator, I just realized you are importing stuff from Generator in this file. We try to not go in that direction, importing implementation details from the Generator, instead we only import Generator to call it, the rest of deps goes in another direction.
SO yeah, we need to move this logic to Generator.

waspc/src/Wasp/Generator/Common.hs
Martinsos: This is client specific dep though, right? So it should not be here but some lower level, under client Generator stuff? Or does SDK complicate this? But then maybe it should go under SDK? Or maybe it is ok here? OK you will know better, but wanted to check with you.

waspc/src/Wasp/Project/Analyze.hs
```diff
@@ -81,11 +77,11 @@ analyzeWaspFileContent prismaSchemaAst = return . left (map getErrorMessageAndCt
 constructAppSpec ::
   Path' Abs (Dir WaspProjectDir) ->
   CompileOptions ->
-  PackageJson ->
+  EC.ExternalConfigs ->
   Psl.Schema.Schema ->
   [AS.Decl] ->
   IO (Either [CompileError] AS.AppSpec, [CompileWarning])
-constructAppSpec waspDir options packageJson parsedPrismaSchema decls = do
+constructAppSpec waspDir options externalConfigs parsedPrismaSchema decls = do
```
Martinsos, line 84: Ok, I get now why you introduced ExternalConfigs, to make it easier to forward tihs stuff. I guess that makes sense. Maybe it should go under Wasp.Project though hm.

waspc/src/Wasp/Util/Json.hs
Martinsos: Cool! Could you add some of this GH comment to the actual comment in the code? I think whoever will be reading this code will benefit from knowing more about why this was done (Aeson can't handle it) and why it was needed.

waspc/headless-test/examples/todoApp/tsconfig.json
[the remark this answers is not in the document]
infomiho: Yep, the tests started failing because we started validation the `tsconfig.json` - we forgot to update it before. Now it's okay 👍

waspc/src/Wasp/ExternalConfig/PackageJson.hs
```diff
@@ -0,0 +1,127 @@
...
+        else [incorrectVersionMessage]
+    _rest -> case validationType of
+      RequiredPackage -> [requiredPackageMessage "dependencies"]
+      RequiredDevPackage -> [requiredPackageMessage "devDependencies"]
+      OptionalPackage -> []
+  where
+    depsToCheck = case validationType of
+      RequiredPackage -> [dependencies packageJson]
+      RequiredDevPackage -> [devDependencies packageJson]
+      -- Users can install packages that don't need to be strictly in dependencies or devDependencies
```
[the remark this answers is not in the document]
infomiho: We don't control where the user installs the dep: production or development deps. If they install some deps like `react-router-dom` either as production or development - they will mess up their project somewhat. Your name is better `HasExactVersionIfListed` -> if the user added it - make sure it has the correct version.

waspc/src/Wasp/ExternalConfig/PackageJson.hs
```diff
@@ -0,0 +1,127 @@
...
+
+    incorrectVersionMessage :: String
+    incorrectVersionMessage =
+      unwords
+        ["The", show packageName, "package must have version", show expectedPackageVersion, "in package.json."]
+
+    requiredPackageMessage :: String -> String
+    requiredPackageMessage packageJsonLocation =
+      unwords
+        [ "The",
```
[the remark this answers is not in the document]
infomiho: That's why we are using `show` to wrap it with quotes. Or you wanted specifically the single quotes?

waspc/src/Wasp/ExternalConfig/PackageJson.hs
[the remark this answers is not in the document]
Martinsos: Right, but is it possible that we care about that dep only if they install it as non-dev, or maybe dev dep? Maybe we need react-router-dom to be specific version if they install it as normal dep, but we don't care if they install it as dev-dep?

Following up on this: we require them to have `react` installed as non-dev dep with specific version. What if they also install it as dev dep with some other version? Is that something we want to throw warning upon?

Yeah this is the direction I am inquirying about: seems to me like sometimes we care if it is dev or non-dev dep, but sometimes we say "any of those, dev or non-dev", and it doesn't seem super clear to me how we are making those decisions, seems like we haven't thought it through completely. Although I get why, it is a bit confusing / hard to define, but it is probably worth giving it a try and thinking it all the way through.

waspc/src/Wasp/ExternalConfig/PackageJson.hs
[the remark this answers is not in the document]
Martinsos: Aha, ok yeah `show` does that I forgot! ALl right, all good then!

waspc/src/Wasp/ExternalConfig.hs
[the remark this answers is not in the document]
Martinsos: I left another comment where I wrote I get now why you have this data type.

waspc/src/Wasp/ExternalConfig/PackageJson.hs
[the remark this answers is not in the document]
infomiho: > Right, but is it possible that we care about that dep only if they install it as non-dev, or maybe dev dep? Maybe we need react-router-dom to be specific version if they install it as normal dep, but we don't care if they install it as dev-dep?

`npm install` doesn't care if it's a dev dep or prod dep. `npm ci` cares, but users don't use that. So for our use case "let's prevent users from messing up their Wasp project" - we need to validate both equally. 

> Following up on this: we require them to have react installed as non-dev dep with specific version. What if they also install it as dev dep with some other version? Is that something we want to throw warning upon?

That's not possible. If you try installing React as a dev dep, it will just move from deps to dev deps.

To conclude: some deps have strict location (dep vs. dev dep) and version requirements but for some the location is irrelevant due to `npm install` not caring where it's installed (it all goes to one `node_modules`) and only the version needs to be validated.

waspc/src/Wasp/Util.hs
```diff
@@ -270,3 +271,9 @@ secondsToMicroSeconds = (* 1000000)
 
 findDuplicateElems :: Ord a => [a] -> [a]
 findDuplicateElems = map head . filter ((> 1) . length) . group . sort
+
+validateToEither :: (Monad m) => (a -> m [String]) -> a -> m (Either [String] a)
```
infomiho, line 275: I'm not sure what to name this ... it takes a validation function that returns a list of errors and based on that either returns the input or the errors. 

`listValidationToEither`, `validateListToEither`, `runValidation`, `listValidationToEither`?

waspc/src/Wasp/Generator/Common.hs
Martinsos: Ok, that makes sense to me! Resolve this comment when happy.

waspc/src/Wasp/ExternalConfig/PackageJson.hs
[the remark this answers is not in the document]
Martinsos: Oooooh ok, I actually had this false idea that devDependencies are somehow better isolated from the normal ones, I get now how it works. Thanks, ok this makes sense then! Feel free to resolve this comment.

waspc/src/Wasp/Project/ExternalConfig/TsConfig.hs
```diff
@@ -0,0 +1,42 @@
...
+
+readTsConfigFile :: Path' Abs (File TsConfigFile) -> IO (Either [String] T.TsConfig)
+readTsConfigFile tsConfigFile = do
+  tsConfigContent <- IOUtil.readFileBytes tsConfigFile
+
+  parseResult <- parseJsonWithComments . BS.toString $ tsConfigContent
+
+  case parseResult of
+    Right tsConfig -> return $ Right tsConfig
+    Left err -> return $ Left ["Failed to parse tsconfig.json file: " ++ err]
```
Martinsos, line 42: You could combine these two into just one function, to make it a bit simpler, there doesn't seem to be any need to have `findTsConfigFile` as a separate step.

You could probably even completely ditch "finding" of the file and just go for the reading of it, and if there is no file, reading will fail anyway. But maybe this way we get nicer error message? Yeah, that might be so, not sure if worth it though.

I am just looking how to simplify this file, it seems like it has a lot of code but doesn't do much. But maybe that is ok.

waspc/src/Wasp/Util/Json.hs
```diff
@@ -0,0 +1,18 @@
+module Wasp.Util.Json (parseJsonWithComments) where
+
+import Data.Aeson (FromJSON)
+import System.Exit (ExitCode (..))
+import qualified System.Process as P
+import Wasp.Util.Aeson (decodeFromString)
+
+-- | Uses Node.js to parse JSON with comments by treating it as a JavaScript object.
+-- We used this technique because Aeson can't handle it and we didn't want to write
```
Martinsos, line 9:
```suggestion
-- We use this technique because Aeson can't read JSON with comments and we didn't want to write
```

waspc/src/Wasp/Generator/ExternalConfig/PackageJson.hs
```diff
@@ -0,0 +1,56 @@
...
+  ( validatePackageJson,
+  )
+where
+
+import qualified Data.Map as M
+import qualified Wasp.ExternalConfig.PackageJson as P
+import Wasp.Generator.Common (prismaVersion)
+import Wasp.Generator.WebAppGenerator.Common (reactRouterVersion)
+
+validatePackageJson :: P.PackageJson -> IO [String]
```
Martinsos, line 11: Let's maybe add `type ErrorMsg = String` here.

Martinsos, line 11: Looking at it, this actually isn't `IO` method, is it? It does only pure calculations. So why do we have it return `IO`? Is it a leftover / mistake?

waspc/src/Wasp/Generator/ExternalConfig/TsConfig.hs
```diff
@@ -0,0 +1,56 @@
+{-# LANGUAGE FlexibleInstances #-}
+
+module Wasp.Generator.ExternalConfig.TsConfig
+  ( validateTsConfig,
+  )
+where
+
+import qualified Wasp.ExternalConfig.TsConfig as T
+
+validateTsConfig :: T.TsConfig -> IO [String]
```
Martinsos, line 10: Same thing like for validatePackageJson, if I am correct this doesn't have to be `IO`.

waspc/src/Wasp/Project/ExternalConfig/PackageJson.hs
```diff
@@ -0,0 +1,37 @@
...
+    packageJsonInWaspProjectDir,
+  )
+import Wasp.Util (validateToEither)
+import qualified Wasp.Util.IO as IOUtil
+
+analyzePackageJsonFile :: Path' Abs (Dir WaspProjectDir) -> IO (Either [String] P.PackageJson)
+analyzePackageJsonFile waspProjectDir = runExceptT $ do
+  packageJsonFile <- ExceptT findPackageJsonFileOrError
+  packageJson <- ExceptT $ readPackageJsonFile packageJsonFile
+  ExceptT $ validateToEither validatePackageJson packageJson
```
Martinsos, line 26: I was looking at this call of validateToEither here to figure out if we might not need it, since it seemed to me it is oddly specific function and we could probably replace it with something nicer.

I was looking to do something like following instead:
```hs
ExceptT $ <sometransformation> $ validatePackageJson packageJson
```
So `validatePackageJson packageJson` gives us `IO [String]`. If this array is not empty , we want to shortcircuit the computation with this array sinced it is a list of errors. We could do this with following function:
```hs
\errors -> if null errors then Right () else Left errors
```

I guess we could name it like this:
```hs
errorsListToEither [] = Right ()
errorsListToEither _ = Left errors
```

And then we map it:
```hs
ExceptT $ errorsListToEither <$> validatePackageJson packageJson
```
actually, fmap it, since it doesn't operate in `IO`.

Although, I noticed in another comment that `validatePackageJson packageJson` doesn't even have to be `IO`, so that will also change things here, but I think the idea will remain the same.

Another approach which I have only a rough hint it might work but you can give it a try and see if it makes sense or not, buit I think `ExceptT` might be instance of ErrorMonad (or is it MonadError?), therefore supporting stuff like `throw` (or is it `throw<something>`?) and `try` or something like it. If that is so, you might be able to just do something like
```hs
case validatePackageJson packageJson of
  [] -> pure packageJson
  errors -> throwError errors
```
I assumed here that you made `validatePackageJson` to not be `IO`.

Although `throwError` here you could probably just replace with `ExceptT . pure . Left`, that is what it will be doing in the background I believe.

Martinsos (commented):
@infomiho looking good! I think all the big stuff is done, there is a couple of smaller things left that I commented on, but major stuff is good.

waspc/src/Wasp/ExternalConfig/PackageJson.hs
```diff
@@ -11,15 +19,21 @@ import qualified Wasp.AppSpec.App.Dependency as D
...
 getDevDependencies :: PackageJson -> [Dependency]
 getDevDependencies packageJson = D.fromList $ M.toList $ devDependencies packageJson
+
+type DependenciesMap = Map PackageName PackageVersion
+
+type PackageName = String
+
+type PackageVersion = String
+
+instance FromJSON PackageJson
```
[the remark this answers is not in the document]
infomiho: It was `{-# LANGUAGE DeriveAnyClass #-}` :)

waspc/src/Wasp/Generator/ExternalConfig/PackageJson.hs
```diff
@@ -0,0 +1,56 @@
...
+        validate ("react-router-dom", show reactRouterVersion, HasExactVersionIfListed)
+      ]
+  where
+    validate = validateDep packageJson
+
+data PackageValidationType = IsListedWithExactVersion | IsListedAsDevWithExactVersion | HasExactVersionIfListed
+
+validateDep :: P.PackageJson -> (P.PackageName, P.PackageVersion, PackageValidationType) -> [String]
+validateDep packageJson (packageName, expectedPackageVersion, validationType) = case validationType of
+  IsListedWithExactVersion -> checkDeps [P.dependencies packageJson] [requiredPackageMessage "dependencies"]
```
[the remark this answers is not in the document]
infomiho: That's a nice change, cool! It reads nicer as well - package name + version as a tuple and then an another param.

waspc/src/Wasp/Project/ExternalConfig/TsConfig.hs
infomiho: I feel like it's simple enough - the finding of the file produces a `Maybe` and this way I can handle it at the top level vs. handling it inside of `readTsConfigFile` 🤷

waspc/src/Wasp/Project/ExternalConfig/PackageJson.hs
infomiho: I went with something super simple:

```haskell
case validateTsConfig tsConfig of
  [] -> return tsConfig
  errors -> throwError errors
  ```

Thanks for the tip, it makes sense to inline it.

waspc/src/Wasp/Util.hs
infomiho: Inlined the function 👍

waspc/src/Wasp/Project/ExternalConfig/TsConfig.hs
Martinsos: Ok, let's leave it then!

waspc/src/Wasp/Generator/ExternalConfig/PackageJson.hs
Martinsos: One thought on top of this: often, in our code, we just use Strings as error messages. General advice is that this is not good, you should use more structured types, but the truth is, for our needs this seems to be right at the moment.
However, it is still much nicer to see `ErrorMsg` in output type than `String` because it tells me what it is, which is why `type ErrorMsg = String`.
Now, sometimes it is nice to call it more specifically: `type CompilerError = String`, or `type ParsingError = String`, but in many parts of the code we don't want to bother iwth that and just doing `type ErrorMsg = String` is enough. But that means we have a lot of places where we do the same thing: define `type ErrorMsg = String`. So I wonder if we should just have central `type ErrorMsg = String` in some `Wasp.Utils` or something like that? We don't have to do this in this PR, but wanted to mention this now since I was talking about this. What do you think?

waspc/src/Wasp/Project/ExternalConfig/PackageJson.hs
Martinsos: Ok perfect! I wrote a lot above in the comment in case train of thought is useful, but yeah this is the optimal solution at the end.

Martinsos (approved):
@infomiho I think this is it! Left couple of small comments but none require any changes I believe, give them a quick look and them merge when happy.

waspc/src/Wasp/Generator/ExternalConfig/PackageJson.hs
infomiho: I'll turn this line of thinking into an issue 👍

infomiho: https://github.com/wasp-lang/wasp/issues/2310

PR #2299, opened 2024-09-18
TS SDK Setup

sodic:
## Notes for reviewers
You can ignore the todos (they're mostly notes for me and I'll soon be moving them somewhere else).
You can comment on them if you want, but it isn't necessary.

The `wasp ts-setup` command is unfinished (more on that below). We can discuss how functional it has to be before the release.

## Context
This PR is complementary to #2276.

It contains the changes that enable the user to create and work in a TS SDK project that looks like this:
```
├── main.wasp.mts         // Wasp config file
├── package.json
├── package-lock.json
├── public
├── schema.prisma
├── src
├── tsconfig.src.json     // tsconfig for src
├── tsconfig.json         // Composite  tsconfig
├── tsconfig.wasp.json    // tsconfig for the Wasp config file
├── vite.config.ts
└── node_moduels
    ├── ...
    └── wasp-config       // The package for the TS sdk
``` 
Those changes include:
1. The `wasp-config` packages that servers as an API for both our users and `waspc`.
3. A `wasp ts-setup` command that sets up the project to work with a config in TypeScript (TS SDK).

It's just the blue parts from this graph plus the prep work necessary for making the whole thing functional. The idea is to review this PR by **treating the purple parts as a black box** (that part is covered in #2276).

<p align="center">
    
 </p>


waspc/packages/wasp-config/src/appSpec.ts
```diff
@@ -0,0 +1,210 @@
+/** This module is a mirror implementation of AppSpec Decls in TypeScript.
+ * The original implemention is in Haskell (waspc).
+ *
+ * IMPORTANT: Do not change this file without updating the AppSpec in waspc.
```
sodic, line 4: I'll add this comment there too.

It would be nice if AppSpec came from a single specification independent of TS and HS.

waspc/packages/wasp-config/src/appSpec.ts
```diff
@@ -0,0 +1,210 @@
...
+  | { declType: 'App'; declName: string; declValue: App }
+  | { declType: 'Page'; declName: string; declValue: Page }
+  | { declType: 'Route'; declName: string; declValue: Route }
+  | { declType: 'Query'; declName: string; declValue: Query }
+  | { declType: 'Action'; declName: string; declValue: Action }
+  | { declType: 'App'; declName: string; declValue: App }
+  | { declType: 'Job'; declName: string; declValue: Job }
+  | { declType: 'Api'; declName: string; declValue: Api }
+  | { declType: 'ApiNamespace'; declName: string; declValue: ApiNamespace }
+  | { declType: 'Crud'; declName: string; declValue: Crud }
```
sodic, line 17: This looks a little repetitive, but removing the repetition with a complicated generic isn't worth it IMO.

waspc/packages/wasp-config/src/mappers.ts
```diff
@@ -0,0 +1,443 @@
...
+    ),
+  })
+
+  for (const [actionName, actionConfig] of actions.entries()) {
+    decls.push({
+      declType: 'Action',
+      declName: actionName,
+      declValue: mapOperationConfig(actionConfig, parseEntityRef),
+    })
+  }
```
sodic, line 75: These for loops are also repetitive, and I'll probably rewrite them as a generic map, and add a compile-time check making sure I've covered all the options.

But let's release it first and then deal with that.

waspc/packages/wasp-config/src/_private.ts
sodic, line 1: I'm not a fan of a file with a single constant, but it's the only solution to package-private method that I could think of (see the comment above `App#getSpec`.

waspc/packages/wasp-config/src/run.ts
sodic, line 1: This is what we call from `waspc`.

waspc/packages/wasp-config/src/index.ts
sodic, line 1: This is the file that the user imports.

I could have exposed `userApi` directly, but doing so through `index.ts`:
1. Follows the convention from `NodePacakgeFFI`.
2. Makes it more explicit/expected.

waspc/src/Wasp/Generator/NpmInstall.hs
```diff
@@ -1,5 +1,6 @@
 module Wasp.Generator.NpmInstall
-  ( installNpmDependenciesWithInstallRecord,
+  ( installProjectNpmDependencies,
```
sodic, line 2: I had to export this and kind of break encapsulation. @Martinsos that cool?

waspc/packages/wasp-config/src/appSpec.ts
Martinsos: We could generate it from Haskell! Shouldn't do it now, but it should be doable in the future. That said, I am not sure if that is smart or not. We should have our hands free here in TS to do as much type level stuff as we want and to also adjust things as we like. So ok, not all should be genreated. But maybe a small part of it should be generated, enough to ensure that two things are in sync. Which maybe is this very file at the end, yeah all right.

waspc/packages/wasp-config/src/index.ts
Martinsos: Yeah, makes sense to me.

waspc/packages/wasp-config/src/mappers.ts
Martinsos: Sure, I in general just skimmed this file, let's release and then we can work on making it perfect. I am sure we can work on naming, code itself, readability (splitting it up a bit), ... .

waspc/packages/wasp-config/src/mappers.ts
```diff
@@ -0,0 +1,443 @@
...
+): AppSpec.Query
+function mapOperationConfig(
+  config: User.ActionConfig,
+  parseEntityRef: RefParser<'Entity'>
+): AppSpec.Action
+function mapOperationConfig(
+  config: User.ActionConfig | User.QueryConfig,
+  parseEntityRef: RefParser<'Entity'>
+): AppSpec.Action | AppSpec.Query {
+  // TODO: How to make sure I've destructured everything?
```
Martinsos, line 135: Interesting, I was asking myslf the same asme thing while reading this. I guess we use `...rest` and then somehow say that its type has to be empty array or something like taht?

waspc/packages/wasp-config/src/mappers.ts
Martinsos, line 1: Not the best file name :D! But we can work on name, content and all after the release, as you said.

waspc/packages/wasp-config/src/_private.ts
Martinsos: I couldn't quite get how is this useful / important?

waspc/packages/wasp-config/src/run.ts
```diff
@@ -0,0 +1,69 @@
...
+function parseEntityNamesJson(entitiesJson: string): string[] {
+  const entities = JSON.parse(entitiesJson)
+  if (!Array.isArray(entities)) {
+    throw new Error('The entities JSON must be an array of entity names.')
+  }
+  return entities
+}
+
+function serialize(appConfig: Decl[]): string {
+  return JSON.stringify(appConfig)
```
Martinsos, line 68: Ok, so Decl's are implemented exactly in such way that they match FromJSON of Haskell's AppSpec? We should mention that in the appSpec.ts .

waspc/packages/wasp-config/src/appSpec.ts
```diff
@@ -0,0 +1,210 @@
+/** This module is a mirror implementation of AppSpec Decls in TypeScript.
```
Martinsos, line 1: More specifically, they mirror the implementation of FromJSON for AppSpec Decls in Haskell, right?

waspc/packages/wasp-config/src/run.ts
Martinsos, line 1: Just skimmed this. Found it kind of hard to read, I feel like we could improve it likely. But good for now I guess!

waspc/packages/wasp-config/src/userApi.ts
```diff
@@ -0,0 +1,278 @@
+/** This module defines the user-facing API for defining a Wasp app.
+ */
+import * as AppSpec from './appSpec.js'
+import { GET_USER_SPEC } from './_private.js'
+
+export class App {
+  #userSpec: UserSpec;
+
+  // NOTE: Using a non-public symbol gives us a pacakge-private property.
```
Martinsos, line 9: Why is this important?

waspc/packages/wasp-config/src/userApi.ts
```diff
@@ -0,0 +1,278 @@
...
+  cruds: Map<string, Crud>
+  db?: DbConfig
+  emailSender?: EmailSenderConfig
+  jobs: Map<string, JobConfig>
+  pages: Map<string, PageConfig>
+  queries: Map<string, QueryConfig>
+  routes: Map<string, RouteConfig>
+  server?: ServerConfig
+  websocket?: WebsocketConfig
+}
```
Martinsos, line 278: So some stuff was left in app, like head, title, version, but some stuff got taken out, like auth, client, server, ... . How is that? That is in order to be able to define these separately so they can reference other stuff? We should probably figure out how to best draw the line there, so it is not just "does this reference something else or not" but instead it should also be logical from the user's side.

waspc/packages/wasp-config/tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+    "noImplicitOverride": true,
+
+    "module": "NodeNext",
+    "outDir": "dist",
+    "declarationMap": true,
+
+    "declaration": true,
+    "lib": ["es2022"]
+  },
+   // better structure in output
```
Martinsos, line 23: Maybe you could elaborate this one a bit more.

waspc/src/Wasp/Generator/NpmInstall.hs
Martinsos: Yeah, I guess it is fine! Maybye put it as second in the list of exported stuff.
Also, you can always make the move of creating Wasp.Generator.NpmInstall.Internal and putting it there, if you do want to make it clear it is an internal thing but still allow usage if somebody really knows what they are doing. Maybe you don't have to do this here though.

waspc/src/Wasp/NodePackageFFI.hs
```diff
@@ -60,22 +69,30 @@ scriptInPackageDir = [relfile|dist/index.js|]
 -- If the package does not have its dependencies installed yet (for example,
 -- when the package is run for the first time after installing Wasp), we install
 -- the dependencies.
-getPackageProcessOptions :: Package -> [String] -> IO P.CreateProcess
+-- TODO: How would it not have npm dependencies installed if we always to it in
+-- install_packages_to_data_dir.sh?
```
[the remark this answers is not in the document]
Martinsos: Hm can't remember -> does `install_packages_to_data_dir.sh` also install their npm deps? And who would anyway install their npm deps in release version of wasp CLI -> they don't come packaged with wasp together with their node_modules, right?

waspc/src/Wasp/NodePackageFFI.hs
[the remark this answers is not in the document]
Martinsos: Hm do they really share any functionality? I am looking at it, and maybe I miseed sometihng, but it doesnt' seem to me like "installablePackage" has much functionality really. The only tihng it can do is returns its path, right?

So far, it was Node packages that got shipped with Wasp and you could call them from Wasp CLI Haskell code. ok.

Now you added a node package that gets shipped with Wasp, so that part is the same, but the is not that we call it from Wasp's Haskell code, instead we just have it there and it is used from somewhere else (in this case from the main.wasp.mts right?). Althought wait, we do run it, don't we? We run `run.ts`? So is not also runnable in this case, in a way?

Yeah I think I would certainly split it into two modules if there is ont much shared functinality. And I would try to explain it in the comment. I am not sure what is the best name, is "installable" right. Arent they all installable? Or no because these others can't be used from user's JS code? We need to figure out there "true nature" :D so we can be sure about the names.

waspc/src/Wasp/NodePackageFFI.hs
```diff
@@ -34,6 +36,8 @@ data Package
     PrismaPackage
   | WaspStudioPackage
 
+data InstallablePackage = WaspConfigPackage
```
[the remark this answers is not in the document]
Martinsos: Yeah I think we will need some more comments, but I explained above. We need to make it clear what the difference is, what is what.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/Templating.hs
```diff
@@ -43,8 +44,16 @@ replaceTemplatePlaceholdersInPackageJsonFile appName projectName projectDir =
...
-  updateFileContentWith (replacePlaceholders waspTemplateReplacements)
+replaceTemplatePlaceholdersInFileOnDisk appName projectName file = do
+  waspConfigPackagePath <- getPackageInstallationPath WaspConfigPackage
+  let waspTemplateReplacements =
+        [ ("__waspConfigPath__", waspConfigPackagePath),
+          ("__waspAppName__", show appName),
+          ("__waspProjectName__", show projectName),
+          ("__waspVersion__", defaultWaspVersionBounds)
+        ]
+  -- TODO: We do this in all files, but not all files have all placeholders
```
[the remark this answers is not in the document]
Martinsos: But this how it was before also, right? Yeah, that is fine then. Btw this is afunction that is called upon a file, logic that decided which files this will be called upon is not in this function, so this comment should be referencing some other peice of code I guess, that I don't see from here at the moment.

Martinsos (commented):
Ok, did a quick shallow review, let's take of stuff that is important for release, and the rest we can do after it!

waspc/src/Wasp/NodePackageFFI.hs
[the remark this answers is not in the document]
sodic: Turns out that install_packages_to_data_dir doesn't really install dependencies, it just verifies that the build works. Document this properly.

waspc/cli/src/Wasp/Cli/Command/TsConfigSetup.hs
```diff
@@ -0,0 +1,28 @@
+module Wasp.Cli.Command.TsConfigSetup (tsConfigSetup) where
+
+import Control.Concurrent (newChan)
+import Control.Monad.IO.Class (liftIO)
+import Wasp.Cli.Command (Command, require)
+import Wasp.Cli.Command.Require (InWaspProject (InWaspProject))
+import Wasp.Generator.NpmInstall (installProjectNpmDependencies)
```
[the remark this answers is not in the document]
sodic: I will add a comment.

waspc/data/Cli/templates/basic/package.json
```diff
@@ -5,6 +5,7 @@
     "react": "^18.2.0"
   },
   "devDependencies": {
+    "wasp-config": "file:__waspConfigPath__",
```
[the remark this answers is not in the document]
sodic: This has been removed since.

waspc/cli/src/Wasp/Cli/Command/TsConfigSetup.hs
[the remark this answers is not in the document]
sodic: This has been removed since.

waspc/packages/wasp-config/src/appSpec.ts
```diff
@@ -0,0 +1,210 @@
...
+  emailVerification: EmailVerificationConfig
+  passwordReset: PasswordResetConfig
+}
+
+export type EmailSender = {
+  provider: EmailProvider
+  defaultFrom?: EmailFromField
+}
+
+// TODO: duplication
```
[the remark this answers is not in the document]
sodic: No longer duplication

waspc/packages/wasp-config/src/userApi.ts
sodic: IDE completion, discussed in person and added a comment.

waspc/cli/exe/Main.hs
```diff
@@ -51,6 +52,7 @@ main = withUtf8 . (`E.catch` handleInternalErrors) $ do
         ["start"] -> Command.Call.Start
         ["start", "db"] -> Command.Call.StartDb
         ["clean"] -> Command.Call.Clean
+        ["ts-setup"] -> Command.Call.TsSetup
```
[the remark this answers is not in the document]
sodic: Covered here: https://github.com/wasp-lang/wasp/issues/2390

waspc/cli/exe/Main.hs
```diff
@@ -102,6 +104,7 @@ main = withUtf8 . (`E.catch` handleInternalErrors) $ do
     Command.Call.Start -> runCommand start
     Command.Call.StartDb -> runCommand Command.Start.Db.start
     Command.Call.Clean -> runCommand clean
+    Command.Call.TsSetup -> runCommand tsConfigSetup
```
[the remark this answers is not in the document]
sodic: Tracked here: https://github.com/wasp-lang/wasp/issues/2391

waspc/cli/src/Wasp/Cli/Command/TsConfigSetup.hs
```diff
@@ -0,0 +1,28 @@
...
+
+-- | Prepares the project for using Wasp's TypeScript SDK.
+tsConfigSetup :: Command ()
+tsConfigSetup = do
```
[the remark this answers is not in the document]
sodic: Tracked here: https://github.com/wasp-lang/wasp/issues/2390

waspc/cli/src/Wasp/Cli/Command/TsConfigSetup.hs
```diff
@@ -0,0 +1,28 @@
...
+  -- - Edit package.json to contain the SDK package
+  -- - Adapt TSconfigs to fit with the TS SDK project structure
+  liftIO $
+    -- NOTE: We're only installing the user's package.json dependencies here
+    -- This is to provide proper IDE support for users working with the TS SDK
+    -- (it needs the `wasp-config` package).
+    -- Calling this function here shouldn't break anything for later
+    -- installations.
+    -- TODO: What about doing this during Wasp start? Can we make Wasp start
+    -- pick up whether the user wants to use the TS SDK automatically?
```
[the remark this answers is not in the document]
sodic: Tracked here: https://github.com/wasp-lang/wasp/issues/2390

waspc/src/Wasp/NodePackageFFI.hs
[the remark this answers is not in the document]
sodic: We discussed this in person and agreed that they indeed share functionality (setup and installing dependencies), so we decided to keep them as is.

I'll add comments that explain the difference.

waspc/src/Wasp/Generator/NpmInstall.hs
sodic: This has been removed since.

waspc/packages/wasp-config/src/mappers.ts
sodic: I solved the root issue in another way, we discussed it in person: https://github.com/wasp-lang/wasp/pull/2393/files#diff-0e8a94a09bb3cfb7a0d8ee9d0e1779d2ae6e871002a542711326f56e68707d0eR243

waspc/packages/wasp-config/src/mappers.ts
sodic: Changed it: https://github.com/wasp-lang/wasp/pull/2393/files#diff-c1eeb9771e30edfc872c9b4737350fe2a7c4f30a947ad450d9423e4a9e80657aR6

waspc/packages/wasp-config/src/run.ts
sodic: I improved it, please check and tell me what you don't like: https://github.com/wasp-lang/wasp/pull/2393#discussion_r1863726625

waspc/packages/wasp-config/tsconfig.json
sodic: I shouldn't have actually even added this comment. This option just points to "where the source is", so this is a completely normal use case. 

I've now removed it.

waspc/packages/wasp-config/src/userApi.ts
sodic: > We should probably figure out how to best draw the line there, so it is not just "does this reference something else or not" **but instead it should also be logical from the user's side.**

Yeah, that's what I tried to do.

`auth` is the only part that had to be extracted due to technical limitations, but we also wanted to extract it (because it made sense for the user and the app declaration was crowded). The rest (client, server) were extracted purely because it made sense for the user.

Martinsos: OK, it kind of makes sense to me, but if you had to explain it, how would you explain what goes into `app` and what is a standalone thing? How are those decisions being made? Let's say I want to add a new feature that is candidate for `app`, how will I decide?

waspc/packages/wasp-config/src/mappers.ts
sodic: Took care of this in https://github.com/wasp-lang/wasp/pull/2393

waspc/packages/wasp-config/src/userApi.ts
sodic: Let's say the rules are:
- Everything an app can have more of is its own declaration (servers, clients, databases).
- Auth has to be its own declaration because of technical limitations :new_moon_with_face:

PR #2300, opened 2024-09-20
add new blog post

mmaksimovic:
Let's add another blog post :)


web/blog/2024-09-24-why-your-emails-arent-getting-delivered-and-how-to-fix-this-problem.md
```diff
@@ -0,0 +1,71 @@
+---
+title: "Why Your Emails Aren't Getting Delivered and How to Fix This Problem"
```
matijaSos, line 2: At first, just reading the title, I thought this is about emails I'm sending personally or marketing emails. But actually, its transcational emails that my SaaS sends, right? Since SaaS builders are our audience, can we make the title more targeted for them?

E.g. "Why your SaaS's transactional emails aren't getting delivered", or "How to improve your the email delivery rate for your new SaaS with a fresh domain", sth along these lines.

web/blog/2024-09-24-why-your-emails-arent-getting-delivered-and-how-to-fix-this-problem.md
```diff
@@ -0,0 +1,71 @@
...
+tags: [webdev, tips, emails]
+---
+import ReactPlayer from 'react-player'
+import Link from '@docusaurus/Link';
+import useBaseUrl from '@docusaurus/useBaseUrl';
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+import ImgWithCaption from './components/ImgWithCaption'
+
+If you’re new to [Wasp](https://wasp-lang.dev/), and you’ve just built and deployed your app, it’s possible that your emails aren’t reaching your users. Don’t be too alarmed. This is a common issue that affects many developers who have newly registered domains.
```
matijaSos, line 14: The advice below isn't specific for Wasp, right? I'd then also make the intro more general, for any SaaS builder / indie hacker. I'd start with something like a problem statement, e.g. "You've been hacking at your newest SaaS idea for a couple of weeks and you finally released it to the world. But, soon, you start getting complaints - the transactional emails your app sends (e.g. for signup or X/Y/Z) aren't getting delievered! Do not worry - that's a common problem ..."

matijaSos: What was the problem that Wasp users encountered? We can use that use case probably.

matijaSos (commented):
Good stuff! This would be a SEO intented piece primarily, right? We might even link to it from our docs where it makes sense.

My main idea/remark here is to make it more tailored for our audience, which are SaaS builders / indie hackers. in this case. They will have a fresh domain for their product and will be sending out both transactional and marketing emails.

web/blog/2024-09-24-why-your-emails-arent-getting-delivered-and-how-to-fix-this-problem.md
matijaSos: After reading the rest, I see it's not only transactinoal emails. But I'd scope it for SaaS builder crowd.

web/blog/2024-09-24-why-your-emails-arent-getting-delivered-and-how-to-fix-this-problem.md
```diff
@@ -0,0 +1,71 @@
...
+
+In other words, your users and their email servers pulled a Gandalf move on you. No worries, this is something you can fix!
+
+![https://media2.giphy.com/media/WpD30tFjzosDn7amXq/giphy.gif?cid=7941fdc68jh5kd0zboo37lcp48knhd8bewokjgz7kc9juwov&ep=v1_gifs_search&rid=giphy.gif&ct=g](https://media2.giphy.com/media/WpD30tFjzosDn7amXq/giphy.gif?cid=7941fdc68jh5kd0zboo37lcp48knhd8bewokjgz7kc9juwov&ep=v1_gifs_search&rid=giphy.gif&ct=g)
```
infomiho, line 18: FYI it would be better to put some description inside of `[]` to make our blog more accessible for screen readers.

infomiho: Also, let's check if Giphy allows direct linking to gifs - if not, download them and link them locally :)

mmaksimovic: Oh yeah, they actually encourage embedding :) 

If you take it away from their platform it counts as taking it away from them. 

The Gandalf one is owned by HBO for example :)

web/blog/2024-09-24-why-your-emails-arent-getting-delivered-and-how-to-fix-this-problem.md
mmaksimovic: Yup, you're right, I'll clarify the use case we had and make the first few lines more neutral :)

web/blog/2024-09-24-why-your-emails-arent-getting-delivered-and-how-to-fix-this-problem.md
mmaksimovic: Why Your SaaS Emails Aren’t Reaching Users and How to Fix This Issue

web/blog/2024-09-24-why-your-emails-arent-getting-delivered-and-how-to-fix-this-problem.md
infomiho: Good to know!

PR #2301, opened 2024-09-20
Wasp 0.15.0 migration docs

infomiho:
- Moves migrations to a separate category in the sidebar and Markdown files into a separate folder
- Collapses the migration and misc categories by default
- Adds migration instructions from 0.14.X to 0.15.X

Left to do:
- [x] Test the migration docs after 0.15.0 RC
- [x] Update docs
  - [x] Add `NavLink` - `isActive` is deprecated (we use it in Open SaaS)
- [x] Users need to remove `@types/react-router-dom`: it's irrelevant and incorrect - but it doesn't mess up the app. Should we add validation rules for that?


web/docusaurus.config.js
```diff
@@ -280,6 +282,18 @@ module.exports = {
             from: '/docs/tutorials/todo-app',
             to: '/docs/tutorial/create',
           },
+          {
```
Martinsos, line 285: Nice job on grouping migrations in top level category and collapsing it by default!

Martinsos (approved):
LGTM!

PR #2303, opened 2024-09-24
Official Mailgun SDK + EU host support

### LINKED ISSUE #2272: Problem with using Mailgun and EU region
**Describe the bug**
I'm using mailgun and eu region which forces me to change host for the API and it seems like that this is not possible to setup otherwise I'm getting Forbidden error.

From documentation:
```
host - the mailgun host (default: 'api.mailgun.net'). Note that if you are using the EU region the host should be set to 'api.eu.mailgun.net'
```
https://github.com/paralect/mailgun-js

**Expected behavior**
Would be nice to be able to define it as env property and pass ti to mailgun

infomiho:
Closes #2272 

This PR:
- uses the official Mailgun SDK (we didn't use it before because it didn't have Typescript support)
- adds a new env var `MAILGUN_API_URL` to allows users to set their API url to the EU variant if they need

### Left to do

- [x] Update docs

### DIFF
--- a/waspc/data/Generator/templates/sdk/wasp/server/email/core/providers/mailgun.ts
+++ b/waspc/data/Generator/templates/sdk/wasp/server/email/core/providers/mailgun.ts
@@ -1,4 +1,4 @@
-import { NodeMailgun } from "ts-mailgun";
+import Mailgun from 'mailgun.js';
 import { getDefaultFromField } from "../helpers.js";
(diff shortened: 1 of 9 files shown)


Martinsos (approved):
LGTM!

Martinsos:
TODO: Update ChangeLog?

infomiho:
I'll update the Changelog in #2301

PR #2311, opened 2024-09-27
Validation for @default attribute on userEntity

### LINKED ISSUE #2308: Ensure `userEntity` has `autogenerate` on the `@id` field
Our code expects to see a model like this:
```prisma
model User {
}
```

We are validating the `@id` attribute - but not the `@default` attribute: https://github.com/wasp-lang/wasp/blob/5a79eca5c2362138a01211b8e22b9a2bb95d3e35/waspc/src/Wasp/AppSpec/Valid.hs#L131

We should validate it because users might not put it: https://ptb.discord.com/channels/686873244791210014/1288908918315024468/1289161033499414591

infomiho:
Closes #2308

### DIFF
--- a/waspc/src/Wasp/AppSpec/Entity.hs
+++ b/waspc/src/Wasp/AppSpec/Entity.hs
@@ -7,8 +7,6 @@ module Wasp.AppSpec.Entity
     getPslModelBody,
     getIdField,
     getIdBlockAttribute,
-    isFieldUnique,
-    -- only for testing:
     doesFieldHaveAttribute,
   )
 where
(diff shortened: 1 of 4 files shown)


waspc/src/Wasp/AppSpec/Entity.hs
```diff
@@ -54,18 +52,15 @@ getPslModelBody = pslModelBody
 getIdField :: Entity -> Maybe Psl.Model.Field
 getIdField = findIdField . getPslModelBody
 
-isFieldUnique :: String -> Entity -> Maybe Bool
```
infomiho, line 57: It was unused.

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -128,12 +129,20 @@ validateUserEntity spec =
   case App.auth (snd $ getApp spec) of
     Nothing -> []
     Just auth ->
-      [ GenericValidationError $ "Entity '" ++ userEntityName ++ "' (referenced by app.auth.userEntity) must have an ID field (specified with the '@id' attribute)"
-        | isNothing idFieldType
-      ]
+      concat
```
infomiho, line 132: WIP: I'll rewrite this to be nicer

waspc/src/Wasp/AppSpec/Entity.hs
```diff
@@ -54,18 +50,5 @@ getPslModelBody = pslModelBody
...
-isFieldUnique fieldName = doesFieldHaveAttribute fieldName "unique"
-
-doesFieldHaveAttribute :: String -> String -> Entity -> Maybe Bool
```
infomiho, line 60: This was a weirdly specific function which I tried to rewrite but in the end it was easier just to use `doesPslFieldHaveAttribute` directly in the one place where we needed to check if a field has an attribute.

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -128,14 +129,18 @@ validateUserEntity spec =
...
     Just auth ->
-      [ GenericValidationError $ "Entity '" ++ userEntityName ++ "' (referenced by app.auth.userEntity) must have an ID field (specified with the '@id' attribute)"
-        | isNothing idFieldType
-      ]
+      case Entity.getIdField userEntity of
+        Nothing -> [userEntityMissingIdFieldError]
+        Just idField ->
+          if Psl.Util.doesPslFieldHaveAttribute "default" idField
+            then []
+            else [userEntityIdFieldMissingDefaultAttrError]
```
Martinsos, line 137: Nice! Might be tempting to use Either here as a monad? Btw there is even this thing: https://hackage.haskell.org/package/validation .
But probably just not to complicate, it is simple enough for now, I am writing this just to give ideas for the future!

Martinsos (approved):
LGTM! Left one tiny comment.

PR #2316, opened 2024-10-02
Cleanup & bump some deps

infomiho:
This is a hygiene PR:
- I went through all deps where we didn't have the `^` (backwards compatible) version defined
- Removed from web app since not used (things moved to SDK):
  - `react-hook-form`
  - `mitt`
  - `superjson`
 - Removed from server since not used:
    - `rate-limiter-flexible`
    - `standard`
- Removed `uuid` from websockets deps (not used)
- Updated versions:
  - `express`: `4.18.1` -> `4.21.0`, [some security fixes, deps updates, cleanup](https://github.com/expressjs/express/releases)
  - `superjson`: `1.12.1` -> `2.2.1`, major bump due to Node.js 16 being min. required version - this is OK with Wasp
 - Allowed `dotenv` version bumps
 - Using a single source of truth for deps which versions are repeated twice (SDK + either web app or server)
 - Tested with all example apps.


waspc/src/Wasp/Generator/SdkGenerator/WebSocketGenerator.hs
```diff
@@ -53,6 +53,7 @@ genWebSocketProvider spec = return $ C.mkTmplFdWithData [relfile|client/webSocke
     tmplData = object ["autoConnect" .= map toLower (show shouldAutoConnect)]
 
 depsRequiredByWebSockets :: AppSpec -> [AS.Dependency.Dependency]
-depsRequiredByWebSockets spec
-  | AS.WS.areWebSocketsUsed spec = AS.WS.sdkDepsRequiredForWebSockets
-  | otherwise = AS.Dependency.fromList [("uuid", "^9.0.0")]
```
infomiho, line 58: Not used, removed it.

waspc/src/Wasp/Generator/ServerGenerator.hs
```diff
@@ -145,12 +147,12 @@ npmDepsForWasp spec =
         AS.Dependency.fromList
           [ ("cookie-parser", "~1.4.6"),
             ("cors", "^2.8.5"),
-            ("express", "~4.18.1"),
+            ("express", expressVersionStr),
             ("morgan", "~1.10.0"),
-            ("dotenv", "16.0.2"),
+            ("dotenv", "^16.0.2"),
```
infomiho, line 152: Allows feat version bumps, why not, we do it for other deps.

Martinsos: Hm yeah I am always wondering about this. On one hand, this means that at any point some of these deps can introduce a bug (they happen even in patches, so certaily also in minor updates). On the other hand, by fixing the version we haven't really solved that problem anyway, because of the deps of deps -> real solution to avoid unexpected updates messing stuff up is only by doing dep freezing/locking.
Ok, makes sense to keep it all same now (`^`) and then we will figure out in the future what is the best solution.

waspc/src/Wasp/AI/GenerateNewProject/InitialFiles.hs
```diff
@@ -21,6 +21,7 @@ import Wasp.AI.GenerateNewProject.Common
     getProjectPrimaryColor,
   )
 import Wasp.AI.GenerateNewProject.Plan (PlanRule)
+import Wasp.Generator.WebAppGenerator.Common (reactVersion)
```
Martinsos, line 24: Ha, at first I wanted to comment that this is a bit wild, that Wasp/AI code imports react version from the Generator, but then I realized it actually makes sense.
Hm I am not sure, part of me wants to not have this dependency, to keep AI somewhat independent of Generator, but yeah this does keep us in sync automatically, so it might be fine.

What about other deps here, just below: typescript, prisma -> you are not pulling those versions from Generator. Following this way of reasoning, we should also import those? I guess what I would say: let's import all of let's import none. I am ok wtih importing none also. Or at least have TODO toimport those although that is a bit lazy.

waspc/src/Wasp/Generator/ServerGenerator/Common.hs
```diff
@@ -132,3 +133,8 @@ defaultServerPort = 3001
 
 defaultDevServerUrl :: String
 defaultDevServerUrl = "http://localhost:" ++ show defaultServerPort
+
+-- TODO: update this to use Wasp.SemanticVersion when we'll have support
```
Martinsos, line 137: Aha! Well interesting that I haven't implemented `~` in our semver. I haven't left any TODOs there, which means I thought I implemented everything I should. Truth is, `~`, `^`, `1.2.3 - 3.4.5`, `1.2.x`, ... -> those are "sugars" (https://github.com/npm/node-semver?tab=readme-ov-file#advanced-range-syntax) on top of basic primitives. So one could say our implementation of SemanticVersion is complete but it doesn't have this "sugar" layer, these advanced ways to define ranges, so we could add that on top in the future. But then I see that I did implement `^` :D. I wonder why I implemented just that one and not the rest. Ok. Maybe you can put a TODO in SemanticVersion.hs to implement other advanced ranges (https://github.com/npm/node-semver?tab=readme-ov-file#advanced-range-syntax) besides `^`?

waspc/src/Wasp/Generator/WebAppGenerator.hs
```diff
@@ -134,6 +138,7 @@ npmDepsForWasp _spec =
             ("@types/react", "^18.0.37"),
             ("@types/react-dom", "^18.0.11"),
             ("@vitejs/plugin-react", "^4.2.1"),
+            -- NOTE: used in the validat-env.mjs script
```
Martinsos, line 141: typo -> should be `validate` probably?

Martinsos (approved):
LGTM, nice work @infomiho ! Left some small comments but nothing major.

waspc/src/Wasp/AI/GenerateNewProject/InitialFiles.hs
infomiho: I've went with _importing none_ since for all other templates we are not dynamically injecting the versions but rather we'll validate them when this is done: #2317

waspc/src/Wasp/Generator/ServerGenerator/Common.hs
infomiho: I've left a comment in the SemanticVersion module 👍

infomiho:
@Martinsos I've cleaned up things bit more, tested with all example apps + Open SaaS template.

waspc/data/Generator/templates/server/README.md
```diff
@@ -1,5 +1 @@
 `npm start` to run server in development mode (it reloads on changes).
-
```
infomiho, line 2: These commands don't exist, so I removed this bit from the server README (even though no one reads that, but why not).

waspc/data/Generator/templates/server/package.json
```diff
@@ -13,8 +13,7 @@
     "validate-env": "node -r dotenv/config ./scripts/validate-env.mjs",
     "db-seed": "npm run bundle && node --enable-source-maps -r dotenv/config bundle/dbSeed.js",
     "db-migrate-prod": "prisma migrate deploy --schema=../db/schema.prisma",
-    "start-production": "{=& startProductionScript =}",
-    "standard": "standard"
```
infomiho, line 17: It was here since 2020 but it's not user accessible or useful 🤷

Martinsos (approved):
LGTM!

PR #2322, opened 2024-10-04
Update tsconfig.json validation messages

infomiho:
We didn't output the fully qualified field name for the missing error message e.g. we sad "module" is missing vs. "compilerOptions.module" is missing. 

This is maybe a bit over engineered, but I thought it looked nicely expressed with the type.


waspc/src/Wasp/Generator/ExternalConfig/TsConfig.hs
```diff
@@ -21,7 +22,15 @@ instance IsJavascriptValue Bool where
...
-type FieldName = String
+-- | Represents a fully qualified field name in a JSON object.
+-- For example, for the field "module" in the "compilerOptions" object,
+-- the fully qualified field name would be "compilerOptions.module".
+data FullyQualifiedFieldName = FieldName FieldPath
+
+type FieldPath = [String]
+
+instance Show FullyQualifiedFieldName where
+  show (FieldName fieldPath) = show $ intercalate "." fieldPath
```
Martinsos, line 33:
```suggestion
  show (FieldName fieldPath) = intercalate "." fieldPath
```

I think this extra `show` was redundant -> `intercalate` will already return a string in this case!

If you do wnat to have extra set of quotes around it, I would that below in the place of call, not add them here with `show`. Although I guess it was just a leftover.

Martinsos (approved):
I like it, LGTM! One small potential fix.

Martinsos (approved):

PR #2337, opened 2024-10-10
Moved examples/thoughts to Wasp TS config.

Martinsos:


sodic (approved):
You know what I think about the formatting.

The rest is OK.

PR #2347, opened 2024-10-20
Update Docs: Netlify Deploy Should Be Through CLI, Not Website

### LINKED ISSUE #1982: Properly document our client deployment
In production, our client is just a bunch of static files. When you input a non-existent file path into the address bar (or the browser tries to redirect to it), the navigation fails with a 404 - no surprises there.

You can see this behavior by serving the client files locally and trying to access http://localhost:3000/login directly from the address bar.

Therefore: **for Wasp client deployments to work, the server that's serving the static files must return `index.html` for all extension-less paths** (regardless of the path requested by the browser).
This way, the static server delegates all routing to the React Router on the frontend, which then knows what to do. The extension-less part is important because we still want to properly serve all other static resources (CSS, HTML, JS, images, etc.)

To solve this problem on Fly.io, we've configured our Go server to fall back onto `index.html` for all non-existent paths:
```Dockerfile
FROM pierrezemb/gostatic
CMD [ "-fallback", "index.html" ]
COPY ./dist/ /srv/http/
```

What needs to be done:
1. **Decide whether `fallback` is the best option** - Fallback means _"return `index.html` when you can't find the requested path on disk._ What we truly want is "return `index.html` for all paths without an extension." A possible edge case where `fallback` becomes problematic is a user-defined route that accidentally resolves to an existing extension-less file on disk (unlikely). Fallback might also have implications on response status codes.
2. **Modify our docs** - Our docs say *"Since the app's frontend is just a bunch of static files, you can deploy it to any static hosting provider."*. But that's not all there is to it. The docs should also mention that your hosting provider must make sure to resolve all paths to `index.html`. Perhaps we should include examples. 

Easy solutions:
 - `npx serve -s` - the `-s` or `--single` flag is meant for serving single page apps. The `--help` command says for the flag: _Rewrite all not-found requests to `index.html`_.

> [!NOTE]
> Since our docs don't mention this problem at all, and many user deploy the frontend on their own (without using Fly.io), how come we haven't heard any complaints? Are most servers configured to fall back onto `index.html` by default? Or are they hitting the issue and giving up?

### LINKED ISSUE #2344: Google Auth Redirection Not Working (Netlify & AWS)
**Describe the bug**

I am currently trying to deploy an app built with Open-Saas on Netlify and AWS EC2 instance with Nginx. Everything was working in the developer environment, but when I try logging in via Google it breaks. The Google Auth screen shows up normally and gives a response that results in several redirects, but after signing in it eventually redirects to https://myapp.netlify.app/oauth/callback, where it is met with a 404 (the same redirection returns a 304 in local developing environment). I have posted the screenshots from Google Dev console. At the time of failure, the url showing up in chrome is https://myapp.netlify.app/oauth/callback#{long_token}

**To Reproduce**
Use the [Open Saas Template](https://opensaas.sh/) to deploy a project frontend on Netlify and Docker container on an AWS EC2 instance with Nginx using Google Auth. The Http requests will break due to Netlify's security policy (Https requests will also be banned if you use a self-generated certificate), so add a custom domain name and use openssl to generate a trusted certificate and host a redirect on CloudFront.

Signing in via Google Auth will result in an error. In partilcular, it will result in a Netlify's "This page does not exist" error.

**Expected behavior**
Signing in via Google Auth results in a website error page / redirects to the example ai app.

**Screenshots**
Here is the error page I got:
Screenshot 2024-10-18 at 03 10 05

Here is the expected error page (Ideally it should sign me in and go to the app but even if Google Auth fails we should see this error instead of the previous one):
Screenshot 2024-10-18 at 03 12 41

Here are the two redirect requests being met with different responses:
image
image

 - OS: Mac
 - Browser: Chrome
 - Version Version: 129.0.6668.91 (Official Build) (arm64)

**Additional context**
I went though the docs and previous questions but was not able to find much helpful information. I imagine it has something to do with the netlify.toml file but am not sure what to change. Please let me know if there are any other information I can provide. Any suggestion would be much appreciated!

Bojun-Feng:
### Description

Fixes #2344
Fixes #1982

### DIFF
--- a/web/docs/advanced/deployment/manually.md
+++ b/web/docs/advanced/deployment/manually.md
@@ -271,6 +271,17 @@ If you follow the instructions above, Netlify will use `netlify.toml` file that 
 If you instead use another method of deployment to Netlify, e.g. do it via CI, make sure that Netlify picks up that `netlify.toml` file, or configure URL redirecting yourself manually on Netlify.
 
 It is recommended to deploy through the Netlify CLI in Github Actions. The first deploy can be through the website or manually to get a `NETLIFY_SITE_ID`, the following deploys can then be automatic.
+:::
+
+:::note
+Make sure you set this URL as the `WASP_WEB_CLIENT_URL` environment variable in your server hosting environment (e.g., Fly.io or Heroku).
+:::
+
+### Deploying through Github Actions
+
+To enable automatic deployment of the frontend whenever you push to the `main` branch, you can set up a GitHub Actions workflow. To do this, create a file in your repository at `.github/workflows/deploy.yaml`. Feel free to rename `deploy.yaml` as long as the file type is not changed.
+
+Here’s an example configuration file to help you get started. This example workflow will trigger a deployment to Netlify whenever changes are pushed to the main branch.
 
 <details>
 <summary>Example Github Action (Updated for 0.15.0)</summary>
@@ -327,32 +338,12 @@ jobs:
 <details>
 <summary>How do I get the Environment Variables?</summary>
 
-- **`NETLIFY_AUTH_TOKEN`**: This is the token that allows GitHub Actions to authenticate with your Netlify account. You can generate this from the **Netlify User Settings**:
-  - Go to your [Netlify User Settings Page](https://app.netlify.com/user/settings).
-  - Scroll down to **Personal Access Tokens**.
-  - Click on **New Access Token** to create a new token.
-  
-- **`NETLIFY_SITE_ID`**: This is the unique identifier for your Netlify site. You can obtain this from the **Site Settings** of your site in the Netlify dashboard:
-  - Go to your **Netlify dashboard**.
-  - Click on the specific site you want to deploy.
-  - Go to **Site Settings** → **General** → **Site Information** → **Site ID**.
+- **`NETLIFY_AUTH_TOKEN` & `NETLIFY_SITE_ID`**: They can be configured through Netlify.
 
 - **`WASP_SERVER_URL`**: This is the link that points to your backend and is generally only available after **deploying the backend**. This variable can be skipped when the backend is not functional or not deployed, but be aware that backend-dependent functionalities may be broken.
 
-After obtaining the environment variables, you need to store these values securely in GitHub Secrets. To add these:
-
-1. Go to your **GitHub repository**.
-2. Navigate to **Settings** → **Secrets and variables** → **Actions** → **New repository secret**.
-3. Add the following secrets:
-   - `NETLIFY_AUTH_TOKEN`
-   - `NETLIFY_SITE_ID`
-   - `WASP_SERVER_URL`
+After obtaining the environment variables, you need to store these values securely in GitHub Secrets.
 </details>
-:::
-
-:::note
-Make sure you set this URL as the `WASP_WEB_CLIENT_URL` environment variable in your server hosting environment (e.g., Fly.io or Heroku).
-:::
 
 ## Railway (server, client and database)
 
--- a/web/docs/advanced/deployment/manually.md
+++ b/web/docs/advanced/deployment/manually.md
@@ -268,7 +268,7 @@ That is it! Your client should be live at `https://<app-name>.netlify.app` ✨
 :::caution
 Please do not set up the continuous deploy through the Netlify website. When deployed in this way, Netlify doesn't redirect URLs toward `index.html` but instead returns 404, which can break the user authentication functionality. Wasp by default comes with a `netlify.toml` file in `.wasp/build/web-app` that configures Netfliy correctly, but the configuration may break when deployed through the Netlify website.
 
-It is recommended to deploy through the Netlify CLI in Github Actions. The first deploy can be through the website or manually through get a `NETLIFY_SITE_ID`, then the following deploys can be automatic.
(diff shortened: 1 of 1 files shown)


Bojun-Feng:
@Martinsos I have added a warning with my docker file attached as an example. Can we have this merged please?

Let me know if anything else needs to be changed!

web/docs/advanced/deployment/manually.md
```diff
@@ -265,6 +265,89 @@ netlify deploy --prod
 
 That is it! Your client should be live at `https://<app-name>.netlify.app` ✨
 
+:::caution
+Please do not set up the continuous deploy through the Netlify website. When deployed in this way, Netlify doesn't redirect URLs toward `index.html` but instead returns 404, which can break the user authentication functionality. Wasp by default comes with a `netlify.toml` file in `.wasp/build/web-app` that configures Netfliy correctly, but the configuration may break when deployed through the Netlify website.
```
Martinsos, line 269: I would suggest the following change:

```suggestion
:::caution Redirecting URLs toward `index.html`
If you follow the instructions above, Netlify will use `netlify.toml` file that Wasp generates by default in `.wasp/build/web-app/`. This will correctly configure Netlify to redirect URLs toward `index.html`, which is important since Wasp is SPA.

If you instead use another method of deployment to Netlify, e.g. do it via CI, make sure that Netlify picks up that `netlify.toml` file, or configure URL redirecting yourself manually on Netlify.
```

web/docs/advanced/deployment/manually.md
```diff
@@ -265,6 +265,89 @@ netlify deploy --prod
...
+
+1. Go to your **GitHub repository**.
+2. Navigate to **Settings** → **Secrets and variables** → **Actions** → **New repository secret**.
+3. Add the following secrets:
+   - `NETLIFY_AUTH_TOKEN`
+   - `NETLIFY_SITE_ID`
+   - `WASP_SERVER_URL`
+</details>
+:::
+
```
Martinsos, line 350: All right, instructions for deploying to Netlify via CLI!

So tricky part with docs is that on one hand, it is good for users to put a lot of stuff in them, because it usually helps, but on the other hand, that means more stuff to maintain and that will potentially become incorrect or stale with time. So we need to choose carefully how much to put.

I would say this is on the border of it, since it is quite Github Action and Netlify specific. i wonder if it might be better having this as a separate repo that has example app with tha tgithub action, so it is also tested immediately that way -> but ok that would require some setup on our side now.

Ok, I would say let's keep it and see, why not. But I would make it a separate section, so `### Automatic deployment via Github Actions`, and this would go under it.
Also, I would shorten the instructions on how to obtian the Netlify secrets, because those can easily become incorrect if Netfliy changes their UI. I would instead just say that they can obtain them from Netlify and that is it, they will figure it out.

Martinsos (commented):
Thanks, this is looking quite good! Left some comments, nothing super major, let's work a bit on those and we should be able to merge!

Martinsos (approved):
All right, looking good, thanks for this!

web/docs/advanced/deployment/manually.md
```diff
@@ -265,10 +265,86 @@ netlify deploy --prod
...
+If you instead use another method of deployment to Netlify, e.g. do it via CI, make sure that Netlify picks up that `netlify.toml` file, or configure URL redirecting yourself manually on Netlify.
+
+It is recommended to deploy through the Netlify CLI in Github Actions. The first deploy can be through the website or manually to get a `NETLIFY_SITE_ID`, the following deploys can then be automatic.
+:::
+
 :::note
 Make sure you set this URL as the `WASP_WEB_CLIENT_URL` environment variable in your server hosting environment (e.g., Fly.io or Heroku).
 :::
 
+### Deploying through Github Actions
```
Martinsos, line 280: @infomiho I just merged this PR by a contributor. They made it clear that if deploying on Netlify, we need redirectin to be correctly set (either manually or using our netlify.toml).

They also added this section though, which shows how to exactly do it via Github Action. I approved it and merged, but I do wonder if this is too much for this part of the docs, are we going too far / doing too much here. Since you are now redesigning deployment docs I wanted to hear what you think. Maybe we will have a specific Github Actions section and this will be better explained there directly, instead of here under Netlify? Should we matbe (re)move this part?

Bojun-Feng: @Martinsos @infomiho Just wanted to hear your opinion on something similar: I am interested in writing documentation to deploy OpenSaaS on AWS (either EC2 or Lightsail), probably also including GitHub Actions to handle CI/CD.

I imagine I will briefly describe the peculiarities related to AWS, but maybe the GitHub Actions code is a bit too much? If that is the case I can put the detailed code that works with a particular version of OpenSaaS into a separate blog post, and only cover the general steps in the docs.

Martinsos: @Bojun-Feng btw, @infomiho is on the conference from tomorrow so it might take him a bit longer to respond to this, but he should get to it the next week latest.

As for your question: Ok nice, that would be cool, the docs for AWS! Btw why OpenSaas, why not Wasp? I don't think there is anything OpenSaas specific? Or is there?

I think you are right, Github Action might be too much, especially if it has a lot of stuff that is "general" and not specific to AWS. But as you said, that could be a blog post for example, you are right! Btw we also ahve this https://github.com/wasp-lang/learning-materials -> sometimes we put stuff like this there. Hard to say at the moment if blog post or this is better, @infomiho might have some good ideas.

PR #2362, opened 2024-10-29
Validate env vars with Zod

### LINKED ISSUE #1067: Defining environment variable types using Zod
This is what T3 does, we can take a look at them for inspiration. We should explore sense and either do it or close this issue.

infomiho:
This PR will enable us to have strict env vars validation with Zod for Wasp's internal env vars. Also, it will enable users to define their own env var validation. Wasp will use the Zod schemas to output errors for missing env vars.

### Left to do
- [x] After the strict null checks is merged, make sure that the types are correct #2360 
- [x] Make sure all env vars in our code are used through the `env` object and not directly
- [x] Experiment with enabling users to add their own env vars validation https://github.com/wasp-lang/wasp/pull/2363
- [x] Update docs https://github.com/wasp-lang/wasp/pull/2450
  - Make sure to mention the difference between `env` and `process.env` 
- [x] Fix `keycloak.env.KEYCLOAK_REALM_URL` leftover 

Closes #1067


waspc/data/Generator/templates/Dockerfile
```diff
@@ -58,12 +58,12 @@ COPY --from=server-builder /app/node_modules ./node_modules
...
 COPY --from=server-builder /app/.wasp/out/sdk .wasp/out/sdk
-# Copying 'server/node_modules' because 'validate-env.mjs' executes independent
-# of the bundle and references the dotenv package.
+# Copying 'server/node_modules' because we require dotenv package to
+# load environment variables
+# TODO: replace dotenv with native Node.js environment variable loading
 COPY --from=server-builder /app/.wasp/build/server/node_modules .wasp/build/server/node_modules
 COPY --from=server-builder /app/.wasp/build/server/bundle .wasp/build/server/bundle
 COPY --from=server-builder /app/.wasp/build/server/package*.json .wasp/build/server/
-COPY --from=server-builder /app/.wasp/build/server/scripts .wasp/build/server/scripts
```
infomiho, line 66: We no longer need `validate-env.mjs` since we are validating the server env vars with Zod now.

waspc/data/Generator/templates/sdk/wasp/client/config.ts
```diff
@@ -1,7 +1,7 @@
-{{={= =}=}}
 import { stripTrailingSlash } from '../universal/url.js'
+import { env } from './env.js'
 
-const apiUrl = stripTrailingSlash(import.meta.env.REACT_APP_API_URL) || '{= defaultServerUrl =}';
```
infomiho, line 4: `env` objects is `import.meta.env` but validated. The default values are also defined in the Zod schema.

waspc/data/Generator/templates/sdk/wasp/env/index.ts
```diff
@@ -0,0 +1,29 @@
...
+const redColor = '\x1b[31m'
+
+export function ensureEnvSchema<Schema extends z.ZodTypeAny>(
+  data: unknown,
+  schema: Schema
+): z.infer<Schema> {
+  try {
+    return schema.parse(data)
+  } catch (e) {
+    // TODO: figure out how to output the error message in a better way
```
infomiho, line 12: | Screenshot 2024-11-29 at 09 04 34 |
| --- |
| How it looks on the server |

| Screenshot 2024-11-29 at 09 14 05 |
| --- |
| Error on the client |

I'm not sure if this is good enough?

waspc/data/Generator/templates/sdk/wasp/server/auth/oauth/provider.ts
```diff
@@ -1,23 +1,19 @@
...
+  OAuthClient extends OAuth2Provider | OAuth2ProviderWithPKCE
 >({
   id,
   displayName,
-  env,
   oAuthClient,
 }: {
   id: string;
   displayName: string;
-  env: Env;
```
infomiho, line 14: Instead of OAuth env vars having their special validation and living in a special place, they are now used directly from the `env` object.

waspc/data/Generator/templates/sdk/wasp/server/config.ts
```diff
@@ -1,94 +1,47 @@
...
-  port: number;
-  databaseUrl: string | undefined;
-}
+const frontendUrl = stripTrailingSlash(env.WASP_WEB_CLIENT_URL)
+const serverUrl = stripTrailingSlash(env.WASP_SERVER_URL)
 
-type EnvConfig = BaseConfig & {
-  frontendUrl: string;
-  serverUrl: string;
+const allowedCORSOriginsPerEnv: Record<NodeEnv, string | string[]> = {
```
infomiho, line 25: Default CORS values per NODE_ENV, same as before.

waspc/data/Generator/templates/sdk/wasp/server/config.ts
```diff
@@ -1,94 +1,47 @@
...
-    serverUrl,
-    allowedCORSOrigins: '*',
-    {=# isAuthEnabled =}
-    auth: {
-      jwtSecret: 'DEVJWTSECRET'
-    }
-    {=/ isAuthEnabled =}
+const allowedCORSOrigins = allowedCORSOriginsPerEnv[env.NODE_ENV]
+
+const config: Config = {
```
infomiho, line 31: Simplified the creation of the `config` object, instead of having three separate things `all`, `development`, `production`, now we construct a single object and vary some of the values (e.g. allowed CORS origins) based on `NODE_ENV`.

waspc/data/Generator/templates/sdk/wasp/server/email/index.ts
```diff
@@ -7,24 +8,24 @@ import { EmailSender } from "./core/types.js";
 {=# isSmtpProviderUsed =}
 const emailProvider = { 
     type: "smtp",
-    host: process.env.SMTP_HOST!,
-    port: parseInt(process.env.SMTP_PORT!, 10),
-    username: process.env.SMTP_USERNAME!,
-    password: process.env.SMTP_PASSWORD!,
+    host: env.SMTP_HOST,
+    port: env.SMTP_PORT,
```
infomiho, line 12: Why did `parseInt(process.env.SMTP_PORT!, 10)` turn into `env.SMTP_PORT`? 

Because Zod not only validates that `SMTP_PORT` exists (no need for `!` anymore, here) but also casts the string to a number.

waspc/data/Generator/templates/sdk/wasp/server/env.ts
```diff
@@ -0,0 +1,141 @@
+{{={= =}=}}
```
infomiho, line 1: All of env Wasp defined env variables live in this file - single central location that validates them. All env usage in the Wasp codebase should use this validated `env` object. 

Because we have different needs when in dev and production, we need to have different validations and default values (e.g. server URL is `http://localhost:3001` in dev by default but in production it is required!).

So, we have:
- `serverCommonSchema` (same for both dev and prod)
- `serverDevSchema` (used in dev, has defaults for some env vars)
- `serverProdSchema` (used in prod)

And then we have this bit of Zod "magic"
```ts
const serverEnvSchema = z.discriminatedUnion('NODE_ENV', [
  serverDevSchema.merge(serverCommonSchema),
  serverProdSchema.merge(serverCommonSchema)
])
```
which based on `NODE_ENV` picks one of the other schema to use. On type level it looks smth like this:
```ts
type Env =
  | {
      NODE_ENV: 'development'
      // optional in dev
      SOMETHING: string | undefined
    }
  | {
      NODE_ENV: 'production'
      // required in prod
      SOMETHING: string
    }
```

waspc/examples/todoApp/src/App.tsx
```diff
@@ -16,6 +16,7 @@ export function App() {
 
   const connectionIcon = isConnected ? '🟢' : '🔴'
 
+  // TODO: enable users to define their own client env vars
```
infomiho, line 19: Did it in a follow-up PR: #1067

waspc/data/Generator/templates/react-app/package.json
```diff
@@ -7,9 +7,8 @@
   {=& depsChunk =},
   {=& devDepsChunk =},
   "scripts": {
-    "start": "npm run validate-env && vite",
```
infomiho, line 10: Removing the `validate-env.mjs` script in favor of doing it inside of a Vite plugin: https://github.com/wasp-lang/wasp/pull/2392

waspc/data/Generator/templates/sdk/wasp/server/env.ts
Martinsos, line 1: Is it bad that this is so centralized? I imagine taht on Haskell side you have to import a ton of things from the codebase in the generator that does this in order to side what goes in and what doesn't. Part of me is tempted to look for a more decentralized appraoch, where each "module" (auth providers, job, email, ...) defines its own env vars and they are somehow all registered in this central place, sure, but they are not all defined in this central place, but it is done via some other mechanism. Maybe that would cause cyclic deps hm, I guess it depends on how you execute it. I haven't thought it through more than this, but you probably did, so I am interested what was your reasoning, why you chose this approach, what are pros/cons.

waspc/src/Wasp/Generator/EmailSenders.hs
```diff
@@ -0,0 +1,23 @@
...
+import qualified Wasp.AppSpec.App.EmailSender as AS.App.EmailSender
+
+getEnabledEmailProvidersJson :: AS.App.EmailSender.EmailSender -> Aeson.Value
+getEnabledEmailProvidersJson emailSender =
+  object $
+    makeProviderJson
+      <$> providersKeyAndName
+  where
+    providersKeyAndName =
+      [ ("isSmtpProviderUsed", AS.App.EmailSender.SMTP),
```
Martinsos, line 17: Why do you use `used` here when you used `enabled` in other places? Btw enabled sounds better to me -> we know if they are enabled, not if they are used. Not a biggie though, both terms can work, but would be nice if terminology was synced I think.

Martinsos (commented):
@infomiho looks quite solid to me, improvements in the codebase are visible already! Nice stuff.

I left some comments.

Some general stuff I also had on mind:
1. What if users try to access our env vars via `process.env`? I know they can, nothing bad will really happen, but we would rather they use them via our `env` object, right, so they have better DX (validation happened, types)? This becomes even more important once they start defining their own variables I guess (which is another PR)? I don't think we should in any way forbid them from using process.env of course, that is too crucial of a thing to touch, but I wonder if we should do anything. Maybe monkey patch it in some way so in dev they get a warning that they should use our env object instead? Sounds a bit wild.
2. process.env and our env object -> are the values in them the same? Shoudl they be? I am guessing they are not the same because of zod default values -> so in `env`, values will be set to those defaults if they are empty, right? While in process.env, they will not be. This could be confusing for users -> you access variable from `env`, it has some value, but you know you haven't set it via env, nor anybody else. So how can it have value? Hm. Maybe just documenting this behaviour is good enough.
3. Just to confirm, what is the behaviour in production now? If some of the env vars are missing and they are required, server will print error and crash? What about in development, it will print errors and also crash or continue?

waspc/data/Generator/templates/Dockerfile
```diff
@@ -55,15 +55,15 @@ WORKDIR /app
...
 COPY --from=server-builder /app/node_modules ./node_modules
-# Copying the SDK because 'validate-env.mjs' executes independent of the bundle
+# Copying the SDK because the server bundle doesn't bundle the SDK
 # and references the 'wasp' package.
 COPY --from=server-builder /app/.wasp/out/sdk .wasp/out/sdk
-# Copying 'server/node_modules' because 'validate-env.mjs' executes independent
-# of the bundle and references the dotenv package.
+# Copying 'server/node_modules' because we require dotenv package to
+# load environment variables
+# TODO: replace dotenv with native Node.js environment variable loading
```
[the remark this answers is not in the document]
infomiho: It's always nicer to use natively supported methods than a package to do the same thing. It's available since [Node.js 20](https://nodejs.org/en/blog/release/v20.6.0) and we should probably go for it when Node.js 20 becomes our minimum version.

waspc/data/Generator/templates/sdk/wasp/server/config.ts
infomiho: I think the previous approach we clever since it envisioned many different options for dev and for production, but it in the end, we only vary the CORS origins between dev and production. So, it seemed like an overkill and I know I had these complex types to capture the different objects. But when I looked closer, it turned out this simple `Config` type is enough.

waspc/data/Generator/templates/sdk/wasp/server/env.ts
```diff
@@ -0,0 +1,142 @@
...
+    required_error: 'WASP_WEB_CLIENT_URL is required',
+  })
+  .url({
+    message: 'WASP_WEB_CLIENT_URL must be a valid URL',
+  })
+
+const jwtTokenSchema = z
+  .string({
+    required_error: 'JWT_SECRET is required',
+  })
```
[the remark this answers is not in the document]
infomiho: I use them as building blocks for the dev and production schemas below.

waspc/data/Generator/templates/sdk/wasp/server/env.ts
```diff
@@ -0,0 +1,142 @@
...
+})
+
+const serverProdSchema = z.object({
+  NODE_ENV: z.literal('production'),
+  WASP_SERVER_URL: serverUrlSchema,
+  WASP_WEB_CLIENT_URL: clientUrlSchema,
+  {=# isAuthEnabled =}
+  JWT_SECRET: jwtTokenSchema,
+  {=/ isAuthEnabled =}
+})
```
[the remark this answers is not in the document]
infomiho: To get the benefit of Zod's inferred types, things have to be known on compile time, so I can't do dynamic default values. If I wanted to do that, I'd have to say that both in dev and production something is optional and then dynamically only set default values for dev - but in reality, they have different rules e.g. in dev something is optional but in production it's required. And I need to declare that and make it known when Zod tries to infer types i.e. on compile time. 

There is duplication here, but this is the best way to get nice looking types between various `NODE_ENV` values. I haven't found anything nicer, I've looked at how others used Zod and this is how they did it as well.

waspc/data/Generator/templates/sdk/wasp/server/env.ts
infomiho: I think we could build such a system, each part of the framework declares what it needs and declares that for development and production. Then we assemble it in this file by importing those schemas and merge them into one large schema which we use to validate the `process.env` object.

I feel like it could be useful if we start splitting Wasp into multiple full-stack modules or smth like that. Each module declares what it requires. But ... I think it's maybe an overkill for this PR. This PR wanted to use Zod to validate env vars and make sure all framework code uses validated env vars. I think the decentralised env var schemas thing could be a nice improvement if we decide it's worth the extra complexity in the future.

waspc/src/Wasp/Generator/EmailSenders.hs
infomiho: I agree, it's a better and more precise name 👍

waspc/data/Generator/templates/sdk/wasp/env/index.ts
```diff
@@ -0,0 +1,31 @@
...
+  data: unknown,
+  schema: Schema
+): z.infer<Schema> {
+  try {
+    return schema.parse(data)
+  } catch (e) {
+    if (e instanceof z.ZodError) {
+      const errorOutput = [
+        '',
+        '╔═════════════════════════════╗',
```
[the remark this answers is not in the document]
infomiho: Gotcha, let me do another iteration of this output 👍 

I'll create an issue for making this more universal.

infomiho:
@Martinsos 

> What if users try to access our env vars via process.env?

They can, it's part of the Node.js platform - we don't want to deny them that. We offer a convenience method of accessing Wasp defined env variables - which are validated and typed. Monkey patching sounds like it could break some other libraries. that users might want to use in the future - so I think that's not the way to go.

> process.env and our env object -> are the values in them the same?

One of them is a raw object, one of them is our validated and typed object. We encourage users to use our `env` object, while they are free to use the source `process.env` object as well. We can mention in the docs that there will difference so they are not surprised 👍 

> Just to confirm, what is the behaviour in production now? If some of the env vars are missing and they are required, server will print error and crash? What about in development, it will print errors and also crash or continue?

Failed validation crashes the server / client - in both development and production. This way it's clear upfront what's wrong vs. having some sort of runtime errors that might not be clear to our users.

waspc/data/Generator/templates/sdk/wasp/env/index.ts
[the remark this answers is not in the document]
infomiho: Screenshot 2024-12-16 at 10 32 37

Here's the updated version. Related issue: https://github.com/wasp-lang/wasp/issues/2420

waspc/data/Generator/templates/sdk/wasp/env/index.ts
Martinsos: I would probably experiment with the title not being a box, but just being underlined title. When it is a box it feels like it is a standalone thing, that stuff under it does not belong to it. Also, indent the bullet points, to make it even clearer they are "under" it.

So
```
===== Env vars validation failed =====
 - REACT_APP_WS_URL is required
 - SOME_VAR is required
======================================
```

waspc/data/Generator/templates/sdk/wasp/server/config.ts
Martinsos: Ok, makes sense, we can always modify it if needed!

waspc/data/Generator/templates/sdk/wasp/env/index.ts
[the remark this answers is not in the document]
Martinsos: I see this now -> nice! Why don't you use nice characters for the left border? `╔` and `║` and `╚`?

waspc/data/Generator/templates/sdk/wasp/server/env.ts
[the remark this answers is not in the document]
Martinsos: Aha they are reused? Ok then!

waspc/data/Generator/templates/sdk/wasp/server/env.ts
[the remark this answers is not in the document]
Martinsos: All right, it is quite small and local so not a big deal!

waspc/data/Generator/templates/sdk/wasp/server/env.ts
Martinsos: Hm ok yeah, if you don't feel liek this makes our code much worse, then it is ok I believe. We can see how this behaves!

Martinsos:
> @Martinsos
> 
> > What if users try to access our env vars via process.env?
> 
> They can, it's part of the Node.js platform - we don't want to deny them that. We offer a convenience method of accessing Wasp defined env variables - which are validated and typed. Monkey patching sounds like it could break some other libraries. that users might want to use in the future - so I think that's not the way to go.
> 
> > process.env and our env object -> are the values in them the same?
> 
> One of them is a raw object, one of them is our validated and typed object. We encourage users to use our `env` object, while they are free to use the source `process.env` object as well. We can mention in the docs that there will difference so they are not surprised 👍
> 
> > Just to confirm, what is the behaviour in production now? If some of the env vars are missing and they are required, server will print error and crash? What about in development, it will print errors and also crash or continue?
> 
> Failed validation crashes the server / client - in both development and production. This way it's clear upfront what's wrong vs. having some sort of runtime errors that might not be clear to our users.

I believe @sodic suggested we more often quote only small parts of the original message, but in this case I would prefer if you quoted the whole thing, because you left out a lot and I had to go back to my original message to see what I actually wrote, which was not evident here and was missing from the discussion if somebody reads it. Not a biggie though.

Ok, I agree about not touching process.env is probably the best -> btw I advised monkey patching only to print something in dev, exactly because we don't know who might use it, so not to be intrusive, but I guess even that could be surprising in a way, so yeah better not to touch it probably at all.
The thing I think will be confusing is that they will just not be aware of `env`, and neither will LLM they are using, so we will often be seeing people using `process.env` instead of `env`, because they didn't even know `env` exists. So I am tempted to try to do sometihng about it. Monkey patching does sound intrusive tough, and linter rules or something sound too complex.

On the second thing -> You explained shortly the difference between process.env and env + said we can document it, which is what I already knew / suggested above, but you didn't answer the most interesting (to me) part of my question about default values. That from my perspective is the biggest difference and it could be considered quite unexpected, therefore a source of confusion / wrong assumptions. Have you considered that, and if so, what are your thoughts?

Dev/production craches -> ok great, that is what we want, I was just double checking that is how we did it.

Martinsos (approved):
Left some comments but approved!
I would also suggest that @sodic takes a look before merging.

waspc/data/Generator/templates/sdk/wasp/env/index.ts
infomiho: This screenshots are old news :) the code has changed and it doesn't output a box anymore.

waspc/data/Generator/templates/sdk/wasp/env/index.ts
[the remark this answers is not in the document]
infomiho: I tried it and I didn't like it actually!

The box looks _too polished_ but it's missing the right border, so this look is kinda _less polished_ and the missing right border doesn't bother me that much visually 🤷

infomiho:
> > @Martinsos
> > > What if users try to access our env vars via process.env?
> > 
> > 
> > They can, it's part of the Node.js platform - we don't want to deny them that. We offer a convenience method of accessing Wasp defined env variables - which are validated and typed. Monkey patching sounds like it could break some other libraries. that users might want to use in the future - so I think that's not the way to go.
> > > process.env and our env object -> are the values in them the same?
> > 
> > 
> > One of them is a raw object, one of them is our validated and typed object. We encourage users to use our `env` object, while they are free to use the source `process.env` object as well. We can mention in the docs that there will difference so they are not surprised 👍
> > > Just to confirm, what is the behaviour in production now? If some of the env vars are missing and they are required, server will print error and crash? What about in development, it will print errors and also crash or continue?
> > 
> > 
> > Failed validation crashes the server / client - in both development and production. This way it's clear upfront what's wrong vs. having some sort of runtime errors that might not be clear to our users.
> 
> I believe @sodic suggested we more often quote only small parts of the original message, but in this case I would prefer if you quoted the whole thing, because you left out a lot and I had to go back to my original message to see what I actually wrote, which was not evident here and was missing from the discussion if somebody reads it. Not a biggie though.
> 
> Ok, I agree about not touching process.env is probably the best -> btw I advised monkey patching only to print something in dev, exactly because we don't know who might use it, so not to be intrusive, but I guess even that could be surprising in a way, so yeah better not to touch it probably at all. The thing I think will be confusing is that they will just not be aware of `env`, and neither will LLM they are using, so we will often be seeing people using `process.env` instead of `env`, because they didn't even know `env` exists. So I am tempted to try to do sometihng about it. Monkey patching does sound intrusive tough, and linter rules or something sound too complex.
> 
> On the second thing -> You explained shortly the difference between process.env and env + said we can document it, which is what I already knew / suggested above, but you didn't answer the most interesting (to me) part of my question about default values. That from my perspective is the biggest difference and it could be considered quite unexpected, therefore a source of confusion / wrong assumptions. Have you considered that, and if so, what are your thoughts?
> 
> Dev/production craches -> ok great, that is what we want, I was just double checking that is how we did it.

Replying to the full message but I'm only replying to the last question.

People use the `config` object today to get the `serverUrl` and `frontendUrl` nowadays. So they are used to using higher level abstractions. `env` object is something we made up, we document it and say it has default values for certain things. People can use it and there they get the default values. If they use the platform primitive `process.env`, it's just what they expect. I don't see this is a potential source of confusion. I'd rather do less at first than more. Let's see how people use this and then put extra guard rails if we see it as a source of problems.

waspc/data/Generator/templates/sdk/wasp/env/index.ts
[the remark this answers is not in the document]
Martinsos: Ok yeah I thought that might be the case. Although this also looks a bit weird. But ok, it is good enough for now, I also don't have better ideas at the moment!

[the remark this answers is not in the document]
Martinsos: Some quick ideas though:

```
===== Env vars validation failed =====
> - some text
> - some more text
> - third problem
======================================
```

or even just

```
===== Env vars validation failed =====
 - some text
 - some more text
 - third problem
======================================
```

There will probalby never be more then a couple of these errors, so if you remove those empty lines and make it more compact, you get it all to look quite tight and connected, I think left border might not even be needed then.

Martinsos:
> People use the `config` object today to get the `serverUrl` and `frontendUrl` nowadays. So they are used to using higher level abstractions. `env` object is something we made up, we document it and say it has default values for certain things. People can use it and there they get the default values. If they use the platform primitive `process.env`, it's just what they expect. I don't see this is a potential source of confusion - maybe I'm wrong. I'd rather do less at first than more. Let's see how people use this and then put extra guard rails if we see it as a source of problems.

That makes sense to me -> I wanted to hear your thinking on this. Ok, if we make it clear in the docs that it might have default values, then I think that is fine.
Btw I get what you mean by `config`, but the thing with `env` and `process.env` is that they have very similar names and are even more similar in concept, so it is more likely that people will expect them to behave the same, that is why I think it is important to be careful about it.

Martinsos (approved):
Approved! One or two small comments left, pls take a look but then do as you like.

waspc/data/Generator/templates/sdk/wasp/client/index.ts
```diff
@@ -11,4 +11,7 @@ export enum HttpMethod {
 export type Route = { method: HttpMethod; path: string }
 
 // PUBLIC API
-export { config, ClientConfig } from './config'
+export { config, ClientConfig } from './config.js'
+
+// PUBLIC API
+export { env } from './env.js'
```
sodic, line 17: Just please add the public API changes to the annoying table.

waspc/data/Generator/templates/sdk/wasp/env/index.ts
```diff
@@ -0,0 +1,29 @@
...
+         errorOutput.push(`| - ${error.message}`)
+      }
+      errorOutput.push('|')
+      errorOutput.push('|════════════════════════════════')
+      console.error(redColor, errorOutput.join('\n'))
+      throw new Error('Error parsing environment variables')
+    } else {
+      throw e
+    }
+  }
```
sodic, line 28: I like how you made this very visible, but you having to do that is a nice reminder that our output is very noisy. I hope we clean that up soon

waspc/data/Generator/templates/sdk/wasp/env/index.ts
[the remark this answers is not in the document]
sodic: I'd drop all the `|` at the beginning of lines. It's weird and adds no extra info.

Martin's last suggestion is my favorite.
Obviously, I won't argue over this, so you can choose whatever you want :)

waspc/data/Generator/templates/sdk/wasp/server/env.ts
sodic: > I think we could build such a system, each part of the framework declares what it needs and declares that for development and production. Then we assemble it in this file by importing those schemas and merge them into one large schema which we use to validate the process.env object.

I wanted to suggest exactly this!

I see it as a clear improvement, but it's perfectly fine not to handle it now.

waspc/data/Generator/templates/sdk/wasp/server/env.ts
```diff
@@ -0,0 +1,142 @@
...
+
+const clientUrlSchema = z
+  .string({
+    required_error: 'WASP_WEB_CLIENT_URL is required',
+  })
+  .url({
+    message: 'WASP_WEB_CLIENT_URL must be a valid URL',
+  })
+
+const jwtTokenSchema = z
```
sodic, line 108: This one should also be under `isAuthEnabled` I believe. I see it's unused in the tests.

sodic (commented):
Nice, I'm a big fan of this change!

I don't have many comments, Martin covered it well.

waspc/data/Generator/templates/sdk/wasp/server/env.ts
```diff
@@ -0,0 +1,142 @@
...
+  GITHUB_CLIENT_ID: z.string({
+    required_error: 'GITHUB_CLIENT_ID is required',
+  }),
+  GITHUB_CLIENT_SECRET: z.string({
+    required_error: 'GITHUB_CLIENT_SECRET is required',
+  }),
+  {=/ enabledAuthProviders.isGitHubAuthEnabled =}
+  {=# enabledAuthProviders.isDiscordAuthEnabled =}
+  DISCORD_CLIENT_ID: z.string({
+    required_error: 'DISCORD_CLIENT_ID is required',
```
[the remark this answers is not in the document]
infomiho: I've improved the "it's required when using X" bit, but I'm having a hard time including anything else e.g. JWT_SECRET is used to sign the tokens etc. I'd like to leave that out to keep the error as generic as possible.

waspc/data/Generator/templates/sdk/wasp/prettier.config.js
```diff
@@ -0,0 +1,7 @@
+// Used for internal Wasp development only, not copied to generated app.
+module.exports = {
```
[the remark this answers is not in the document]
infomiho: It turns out you can :)

waspc/data/Generator/templates/Dockerfile
```diff
@@ -55,15 +55,15 @@ WORKDIR /app
 # Copying the top level 'node_modules' because it contains the Prisma packages
 # necessary for migrating the database.
 COPY --from=server-builder /app/node_modules ./node_modules
-# Copying the SDK because 'validate-env.mjs' executes independent of the bundle
+# Copying the SDK because the server bundle doesn't bundle the SDK
```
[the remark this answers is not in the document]
infomiho: Rollup doesn't bundle the `wasp/*` deps because they are considered "external" deps. I tried making them "internal" and it still didn't want to bundle them. Maybe because they are from `node_modules`. So, yep, we still need to copy over the SDK in the build context.

waspc/data/Generator/templates/server/package.json
```diff
@@ -7,10 +7,9 @@
   "comment-filip": "The server.js location changed because we have now included client source files above .wasp/out/server/src.",
   "scripts": {
     "bundle": "rollup --config --silent",
-    "start": "npm run validate-env && node --enable-source-maps -r dotenv/config bundle/server.js",
+    "start": "node --enable-source-maps -r dotenv/config bundle/server.js",
     "bundle-and-start": "npm run bundle && npm run start",
     "watch": "nodemon --exec 'npm run bundle-and-start || exit 1'",
-    "validate-env": "node -r dotenv/config ./scripts/validate-env.mjs",
```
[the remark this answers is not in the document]
infomiho: Nope, they are not using it... not yet at least :)

waspc/data/Generator/templates/Dockerfile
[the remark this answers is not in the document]
infomiho: The reason why we don't bundle `wasp/` is my poor configuration. I didn't realize that we needed one extra Rollup plugin to enable Rollup to bundle packages from `node_modules`.

I'll create an issue for this since this would be nice to have - so there is no need to copy over the SDK anymore. https://github.com/wasp-lang/wasp/issues/2446

PR #2372, opened 2024-11-15
Fix building with TS config

### LINKED ISSUE #2368: Fix TS Config build
`wasp build` doesn't work when using Wasp's new TS config.

**The core problem** behind the bug is our mishandling of dev dependencies and production dependencies. `wasp-config` has no business being included in Docker containers. Fixing that would take a while, probably a week or more.

**The trigger for the bug** is the fact that `wasp-config` is a local project outside the Docker's build context. We could prevent this by moving wasp-config into the .wasp folder, which would also take some time.

**Possible quick fix:** Read the packge.json in Haskell during `wasp-build` and kick out the problematic `wasp-config` package (which isn't required after `wasp build` finishes).

sodic:
Fixes #2368

### DIFF
--- a/waspc/cli/src/Wasp/Cli/Command/Build.hs
+++ b/waspc/cli/src/Wasp/Cli/Command/Build.hs
@@ -18,7 +18,7 @@ import Wasp.Cli.Command.Message (cliSendMessageC)
 import Wasp.Cli.Command.Require (InWaspProject (InWaspProject), require)
 import Wasp.Cli.Message (cliSendMessage)
 import Wasp.CompileOptions (CompileOptions (..))
-import qualified Wasp.Generator.Common (ProjectRootDir)
+import Wasp.Generator.Common (ProjectRootDir)
 import Wasp.Generator.Monad (GeneratorWarning (GeneratorNeedsMigrationWarning))
 import Wasp.Generator.SdkGenerator.Common (sdkRootDirInGeneratedCodeDir, sdkRootDirInProjectRootDir)
 import qualified Wasp.Message as Msg
@@ -131,7 +131,7 @@ build = do
 
 buildIO ::
   Path' Abs (Dir WaspProjectDir) ->
-  Path' Abs (Dir Wasp.Generator.Common.ProjectRootDir) ->
+  Path' Abs (Dir ProjectRootDir) ->
   IO ([CompileWarning], [CompileError])
 buildIO waspProjectDir buildDir = compileIOWithOptions options waspProjectDir buildDir
   where
--- a/waspc/cli/src/Wasp/Cli/Command/Build.hs
+++ b/waspc/cli/src/Wasp/Cli/Command/Build.hs
@@ -120,10 +120,10 @@ build = do
     removeAllMentionsOfWaspConfig :: Value -> Value
     removeAllMentionsOfWaspConfig packageLockJsonObject =
       packageLockJsonObject
-        & key "packages" . _Object %~ HM.filterWithKey packageNameDoesntMentionWaspConfig
         & key "packages" . key "" %~ removeWaspConfigDevDependency
-      where
-        packageNameDoesntMentionWaspConfig package _ = not ("wasp-config" `isInfixOf` package)
+        & key "packages" . _Object
+          %~ HM.filterWithKey
+            (\packageKey _ -> not ("wasp-config" `isInfixOf` packageKey))
 
     removeWaspConfigDevDependency :: Value -> Value
     removeWaspConfigDevDependency original =
--- a/waspc/cli/src/Wasp/Cli/Command/Build.hs
+++ b/waspc/cli/src/Wasp/Cli/Command/Build.hs
@@ -10,8 +10,10 @@ import Control.Monad.IO.Class (liftIO)
 import Data.Aeson (Value (..))
 import Data.Aeson.Lens
 import qualified Data.HashMap.Strict as HM
-import Data.Text (isInfixOf)
+import Data.List (isSuffixOf)
+import Data.Text (Text, unpack)
 import StrongPath (Abs, Dir, Path', castRel, (</>))
+import qualified System.FilePath as FP
 import Wasp.Cli.Command (Command, CommandError (..))
 import Wasp.Cli.Command.Compile (compileIOWithOptions, printCompilationResult)
 import Wasp.Cli.Command.Message (cliSendMessageC)
@@ -75,16 +77,17 @@ build = do
 
   liftIO (prepareFilesNecessaryForDockerBuild waspProjectDir buildDir) >>= \case
     Left err -> throwError $ CommandError "Failed to prepare files necessary for docker build" err
-    Right () ->
-      cliSendMessageC $
-        Msg.Success "Your wasp project has been successfully built! Check it out in the .wasp/build directory."
+    Right () -> return ()
+
+  cliSendMessageC $
+    Msg.Success "Your wasp project has been successfully built! Check it out in the .wasp/build directory."
   where
     prepareFilesNecessaryForDockerBuild waspProjectDir buildDir = runExceptT $ do
       -- Until we implement the solution described in https://github.com/wasp-lang/wasp/issues/1769,
-      -- we're copying all files and folders necessary for the build into the .wasp/build directory.
+      -- we're copying all files and folders necessary for Docker build into the .wasp/build directory.
       -- We chose this approach for 0.12.0 (instead of building from the project root) because:
-      --   - The build context remains small (~1.5 MB vs ~900 MB).
-      --   - We don't risk copying possible secrets from the project root into the build context.
+      --   - The Docker build context remains small (~1.5 MB vs ~900 MB).
+      --   - We don't risk copying possible secrets from the project root into Docker's build context.
       --   - The commands for building the project stay the same as before
       --     0.12.0, which is good for both us (e.g., for fly deployment) and our
       --     users  (no changes in CI/CD scripts).
@@ -114,19 +117,32 @@ build = do
 
       -- A hacky quick fix for https://github.com/wasp-lang/wasp/issues/2368
       -- We should remove this code once we implement a proper solution.
-      ExceptT $ updateJsonFile removeWaspConfigDevDependency packageJsonInBuildDir
-      ExceptT $ updateJsonFile removeAllMentionsOfWaspConfig packageLockJsonInBuildDir
+      ExceptT $ updateJsonFile removeWaspConfigFromDevDependenciesArray packageJsonInBuildDir
+      ExceptT $ updateJsonFile removeAllMentionsOfWaspConfigInPackageLockJson packageLockJsonInBuildDir
 
-    removeAllMentionsOfWaspConfig :: Value -> Value
-    removeAllMentionsOfWaspConfig packageLockJsonObject =
+    removeAllMentionsOfWaspConfigInPackageLockJson :: Value -> Value
+    removeAllMentionsOfWaspConfigInPackageLockJson packageLockJsonObject =
+      -- We want to:
+      --   1. Remove the `wasp-config` dev dependency from the root package in package-lock.json.
+      --   This is at `packageLock["packages"][""]["wasp-config"]`.
+      --   2. Remove all package location entries for the `wasp-config` package
+      --   (i.e., entries whose location keys end in `/wasp-config`).
+      --   Example locations include:
+      --      packageLock["packages"]["../../data/packages/wasp-config"]
+      --      packageLock["packages"]["node_modules/wasp-config"]
+      --      packageLock["packages"]["/home/filip/../wasp-config"]
       packageLockJsonObject
-        & key "packages" . key "" %~ removeWaspConfigDevDependency
+        & key "packages" . key "" %~ removeWaspConfigFromDevDependenciesArray
         & key "packages" . _Object
           %~ HM.filterWithKey
-            (\packageKey _ -> not ("wasp-config" `isInfixOf` packageKey))
+            (\packageLocation _ -> not $ isWaspConfigPackageLocation packageLocation)
 
-    removeWaspConfigDevDependency :: Value -> Value
-    removeWaspConfigDevDependency original =
+    isWaspConfigPackageLocation :: Text -> Bool
+    isWaspConfigPackageLocation packageLocation =
+      (FP.pathSeparator : "wasp-config") `isSuffixOf` unpack packageLocation
+
+    removeWaspConfigFromDevDependenciesArray :: Value -> Value
+    removeWaspConfigFromDevDependenciesArray original =
       original & key "devDependencies" . _Object . at "wasp-config" .~ Nothing
 
 buildIO ::
--- a/waspc/cli/src/Wasp/Cli/Command/Build.hs
+++ b/waspc/cli/src/Wasp/Cli/Command/Build.hs
@@ -3,10 +3,17 @@ module Wasp.Cli.Command.Build
   )
 where
 
+import Control.Lens
 import Control.Monad (unless, when)
-import Control.Monad.Except (throwError)
+import Control.Monad.Except (ExceptT (ExceptT), liftEither, runExceptT, throwError)
 import Control.Monad.IO.Class (liftIO)
-import StrongPath (Abs, Dir, Path', castRel, (</>))
+import Data.Aeson (Object, Value (..), eitherDecode, encode, (.=))
+import Data.Aeson.Lens
+import Data.Aeson.Types (object)
+import qualified Data.HashMap.Strict as HM
+import Data.Maybe (fromJust)
+import Data.Text (Text, isInfixOf)
+import StrongPath (Abs, Dir, File, Path', castRel, (</>))
 import Wasp.Cli.Command (Command, CommandError (..))
 import Wasp.Cli.Command.Compile (compileIOWithOptions, printCompilationResult)
 import Wasp.Cli.Command.Message (cliSendMessageC)
@@ -20,6 +27,7 @@ import qualified Wasp.Message as Msg
 import Wasp.Project (CompileError, CompileWarning, WaspProjectDir)
 import Wasp.Project.Common (buildDirInDotWaspDir, dotWaspDirInWaspProjectDir, generatedCodeDirInDotWaspDir, packageJsonInWaspProjectDir, packageLockJsonInWaspProjectDir, srcDirInWaspProjectDir)
 import Wasp.Util.IO (copyDirectory, copyFile, doesDirectoryExist, removeDirectory)
+import qualified Wasp.Util.IO as IOUtil
 
 -- | Builds Wasp project that the current working directory is part of.
 -- Does all the steps, from analysis to generation, and at the end writes generated code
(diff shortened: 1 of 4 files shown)


waspc/cli/src/Wasp/Cli/Command/Build.hs
```diff
@@ -58,40 +73,65 @@ build = do
...
+          (waspProjectDir </> packageLockJsonInWaspProjectDir)
+          packageLockJsonInBuildDir
 
-      copyDirectory
-        (waspProjectDir </> dotWaspDirInWaspProjectDir </> generatedCodeDirInDotWaspDir </> sdkRootDirInGeneratedCodeDir)
-        (buildDir </> sdkRootDirInGeneratedCodeDir)
+      -- A hacky quick fix for https://github.com/wasp-lang/wasp/issues/2368
+      -- We should remove this code once we implement a proper solution.
+      ExceptT $ updateJsonFile removeWaspConfigDevDependency packageJsonInBuildDir
+      ExceptT $ updateJsonFile removeAllMentionsOfWaspConfig packageLockJsonInBuildDir
```
sodic, line 118: I originally had two functions which I called from the main body of the build function:
- `copyFilesNecessaryForDockerBuild` that copies everything.
- `editNpmFilesForDockerBuild` that edits package.json and package-lock.json.

This looked a little cleaner but forced me to define `packageJsonInBuildDir` and `packageLockJsonInBuildDir` twice, so I decided to stick with a single function (since the second one heavily relies on the first one being called beforehand, and they share some variables).

waspc/cli/src/Wasp/Cli/Command/Build.hs
```diff
@@ -58,40 +73,65 @@ build = do
...
+        & key "packages" . _Object
+          %~ HM.filterWithKey
+            (\packageKey _ -> not ("wasp-config" `isInfixOf` packageKey))
 
-      copyFile
-        (waspProjectDir </> packageLockJsonInWaspProjectDir)
-        (buildDir </> castRel packageLockJsonInWaspProjectDir)
+    removeWaspConfigDevDependency :: Value -> Value
+    removeWaspConfigDevDependency original =
+      original & key "devDependencies" . _Object . at "wasp-config" .~ Nothing
```
sodic, line 130: This is where all the lens stuff is.

waspc/src/Wasp/Util/Json.hs
```diff
@@ -16,3 +25,12 @@ parseJsonWithComments jsonStr = do
...
     _exitFailure -> return $ Left stderr
+
+updateJsonFile :: (Value -> Value) -> Path' Abs (File a) -> IO (Either String ())
+updateJsonFile updateFn jsonFilePath = runExceptT $ do
+  jsonContent <- ExceptT $ eitherDecode <$> IOUtil.readFileBytes jsonFilePath
+  let updatedContent = updateFn jsonContent
+  liftIO $ writeJsonValue jsonFilePath updatedContent
+
+writeJsonValue :: Path' Abs (File f) -> Value -> IO ()
+writeJsonValue file = IOUtil.writeFileBytes file . encode
```
sodic, line 36: I thought about whether I should put this here or in IOUtil, but I decided this makes more sense in the end.

waspc/cli/src/Wasp/Cli/Command/Build.hs
```diff
@@ -58,40 +73,65 @@ build = do
...
 
-  liftIO $ copyUserFilesNecessaryForBuild waspProjectDir buildDir
-
-  cliSendMessageC $
-    Msg.Success "Your wasp project has been successfully built! Check it out in the .wasp/build directory."
+  liftIO (prepareFilesNecessaryForDockerBuild waspProjectDir buildDir) >>= \case
+    Left err -> throwError $ CommandError "Failed to prepare files necessary for docker build" err
+    Right () ->
+      cliSendMessageC $
+        Msg.Success "Your wasp project has been successfully built! Check it out in the .wasp/build directory."
```
infomiho, line 80:
```suggestion
  liftIO (prepareFilesNecessaryForDockerBuild waspProjectDir buildDir) >>= \case
    Left err -> throwError $ CommandError "Failed to prepare files necessary for docker build" err
    Right () -> return ()
  cliSendMessageC $
    Msg.Success "Your wasp project has been successfully built! Check it out in the .wasp/build directory."
```
Maybe pull out the success msg so in the future we can add more steps before without a rewrite?

waspc/cli/src/Wasp/Cli/Command/Build.hs
```diff
@@ -58,40 +73,65 @@ build = do
...
-      copyFile
-        (waspProjectDir </> packageJsonInWaspProjectDir)
-        (buildDir </> castRel packageJsonInWaspProjectDir)
+    removeAllMentionsOfWaspConfig :: Value -> Value
+    removeAllMentionsOfWaspConfig packageLockJsonObject =
+      packageLockJsonObject
+        & key "packages" . key "" %~ removeWaspConfigDevDependency
+        & key "packages" . _Object
+          %~ HM.filterWithKey
+            (\packageKey _ -> not ("wasp-config" `isInfixOf` packageKey))
```
infomiho, line 126: Could you maybe explain what is going on here, maybe in a comment above the since it's quite hard to grasp.

"Here we are removing this from X and then we are doing ..."

infomiho (approved):
LGTM, I'd just write a comment above the lens stuff (the intention of the code)

waspc/cli/src/Wasp/Cli/Command/Build.hs
Martinsos: Ok, yeah I guess it doesn't make a big difference. Although you could also have written helper functions that produce `packageJsonInBuildDir` and the other one for you when needed, to avoid duplication.

What I wonder more about is the name `prepareFilesNecesasryForDockerBuild`-> some of them are necessary for docker build, but that is not their only purpose, right? If we decided we are not producing a Dockerfile for users anymore, and they are supposed to write their own, could I just delete this whole function and would everything work fine, in the sense that all it is doing is only needed for Dockerfile? If so, then this name is fine. If not, I would ideally split logic into two pieces: first that copies over what needs to be copied in general, regardless if it will be used in Dockerfile or not, and then the other part is additional logic that is neccessary specifically in order to support Dockerfile.

waspc/cli/src/Wasp/Cli/Command/Build.hs
```diff
@@ -58,40 +73,65 @@ build = do
...
-        (buildDir </> sdkRootDirInGeneratedCodeDir)
+      -- A hacky quick fix for https://github.com/wasp-lang/wasp/issues/2368
+      -- We should remove this code once we implement a proper solution.
+      ExceptT $ updateJsonFile removeWaspConfigDevDependency packageJsonInBuildDir
+      ExceptT $ updateJsonFile removeAllMentionsOfWaspConfig packageLockJsonInBuildDir
 
-      copyFile
-        (waspProjectDir </> packageJsonInWaspProjectDir)
-        (buildDir </> castRel packageJsonInWaspProjectDir)
+    removeAllMentionsOfWaspConfig :: Value -> Value
```
Martinsos, line 120: I would be tempted to name this `removeAllMentionsOfWaspConfigInPackageLockJson`, to make this assumption / requirement clear. Because it does assume that it is operating on package lock json. But it is a local function usable only in limited scope so no biggie, shorter name can also work fine, it is just that if you look solely at it, you are missing that piece of information (arg name does hint at it, but I like to look at signature + comments as the sole API of the function, at least in Haskell).

waspc/cli/src/Wasp/Cli/Command/Build.hs
Martinsos: Yeah hm. I guess this is on us since we don't know lenses well, but they are advanced Haskell concept I guess, at least for our team at the moment.

Reading this, I get taht in object, under "packages" key, we remove Wasp config dev dependency, althogh that `key ""` confuses me, what is empty key hm? If there wasn't `key ""` in that first part, I think it would need no comment.

Second part also starts good, `key "packages"`, I can disregard `_Object` or take it as "ok, in the object", then we filter by keys I guess, ok then `packageKey` confuses me, I would either go with `key` or `packageName`, and then we say key has to have `wasp-config` as infix. Ok, so we filter/keep those without wasp-config as infix. This was actually readable, although I would change `packageKey` as I suggested + maybe just go with
```hs
HM.filterWithKey (not . ("wasp-config" `isInfixOf`))
```
to reduce the noise.

I would also add short comments though, just to be safe, but would also make code a bit more readable as suggested.

Btw I am always wary when I see usage of `isInfixOf` like this, or prefix or whatever, as often it is too wide of a net. Why `isInfixOf`? Why do you can't be more specific, I guess there are some extra characters in this string? But `isInfixOf` is bad because what if they package called `owasp-configuration` that has nothing to do with wasp? Yeah, doesn't seem great, I would make it more specific.

waspc/cli/src/Wasp/Cli/Command/Build.hs
```diff
@@ -58,40 +73,65 @@ build = do
...
+        & key "packages" . key "" %~ removeWaspConfigDevDependency
+        & key "packages" . _Object
+          %~ HM.filterWithKey
+            (\packageKey _ -> not ("wasp-config" `isInfixOf` packageKey))
 
-      copyFile
-        (waspProjectDir </> packageLockJsonInWaspProjectDir)
-        (buildDir </> castRel packageLockJsonInWaspProjectDir)
+    removeWaspConfigDevDependency :: Value -> Value
+    removeWaspConfigDevDependency original =
```
Martinsos, line 129: Similar applies as for function above.

Ok interesting, this function is a bit more general then the one above -> the one above operates on packageLockJson, while this one operates on any object that has devDependecies in it. I would, similar like for function above, make it clear what this function expects to operate on, in its name, and if not, at least in the comment.

waspc/cli/src/Wasp/Cli/Command/Build.hs
Martinsos: Ok, this is readable I think! I do wonder if there is not sometihng, at lens level, that would allow you to unset `wasp-config` field, instead of setting it to Nothing? I guess this works, so not a biggie, but that would be even more readable + a bit more direct. If not, no worries. I mean it is a key in hashmap so I am sure it is doable hm.

waspc/src/Wasp/Util/IO.hs
```diff
@@ -115,6 +116,9 @@ readFileStrict = T.IO.readFile . SP.toFilePath
 writeFile :: Path' Abs (File f) -> String -> IO ()
 writeFile = P.writeFile . SP.fromAbsFile
 
+writeFileBytes :: Path' Abs (File f) -> B.ByteString -> IO ()
+writeFileBytes = B.writeFile . SP.fromAbsFile
```
Martinsos, line 120: Looking at this it looks as if it should be under sometihng like StrongPath/IO, but ok, I see other similar stuff is here, so this is consistent with it, we would have to move evertying with would now be hassle.

waspc/src/Wasp/Util/Json.hs
```diff
@@ -16,3 +25,12 @@ parseJsonWithComments jsonStr = do
   case exitCode of
     ExitSuccess -> return $ decodeFromString response
     _exitFailure -> return $ Left stderr
+
+updateJsonFile :: (Value -> Value) -> Path' Abs (File a) -> IO (Either String ())
+updateJsonFile updateFn jsonFilePath = runExceptT $ do
+  jsonContent <- ExceptT $ eitherDecode <$> IOUtil.readFileBytes jsonFilePath
+  let updatedContent = updateFn jsonContent
+  liftIO $ writeJsonValue jsonFilePath updatedContent
```
Martinsos, line 33:
```suggestion
  liftIO $ writeJsonValue jsonFilePath $ updateFn jsonContent
```

Martinsos (commented):
@sodic looks good! This actually turned out to be quite simple merely a bit of manipulation on json files and that is it really. Code is also quite simple due to lenses, couple of lines.

Most stuff I commented are minor improvements in code cleanliness, but one is actually a potentiall bug in hiding (although not likely to happen), which is why I didn't approve it yet. Let's fix that and whatever code improvements you like and then we can do it.

waspc/cli/src/Wasp/Cli/Command/Build.hs
sodic: I couldn't find it, and I don't think it's worth exploring further since this is a temporary fix and we want to get it out ASAP.

If that's ok with you, please resolve

waspc/cli/src/Wasp/Cli/Command/Build.hs
sodic: Sure, that's what we had before. I'll change it back.

waspc/cli/src/Wasp/Cli/Command/Build.hs
sodic: > "Here we are removing this from X and then we are doing ..."

Sure, will do!

> althogh that key "" confuses me, what is empty key hm?

[Answer](https://youtu.be/WyYC9qRlJxY?t=205)

I'll add a comment for that too.

image

> Btw I am always wary when I see usage of isInfixOf

What would you suggest? Here are the keys it has to remove:
```
../../data/packages/wasp-config
node_modules/wasp-config
/home/filip/.../wasp-config
<Other possible paths on user machines>
```
`isSuffixOf` with `/wasp-config` (notice the slash) might be enough, but arguably it's better to delete something extra than miss something, so keep that in mind (I don't know what might happen with a trailing slash, although it's safe to guess we won't have to deal with it).

Finally, keep in mind this is code we want to get rid of in a couple of weeks tops, so getting to clever here probably isn't worth it.

Either way, I'll document the strings I expect to match, you would have been able to give a concrete suggestion immediately if I'd done so in the first place.

> To reduce the noise:
> ```haskell
> HM.filterWithKey (not . ("wasp-config" `isInfixOf`))
>```

Hm, I don't think this will work.

I remember trying all kinds of compositions and functions, and lambda ended up the cleanest one that compiles. I think this one won't work because it doesn't take into account that the callback receives two arguments (key and value).

I managed to avoid the lambda with some combination of composition, `const`, and `uncurry`, but it was much less readable IMO.

>  I would either go with `key` or `packageName`

`packageName` would be incorrect (see examples of keys I listed above). `key` is ok, but sounded less informative than `packageKey`. I can change it to `key` if that's what you prefer, I'm fine either way.

> I would also add short comments though, just to be safe, but would also make code a bit more readable as suggested.

I'm sorry I can't resist... You know you're working in a state-of-the-art language when you need comments to understand code that changes a JSON property :smile: 

And yes, I agree we need comments, I'll add some.

waspc/cli/src/Wasp/Cli/Command/Build.hs
sodic: > Although you could also have written helper functions that produce packageJsonInBuildDir and the other one for you when needed, to avoid duplication.

I've considered this but decided it would fall into overengineering. Do you think it would be an improvement?

> If we decided we are not producing a Dockerfile for users anymore, and they are supposed to write their own, could I just delete this whole function and would everything work fine, in the sense that all it is doing is only needed for Dockerfile?

Yes, I believe that's the case. The only reason we're copying these files is because we need them in Docker's build context. **If we weren't using docker, we wouldn't need this function.** This is already documented at the [beginning of the function](https://github.com/wasp-lang/wasp/pull/2372/files#diff-ea4b4106c35cd1478bae7f04a9c2a368284228df2e37ae27aa4894b8c8f773c9R83), and we soon plan to remove it altogether.

Did I understand the question correctly?

waspc/cli/src/Wasp/Cli/Command/Build.hs
sodic: Yes, nice catch. I'll change the name.

waspc/cli/src/Wasp/Cli/Command/Build.hs
sodic: Yeah, I actually reuse this one inside `removeAllMentionsOfWaspConfig`.

Which name would you suggest for this one? `removeWaspConfigFromDevDependenciesArray`?

waspc/src/Wasp/Util/IO.hs
sodic: Ok, I'm resolving

waspc/cli/src/Wasp/Cli/Command/Build.hs
Martinsos: Yup, good with me.

waspc/cli/src/Wasp/Cli/Command/Build.hs
Martinsos: Oh wow it is actually a "" key, ok I didn't expect that, thought it is some kind of quirky thing you had to do to get the lenses to do what you want, but it is actually exactly what it looks like.

infixOf -> seems like you want to target a path that ends up with `/wasp-config`. So then go with that -> check if `/wasp-config` is a suffix. Not sure about the trailing slash? Add logic for that, can be just simple OR. Be careful not to hardcode file seprator, use the one from FilePath.</> so it is system agnostic (unless they are guaranteed to be posix by package.json, in that case go with posix).
This is temporary fix, but you how temporary fixes are, they can stay in longer than one would expect. Also, this is so simple and easy to fix, and final solution is equally simple, we are certainly not overengineering here. And, there is value in me pointing this out to you, because you are less likely to make the same mistake next time.
Better to delete too much than too little -> probably depends on the context, but I don't agree in this case, because this can result in some nasty bugs, in the sense that they are hard to figure out because this is super unexpected behaviour. And I don't think going with suffix we have much chance to miss something, I would call it strictly and obviously better solution with the same level of complexity.

That lambda with packageKey -> my bad, I missed that `_` there, thought it is just one argument, I agree not worth bothering with it.

`packageKey` -> ok, so those are paths actually from what I get? I mean what is that, how would you describe if I you had to? `key` is correct because it is key in the object. `packageKey` doesn't sound correct because I doubt that thing is called `package key`. Obviously also not a name, as you pointed out -> when I was suggesting that I didn't really know what is it. Now I would be tempted to call it `packagePath` or `packageUri` or something like that. I would go with `key` if you wish to keep it simple and correct, or I would check the npm docs, figure out what can be in that string + waht they call it, and call it like that.

I know you were making a joke regarding JSON and comments, but what lenses do here is quite impressive to me actually, you can't do taht in Javascript in one line of code while not mutating the origianl object, and spread operators can also be a pain to write when you have to go couple of levels deep, while ont being reusable or anything.

waspc/cli/src/Wasp/Cli/Command/Build.hs
Martinsos: Yup, took me a bit to figure out it is reused. Yes, that sounds like ok name.

waspc/src/Wasp/Util/Json.hs
Martinsos: Keep it as you like, this was just an idea. I like this better because it just flows, the operations, from right to left, in one line. While in two lines it is broken a bit. But really this super silly nitpicking, I am aware, so whatever you like better.

waspc/cli/src/Wasp/Cli/Command/Build.hs
sodic: `unless` was ugly because I'd have to force `fromLeft` with a bogus return value, so I went with @infomiho's suggestion.

Martinsos (approved):
LGTM!

PR #2375, opened 2024-11-19
Update env variables docs

### LINKED ISSUE #1894: Document our environment variables

infomiho:
Updates the Project and Deployment env variables docs. I've tried to deduplicate as much as possible. I've enumerate all the env variables that Wasp defines.

Closes #1894


web/docs/project/env-vars.md
```diff
@@ -2,61 +2,165 @@
...
 </Tabs>
 
+Read more about the `env` object in the [API reference](#client-env-vars-1).
 
-Check below on how to define them.
+### Wasp Client Env Vars
 
+Here are the client env vars that Wasp defines:
+
+#### General Configuration
```
Martinsos, line 49: We haven't introduced earlier this concept of "general configuration env vars" though, right? Should we, or it is self evident?

web/docs/project/env-vars.md
```diff
@@ -2,61 +2,165 @@
...
+
+<EnvVarsTable>
+  <EnvVar name="REACT_APP_API_URL" type="URL" isRequired={false} defaultValue="http://localhost:3001" note="The client uses this as the server URL." />
+</EnvVarsTable>
```
Martinsos, line 54: Should we mention that this is it, only one? Somehow it feels like waht now, where is the rest. Maybe we can mention that this is it but you can also deifne your own. Ok maybe that is silly, I am not sure, but this somehow felt unfinished.

web/docs/project/env-vars.md
```diff
@@ -2,61 +2,165 @@
...
-Check below on how to define them.
+Read more about the `env` object in the [API reference](#server-env-vars-1).
+
+### Wasp Server Env Vars
+
+#### General Configuration
+
+<EnvVarsTable>
+    <EnvVar name="DATABASE_URL" type="String" isRequired={true} note="The URL of the PostgreSQL database you want your app to use." />
+    <EnvVar name="WASP_WEB_CLIENT_URL" type="URL" isRequired={true} note="Used as your client URL. The server needs to know about it to properly configure Same-Origin Policy (CORS) headers." />
```
Martinsos, line 87: And some other stuff I think, for example provide a link to the client in some automatic emails (auth)? Maybe not any more if flow is completely server-side, but I would maybe state is to that it sound slike CORS is not or might not be the only thing this is important for. I wonder if we should even explain why server needs it -> that info can easily become obsolete, maybe better to skip it? Not a biggie though.

web/docs/project/env-vars.md
```diff
@@ -2,61 +2,165 @@
...
+    <EnvVar name="MAILGUN_API_KEY" type="String" isRequired={true} note="The Mailgun API key." />
+    <EnvVar name="MAILGUN_DOMAIN" type="String" isRequired={true} note="The Mailgun domain." />
+    <EnvVar name="MAILGUN_API_URL" type="URL" isRequired={false} note={
+      <span>Useful if you want to use the EU API endpoint (<code>https://api.eu.mailgun.net</code>).</span>
+    } />
+</EnvVarsTable>
+
+#### Auth
+
+You can use this env while developing to skip email verification. This is useful when you are testing the email sending functionality in development:
```
Martinsos, line 131: You are talking here about a single specific env var? That is a bit weird because below we have a whole table. Feels like you are talking about the whole table. Maybe put it all in the note for that one instead?

Martinsos, line 131:
```suggestion
You can use this env var during development to skip email verification. This is useful when you are testing the email sending functionality in development:
```
The part "this is useful when you are testing the email sending functionality in development" also confuses me -> how is this useful if I am testing the email sending functionality? I guess it is useful if I am not testing email sending functinoality? Hmmmm

web/docs/project/env-vars.md
```diff
@@ -2,61 +2,165 @@
...
+
+<!-- TODO: this not really Boolean, it's really a string that accepts "true" as only true value -->
```
Martinsos, line 133: Ay yay let's make that clear?

web/docs/project/env-vars.md
```diff
@@ -2,61 +2,165 @@
...
+
+#### Jobs
+
+<EnvVarsTable>
+    <EnvVar name="PG_BOSS_NEW_OPTIONS" type="String" isRequired={false} note={
+        <span>It's parsed as JSON. Enables you to provide <a href="../advanced/jobs#declaring-jobs">custom config</a> for PgBoss.</span>
+    } />
+</EnvVarsTable>
+
+#### OAuth Configuration
```
Martinsos, line 147: Can we put this under Auth header?

web/docs/project/env-vars.md
```diff
@@ -2,61 +2,165 @@
...
+If you are using OAuth, you need to provide the following environment variables:
+
+<EnvVarsTable>
+    <EnvVar name="<PROVIDER_NAME>_CLIENT_ID" type="String" isRequired={true} note="The client ID provided by the OAuth provider." />
+    <EnvVar name="<PROVIDER_NAME>_CLIENT_SECRET" type="String" isRequired={true} note="The client secret provided by the OAuth provider." />
+</EnvVarsTable>
+
+ `<PROVIDER_NAME>` is the uppercase name of the provider you are using. For example, if you are using Google OAuth, you need to provide the `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET` environment variables.
+
+If you are using [Keycloak](../auth/social-auth/keycloak.md), you'll need to provide one extra environment variable:
```
Martinsos, line 158: Should we create special header for Keycloak?

web/docs/project/env-vars.md
```diff
@@ -86,13 +190,11 @@ In the root of your Wasp project you can create two distinct files:
     REACT_APP_SOME_VAR_NAME=somevalue
     ```
 
+    <ClientEnvVarsNote />
+
 `.env.server` should not be committed to version control as it can contain secrets, while `.env.client` can be versioned as it must not contain any secrets.
 By default, in the `.gitignore` file that comes with a new Wasp app, we ignore all dotenv files.
 
-:::info Dotenv files
```
Martinsos, line 92: Where did this go, it wasn't useful?

web/docs/project/env-vars.md
```diff
@@ -104,43 +206,70 @@ Defining environment variables in this way can be cumbersome even for a single p
 
 ## Defining Env Vars in Production
 
-While in development, we had the option of using `.env.client` and `.env.server` files which made it easy to define and manage env vars.
```
Martinsos, line 107: Aha this went away because it is now mentinoed in deployment/env-vars.md?

web/docs/project/env-vars.md
```diff
@@ -104,43 +206,70 @@ Defining environment variables in this way can be cumbersome even for a single p
...
-However, for production, `.env.client` and `.env.server` files will be ignored, and we need to provide env vars differently.
+Defining env variables in production will depend on where you are deploying your Wasp project. In general, you will define them via mechanisms that your hosting provider provides.
+
+We'll talk about how to define env vars for each deployment option in the [deployment section](../deployment/env-vars.md).
```
Martinsos, line 211: We'll talk -> why is it future tense?

web/docs/project/env-vars.md
```diff
@@ -104,43 +206,70 @@ Defining environment variables in this way can be cumbersome even for a single p
...
+
+## Custom Env Var Validations
+
+TODO: when the Zod validation PRs are merged, describe how users can define their own validations.
 
-![Env vars usage in development and production](/img/env/prod_dev_fade_2.svg)
```
Martinsos, line 110: What happened to this image?

web/docs/project/EnvVarsTable.tsx
Martinsos, line 1: Nice this is cool! I found it a bit weird that I have to construct EnvVar elements outside and then provide them as children, to me it sounded more natural to provide a list of EnvVar objects to the EnvVarsTable and then it would call `.map` on those and produce a bunch of EnvVar elements that show those. But ok it is quite unimportant.

Martinsos (commented):
Nice stuff! Once we solve all these comments, I will take another look but more from the high above, to get a better grasp for where is what and what are the relations, now I was focusing more on the details. But I think the question is similar like in other PR: Do we do A.P B.P and then link from P to those, or do we do P.A and P.B and then link from A and B to those, where are parts of docs about Production and A and B are some features (check the other PR for more info).

web/docs/deployment/deployment-methods/_addExternalAuthEnvVarsReminder.md
```diff
@@ -1,3 +1,3 @@
 :::tip Using an external auth method?
-If your app is using an external authentication method(s) supported by Wasp (such as [Google](../../auth/social-auth/google#4-adding-environment-variables) or [GitHub](../../auth/social-auth/github#4-adding-environment-variables)), make sure to additionally set the necessary environment variables specifically required by these method(s).
+If your app is using an external authentication method(s) supported by Wasp (such as [Google](/auth/social-auth/google.md#4-adding-environment-variables) or [GitHub](/auth/social-auth/github.md#4-adding-environment-variables)), make sure to additionally set the necessary environment variables specifically required by these method(s).
```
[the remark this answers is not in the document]
infomiho: I've been using this for some time now and it works great 👍 , it's explained in Writing Guide https://github.com/wasp-lang/wasp/blob/main/web/docs/writingguide.md#linking-to-pages-in-the-docs

web/docs/project/env-vars.md
infomiho: It didn't feel useful, it talks about the specific Node.js package and they are not going to use it directly. I think the env vars file format is self-explanatory.

web/docs/project/env-vars.md
infomiho: I guess so, yes 😄 I moved things around so it made sense when reading the docs.

web/docs/project/env-vars.md
infomiho: It moved to the deployment env vars page :)

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -56,37 +56,8 @@ You'll have to [switch to PostgreSQL](../../data-model/backends#migrating-from-s
...
-  You only need this environment variable if you're using Wasp's `auth` features.
-  Set it to a random string at least 32 characters long (you can use an [online generator](https://djecrety.ir/)).
-
-- `PORT`
-
-  The server's HTTP port number. This is where the server listens for requests (default: `3001`).
-
-
-<AddExternalAuthEnvVarsReminder />
+Check the [required server env variables](../env-vars.md#server-env-vars) and make sure they are set up in your production environment.
```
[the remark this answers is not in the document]
infomiho: This section is now:

> There's a Dockerfile that defines an image for building the server in the `.wasp/build` directory.

> To run the server in production, deploy this Docker image to a hosting provider and make sure the required env variables are correctly set up. Usually, you use the provider's dashboard UI or a CLI tool to set up these env variables.

> Check the [required server env variables](../env-vars.md#server-env-vars) and make sure they are set up for your server.

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -96,8 +67,6 @@ While these are the general instructions on deploying the server anywhere, we al
 
 The command above will build the web client and put it in the `build/` directory in the `.wasp/build/web-app/`.
 
-This is also the moment to provide any additional env vars for the client code, next to `REACT_APP_API_URL`. Check the [env vars docs](../../project/env-vars#client-env-vars-1) for more details.
```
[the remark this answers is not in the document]
infomiho: Yes 👍

web/docs/project/env-vars.md
infomiho: This is where we introduce it :)

web/docs/project/env-vars.md
infomiho: It's following the same structure as the server env vars, I think this is understandable.

web/docs/project/env-vars.md
infomiho: I'll update this once we merge the env vars PR since the type depends on what we decide we want to do there. #2362

web/docs/project/env-vars.md
infomiho: I've moved the `Auth` stuff to `Development` and now this feels better i.e. I named it `OAuth providers`.

web/docs/project/env-vars.md
infomiho: I feel like it would confuse people, just keeping it under `OAuth Providers` makes it feel connected to the rest of the stuff.

web/docs/project/EnvVarsTable.tsx
infomiho: Screenshot 2024-12-11 at 15 13 10

I got inspired by ShadCN's way of composing things.

web/docs/deployment/deployment-methods/_addExternalAuthEnvVarsReminder.md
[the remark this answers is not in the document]
Martinsos: That is suprising to me because it sounds like it goes against theri docs. Just to confirm very explicitly: you can confirm this works if we release new versioned docs and these become numbered as let's say 0.15? Will these links, when clicked from 0.15 version of docs, correctly take you to the 0.15 version of these files, or instead incorrectly to the most current docs?

web/docs/project/env-vars.md
Martinsos: Hm but why did I ask this then? I believe I asked it because you don't really explain it at any moment, you just go for it and start listing them. That is when I was reading it, I was thinking "oh ok general configuration, what is that" and then bam, vars immediately.
Which is why I asked above, should we explain what it is or you think it is self evident - when somebody sees "general configruation", they understand what these env vars are?

web/docs/project/env-vars.md
Martinsos: Did this ever happen (I don't see a rocket :rocket: )?

web/docs/project/env-vars.md
Martinsos: Ok, as long as you don't forget about it!

web/docs/project/env-vars.md
Martinsos: They might need to understand how to use it though, to for example pass JSON values and similar, we had those situations, how to quote stuff. So I thoguht it is good to introduce the concept, since we are using it as na important part of Wasp, so it is clear this is not something we made up, but instead there are more resources behind it that they can look for. Drop it if you really want, but to me it seems useful to keep it.

web/docs/project/EnvVarsTable.tsx
Martinsos: Right, but they can combine multiple things probably. You instead have a single item that can go in. So no composability, but you opened space for throwing in random elements that don't make sense.

Martinsos (commented):
I did another round of comments. Once we complete those, I will as a final step take a higher look from above, probably run the docs and see how they read now.
As a preparation for that, what do you say about what I asked here https://github.com/wasp-lang/wasp/pull/2375#pullrequestreview-2469721287 ? What is your view on the current structure, how do you feel about it?

web/docs/project/env-vars.md
infomiho: Yep, it happened, sorry 🙋‍♂️

web/docs/project/env-vars.md
infomiho: I get what you mean, if people know what's underneath they can Google better. 

The env variables are first parsed by Haskell and then outputted to the `.env` files in the server and the client apps. If we wanted to provide full information, we should first mention Haskell's [dotenv](https://hackage.haskell.org/package/dotenv) package and then our logic for outputting the vars i.e. our `envVarsToDotEnvContent` function. And then we can mention that the variables are ingested by Node.js with the JS [dotenv](https://www.npmjs.com/package/dotenv) package.

I dislike giving this many details to users because 99% of them don't really need them and those with issues will most likely ask about it in our Discord.

web/docs/project/EnvVarsTable.tsx
infomiho: Compare
Screenshot 2024-12-16 at 15 55 03
vs.
Screenshot 2024-12-16 at 15 55 11

One of them is more readable to me, so I'd like to keep the current API. It's our internal component for docs, no need to make it so strict IMHO.

web/docs/project/env-vars.md
Martinsos: I wrote this because when I was reading it I was confused for a bit, so I didn't make it up. Am I representative of a typical user, hard to say. Which is why I asked how you feel about it.

If I would write sometihng, I wouldn't write what you said because it is a bit much, especially the last part with "hence". I might write someting like "Env variables used by Wasp to configure general aspects of your app, not tied to one specific feature.".
But again, is that useful or not -> not sure. I believe I provided my perspective well enough here, I will leave the final decision to you.

web/docs/project/EnvVarsTable.tsx
Martinsos: It is more readable because you used different vars! The second example has less vars, the first and second, and third and forth, have equal length of name, and they all have much shorter notes. So there is less content, it is better aligned, and on top of all that it allowed you to zoom in so that also looks better.
Finally, your editor seems to do a better job with syntax coloring in the second example, but I don't see why it couldn't do coloring well also in the first example, I just tried this in my emacs and it does the coloring properly. Maybe that `<span>` in that note causes you issues.

Take a look at what it looks in my emacs, when I use the same env vars:

image

image

Those look exactly the same regarding readibility to me. If anything, you could argue that the second one has a bit less noise due to less repetitive HTML tags, although I don't think it matters.

Also, while doing the comparison, you ignored my argument about the second approach (one with children) allowing for elements that you might not expect, while the first one doesn't allow that.

Keep the API as you like it, as I said in my initial comment, this is quite unimportant, but still I think this discussion is useful as a learning opportunity and didn't want it to end with wrong arguments.

Martinsos (commented):
Responded again.

web/docs/project/EnvVarsTable.tsx
infomiho: Updated the API 👍

web/docs/project/env-vars.md
infomiho: Added `These are some general env variables used for various Wasp features:` 👍

Martinsos (approved):

PR #2376, opened 2024-11-19
Production database docs

infomiho:
The idea was to talk about general production database concepts: what Wasp expects, how migrations work and how to connect to the production DB.

### DIFF
--- a/web/docs/deployment/database.md
+++ b/web/docs/deployment/database.md
@@ -2,10 +2,66 @@
 title: Database
 ---
 
-TODO: we'll explain the database in the context of deployment
+In this section, we'll discuss what happens with the database when your app goes live. When you develop your app locally, you probably use the dev database that you start with the `wasp db start` command. However, when it's time to deploy your app, you'll need to set up a production database.
 
-TODO: the only requirement from Wasp's point of view - it has to be a PostgreSQL database accessible from the server via the `DATABASE_URL` environment variable
+### Production database requirements
 
-TODO: we'll talk about migrations: how are they created in development, how are they applied in production, how to debug failed migrations
+The server app that Wasp generates uses a PostgreSQL database. The only requirement from Wasp's point of view is that the database is accessible from the server via the `DATABASE_URL` environment variable.
 
-TODO: we'll talk about how to open the DB studio in development and production
+It can be a PostgreSQL database running on the same server as the server app, or it can be a managed PostgreSQL database service like [Fly Postgres](https://fly.io/docs/postgres/), [AWS RDS](https://aws.amazon.com/rds/), or some other service.
+
+## Migrations
+
+Every time you make a change in your [Prisma schema](../data-model/prisma-file.md) e.g. adding a new model, changing a field type, etc., you need to create a migration. Migrations are some code that describes the change you made in the schema, and they are used to apply the change to the database.
+
+The benefit of migrations is that you can apply the same change to multiple databases, e.g. your local development database and your production database. 
+
+### Creating migrations
+
+After you made a change in the Prisma schema, you can create a migration by running the following command:
+
+```bash
+wasp db migrate-dev
+```
+
+This command will create a new migration in the `migrations` directory. The migration is a set of SQL commands that describe the change you made in the schema.
+
+### Applying migrations
+
+**In development**, the migrations are applied as soon as you run the `wasp start` command. 
+
+**In production**, the server app first checks if there are any new migrations that need to be applied, and if there are, it applies them before starting the server. This way, the database schema is always in sync with the Prisma schema.
+
+The migrations might fail to apply if there is a conflict with the existing data in the database. In that case, you'll need to fix the migration and try again.
+
+### Debugging failed migrations
+
+If a migration fails to apply, the server app will log the error message and stop. You should then connect to the production database and see what went wrong. If you check the `_prisma_migrations` table, you'll see the failed migration there. 
+
+You can try resolving the erorr e.g. if you tried adding a `@unique` constraint to a field that already has duplicate values:
+1. Remove any duplicate values from the database
+2. Remove the failed migration from the `_prisma_migrations` table
+3. Try applying the migration again by restarting the server app
+
+:::tip Viewing the `_prisma_migrations` table
+
+You can't use the `wasp db studio` command to view the `_prisma_migrations` table in the production database, but you can use a database management tool like [DBeaver](https://dbeaver.io/) or [pgAdmin](https://www.pgadmin.org/).
+:::
+
+## Connect to the production database
+
+**In development**, you can use the `wasp db studio` command to open a web-based database management tool that allows you to inspect the database. 
+
+You can use the same tool to inspect the **production database**, but you'll need to set the `DATABASE_URL` environment variable to point to the production database. Set the variable in the `.env.server` file in the root of your project.
+
+```bash title=".env.server"
+DATABASE_URL="postgresql://user:password@host:port/dbname"
+```
+
+Then you can start the database studio with the following command:
+
+```bash
+wasp db studio
+```
+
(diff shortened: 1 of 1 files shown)


Martinsos:
I am linking my issue from another PR that you said we can discuss here: https://github.com/wasp-lang/wasp/pull/2373#discussion_r1863053056 .

web/docs/deployment/database.md
```diff
@@ -2,10 +2,66 @@
 title: Database
 ---
 
-TODO: we'll explain the database in the context of deployment
+In this section, we'll discuss what happens with the database when your app goes live. When you develop your app locally, you probably use the dev database that you start with the `wasp db start` command. However, when it's time to deploy your app, you'll need to set up a production database.
```
Martinsos, line 5:
```suggestion
In this section, we'll discuss what happens with the database when your app goes live. When you develop your app locally, you probably use the local dev database that you start with the `wasp db start` command or run on your own. However, when it's time to deploy your app, you'll need to set up a production database.
```

web/docs/deployment/database.md
```diff
@@ -2,10 +2,66 @@
...
 
-TODO: the only requirement from Wasp's point of view - it has to be a PostgreSQL database accessible from the server via the `DATABASE_URL` environment variable
+### Production database requirements
 
-TODO: we'll talk about migrations: how are they created in development, how are they applied in production, how to debug failed migrations
+The server app that Wasp generates uses a PostgreSQL database. The only requirement from Wasp's point of view is that the database is accessible from the server via the `DATABASE_URL` environment variable.
```
Martinsos, line 9:
```suggestion
The server app that Wasp generates uses a PostgreSQL database. The only requirement from Wasp's point of view is that the database is accessible from the server via the `DATABASE_URL` server environment variable.
```

web/docs/deployment/database.md
```diff
@@ -2,10 +2,66 @@
...
+wasp db migrate-dev
+```
+
+This command will create a new migration in the `migrations` directory. The migration is a set of SQL commands that describe the change you made in the schema.
+
+### Applying migrations
+
+**In development**, the migrations are applied as soon as you run the `wasp start` command. 
+
+**In production**, the server app first checks if there are any new migrations that need to be applied, and if there are, it applies them before starting the server. This way, the database schema is always in sync with the Prisma schema.
```
Martinsos, line 33: Actually it is not a server app doing this, it is our Dockerfile, right? I would make it clera this behaviour is present only if they use our Dockerfile.

web/docs/deployment/database.md
```diff
@@ -2,10 +2,66 @@
...
 
-TODO: we'll talk about how to open the DB studio in development and production
\ No newline at end of file
+It can be a PostgreSQL database running on the same server as the server app, or it can be a managed PostgreSQL database service like [Fly Postgres](https://fly.io/docs/postgres/), [AWS RDS](https://aws.amazon.com/rds/), or some other service.
+
+## Migrations
+
+Every time you make a change in your [Prisma schema](../data-model/prisma-file.md) e.g. adding a new model, changing a field type, etc., you need to create a migration. Migrations are some code that describes the change you made in the schema, and they are used to apply the change to the database.
+
+The benefit of migrations is that you can apply the same change to multiple databases, e.g. your local development database and your production database.
```
Martinsos, line 17:
```suggestion
The benefit of migrations is that you can apply the same change to multiple databases. Typically, this is used to apply the migration in development that your teammate produced, or to apply the migration in production after you have developed and tested it in development. 
```

web/docs/deployment/database.md
```diff
@@ -2,10 +2,66 @@
...
+
+**In development**, the migrations are applied as soon as you run the `wasp start` command. 
+
+**In production**, the server app first checks if there are any new migrations that need to be applied, and if there are, it applies them before starting the server. This way, the database schema is always in sync with the Prisma schema.
+
+The migrations might fail to apply if there is a conflict with the existing data in the database. In that case, you'll need to fix the migration and try again.
+
+### Debugging failed migrations
+
+If a migration fails to apply, the server app will log the error message and stop. You should then connect to the production database and see what went wrong. If you check the `_prisma_migrations` table, you'll see the failed migration there.
```
Martinsos, line 39: server app -> is that the term we use? Might be ok, just not sure I hread it before. I was a bit tempted to add Wasp here, "Wasp server" or "Wasp server app", to make it clearer, but I am not 100% sure if that is an improvement or not.

web/docs/deployment/database.md
```diff
@@ -2,10 +2,66 @@
...
+
+**In production**, the server app first checks if there are any new migrations that need to be applied, and if there are, it applies them before starting the server. This way, the database schema is always in sync with the Prisma schema.
+
+The migrations might fail to apply if there is a conflict with the existing data in the database. In that case, you'll need to fix the migration and try again.
+
+### Debugging failed migrations
+
+If a migration fails to apply, the server app will log the error message and stop. You should then connect to the production database and see what went wrong. If you check the `_prisma_migrations` table, you'll see the failed migration there. 
+
+You can try resolving the erorr e.g. if you tried adding a `@unique` constraint to a field that already has duplicate values:
```
Martinsos, line 41: Can we link to Prisma docs that have more details on this process? Handling failed migrations? And migrations in general? I don't mind that we explain it here a bit ourselves, but if there are materials about it in Prisma docs, I would like to them for more info.

web/docs/deployment/database.md
```diff
@@ -2,10 +2,66 @@
...
+DATABASE_URL="postgresql://user:password@host:port/dbname"
+```
+
+Then you can start the database studio with the following command:
+
+```bash
+wasp db studio
+```
+
+If you are looking how to connect to a Fly.io production database, we wrote a [guide](https://github.com/wasp-lang/learning-materials/?tab=readme-ov-file#running-wasp-db-studio-on-production-db) on how to do that.
```
Martinsos, line 67: This is quite dangerous though, they can easily forget to unset that value and then in development to stuff on production database. So if they are doing this, I would super strongly emphasize to unset it as soon as they are done. That said, I would prefer suggesting different approach. For example setting it in shell? That way it stays only for the current session. They could even use it only per command, as prefix `DATABASE_URL=<production_url> wasp db studio`, I think that would work? Maybe they can create .sh file for it also (although they have to be careful not to commit it). Maybe safest is just to tell them to set env var in session.

Martinsos (commented):
@infomiho looking good in general, we can easily more when needed now that we have this page in docs, I think it is good enough for now! I added some comments, some less some more crucial.

web/docs/deployment/database.md
infomiho: Well actually, it's the `start-production` npm script. It runs `npm run db-migrate-prod` before starting the server. I'll mention this.

web/docs/deployment/database.md
infomiho: I keep using it, server app and the client app. We are in Wasp context, which other server app can exist :)

web/docs/deployment/database.md
infomiho: Looking at the Prisma docs, they rely heavily on `prisma` commands, I'm afraid if I link to those, people will start running the `prisma` commands and get stuck. https://www.prisma.io/docs/orm/prisma-migrate/workflows/patching-and-hotfixing#failed-migration

web/docs/deployment/database.md
infomiho: Good stuff!

web/docs/deployment/database.md
Martinsos: Hm crap. I would still link to them I think, maybe warn them. But ok, whatever you think is best here is ok with me!

Martinsos (approved):

PR #2382, opened 2024-11-25
Update Posthog script

infomiho:
Updates the `web` Posthog script to latest version.


web/static/scripts/posthog.js
```diff
@@ -1,2 +1,8 @@
...
+posthog.init(
+  'CdDd2A0jKTI2vFAsrI9JWm3MqpOcgHz1bMyogAcwsE4',
+  {
+    api_host:'https://us.i.posthog.com',
+    person_profiles: 'identified_only'
```
infomiho, line 6: @mmaksimovic This option comes by default and it means "Don't create person profiles for anon users" which I think makes sense for us since we capture only anon users.

Issue #2024, opened 2024-05-08
Run headless tests on `waspc/examples/todoApp`


sodic:
The app we run headless tests on (`headless-test/examples/todoApp`) is a copy of `waspc/examples/todoApp` that diverged over time.

Having multiple copies of "apps for testing" is a maintenance burden. We should combine these two apps into a single app under `waspc/examples/todoApp` and run the tests on that one. 

Related #1976 
Related #2011 
See this when implementing: https://github.com/wasp-lang/wasp/pull/2044#discussion_r1636908391

PR #2389, opened 2024-11-28
CI/CD deployment docs

infomiho:
Talks about how users can test their apps in the CI and about packaging their apps in Docker images for CD.

### Left to do:

- [x] Talk about building the client in the CI https://github.com/wasp-lang/wasp/pull/2416#discussion_r1892938112


web/docs/deployment/ci-cd.md
```diff
@@ -1,7 +1,113 @@
...
+Setting up a CI/CD pipeline is an optional but highly recommended part of deploying applications.
 
-TODO: we'll talk about how to e2e test Wasp apps in the CI
+**Continuous Integration (CI)** involves verifying/testing code changes through an automated process whenever code is pushed to the repository. This helps us catch bugs early and make sure that our app works.
 
-TODO: we'll talk about building the server as a Docker image in the CI
\ No newline at end of file
+**Continuous Deployment (CD)** refers to the automatic deployment of code changes to the production environment. This is commonly know as "push to deploy" and frees developers from having to manually deploy code changes.
+
+## Running tests in CI
```
Martinsos, line 10: I wonder if this should be under deployment, since CI is a separate thing from deployment.

Maybe it would be better if we had a part of docs called Testing and then tehre we can have CI part?

web/docs/deployment/ci-cd.md
```diff
@@ -1,7 +1,113 @@
 ---
 title: CI/CD Scenarios
 ---
+Setting up a CI/CD pipeline is an optional but highly recommended part of deploying applications.
 
-TODO: we'll talk about how to e2e test Wasp apps in the CI
+**Continuous Integration (CI)** involves verifying/testing code changes through an automated process whenever code is pushed to the repository. This helps us catch bugs early and make sure that our app works.
```
Martinsos, line 6:
```suggestion
**Continuous Integration (CI)** involves verifying/testing code changes during development through an automated process whenever code is pushed to the repository. This helps us catch bugs early and make sure that our app works.
```

web/docs/deployment/ci-cd.md
```diff
@@ -1,7 +1,113 @@
...
+**To run e2e tests with Wasp in the CI**, you'll need to:
+1. Install Wasp in the CI environment.
+2. Run your app (with the database) in the CI environment.
+3. Run the e2e tests against the running app.
+
+#### Example app
+
+We'll show you how to run end-to-end tests in CI using the [Github Actions](https://github.com/features/actions) as our CI and the [Playwright](https://playwright.dev/) as our e2e testing framework.
+
+1. Check our example app and its e2e tests in the [e2e-tests](https://github.com/wasp-lang/e2e-test-example/tree/main/e2e-tests) directory.
```
Martinsos, line 26: Hmmm should this point to release branch? Usually in docs we aim to point at release branch, since we know it is in sycn with the released version of the docs.

web/docs/deployment/ci-cd.md
```diff
@@ -1,7 +1,113 @@
...
+        await expect(page.locator("body")).toContainText("No tasks yet.");
+
+        // Add a task
+        await page.fill('input[name="description"]', "First task");
+        await page.click('input:has-text("Create task")');
+        await expect(page.locator("body")).toContainText("First task");
+      });
+    });
+    ```
+    </details>
```
Martinsos, line 52: You can probably drop this? It is just a small snippet, and it risks becoming out of date, we have to maintain it, real code is anyway in that example app you linked to.

web/docs/deployment/ci-cd.md
```diff
@@ -1,7 +1,113 @@
...
+
+The most common way to package your app for deployment is using Docker images. This way you can easily deploy the same image to different environments (staging, production, etc.).
+
+**To build the app as a Docker image**, you'll need to:
+1. Install Docker in the CI environment.
+2. Build the app with `wasp build`.
+3. Build the Docker image and push it to a Docker registry:
+    - for our server app
+    - for our client app
+4. Notify your deployment provider to deploy the new image.
```
Martinsos, line 78: This can depend on their hostirng provider, right? If I am correct some of them have their own registries so this looks more like just a single command, or maybe you just give them a branch with Dockerfile, stuff like that?
Also client app doesn't have to be in Dockerfile.
Also might be worth  mentiong here that we provide Dockerfile for the server.

And now that I am writnig all this, sounds to me like this is quite complex and we should instead point them to the part of deployment docs that describes it in details? Maybe just be very simple here, try not to assume much, link to the part with dpeloying Wasp app, and say that they will want to use whatever method their hosting providers reocmmends for CD. And an example deployment of ours, that is cool.

Martinsos (commented):
Looking good! This one feels a bi tless claer to me, regarding what should be here. Seems every easy to me to go into too much details, or to not give enough info. Since it is more CI/hosting provider specific than Wasp specific. So I think it is ok to keep this as light as we can.

web/docs/deployment/ci-cd.md
infomiho: Can we talk about that when we have all the deployment docs merged? I don't want to spawn a new `Testing docs` effort. Let's keep this here for now and then we can rework the Testing docs in the future and move this bit there.

web/docs/deployment/ci-cd.md
infomiho: I doesn't only happen during development. It can be also running before the deployment pipeline.

web/docs/deployment/ci-cd.md
infomiho: I wanted to include this code example here to motivate people to write these tests. The example app is nice, but this is available while going through the docs and maybe better understand what e2e tests mean without leaving the docs.

I get your concern about the code being outdated - but the example app code will get outdated over time as well :)

web/docs/deployment/ci-cd.md
infomiho: I've rewritten this line as `For some providers: notify them to deploy the new app version.`

> Also client app doesn't have to be in Dockerfile.

I've mentioned an alternative way to deploy the client 👍 

> sounds to me like this is quite complex 

Yep, it's quite complex, that's why I wanted to explain the concept from a higher level a bit. We our self-hosted deployment docs that talk a lot about using CD.

web/docs/deployment/ci-cd.md
Martinsos: Ok, I will leave that to you to figure out! I thought maybe creating this new page in docs is clear cut, but if that is too much for now, I get it.

web/docs/deployment/ci-cd.md
Martinsos: All right I missed that! Hos iw that, don't we have example app with e2e tests in our `wasp` repo? How is it that we also have this external one? Should we be linking to it?

web/docs/deployment/ci-cd.md
Martinsos: My feeling is that maintanability trumps the other arguments here. We will keep the example app up to date though! At least we should, I would expect that.

Martinsos (approved):
Approved, but check a couple of new comments.

web/docs/deployment/ci-cd.md
infomiho: We have e2e tests running in our `wasp` repo but it's quite complex IMHO so I wanted to offer a simple example repo with the e2e tests so people can just grab them and use them. Maybe in the future when we work on https://github.com/wasp-lang/wasp/issues/2024 we can link to the consolidated app.

web/docs/deployment/ci-cd.md
infomiho: Created an issue: https://github.com/wasp-lang/wasp/issues/2439

PR #2393, opened 2024-11-29
TS Config cleanup - Part 1

sodic:


waspc/packages/wasp-config/src/run.ts
```diff
@@ -77,14 +96,14 @@ function parseProcessArguments(args: string[]): {
...
   }
   return entities
 }
 
-function serialize(appConfig: Decl[]): string {
-  return JSON.stringify(appConfig)
-}
+type Result<Value, Error> =
+  | { status: 'ok'; value: Value }
+  | { status: 'error'; error: Error }
```
sodic, line 109: I'm using this for normal control flow errors (e.g., user forgets something).

I'm using exceptions for unexpected errors/compiler bugs. For example, an exception is thrown if `waspc` doesn't pass the correct arguments to this executable.

For the normal control flow, I am using the `Result` type instead of exceptions because:
- The error users see is nicer (no stack trace)
- They can slip through the type system.

waspc/packages/wasp-config/src/mapUserSpecToAppSpecDecls.ts
```diff
@@ -64,60 +71,39 @@ export function mapUserSpecToDecls(
...
+  type: DeclType,
+  configToDeclValue: (
+    config: T
+  ) => AppSpec.GetDeclForType<DeclType>['declValue']
+) {
+  return [...configs].map(([name, config]) => ({
+    declType: type,
+    declName: name,
+    declValue: configToDeclValue(config),
+  }))
```
sodic, line 106: This might look too complex compared to what we had before (the `for` loops), but comes with many benefits:

#### No repetition
We don't repeat the same map iteration and `Decl` object construction a bunch of times.

#### The entire `decls` array is built at once, without pushing
Because we're functional bros.


#### The compiler ensures we don't forget to include any declarations
This is the biggest benefit IMO, courtesy of the `makeDeclsArray` function.
For example, see what happens if we forget to include Crud:

image

So, when we add a new declaration, the compiler tells us what to implement.

#### The compiler ensures we don't mix anything up
`mapToDecls` ensures that the `Decl` returned by the mapping function matches the `DeclType` provided as the second argument, so we can't mix different `Decl` types and values (we had this before, but it's preserved):

image

waspc/packages/wasp-config/src/mapUserSpecToAppSpecDecls.ts
```diff
@@ -64,60 +71,39 @@ export function mapUserSpecToDecls(
...
+}
 
-  return decls
+function mapToDecls<T, const DeclType extends AppSpec.Decl['declType']>(
+  configs: Map<string, T>,
+  type: DeclType,
+  configToDeclValue: (
+    config: T
+  ) => AppSpec.GetDeclForType<DeclType>['declValue']
+) {
```
sodic, line 101: I can't specify a return type of this function because of some weird error, but I stopped digging into it.

If you're curious, here's the [playground](https://www.typescriptlang.org/play/?#code/FAUwHgDg9gTgLgAjgTwiBBxEcCiA7OGZAMVgGkBLPAEwB5KaFw4QaBnBfQ5AbQHIA1lWp8AugD4EAXk5hCAQwDGcWsAScCRADRqEAbwRCaALgQNqCAL7BxwUJFiIUaDd2m6APvsPDTfNoRUAOZ8ANwIAG7yADYAriCmATDBVp7eRtR+eLEAtgBGIDBhkTHxptn5hal2AGaxeMoUUHgI1FAAylA52AAWwfTCTHKs1BxcRPwZYuIAFBmm5lolcQmY2OMk5MIDNOL8UStiAJSmWLiamzDmO9SSerow2LEwLffq6hk678vxutaWQA).

waspc/packages/wasp-config/src/mapUserSpecToAppSpecDecls.ts
```diff
@@ -64,60 +71,39 @@ export function mapUserSpecToDecls(
...
+}
 
-  for (const [crudName, crudConfig] of spec.cruds.entries()) {
-    decls.push({
-      declType: 'Crud',
-      declName: crudName,
-      declValue: mapCrud(crudConfig, parseEntityRef),
-    })
-  }
+function makeDeclsArray(decls: {
```
infomiho, line 89: Sweet

waspc/packages/wasp-config/src/mapUserSpecToAppSpecDecls.ts
```diff
@@ -64,60 +71,39 @@ export function mapUserSpecToDecls(
...
+  [Type in AppSpec.Decl['declType']]: AppSpec.GetDeclForType<Type>[]
+}): AppSpec.Decl[] {
+  return Object.values(decls).flatMap((decl) => [...decl])
+}
 
-  return decls
+function mapToDecls<T, const DeclType extends AppSpec.Decl['declType']>(
```
infomiho, line 95: What does `const` do here?

waspc/packages/wasp-config/src/mapUserSpecToAppSpecDecls.ts
```diff
@@ -228,9 +212,10 @@ function mapAuth(
   return {
     userEntity: parseEntityRef(userEntity),
     // TODO: Abstract away this pattern
-    ...(externalAuthEntity && {
-      externalAuthEntity: parseEntityRef(externalAuthEntity),
-    }),
+    externalAuthEntity:
```
infomiho, line 215: For entities you did `entities: entities && entities.map(parseEntityRef),` why are you are explicitly checking for `undefined`?

waspc/packages/wasp-config/src/userApi.ts
```diff
@@ -7,13 +7,15 @@ export class App {
   #userSpec: UserSpec;
 
   // NOTE: Using a non-public symbol gives us a pacakge-private property.
```
infomiho, line 9:
```suggestion
  // NOTE: Using a non-public symbol gives us a package-private property.
```

waspc/src/Wasp/AppSpec.hs
```diff
@@ -58,6 +58,10 @@ import qualified Wasp.SemanticVersion as SV
 -- describing the web app specification with all the details needed to generate it.
 -- It is standalone and de-coupled from other parts of the compiler and knows nothing about them,
 -- instead other parts are using it: Analyzer produces AppSpec while Generator consumes it.
+--
+-- IMPORTANT: Do not change this data structure without updating the AppSpec in
```
infomiho, line 62: I had this random 💩 idea, run a LLM on each commit if this files changed or the TS file changed. "Are these two things  the same implementation?"

waspc/tools/install_packages_to_data_dir.sh
```diff
@@ -9,6 +9,9 @@ dir=$(CDPATH= cd -- "$(dirname -- "$0")" && pwd)
 for package in $(ls "$dir/../packages"); do
   package_dir="$dir/../packages/$package"
   if [[ -d "$package_dir" ]]; then
+    # We're only installing the dependencines here to verify that the build
+    # works, that's why the node_modules folder is removed immediately after.
+    # The real dependency installatino happens in Haskell.
```
infomiho, line 14:
```suggestion
    # The real dependency installation happens in Haskell.
```

infomiho (approved):
LGTM

Some nice code improvements. Left some minor comments.

waspc/packages/wasp-config/src/mapUserSpecToAppSpecDecls.ts
sodic: Nothing, it's a leftover I think. Nice catch.

waspc/packages/wasp-config/src/run.ts
Martinsos: Nice! If only there was a language that already used this kind of approach for errors/exceptions :D!

waspc/packages/wasp-config/src/mapUserSpecToAppSpecDecls.ts
sodic: Because it can technically be an empty string.

PR #2395, opened 2024-12-03
Change the Arrivederci to Goodbye, and fix CSS wrapping issues with titles

NathanaelA:
### Description

Change the word: "Arrivederci" to "Goodbye" to make it more universal

Fixed wrapping issue with titles in same section:
Before:
image

After:
image

### DIFF
--- a/web/src/css/custom.css
+++ b/web/src/css/custom.css
@@ -152,3 +152,10 @@
   height: 100%;
   border: 0;
 }
+
+/***** HowItWorks *****/
+.featureTitle {
+  line-height: 1.73;
+  box-decoration-break: clone;
+  -webkit-box-decoration-break: clone;
+}
--- a/web/src/css/custom.css
+++ b/web/src/css/custom.css
@@ -152,10 +152,3 @@
   height: 100%;
   border: 0;
 }
-
-/***** HowItWorks *****/
-.featureTitle {
-  line-height: 1.73;
-  box-decoration-break: clone;
-  -webkit-box-decoration-break: clone;
-}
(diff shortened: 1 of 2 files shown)


web/src/css/custom.css
```diff
@@ -152,3 +152,10 @@
   height: 100%;
   border: 0;
 }
+
+/***** HowItWorks *****/
+.featureTitle {
+  line-height: 1.73;
+  box-decoration-break: clone;
+  -webkit-box-decoration-break: clone;
+}
```
Martinsos, line 161: Since we use tailwind, could we move some of this stuff to tailwind? For example it has [leading](https://tailwindcss.com/docs/line-height) class for line height. `leading-relaxed` is 1.625, maybe that is a good value? How did you pick exactly 1.73?

There is also [box-decoration-clone](https://tailwindcss.com/docs/box-decoration-break)!

So yeah, let's use those two and let's put them directly into the JSX, next to other classes.

Martinsos (changes requested):
Nice fix @NathanaelA , thank you! The only thing we should do is make sure we use Tailwind, since that is what we use here: I linked more details on how to do that.

web/src/css/custom.css
NathanaelA: 1.625 would cause them to merge, I test by testing multiple values -- 1.73 was the point where the two boxes do NOT overlap. 

I can add "box-decoration-clone" and remove it from this css, but I think you should leave the line-height otherwise it will still look poor using 1.625...

Martinsos: There is also a way to set artibtrary values in Tailwind, with leading-[1.73], we could do that?

NathanaelA: @Martinsos - I've updated the PR to use only Tailwind properties and removed the changes from the custom.css.

Martinsos (approved):
Awesome, thanks @NathanaelA !

PR #2402, opened 2024-12-08
Support in Wasp for external/community starter templates

Martinsos:
I started this PR with the idea to add support for external starter templates to Wasp CLI. I imagined that we would allow users to use them in the similar fashion as right now we do for our official templates, it is just that they would specify them with sometihng like `github:<repo_owner>/<repo_name>[/template/dir]` and we would pull the template from there.

I implemented ability to this via `wasp new -t <repo_owner>/<repo_name>[/template/dir]`, and also ended up refactoring a bit bigger parts of the starter templates logic in CLI while at it, both to support it and to improve the code in general.

However, I then at some point realized that I am not sure if this is useful at all! When you think about it, what do they get this way compared to just going to a github repo of that starter template and cloning that or using github's "clone as a template" feature? For official templates, we want to make sure they can be discovered and easily used from the CLI, and that makes sense, and we also insert name of the app into them and correct Wasp version. But for community tempaltes, that isn't that important I think -> We would expected users to go out of Wasp CLI and check out that template anyway, read the instrutions there, ... . They probably found out about it in some other fashion also. And I think maintainers of these community templates are not likely to be interested in using the rules that come with how our CLI pulls in the tempaltes -> I don't think they will use the template placeholders for wasp version and app name, nor can I see them updating templates immediately on every wasp version (which wasp CLI expects from official tempaltes, since it pulls them by specific version tag). I expect community templates will instead on their `main` be at the latest wasp version that they managed to update to and that is it, and they will be late a bit after each Wasp release, that is good enough and simple enough, I would actually recommend doing it that way.

So, I concluded that having this ability to pull in the repo from our CLI via `github:...` is actually not useful! What do you think, do you agree?

However, while implementing this PR, it turned out that most work was other stuff, which is improvments to the rest of the code / docs. Everything taht I implmented in this PR:
1. Added ability to do `wasp new -t github:...`.
2. Refactored and improved the code of StarterTemplates modules in wasp CLI Haskell code.
3. Updated CLI instructions on tempaltes to be correct and up to date.
4. Updated docs on starter templates to have a "Community Templates" section that lists Roke, the template by @wardbox .
5. Added an additional interactive option to the starter template menu in `wasp new` that is called "community template" that informs users about existence of community templates, to help with discovery.

As it turns out, (1) is actually only a small part of it.

So, what I did at the end is create this PR that contains all the changes from above, although I would suggest that I remove (1) (which is mostly changes in TemplateId.hs), but I would merge the rest.

What I would like you to do:
1. Review the PR.
2. Let me know if you would keep the `github:` functionality or not. I vote for not, and will kick it out if you agree.
3. Let me know how you find the addition of community templates link in interactive menu, is that all right. Once picked, it throws an error, but I felt like making that not be an error is not worth it at the moment, and that it is clear enough.

Additional option in interactive menu:
image

Addition to docs:
image


infomiho:
Great analysis @Martinsos! 

1. I think you are right that if we concluded that a feature is not really useful (`github:` templates), **we should remove it**. We don't want to maintain features we don't use or even get stuck with a iteration of a feature that might be okay now, but later on with new info - we find it we want to reimplement it.  
2. I dislike the **community template** option in the menu. Somebody will want to try out community templates and then they get an error. "Oh well, this guys have some sort of a bug". I'd rather invest some time into outputting some extra text to say "Hey, these are our official templates - check out the community templates as well" in some sort of "Footer" of the listed options. 

I'll review the code next :)

waspc/ChangeLog.md
```diff
@@ -1,11 +1,15 @@
 # Changelog
 
-## Next
+## [WIP] 0.16.0
```
infomiho, line 3: What was wrong with Next? 😢 I kinda liked it, you don't have to think about what the next version will be

waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs
```diff
@@ -46,20 +46,21 @@ createProjectOnDisk
...
-        liftIO $ createProjectOnDiskFromLocalTemplate absWaspProjectDir projectName appName $ _path metadata
-      AiGeneratedStarterTemplate ->
-        AI.createNewProjectInteractiveOnDisk absWaspProjectDir appName
+      GhRepoStarterTemplate ghRepoRef tmplDirPath _ ->
+        createProjectOnDiskFromGhRepoTemplate absWaspProjectDir projectName appName ghRepoRef tmplDirPath
+      LocalStarterTemplate tmplDirPath _ ->
+        liftIO $ createProjectOnDiskFromLocalTemplate absWaspProjectDir projectName appName tmplDirPath
+      AiGeneratedStarterTemplate waspAppAiGenerator _ ->
+        case waspAppAiGenerator of
+          WaspAI -> AI.createNewProjectInteractiveOnDisk absWaspProjectDir appName
```
infomiho, line 57: Why did we generalise this, do we expect different generators in the future?

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/ArgumentsParser.hs
```diff
@@ -33,7 +33,7 @@ parseNewProjectArgs newArgs = parserResultToEither $ execParserPure defaultPrefs
       Opt.strOption $
         Opt.long "template"
           <> Opt.short 't'
-          <> Opt.metavar "TEMPLATE_NAME"
+          <> Opt.metavar "TEMPLATE"
```
infomiho, line 36: Do we maybe want to go with `TEMPLATE_ID` to keep it consistent with the record above?

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/Skeleton.hs
```diff
@@ -0,0 +1,18 @@
...
+
+import Data.Text (Text)
+import StrongPath (File', Path, Rel, System, reldir, (</>))
+import qualified Wasp.Data as Data
+import Wasp.Project.Common (WaspProjectDir)
+import Wasp.Util.IO (listDirectoryDeep, readFileStrict)
+
+readWaspProjectSkeletonFiles :: IO [(Path System (Rel WaspProjectDir) File', Text)]
+readWaspProjectSkeletonFiles = do
+  skeletonFilesDir <- (</> [reldir|Cli/templates/skeleton|]) <$> Data.getAbsDataDirPath
```
infomiho, line 16: I had trouble parsing this line at first, can write in way that it reads like a path i.e. from left to right?

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/StarterTemplate.hs
```diff
@@ -0,0 +1,61 @@
...
+type TemplateDirPath = Path' Rel' Dir'
+
+data TemplateMetadata = TemplateMetadata
+  { _tmplName :: !String,
+    _tmplDescription :: !String,
+    _tmplBuildStartingInstructions :: !StartingInstructionsBuilder
+  }
+
+-- | AI generators that can generate Wasp apps.
+data WaspAppAiGenerator = WaspAI
```
infomiho, line 33: Are we overengineerring here a bit? Are we working on other AI generators for Wasp or looking to support other AI external AI generators? Sorry if I missed that we are doing something like that.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/StarterTemplateId.hs
```diff
@@ -0,0 +1,115 @@
...
+  GhRepoTemplateUri {} -> IdTypeGhRepoTemplateUri
+
+getStarterTemplateIdTypeDescription :: StarterTemplateIdType -> String
+getStarterTemplateIdTypeDescription = \case
+  IdTypeFeaturedTemplateName -> "a featured template name (e.g. \"saas\")"
+  IdTypeGhRepoTemplateUri -> "a github URI (github:<repo_owner>/<repo_name>[/dir/to/template])"
+
+-- | Given a template id (as string), it will obtain the information on the template that this id references.
+-- It will throw if the id is invalid (can't be parsed, or information on the template can't be obtain based on it).
+getStarterTemplateByIdOrThrow :: (MonadError String m) => [ST.StarterTemplate] -> String -> m ST.StarterTemplate
```
infomiho, line 54: Cool way of resolving the Github based templates 👍

infomiho (approved):
LGTM

Left some minor comments. I like the `github:` implementation - but - as I said, let's remove features we don't believe would be used by our users.

waspc/ChangeLog.md
Martinsos: Next is cool! But in this case I knew what next version is going to be, so I thought why not write it down? But yeah I could have left Next you are right, wouldn't be an issue.

Martinsos:
So when I wrote this PR I also wrote 9 comments for the reviewers to make reviewing easier, but they stayed on "pending"! Too bad, I pushed them now.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/StarterTemplate.hs
Martinsos: Yes, that is my secret project :D! Kidding, it is a tiny bit of overengienerg that allowed me to treat AI starter template same as the rest while keeping compile time safety. Commented on it also in another comment in more details.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/StarterTemplateId.hs
Martinsos: Thanks! Will likely be throwing away most of this though.

vincanger:
Yeah @Martinsos I agree it's not really useful. What could be cool though is to have a section on our docs or webpage that features community templates to give those creators some recognition and perhaps inspire others to make their own.

web/docs/project/starter-templates.md
```diff
@@ -82,8 +84,23 @@ wasp new <project-name> -t todo-ts
...
+project description. It will automatically generate your data model, auth, queries, actions and React pages.
 
 _You will need to provide your own OpenAI API key to be able to use this template._
 
-**Features:** Generated using OpenAI's GPT models, Auth (username/password), Queries, Actions, Pages, Full-stack Type Safety
\ No newline at end of file
+**Features:** Generated using OpenAI's GPT models, Auth (username/password), Queries, Actions, Pages, Full-stack Type Safety
+
+
+## Community templates #{community-templates}
```
Martinsos, line 94: This id in #{} got rendered in docs! See what is with that.

infomiho: What do you mean? How does it look like?

Martinsos:
Too ahead of its time (hehe), closing.

Issue #2398, opened 2024-12-04
When deploying on fly, server PORT is set to 8080 instead of 3000?


Martinsos:
We had two users encounter this situation. It is not clear how that happened -> they used `wasp deploy fly deploy` and that should have set PORT to 3000, not 8080. We should investigate to figure out what could be causing this.

One noticable thing is that on the first deploy one of the users forgot some other env vars, so they redeployed a couple of times till they got them all set correctly. Might it be that in this process, PORT somehow got messed up? Or it was never set to 3000?

Related discord convo: https://discord.com/channels/686873244791210014/1313215934844113000/1313215934844113000 .

vincanger:
This just happened to me as well:

1. Ran `wasp deploy fly launch cursor-template ams --org wasp`
2. Encountered the following error:
```sh
$ flyctl postgres create --name cursor-template-db --region ams --vm-size shared-cpu-1x --initial-cluster-size 1 --volume-size 1 --org wasp
/bin/bash: flyctl: command not found
🚀  There was an error running "wasp deploy fly create-db <region>". Please review the error and try again (if appropriate). 
```
3. Ran `wasp deploy fly create-db ams` and encountered error that I forgot to add the org slug so 
4. Ran `wasp deploy fly create-db ams --org wasp` && `wasp deploy fly deploy --org wasp`
5. Checked port and it was set to `8080` so I changed it to `3000` and it worked and is now live at https://cursor-template-client.fly.dev/

Martinsos:
Hm ok we certainly have to look into this then! Btw what do you think caused that first error, any ideas? Ok so you ran "launch" -> and that error popped out at which moment? Woudl would it say that command is not found hm. What is your version of flyctl?

vincanger:
Fly at the time was having an issue so perhaps the db creation and app deployment overall failed because of that? Hard to say. 

Not at my laptop now but I believe I’m on the latest flyctl version.

jturner:
I also just ran my first `wasp deploy fly launch ptcgp-friends ord` for my app and had the same issue where the server was running on 8080 but fly.io wanted it on 3000. I ran `wasp deploy fly cmd secrets set PORT=3000 --context=server` to fix this.

Otherwise I encountered no other errors when running the launch command.

PR #2403, opened 2024-12-09
Force internal port to 8080

jturner:
### Description

This is a proposed fix for https://github.com/wasp-lang/wasp/issues/2398. This is similar to a fix that was made to the client setup. Another option would be to change like 120 to `PORT=3000` but my guess is older version of `flyctl` used port 8080 and don't want to break those.

### DIFF
--- a/waspc/packages/deploy/src/providers/fly/setup/setup.ts
+++ b/waspc/packages/deploy/src/providers/fly/setup/setup.ts
@@ -85,6 +85,7 @@ async function setupServer(deploymentInfo: DeploymentInfo<SetupOptions>) {
 	await $`flyctl launch --no-deploy ${launchArgs}`;
 
 	const minMachinesOptionRegex = /min_machines_running = 0/g;
+	const internalPortOptionRegex = /internal_port = \d+/g;
 
 	if (!doesLocalTomlContainLine(minMachinesOptionRegex)) {
 		await question(`\n⚠️  There was a possible issue setting up your server app.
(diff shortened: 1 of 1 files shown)


waspc/packages/deploy/src/providers/fly/setup/setup.ts
```diff
@@ -109,6 +110,24 @@ Press any key to continue or Ctrl+C to cancel.`);
...
 
+	if (!doesLocalTomlContainLine(internalPortOptionRegex)) {
+		await question(`\n⚠️  There was an issue setting up your server app.
+We tried modifying your server fly.toml to set ${boldText(
+		'internal_port = 8080',
+	)}, but couldn't find the option ${boldText(
+	'internal_port',
+)} in the fly.toml.
+
+This means your server app might not be accessible.
```
infomiho, line 121: I got confused by this part, what are we trying to say here: if the `internal_port` option is not present in the Fly TOML file the server app might not work correctly because Fly won't know how to reach it?

waspc/packages/deploy/src/providers/fly/setup/setup.ts
```diff
@@ -109,6 +110,24 @@ Press any key to continue or Ctrl+C to cancel.`);
...
+
+Contact the Wasp Team at our Discord server if you need help with this: https://discord.gg/rzdnErX
+
+Press any key to continue or Ctrl+C to cancel.`);
+	} else {
+		// the default fly.toml assumes port 8080 (or 3000, depending on flyctl version).
```
infomiho, line 127: I feel like this comment is redundant - we are setting the value to something we want, the default is not really important to know for the reader.

infomiho: In this file, here: https://github.com/wasp-lang/wasp/blob/a0f108b8f07b00b9b55e094e114148e1a3a78588/waspc/packages/deploy/src/providers/fly/setup/setup.ts#L139 we are also setting the server env var to `8080` and we want these two to change at the same time. So, I'd extract the port information into a variable and use it in both places e.g. `serverPort` or similar.

waspc/packages/deploy/src/providers/fly/setup/setup.ts
```diff
@@ -85,6 +85,7 @@ async function setupServer(deploymentInfo: DeploymentInfo<SetupOptions>) {
 	await $`flyctl launch --no-deploy ${launchArgs}`;
 
 	const minMachinesOptionRegex = /min_machines_running = 0/g;
+	const internalPortOptionRegex = /internal_port = \d+/g;
```
infomiho, line 88: We already defined this regex in the `setupClient` fn, and since it's the same thing, I'd extract this regex to the top and use it both `setupServer` and `setupClient` functions.

infomiho (commented):
Thank you for the contribution! 

Left some comments :)

infomiho:
Thank you @jturner for your effort! Due to the high priority of this fix, I've taken over the PR: #2438

PR #2410, opened 2024-12-10
TS Config cleanup - Part 3

sodic:
Fixing stuff from #2276 and personal notes.

### DIFF
--- a/waspc/src/Wasp/Project/WaspFile.hs
+++ b/waspc/src/Wasp/Project/WaspFile.hs
@@ -21,6 +21,8 @@ import Wasp.Project.Common
     WaspFilePath (..),
     WaspProjectDir,
   )
+import Wasp.Project.WaspFile.TypeScript (analyzeWaspTsFile)
+import Wasp.Project.WaspFile.WaspLang (analyzeWaspLangFile)
 import qualified Wasp.Psl.Ast.Schema as Psl.Schema
 import qualified Wasp.Util.IO as IOUtil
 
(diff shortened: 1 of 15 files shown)


waspc/src/Wasp/Project/WaspFile.hs
```diff
@@ -117,9 +117,23 @@ compileWaspTsFile waspProjectDir tsconfigNodeFileInWaspProjectDir waspFilePath =
...
+          --   suffer bad DX (no IDE support in *.wasp.ts file).
           [ "tsc",
             "-p",
             fromAbsFile (waspProjectDir </> tsconfigNodeFileInWaspProjectDir),
+            -- The tsconfig.wasp.json file has the noEmit flag on.
+            -- The file only exists IDE support, and we don't want users to
+            -- accidentally chage the outDir.
+            --
+            -- Here, to actually generate the JS file in the desired location,
+            -- we must turn off the noEmit flag and specify the outDir.
```
sodic, line 137: Original discussion: https://github.com/wasp-lang/wasp/pull/2276#discussion_r1775758517

waspc/src/Wasp/Psl/Ast/Schema.hs
```diff
@@ -51,6 +49,3 @@ getGenerators schema = [generator | generator@((Psl.ConfigBlock.ConfigBlock Psl.
 
 getConfigBlocks :: Schema -> [ConfigBlock]
 getConfigBlocks (Schema blocks) = [configBlock | ConfigBlock configBlock <- blocks]
-
-getModelNames :: Schema -> [String]
-getModelNames schema = map Model.getName $ getModels schema
```
sodic, line 56: Discussed here: https://github.com/wasp-lang/wasp/pull/2276#discussion_r1776737849

infomiho (approved):
LGTM

waspc/src/Wasp/Project/WaspFile.hs
```diff
@@ -117,9 +118,23 @@ compileWaspTsFile waspProjectDir tsconfigNodeFileInWaspProjectDir waspFilePath =
...
           waspProjectDir
           "npx"
+          -- We're using tsc to compile the *.wasp.ts file into a JS file.
+          --
+          -- The compilation rules mostly come from tsconfig.wasp.json but also
+          -- include several overrides:
+          --   - We don't keep all the rules in tsconfig.wasp.json because it
+          --   would give users a way to break things.
+          --   - We don't keep all the rules here (inline) because users would
+          --   suffer bad DX (no IDE support in *.wasp.ts file).
```
Martinsos, line 128: I found this confusing to read, had to read it a couple of times to understand what it is about.

The problem was that you said "include several overrides:" and then bullet points, but those bullet points were not overrides, as I expected, but instead they were reasons why it makes sense to have these overrides. Which I don't think is what a reader expects, so I was reading and reading trying to undersatnd how is this connected together.

I would instead say that yes, the main source of rules is tsconfig.wasp.json, but we add some additional rules on top of that. And reason for that is that we don't want  to keep all these rules in ... because of ..., and also is not all here because ... .

Although I think better way to put it would be to just say that rules in tsconfig.wasp.json are aimed at user and providing them with the best IDE support, while here we have a bit different requirements, and that is the reason for additional rules we add on top.

Martinsos (approved):
Approved! But left one comment, check it out please.

PR #2413, opened 2024-12-12
Waspello: Tilt cards when dragging

### LINKED ISSUE #338: [example/Waspello] Make list tilt while dragged'n'dropped like in Trello
This is how it works in Trello:
trello_dnd

This is in Waspello:
waspello_list_dnd

**NOTE**: For the implementation of drag'n'drop functionality in Waspello, we are using [react-beautiful-dnd](https://github.com/atlassian/react-beautiful-dnd) maintained by Atlassian (which owns Trello), so I believe with it we should able to implement the same functionality Trello has.

The problems with Waspello are:
* ~~**user can't drag by focus area of list's name** -> it automatically goes into edit mode (turns into input)~~ fixed by #785
* **list does not get tilted while dragged** (wasn't implemented yet but shouldn't be too complicated)

Disregard the slight lag when the list is dropped to the new position, that is due to no support for optimistic UI updates in Wasp yet (currently discussed in #63).

dabrorius:
### Description

This PR implements tilting while dragging for both full lists and individual cards. Fixes #338 

Note:
- Dragging of lists is completely broken in development due to `React.StrictMode`. This is unrelated to this PR and is fixed in https://github.com/wasp-lang/wasp/pull/2414.
- Dragging lists is glitchy, after dragging the lists briefly switch to original position. This is also present on the [deployed app](https://waspello-demo.netlify.app/) and is not related to this PR.

https://github.com/user-attachments/assets/6e912e04-209d-4be5-8147-f8d8626429e1


examples/waspello/src/cards/Main.css
```diff
@@ -235,6 +235,10 @@ button.open-add-list {
   border-radius: 3px;
 }
 
+.list--dragging {
```
infomiho, line 238: I think we should add transitions to make this look smoother:

```css
.list {
  ...
  transition: transform 200ms ease-in-out;
}

.list--dragging {
  transform: rotate(2deg);
}

.list--drop-animating {
  transform: rotate(0deg);
}
```

I'd do the same for the `.list-card` component. 

Also, potential improvement in readability is to maybe declare CSS vars for the angle e.g. `--list-rotate-angle: 2deg` and then use it `var(--list-rotate-angle)`.

examples/waspello/src/cards/MainPage.jsx
```diff
@@ -248,7 +248,7 @@ const List = ({ list, index, cards }) => {
           {...provided.draggableProps}
           {...provided.dragHandleProps}
         >
-          <div className='list'>
+          <div className={`list ${snapshot.isDragging && 'list--dragging'}`}>
```
infomiho, line 251: I'd implement a helper function for this e.g. 

```ts
function getListClassName(snapshot) {
  const classes = ['list']
  if (snapshot.isDragging) {
    classes.push('list--dragging')
  }
  if (snapshot.isDropAnimating) {
    classes.push('list--drop-animating')
  }
  return classes.join(' ')
}
```

This way:
1. There is no space after `list` if `list--dragging` is not present. 
2. It's easier to compose multiple classes

You'll notice I used `isDropAnimating` as well, this enables us to have a smooth start:
```
(no rotation) -> (rotation)
```
and a smooth end:
```
(rotation) -> (no rotation)
```

https://github.com/user-attachments/assets/0e22888d-7fe2-40a0-bf37-2b652dc6e530


https://github.com/user-attachments/assets/a9c4fab4-6264-48e0-ba29-1d57f52d5234

infomiho (commented):
Good job!

I've added some comments to maybe improve it a bit :)

examples/waspello/src/cards/Main.css
dabrorius: I added suggested changes. I didn't extract the `2deg` to a separate variable as it is used only once (when dragging the whole list 4 degs are used), moving its definition to a different place could make the code harder to follow.

dabrorius: P.S. I still need to test the whole list dragging, but it won't work until https://github.com/wasp-lang/wasp/pull/2414 is resolved and merged.

examples/waspello/src/cards/MainPage.jsx
dabrorius: I just noticed that `classnames` npm package is included and used in other places. Should have used that 🤦
I'll update the code.

PR #2417, opened 2024-12-14
Update README docs

komyg:
### Description

These are a few suggested updates to the Waspc README, with problems that I faced when setting up the project for the first time.

### DIFF
--- a/waspc/README.md
+++ b/waspc/README.md
@@ -39,7 +39,7 @@ the project. The best way to install it is via
 is a convenient way of installing and selecting versions of `cabal`, `hls` and
 `ghc`.
 
-:warning: If you are on a Mac, we still recommend you use the official [ghcup](https://www.haskell.org/ghcup/) installer over Homebrew, as the initial setup will be easier, and you won't have to worry about setting the correct environment variables.
+:warning: On Mac, we recommend using the official [ghcup](https://www.haskell.org/ghcup/) installer over Homebrew, as it works out of the box.
 
 In [cabal.project](cabal.project) we're explicitly saying what compiler version
 to build Wasp with.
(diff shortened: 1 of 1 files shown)


waspc/README.md
```diff
@@ -39,16 +39,16 @@ the project. The best way to install it is via
...
 to build Wasp with.
 
 ```cabal
 with-compiler: ghc-x.y.z
 ```
 
-Ensure that you have this [ghc
-version](https://downloads.haskell.org/ghc/latest/docs/html/users_guide/intro.html#ghc-version-numbering-policy)
-installed.
+Please ensure that you have this ghc version installed.
```
komyg, line 51: Note: I removed the link here, because I found it confusing. I expected that it would take me to a page with the correct version of GHC, but it actually took me to their semantic version policy.

waspc/README.md
```diff
@@ -39,16 +39,16 @@ the project. The best way to install it is via
 is a convenient way of installing and selecting versions of `cabal`, `hls` and
 `ghc`.
 
+:warning: If you are on a Mac, we still recommend you use the official [ghcup](https://www.haskell.org/ghcup/) installer over Homebrew, as the initial setup will be easier, and you won't have to worry about setting the correct environment variables.
```
komyg, line 42: I've added this bit, because I tried to install ghcup through Homebrew, and the env vars were all wrong.

Martinsos, line 42:
```suggestion
:warning: On Mac, we recommend using the official [ghcup](https://www.haskell.org/ghcup/) installer over Homebrew, as it works out of the box.
```

Nice! A small suggestion, since the details about env vars is a bit out of blue / potentially overly specific.

waspc/README.md
Martinsos: Makes sense, I wonder why we put this link in the first place! Doesn't seem to be very helpful here.

Martinsos (approved):
Thanks, looks good to me! I approved it, left one minor comment.

Martinsos (approved):
Awesome, thanks!

Issue #2406, opened 2024-12-09
Ensure users don't create a `.wasp` file


sodic:
Wasp relies on creating and managing a `.wasp` directory. If a user creates **a file** with the same name, Wasp breaks. 

We should fix this.

Original discussion: https://github.com/wasp-lang/wasp/pull/2276#discussion_r1774799665

komyg:
Hey, I've opened a PR with a possible fix. Please let me know what you think.

PR #2418, opened 2024-12-14
Ensure users don't create a .wasp file

komyg:
### Description

Fixes: https://github.com/wasp-lang/wasp/issues/2406


waspc/src/Wasp/Project/Analyze.hs
```diff
@@ -245,6 +250,14 @@ constructAppSpec waspDir options externalConfigs parsedPrismaSchema decls = do
 
   return $ runValidation ASV.validateAppSpec appSpec
 
+waspDirExists :: Path' Abs (Dir WaspProjectDir) -> IO (Either String (Path' Abs (Dir WaspProjectDir)))
+waspDirExists waspDir = do
```
sodic, line 254: This function is called `waspDirExists`, implying that it checks whether a `.wasp` directory exists and returns either an error or a path to the directory.

But, in reality, the function does something else:
- If a `.wasp` directory does exist, it returns a path to the existing `.wasp` directory (so far so good)
- If the `.wasp` directory doesn't exist, it returns a path to the non-existing `.wasp` directory (this is the first weird part)
- If a `.wasp` file exists, it returns an error saying that `.wasp` is a file (but the function's name implies it should only care about whether the `.wasp` directory exists and report on that).

Imagine what happens when someone else tries to use this function in a different context (e.g., to really check whether the `.wasp` directory exists). I think they'd end up very confused :)

Anyway, for this issue, we don't care whether the `.wasp` directory exists or not. It can exist, it doesn't need to exist - it's all the same to us.

We only care about whether a `.wasp` file exists, so we should write a function to check for that, and name it accordingly.

waspc/src/Wasp/Project/Analyze.hs
```diff
@@ -245,6 +250,14 @@ constructAppSpec waspDir options externalConfigs parsedPrismaSchema decls = do
...
+  let waspDotWaspPath = waspDir </> [relfile|.wasp|]
+  isFile <- IOUtil.doesFileExist waspDotWaspPath
+  if isFile
+    then return $ Left "The path to the Wasp project is a file, but it should be a directory."
```
sodic, line 258: The "Wasp project" is the user's project, not the stuff in the `.wasp` directory.

Also, users don't really care about the `.wasp` directory, so this message gives out unnecessary information.  If I read it, I'd think Wasp expects me to create a `.wasp` directory (which isn't true).

We should tell users something that conveys the message "Hey, your `.wasp` file can't be called `.wasp`, it needs to have a proper name," no need to mention the `.wasp` directory at all.

waspc/src/Wasp/Project/Analyze.hs
```diff
@@ -77,21 +78,25 @@ analyzeWaspProject ::
   CompileOptions ->
   IO (Either [CompileError] AS.AppSpec, [CompileWarning])
 analyzeWaspProject waspDir options = do
-  waspFilePathOrError <- left (: []) <$> findWaspFile waspDir
+  dirResult <- waspDirExists waspDir
+  case dirResult of
+    Left err -> return (Left [err], [])
+    Right _ -> do
+      waspFilePathOrError <- left (: []) <$> findWaspFile waspDir
```
sodic, line 85: A better place to handle this check is the function `findWaspFile`.

After all, we don't care about the directory at all. All we care about is:
- Do the users have a `*.wasp` file?
- If they do, is it called something other than just `.wasp`?
- If not, tell them to name it properly.

That's on me, the issue could have been explained better. Sorry for the troubles :)

So, to summarize one last time:
> While looking for a `*.wasp` file, we should also ensure that the found file isn't called `.wasp`

While you're at it, you might also want to check whether there are multiple `.wasp` files and report that. 

Because, in theory, Wasp will only find one of  `*.wasp` files. Let's say we have two of them: `something.wasp` and `.wasp`. `findWaspFile` could return the `something.wasp`, letting `.wasp` slip under the radar and cause problems later.

And even if it didn't, allowing multiple `.wasp` files can be confusing.

sodic (changes requested):
Hey @komyg, thanks for the effort!

You have the right idea, and I see you know how to use both Haskell and our codebase (props for using `relfile` and `IOUtil.doesFileExist`), so I have no doubt we'll be merging this soon.

There are some improvements we can do, I left three requests (the biggest of which is moving the logic into `findWaspFile`).

Let me know if you have any questions :)

waspc/src/Wasp/Project/Analyze.hs
komyg: My original plan for this implementation was to create a sort of pipeline in which the result of the previous function would go into the next function, as it is usually done in [railway oriented programming](https://blog.logrocket.com/what-is-railway-oriented-programming/). 

This way, I would be able to avoid having two `case` statements in the `analyzeWaspProject`, because the next `waspFilePathOrError` would only be called if its input was a *right* value, otherwise it will pass through the *left* value unchanged.

Unfortunately, I was unable to do that, and I forgot to change the return type of this function to `boolean` or an empty *right* value.

waspc/src/Wasp/Project/Analyze.hs
komyg: Ok, I've re-phrased the message as: *Invalid file name for the .wasp file. Please rename it to [something].wasp.*. What do you think?

komyg:
Hey @sodic, thank you for your review, it was very through! 

I believe I addressed most of your comments. Could you review again, please?

The only thing left is making sure we don't have more than one .wasp file. I am currently working on that, but it is not done yet. If I can get it done before your next review, then I can be a part of this PR, otherwise, I can make it into another PR.

komyg:
Hey @sodic, I managed to check if we have more than one `*.wasp` or `*.wasp.ts` files in the `.wasp` folder. Could you review again and tell me what you think, please?

waspc/src/Wasp/Project/WaspFile.hs
```diff
@@ -28,24 +29,29 @@ import qualified Wasp.Util.IO as IOUtil
...
 findWaspFile waspDir = do
-  files <- fst <$> IOUtil.listDirectory waspDir
-  return $ case (findWaspTsFile files, findWaspLangFile files) of
-    (Just _, Just _) -> Left bothFilesFoundMessage
-    (Nothing, Nothing) -> Left fileNotFoundMessage
-    (Just waspTsFile, Nothing) -> Right waspTsFile
-    (Nothing, Just waspLangFile) -> Right waspLangFile
+  let dotWaspPath = waspDir </> [relfile|.wasp|]
+  isFile <- IOUtil.doesFileExist dotWaspPath
+  if isFile
```
sodic, line 34: Nice!

This works, but, instead of having two IO procedures (`doesFileExist` and `listDirectory`) that share implicit information, it's better to have a single IO call and check all the constraints on its result.

More precisely, the logic after `IOUtil.listDirectory` assumes someone has already checked the `.wasp` file situation, but this knowledge isn't communicated through the type system. This isn't too bad (it's all in the same function after all), but we could improve it.

So, let's try to keep all the IO querying at the start of the function, and all the logic at its end.

I'll help you out here (since StrongPath can be difficult to deal with :).

Here's a rough sketch of what I had in mind:
```haskell
findWaspFile :: Path' Abs (Dir WaspProjectDir) -> IO (Either String WaspFilePath)
findWaspFile projectDir = do
  filesInProjectDir <- fst <$> IOUtil.listDirectory projectDir
  let filesEndingWithWasp = findAllFilesWithSuffix ".wasp" filesInProjectDir
      filesEndingWithWaspTs = findAllFilesWithSuffix ".wasp.ts" filesInProjectDir
  return $ case (filesEndingWithWasp, filesEndingWithWaspTs) of
    -- We'll check everyhing here
    
 
-- You can place this function in Wasp.Util.StrongPath, it might be useful elsewhere :)
findAllFilesWithSuffix :: String -> [Path p r (File f)] -> [Path p r (File f)]
findAllFilesWithSuffix extension = filter ((extension `isSuffixOf`) . toFilePath)
```

The proposed separation carries an extra benefit. If all the IO stuff is grouped together, it's easy to extract all the logic into a pure function and test it with unit tests.

You can do this too if you want, but don't feel like you have to :)

waspc/src/Wasp/Project/WaspFile.hs
```diff
@@ -28,24 +29,29 @@ import qualified Wasp.Util.IO as IOUtil
...
+    then return $ Left "Invalid file name for the .wasp file. Please rename it to [something].wasp."
+    else do
+      files <- fst <$> IOUtil.listDirectory waspDir
+      return $ case (findWaspTsFile files, findWaspLangFile files) of
+        (tsFiles, langFiles)
+          | not (null tsFiles) && not (null langFiles) -> Left bothFilesFoundMessage
+          | null tsFiles && null langFiles -> Left fileNotFoundMessage
+          | [waspTsFile] <- tsFiles, null langFiles -> Right waspTsFile
+          | null tsFiles, [waspLangFile] <- langFiles -> Right waspLangFile
+          | otherwise -> Left multipleFilesFoundMessage
```
sodic, line 44: This is a very nice display of Haskell's power!

That said, I'd probably stick to pattern matching as much as possible (as opposed to conditional checks) - they're more declarative and better at describing mappings that don't involve too much logic.

Here's what I had in mind:

```haskell
return $ case (filesEndingWithWasp, filesEndingWithWaspTs) of
  ([], []) -> Left fileNotFoundMessage
  (_ : _, _ : _) -> Left bothFilesFoundMessage
  ([], waspTsFiles) -> case waspTsFiles of
    [singleWaspTsFile] -> Right . WaspTs $ castFile (projectDir </> singleWaspTsFile)
    multipleWaspTsFiles -> -- ...
  (waspLangFiles, []) -> case waspLangFiles of
    [singleWaspLangFile]
      | toFilePath (basename singleWaspLangFile) == ".wasp" -> -- ... Double check whether this works please
      | otherwise -> Right . WaspLang $ castFile (projectDir </> singleWaspLangFile)
    multipleWaspLangFiles -> -- ...
```
Pattern matches also help Haskell ensure you've covered all the options. If you use conditional checks (e.g., `null tsFiles`, `[waspTsFile] <- tsFiles`, ...), the compiler can't check the exhaustiveness of your checks.

waspc/src/Wasp/Project/Analyze.hs
sodic: Understood!

Thanks for sharing your thoughts. I enjoyed reading that article.

waspc/src/Wasp/Project/Analyze.hs
sodic: I like it, I'd probably say:
```
Your Wasp file can't be called just '.wasp'. Please rename it to [something].wasp
```
It's a little more direct and tells you exactly what the problem is. But we can keep your message if you prefer it.

sodic (commented):
Nice work, Felipe!

I appreciate the patience. I tried to make it up to you with a thorough review.

Let me know what you think!

waspc/src/Wasp/Project/Analyze.hs
komyg: Ok, I've changed it to your version/

komyg:
Hey @sodic, I've refactored the code based on your suggestions. 

It does look better (and more declarative) using the pattern matching you suggested, however I had to add a `if` to check if the file is called just `.wasp` or `.wasp.ts`, which is invalid. Maybe there is a better way to account for these cases?

Anyway, could you review again, please?

sodic (approved):
Hey @komyg, great work! This is almost ready.

Just a couple of small fixes and we're good to go.

About your question:

> however I had to add a if to check if the file is called just .wasp or .wasp.ts, which is invalid. Maybe there is a better way to account for these cases?

`if` expressions are a fine way to do this, but I personally prefer guards:
```haskell
(waspLangFiles, []) -> case waspLangFiles of
  [singleWaspLangFile]
    | fromRelPath (basename singleWaspLangFile) == ".wasp" -> Left "Name too short"
    | otherwise -> Right . WaspLang $ castFile (projectDir </> singleWaspLangFile)
```
You're free to keep the if if you like it. This is just my preference.

komyg:
Hey @sodic, I've made the changes you requested. Can you review again, please?

waspc/src/Wasp/Util/StrongPath.hs
```diff
@@ -33,3 +37,9 @@ splitAbsExtension path =
 splitRelExtension :: MonadThrow m => SP.Path' (SP.Rel b) (SP.File a) -> m (SP.Path' (SP.Rel b) (SP.File c), String)
 splitRelExtension path =
   first SP.fromPathRelFile <$> P.splitExtension (SP.toPathRelFile path)
+
+findAllFilesWithSuffix :: String -> [SP.Path p r (SP.File f)] -> [SP.Path p r (SP.File f)]
+findAllFilesWithSuffix extension = filter ((extension `isSuffixOf`) . SP.toFilePath)
+
+getFilename :: SP.Path' (SP.Rel b) (SP.File a) -> String
+getFilename = takeFileName . SP.toFilePath
```
sodic, line 45: I don't think we need this function. `fromRelFile` from `StrongPath` does the same thing. We can just use that in all places we're using this.

Or am I missing something?

sodic (commented):
There's just one slight regression we should fix and I'm merging.

waspc/src/Wasp/Util/StrongPath.hs
komyg: I've removed it.

sodic (approved):
All's good to go!

@komyg Great work and thanks for the fix!

PR #2428, opened 2024-12-20
Add an e2e test for the TS config

sodic:
This turned out to be more work than I had anticipated.

I imagined I'd just create a new project and commit build results. It turned out I had to specify everything in Haskell strings.

Can we revise this?

Ideally, we could create a Wasp project and take a snapshot of the result. Why do it from Haskell?


infomiho:
> I imagined I'd just create a new project and commit build results. It turned out I had to specify everything in Haskell strings.

How would the test runner recreate the project with the _latest_ CLI version?

PR #2449, opened 2025-01-11
[WIP] Brought back APIs, moved route generation to proper v7 convention, allowed for user decision on SSR

AlemTuzlak:
### Description

Continuation of moving to react-router v7 PoC with some improvements, notably:
- Allow for full on SSR (depending on if the user wants it or not
- Allow for expanded routing convention where you could create layouts etc.

4. [x] **Breaking change** (fix or feature that would cause existing functionality to not work as expected).


waspc/data/Generator/templates/react-app/src/catchall.tsx
```diff
@@ -1,5 +0,0 @@
-import { router } from './router'
-
-export default function Component() {
-  return router;
```
AlemTuzlak, line 4: the reason this was removed was due to the fact that you would create a browserRouter inside the already existing react-router router. I am not sure if this worked before, I'd expect it to not work due to nested routers but maybe it did.

AlemTuzlak: `entry.client.ts` creates the `HydratedRouter` and wraps the whole app with it

waspc/data/Generator/templates/react-app/src/root.tsx
```diff
@@ -35,6 +36,9 @@ export function Layout({
...
 }
 
+
 export default function Root() {
   return <Outlet />;
-}
\ No newline at end of file
+}
+
+export const ErrorBoundary = DefaultRootErrorBoundary
```
AlemTuzlak, line 44: brought back the error boundary for the whole app

waspc/data/Generator/templates/react-app/src/router.tsx
AlemTuzlak, line 1: removed in favor of `routes.ts`

waspc/data/Generator/templates/react-app/src/routes.ts
```diff
@@ -2,8 +2,34 @@ import {
...
+  {=/ routes =}
+} as const;
+
+const waspDefinedRoutes = [
+  {=# isExternalAuthEnabled =}
+  route("{= oAuthCallbackPath =}", "./auth/pages/OAuthCallback"),
+  {=/ isExternalAuthEnabled =}
+]
+const userDefinedRoutes = Object.entries(routes).map(([routeKey, route]) => {
+  // TODO: This should be the path to the route module file (eg. /src/routes/home.tsx)
```
AlemTuzlak, line 20: I would need a little bit of help here, the current api of the `route` function is to expect the url segment as the first arg and the path to the module as the second, so eg:
```ts
route("login", "./app/login.tsx")
```
Not sure how you generate routes but I'm pretty sure you have to have a path to the actual route, all you have to do is replace the `route` argument with the actual correct thing, so just adapt 
```ts
export const routeNameToRouteComponent = {
  {=# routes =}
  {= name =}: {= targetComponent =},
  {=/ routes =}
} as const;
```
to actually work with modules I guess

waspc/data/Generator/templates/react-app/vite.config.ts
```diff
@@ -16,7 +16,7 @@ const _waspUserProvidedConfig = {};
 
 const defaultViteConfig = {
   base: "{= baseDir =}",
-  plugins: [reactRouter()],
+  plugins: [reactRouterDevtools(), reactRouter()],
```
AlemTuzlak, line 19: I sneakily added in dev tools for react-router, don't tell the contributors! Jokes aside, remove if you don't like it, or think it's useless

waspc/data/Generator/templates/sdk/wasp/api/index.ts
AlemTuzlak, line 1: so this file is a bit suspicious in terms of what it should look like, I see you use axios, which is perfectly fine, but using browser APIs locks you into **drumroll** client-side rendering.

I would need to understand what exactly is the scope of this functionality to help out and change it, from what I see it's supposed to handle authentication. If that's the case and you want to support both SSR and CSR you will need to do the following:
1. Add two storages (session storage on the server, the same storage you've been using up until now for the client)
2. Make the code understand the operation context and fetch from the correct storage.

AlemTuzlak: After looking at this for a second I realise there's no need to change this file, but rather the storage interface, I'll elaborate more there

waspc/data/Generator/templates/sdk/wasp/core/storage.ts
```diff
@@ -6,45 +6,50 @@ export type DataStore = {
   clear(): void
 }
 
-function createLocalStorageDataStore(prefix: string): DataStore {
+function createStorageDataStore(prefix: string): DataStore {
```
AlemTuzlak, line 9: I made this generic as if you want to support SSR you'll need to not rely on web api's

waspc/data/Generator/templates/sdk/wasp/core/storage.ts
```diff
@@ -6,45 +6,50 @@ export type DataStore = {
...
   }
 }
 
-export const storage = createLocalStorageDataStore('wasp')
+export const storage = createStorageDataStore('wasp')
 
-function ensureLocalStorageIsAvailable(): void {
-  if (!window || !window.localStorage) {
-    throw new Error('Local storage is not available.')
+// TODO: Make this function work in the server context as well.
```
AlemTuzlak, line 49: this will be a bit trickier, if you want to use the react-router built in session handling it would require the request object as well which would have to be piped to here. Up for discussion with you guys what you'd want to do

waspc/data/Generator/templates/react-app/react-router.config.ts
```diff
@@ -1,6 +1,12 @@
...
+let isSSR = false;
+
+{=# ssr.isDefined =}
+ isSSR = Boolean("{= ssr.value =}")
+{=/ ssr.isDefined =}, 
+
 export default {
   appDirectory: "src",
-  ssr: false,
+  ssr: isSSR,
```
AlemTuzlak, line 11: I'm just a good ol' JS dev, I don't speak in haskell, but you'll get the idea

waspc/data/Generator/templates/react-app/src/routes.ts
```diff
@@ -2,8 +2,34 @@ import {
...
-  // * matches all URLs, the ? makes it optional so it will match / as well
-  route("*?", "catchall.tsx"),
+  layout(
+    // TODO: This should maybe always be defined, if not, 
+    // TODO: the code should be (if defined => layout, otherwise => spread the routes normally in the array)
+    {=# rootComponent.isDefined =}
+    "{= rootComponent.importIdentifier =}"
+    {=/ rootComponent.isDefined =}, 
+    [...waspDefinedRoutes, ...userDefinedRoutes]
+  ),
```
AlemTuzlak, line 34: brought back the functionality from the router.ts but the issue is that layout needs to be defined, so it will either be always defined OR:
```ts
isDefined ? layout(path, children) : children
```

AlemTuzlak: also, another cool this opens up as a possibility for you guys is to add layout routes etc as functionalities in wasp, eg in your wasp config you could do something like:
layout RootLayout { path: "/", to: MainPage }
route RootRoute { path: "/", to: MainPage, layout: RootLayout }

This would increase complexity obviously on your side, but would also make users happy, I'd assume

infomiho:
@AlemTuzlak Hey, sorry for the _extra long wait_ for our response! Thank you for your interest in Wasp, it means a lot. Thank you for the some of insights in this PR that will help us once we decide to tackle SSR in Wasp 🙏 

I've tried running this locally and I found out that we'll need to update a couple of things to get the RR7 fully functional. Instead of going back forth in this PR with your (I'm not sure you'd be up for pair programming in Haskell 😂 ) I'll close this PR and link to it as a source of knowledge when we go ahead with SSR (soon). 

When we have some functional SSR implementation, I'd love to ping you and get your feedback if it all makes sense, if that's okay with you?

PR #2450, opened 2025-01-12
Docs for user defined env vars validation

infomiho:
This PR will cover:
- [x] The way users access the validated `env` object on the server and the client
- [x] How users can define their custom env vars validation

### DIFF
--- a/web/docs/project/env-vars.md
+++ b/web/docs/project/env-vars.md
@@ -399,7 +399,7 @@ const analyticsId: string = env.REACT_APP_ANALYTICS_ID
 </TabItem>
 </Tabs>
 
-You can use `import.meta.env.REACT_APP_SOME_VAR_NAME` directly in your code, but it's not recommended because it's not validated and can lead to runtime errors if the env var is not defined.
+You can use `import.meta.env.REACT_APP_SOME_VAR_NAME` directly in your code. We don't recommend this. `import.meta.env` isn't validated and missing env vars can cause runtime errors.
 
 ### Server Env Vars
 
--- a/web/docs/project/env-vars.md
+++ b/web/docs/project/env-vars.md
@@ -229,7 +229,7 @@ If your code requires some environment variables, you usually want to ensure tha
 
 :::
 
-Let's look at an example of defining env vars validation:
+Take a look at an example of defining env vars validation:
 
 <Tabs groupId="js-ts">
 <TabItem value="js" label="JavaScript">
--- a/web/docs/project/env-vars.md
+++ b/web/docs/project/env-vars.md
@@ -218,15 +218,15 @@ We talk about how to define env vars for each deployment option in the [deployme
 
 ## Custom Env Var Validations
 
-If your code requires some custom env variables, you might want to ensure that they are defined and have the correct values. You can define your env vars validation by defining a [Zod object schema](https://zod.dev/) and importing it in the `main.wasp` file.
+If your code requires some environment variables, you usually want to ensure that they are correctly defined. In Wasp, you can define your environment variables validation by defining a [Zod object schema](https://zod.dev/?id=basic-usage) and telling Wasp to use it.
 
 :::info What is Zod?
 
-[Zod](https://zod.dev/) is a library that enables you to define what you expect from your data. For example, you can use Zod to define that a string should be an email address, or that a number should be between 0 and 100.
+[Zod](https://zod.dev/) is a library that lets you define what you expect from your data. For example, you can use Zod to define that a value should be a string that's a valid email address or that a value should be a number between 0 and 100.
 
 :::
 
-This is how you can define a custom env var validation schema for the client and the server:
+Let's look at an example of defining env vars validation:
 
 <Tabs groupId="js-ts">
 <TabItem value="js" label="JavaScript">
@@ -268,6 +268,8 @@ export const clientEnvValidationSchema = defineEnvValidationSchema(
 )
 ```
 
+We are using the `defineEnvValidationSchema` helper to get type-checking for the Zod schema.
+
 </TabItem>
 </Tabs>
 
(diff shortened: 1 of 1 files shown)


web/docs/project/env-vars.md
```diff
@@ -217,21 +218,156 @@ We talk about how to define env vars for each deployment option in the [deployme
...
+
+export const serverEnvValidationSchema = z.object({
+  STRIPE_API_KEY: z.string({
+    required_error: 'STRIPE_API_KEY is required.',
+  }),
+})
+
+export const clientEnvValidationSchema = z.object({
+  REACT_APP_NAME: z.string().default('TODO App'),
+})
```
sodic, line 245: The API should not change depending on whether the user uses TS or JS. We should always tell users to call `defineEnvValidationSchema`.

Since JS and TS don't differ in runtime, neither should our API. The only differences should be in the type space.

It's better for both us and our users:
- Users will appreciate (or maybe even expect) the consistency.
- Two APIs mean maintaining two runtime behaviors/execution paths. The difference might be trivial, but it's there. Plus, we might one day decide to do something extra in `defineEnvValidationSchema`.

I noticed the same pattern with `userSignupFields`, and I'd like to change it there too (and all other places, if there are more).

> [!NOTE]
> I know [Vue does it like this](https://vuejs.org/guide/typescript/overview#definecomponent), but I still don't like it :)
>
> I guess they did it to stay somewhat backward-compatible. If they could start over, they'd probably make it consistent (but this is just my theory, maybe they prefer not to burden the JS users).

web/docs/project/env-vars.md
```diff
@@ -217,21 +218,156 @@ We talk about how to define env vars for each deployment option in the [deployme
...
+)
+
+export const clientEnvValidationSchema = defineEnvValidationSchema(
+  z.object({
+    REACT_APP_NAME: z.string().default('TODO App'),
+  })
+)
+```
+
+We are using the `defineEnvValidationSchema` helper to get type-checking for the Zod schema.
```
sodic, line 271: I suggest changing the subject to emphasize a point and address the user directly:
```suggestion
The defineEnvValidationSchema` function ensures your Zod schema is type-checked.
```

web/docs/project/env-vars.md
```diff
@@ -217,21 +218,156 @@ We talk about how to define env vars for each deployment option in the [deployme
 
 ## Custom Env Var Validations
 
-TODO: when the Zod validation PRs are merged, describe how users can define their own validations.
+If your code requires some environment variables, you usually want to ensure that they are correctly defined. In Wasp, you can define your environment variables validation by defining a [Zod object schema](https://zod.dev/?id=basic-usage) and telling Wasp to use it.
+
+:::info What is Zod?
+
+[Zod](https://zod.dev/) is a library that lets you define what you expect from your data. For example, you can use Zod to define that a value should be a string that's a valid email address or that a value should be a number between 0 and 100.
```
sodic, line 225:
```suggestion
[Zod](https://zod.dev/) is a library that lets you define what you expect from your data. For example, you can use Zod to define that:
- A value should be a string that's a valid email address.
- A value should be a number between 0 and 100.
- etc.
```
Plus, maybe add an explanation why that is helpful (something along the lines of "this is great because it warns us about improperly defined environment variables as soon as possible instead of failing somewhere else down the line").

web/docs/project/env-vars.md
```diff
@@ -217,21 +218,156 @@ We talk about how to define env vars for each deployment option in the [deployme
...
+
+:::
+
+Let's look at an example of defining env vars validation:
```
sodic, line 229:
```suggestion
Take a look at an example of defining env vars validation:
```
Addresses the user directly, but no big deal. We use "let's take a look at" all the time, although we might consider switching it up for something like this.

web/docs/project/env-vars.md
```diff
@@ -217,21 +218,156 @@ We talk about how to define env vars for each deployment option in the [deployme
...
+  client: {
+    envValidationSchema: import { clientEnvValidationSchema } from "@src/env",
+  },
+  server: {
+    envValidationSchema: import { serverEnvValidationSchema } from "@src/env",
+  },
+}
+```
+
+We defined schemas for both the client and the server env vars and told Wasp to use them. Wasp merges your env validation schemas with the built-in env vars validation schemas when it validates the `process.env` object on the server and the `import.meta.env` object on the client.
```
sodic, line 288: First sentence: we -> you.

The rest of the text already uses the second person, so that's great.

web/docs/project/env-vars.md
```diff
@@ -217,21 +218,156 @@ We talk about how to define env vars for each deployment option in the [deployme
...
+import { env } from 'wasp/server'
+
+const stripeApiKey: string = env.STRIPE_API_KEY
+```
+
+Read more about the env object in the [API Reference](#api-reference).
 
 ## API Reference
 
+There are **Wasp-defined** and **user-defined** env vars. Wasp has built-in validation for its env vars and you can define your own validation for your env vars.
```
sodic, line 302: I'd flip the ordering to better emphasize the main point:
```suggestion
There are **Wasp-defined** and **user-defined** env vars. Wasp already comes with built-in validation for Wasp-defined env vars. For your env vars, you can define your own validation.
```

web/docs/project/env-vars.md
```diff
@@ -240,45 +376,101 @@ console.log(env.REACT_APP_SOME_VAR_NAME)
...
+// User-defined
+const analyticsId: string = env.REACT_APP_ANALYTICS_ID
 ```
 
 </TabItem>
 </Tabs>
 
-The `env` object is a validated object that Wasp provides to access client env vars.
-
 You can use `import.meta.env.REACT_APP_SOME_VAR_NAME` directly in your code, but it's not recommended because it's not validated and can lead to runtime errors if the env var is not defined.
```
sodic, line 389:
```suggestion
You can use `import.meta.env.REACT_APP_SOME_VAR_NAME` directly in your code. We don't recommend this. `import.meta.env` isn't validated and missing env vars can cause runtime errors.
```

web/docs/project/env-vars.md
```diff
@@ -240,45 +376,101 @@ console.log(env.REACT_APP_SOME_VAR_NAME)
...
+// User-defined
+const stripeApiKey: string = env.STRIPE_API_KEY
 ```
 
 </TabItem>
 </Tabs>
 
-The `env` object is a validated object that Wasp provides to access server env vars.
-
 You can use `process.env.SOME_SECRET` directly in your code, but it's not recommended because it's not validated and can lead to runtime errors if the env var is not defined.
```
sodic, line 476: Same suggestion as above. This sentence is kind of a mouthful (that's what she said).

web/docs/project/env-vars.md
```diff
@@ -240,45 +376,101 @@ console.log(env.REACT_APP_SOME_VAR_NAME)
...
 </TabItem>
 <TabItem value="ts" label="TypeScript">
 
-```ts title="src/App.ts"
+```ts title="src/stripe.ts"
 import { env } from 'wasp/server'
 
-console.log(env.SOME_SECRET)
+// Wasp-defined
+const serverUrl: string = env.WASP_SERVER_URL
```
sodic, line 467: Why the `: string` part?

Maybe you wanted to emphasize that "this is now TypeScript." I understand the motivation, but we should document the TS code our users should write. If that's the same as JS code, so be it.

It's even better if the type-checker does its job silently. The fewer changes happening when you toggle JS/TS in the docs, the better - it means TS and Wasp are doing a good job.

Applies to other code boxes too.

sodic (changes requested):

web/docs/project/env-vars.md
infomiho: Good point 👍 

I agree with both this and the `userSignupFields` change, created an issue: https://github.com/wasp-lang/wasp/issues/2455

web/docs/project/env-vars.md
infomiho: I had the intent of showing it's no longer `string | undefined` but actually a validated `string` but if that didn't come across, I'll remove it 👍

PR #2444, opened 2025-01-10
Treat user code as a referenced project

sodic:
Fixes #2247

PR #2461, opened 2025-01-22
Make Wasp TS Config projects work with TS project references

sodic:
Extends #2444 support to Wasp TS config projects,


infomiho (approved):
Some suggestions, but good to go

Issue #2464, opened 2025-01-23
Add `tailwindcss` as an explicit dependecy and validate that it exists in user's `package.json`


infomiho:
Same issue as https://github.com/wasp-lang/wasp/issues/2408

Since Tailwind v4 was [released](https://github.com/tailwindlabs/tailwindcss/releases/tag/v4.0.0) all the Tailwind specific packages that require Tailwind as a peer dep will get the latest (v4) but we only support v3 at the moment. This happens because we don't say explicitly in the user's `package.json` that we require Tailwind v3.

We should:
- [ ] Update Open Saas to include the `tailwindcss@^3.2.7` in the `package.json` https://github.com/wasp-lang/open-saas/pull/356
- [ ] Update `embeddings` starter template since it uses Tailwind as well
- [ ] Add validation rules for `package.json` to require the `tailwindcss` dep if Tailwind is used https://github.com/wasp-lang/wasp/pull/2465

PR #2465, opened 2025-01-23
Validates `tailwindcss` if Tailwind is used

infomiho:
~~This PR aims to add validation for the `tailwindcss` dep if Tailwind is used. For more details: #2464~~

This PR did a refactor of the `package.json` and `tsconfig.json` validation:
- splits the previous implementation into two pieces:
  - reading of the external config files lives in `Wasp.Project`
  - validation lives in `Wasp.Generator.Valid`
- adds the contents of the external config files to the `AppSpec` so it can validated just before running the Generator

Left to do:
- [x] Resolve any conflicts related to `PackageJson` and `TsConfig` modules
- [x] ~~Validate that the Tailwind dep is in `devDependencies`~~


waspc/src/Wasp/AppSpec.hs
```diff
@@ -149,10 +148,6 @@ resolveRef spec ref =
     $ find ((== refName ref) . fst) $
       getDecls spec
 
-doesConfigFileExist :: AppSpec -> Path' (Rel WaspProjectDir) File' -> Bool
```
infomiho, line 152: I needed the information if Tailwind is used (which is computed using `doesConfigFileExist`) _before we have the AppSpec_ so I pulled it out of here and rewritten it so it doesn't need the AppSpec.

waspc/src/Wasp/Generator/ConfigFile.hs
```diff
@@ -24,12 +26,16 @@ asProjectRootDirConfigFile = (webAppRootDirInProjectRootDir </>) . castRel
...
+isTailwindUsed :: [CF.ConfigFileRelocator] -> Bool
+isTailwindUsed configFileRelocators =
+  doesConfigFileExist tailwindConfigFile
+    && doesConfigFileExist postcssConfigFile
+  where
+    doesConfigFileExist :: Path' (Rel WaspProjectDir) File' -> Bool
+    doesConfigFileExist file =
+      isJust $ find ((==) file . CF._pathInWaspProjectDir) configFileRelocators
 
--- | Establishes the mapping of what config files to copy and where from/to.
```
infomiho, line 32: Ormolu

waspc/src/Wasp/Generator/ConfigFile.hs
```diff
@@ -24,12 +26,16 @@ asProjectRootDirConfigFile = (webAppRootDirInProjectRootDir </>) . castRel
...
-isTailwindUsed :: AppSpec -> Bool
-isTailwindUsed spec =
-  doesConfigFileExist spec tailwindConfigFile
-    && doesConfigFileExist spec postcssConfigFile
+isTailwindUsed :: [CF.ConfigFileRelocator] -> Bool
+isTailwindUsed configFileRelocators =
+  doesConfigFileExist tailwindConfigFile
+    && doesConfigFileExist postcssConfigFile
+  where
+    doesConfigFileExist :: Path' (Rel WaspProjectDir) File' -> Bool
```
infomiho, line 34: This is only used for `isTailwindUsed` so I inlined it even though it might be useful in the future - let's pull it out when it becomes necessary?

waspc/src/Wasp/Generator/ExternalConfig/PackageJson.hs
```diff
@@ -7,22 +7,29 @@ import qualified Data.Map as M
 import qualified Wasp.ExternalConfig.PackageJson as P
 import Wasp.Generator.Common (prismaVersion)
 import Wasp.Generator.ExternalConfig.Common (ErrorMsg)
+import Wasp.Generator.SdkGenerator.Common (tailwindCssVersion)
 import Wasp.Generator.WebAppGenerator.Common (reactRouterVersion, reactVersion)
 
-validatePackageJson :: P.PackageJson -> [ErrorMsg]
-validatePackageJson packageJson =
+validatePackageJson :: Bool -> P.PackageJson -> [ErrorMsg]
+validatePackageJson isTailwindUsed packageJson =
```
infomiho, line 14: I had to thread through the `isTailwindUsed` info unfortunately. Some sort of Reader would be useful here, but I didn't want to over-engineer it.

waspc/src/Wasp/Project/Analyze.hs
```diff
@@ -61,25 +66,25 @@ analyzeWaspProject waspDir options = do
           analyzeWaspFile waspDir prismaSchemaAst waspFilePath >>= \case
             Left errors -> return (Left errors, [])
             Right declarations ->
-              EC.analyzeExternalConfigs waspDir (getSrcTsConfigInWaspProjectDir waspFilePath) >>= \case
+              EC.analyzeExternalConfigs isTailwindUsed waspDir (getSrcTsConfigInWaspProjectDir waspFilePath) >>= \case
```
infomiho, line 69: I wasn't sure if I wanted to thread the `configFiles` all the way down to `validatePackageJson` and the call the `G.CF.isTailwindUsed` check. Since `validatePackageJson` doesn't need the list of configs, I decided not to.

waspc/src/Wasp/Generator/ExternalConfig/PackageJson.hs
```diff
@@ -7,22 +7,29 @@ import qualified Data.Map as M
...
+        validate ("react", show reactVersion) IsListedWithExactVersion,
+        validate ("react-dom", show reactVersion) IsListedWithExactVersion
+      ]
+        ++ tailwindValidations
+    )
   where
     validate = validateDep packageJson
 
+    tailwindValidations =
+      [ validate ("tailwindcss", show tailwindCssVersion) IsListedWithExactVersion | isTailwindUsed
```
infomiho, line 30: If Tailwind is used, validate that the `tailwindcss` package is present in the `package.json` with the correct version.

waspc/src/Wasp/Generator/ExternalConfig/PackageJson.hs
Martinsos: I get what you mean, and I agree it would be much going for Reader now.
That said, having just a Bool like this is not very nice, what if we add one or two more conditionals like this, would we have 2 more arguments? I would immediately go for grouping this into a record, even it if has just one field for now. Might be called something like `ValidationConfig` or `ValidationPreconditions` or `ValidationContext` (I like that one!) or whatever you think is the best name. That will also make moving to Reader easier in the future if we want it, since you would just move that record into Reader. But it will also make this function immediately more readable and extendable.

waspc/src/Wasp/Project/Analyze.hs
Martinsos: Hm maybe they don't need it all, but this function is called `analyzeExternalConfigs` and now it takes a bool, waspDir, and then a specific config file (srcTsConfig). That is now quite messy, because you could look at those args as:
arg#1: info about config file
arg#2: context
arg#3: config file

Usually in Haskell we aim for "context" args at start, and then "core" args after that.
Since this function is about config files, any amount of them, then I would also find it more natural that it takes a list or record of config files, note one per argument.

So what I would consider is turning this function's signature into `:: WaspDir -> ConfigFiles -> ReturnType` (I used a bit pseudo names here).

And it's not a problem if it takes a bit too much info -> it will analyze what it needs to analyze. We don't want to guess how it will analyze them, that is implementation detail.

waspc/src/Wasp/Project/Analyze.hs
```diff
@@ -61,25 +66,25 @@ analyzeWaspProject waspDir options = do
...
 constructAppSpec ::
   Path' Abs (Dir WaspProjectDir) ->
   CompileOptions ->
   EC.ExternalConfigs ->
+  [CF.ConfigFileRelocator] ->
   Psl.Schema.Schema ->
   [AS.Decl] ->
   IO (Either [CompileError] AS.AppSpec, [CompileWarning])
-constructAppSpec waspDir options externalConfigs parsedPrismaSchema decls = do
+constructAppSpec waspDir options externalConfigs configFiles parsedPrismaSchema decls = do
```
Martinsos, line 81: What confuses me a bit is that this function takes both external configs and congif file relocators. Actually those two kind of make sense -> I guess it takes them, and then rules how to relocate them, I am not up to date with this piece of code completely so I am not quite sure -> but then for relocators you use `configFiles` name and that confuses me absolutely, because you said relocators but now it is files.

waspc/src/Wasp/Project/ExternalConfig.hs
```diff
@@ -23,11 +23,12 @@ data ExternalConfigs = ExternalConfigs
   deriving (Show)
 
 analyzeExternalConfigs ::
+  Bool ->
```
Martinsos, line 26: In general, seeing a Bool like this, where I can't tell from signature at all what it could be, is a bit suspicious. I commented more above why this one is probably not the right choice here.

waspc/src/Wasp/Project/ExternalConfig/PackageJson.hs
```diff
@@ -18,11 +18,11 @@ import Wasp.Project.Common
   )
 import qualified Wasp.Util.IO as IOUtil
 
-analyzePackageJsonFile :: Path' Abs (Dir WaspProjectDir) -> IO (Either [String] P.PackageJson)
-analyzePackageJsonFile waspProjectDir = runExceptT $ do
+analyzePackageJsonFile :: Bool -> Path' Abs (Dir WaspProjectDir) -> IO (Either [String] P.PackageJson)
```
Martinsos, line 21: We should probably rework this signature also, with similar approach as to how I suggested in other comments.

waspc/src/Wasp/Generator/ConfigFile.hs
Martinsos: If you know it is a general utility, I would pull it out immediately, since then it is easier for somebody else to see it there and use it. If it was more tied down to this specific place then mabye not, but it seems quite ready to isolate right now, and I can easily imagine somebody else will want to check if certain congfi file exists, so yeah, let's go for it right now.

waspc/src/Wasp/AppSpec.hs
Martinsos: How is it that you needed it before?
Are we getting into a situtation where we need a lot of info from AppSpec before AppSpec exists, pointing at some potential design issue, or do you think this is ok?

Martinsos (changes requested):
All makes sense, but I believe we can improve data flow / function signatures a bit, left comments on that!

infomiho:
@Martinsos could you take another look? I've read through your comments and took another stab at the implementation, let me summarize my thought process and what I did. 

First, I'll jump into my thoughts about the way we support Tailwind and then I'll go into the `ExternalConfigAnalysisContext` 

1. Before we can construct the `AppSpec`, we need to analyze/validate the `package.json` and `tsconfig.json` files because they need to be **valid** before we continue with generating the project. (We call these `externalConfigs`)
2. Before we can analyze the `package.json` we need to know if Tailwind is used because we want to validate Tailwind CSS only if it's used.
3. Users enable Tailwind CSS implicitly by defining `tailwind.config.cjs` _and_ `postcss.config.cjs` in their project dir. 
4. We devised a general way of relocating user defined configs, you specify the source path and the target path. We use this only for Tailwind related config files. (We call these `configFiles` with only Tailwind and PostCSS configs inside)

We were going for the magical DX where we install Tailwind dep in the background when we detect the Tailwind config files. Users didn't have to install the `npm` package themselves. Given the new validation rules for the `tailwindcss` dep (which we _have to have_ due to peer deps issues) users need to have the `tailwindcss` dep in their `package.json` anyways. So, Wasp is not really doing anything anymore. When we migrate to Tailwind 4, I propose we make Tailwind a user-land feature: #2483 

In general, I feel like the mechanism of configuring your Wasp project by Wasp detecting presence of certain files like `tailwind.config.cjs` and `postcss.config.cjs` is kinda confusing. Based on our experience with users wanting to have more control, I believe the more explicit configuration the better. So, I think we should maybe get rid of the `ConfigFileRelocator` concept all together.

Given that I think we should get rid of `ConfigFileRelocator` as a concept, I've made parts of it less general and made it specific to Tailwind. It's much more clear to me what is now going on in the AppSpec, there is no longer `configFiles` but `tailwindConfigFilesRelocators` - super specific. Let me know what you think.

Also, I liked your suggestion to have a `...Context` for the external config files analysis, so I did just that. All the related functions now use the `ExternalFilesAnalysisContext` object which made it much easier to implement.

I've also tried to introduce the `AppSpecConstructionContext` so maybe make it easier in the future to introduce new into without introducing new function params. I'm not sure about this one, looks okay to me, let me know what you think.

Martinsos (commented):
@infomiho I think this is a step in right direction, but you might have went a bit too far with it.

The point of the "context" was to enable passing around some data without having to pass every single argument individually, where arguments were really implementation details of other, deeper layers of logic, specifically specific config files like tailwind and so on -> top-level external config files fuction doesn't want to list every congif file as its argument, but take them as one big data chunk that it can later dissect and delegate to smaller functions if it wants.

That was kind of the idea. Also you put in some stuff that might not even belong there, like waspDir -> I am not sure about that yet though. But I would say, look at the idea of "context" as a way to wire stuff through while controlling amount of information that different layers of logic need to know about, and also look at it more from the "site of call" then from its own position -> the context shouldn't be a standalone thing that is used in other places, conceptually, unless it is indeed super clear and standalone concept (which I don't think is true here), but instead should be kind of dependant / modeled by the usage site / consumer, in this case your validation logic in ExternalCOnfig module.

I commented more in the individual comments, read them all before you start acting on any of them, to get the full picture.

We can also discuss further if needed, if I was unclear about something -> I have a feeling here what is right / wrong, but I am a bit vague in telling you exactly what it should look like, because I don't have a clear picture myself yet, I would have to get my hands dirtier if I wanted to figure that out exactly.

Martinsos:
> 1. Before we can construct the `AppSpec`, we need to analyze/validate the `package.json` and `tsconfig.json` files because they need to be **valid** before we continue with generating the project. (We call these `externalConfigs`)

But if you need to validate the package.json and tsconfig.json files before generating the project, you could also do that after the construction of AppSpec and still before the generation of the project, that is our Validate.hs step. Or is there some other reason to do it before?

waspc/src/Wasp/AppSpec.hs
infomiho: I've now moved the job of validating the `package.json` and `tsconfig.json` after the AppSpec is constructed (I've now included the contents of those files in the AppSpec since they describe the user input for the generation step, so it made sense to do it like that)

waspc/src/Wasp/Generator/DepVersions.hs
```diff
@@ -0,0 +1,19 @@
+module Wasp.Generator.DepVersions
```
infomiho, line 1: Extracted the versions from `Common` to prevent cyclical imports.

infomiho:
I've now updated the concept:
1. We read the external configs (package.json, tsconfig.json) in the analyze phase and store them in the AppSpec
2. We run the validation of the external configs _just before_ we run the generator
3. The validations were specific to the generator, so they live in the Generator module

waspc/src/Wasp/Generator/Valid/PackageJson.hs
```diff
@@ -35,28 +39,38 @@ validateRuntimeDependencies packageJson =
...
     [ validateDevelopment ("vite", show viteVersion),
       validateDevelopment ("prisma", show prismaVersion)
     ]
+      ++ tailwindValidations
   where
-    validateDevelopment spec = validatePackageJsonDependency packageJson spec RequiredDevelopment
+    validateDevelopment packageSpec = validatePackageJsonDependency packageJson packageSpec RequiredDevelopment
+
+    tailwindValidations =
+      [ validateDevelopment ("tailwindcss", show tailwindCssVersion) | TCF.isTailwindUsed spec
```
infomiho, line 56: The main point of this PR is this one line :D

Martinsos: Oh so we are still validating Tailwind, I thought we might give up on that due to moving Tailwind to user land, and just focus in this PR on the useful refactors we did on the way?

waspc/src/Wasp/Generator.hs
```diff
@@ -41,15 +42,18 @@ import Wasp.Util ((<++>))
...
-      (setupGeneratorWarnings, setupGeneratorErrors) <- runSetup spec dstDir sendMessage
-      return (generatorWarnings ++ setupGeneratorWarnings, setupGeneratorErrors)
+      case generatorResult of
+        Left generatorErrors -> return (generatorWarnings, toList generatorErrors)
+        Right fileDrafts -> do
+          synchronizeFileDraftsWithDisk dstDir fileDrafts
+          writeDotWaspInfo dstDir
+          (setupGeneratorWarnings, setupGeneratorErrors) <- runSetup spec dstDir sendMessage
+          return (generatorWarnings ++ setupGeneratorWarnings, setupGeneratorErrors)
+    appSpecValidationErrors -> return ([], appSpecValidationErrors)
```
Martinsos, line 56: Tiny thing, but when I can I prefer to start with the error case, becuase usually it is much shorter and I don't have to go hunting down for it.

So in this case, you could do

```hs
writeWebAppCode spec dstDir sendMessage = do
  case validateAppSpec spec of
    errs@(_:_) -> return ([], errs)
    [] -> do
      let (generatorWarnings, generatorResult) = runGenerator $ genApp spec

      case generatorResult of
        Left generatorErrors -> return (generatorWarnings, toList generatorErrors)
        Right fileDrafts -> do
          synchronizeFileDraftsWithDisk dstDir fileDrafts
          writeDotWaspInfo dstDir
          (setupGeneratorWarnings, setupGeneratorErrors) <- runSetup spec dstDir sendMessage
          return (generatorWarnings ++ setupGeneratorWarnings, setupGeneratorErrors)
```

waspc/src/Wasp/Generator/Valid.hs
```diff
@@ -0,0 +1,13 @@
+module Wasp.Generator.Valid where
```
Martinsos, line 1: It is anti pattern to blidnly export everthing from the module in Haskell.
Instead, use `()` and explicitly list in it what you are exporting.

waspc/src/Wasp/Generator/Valid/PackageJson.hs
```diff
@@ -18,14 +22,14 @@ data PackageRequirement
 
 type PackageSpecification = (P.PackageName, P.PackageVersion)
 
-validatePackageJson :: P.PackageJson -> [ErrorMsg]
-validatePackageJson packageJson =
-  validateRuntimeDependencies packageJson
-    ++ validateDevelopmentDependencies packageJson
-    ++ validateOptionalDependencies packageJson
+validatePackageJson :: AppSpec -> [GeneratorError]
```
Martinsos, line 25: I am curious, it seems that all these validate functions operate exclusivly on PackageJson. Would it then make sense to actually have them take only PackageJson, and not the whole AppSpec, as it was before? What was you rmotivation to change that? How likely is it that they will ever need sometihng out of purely PackageJson?

waspc/src/Wasp/Generator/Valid/TsConfig.hs
```diff
@@ -32,12 +34,14 @@ type FieldPath = [String]
 instance Show FullyQualifiedFieldName where
   show (FieldName fieldPath) = intercalate "." fieldPath
 
-validateSrcTsConfig :: T.TsConfig -> [ErrorMsg]
-validateSrcTsConfig tsConfig =
-  validateRequiredField (FieldName ["include"]) (T.include tsConfig) ["src"]
-    ++ validateCompilerOptions (T.compilerOptions tsConfig)
+validateSrcTsConfig :: AppSpec -> [GeneratorError]
```
Martinsos, line 37: Hm you also went more general with arugments here. Not sure if that is good or not, similar like I asked above.

waspc/src/Wasp/Project/ExternalConfig/PackageJson.hs
```diff
@@ -18,21 +17,18 @@ import Wasp.Project.Common
   )
 import qualified Wasp.Util.IO as IOUtil
 
-analyzePackageJsonFile :: Path' Abs (Dir WaspProjectDir) -> IO (Either [String] PackageJson)
-analyzePackageJsonFile waspProjectDir = runExceptT $ do
+readPackageJsonFile :: Path' Abs (Dir WaspProjectDir) -> IO (Either [String] PackageJson)
```
Martinsos, line 20: You should proably convert the type to `Either String PackageJson` -> doesn't make much sense that it can return mulitple errors.

Same goes for the function below.

Martinsos (commented):
@infomiho reviewed again! Looks good in general, left some smaller comments, I think we are close to done. Please check them out.

Also, could you please go through the older convos and either resolve them if they are not revelant anymore, or comment on them if they need commenting? There is quite a few of those and even though the code the point to is outdated, I am not confident enough to disregard them since I don't have enough context any more.

waspc/src/Wasp/Generator/Valid/PackageJson.hs
infomiho: We still need this validation because without this dep in the `package.json` the Tailwind setup breaks. Until we move it 100% to the user-land, we need to have the validation.

waspc/src/Wasp/Generator/Valid/TsConfig.hs
infomiho: You are right, I had the initial idea of "hey this works on app spec validation, it should get the whole thing" but in reality, these are already specialized functions that don't need the whole spec.

waspc/src/Wasp/Generator/Valid/PackageJson.hs
infomiho: We need two pieces of info: package json contents + if tailwind is used, those live in the spec. Now the question is raised, do we send the `isTailwindUsed` as an extra bool param or do we construct an object called "PackageJsonValidationContext" with just that info. I'll go for the latter and let's see what you think :)

waspc/src/Wasp/Generator.hs
Martinsos: Yup, that means the non-empty list!
Hm I get what you mean with readibility. It is one of those things where I am not sure if we should "spare" people or just have them learn what it means. Lists are so basic in Haskell that I don't really mind having this. Even if you don't undrstand what it is, thre are only two cases, the second is empty list -> not hard to guess what this one is. There is also `errs` that kind of makes it clear it is "hey we found some errors".

In Haskell I don't try to understand every single symbol in the code, I instead rely on compiler and on types + some understanding of the code I am reading, that is enough for me to undrestand what is happening, and then usually I will figure out exact logic bit by bit only if I need that level of understanding (whic I often don't).

Do as you like here of rcouse, as I said, just an idea!

waspc/src/Wasp/Generator/Valid/PackageJson.hs
Martinsos: Ok, in that case let's go for whole spec, as you have now, sounds good.

waspc/src/Wasp/Generator/Valid/TsConfig.hs
Martinsos: Yeah it can be hard to make taht decision. It is more about the semantics than what really happens. Up to you to decide what is right, I am ok with both.

waspc/src/Wasp/Generator/Valid/PackageJson.hs
infomiho: I've actually went the other way - for the `tsconfig.json` validation, we only need the ts config contents. For the `package.json` we need the contents + info about tailwind (which I've packed into a context object) which looked clean to me.

infomiho:
@Martinsos **I think we should remove the Tailwind dep validation because I made an error earlier.** I've updated all the starters to include `tailwindcss` but in the `dependencies` section and not the `devDepedencies` section (as recommended by the [official docs](https://v3.tailwindcss.com/docs/installation)). 

With this PR, we want to do the "right thing" and validate that the `tailwindcss` dep is in the `devDepedencies`. If we now include this validation that means it will give an error for _all projects_ and people will have to move the dep. 

The issue for me is that people won't get any benefit by moving the dep other then to be semantically correct by including the dep in the right place. Wasp apps work correctly even with that dep in the `dependencies`. 

**Given that we want to make Tailwind a user land feature soon**, and almost all Wasp apps are functional already, maybe we shouldn't annoy people with this temporary change. We could just [update the docs](https://wasp.sh/docs/project/css-frameworks#enabling-tailwind-step-by-step) to say they should add the dep.

**Alternatively**, we go ahead with this change, say this is a breaking change, include it in the migration guide and people will just do it. I'm not sure what is the size of annoyance.

FranjoMindek:
The fact that v4 now recommends it as `dependency` while v3 was `devDependency` confuses things more:
https://github.com/tailwindlabs/tailwindcss/discussions/13854

Might be best to just drop tailwind to userland and not validate.

Martinsos:
> @Martinsos **I think we should remove the Tailwind dep validation because I made an error earlier.** I've updated all the starters to include `tailwindcss` but in the `dependencies` section and not the `devDepedencies` section (as recommended by the [official docs](https://v3.tailwindcss.com/docs/installation)).
> 
> With this PR, we want to do the "right thing" and validate that the `tailwindcss` dep is in the `devDepedencies`. If we now include this validation that means it will give an error for _all projects_ and people will have to move the dep.
> 
> The issue for me is that people won't get any benefit by moving the dep other then to be semantically correct by including the dep in the right place. Wasp apps work correctly even with that dep in the `dependencies`.
> 
> **Given that we want to make Tailwind a user land feature soon**, and almost all Wasp apps are functional already, maybe we shouldn't annoy people with this temporary change. We could just [update the docs](https://wasp.sh/docs/project/css-frameworks#enabling-tailwind-step-by-step) to say they should add the dep.
> 
> **Alternatively**, we go ahead with this change, say this is a breaking change, include it in the migration guide and people will just do it. I'm not sure what is the size of annoyance.

I agree, good call! I would maybe still merge he good refactorings from this PR? What do you think, is it worth? If not, we can drop the whole PR of course.

PR #2472, opened 2025-01-28
Fixes 0.16.0 issues

infomiho:
Fixes two issues found during 0.16.0-rc2 QA:
- [x] Wasp AI template didn't have all the necessary deps
- [x] Using JS files to define operations didn't work due to a TS bug (linked in the comment above the fix)

### DIFF
--- a/waspc/src/Wasp/AI/GenerateNewProject/InitialFiles.hs
+++ b/waspc/src/Wasp/AI/GenerateNewProject/InitialFiles.hs
@@ -150,7 +150,10 @@ generatePackageJson newProjectDetails =
         "type": "module",
         "dependencies": {
           "wasp": "file:.wasp/out/sdk/wasp",
-          "react": "^18.2.0"
+          "react": "^18.2.0",
+          "react-dom": "^18.2.0",
+          "react-router-dom": "^6.26.2",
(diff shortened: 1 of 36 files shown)


waspc/src/Wasp/AI/GenerateNewProject/InitialFiles.hs
```diff
@@ -150,7 +150,10 @@ generatePackageJson newProjectDetails =
         "type": "module",
         "dependencies": {
           "wasp": "file:.wasp/out/sdk/wasp",
-          "react": "^18.2.0"
+          "react": "^18.2.0",
+          "react-dom": "^18.2.0",
+          "react-router-dom": "^6.26.2",
+          "tailwindcss": "^3.2.7"
```
sodic, line 156: Tailwind wasn't here before. Did you add it on purpose?

sodic (approved):
Great fix!

Can you please add a type "e2e test" in `waspc/todoApp` to prevent regressions (define an operation in JS and assert proper types are inferred though our custom type tests)?

Best to do it now while this is still fresh, but we can merge this first and then take care of it in a different PR.

sodic (approved):
LGTM

PR #2476, opened 2025-01-30
Add Remark plugin for Wasp version search & replace

infomiho:
Added a simple plugin to search & replace `{latestWaspVersion}` string in code blocks with the actual latest value based on `versions.json` file.

It uses the sem ver range format e.g. `^0.16.0` because this is what we needed. If we need to expand it, it's easy enough to update the plugin. 

I've tested it in dev, prod build and adding a new version e.g. 0.17.0 and it worked well.


sodic (approved):
Niceara

PR #2478, opened 2025-01-31
Adds headless tests for example apps

infomiho:
- [x] Adds headless tests for
  - [x] `examples/waspello`
  - [x] `examples/waspleau`
  - [x] `examples/websockets-realtime-voting`
  - [x] `examples/tutorials/TodoApp`
  - [x]  `examples/tutorials/TodoAppTs`


examples/hackathon-submissions/.gitignore
```diff
@@ -11,4 +11,8 @@ node_modules/
 !.env.*.example
 
 # Local Netlify folder
-.netlify
\ No newline at end of file
+.netlify
+
+# Headless tests
+!.env.*.headless
+test-results/
```
sodic, line 18: Hmm, I may have changed my mind on this grouping :sweat_smile: 

It makes sense to kepp all the headless stuff together. But `netlify` should definitely be pushed up.

examples/hackathon-submissions/headless-tests/playwright.config.ts
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// require('dotenv').config();
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+export default defineConfig({
```
sodic, line 12: Is this the same for all apps? Should we do something about it?

examples/hackathon-submissions/src/projects.ts
```diff
@@ -23,6 +23,7 @@ export const submitProject: SubmitProject<z.infer<typeof submitProjectInputSchem
         country,
         website,
         image,
+        approved: process.env.HEADLESS_TESTING === 'true',
```
sodic, line 26: What's happening here?

examples/streaming/headless-tests/tests/simple.spec.ts
sodic, line 1: Since this app is supposed to test streaming, can we somehow write the test to check for that (not just the final content)?

examples/tutorials/TodoApp/headless-tests/tests/helpers.ts
```diff
@@ -0,0 +1,43 @@
...
+  await page.goto("/signup");
+
+  await page.waitForSelector("text=Create a new account");
+
+  await page.locator("input[name='username']").fill(username);
+  await page.locator("input[type='password']").fill(password);
+  await page.locator("button").click();
+}
+
+export async function performLogin(
```
sodic, line 16: We could probably skip on testing the signup and the login in all apps. It's a lot of repetitive code that brings little benefit I'd say.

However, if and where we test for it (e.g., in `waspc/todoApp`), we should also test for unsuccessful logins and signups.

examples/tutorials/TodoApp/headless-tests/tests/simple.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+    await page.locator("input[type='submit']").click();
+    // Expect to see the task on the page
+    await expect(page.locator("body")).toContainText(randomTask);
+    // Check the task as done input[type="checkbox"]
+    await page.locator("input[type='checkbox']").click();
+    // Reload the page
+    await page.reload();
+    // Expect the task to be checked
+    await expect(page.locator("input[type='checkbox']")).toBeChecked();
+  });
```
sodic, line 54: The comments are probably gpt leftovers, we can remove them :)

examples/tutorials/TodoAppTs/headless-tests/tests/simple.spec.ts
sodic, line 1: For this one, we should probably test everything we tell them to do in the tutorial.

sodic (commented):
Excellent work!

But I can't decide whether we should remove most of the tests. They're pretty repetitive and unlikely to fail if `waspc/todoApp` works. Then again, we run them manually every time...

Another big repetition is the playwright config, so we should decide whether to do something about that too.

examples/tutorials/TodoApp/headless-tests/tests/helpers.ts
infomiho: Yep, testing login and signup is something we do along the way because we need to authenticate anyways to test the apps.

examples/tutorials/TodoApp/headless-tests/tests/simple.spec.ts
infomiho: Yep, probably me telling Copilot what to write (or I generated this with Playwright's extension, not sure).

examples/waspello/src/auth/SignupPage.tsx
```diff
@@ -7,18 +7,16 @@ import addWaspSourceHeader from '../common/addWaspSourceHeader'
```
infomiho, line 7: There was a type error (which I didn't notice) when using this code with the latest released Wasp, but this code works with `main` Wasp CLI. Cleaned up and converted to Typescript just in case.

examples/waspello/tsconfig.json
```diff
@@ -11,38 +11,18 @@
 {
```
infomiho, line 11: Migrated to latest `tsconfig.json`

examples/hackathon-submissions/headless-tests/playwright.config.ts
infomiho: It would be nice if we had some sort of Wasp `npm` package we can install and import these kind of common setups. But currently we don't have it, so let's keep it simple for now IMHO.

examples/hackathon-submissions/src/projects.ts
infomiho: This example app is now gone :)

examples/streaming/headless-tests/tests/simple.spec.ts
infomiho: I've migrated `streaming` to `todoApp` with a test that you mention, you'd be pleased 👍

examples/tutorials/TodoAppTs/headless-tests/tests/simple.spec.ts
infomiho: We maybe be able to do this once we have this #2666 

We could perform headless tests at various tutorial steps, until then, we are just testing the final app as we would do it manually to keep things simple.

examples/hackathon-submissions/.gitignore
infomiho: This app is now gone, but the headless ignore patterns are grouped together in all other apps as well 👍

examples/websockets-realtime-voting/src/pages/MainPage.tsx
```diff
@@ -1,13 +1,21 @@
-import { type ServerToClientPayload, useSocket, useSocketListener } from "wasp/client/webSocket";
```
infomiho, line 1: Most of this file is formatting, but I added some classes for easier headless testing

examples/waspello/src/auth/LoginPage.tsx
```diff
@@ -28,11 +28,12 @@ const LoginPage = (props) => {
       setPasswordFieldVal('')
 
       navigate('/')
-    } catch (err) {
-      if (err.statusCode == 422) {
+    } catch (err: any) {
+      // TODO: Update this to check against WaspHttpError https://github.com/wasp-lang/wasp/issues/2767
```
sodic, line 32: Can we mention this in the issue? We'll otherwise forget to update it.

examples/waspello/src/auth/SignupPage.tsx
```diff
@@ -30,8 +28,9 @@ const SignupPage = (props) => {
       setPasswordFieldVal('')
 
       navigate('/')
-    } catch (err) {
-      if (err.statusCode == 422) {
+    } catch (err: any) {
+      // TODO: Update this to check against WaspHttpError https://github.com/wasp-lang/wasp/issues/2767
```
sodic, line 32: Also please mention it in the issue.

sodic (commented):
Went through it again.

Everything looks ok functionally, but I'm still uneasy about the duplication:
- `playwright.config.ts` files seem to be identical, at least we intend them to be. But they're already drifting apart through small unintended differences (e.g., quotes and variables, check screenshot below).
- I believe the same applies for `simple.spec.ts` files.

Can we do anything about this? It doesn't need to be fancy. For example, we could symlink them all to the same file. 

image

infomiho:
> Can we do anything about this? It doesn't need to be fancy. For example, we could symlink them all to the same file.

I guess we are trying to solve the issue of sharing code between multiple different Wasp apps. I'm not sure symlinking is the best way to go (it feels OS specific to me, correct if I'm wrong). This is one argument for publishing the `wasp-app-runner` as an `npm` package which would enable use to centralise our common headless apps setup. 

Maybe we can locally install the `wasp-app-runner` package, but then I'd rename it into `wasp-headless-tests-utils` or smth like that, so it reflects it's wider scope.

sodic (approved):
Talked out of band, we'll handle the duplication separately. @infomiho Please link the issue when you create it.

Good to go!

infomiho:
Great, thanks! I'll merge this then and here's the issue for deduplication effort: https://github.com/wasp-lang/wasp/issues/2790

PR #1626, opened 2024-01-10
Implement Wasp SDK generation 

sodic:

PR #2486, opened 2025-02-05
Rename `ext-src` to `src` in SDK

sodic:


waspc/src/Wasp/Generator/SdkGenerator/Common.hs
```diff
@@ -73,9 +73,6 @@ relDirToRelFileP path = fromJust $ SP.parseRelFileP $ removeTrailingSlash $ SP.f
 makeSdkImportPath :: Path Posix (Rel SdkRootDir) File' -> Path Posix (Rel s) File'
 makeSdkImportPath path = [reldirP|wasp|] </> path
 
-extCodeDirInSdkRootDir :: Path' (Rel SdkRootDir) Dir'
-extCodeDirInSdkRootDir = [reldir|ext-src|]
```
sodic, line 77: We accidentally had two constants describing the same thing, so I removed this one.

waspc/data/Generator/templates/react-app/tsconfig.node.json
```diff
@@ -8,7 +8,6 @@
     "allowSyntheticDefaultImports": true,
   },
   "include": [
-    "vite.config.ts",
-    "./src/ext-src/vite.config.ts"
```
infomiho, line 12: Why did you drop this?

sodic: `./src/ext-src/vite.config.ts` isn't a thing since #1626, but we forgot to remove it from here.

These days, we're importing the user's config directly:
```ts
import customViteConfig from '../../../vite.config'
const _waspUserProvidedConfig = customViteConfig
```
Hm, should we perhaps add `../../../vite.config` to the `include` array? Vite doesn't type-check anything anyway.

infomiho: I've put `@ts-ignore` about the import of user's Vite config because it's outside of the root (TS complained) - maybe that has changed? If not, then the include might not be necessary since it's ignored anyways?

infomiho (approved):
Tested it locally, built with Docker. All good.

PR #2457, opened 2025-01-17
Make Wasp work with tsconfig path aliases

sodic:

PR #2493, opened 2025-02-07
Bundle the SDK together with the server

### LINKED ISSUE #2446: Improve our Rollup config for server bundle to include `wasp/` modules
We bundle our server app to give users more freedom with how they write their imports paths i.e. they can use the `.js` suffix or can leave it out.

One thing we noticed is that in our bundle, the `wasp/` package is not bundled and we have to copy over the Wasp SDK when building the server Docker image: https://github.com/wasp-lang/wasp/pull/2362#discussion_r1893275711

We are missing the`@rollup/plugin-node-resolve` plugin which enables Rollup to bundle packages from `node_modules`.

Here's a working `rollup.config.js` file that bundles the Wasp SDK:

```js
import esbuild from 'rollup-plugin-esbuild'
import nodeResolve from "@rollup/plugin-node-resolve";

export default [
  createBundle('src/server.ts', 'bundle/server.js'),
  createBundle('src/dbSeed.ts', 'bundle/dbSeed.js'),
]

/**
 * @param {string} inputFilePath
 * @param {string} outputFilePath
 * @returns {import('rollup').RollupOptions}
 */
function createBundle(inputFilePath, outputFilePath) {
  return {
    input: inputFilePath,
    output: {
      file: outputFilePath,
      format: 'es',
      sourcemap: true,
    },
    plugins: [
      esbuild({
        target: 'esnext',
      }),
      // Using `modulesOnly` because I found it here: https://stackoverflow.com/questions/45707711/does-rollup-bundle-node-modules-into-bundle-js
      nodeResolve({
        modulesOnly: true
      })
    ],
    external: (id) => {
      // Anything that's not a relative or absolute import or a wasp/ module - should not be bundled e.g. Express won't be bundled
      return !id.startsWith('.') && !id.startsWith('/') && !id.startsWith('wasp/')
    },
    treeshake: 'smallest'
  }
}
```

### LINKED ISSUE #2492: Relative imports in user code combined with imports from `wasp/server/operations` break `wasp start`
This is a problem we've had for a while but haven't properly documented. Users have also been known to hit it (example below).

## Steps to reproduce
Create a project with Wasp 0.16 and make the following edits:

<details><summary>./src/main.wasp</summary>

```c
app nuno {
  wasp: {
    version: "^0.16.0"
  },
  title: "nuno"
}

route RootRoute { path: "/", to: MainPage }
page MainPage {
  component: import { MainPage } from "@src/MainPage"
}

query getAllTasks {
  fn: import { getAllTasks } from "@src/error/queries"
}
```

</details>
<details><summary>./src/queries.js</summary>

```js
import { sayHi } from './util'
import { getAllTasks as getAllTasksWasp } from 'wasp/server/operations'

export const getAllTasks = () => {
  console.log(sayHi, getAllTasksWasp)
  return []
}
```

</details>

</details>
<details><summary>./src/util.js</summary>

```js
export function sayHi() {
  console.log('`Hello!')
}
```

</details>

Run `wasp start` and you should see the error:
```
[ Server ] > server@0.0.0 start
[ Server ] > node --enable-source-maps -r dotenv/config bundle/server.js
[ Server ]
[ Server!] node:internal/errors:496
[ Server!]     ErrorCaptureStackTrace(err);
[ Server!]     ^
[ Server!]
[ Server!] Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/tmp/nuno/.wasp/out/sdk/wasp/dist/ext-src/error/util' imported from /tmp/nuno/.wasp/out/sdk/wasp/dist/ext-src/error/queries.js
[ Server!]     at __node_internal_captureLargerStackTrace (node:internal/errors:496:5)
[ Server!]     at new NodeError (node:internal/errors:405:5)
[ Server!]     at finalizeResolution (node:internal/modules/esm/resolve:327:11)
[ Server!]     at moduleResolve (node:internal/modules/esm/resolve:980:10)
[ Server!]     at defaultResolve (node:internal/modules/esm/resolve:1206:11)
[ Server!]     at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:404:12)
[ Server!]     at ModuleLoader.resolve (node:internal/modules/esm/loader:373:25)
[ Server!]     at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:250:38)
[ Server!]     at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:76:39)
[ Server!]     at link (node:internal/modules/esm/module_job:75:36) {
[ Server!]   url: 'file:///tmp/nuno/.wasp/out/sdk/wasp/dist/ext-src/error/util',
[ Server!]   code: 'ERR_MODULE_NOT_FOUND'
[ Server!] }
[ Server!]
```

## Explanation

Wasp bundles the server with Rollup, meaning that **server code (and the user code it imports, e.g., operations) can use extensionless relative imports** (e.g., `import { sayHi } from './util'`).

Wasp doesn't bundle the SDK (the `wasp` package), meaning that code in the **SDK cannot use extensionless relative imports**.

Since Wasp copies all user code into the SDK (more precisely, into `wasp/ext-src`), it should mean that user code cannot use extensionless imports after all. But, as long as nothing references the offending code, no one will notice and everything will work.

In practice, this means that user code _can_ use extensionless relative imports as long as some other part of user code doesn't indirectly import the offending code from `wasp/ext-src` (e.g., by importing `wasp/server/operations`).

You could be working on your app for weeks, happily using extensionless relative imports. And then, one day, you import something from `wasp/server/operations`, which breaks your app in 100 places that have nothing to do with the change you just made (this is what happened to a [user with OpenSaas](https://discord.com/channels/686873244791210014/1337075786846437496/1337418335746396170))

This behavior was also the root cause of this bug (albeit with slight differences): https://github.com/wasp-lang/wasp/issues/2096#issuecomment-2263566708

Image

### Why isn't this caught during SDK build
At the time of writing, the SDK is built with TypeScript.

TypeScript can resolve extensionless files and doesn't alter import paths when outputting JS code, meaning that the error can only be caught in runtime.

## Solution
The long term solution is figuring out how precisely how we want to treat and build the SDK.
The short-term solution is including the `wasp` package in the server's bundle  (as we do for the client). @infomiho talks about it here: #2446

sodic:
Fixes #2492 (short term fix).
Fixes #2446

Inspiration: #2457 

image


waspc/data/Generator/templates/Dockerfile
```diff
@@ -55,9 +55,6 @@ WORKDIR /app
 # Copying the top level 'node_modules' because it contains the Prisma packages
 # necessary for migrating the database.
 COPY --from=server-builder /app/node_modules ./node_modules
-# Copying the SDK because the server bundle doesn't bundle the SDK
-# and references the 'wasp' package.
-COPY --from=server-builder /app/.wasp/out/sdk .wasp/out/sdk
```
sodic, line 60: @infomiho I removed this, and it works, but I think I might even want to keep the old code

I don't like it either way. What do you think?

infomiho: Let's make it as precise as possible IMHO, if we don't need, don't include it in the context.

infomiho (approved):
Tested out locally, works okay. I'm okay with whichever decision make with regard to copying the SDK in `node_modules` or not.

waspc/data/Generator/templates/Dockerfile
sodic: Removed it in the end. It will help us catch bundle errors on time, instead of risking things working accidentally.

PR #2505, opened 2025-02-12
Activate project references in the client

sodic:
Turns out I forgot to activate `tsc --build` (TS build with project references) in the the generated web app. Once I did, it turned out we really did need that `ts-ignore` :sweat_smile:


Martinsos:
This one is close to merging, right? Just needs updated e2e tests?

PR #2507, opened 2025-02-12
Add default value for `NODE_ENV`

sodic:
Fixes this:

image

@infomiho Couldn't find a better way to set a default value. Is this ok?


waspc/data/Generator/templates/sdk/wasp/server/env.ts
sodic, line 1: @infomiho I tried to define list of all necessary env vars and somehow ensure (in the type system) that both `serverProdSchema` and `serverDevSchema`:
1. Define constraints for all vaules (`NODE_ENV`, `WASP_SERVER`...).
2. Can only make the existing constraints stricter.
3. Define no other constraints (to prevent typos).

I was hoping Zod supports something that, but seemingly none of its methods (e.g., `extend`, `merge`) fit my use case.

I also tried to use a type definition and `satisfies`, but couldn't make it work. GPTs kept suggesting complicated custom generic functions, so I gave up (don't think it's worth it)

It would work something like this (I'm making up the `narrow` method):
```typescript
// This doesn't have to be Zod, it could just be a normal type, perhaps that's even better.
// (but I couldn't make that work either).
const envRequirements = z.object({
  NODE_ENV: z.enum(['development', 'production']),
  WASP_SERVER_URL: serverUrlSchema,
  WASP_WEB_CLIENT_URL: clientUrlSchema,
  JWT_SECRET: z.string(),
});

const serverDevSchema = envRequirements.narrow({
  NODE_ENV: z.literal('development'),
  WASP_SERVER_URL: serverUrlSchema
    .default('http://localhost:3001'),
  WASP_WEB_CLIENT_URL: clientUrlSchema
    .default('http://localhost:3000/'),
  JWT_SECRET: jwtTokenSchema
    .default('DEVJWTSECRET'),
})

const serverProdSchema = envRequirements.narrow({
  NODE_ENV: z.literal('production'),
  WSP_SERVER_URL: serverUrlSchema, // error: typo
  WASP_SERVER_URL: z.string(), // error: widens constraints
  // Error for missing JWT_SECRET, no constraints defined
})
```
Did I miss something, can Zod do this?
If not, I'll keep it as is. A custom generic function is too much IMO.

infomiho: Too bad we couldn't do it like you imagined, I see [unions in Zod 4 will get some love](https://github.com/colinhacks/zod/issues/3407), but it's unclear when it will go out - that might help us in the future.

The solution you implemented is good enough, I don't see anything better you could have done.

infomiho (approved):
LGTM, tested locally 👍

Issue #1864, opened 2024-03-07
Investigate how much we can mix server and client code


sodic:
As of Wasp 0.12.0-ish, we can mix client and server code in the same file.

For example, the `main.wasp` can this:
```c
route SomeRoute { path: "/something", to: SomePage }
page SomePage {
  component: import { SomeComponent } from "@src/api"
}

// ...

api fooBar {
  fn: import { fooBar } from "@src/api",
  httpRoute: (GET, "/foo/bar")
} 
```

The `src.tsx` file can contain this:
```tsx
import type { FooBar } from 'wasp/server/api'

export const fooBar: FooBar = (req, res, context) => {
  res.json({ msg: 'Hello friend' })
}

export const SomeComponent = () => {
  return <div>Some gmali</div>
}
```
And everything works (even when built and deployed).

Gotchas:
 - Hot reloading on the server doesn't work (i.e,. the server ignores changes made to the file).
   This probably happens because we're not watching `.tsx` files.
    - Related issue: #1844)
 - It's even possible to use React on the server:
    ```ts
    import type { FooBar } from 'wasp/server/api'
    import React from 'react' // necessary

    export const fooBar: FooBar = (req, res, context) => {
      console.log(SomeComponent())
      res.json({ msg: 'Hello friend' })
    }

   export const SomeComponent = () => {
     return <div>Some gmali</div>
   }
   ```
   If you forget to import react, you get a runtime error (neither the IDE or Wasp complain in compile time). We should investigate this as well.

Issue #1844, opened 2024-03-01
Investigate how our file watching works


sodic:
With Wasp 0.12.0 and the restructuring (#1688), we've removed the constraint of separating server files from client files.

We should investigate the impact this had on our watching procedures and hot reloads.

Goals:
- The server rebuilds only on changes that affect the server framework code.
- The SDK rebuilds only on changes that affect the SDK code.
- The web app rebuilds only on changes that affect the web-app framework code.

Questions, on each change, what does:
- Nodemon watch?
- TSC compile (in SDK)?
- Rollup rebuild?
- Vite watch?
- Vite rebuild?

Relevant `todo` in the code: https://github.com/wasp-lang/wasp/blob/f2721e767931efdfc0731a877239ba54eafeba9e/waspc/src/Wasp/Analyzer/Evaluator/Evaluation/TypedExpr/Combinators.hs#L176

Martinsos:
We also seem to be having an issue with too many files being watched with the recent projects, check out this issue: https://github.com/wasp-lang/wasp/issues/1919 .

This might be because we have some files being watched redundantly.

There are commands that track how many files are watched by which process, so we should se that to observe what is happening.

napon:
I'm running into this issue in my project. Let me know if you need me to share any logs to help debug the issue further!

sodic:
Hi @napon, thanks for reporting!

Just to ensure I got it right, you're talking about "too many files being watched" issue?

sodic:
Hmm, I gotta admit I'm not familiar enough with the tools used to debug these kinds of issues, so I wouldn't know what to ask for.

@napon Any chance you'd be ready to share your project with me and let me play around with debugging it locally?

infomiho:
Related issue: https://github.com/wasp-lang/wasp/issues/1843

infomiho:
I'd like to tackle the `nodemon` part of this issue soon since it's causing server crashes:
- for example, add a new `query` + add implementation **at the same time** (happens when using LLMs)
- the SDK starts rebuilding
- the server's `nodemon` watcher also sees changes, tries rebuilding
- the server crashes
- the SDK finishes building
- the server in now stuck because `nodemon` doesn't detect that SDK is _now_ ready

We'd ideally orchestrate server watching + rebuilds + restarts ourselves _somehow_:
- either we drop `nodemon` and implement something in Haskell
- or we figure out a way to configure `nodemon` or something similar to work as expected.

PR #2508, opened 2025-02-13
Handle relative path from web app to project root in a single place

sodic:
There are still two places of duplication left:
- `ServerGenerator/JsImport.hs`
- `WebAppGenerator/JsImport.hs`

But those two are too complicated to handle now (it would require rearraning many imports/modules) . I have bigger plans that will also take care of those two, so I'm leaving them alone for now.


waspc/data/Generator/templates/react-app/vite/detectServerImports.ts
```diff
@@ -49,7 +47,6 @@ function parsePathToUserCode(
 // e.g. /Users/{username}/dev/wasp/waspc/examples/todoApp
 // which contains machine specific info like the username which is different in the CI and locally.
 // This breaks our e2e tests in the CI because the path is different.
-function getWaspProjectDirAbsPathFromCwd(): string {
-  const webAppDirAbsPath = process.cwd()
-  return path.join(webAppDirAbsPath, '{= waspProjectDirFromWebAppDir =}')
+function getWaspProjectDirAbsPathWhileInWebAppDir(): string {
```
sodic, line 50: I changed this function's nam to clearly communicate its assumptions.

waspc/data/Generator/templates/react-app/vite/detectServerImports.ts
```diff
@@ -49,7 +47,6 @@ function parsePathToUserCode(
...
+  return path.resolve(resolveProjectPath('./'))
```
sodic, line 51: @infomiho Double check if it's ok that I import this function here.

I tested it out and it works.

waspc/data/Generator/templates/server/nodemon.json
```diff
@@ -1,12 +1,13 @@
...
   },
   "watch": [
     "src/",
-    "../../../src/",
+    "{= relativeUserSrcDirPath =}",
     ".env"
   ],
-  "comment-filip": "We now have to watch ../../../src/ because we're importing client files directly",
+  "comment-filip": "We now have to watch {= relativeUserSrcDirPath =} because we're importing client files directly",
   "ext": "ts,mts,js,mjs,json"
```
infomiho, line 12: @Zeko369 mentioned that the server doesn't restart on TSX file changes, should we extend this list? Does it make sense for server to restart on TSX changes since we are not rendering anything on the server?

waspc/data/Generator/templates/react-app/vite/detectServerImports.ts
infomiho: That helper has been used in the user's `vite.config.ts` because it was used from `web-app`. This serves the exact same purpose - it's a good use!

waspc/data/Generator/templates/react-app/vite/detectServerImports.ts
```diff
@@ -37,7 +35,7 @@ function parsePathToUserCode(
   importerPath: string
 ): RelativePathToUserCode | null {
   const importerPathRelativeToWaspProjectDir = path.relative(
-    waspProjectDirAbsPath,
+    getWaspProjectDirAbsPathWhileInWebAppDir(),
     importerPath
   )
   return importerPathRelativeToWaspProjectDir.startsWith('src/')
```
infomiho, line 41: Do we maybe want to template the `src` dir? I seem to have forgotten to use the `srcDirInWaspProjectDir` variable.

Martinsos:
What do we need to do to merge this one?

waspc/data/Generator/templates/server/nodemon.json
sodic: Already covered by #1844 and #1864

PR #2511, opened 2025-02-13
Inline Google Fonts to reduce Network requests

infomiho:
Optimising the web a bit more: https://johnfraney.ca/blog/improve-page-speed-google-fonts/


Martinsos (commented):
LGTM

PR #2515, opened 2025-02-17
Fixes MDX components to have correct links in versioned docs

infomiho:
- `AuthMethodsGrid` wasn't versioned (e.g. we would show latest list in 0.12.0 version)
- `SocialMethodsGrid` didn't use relative links
- `DeploymentMethodsGrid` didn't use relative links


sodic:
Nice work.

How did you catch this? Any chance of avoiding it in the future?

infomiho:
I saw that the Algolia crawler couldn't index some URLs so I investigated. The root cause was writing absolute URLs and not versioning one of the components. We could add a note in the README for future work not to do this.

sodic:
> We could add a note in the README for future work not to do this.

Any way to enforce it while building the docs?

infomiho:
> Any way to enforce it while building the docs?

I haven't researched this but it could be possible that we can write a plugin for Docusaurus.

PR #2534, opened 2025-03-05
Adds Wasp app runner for headless tests

infomiho:
Wasp app runner is:
> A robust script to run your Wasp application with PostgreSQL, including automated database setup and migrations.

This will help with running headless tests on our example apps.


Martinsos:
How are we doing here, need somebody to check this out @infomiho , or is it still cooking?

wasp-app-runner/src/index.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  return processManager.spawnWithLog({
+    name: "install-wasp-cli",
+    cmd: "cabal",
+    args: ["install", "--overwrite-policy=always"],
+    cwd: getWaspcDirAbsPath(),
+  });
+}
+
+async function main() {
```
sodic, line 22: Hmm, this program's flow is a little difficult to follow.

Who parses the command line options?
`main` relies on `options`, which relies on `getOptions`, which relies on `argv`, which is just some local constant. The control flow implicitly extends from functions' bodies onto global constants instead of going top-down.

Here's what I find intutitive.

I'd expect `main` to call `getOptions`. It can already do this since `options` aren't exported (which is good).

After `main` gets all the CLI options, they should no longer travel together - nothing really connects them other than being configurable by the CLI (which is finished at that point). Functions `executeWithDb` and `setupEnvFiles` shouldn't get or know about all the CLI options anyway.

wasp-app-runner/src/index.ts
```diff
@@ -0,0 +1,61 @@
...
+import { getOptions } from "./cli.js";
+import { checkDependencies } from "./dependencies.js";
+import { processManager } from "./process.js";
+import { setupEnvFiles } from "./env.js";
+import { getWaspcDirAbsPath } from "./path.js";
+import { executeWithDb } from "./db/index.js";
+
+const options = getOptions();
+
+async function installWaspCli() {
```
sodic, line 11: I recommend moving the most important stuff (`main`) to the top of the file.

wasp-app-runner/src/logging.ts
```diff
@@ -0,0 +1,20 @@
+type LogType = "error" | "warn" | "info" | "success" | "debug";
+
+const logTypeToColor: Record<LogType, number> = {
+  error: 31,
+  warn: 33,
+  info: 36,
+  success: 32,
+  debug: 90,
+};
```
sodic, line 9: I'd make this local to the function.

wasp-app-runner/src/path.ts
```diff
@@ -0,0 +1,7 @@
+import { join } from "path";
+
+// Gets the absolute path to the waspc directory __from the current file__.
+export function getWaspcDirAbsPath() {
+  const currentDirAbsPath = new URL(".", import.meta.url).pathname;
+  return join(currentDirAbsPath, "../../waspc");
+}
```
sodic, line 7: Nice trick!

wasp-app-runner/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+// Based on https://www.totaltypescript.com/tsconfig-cheat-sheet
```
sodic, line 1: I was just about to suggest this!

examples/todo-typescript/src/Main.css
```diff
@@ -70,4 +71,4 @@ li {
   flex-direction: row;
   justify-content: space-between;
   align-items: center;
-}
+}
```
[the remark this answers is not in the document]
infomiho: VS Code didn't insert the new line by default for me, I've now configured it to do so for CSS files 👍

wasp-app-runner/src/process.ts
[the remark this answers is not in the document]
infomiho: Looked into some libs that deal with spawning processes, they don't abstract away things _that much_ so I vote we keep things as basic as possible.

wasp-app-runner/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+      pathToApp: argv.appPath,
+    });
+
+    await executeWithDb(
+      {
+        appName,
+        dbType,
+        pathToApp: argv.appPath,
+      },
+      async ({ extraEnv }) => {
```
sodic, line 52: This flow seems a little overly complex: This function sends an arrow function to another function, which then calls the received arrow function with an argument, and the arrow function finally passes those arguments to the function that does the actual work.

I understand what's happening and why we did it (there might be some extra work to do when calling wasp with a specific database), but I think we can make it a little simpler:

```ts
// This function does all the setup currently done in executeDb
// but without calling the app.
const extraEnv = setUpDatabase({ appName, pathToApp, dbType })

await migrateDb({
    waspCliCmd,
    pathToApp,
    extraEnv,
});

await startApp({
    waspCliCmd,
    pathToApp,
    extraEnv,
});

```
We should then also change the type `RunAppWithDbFn` to `setUpDatabase`:

```ts
export type SetUpDatabase = (
  options: { appName: string; pathToApp: string  },
) => Promise<EnvVars>;
```

I don't think we lose anything (db setup functions can still do side effects) and the flow is much simpler (we're not dragging callbacks all the way down to the database initialization functions). How does that look to you, did I miss something?

wasp-app-runner/src/waspCli.ts
```diff
@@ -0,0 +1,121 @@
...
+}
+
+export function startApp({
+  waspCliCmd,
+  pathToApp,
+  extraEnv,
+}: {
+  waspCliCmd: string;
+  pathToApp: string;
+  extraEnv: Record<string, string>;
```
sodic, line 33: I recommend using for this type since it's mentioned so often. I'd also go for an index signature to name the key:
```ts
type EnvVars = { [envVarName: string]: string }
```

wasp-app-runner/src/waspCli.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  if (!appNameMatch) {
+    log("get-app-info", "error", `Failed to get app name`);
+    process.exit(1);
+  }
+
+  if (!dbTypeMatch) {
+    log("get-app-info", "error", `Failed to get database type`);
+    process.exit(1);
+  }
```
sodic, line 84: Can we check for `null` explicitly?

I am generally not a fan of coercions in boolean expressions. I wouldn't normally mention it, but in this case, we're dealing with regexes which can match empty strings.

Of course, if this happens, something else is probably wrong. Still, it's best if our conditions communicate exactly what we want to say. And in this case, I believe we want to say "exit if you haven't found a match".

sodic (commented):
Nice fixes, but I still have some suggestions :)

wasp-app-runner/README.md
```diff
@@ -0,0 +1,52 @@
+# Wasp Application Runner
+
+A robust script to run your Wasp application with PostgreSQL, including automated database setup and migrations.
```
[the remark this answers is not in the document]
infomiho: image

wasp-app-runner/src/db/postgres.ts
```diff
@@ -0,0 +1,194 @@
...
+  const DATABASE_URL = await ensurePostgresContainer({ appName, pathToApp });
+
+  log("postgres", "info", `Using DATABASE_URL: ${DATABASE_URL}`);
+
+  return runApp({
+    extraEnv: { DATABASE_URL },
+  });
+};
+
+async function ensurePostgresContainer({
```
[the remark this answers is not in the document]
infomiho: We start the Postgres container in attached mode (not detached), this means that the DB containers gets killed when the user kills the `wasp-app-runner`. 

I've adjusted the code based on your comment 👍

waspc/examples/todoApp/.env.client.headless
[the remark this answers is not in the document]
infomiho: We don't really need these values for headless tests, I'll remove the file 👍

wasp-app-runner/src/process.ts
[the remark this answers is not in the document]
infomiho: The code is `spawn`ing new processes and collecting them in case the main process exists, it makes sure to kill all the children processes. 

I've now extracted the process management logic into one class and the spawn wrappers are two separate functions that are not complex. I hope this is better for you now.

wasp-app-runner/src/dependencies.ts
```diff
@@ -0,0 +1,26 @@
...
+        `Required command '${cmd}' not found. Please install it.`
+      );
+      process.exit(1);
+    }
+  }
+}
+
+function commandExists(command: string): Promise<boolean> {
+  return new Promise((resolve) => {
+    exec(`command -v ${command}`, (error) => {
```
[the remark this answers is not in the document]
infomiho: I'm not sure, it just failed for some reason. I think it's okay to keep using `exec` for this. It's our own code without any user input and it's used internally.

wasp-app-runner/src/env.ts
```diff
@@ -0,0 +1,22 @@
...
+import { copyFileSync, existsSync } from "fs";
+import { log } from "./logging.js";
+
+export function setupEnvFiles({ pathToApp }: { pathToApp: string }) {
+  const envFiles = [
+    { headless: ".env.client.headless", target: ".env.client" },
+    { headless: ".env.server.headless", target: ".env.server" },
+  ];
+
+  for (const { headless, target } of envFiles) {
```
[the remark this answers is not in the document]
infomiho: We have a few options IMHO:
1. Backing up the old files before overwriting it (`.env.server.backup` or smth like that) so the user can restore the old values
2. Asking before overwriting it, so the user can a chance to manually save the values if needed
3. Keeping it as-is - these are example apps for which we should have the env vars in https://www.dotenv.org/ anyways
4. Not doing this if not specified with a CLI flag e.g. `--copy-headless-env-files`

What would you pick?

wasp-app-runner/src/process.ts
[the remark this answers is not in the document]
sodic: It's probably me, so if it's clear to you, we can merge it.

wasp-app-runner/src/process.ts
```diff
@@ -0,0 +1,128 @@
...
+    if (index !== -1) {
+      this.children.splice(index, 1);
+    }
+  }
+
+  private cleanExit(reason: string) {
+    log("shutdown", "info", `Received ${reason}. Cleaning up...`);
+    this.children.forEach((child) => {
+      if (!child.killed) {
+        child.kill();
```
sodic, line 36: Talked in person with @infomiho. We sometimes we don't sucessfully kill all the children, causing the parent process to exit without removing the container)?

Let's log something when `child.kill()` returns false and see whether that causes the problem.

wasp-app-runner/src/dependencies.ts
```diff
@@ -0,0 +1,38 @@
...
+        "setup",
+        "error",
+        `Required command '${cmd}' not found. Please install it.`
+      );
+      process.exit(1);
+    }
+  }
+
+  if (buildWaspCli) {
+    await installWaspCli();
```
sodic, line 28: We talked in person and agreed we'll remove the installation details from this project (Unix philosophy and all that).

wasp-app-runner/src/types.ts
sodic, line 1: I'm not a big fan of having `types` files.

I'd call it `utils.ts` or something (it's still frowned upon, but it at least describes what the stuff inside is for).

waspc/examples/todoApp/headless-tests/playwright.config.ts
```diff
@@ -44,11 +44,11 @@ export default defineConfig({
...
   webServer: {
-    command: 'npm run headless:start',
-
+    command: 'wasp-app-runner --path-to-app=../',
     // Wait for the backend to start
     url: 'http://localhost:3001',
     reuseExistingServer: !process.env.CI,
-    timeout: 120 * 1000,
+    timeout: 180 * 1000,
+    gracefulShutdown: { signal: 'SIGTERM', timeout: 500 },
```
sodic, line 52: Maybe SIGINT is even better? Or maybe it wouldn't work?

sodic (commented):
Left some comments, mostly cosmetics (other than what we discussed in person).

waspc/examples/todoApp/headless-tests/playwright.config.ts
infomiho: > SIGINT is a signal used to request a graceful termination of a process, typically initiated by the user. SIGTERM is a signal used to request a process to terminate gracefully, which can be sent by other processes or the system itself.

Based on [this](https://eitca.org/cybersecurity/eitc-is-lsa-linux-system-administration/linux-processes/process-signals/examination-review-process-signals/explain-the-difference-between-sigint-sigterm-and-sigkill-signals-in-linux/), I'd keep `SIGTERM`

wasp-app-runner/src/types.ts
infomiho: I like having `types.ts` in various folders e.g. `/types.ts` and `/db/types.ts` and they are "useful types" used at that level of the app and deeper. I don't want to use `utils.ts` because that sounds to be that it might contain some runtime code.

wasp-app-runner/src/process.ts
infomiho: Killing the process that started the container with `docker run` does not necessarily kill the Docker container. It might send a signal, it might not, it depends.

More info here: https://stackoverflow.com/a/50034061

Alternatively, we could upgrade our system in couple of ways:
- run `docker stop` command to stop the PostgreSQL container when cleaning up
- use Docker API directly from Node.js via https://github.com/apocas/dockerode

I'm on the fence about what to do here, I'd keep this _as simple as possible_ since it's our internal tool and the simplest option for me would be to just accept that fact that sometimes we'd have to manually clean up the database container. We can evolve the approach it start getting in our way. What do you think?

infomiho:
@sodic I've decided to remove the env files management from the `wasp-app-runner` to make the tool even simpler.

wasp-app-runner/src/process.ts
sodic: We can leave it like this. Please just add this explanation to the code (and possibly link to the github comment).

wasp-app-runner/src/env.ts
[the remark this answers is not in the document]
sodic: I see you removed env vars from the runner. Makes sense, I'm resolving.

wasp-app-runner/src/db/postgres.ts
```diff
@@ -0,0 +1,193 @@
...
+    // If we awaited here, we would block the main thread indefinitely.
+    .then(({ exitCode, stderrData }) => {
+      if (exitCode !== 0) {
+        log("postgres", "error", stderrData);
+        const extraInfo = getExtraInfoOnPostgresStartError({
+          originalErrorText: stderrData,
+          containerName,
+          port,
+        });
+        if (extraInfo) {
```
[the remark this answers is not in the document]
infomiho: Yep, you are right, we should be as strict as possible with our code 👍

wasp-app-runner/src/types.ts
infomiho: I understand the domain driven file structure organisation and I quite like it. Again, the `types.ts` file for me has this special aura that it contains types related to the folder. I'll just keep it as-is for now since the files are really small.

Issue #2514, opened 2025-02-15
`sanitizeAndSerializeProviderData` is missing `await` in seeding docs


infomiho:
Here: https://wasp.sh/docs/data-model/backends#writing-a-seed-function

Maybe it wouldn't be bad to rerun the seeding example code to verify it's up to date with latest Wasp version.

Pay attention to our writing guide: https://github.com/wasp-lang/wasp/blob/main/web/docs/writingguide.md

seppulcro:
Here's a more detailed example that I landed on while testing out wasp and going through the docks:

```
import { sanitizeAndSerializeProviderData } from "wasp/server/auth";
import { PrismaClient, Role, User } from "@prisma/client";

export const devSeedSimple = async (prisma: PrismaClient) => {
  await createUser(prisma, {
    username: "demo",
    password: "demo12345",
    role: Role.ADMIN,
  });
};

type CreateUserInput = {
  username: string;
  password: string;
} & Pick<User, "role">;

async function createUser(
  prisma: PrismaClient,
  data: CreateUserInput
): Promise<User> {
  const newUser = await prisma.user.create({
    data: {
      auth: {
        create: {
          identities: {
            create: {
              providerName: "username",
              providerUserId: data.username,
              providerData: await sanitizeAndSerializeProviderData({
                hashedPassword: data.password,
              }),
            },
          },
        },
      },
      role: data.role,
    },
  });

  return newUser;
}
```

This is working and intended and:
- They can use types/enums from prisma schema
- Shows users they can use type intersection if their model extends basic functionallity

FranjoMindek:
Fixed with #2535

Issue #1071, opened 2023-03-21
Improve docs on the session prop


sodic:
Wasp automatically passes the `user` prop to authenticated components (i.e., pages with `auth: true`).

We show this functionality in the todoApp tutorial, but our documentation only mentions the `useAuth` hook (which is a suboptimal way of getting the user because the hook can't know whether it's being called in an authenticated component).

Our documentation should clearly instruct our users to use the `user` prop whenever possible and only resort to the `useAuth` hook when they want to do something user-related in a non-authenticated component (e.g., headers).

Pay attention to our writing guide: https://github.com/wasp-lang/wasp/blob/main/web/docs/writingguide.md

PR #2535, opened 2025-03-06
Update outdated docs, improve naming, prefer user props over auth hook in docs

FranjoMindek:
- Adds `await` for `sanitizeAndSerializeProviderData` calls
- Renames `data-model/backends` to `data-model/databases` 
- Docs guide users to use `user` props rather than `useAuth()`, also reflected on examples

Related issues: https://github.com/wasp-lang/wasp/issues/2514, https://github.com/wasp-lang/wasp/issues/1071

### DIFF
--- a/web/docs/advanced/jobs.md
+++ b/web/docs/advanced/jobs.md
@@ -233,7 +233,7 @@ The Job declaration has the following fields:
   :::note Job executors
   Our jobs need job executors to handle the _scheduling, monitoring, and execution_.
 
-  `PgBoss` is currently our only job executor, and is recommended for low-volume production use cases. It requires that your database provider is set to `"postgresql"` in your `schema.prisma` file. Read more about setting the provider [here](../data-model/backends.md#postgresql).
+  `PgBoss` is currently our only job executor, and is recommended for low-volume production use cases. It requires that your database provider is set to `"postgresql"` in your `schema.prisma` file. Read more about setting the provider [here](../data-model/databases.md#postgresql).
   :::
 
   We have selected [pg-boss](https://github.com/timgit/pg-boss/) as our first job executor to handle the low-volume, basic job queue workloads many web applications have. By using PostgreSQL (and [SKIP LOCKED](https://www.2ndquadrant.com/en/blog/what-is-select-skip-locked-for-in-postgresql-9-5/)) as its storage and synchronization mechanism, it allows us to provide many job queue pros without any additional infrastructure or complex management.
--- a/web/docs/auth/overview.md
+++ b/web/docs/auth/overview.md
@@ -184,7 +184,7 @@ There are two ways to access the `user` object on the client:
 
 #### Using the `user` prop
 
-If the page's declaration sets `authRequired` to `true`, the page's React component receives the `user` object as a prop:
+The preferred way to fetch user data on authenticated pages is by using the `user` prop. If the page's declaration sets `authRequired` to `true`, the page's React component receives the `user` object as a prop:
 
 <Tabs groupId="js-ts">
 <TabItem value="js" label="JavaScript">
@@ -313,7 +313,7 @@ export function Main() {
 </Tabs>
 
 :::tip
-Since the `user` prop is only available in a page's React component: use the `user` prop in the page's React component and the `useAuth` hook in any other React component.
+While the `user` prop is only available in an authenticated page's React component, it is the preferred option because we can be certain that the user exists on that page. On the other hand, `useAuth` can be used when the user might not exist, making it better suited for non-authenticated components.
 :::
 
 ### On the server
--- a/web/docs/auth/social-auth/overview.md
+++ b/web/docs/auth/social-auth/overview.md
@@ -215,7 +215,7 @@ export const userSignupFields = defineUserSignupFields({
 
 #### 3. Showing the Correct State on the Client
 
-You can query the user's `isSignupComplete` flag on the client with the [`useAuth()`](../../auth/overview) hook.
+You can check the `isSignupComplete` flag on the `user` object.
 Depending on the flag's value, you can redirect users to the appropriate signup step.
 
 For example:
@@ -227,12 +227,9 @@ For example:
 <TabItem value="js" label="JavaScript">
 
 ```jsx title=src/HomePage.jsx
-import { useAuth } from 'wasp/client/auth'
 import { Navigate } from 'react-router-dom'
 
-export function HomePage() {
-  const { data: user } = useAuth()
-
+export function HomePage({ user }) {
   if (user.isSignupComplete === false) {
     return <Navigate to="/edit-user-details" />
   }
@@ -245,12 +242,9 @@ export function HomePage() {
 <TabItem value="ts" label="TypeScript">
 
 ```tsx title=src/HomePage.tsx
-import { useAuth } from 'wasp/client/auth'
 import { Navigate } from 'react-router-dom'
 
-export function HomePage() {
-  const { data: user } = useAuth()
-
+export function HomePage({ user }: { user: AuthUser }) {
   if (user.isSignupComplete === false) {
     return <Navigate to="/edit-user-details" />
   }
--- a/web/static/_redirects
+++ b/web/static/_redirects
@@ -8,6 +8,7 @@
 /docs/deploying                    /docs/deployment/intro                    301
 /docs/advanced/deployment/overview /docs/deployment/intro                    301
 /docs/integrations/css-frameworks  /docs/project/css-frameworks             301
+/docs/data-model/backends          /docs/data-model/databases               301
 /docs/guides/websockets            /docs/advanced/web-sockets               301
 /docs/guides/testing               /docs/project/testing                    301
 /docs/guides/middleware-customization /docs/advanced/middleware-config       301
--- a/web/docs/data-model/databases.md
+++ b/web/docs/data-model/databases.md
@@ -225,7 +225,7 @@ async function createUser(prisma, data) {
             create: {
               providerName: 'username',
               providerUserId: data.username,
-              providerData: sanitizeAndSerializeProviderData({
+              providerData: await sanitizeAndSerializeProviderData({
                 hashedPassword: data.password
               }),
             },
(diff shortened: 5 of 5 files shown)


web/docs/advanced/jobs.md
```diff
@@ -233,7 +233,7 @@ The Job declaration has the following fields:
   :::note Job executors
   Our jobs need job executors to handle the _scheduling, monitoring, and execution_.
 
-  `PgBoss` is currently our only job executor, and is recommended for low-volume production use cases. It requires that your database provider is set to `"postgresql"` in your `schema.prisma` file. Read more about setting the provider [here](../data-model/backends.md#postgresql).
+  `PgBoss` is currently our only job executor, and is recommended for low-volume production use cases. It requires that your database provider is set to `"postgresql"` in your `schema.prisma` file. Read more about setting the provider [here](../data-model/databases.md#postgresql).
```
Martinsos, line 236: You learned aboutthis by trying to follow the link?
If so, I wish we knew about this link being wrong without having to manually figure it out! I thought Docusaurus warns about wrong internal links during building hm, but I guess it doesn't or it didn't for this one for some reason?

web/static/_redirects
```diff
@@ -8,6 +8,7 @@
 /docs/deploying                    /docs/deployment/intro                    301
 /docs/advanced/deployment/overview /docs/deployment/intro                    301
 /docs/integrations/css-frameworks  /docs/project/css-frameworks             301
+/docs/data-model/backends          /docs/data-model/databases               301
```
Martinsos, line 11: Hm but this is probably not 1 on 1 mapping? I am guessing that backends was wider than just databases? If it is just name change then makews sense, but if not and backends was wider, maybe better not to map at all?

Martinsos (commented):
Looks good to me! I will leave approving to @infomiho though, since he knows the Auth part better.

web/docs/advanced/jobs.md
sodic: Docosaurus throws an error, yes.

Why do you think it doesn't warn about it?

web/docs/auth/overview.md
```diff
@@ -313,7 +313,7 @@ export function Main() {
 </Tabs>
 
 :::tip
-Since the `user` prop is only available in a page's React component: use the `user` prop in the page's React component and the `useAuth` hook in any other React component.
+While the `user` prop is only available in an authenticated page's React component, it is the preferred option because we can be certain that the user exists on that page. On the other hand, `useAuth` can be used when the user might not exist, making it better suited for non-authenticated components.
 :::
```
sodic, line 317: I'd move this to the beginning of the chapter. It's best to warn users about a better approach at the beginning.

And we can lose the tip - the text is part of the document's theme.

web/docs/auth/overview.md
```diff
@@ -184,7 +184,7 @@ There are two ways to access the `user` object on the client:
 
 #### Using the `user` prop
 
-If the page's declaration sets `authRequired` to `true`, the page's React component receives the `user` object as a prop:
+The preferred way to fetch user data on authenticated pages is by using the `user` prop. If the page's declaration sets `authRequired` to `true`, the page's React component receives the `user` object as a prop:
```
sodic, line 187:
```suggestion
If the page's declaration sets `authRequired` to `true`, the page's React component receives the `user` object as a prop. This is the simplest way to access the user inside an authenticated page:
```

web/docs/advanced/jobs.md
FranjoMindek: I changed this link because I renamed the file from backends to databases. I've added rest of them now.

Martinsos: Ah I didn't realize you renamed the file! That is why I asked about why we didn't get the warning before, because I thought it is an old broken link that you picked up and fixed. Ok got it.

Martinsos:
@FranjoMindek did I get something wrong, or did you add quite some changes after my last review -> what happened there? Are those still in the initial scope of this PR, or are these additional fixes?

FranjoMindek:
Initial commit didn't capture all places referencing the old name of the document. I've added rest of them.
Filip also wanted to word things a bit differently so I've added those too.
Lastly, we found mentions of "wasp-lang" instead of wasp in some old document so Filip said I should fix those too.

waspc/src/Wasp/Generator/DbGenerator.hs
```diff
@@ -258,7 +258,7 @@ generatePrismaClient spec projectRootDir = do
     generatePrismaClientIfEntitiesExist :: IO (Maybe GeneratorError)
     generatePrismaClientIfEntitiesExist
       | entitiesExist =
-          either (Just . GenericGeneratorError) (const Nothing) <$> DbOps.generatePrismaClient projectRootDir
+        either (Just . GenericGeneratorError) (const Nothing) <$> DbOps.generatePrismaClient projectRootDir
```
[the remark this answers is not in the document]
FranjoMindek: There is mention of it in the docs, I think I've set it up but I'll check it out.

web/docs/auth/overview.md
```diff
@@ -1173,7 +1170,7 @@ The extra fields can be either **objects** or **render functions** (you can comb
 
    The render function has the following signature:
 
-   ```ts
+   ```
```
infomiho, line 1173: Maybe we can improve this code to be valid TS so we get syntax highlighting? That could be a nice improvement for readers.

FranjoMindek: What would you write? 
If we want syntax highlighting we need something like:
```ts
type renderFn = (form: UseFormReturn, state: FormState) => React.ReactNode
```

But this may lead on readers that such type exists.

web/docs/advanced/jobs.md
Martinsos: Feel free to resolve this comment once you read it.

web/docs/auth/overview.md
infomiho: I'd write it like that:
```ts
type RenderFn = (form: UseFormReturn, state: FormState) => React.ReactNode
```
We are not writing an import statement, I think the readers would get it 👍

FranjoMindek: Now that I've looked into the source code, we have this:
```ts
// PRIVATE API
export type AdditionalSignupFieldRenderFn = (
  hookForm: UseFormReturn<LoginSignupFormFields>,
  formState: FormState
) => React.ReactNode
```

Even though this is a private API I still think it would be preferable to include the same as the source code. It could be helpful to possible future contributors or anyone who wants to dig deeper into wasp. I would omit the `// PRIVATE API` comment and `export` for the docs.
@sodic you had some reservations around this, do you think this is fine?

web/docs/auth/social-auth/overview.md
```diff
@@ -215,7 +215,8 @@ export const userSignupFields = defineUserSignupFields({
 
 #### 3. Showing the Correct State on the Client
 
-You can query the user's `isSignupComplete` flag on the client with the [`useAuth()`](../../auth/overview) hook.
+You can check the `isSignupComplete` flag on the `user` object.
+If you're on an authenticated page the `user` can be fetched through the [props](../../auth/overview#getting-the-user-in-authenticated-routes), otherwise use the [`useAuth()`](../../auth/overview#getting-the-user-in-non-authenticated-routes) hook instead.
```
sodic, line 219: I'd change it to something like this:
```suggestion
Authenticated pages come with the [`user` prop](../../auth/overview#getting-the-user-in-authenticated-routes) which gives you access to the current user. If the `user` prop is out of reach, fetch the current user with the  [`useAuth()` hook](../../auth/overview#getting-the-user-in-non-authenticated-routes).
```
Explanation:
- Active voice is better than passive voice
- Shorter sentences are easier to parse
- "fetched" implies a network request, which the prop doesn't do.
- URLS look more natural when they span the entire noun phrase (e.g., "useAuth hook" instead of "useAuth").
- Other stuff just has to do with the flow and is probably a little subjective.

I also liked your previous explanation and we seemed to have lost it. You can append it here:
```
As a rule of thumb:
- If a component only makes sense for authenticated users, it's best to rely on the `user` prop.
- If a component works for both authenticated and non-authenticated users and needs information about the current user, reach for the `useAuth` hook.

sodic: Also, seems like we repeat the same thing (hook vs prop) both here (in social auth overview) and in the regular auth overview.

@infomiho Is this on purpose or did we miss it?

web/docs/auth/overview.md
sodic: Hm, maybe too much for my taste. I'd go with what you had originally (just the type signature) and maybe even inline it into the text.

That said, I am fine with whatever you like best.

One final piece of advice, ask yourself this: _"If I could have proper syntax highlighting (or none) for all options, which one would I pick?"_ And then go with that one.

We always want the best possible content, not the content that makes our tools the happiest :)

Anyway, pick whatever you like and feel free to resolve!

sodic (approved):
Nice work. I'm approving but left some suggestions

web/docs/auth/social-auth/overview.md
sodic: Yeah, it's not ideal. Because it's not really about routes/pages. It's about any component inside the tree:
- Anything that's in the tree whose root is an authenticated page should use the prop (and we should pass it down accordingly)
- Other stuff should use `useAuth`

As I was writing the first point, something seemed off. @FranjoMindek What would be the best way to do this, if we had a clean slate? Would the user always sit in the context (instead of getting passed via the props)?

On the topic on where to put this rule of thumb... If you can't fit it anywhere, we can leave it out. Like you said, we already have those titles, they probably explain it well enough.

FranjoMindek: > ### Getting the `user` in non-authenticated routes
Now that I think about it, it may lead people to think that it can't be used in authenticated routes.
We do say that it's a `useQuery` wrapper, but 🤷

infomiho (approved):

web/docs/auth/social-auth/overview.md
sodic: Ok, let's merge this know and we'll come back to this discussion at a different time (has nothing to do with the PR anyway).

Issue #2436, opened 2025-01-03
Use Node.js 20 everywhere since Node.js 18 is soon EOL


infomiho:
Based on this: https://endoflife.date/nodejs

We are using Node.js 18 in our Dockerfile for the server image. 

Also, we'll need to upgrade our `oldestWaspSupportedNodeVersion = SV.Version 18 18 0` version to the Node.js 20 so users will get an error that they need to upgrade their Node.js version.

PR #2537, opened 2025-03-07
Upgrade Node to 20

FranjoMindek:
Related Issues: https://github.com/wasp-lang/wasp/issues/2436


waspc/src/Wasp/Node/Version.hs
```diff
@@ -25,7 +25,7 @@ import Wasp.Util (indent)
 --   - /waspc/.nvmrc
 --   - /web/docs/introduction/getting-started.md -> "Requirements" section.
 oldestWaspSupportedNodeVersion :: SV.Version
-oldestWaspSupportedNodeVersion = SV.Version 18 18 0
+oldestWaspSupportedNodeVersion = SV.Version 20 18 0
```
Martinsos, line 28: Hm why is it 20.18.0? There was some reason we had for specifically 18.18.0 although I can't remember exactly, but I doubt that we have similar/same reason for exactly the same combo of minor version and patch -> sounds instead like you updated the first number without considering the rest?

waspc/data/Generator/templates/Dockerfile
```diff
@@ -3,7 +3,7 @@
 #   Because if not, we had situations where it would use the different version
 #   locally and on Github CI. This way we ensure exact version is used,
 #   and also have control over updating it (instead of update surprising us).
-FROM node:{= nodeVersion =}-alpine3.17 AS node
+FROM node:{= nodeVersion =}-alpine3.21 AS node
```
Martinsos, line 6: I am curious, why is 3.21 the right version here? That is what they support up there on Dockerhub?

We should make sure that wasp apps that are built and then use this dockerfile run correctly. I can't remember if we have automatic test for that, I think we might not -> please check if we do (in CI), if we don't, you will have to test it manually. You might also open an issue for automating it, if we don't have one yet.

Martinsos (changes requested):
Looks almost done! There is one thing that is likely not correct so I requested changes, and also it might need some more testing.

Regarding making sure you did updates on all the needed places -> could you explain how you went about it, to help me build confidence in your process? Because just looking at PR, I am wondering if we missed something somewhere, something that depended on version of node, but wasn't captured via tests or types. E.g. npm version, should we update that also? Do we even restrict that, we did before, maybe not any more. Maybe something else? Likely not hm.

waspc/src/Wasp/Node/Version.hs
FranjoMindek: I can change it to 20 0 0. I just thought you wanted to be very specific for reproducibility. It just happens so that the 20.18 is the latest of version 20.

waspc/data/Generator/templates/Dockerfile
FranjoMindek: I've asked Filip about that. 
So for alpine image we can either `use node:20-alpine` (latest) or `use node:20-alpine-xyz`.
Last time there was specific version so I asked Filip which one of the above do we want to use. He said probably the fixed version one because reproducibility. The latest one is `node:20-alpine3.21` for node 20.

Martinsos: We certainly want a fixed versino, because before it wasn't fixed and something changed and dockerfile stopped working.
I was curious about why specifically 21 -> but ok, you said that is what they have for node:20. Makes sense.
Regarding testing, I will repeat what I asked above -> did you test taht this specific Dockerfile works correctly?

FranjoMindek: Good catch, I've tested it and it worked for node version "20", but then I changed that to "20.18" but forgot to test it again and it doesn't work anymore.
`20.18.0-alpine3.20` works well though so I'll change it to that.

waspc/src/Wasp/Node/Version.hs
FranjoMindek: I think that that is fine. Since if we dont restrict ourselves to specific version of node we must also propagate that further. E.g. alpine in dockerfile which differs depending on node. Then everything must be relative.

Martinsos: Hm sorry I didn't get this actually. So we can go with 20.18.0 or with 20.0.0. What is your final choice, and why?

waspc/data/Generator/templates/Dockerfile
Martinsos: Ok that is weird, so `20.18.0-alpine3.21` didn't work? We should maybe add a comment for that -> we could easily want to update it in the future and stumble into it again. Why didn't it work?

FranjoMindek: Initially I've tested on node version `20`, without specifying minor or patch. 

I didn't realize that wasp would still specify `20.0.0` for the e.g. dockerfile, instead of `20`.
When I tested it with actual wasp build, rather than quick dockerfile number change, it didn't work.

There is a bit of problem of which node-alpine docker images actually exist.
For the `20.18.0` version `20.18.0-alpine3.20` exists but `20.18.0-alpine3.21` doesn't.
While for the `20.18` or  `20.18.3` version both alpine variations exist.
For all of them the `alpine` without version exists. e.g. `20.18.0-alpine`.

As it currently stands, we can either:
1. Find specific version of alpine for the specific node version
2. Relax the constraint on the alpine version
     - `node:{= nodeVersion =}-alpine3.20` -> `node:{= nodeVersion =}-alpine`
     - we no longer have to adjust alpine for each version, as it will grab the ?latest? one for that version
     
Maybe the version markers (major, minor, patch) that are 0 should be ignored when converting to strings?
E.g.
`oldestWaspSupportedNodeVersion = SV.Version 20 18 0` would print out to `20.18`?
Do we really need to always specify minor and patch?

waspc/src/Wasp/Node/Version.hs
FranjoMindek: We went with 20.0.0, alpine downgraded to 3.17 to match.

waspc/data/Generator/templates/Dockerfile
FranjoMindek: We decided to stay with fully fixed versions for now.
Less possibilities for stuff breaking.

waspc/data/Generator/templates/Dockerfile
```diff
@@ -50,7 +50,7 @@ RUN cd .wasp/build/server && npm run bundle
 FROM base AS server-production
 # In case they want to use python3 in their app.
 RUN apk add --no-cache python3
-ENV NODE_ENV production
```
Martinsos, line 53: Hm interesting that we had this wrong, nice fix! Was it really working wrong, or was it somehow working correctly still even though syntax was not as it should be? If you don't know doesn't matter, don't spend time testing, I am just curious.

Martinsos (approved):
LGTM!

waspc/data/Generator/templates/Dockerfile
FranjoMindek: Both work correctly, just that the space variant is legacy/outdated variant.
Current docs push for the `ENV name=value` approach.

Issue #2063, opened 2024-05-27
Remove "hello world" on server's "/" route


Martinsos:
Suggestion by the user from Discord:

> Hiding away "hello world" when visiting the server url in production, by returning a 301 redirect to the client url (or something similar would be good)

We can probably just go away with it completely, that is the easiest thing to do -> if they do want to have a route like this, they can add it on their own, it is weird that we create it for them and force them to have it.

We used it kind of as a health check, but if we really need that, we should implement it properly.

sodic:
After removing this, double check we don't mention it in the docs (e.g., the tutorial).

PR #2463, opened 2025-01-22
Merged headless todoApp with waspc todoApp

infomiho:
This PR:
- [x] Merges `waspc/headless-test/examples/todoApp` with `waspc/examples/todoApp`
- [x] Runs headless tests in CI on `waspc/examples/todoApp`

Closes #2024

PR #2538, opened 2025-03-07
Remove hello world route from the server

FranjoMindek:
Related issues: https://github.com/wasp-lang/wasp/issues/2063

### DIFF
--- a/waspc/data/Generator/templates/server/src/routes/index.js
+++ b/waspc/data/Generator/templates/server/src/routes/index.js
@@ -16,6 +16,10 @@ import { rootCrudRouter } from './crud/index.js'
 const router = express.Router()
 const middleware = globalMiddlewareConfigForExpress()
 
+router.get('/', middleware, function (_req, res, _next) {
+  res.status(200).send();
+})
+
 {=# isAuthEnabled =}
 router.use('/auth', middleware, auth)
 {=/ isAuthEnabled =}
--- a/waspc/data/Generator/templates/server/src/routes/index.js
+++ b/waspc/data/Generator/templates/server/src/routes/index.js
@@ -16,10 +16,6 @@ import { rootCrudRouter } from './crud/index.js'
 const router = express.Router()
 const middleware = globalMiddlewareConfigForExpress()
 
-router.get('/', middleware, function (_req, res, _next) {
-  res.json('Hello world')
-})
-
 {=# isAuthEnabled =}
 router.use('/auth', middleware, auth)
 {=/ isAuthEnabled =}
--- a/waspc/data/Generator/templates/server/src/routes/index.js
+++ b/waspc/data/Generator/templates/server/src/routes/index.js
@@ -16,7 +16,7 @@ import { rootCrudRouter } from './crud/index.js'
 const router = express.Router()
 const middleware = globalMiddlewareConfigForExpress()
 
-router.get('/', middleware, function (_req, res, _next) {
+router.get('/', middleware, function (_req, res) {
   res.status(200).send();
 })
 
--- a/waspc/headless-test/README.md
+++ b/waspc/headless-test/README.md
@@ -1,6 +1,7 @@
 ## Headless Test
 
 We are testing if a simple scenario works for the user each time:
+
 - signup
 - login
 - create a new task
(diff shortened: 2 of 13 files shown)


FranjoMindek:
I don't know if we want to add "/health" route in this PR?

Martinsos:
Nice! We don't have to add "health" check for now -> they can easily implement it on their own if they want with API feature in Wasp. We might add it in future if we will have some other mechanism that uses it, but till then let's not have it.

CI is not passing -> I thought it will fail because of CLI e2e tests not being updated (check the README for instructions on that, and also `run` script should be helpful), but I see it failing on Playwright which is weird -> is that because of something actually being wrong in the code, or is it because Playwright was relying on the `/` route on the server, or because it is flaky?

Martinsos (changes requested):
Rejecting because CI is not passing, let's look into that!

Zeko369:
Isn't having something like this in prod kinda dumb, site looks down to traefik/nginx (can't use for uptime/healtcheck)?
CleanShot 2025-03-07 at 22 24 27@2x

302/307 to `WASP_WEB_CLIENT_URL` would be a way better option if you're not for a "OK" health response

Zeko369 (commented):
Isn't having something like this in prod kinda dumb since it's exposing internal details (using express) and site looks down to traefik/nginx (can't use for uptime/healtcheck)?
CleanShot 2025-03-07 at 22 24 27@2x

302/307 to `WASP_WEB_CLIENT_URL` would be a way better option if you're not for a "OK" health response

Martinsos:
> Isn't having something like this in prod kinda dumb since it's exposing internal details (using express) and site looks down to traefik/nginx (can't use for uptime/healtcheck)? CleanShot 2025-03-07 at 22 24 27@2x
> 
> 302/307 to `WASP_WEB_CLIENT_URL` would be a way better option if you're not for a "OK" health response

Is exposing usage of express a big deal? Doesn't sound like it to me. Also, I haven't tried it right now but I am assuming that also trying to open any route (e.g. `/is/this/express`) that doesn't exist will return the same thing "Can't GET ...", which means if they want to check if it is express, they can, regardless of what we do with the `/` route.

Also, if they want to use `/` for healtcheck, they can add such route on their own very easily.
That said, if it turns out a lot of people want it, we could put it back as a default. But I would rather have people ask for it and then we add it, then having it while not a lot of people really need it.

302/307 to web client is interesting idea, if we would be going with some default! Also not sure about it though, simple 200 might be enough.

infomiho:
Just to chime in:
- I think we shouldn't remove the route, but rather r**eplace it with a empty response or "ok" with 200 code**. This way a number of external tools work as expected (Playwright, Nginx/Traefik)
- In the future, we should allow users to set their custom route at `/` to offer complete control over the endpoint.
- IMHO exposing the fact that we use Express isn't that big of a deal (people testing for vulns often do a kitchen sink of probes, so they'll figure it out anyways).

Martinsos:
> Just to chime in:
> 
> * I think we shouldn't remove the route, but rather r**eplace it with a empty response or "ok" with 200 code**. This way a number of external tools work as expected (Playwright, Nginx/Traefik)
> * In the future, we should allow users to set their custom route at `/` to offer complete control over the endpoint.
> * IMHO exposing the fact that we use Express isn't that big of a deal (people testing for vulns often do a kitchen sink of probes, so they'll figure it out anyways).

This makes sense to me. I believe they can already override if they want? How hard is that?

waspc/headless-test/start.js
```diff
@@ -4,12 +4,15 @@ const readline = require('linebyline');
 function spawn(name, cmd, args, done) {
   const spawnOptions = {
     detached: true,
+    // MacOS: uncomment the next line
+    // stdio: ["ignore", "pipe", "pipe"],
```
sodic, line 8: Miho also did something here on #2463, so have a chat with him.

waspc/headless-test/README.md
sodic, line 1: Nice changes!

waspc/data/Generator/templates/server/src/routes/index.js
```diff
@@ -17,7 +17,7 @@ const router = express.Router()
 const middleware = globalMiddlewareConfigForExpress()
 
 router.get('/', middleware, function (_req, res, _next) {
```
sodic, line 19: Since `_next` is positional, you can drop it.

Old code I know, but improvements are always welcome :)

sodic (approved):
Tested it out, works!

Nice work with adding the context to the README. Just make sure to have a chat with @infomiho before merging (and to update the `e2e` tests of course).

@Martinsos The CI is failing because of E2E tests, I told Franjo to only update them after he gets an approval.

Martinsos (approved):

PR #2547, opened 2025-03-12
Extract ANSI coloring into a single module

infomiho:
We wanted to color the `console.log` output and we defined the colors in multiple different places. This brings the colors in one place.

We could go for something like [picocolors](https://github.com/alexeyraspopov/picocolors) which is only a 7 kB dep and get all the colors and a better API.


waspc/data/Generator/templates/sdk/wasp/package.json
```diff
@@ -54,6 +54,8 @@
       "./universal/types": "./dist/universal/types.js",
       {=!  Used by our code, uncodumented (but accessible) for users. =}
       "./universal/validators": "./dist/universal/validators.js",
+      {=!  Used by our code, uncodumented (but accessible) for users. =}
```
Martinsos, line 57: hah uncodumented. But it is an old mistake.

waspc/data/Generator/templates/sdk/wasp/server/email/core/providers/dummy.ts
```diff
@@ -1,7 +1,9 @@
 import { DummyEmailProvider, EmailSender } from "../types";
 import { getDefaultFromField } from "../helpers.js";
 
-const yellowColor = "\x1b[33m%s\x1b[0m";
```
Martinsos, line 4: This isn't just yellow though, is it? This look smore complex, like two things? Ah I get it now, it is a format string, not just yellow color.

waspc/data/Generator/templates/sdk/wasp/universal/ansiColors.ts
```diff
@@ -0,0 +1,13 @@
+export const colors = {
+  red: '\x1b[31m',
+  yellow: '\x1b[33m',
+} as const;
+
+export const resetColor = "\x1b[0m";
+
+// Used with console.log() to colorize the output
+// Example: console.log(getConsoleLogColorTemplate('red'), 'This is red text');
+export function getConsoleLogColorTemplate(colorKey: keyof typeof colors): string {
```
Martinsos, line 10: The naming threw me off a bit. In C, for printf, this is named "format string" from what I know, and I checked MDN docs for console.log and while they call it string with substitutions, they do at one point use term "format string" also, I believe because it is such a mainstream term for this.
So would it make sense to rename this to `getColorizedConsoleFormatString` or something like that? `Log` is not neccessary, since it will work also for warning and error I assume? If they also offer format string?

Btw I used different approach in Haskell codebase: I had functions called something like `style(color, string)` which would add color in front of string and reset at the end. What was cool about this was that I could apply multiple styles: `style(Yellow, style(Bold, "WARNING"))` and use that wherever I want. It seems to me like you locked yourself a bit with going for format string / template. Not a biggie though, just an observation because I was also implementing this before.

Here is my haskell implementation:
```hs
module Wasp.Util.Terminal
  ( Style (..),
    applyStyles,
    styleCode,
    escapeCode,
    resetCode,
  )
where

import Data.List (foldl')

data Style
  = Black
  | Red
  | Green
  | Yellow
  | Blue
  | Magenta
  | Cyan
  | White
  | BlackBg
  | RedBg
  | GreenBg
  | YellowBg
  | BlueBg
  | MagentaBg
  | CyanBg
  | WhiteBg
  | Bold
  | Underline
  | Blink
  deriving (Show, Eq)

-- | Given a string, returns decorated string that when printed in terminal
-- will have same content as original string but will also exibit specified styles.
applyStyles [] str = str
applyStyles _ "" = ""
applyStyles styles str = foldl' applyStyle str styles ++ escapeCode ++ resetCode
  where
    applyStyle s style = escapeCode ++ styleCode style ++ s

styleCode :: Style -> String
styleCode Black = "[30m"
styleCode Red = "[31m"
styleCode Green = "[32m"
styleCode Yellow = "[33m"
styleCode Blue = "[34m"
styleCode Magenta = "[35m"
styleCode Cyan = "[36m"
styleCode White = "[37m"
styleCode BlackBg = "[40m"
styleCode RedBg = "[41m"
styleCode GreenBg = "[42m"
styleCode YellowBg = "[43m"
styleCode BlueBg = "[44m"
styleCode MagentaBg = "[45m"
styleCode CyanBg = "[46m"
styleCode WhiteBg = "[47m"
styleCode Bold = "[1m"
styleCode Underline = "[4m"
styleCode Blink = "[5m" -- Blink does not work in all terminal emulators (e.g. on mac in iTerm2).

escapeCode :: String
escapeCode = "\ESC"

resetCode :: String
resetCode = "[0m"

```

Martinsos (approved):
Nice! I left some comments, but nothing major nor even required -> solve them anyway you like, I am good with it, approved!

Oh and using that small library -> ha we could, although this is so simple that I don't mind us having our own code for this also.

waspc/data/Generator/templates/sdk/wasp/package.json
infomiho: Yeah, we should fix it at some point 👍

waspc/data/Generator/templates/sdk/wasp/universal/ansiColors.ts
infomiho: I wanted to keep as simple as possible since our needs were quite basic based on the places of use. If we start being more creative with the colors and styles in Typescript, I'd then just use the 7kB library which would cover our needs.

I'll rename the function to make more sense, thanks!

PR #2548, opened 2025-03-13
Convert hardcoded strings to variables

### LINKED ISSUE #2220: Refactor database engine definitions
See this comment for details: https://github.com/wasp-lang/wasp/pull/2192#issuecomment-2233177389

FranjoMindek:
Converts all hardcoded strings ("postgresql" and "sqlite) to variable alternatives.

Left out strings include:
1. Filepath strings e.g. `"-v %s:/var/lib/postgresql/data"`, `"postgresql://%s:%s@localhost:%d/%s"`
2. Test `Text` strings e.g.
```hs
getPrismaSchemaWithConfig restOfPrismaSource =
  Util.getPrismaSchema
    [trimming|
      datasource db {
        provider = "postgresql"
        url      = env("DATABASE_URL")
      }
      generator client {
        provider = "prisma-client-js"
      }
      ${restOfPrismaSource}
    |]
```

If we wish to also include variables in tests `Text` strings I can change that.

Fixes #2220

### DIFF
--- a/waspc/ChangeLog.md
+++ b/waspc/ChangeLog.md
@@ -7,6 +7,7 @@
 - Custom values for `theme-color` are now supported ([#2546](https://github.com/wasp-lang/wasp/pull/2546) by @andrsdt).
 - Increased the minimum Node version to 20.0.0 ( [#2537](https://github.com/wasp-lang/wasp/pull/2537) )
 - Invalid CLI commands now properly return non-zero exit code ( [#2522](https://github.com/wasp-lang/wasp/pull/2552) )
+- Prisma's datasource database provider string literals are now variablized ( [#2548](https://github.com/wasp-lang/wasp/pull/2548) )
 
 ### Bug fixes
 
--- a/waspc/src/Wasp/AppSpec/Valid.hs
+++ b/waspc/src/Wasp/AppSpec/Valid.hs
@@ -40,6 +40,7 @@ import Wasp.Generator.Crud (crudDeclarationToOperationsList)
 import Wasp.Node.Version (oldestWaspSupportedNodeVersion)
 import qualified Wasp.Node.Version as V
 import qualified Wasp.Psl.Ast.Model as Psl.Model
+import qualified Wasp.Psl.Db as Psl.Db
 import qualified Wasp.Psl.Util as Psl.Util
 import Wasp.Psl.Valid (getValidDbSystemFromPrismaSchema)
 import qualified Wasp.SemanticVersion as SV
@@ -165,7 +166,7 @@ validateOnlyEmailOrUsernameAndPasswordAuthIsUsed spec =
 validateDbIsPostgresIfPgBossUsed :: AppSpec -> [ValidationError]
 validateDbIsPostgresIfPgBossUsed spec =
   [ GenericValidationError
-      "The database provider in the schema.prisma file must be \"postgresql\" since there are jobs with executor set to PgBoss."
+      ("The database provider in the schema.prisma file must be \"" ++ Psl.Db.pslPostgresqlKeyword ++ "\" since there are jobs with executor set to PgBoss.")
     | isPgBossJobExecutorUsed spec && not (isPostgresUsed spec)
   ]
 
--- a/waspc/src/Wasp/AppSpec/Valid.hs
+++ b/waspc/src/Wasp/AppSpec/Valid.hs
@@ -166,7 +166,7 @@ validateOnlyEmailOrUsernameAndPasswordAuthIsUsed spec =
 validateDbIsPostgresIfPgBossUsed :: AppSpec -> [ValidationError]
 validateDbIsPostgresIfPgBossUsed spec =
   [ GenericValidationError
-      ("The database provider in the schema.prisma file must be \"" ++ Psl.Db.pslPostgresqlKeyword ++ "\" since there are jobs with executor set to PgBoss.")
+      ("The database provider in the schema.prisma file must be \"" ++ Psl.Db.dbProviderPostgresqlStringLiteral ++ "\" since there are jobs with executor set to PgBoss.")
     | isPgBossJobExecutorUsed spec && not (isPostgresUsed spec)
   ]
 
--- a/waspc/src/Wasp/Generator/DbGenerator.hs
+++ b/waspc/src/Wasp/Generator/DbGenerator.hs
@@ -64,11 +64,11 @@ genPrismaSchema ::
   Generator FileDraft
 genPrismaSchema spec = do
   (datasourceProvider :: String) <- case dbSystem of
-    AS.Db.PostgreSQL -> return Pls.Db.pslPostgresqlKeyword
+    AS.Db.PostgreSQL -> return Pls.Db.dbProviderPostgresqlStringLiteral
     AS.Db.SQLite ->
       if AS.isBuild spec
         then logAndThrowGeneratorError $ GenericGeneratorError "SQLite (a default database) is not supported in production. To build your Wasp app for production, switch to a different database. Switching to PostgreSQL: https://wasp.sh/docs/data-model/backends#migrating-from-sqlite-to-postgresql ."
-        else return Pls.Db.pslSqliteKeyword
+        else return Pls.Db.dbProviderSqliteStringLiteral
 
   entities <- getEntitiesForPrismaSchema spec
 
--- a/waspc/src/Wasp/Generator/DbGenerator.hs
+++ b/waspc/src/Wasp/Generator/DbGenerator.hs
@@ -48,6 +48,7 @@ import qualified Wasp.Psl.Ast.Argument as Psl.Argument
 import qualified Wasp.Psl.Ast.ConfigBlock as Psl.Ast.ConfigBlock
 import qualified Wasp.Psl.Ast.Model as Psl.Model
 import qualified Wasp.Psl.Ast.Schema as Psl.Schema
+import qualified Wasp.Psl.Db as Pls.Db
 import qualified Wasp.Psl.Format as Psl.Format
 import qualified Wasp.Psl.Generator.Schema as Psl.Generator.Schema
 import Wasp.Util (checksumFromFilePath, hexToString, ifM, (<:>))
@@ -63,11 +64,11 @@ genPrismaSchema ::
   Generator FileDraft
 genPrismaSchema spec = do
   (datasourceProvider :: String) <- case dbSystem of
-    AS.Db.PostgreSQL -> return "postgresql"
+    AS.Db.PostgreSQL -> return Pls.Db.pslPostgresqlKeyword
     AS.Db.SQLite ->
       if AS.isBuild spec
         then logAndThrowGeneratorError $ GenericGeneratorError "SQLite (a default database) is not supported in production. To build your Wasp app for production, switch to a different database. Switching to PostgreSQL: https://wasp.sh/docs/data-model/backends#migrating-from-sqlite-to-postgresql ."
-        else return "sqlite"
+        else return Pls.Db.pslSqliteKeyword
 
   entities <- getEntitiesForPrismaSchema spec
 
--- a/waspc/src/Wasp/Project/Db.hs
+++ b/waspc/src/Wasp/Project/Db.hs
@@ -11,7 +11,6 @@ where
 
 import StrongPath (Abs, Dir, Path')
 import qualified Wasp.AppSpec as AS
-import qualified Wasp.AppSpec.App.Db as AS.App.Db
 import qualified Wasp.AppSpec.App.Db as AS.Db
 import Wasp.Project.Common (WaspProjectDir)
 import qualified Wasp.Project.Db.Dev.Postgres as DevPostgres
@@ -19,6 +18,7 @@ import qualified Wasp.Project.Db.Dev.Sqlite as DevSqlite
 import qualified Wasp.Psl.Ast.Argument as Psl.Argument
 import qualified Wasp.Psl.Ast.ConfigBlock as Psl.ConfigBlock
 import qualified Wasp.Psl.Ast.Schema as Psl.Schema
+import qualified Wasp.Psl.Db as Psl.Db
 import Wasp.Psl.Generator.Argument (generateExpression)
 import Wasp.Psl.Util (findPrismaConfigBlockKeyValuePair)
 
@@ -30,8 +30,8 @@ makeDevDatabaseUrl ::
 makeDevDatabaseUrl waspProjectDir dbSystem decls = do
   (appName, _) <- AS.getApp decls
   case dbSystem of
-    AS.App.Db.PostgreSQL -> Just $ DevPostgres.makeDevConnectionUrl waspProjectDir appName
-    AS.App.Db.SQLite -> Just DevSqlite.defaultDevDbFile
+    AS.Db.PostgreSQL -> Just $ DevPostgres.makeDevConnectionUrl waspProjectDir appName
+    AS.Db.SQLite -> Just DevSqlite.defaultDevDbFile
 
 databaseUrlEnvVarName :: String
 databaseUrlEnvVarName = "DATABASE_URL"
@@ -50,8 +50,9 @@ data DbSystemParseError = UnsupportedDbSystem String | MissingDbSystem
 getDbSystemFromPrismaSchema :: Psl.Schema.Schema -> Either DbSystemParseError AS.Db.DbSystem
 getDbSystemFromPrismaSchema prismaSchema =
   case getDbProviderFromPrismaSchema prismaSchema of
-    Just (Psl.Argument.StringExpr "postgresql") -> Right AS.App.Db.PostgreSQL
-    Just (Psl.Argument.StringExpr "sqlite") -> Right AS.App.Db.SQLite
+    Just (Psl.Argument.StringExpr provider)
+      | provider == Psl.Db.pslPostgresqlKeyword -> Right AS.Db.PostgreSQL
+      | provider == Psl.Db.pslSqliteKeyword -> Right AS.Db.SQLite
     Just anyOtherExpression -> Left $ UnsupportedDbSystem $ generateExpression anyOtherExpression
     Nothing -> Left MissingDbSystem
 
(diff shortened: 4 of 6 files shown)


waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -165,7 +165,7 @@ validateOnlyEmailOrUsernameAndPasswordAuthIsUsed spec =
 validateDbIsPostgresIfPgBossUsed :: AppSpec -> [ValidationError]
 validateDbIsPostgresIfPgBossUsed spec =
   [ GenericValidationError
-      "The database provider in the schema.prisma file must be \"postgresql\" since there are jobs with executor set to PgBoss."
+      ("The database provider in the schema.prisma file must be \"" ++ AS.Db.providerToString AS.Db.PostgreSQL ++ "\" since there are jobs with executor set to PgBoss.")
```
Martinsos, line 168: Ok, so this is first usage site. I can see there that this string must be exact the same string that Prisma uses to specify postgre as a provider. So you are assuming here that `providerToString` returns Prisma keyword/identifier for Postgre. But that function, `providerToString`, in no way offers that guarantee, be it via implemention, types or docs. So that is a future bug waiting to happen.

waspc/src/Wasp/Generator/DbGenerator.hs
```diff
@@ -63,11 +63,11 @@ genPrismaSchema ::
   Generator FileDraft
 genPrismaSchema spec = do
   (datasourceProvider :: String) <- case dbSystem of
-    AS.Db.PostgreSQL -> return "postgresql"
```
Martinsos, line 66: Again using it for Prisma, ok.

Btw note that if you are to define somewhere in our codebase which string does Prisma use as identifier for postgre, you don't have to do it in AppSpec part of our codebase. It has nothign to do with Wasp or AppSpec, it is Prisma-specific thing. I would instead expect to see such definition somewhere in the part of our codebase where we deine Prisma specific stuff (PSL).

waspc/src/Wasp/Project/Db.hs
```diff
@@ -50,10 +50,14 @@ data DbSystemParseError = UnsupportedDbSystem String | MissingDbSystem
 getDbSystemFromPrismaSchema :: Psl.Schema.Schema -> Either DbSystemParseError AS.Db.DbSystem
 getDbSystemFromPrismaSchema prismaSchema =
   case getDbProviderFromPrismaSchema prismaSchema of
-    Just (Psl.Argument.StringExpr "postgresql") -> Right AS.App.Db.PostgreSQL
-    Just (Psl.Argument.StringExpr "sqlite") -> Right AS.App.Db.SQLite
+    Just (Psl.Argument.StringExpr provider) -> mapProvider provider
     Just anyOtherExpression -> Left $ UnsupportedDbSystem $ generateExpression anyOtherExpression
     Nothing -> Left MissingDbSystem
+  where
+    mapProvider provider
```
Martinsos, line 57: I don't thikn you gained much by extracting this like this -> I have to jump here to read this, and also you have duplciation of a negative case `Left $ ...`. Not terrible, but if avoidable why not.

I would probably instead go for:
```hs
  case getDbProviderFromPrismaSchema prismaSchema of
    Just (Psl.Argument.StringExpr provider)
      | provider == AS.Db.providerToString AS.Db.PostgreSQL -> Right AS.App.Db.PostgreSQL
    Just (Psl.Argument.StringExpr provider)
      | provider == AS.Db.providerToString AS.Db.SQLite -> Right AS.App.Db.SQLite
    Just anyOtherExpression -> Left $ UnsupportedDbSystem $ generateExpression anyOtherExpression
    Nothing -> Left MissingDbSystem
```

Ok this is also not super perfect, it is a bit much, but I quite like it that we don't have duplication of that negative case, that is the main reason for me.

waspc/test/Psl/Parser/ConfigBlockTest.hs
```diff
@@ -25,7 +26,7 @@ spec_parsePslConfigBlock = do
             Psl.ConfigBlock.ConfigBlock
               Psl.ConfigBlock.Datasource
               "db"
-              [ Psl.ConfigBlock.KeyValuePair "provider" $ Psl.Argument.StringExpr "postgresql",
+              [ Psl.ConfigBlock.KeyValuePair "provider" $ Psl.Argument.StringExpr (AS.Db.providerToString AS.Db.PostgreSQL),
```
Martinsos, line 29: Similar argument here as far above -> if everything else here is hardcoded, like `db`, and `env`, and `url`, and simlar, then you can also hardcode these values, doesn't make much sense to have them parametrized and the rest not. It is ok, there are tests of Prisma parser, we can assume prisma syntax here.

waspc/test/Psl/Parser/SchemaTest.hs
```diff
@@ -110,7 +111,7 @@ spec_parsePslSchema = do
                 Psl.ConfigBlock.ConfigBlock
                   Psl.ConfigBlock.Datasource
                   "db"
-                  [ Psl.ConfigBlock.KeyValuePair "provider" $ Psl.Argument.StringExpr "postgresql",
+                  [ Psl.ConfigBlock.KeyValuePair "provider" $ Psl.Argument.StringExpr (AS.Db.providerToString AS.Db.PostgreSQL),
```
Martinsos, line 114: I would also leave it hardcoded here.

Martinsos (changes requested):

waspc/e2e-test/ShellCommands.hs
```diff
@@ -78,7 +79,7 @@ createFile content relDirFp filename = return $ combineShellCommands [createPare
 
 setDbToPSQL :: ShellCommandBuilder ShellCommand
 -- Change DB to postgres by adding string at specific line so it still parses.
-setDbToPSQL = replaceLineInFile "schema.prisma" 2 "  provider = \"postgresql\""
+setDbToPSQL = replaceLineInFile "schema.prisma" 2 ("  provider = \"" ++ AS.Db.providerToString AS.Db.PostgreSQL ++ "\"")
```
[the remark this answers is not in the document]
FranjoMindek: It's my bad. As wasp currently heavily leans on Prisma, I've automatically mapped in my head as that I would use the provider for Prisma stuff too. It makes sense that we want to be separated, especially with future plans. I'll revert it.

waspc/src/Wasp/AppSpec/App/Db.hs
```diff
@@ -20,3 +21,7 @@ data Db = Db
 
 data DbSystem = PostgreSQL | SQLite
   deriving (Show, Eq, Data, Generic, FromJSON)
+
+providerToString :: DbSystem -> String
```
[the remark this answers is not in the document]
FranjoMindek: I've just copied a pattern I've seen on google. I'll make it more explicit in function name on how to use this function. Now the question what would the right name for it be? `providerToId`, `providerToKey`? Maybe its better to use "db" word rather than "provider"? e.g. `dbToKey` or `dbToId`? I've copied the provder part from Prisma, but we don't necessarily need to follow that.

waspc/src/Wasp/Project/Db.hs
FranjoMindek: I suppose then `Just (Psl.Argument.StringExpr provider)` is not needed twice?

```hs
  case getDbProviderFromPrismaSchema prismaSchema of
    Just (Psl.Argument.StringExpr provider)
      | provider == Psl.Db.dbToPrismaProviderKey AS.Db.PostgreSQL -> Right AS.Db.PostgreSQL
      | provider == Psl.Db.dbToPrismaProviderKey AS.Db.SQLite -> Right AS.Db.SQLite
    Just anyOtherExpression -> Left $ UnsupportedDbSystem $ generateExpression anyOtherExpression
    Nothing -> Left MissingDbSystem

```

FranjoMindek:
I've tried a new approach:

- Moved db to prisma provider id mapping to Pls as `dbToPrismaProviderId`
- Prisma specific code just hardcodes prisma stuff
- Wasp code using prisma uses `dbToPrismaProviderId` to get Prisma provider Ids

Reminder to myself: add files manually to cabal to stop yourself from debugging 😭

waspc/e2e-test/ShellCommands.hs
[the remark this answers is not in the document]
Martinsos: Yup, it important to recognize what is Prisma specific, and what is Wasp/AppSpec specific, and what relies on what. Sounds good!

waspc/src/Wasp/AppSpec/App/Db.hs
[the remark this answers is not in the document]
Martinsos: > I've just copied a pattern I've seen on google.

Well, that doesn't sound like the best programming practice :D. Not the copying part per se, but the "just" part in it!
Specifically, what kind of pattern are you talking about -> merely turning something into a String?

Right name -> that is the trick, it depends on what it is used for, what it really is, the design. You don't really understand what it really is, which is why you ended up with such a generic name and why it all together caused issues down the road.

`dbToPrismaProviderId/Key` -> doesn't sound good to me, sounds like you are converting a database to prisma provider id or key. I usually advice to first describe in a sentence or two what the function is doing, as if explaining to a colleague, and then take that as a starting point and whittle it down till you end up with a nice name.

p.s. might be useful read: https://wasp.sh/blog/2023/10/12/on-importance-of-naming-in-programming .

waspc/src/Wasp/Project/Db.hs
Martinsos: Oh yes perfect that is even better!

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -165,7 +166,7 @@ validateOnlyEmailOrUsernameAndPasswordAuthIsUsed spec =
 validateDbIsPostgresIfPgBossUsed :: AppSpec -> [ValidationError]
 validateDbIsPostgresIfPgBossUsed spec =
   [ GenericValidationError
-      "The database provider in the schema.prisma file must be \"postgresql\" since there are jobs with executor set to PgBoss."
+      ("The database provider in the schema.prisma file must be \"" ++ Psl.Db.dbToPrismaProviderId AS.Db.PostgreSQL ++ "\" since there are jobs with executor set to PgBoss.")
```
Martinsos, line 169: It still makes no sense here to take something that is AppSpec concept (AS.Db.PostgreSQL) and then map it to Prisma concept.

We are in this error message specifically telling them that in schema.prisma file, provider must be "postgresql". This has nothing to do with Wasp's AppSpec and how we describe the database there.

What instead you should do is define something like `pslPostgresqlKeyword :: String` or `pslPostgresqlIdentifier :: String` or whatever is the correct term in Prisma's world, and have it have value of `"postgresql"`, and use that here, that is it. You would define that somewhere in Wasp.Prisma.PSL most likely (I didn't check the exact name of modules but I am guessing it goes something like this).

waspc/src/Wasp/Generator/DbGenerator.hs
```diff
@@ -63,11 +64,11 @@ genPrismaSchema ::
...
+    AS.Db.PostgreSQL -> return (Pls.Db.dbToPrismaProviderId AS.Db.PostgreSQL)
```
Martinsos, line 67: Ha yes, even here this is too much. You already cased on PostgreSQL -> you know what you need now here is prisma's PSL keyword for postgresql. So you could just do something like `Wasp.Prisma.PSL.postgresqlIdentifier` or Keyword or whatever is better word and that is it. Same below for SQlite.

If you were instead in a situaiton where you are not doing the casing here, and you want to do `Pls.Db.dbToPrismaProviderId dbSystem`, then it would make sense to have `Pls.Db.dbToPrismaProviderId` (although I would name it better), but we don't actually need taht anywehere from what I see.

waspc/src/Wasp/Psl/Db.hs
```diff
@@ -0,0 +1,10 @@
+module Wasp.Psl.Db
+  ( dbToPrismaProviderId,
+  )
+where
+
+import qualified Wasp.AppSpec.App.Db as AS.Db
```
Martinsos, line 6: This is also indicator that something is a bit weird, importing AppSpec in Wasp.Psl module tree. We kind of want Wasp.Psl to be a standalone concept, ideally not reliant on the rest of Wasp's sepcifics, if possible. So if somebody had to rely on somebody, it would probably be better to have AppSpec rely on Prisma, than the opposite direction.

Martinsos (changes requested):
@FranjoMindek step closer, but conceptually it is still not right!
Looking at the code, you don't really need mapping from AppSpec dbSystem to Prisma's PSL db provider -> instead you just need a way to get the exact PSL keywords/identifiers/literals for PSL db providers like postgresql and sqlite. Which is simpler.

And evne if you wanted to create a function that maps from AppSpec db system to PSL db provider, you would want to put that in AppSpec most likely, maybe in the Generator, almost certainly not in PSL.

I think the main problem is that you are still lacking the bigger picture of the design here, which will come with time. But I do advise possibly taking more time to study the design of the parts you are touching before jumping into coding. These issues are here for you to learn in the process about the system you are modifying, so being thorough while doing them is more important then being fast!

Martinsos:
> @Martinsos I've reverted everything to using constants. What is funny that that was my first approach. But I've tried to make it more "extensible" in a way. But now that I think about it's not required of Prisma to support everything Wasp does. I've also tried to copy what `AppSpec.App.Db.DbSystem` does but that also felt weird in context of constants.
> 
> For the constants file I've put them in `Psl.Db` file. Technically Prisma calls them DB providers, but I'm not sure if I want to name it `Psl.DbProviders` or `Psl.Db`. In a way, I feel like calling it `Psl.Db` makes it easier to find since we have the same-named file in AppSpec, and I think it's enough context for people to know it means Prisma DB providers. What would you prefer in naming? Is following project patterns for price of not having fully clean names okay here?

I took a look at the latest changes -> this is it! This is exactly what was needed to solve the issue this PR was meant to solve. You could try to go for a fancier / more complex design, e.g. having a mapping from AppSpec's "db system" to Prisma's "db provider", but than that should also be approached with attention, and still wouldn't make sense to use it in the places that you modified in this PR, as they had nothign to do with Wasp specifically.

You are asking a good question there, and that is usually not an easy choice: do you go with "better" naming or do you stick with convention for naming that arleady exists in the domain (in this case domain is Prisma / PSL). I would say it usually pays off to stick to the domain, because otherwise you have one more layer of mapping your concepts to the mapping of that domain, but on the other hand if their naming is really bad, it can make sense to bend it a bit.

In this case, we are looking at
```psl
datasource db {
  provider = "postgresql"
  url      = env("DATABASE_URL")
}
```

So we have datasources, then one of them is db, and db has provider. So you could in theory go as far as Psl.Datasource.Db.Provider, but that is probably too granular, and it is just fine to go with Psl.Datasource.Db and then you can specify in it anything related to db datasource, be it provider, or something related to other properties. You abbreviated it a bit further, you went for `Psl.Db` and kicked out the Datasource part -> that might make sense, it depends on the rest of Prisma and if Psl.Db will with time get in "conflict" with a more general concept of Db in Prisma, in which case you will probably want to push it further into Psl.Datasource.Db, but at the moment, since we don't know more, I would say going with `Psl.Db` can be ok, and we will refactor if needed.

The last thing about naming on my side is the name of these constants. I told you to name them `keyword` or `identifier` or `literal` or whatever makes the most sense -> I didn't put enough thought into it at that moment but knew some thought needs to go into it. How did you choose to go with `keyword`?
I am looking now into it, and I forgot that it is `provider = "postgresql"`, not `provider = postgresql`. That certainly eliminates `identifier`, but it also kind of eliminates `keyword` -> it all depends how you define it, what the keyword is, you could argue it is still a keyword hm. But in compiler talk, it is "string literal". So probably the most correct name would be `dbProviderPostgresqlStringLiteral`. Maybe `dbProviderPostgresqlKeyword`, if we say it conceptually goes beyond a mere string and it is also used maybe in some other places ... .
But to be on the safe side, I would probably go with `dbProviderPostgresqlStringLiteral`.

The way you went, `pslPostgresqlKeyword` makes sense if this is a way through the whole PSL to reference Postgresql wherever that is needed. But I am not sure that is the case -> is it referenced anyway except for in database datasource provider? If it is, is it really the exact same keyword/value? Which is why maybe going with `dbProviderPostgresqlStringLiteral` is easier for now, since you know that is true for sure, and we can generalize it in the future if it will be needed.

This is a bit of nitpicking at the end, it is not a bit thing this name, but I think it does make some sense.

Martinsos (commented):
Almost there, check the comment above!

FranjoMindek:
Thanks for the detailed response. Having your train of thought written out really helps out. I think it's great to nitpick, especially while I'm still adjusting to Wasps style of things.

waspc/ChangeLog.md
```diff
@@ -7,6 +7,7 @@
 - Custom values for `theme-color` are now supported ([#2546](https://github.com/wasp-lang/wasp/pull/2546) by @andrsdt).
 - Increased the minimum Node version to 20.0.0 ( [#2537](https://github.com/wasp-lang/wasp/pull/2537) )
 - Invalid CLI commands now properly return non-zero exit code ( [#2522](https://github.com/wasp-lang/wasp/pull/2552) )
+- Prisma's datasource database provider string literals are now variablized ( [#2548](https://github.com/wasp-lang/wasp/pull/2548) )
```
Martinsos, line 10: This is internal refactor so it doesn't neccesary even have to go to Changelog. Although it doesn't hurt I think! Shows we are working on stuff. But maybe also clutters it a bit.

waspc/src/Wasp/Psl/Db.hs
```diff
@@ -0,0 +1,11 @@
+module Wasp.Psl.Db
+  ( dbProviderPostgresqlStringLiteral,
+    dbProviderSqliteStringLiteral,
+  )
+where
+
+dbProviderPostgresqlStringLiteral :: String
```
Martinsos, line 7: Ha yes its ugly name, but it is correct! Good!

Martinsos (approved):
LGTM!

PR #2549, opened 2025-03-13
Check npm version when checking node version

### LINKED ISSUE #2113: Improve error message when correct version of `npm` isn't in `PATH`
If a user starts a `wasp` command and `npm` isn't in `PATH`, this is the error they get:
```
[  Wasp  ] Starting npm installwasp-bin: npm: streamingProcess: exec: invalid argument (Bad file descriptor)
```

We should check if `npm` is in `PATH` and give our users a meaningful message if we don't find it (like we do when we don't find `node`) (or if it is too old).

Thanks @Rotemoyal for letting us know: https://github.com/wasp-lang/wasp/issues/1980#issuecomment-2183980522

FranjoMindek:
Adds minimal npm version check when we check for node version
Builds upon node version code, makes it usable for both npm and node. 

Fixes #2113


FranjoMindek:
Should probably bump the minimum version to 9.6 as this checks for >=. 


Image preview on `npm@9.4`:
image

waspc/src/Wasp/Node/Version.hs
```diff
@@ -42,76 +47,79 @@ type ErrorMessage = String
...
+-- | Base function to check if the user's installed version meets Wasp's version requirement.
+checkVersion :: String -> SV.Version -> (String -> Either P.ParseError SV.Version) -> IO VersionCheckResult
+checkVersion command oldestSupportedVersion parseVersion = do
+  result <- getUserVersion command parseVersion
+  return $ case result of
+    Left errorMsg -> VersionCheckFail errorMsg
+    Right userVersion ->
+      if SV.isVersionInRange userVersion $ SV.Range [SV.gte oldestSupportedVersion]
+        then VersionCheckSuccess
+        else VersionCheckFail $ makeVersionMismatchMessage command userVersion oldestSupportedVersion
```
infomiho, line 66: I think this function should get the oldest support version, the current version and decide if the version is okay. I think it shouldn't concern itself with how the version is parsed, that should be the worry of the caller IMHO.

For example, here's a suggestion how to implement something along those lines:
```haskell
checkInstalledVersionNewerThanOldestSupported :: String -> SV.Version -> IO (Either ErrorMessage SV.Version) -> IO VersionCheckResult
checkInstalledVersionNewerThanOldestSupported commandName oldestSupportedVersion getInstalledVersion = do
  result <- getInstalledVersion
  return $ case result of
    Left errorMsg -> VersionCheckFail errorMsg
    Right userVersion ->
      if SV.isVersionInRange userVersion $ SV.Range [SV.gte oldestSupportedVersion]
        then VersionCheckSuccess
        else VersionCheckFail $ makeVersionMismatchMessage commandName userVersion oldestSupportedVersion
```
1. I gave a longer name so the comment above is not needed (maybe not the best name, but you get the point)
2. The first two arguments are "context arguments" and the last one is the thing that we are "processing", the "object" of our operation
  - This is a common practice in Haskell, putting the thing we are processing as the last argument e.g. `map fn list`
  - This enables us to partially apply the `checkInstalledVersionNewerThanOldestSupported` fn e.g. `checkInstalledVersionNewerThanOldestSupported "node" someVersion` and use it multiple times 
4. We get the command name + the oldest supported version as context
5. The third argument is a function that will give us the current version, it's completely abstracted way (we don't know about the version parser anymore)

The way we call it is like this:
```haskell
getAndCheckUserNodeVersion :: IO VersionCheckResult
getAndCheckUserNodeVersion = checkInstalledVersionNewerThanOldestSupported "node" oldestWaspSupportedNodeVersion $ parseVersionFromCommandOutput "node" ["--version"]
```
You'll notice that the `parseVersionFromCommandOutput "node" ["--version"]` is a bit different than what you wrote, I'll explain it in a comment below.

waspc/src/Wasp/Node/Version.hs
```diff
@@ -42,76 +47,79 @@ type ErrorMessage = String
...
+
+-- | Base function to get the installed version of a command, if any is installed, and returns it.
```
infomiho, line 68: I think the comment above is redundant if we name the function clearly and I think the logic of parsing the version can be inlined since this is the only place where it's used (this also gives us the chance to make the function hide a bit of the implementation details). 

For example, I've tried writing it like this:
```haskell
parseVersionFromCommandOutput :: String -> [String] -> IO (Either ErrorMessage SV.Version)
parseVersionFromCommandOutput command args = do
  result <-
    (Right <$> P.readProcessWithExitCode command args "")
      `catchIOError` ( \e ->
                         if isDoesNotExistError e
                           then return $ Left $ commandNotFoundMessage command
                           else return $ Left $ makeUnknownErrorMessage command e
                     )
  return $ case result of
    Left procErr ->
      Left
        ( unlines
            [ "Running `" ++ commandWithArgs ++ "` failed.",
              indent 2 procErr,
              "Make sure you have `" ++ command ++ "` installed and in your PATH."
            ]
        )
    Right (ExitFailure code, _, stderr) ->
      Left
        ( unlines
            [ "Running `" ++ commandWithArgs ++ "` failed (exit code " ++ show code ++ "):",
              indent 2 stderr
            ]
        )
    Right (ExitSuccess, stdout, _) ->
      parseVersionFromOutput stdout
        & left
          ( \e ->
              "Wasp failed to parse `" ++ command ++ "` version provided by `" ++ commandWithArgs ++ "`.\n"
                <> (show e <> "\n")
                <> "This is most likely a bug in Wasp, please file an issue."
          )
  where
    commandWithArgs = unwords $ command : args

    parseVersionFromOutput :: String -> Either P.ParseError SV.Version
    parseVersionFromOutput = P.parse versionParser ""
      where
        versionParser = skipAnyCharTillMatch SV.versionParser
        skipAnyCharTillMatch p = P.manyTill P.anyChar (P.lookAhead $ P.try p) >> p
```

1. I didn't want to assume that the version will be checked by `{cmd} --version`, make the caller give you all the input you need to determine the version
2. I've named it `parseVersionFromCommandOutput` to make it clear it will parse the version from the command output
3. Inlined the `parseVersionFromOutput` function since it's only relevant to this function

waspc/src/Wasp/Job/Process.hs
```diff
@@ -95,10 +95,13 @@ runNodeCommandAsJob :: Path' Abs (Dir a) -> String -> [String] -> J.JobType -> J
...
-  NodeVersion.getAndCheckUserNodeVersion >>= \case
-    NodeVersion.VersionCheckFail errorMsg -> exitWithError (ExitFailure 1) (T.pack errorMsg)
-    NodeVersion.VersionCheckSuccess -> do
+runNodeCommandAsJobWithExtraEnv extraEnvVars fromDir command args jobType chan = do
+  nodeCheck <- NodeVersion.getAndCheckUserNodeVersion
+  npmCheck <- NodeVersion.getAndCheckUserNPMVersion
+  case (nodeCheck, npmCheck) of
+    (NodeVersion.VersionCheckFail errorMsg, _) -> exitWithError (ExitFailure 1) (T.pack errorMsg)
+    (_, NodeVersion.VersionCheckFail errorMsg) -> exitWithError (ExitFailure 1) (T.pack errorMsg)
+    (NodeVersion.VersionCheckSuccess, NodeVersion.VersionCheckSuccess) -> do
```
infomiho, line 104: Given that in all of the places where we check the Node.js version, we also check the `npm` version, do we maybe want to make this into a single function that says `getAndCheckNodeNpmVersion` and then it either returns `VersionCheckFail` or `VersionCheckSuccess`. 

This ensures that we _always_ check both and reduces the noise in places where we check for the version.

waspc/src/Wasp/Node/Version.hs
```diff
@@ -1,9 +1,11 @@
...
-  ( getAndCheckUserNodeVersion,
-    VersionCheckResult (..),
+  ( VersionCheckResult (..),
+    getAndCheckUserNodeVersion,
+    getAndCheckUserNPMVersion,
+    oldestWaspSupportedNPMVersion,
     oldestWaspSupportedNodeVersion,
-    parseNodeVersionOutput,
     isRangeInWaspSupportedRange,
+    checkVersion,
```
infomiho, line 8: Usually we try to expose from modules as little as possible. Usually we have some "public API" that we expose, but keep the helper functions hidden since they are not useful on their own. If you believe this function will be useful in other places in the codebase, it's okay to export it. 

If we come up with a generic utility function e.g. `parseVersionFromCommandOutput` like I mentioned below, maybe it makes sense to extract this into some more generic module, unrelated to Node.js e.g. `Wasp.Util.CommandVersion` or `Wasp.SemanticVersion.Process` (indicating it's got to something with commands and processes).

infomiho (commented):
Good job on figuring this one out, left some comments.

FranjoMindek:
I've tidied it up a bit, so that the flow is easier to read, at least to me.
Other than that I've reduced number of exports, made it export npm/node check in a single function, and other suggested  changes.

waspc/src/Wasp/Node/Version.hs
```diff
@@ -39,79 +40,102 @@ data VersionCheckResult
...
 
--- | Gets the user's installed node version, if any is installed, and checks that it
--- meets Wasp's version requirement.
+getAndCheckNodeAndNpmVersion :: IO VersionCheckResult
+getAndCheckNodeAndNpmVersion = do
+  nodeVersion <- getAndCheckUserNodeVersion
+  npmVersion <- getAndCheckUserNPMVersion
+  return $ case (nodeVersion, npmVersion) of
+    (VersionCheckSuccess, VersionCheckSuccess) -> VersionCheckSuccess
+    (VersionCheckFail nodeError, VersionCheckFail npmError) -> VersionCheckFail $ nodeError ++ "\n" ++ npmError
```
infomiho, line 49: I have a feeling that I don't want to report both errors at the same time, because it might confuse our users to some degree. 

If they don't have Node.js installed, 99% they don't have `npm` installed as well. So they'll always get the double error message in that case. While it's correct to say they don't have neither installed, the most important thing for them to address is installing Node.js - which will solve the `npm` install as well. 

So I'm leaning towards outputting the Node.js error _first_ and then only the `npm` error if there is no Node.js error. 

The usual case:
1. "Hey you don't have the correct Node.js version"
2. User installs the correct version of Node.js (and most of the time `npm` as well)
3. No more errors

Edge case:
1. Node.js version is fine but `npm` version isn't,  "Hey your `npm` version is wrong"
2. User updates their `npm`
3. No more errors

What do you think? Does that make any sense or am I overthinking here? 😄

infomiho (commented):
Another round of comments

waspc/src/Wasp/Node/Version.hs
FranjoMindek: I've actually had the same concerns here, so I'm glad you're pointing it out.

Only reason why I didn't do it is because of the edge case that that both node and npm can throw error at the same time. If that happens, I'd much rather fix both errors at the same time rather than getting new error after fixing old one.

But yeah, this is more of a preference thing for the project.

waspc/src/Wasp/Node/Version.hs
```diff
@@ -39,79 +40,102 @@ data VersionCheckResult
...
+    getNodeVersion = parseVersionFromCommandOutput "node" ["--version"]
+
+getAndCheckUserNPMVersion :: IO VersionCheckResult
+getAndCheckUserNPMVersion = checkNPMVersion getNPMVersion
+  where
+    checkNPMVersion = checkInstalledVersionNewerThanOldestSupported "npm" oldestWaspSupportedNPMVersion
+    getNPMVersion = parseVersionFromCommandOutput "npm" ["--version"]
+
+checkInstalledVersionNewerThanOldestSupported :: String -> SV.Version -> IO (Either ErrorMessage SV.Version) -> IO VersionCheckResult
+checkInstalledVersionNewerThanOldestSupported command oldestSupportedVersion getInstalledVersion = do
```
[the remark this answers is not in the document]
infomiho: @FranjoMindek FYI the output has `commandName` literal in it:

```
❌ --- [Error] Node/NPM requirement not met: --------------------------------------

Running `node --version` failed.
  `node --version` commandName not found!
Make sure you have `node` installed and in your PATH.

Running `npm --version` failed.
  `npm --version` commandName not found!
Make sure you have `npm` installed and in your PATH.
```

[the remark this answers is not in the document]
FranjoMindek: Replace all mistake. I'll fix it.

waspc/src/Wasp/Node/Version.hs
FranjoMindek: @Martinsos @sodic What do you think? I've asked around communities (with little success). Most answers trended towards show all errors, but most people thought about it in context of validation (e.g. forms). This is a bit different.

infomiho: Maybe @sodic and @cprecioso can chime in and we can have a consensus what we prefer as a team and go with that? 🤷 I'm fine with both solutions.

cprecioso: No strong opinion, small preference for separate errors.
Otherwise we can also add a note in the node error saying npm errors are probably related.

FranjoMindek: I'll do node check first then npm.
If no objections soon I'll consider this resolved. 👍

infomiho (approved):
🚢 it

waspc/src/Wasp/Node/Version.hs
```diff
@@ -39,79 +40,101 @@ data VersionCheckResult
...
+        else VersionCheckFail $ versionMismatchErrorMessage userVersion
+  where
+    versionMismatchErrorMessage :: SV.Version -> ErrorMessage
+    versionMismatchErrorMessage userVersion =
+      unlines
+        [ "Your " ++ commandName ++ " version does not meet Wasp's requirements! You are running " ++ commandName ++ " " <> show userVersion <> ".",
+          "Wasp requires " ++ commandName ++ " version " <> show oldestSupportedVersion <> " or higher."
+        ]
+
+parseVersionFromCommandOutput :: String -> [String] -> IO (Either ErrorMessage SV.Version)
```
Martinsos, line 81: A bit wrong name, it doesn't sound like you are also calling the command, while you are.

waspc/src/Wasp/Node/Version.hs
```diff
@@ -39,79 +40,101 @@ data VersionCheckResult
...
-    ]
+    commandWithArgs = unwords $ commandName : args
+
+    commandNotFoundErrorMessage :: ErrorMessage
+    commandNotFoundErrorMessage = "`" ++ commandWithArgs ++ "` command not found!"
+
+    unkownErrorErrorMessage :: IOError -> ErrorMessage
+    unkownErrorErrorMessage err =
+      unlines
+        [ "An unknown error occurred while trying to run `" ++ commandWithArgs ++ ":",
```
Martinsos, line 109:
```suggestion
        [ "An error occurred while trying to run `" ++ commandWithArgs ++ ":",
```

They don't need to know (or care) that it is unknown to us, at the end we do print it!

Aha I see this is old stuff. Maybe still let's improve it.

waspc/src/Wasp/Node/Version.hs
```diff
@@ -39,79 +40,101 @@ data VersionCheckResult
...
+          indent 2 $ show err
+        ]
+
+    runningCommandFailedErrorMessage :: String -> ErrorMessage
```
Martinsos, line 113: This sounds as if "running command" failed. In the sense that command was running, and then it failed. But we awnt to say that we failed to run it. So it should be `failedToRunCommandErrorMessage`.

waspc/src/Wasp/Node/Version.hs
```diff
@@ -39,79 +40,101 @@ data VersionCheckResult
...
+    runningCommandFailedErrorMessage processError =
+      unlines
+        [ "Running `" ++ commandWithArgs ++ "` failed.",
+          indent 2 processError,
+          "Make sure you have `" ++ commandName ++ "` installed and in your PATH."
+        ]
+
+    commandFailedErrorMessage :: Int -> String -> ErrorMessage
```
Martinsos, line 121: How is this really different than the function above? Sounds same from the error message hm. And functoin name doesn't really help. I guess here we have exit code? Let's name it then `commandFailedWithExitCodeErrorMessage`?

waspc/src/Wasp/Node/Version.hs
```diff
@@ -39,79 +40,101 @@ data VersionCheckResult
...
+          indent 2 commandError
+        ]
+
+    parseVersionFromOutput :: String -> Either P.ParseError SV.Version
+    parseVersionFromOutput = P.parse versionParser ""
+      where
+        versionParser = skipAnyCharTillMatch SV.versionParser
+        skipAnyCharTillMatch p = P.manyTill P.anyChar (P.lookAhead $ P.try p) >> p
+
+    parseVersionFailedErrorMessage :: P.ParseError -> ErrorMessage
```
Martinsos, line 134: I again thought this was about something called "parse version" failing.
I would rename it to "failedToParseVersionErrorMessage".

Martinsos: Or at least `parsingVersionFailedErrorMessage`.

waspc/src/Wasp/Node/Version.hs
Martinsos, line 1: While it is fine, this file also feels a bit like it is overly engineered in a way. Logic is broken down into such small bits that I found it a bit hard to follow. Especially the error messages. It might make sense though, but I wonder what would it look like if we tried to go back a bit and make it a bit more cohesive.

Martinsos: Ok I gave it a try a bit and it is not so easy or straightforward, I give up, let's leave it as it is :D.

waspc/src/Wasp/Node/Version.hs
Martinsos: Actually might be fine, it is not so easy to come with a better name, and it is relatively clear what is happening if you look at the arguments. So I am good with the current name if you don't have something clearly better.

waspc/ChangeLog.md
```diff
@@ -10,6 +10,7 @@
 - Prisma's datasource database provider string literals are now variablized ( [#2548](https://github.com/wasp-lang/wasp/pull/2548) )
 - Better validation of `package.json` ( [#2554](https://github.com/wasp-lang/wasp/pull/2554) )
 - Wasp properly checks NPM version now ( [#2549](https://github.com/wasp-lang/wasp/pull/2549) )
+- Typescript config now accept `HttpRoute` object insteada of an array ( [#2580](https://github.com/wasp-lang/wasp/pull/2580) )
```
Martinsos, line 13:
```suggestion
- Typescript config now accept `HttpRoute` object instead of an array ( [#2580](https://github.com/wasp-lang/wasp/pull/2580) )
```

FranjoMindek: I should've really double checked that before merging. My bad.

FranjoMindek:
@sodic I've made some of the changes we've talked about. You can check it now.
Only question I have is how do you look at the error messages? Is it better to have them as helper function inside `where`, or to keep them outside of the original function?

waspc/src/Wasp/Node/Version.hs
sodic: @infomiho Great comment!

@FranjoMindek If you want to read more on Haskell argument order, check [the Haskell wiki](https://wiki.haskell.org/Parameter_order).

sodic (changes requested):
Excellent work!

I see you're getting the hang of Haskell.

I left some comments, don't hesitate to ping me for further questions :)

waspc/src/Wasp/Node/Version.hs
```diff
@@ -39,79 +40,108 @@ data VersionCheckResult
...
+getAndCheckUserNpmVersion = getAndCheckUserToolVersion "npm" ["--version"] oldestWaspSupportedNpmVersion
+
+getAndCheckUserToolVersion :: String -> [String] -> SV.Version -> IO VersionCheckResult
+getAndCheckUserToolVersion commandName commandArgs oldestSupportedToolVersion = checkToolVersion getToolVersion
+  where
+    checkToolVersion = checkInstalledVersionIsNewerThanOldestSupported commandName oldestSupportedToolVersion
+    getToolVersion = parseVersionFromCommandOutput commandName commandArgs <$> runCommand commandName commandArgs
+
+checkInstalledVersionIsNewerThanOldestSupported :: String -> SV.Version -> IO (Either ErrorMessage SV.Version) -> IO VersionCheckResult
+checkInstalledVersionIsNewerThanOldestSupported commandName oldestSupportedVersion getInstalledVersion = do
```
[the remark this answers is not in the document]
FranjoMindek: Yep, mostly leftovers from past refactors. e.g. before `>>=` or while it still returned `IO`

FranjoMindek:
I have questions regarding current state of PR:
- `Command` as an tuple - I think it might sense to alias it and group it together, but what to do with functions that only user command name? still send the whole tuple?
- `showFullCommand` - what would be the best way to handle that, we use this functionality in almost all messages, how do we name stuff like this in Haskell
- `parseVersionFromCommandOutput` - do you think that the current state which handles whole command result object is okay? technically it does more than parsing. 
- is `checkUserToolVersion` going in right direction

waspc/ChangeLog.md
```diff
@@ -9,7 +9,8 @@
 - Invalid CLI commands now properly return non-zero exit code ( [#2522](https://github.com/wasp-lang/wasp/pull/2552) )
 - Prisma's datasource database provider string literals are now variablized ( [#2548](https://github.com/wasp-lang/wasp/pull/2548) )
 - Better validation of `package.json` ( [#2554](https://github.com/wasp-lang/wasp/pull/2554) )
-- Typescript config now accept `HttpRoute` object insteada of an array ( [#2580](https://github.com/wasp-lang/wasp/pull/2580) )
+- Wasp now validates your NPM version satisfies its requirement ( [#2549](https://github.com/wasp-lang/wasp/pull/2549) )
```
sodic, line 12: Extra space after parentheses.

waspc/src/Wasp/Node/Version.hs
```diff
@@ -39,79 +40,108 @@ data VersionCheckResult
...
+  getInstalledVersion >>= \case
     Left errorMsg -> return $ VersionCheckFail errorMsg
-    Right userNodeVersion ->
+    Right userVersion ->
```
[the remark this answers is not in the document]
sodic: `userVersion` was a better name than version IMO. But this can stay too if you like it. Resolve when happy

waspc/src/Wasp/Node/Version.hs
```diff
@@ -39,79 +40,108 @@ data VersionCheckResult
...
-    Right (ExitFailure code, _, stderr) ->
-      Left
-        ( unlines
-            [ "Running `node --version` failed (exit code " ++ show code ++ "):",
-              indent 2 stderr
-            ]
-        )
+          else VersionCheckFail $ versionMismatchErrorMessage userVersion
+  where
+    versionMismatchErrorMessage :: SV.Version -> ErrorMessage
```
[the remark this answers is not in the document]
sodic: I only now see this.

I recommend naming functions with verbs, and non-function values with nouns.
I know Haskell's STL doesn't follow this (e.g., `words`, `head`), I've always found it annoying :)

Still, it's a useful naming heuristic. No need change if you don't like the suggestion. Resolve when happy.

FranjoMindek:
Made further changes, one thing I'm unsure of is of return type for following function:
```hs
checkUserVersionIsSupported :: String -> SV.Version -> SV.Version -> Either ErrorMessage ()
```
I force it be of Either return type so I can easily chain it with others.
I like that it doesn't return VersionCheckResult and only `checkUserToolVersion` does that.
But it still feels a bit off? Maybe this is okay in Haskell.

FranjoMindek:
Final final final version of npm check.

waspc/src/Wasp/Node/Version.hs
```diff
@@ -39,79 +40,85 @@ data VersionCheckResult
...
+checkUserNpmVersion = checkUserToolVersion "npm" ["--version"] oldestWaspSupportedNpmVersion
+
+checkUserToolVersion :: String -> [String] -> SV.Version -> IO VersionCheckResult
+checkUserToolVersion commandName commandArgs oldestSupportedToolVersion = do
+  userVersion <- getToolVersionFromCommandOutput commandName commandArgs
+  return $ case userVersion of
+    Left errorMsg -> VersionCheckFail errorMsg
+    Right version
+      | version >= oldestSupportedToolVersion -> VersionCheckSuccess
+      | otherwise -> VersionCheckFail $ makeVersionMismatchErrorMessage version
```
sodic, line 65: Naming suggestion:
  - `userVersion` -> `userVersionOrError`
  - `version` -> `userVersion`

Whenever two variables can switch names and still be correct, chances are we're abusing the English languge to avoid dealing with naming :)

More examples of this:
```ts
const directory = foo()
const dir = bar(directory)

// ...

const doesExist = exists()
```

waspc/src/Wasp/Node/Version.hs
```diff
@@ -39,79 +40,85 @@ data VersionCheckResult
...
+    wrapCommandIOErrorMessage innerErr =
+      unlines
+        [ "Running `" ++ fullCommand ++ "` failed.",
+          indent 2 innerErr,
+          "Make sure you have `" ++ commandName ++ "` installed and in your PATH."
+        ]
+
+    wrapCommandExitCodeErrorMessage exitCode commandErr =
+      unlines
+        [ "Running `" ++ fullCommand ++ "` failed (exit code " ++ show exitCode ++ "):",
```
sodic, line 103: Does wasp usually use backticks when quoting code/commands in its output? I'm not sure.

Please check the rest of our messages to ensure its consistent.

sodic (commented):
Great stuff!

I'm approving. Before you merge, ensure you've tested all the paths and are getting the correct/expected messages. It's often difficult to visualize the message without actually seeing it. Whenever I work on something like this, I always ended up changing the code after seeing the message (an extra space here, a missing full stop there, etc.)

I also recommend trying to write some unit tests for the code. That's a great way to evaluate the architecture. If it's difficult to write unit tests for testing the core functionality, chances are the code is too coupled.

sodic (approved):
Forgot to approve :upside_down_face:

waspc/src/Wasp/Node/Version.hs
```diff
@@ -39,79 +40,85 @@ data VersionCheckResult
...
+  return $ case (nodeResult, npmResult) of
+    (VersionCheckSuccess, VersionCheckSuccess) -> VersionCheckSuccess
+    (VersionCheckFail nodeError, _) -> VersionCheckFail nodeError
+    (_, VersionCheckFail npmError) -> VersionCheckFail npmError
+
+checkUserNodeVersion :: IO VersionCheckResult
+checkUserNodeVersion = checkUserToolVersion "node" ["--version"] oldestWaspSupportedNodeVersion
+
+checkUserNpmVersion :: IO VersionCheckResult
+checkUserNpmVersion = checkUserToolVersion "npm" ["--version"] oldestWaspSupportedNpmVersion
```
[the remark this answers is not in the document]
FranjoMindek: True, they weren't because they were exported before. Since we want to always test them together this is no longer the case.

FranjoMindek:
Tried to no-node error message which we discussed, I think I like the additional

```
❌ --- [Error] Node/NPM requirement not met: --------------------------------------

Running `node --version` failed.
  `node --version` command not found!
Make sure you have `node` installed and in your PATH.
```
VS
```

PR #2552, opened 2025-03-14
Invalid CLI commands should return non-zero exit code

### LINKED ISSUE #2491: Invalid `wasp` commands return a zero exit code
Running `wasp asdf` or `wasp db migrate-foo` should:
1. Print help on how to use the command.
3. Return a non-zero exit code to communicate failure to the shell

Their current exit code is 0:
```
$ wasp gmao
USAGE
  wasp <command> [command-args]
 ...
$ echo $?
0
```

FranjoMindek:
Add exitFailure to invalid wasp-cli usages

As far as I can see it works.

```bash
➜  waspc git:(bf/franjo/wasp-cli-exit-code) ✗ cabal run wasp-cli asdasdasdasdasd
USAGE
  wasp <command> [command-args]
...

➜  waspc git:(bf/franjo/wasp-cli-exit-code) ✗ echo $?
1
➜  waspc git:(bf/franjo/wasp-cli-exit-code) ✗ cabal run wasp-cli db migrate
USAGE
  wasp db <command> [command-args]
...

➜  waspc git:(bf/franjo/wasp-cli-exit-code) ✗ echo $?
1
```

Side note: 
I think `Main.hs` is a bit messy currently with the file doing too much work.
There is this note:
```hs
-- TODO: maybe extract to a separate module, e.g. DbCli.hs?
dbCli :: [String] -> IO ()
```
Same should probably be done for AI part.

Fixes #2491

### DIFF
--- a/waspc/ChangeLog.md
+++ b/waspc/ChangeLog.md
@@ -6,6 +6,7 @@
 
 - Custom values for `theme-color` are now supported ([#2546](https://github.com/wasp-lang/wasp/pull/2546) by @andrsdt).
 - Increased the minimum Node version to 20.0.0 ( [#2537](https://github.com/wasp-lang/wasp/pull/2537) )
+- Invalid CLI commands now properly return non-zero exit code ( [#2522](https://github.com/wasp-lang/wasp/pull/2552) )
 
 ### Bug fixes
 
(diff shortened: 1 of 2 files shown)


Martinsos (approved):
LGTM!
Tiny hint: instead of doing
```hs
do
  printUsage
  exitFailure
```
you can do it in one line
```hs
printUsage >> exitFailure
```

You don't have to do, if you find the first one more readable, just wanted to share the tip.

FranjoMindek:
Thanks for the tip.
Still learning haskell, I'll change it.

Martinsos:
> Thanks for the tip. Still learning haskell, I'll change it.

I know which is why I shared it, but also feel free to keep it as it was originally if you like that better / find it more readable!

Issue #2317, opened 2024-10-03
Add extra validation rules for deps in `package.json`


infomiho:
In #2297 we added validation for userland config files e.g. `package.json` and `tsconfig.json`. We validated only package versions that were seen causing issues i.e. Prisma and React Router. 

There is no reason not to validate all the required packages:
- react
- @types/react
- vite
etc.

Basically, add some extra rules that would prevent users from installing e.g. React 19 when it comes out before Wasp migrates to the new version fully.

PR #2554, opened 2025-03-17
Extra validation for dependencies in `package.json`

FranjoMindek:
Adds additional validation for `package.json` file.

I'm not sure what packages exactly do we want to fit here. 
I've made `typescript` and type packages optional because I assumed we want su support non-typescript projects.

I would rather avoid `HasExactVersionIfListed` from old solution since packages should always exist in either `dependencies` or `devDependecies`, but not both.

Also not sure if this would be better approach, interested in Haskell best practices side of things here:

```hs
data PackageDependecyType = Runtime | Development
  deriving (Eq, Show)

data PackageRequirement = Required | Optional
  deriving (Eq, Show)

data PackageValidation = PackageValidation
  { dependencyType :: PackageDependecyType,
    requirement :: PackageRequirement
  } deriving (Eq, Show)
```

Fixes: https://github.com/wasp-lang/wasp/issues/2317


FranjoMindek:
image

Example error when `todoApp` has no `vite` in `devDependencies`, core logic wasn't changed so it works okay.

waspc/src/Wasp/Generator/ExternalConfig/PackageJson.hs
```diff
@@ -5,52 +5,76 @@ where
...
-    validate = validateDep packageJson
+    validate = validatePackageJsonDependency packageJson
 
-data PackageValidationType = IsListedWithExactVersion | IsListedAsDevWithExactVersion | HasExactVersionIfListed
-
-validateDep :: P.PackageJson -> (P.PackageName, P.PackageVersion) -> PackageValidationType -> [String]
-validateDep packageJson (packageName, expectedPackageVersion) = \case
-  IsListedWithExactVersion -> checkDeps [P.dependencies packageJson] [requiredPackageMessage "dependencies"]
-  IsListedAsDevWithExactVersion -> checkDeps [P.devDependencies packageJson] [requiredPackageMessage "devDependencies"]
-  HasExactVersionIfListed -> checkDeps [P.dependencies packageJson, P.devDependencies packageJson] []
```
infomiho, line 32: `HasExactVersionIfListed` was implemented this way because the user can install some dependency as either dev or not dev - we can't control that. So we wanted to ensure, _if the user installs some dep_ either in dev or not dev, it should of certain version not to break Wasp.

If I add `"@types/express": "^5.0.0"` to `dependencies` and not `devDependencies` a package with wrong version will be installed. (In this case, it won't, because we have extra deps check later on in the process).

Let me link to this [law](https://fs.blog/chestertons-fence/).

FranjoMindek: I've understood that intention, but wouldn't it also be nice to point the user into best practices? e.g. install types to devDependecies?

It's my bad that I forgot to cover that path in my solution, before I added optional/required options, it was covered by default by `missingRequiredPackageErrorMessage` error.

I could add a separate check in the `_notListed` path to see if something exist's in the wrong dependency type and notify the user of that.

infomiho: >  but wouldn't it also be nice to point the user into best practices

Yep, we can point out stuff like that in the docs, but I wouldn't try to guess each deps proper place in the compiler, it would be too hard to do correctly / too much work overall.

We need to agree on what is that we are validating: we want to ensure that users don't add deps that will break Wasp with weird errors but rather we want to give them a nice informative error message "Hey you can't use React 19". 

For the required deps, it's quite easy, they have a proper place. 

For "optional" deps it gets tricky, because they aren't really optional deps with a correct place, they are deps that we know that can break Wasp if installed with the wrong version even if they are in the "right" place. 

Maybe the naming should be different, it should probably concern with `CorrectVersionUsed` and the second dimension can be `CorrectLocationUsed`.

infomiho: >  Optional - The package is required to activate optional wasp functionality
 
I'd be more precise here, if a certain Wasp feature is optional, as-in, user can opt-in, and the compiler can detect that user is using that feature, then any dep related to that feature becomes required. Wasp ensures you have the deps installed for the Wasp features you used. 

The deps that were `HasExactVersionIfListed` before are deps that can be present, but they don't have to, Wasp doesn't care if they are not present, but it only cares about the version installed if they are present. Wasp doesn't care about certain deps being in the wrong place since it won't break Wasp. 

If we can model that successfully, I'll be happy with any solution we come up with 👍

waspc/src/Wasp/Generator/ExternalConfig/PackageJson.hs
```diff
@@ -5,52 +5,76 @@ where
...
+data PackageRequirement = Required | Optional
+
+type PackageValidation = (PackageDependecyType, PackageRequirement)
 
 validatePackageJson :: P.PackageJson -> [ErrorMsg]
 validatePackageJson packageJson =
   concat
-    [ validate ("wasp", "file:.wasp/out/sdk/wasp") IsListedWithExactVersion,
-      validate ("prisma", show prismaVersion) IsListedAsDevWithExactVersion,
+    [ -- dependencies
```
infomiho, line 22: I'd maybe get rid of the comments and actually group it in two helper functions e.g. `validateRuntimeDeps` and `validateDevDeps` which both return `[ErrorMsg]` and can then concat them into one final array here:
```haskell
validateRuntimeDeps packageJson ++
  validateDevDeps packageJson
```

waspc/src/Wasp/Generator/ExternalConfig/PackageJson.hs
```diff
@@ -3,54 +3,129 @@ module Wasp.Generator.ExternalConfig.PackageJson
...
+          "in package.json."
+        ]
+
+    wrongDependencyTypeErrorMessage :: ErrorMsg
+    wrongDependencyTypeErrorMessage =
+      unwords
+        [ "Wasp requires package",
+          show packageName,
+          "to be in",
+          packageJsonDependencyKey,
```
infomiho, line 127:
```suggestion
          show packageJsonDependencyKey,
```

I'd add `show` to get the double quotes around the location.

waspc/src/Wasp/Generator/ExternalConfig/PackageJson.hs
```diff
@@ -3,54 +3,129 @@ module Wasp.Generator.ExternalConfig.PackageJson
...
+        M.lookup packageName (P.dependencies packageJson)
+          <|> M.lookup packageName (P.devDependencies packageJson)
+
+    isInWrongDependencyType :: Bool
+    isInWrongDependencyType = case constraint of
+      RequiredRuntime -> M.member packageName (P.devDependencies packageJson)
+      RequiredDevelopment -> M.member packageName (P.dependencies packageJson)
+      Optional -> False
+
+    packageJsonDependencyKey :: String
```
infomiho, line 88:
```suggestion
    expectedPackageJsonLocation :: String
```

I think it would be beneficial to include "expected" to make it more clear what this key means.

waspc/src/Wasp/Generator/ExternalConfig/PackageJson.hs
```diff
@@ -3,54 +3,129 @@ module Wasp.Generator.ExternalConfig.PackageJson
...
+  concat
+    [ validateOptional ("typescript", show typescriptVersion),
+      validateOptional ("@types/react", show reactTypesVersion),
+      validateOptional ("@types/express", show expressTypesVersion)
+    ]
+  where
+    validateOptional spec = validatePackageJsonDependency packageJson spec Optional
+
+validatePackageJsonDependency :: P.PackageJson -> PackageSpecification -> PackageRequirement -> [ErrorMsg]
+validatePackageJsonDependency packageJson (packageName, expectedPackageVersion) constraint =
```
infomiho, line 60:
```suggestion
validatePackageJsonDependency packageJson (packageName, expectedPackageVersion) requirement =
```

I'd be consistent, either we use requirement or constraint (both sound good to me) 👍

waspc/src/Wasp/Generator/ExternalConfig/PackageJson.hs
```diff
@@ -3,54 +3,129 @@ module Wasp.Generator.ExternalConfig.PackageJson
...
+      if actualPackageVersion == expectedPackageVersion
+        then []
+        else [incorrectPackageVersionErrorMessage]
+    _notListed ->
+      if isInWrongDependencyType
+        then [wrongDependencyTypeErrorMessage]
+        else case constraint of
+          RequiredRuntime -> [missingRequiredPackageErrorMessage]
+          RequiredDevelopment -> [missingRequiredPackageErrorMessage]
+          Optional -> []
```
infomiho, line 72:
```haskell
case maybePackageJsonDepedency of
    Just actualPackageVersion ->
      if actualPackageVersion == expectedPackageVersion
        then []
        else [incorrectPackageVersionErrorMessage]
    Nothing ->
      if isInWrongLocation requirement
        then [wrongDependencyTypeErrorMessage]
        else getMissingPackageError requirement
```
Here I'd extract the inline case expression to make it more readable. Also, the `_notListed` is a nice way of writing it, but it prompted me to think that there are more cases than simply just `Nothing` so I'd maybe put `Nothing` to make it explicit.

Abstracting away the `case` (I wrote about this `\case` syntax below): 
```haskell
getMissingPackageError :: PackageRequirement -> [ErrorMsg]
    getMissingPackageError = \case
      RequiredRuntime -> [missingRequiredPackageErrorMessage]
      RequiredDevelopment -> [missingRequiredPackageErrorMessage]
      Optional -> []
```

 I'll comment on some of the other changes below.

waspc/src/Wasp/Generator/ExternalConfig/PackageJson.hs
```diff
@@ -3,54 +3,129 @@ module Wasp.Generator.ExternalConfig.PackageJson
...
-
-    incorrectVersionMessage :: String
-    incorrectVersionMessage =
+    getPackageJsonDependency :: Maybe P.PackageVersion
+    getPackageJsonDependency = case constraint of
+      RequiredRuntime -> M.lookup packageName $ P.dependencies packageJson
+      RequiredDevelopment -> M.lookup packageName $ P.devDependencies packageJson
+      Optional ->
+        M.lookup packageName (P.dependencies packageJson)
+          <|> M.lookup packageName (P.devDependencies packageJson)
```
infomiho, line 80: I like the use of the `<|>` operator here, really advanced stuff!

We have this convention of naming variables of type `Maybe a` like `maybePackageJsonDepedency`. Also, I disliked the fact it had `get` prefix making me think it's a function that requires some arguments.

waspc/src/Wasp/Generator/ExternalConfig/PackageJson.hs
```diff
@@ -3,54 +3,129 @@ module Wasp.Generator.ExternalConfig.PackageJson
...
+
+    isInWrongDependencyType :: Bool
+    isInWrongDependencyType = case constraint of
+      RequiredRuntime -> M.member packageName (P.devDependencies packageJson)
+      RequiredDevelopment -> M.member packageName (P.dependencies packageJson)
+      Optional -> False
```
infomiho, line 86: I'd make this into a function that takes the `requirement` as its argument, it feels kinda off having these values already evaluated even though we'll maybe use them.

Something like this:
```haskell
isInWrongLocation :: PackageRequirement -> Bool
isInWrongLocation = \case
  RequiredRuntime -> M.member packageName (P.devDependencies packageJson)
  RequiredDevelopment -> M.member packageName (P.dependencies packageJson)
  Optional -> False
```

Here I've used a neat trick not to be forced to come up with the param name and just used the [lambda case](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/lambda_case.html).

This is equivalent to writing:
```haskell
```haskell
isInWrongLocation :: PackageRequirement -> Bool
isInWrongLocation requirement = case requirement of
  RequiredRuntime -> M.member packageName (P.devDependencies packageJson)
  RequiredDevelopment -> M.member packageName (P.dependencies packageJson)
  Optional -> False
```

waspc/src/Wasp/Generator/ExternalConfig/PackageJson.hs
```diff
@@ -3,54 +3,129 @@ module Wasp.Generator.ExternalConfig.PackageJson
...
+
+    packageJsonDependencyKey :: String
+    packageJsonDependencyKey = case constraint of
+      RequiredRuntime -> "dependencies"
+      RequiredDevelopment -> "devDependencies"
+      Optional -> "dependencies or devDependencies"
```
infomiho, line 92: Again, I'd go with function that take the requirement as a param:
```haskell
getPackageJsonDependencyKey :: PackageRequirement -> String
getPackageJsonDependencyKey = \case
  RequiredRuntime -> "dependencies"
  RequiredDevelopment -> "devDependencies"
  Optional -> "dependencies or devDependencies"
```

waspc/src/Wasp/Generator/ExternalConfig/PackageJson.hs
```diff
@@ -3,54 +3,129 @@ module Wasp.Generator.ExternalConfig.PackageJson
...
+
+    oppositePackageJsonDependencyKey :: String -> String
+    oppositePackageJsonDependencyKey "dependencies" = "devDependencies"
+    oppositePackageJsonDependencyKey "devDependencies" = "dependencies"
+    oppositePackageJsonDependencyKey _ = error "Unknown dependency key"
```
infomiho, line 97: `error` here is an escape hatch when we can't really pattern match everything. In this case, I'd suggest using the `Maybe a` type to do something like this:
```haskell
getOppositePackageJsonDepedencyKey :: PackageRequirement -> Maybe String
getOppositePackageJsonDepedencyKey = \case
  RequiredRuntime -> Just "devDependencies"
  RequiredDevelopment -> Just "dependencies"
  Optional -> Nothing
```
and then below, you can optionally render the second part of the message:
```haskell
wrongDependencyTypeErrorMessage :: ErrorMsg
    wrongDependencyTypeErrorMessage =
      unwords $
        [ "Wasp requires package",
          show packageName,
          "to be in",
          show $ getPackageJsonDependencyKey requirement
        ]
          ++ ( case getOppositePackageJsonDepedencyKey requirement of
                 Just oppositeKey -> ["and not in", show oppositeKey]
                 Nothing -> []
             )
          ++ ["in package.json."]
```

infomiho (commented):
I like the `PackageRequirement` change 👍 

I've left some comments on how I'd maybe improve the `validatePackageJsonDependency` fn.

waspc/src/Wasp/Generator/ExternalConfig/PackageJson.hs
FranjoMindek: Didn't know it did that for Strings.

waspc/src/Wasp/Generator/ExternalConfig/PackageJson.hs
FranjoMindek: In a way we know that `getOppositePackageJsonDepedencyKey` will always be called with only `RuntimeRequired` or `DevelopmentRequired` because of the previous logic. So this is not really needed.

Do we still want to cover it?

infomiho: > In a way we know 

It's always better to encode this knowledge in types so the compiler will look out for us :)

infomiho (approved):
🚢 it

PR #2563, opened 2025-03-20
Adds Posthog script to usemage.ai

infomiho:
Named the file `piggy.js` to better avoid ad blockers from blocking it.

Branched out from `release` so we can get it out faster.


infomiho:
Usually yes, ad blockers detect certain domains and filenames :)

PR #2566, opened 2025-03-20
Update Ubuntu image in CI

sodic:


.github/workflows/waspc-ci.yaml
```diff
@@ -47,7 +47,7 @@ jobs:
           # We can return to `ubuntu-latest` when enough time has elapsed.
           # Still looking into musl static binaries.
           # Ref: https://github.com/wasp-lang/wasp/issues/650
-          - ubuntu-20.04
+          - ubuntu-latest
```
Martinsos, line 50: @sodic there is a big fat comment above that you should take into account and/or update/delete!

.github/workflows/waspc-ci.yaml
```diff
@@ -188,22 +188,22 @@ jobs:
...
-        if: steps.cache-playwright-browsers.outputs.cache-hit != 'true' && matrix.os == 'ubuntu-20.04'
+        if: steps.cache-playwright-browsers.outputs.cache-hit != 'true' && matrix.os == 'ubuntu-latest'
         run: |
           cd examples/todoApp
           npx playwright install --with-deps
 
       - name: Headless - Run Playwright Tests
         id: headless-run-playwright-tests
-        if: matrix.os == 'ubuntu-20.04'
+        if: matrix.os == 'ubuntu-latest'
```
Martinsos, line 206: Wow we are really repeating this ubuntu version in a ton of places. I believe there is a way to have a variable here in Github Action YAML, maybe this is a good moment to extract it? If not, ok we can do it later, I won't push too much, but am tempted.

Martinsos (changes requested):
Since you didn't delete/update that comment, I wonder if you saw it, and if you didn't, then I am not confident that choice of `ubuntu-latest` is correct, so requesting changes for now!

.github/workflows/waspc-ci.yaml
sodic: > I believe there is a way to have a variable here in Github Action YAML

Unfortunately no - you can have something like a variable (YAML anchor) but you can't use it in strings and conditional clauses. To do that, I'd have to define an env variable and change the syntax of `if` clauses to check the env, which is a bit too much for my taste.

.github/workflows/waspc-ci.yaml
sodic: I've set the ubuntu version to ubuntu-22.04.

How do I check this doesn't have the glibc problem and whether it's even necessary to use the older version of ubuntu?

waspc/src/Wasp/Env.hs
```diff
@@ -22,7 +22,7 @@ type EnvVarName = String
 
 type EnvVarValue = String
 
--- Reads specified dotenv file and returns its values.
+-- Reads the specified dotenv file and returns its values.
```
sodic, line 25: This was just to trigger CI.

Martinsos (commented):
Left some comments!

.github/workflows/waspc-ci.yaml
sodic: > You have set it to ubuntu-latest though, not 22.04, at least excplicitly, right?

Hm, I think you're looking at old code.

I understand the problem, but am wondering how to test it:
> How do I check this doesn't have the glibc problem and whether it's even necessary to use the older version of ubuntu?

Martinsos (approved):
Ok great! Let's just improve the comment (glibc) a bit and we are good.

PR #2567, opened 2025-03-20
waspc/run script can now run wasp-cli from anywhere, not just from waspc/ dir

Martinsos:
I was considering if I should add the `--project-dir` flag to the other `cabal` commands in the `run` script, but I decided against it at the end: it doesnt' make much sense for them, they are all development commands that you are running when doing changes in the waspc/ dir, and it would just unneccesarily complicate the `run` script.


cprecioso (approved):
It works ✅

infomiho:
Nice change 👍

cprecioso:
Yep it does... 🤔
image

infomiho:
I had the Cabal version `3.6.2.0` and @cprecioso had `3.14.1.1` (latest). After I upgraded, the script started working for me 👍 

I think we should all upgrade to latest cabal.

Martinsos:
Aha I wasn't aware this is a relatively new thing!

There doesn't seem to be a way to enforce the minimal version of `cabal-install` (cabal CLI) though. So not sure how to do it. But maybe we can just solve this as we go, I mean if anybody in the team gets stuck on this, we will quickly figure out what is the issue I think.
We could potentially mention it in the README, but then we have to maintain that, ... .

PR #2576, opened 2025-03-27
Deployment script cleanup

infomiho:
- Migrated the deployment scripts to use spaces (to be consistent with the rest of the codebase)
- Added Prettier config
- Bumped version of `zx` lib (we use it to run commands)
  - Added `syncProcessCwd` to keep the behavior we relied on with the old version


cprecioso (approved):
I think for the future we should do the reformatting apart from the other changes, because it's made it a bit hard to see the changes 😅 Even then, I ran `prettier` on the `deploy` folder and still there were some more changes.

Otherwise changes LGTM. I'll leave it approved so you can merge when you've dealt with the missing formatting.

infomiho:
@cprecioso Sorry about that! I did the formatting changes mostly while working on the #2578 PR so this was "side effects" of that. I should have made two PRs:
- one with the version bumps
- second with only formatting + all of the formatting

PR #2580, opened 2025-03-27
Typescript application specification http route changes

### LINKED ISSUE #2532: In Wasp TS config (spec), make `HttpRoute` an object instead of a tuple
## Context
Wasp 0.15.0 introduced Wasp TS spec (config). Check the docs to learn more about it: https://wasp.sh/docs/general/wasp-ts-config

Relevant PRs
- #2276 
- #2299
- #2435

The original RFC is available at https://www.notion.so/wasp-lang/RFC-TS-DSL-SDK-86c4082ab9924c77b07dcb25d4cd95ad

There were obvious improvements we could make in the process, we decided to keep the TypeScript syntax as similar as possible to the existing DSL.

 Now that the TS config works, we can start making those improvements.

## Task
Here's now users can define APIs with custom HTTP routes:
```ts
app.api('barBaz', {
  // ...
  httpRoute: ['GET', '/bar/baz']
});
```
A better API would look something like this:
```ts
app.api('barBaz', {
  // ...
  httpRoute: {
    method: 'GET',
    route: '/bar/baz',
  }
});
```
This will also require updating the docs.

FranjoMindek:
Makes the user API `HttpRoute` of type

```ts
export type HttpRoute = {
  method: AppSpec.HttpMethod
  route: string
}
```

which works better with Typescript, instead of the app spec API

```ts
export type HttpRoute = [HttpMethod, string]
```

Also fixes the eslint to work properly with the `_` prefix for unused vars.

Fixes #2532

### DIFF
--- a/waspc/packages/wasp-config/src/mapUserSpecToAppSpecDecls.ts
+++ b/waspc/packages/wasp-config/src/mapUserSpecToAppSpecDecls.ts
@@ -142,6 +142,10 @@ function mapExtImport(extImport: User.ExtImport): AppSpec.ExtImport {
   }
 }
 
+function mapHttpRoute(httpRoute: User.HttpRoute): AppSpec.HttpRoute {
+  return [httpRoute.method, httpRoute.route]
+}
+
 function mapApiConfig(
   config: User.ApiConfig,
   parseEntityRef: RefParser<'Entity'>
@@ -151,7 +155,7 @@ function mapApiConfig(
     fn: mapExtImport(fn),
     middlewareConfigFn: middlewareConfigFn && mapExtImport(middlewareConfigFn),
     entities: entities && entities.map(parseEntityRef),
-    httpRoute,
+    httpRoute: mapHttpRoute(httpRoute),
     auth,
   }
 }
@@ -333,7 +337,8 @@ function mapServer(server: User.ServerConfig): AppSpec.Server {
   return {
     setupFn: setupFn && mapExtImport(setupFn),
     middlewareConfigFn: middlewareConfigFn && mapExtImport(middlewareConfigFn),
-    envValidationSchema: envValidationSchema && mapExtImport(envValidationSchema),
+    envValidationSchema:
+      envValidationSchema && mapExtImport(envValidationSchema),
   }
 }
 
@@ -343,7 +348,8 @@ function mapClient(client: User.ClientConfig): AppSpec.Client {
     setupFn: setupFn && mapExtImport(setupFn),
     rootComponent: rootComponent && mapExtImport(rootComponent),
     baseDir,
-    envValidationSchema: envValidationSchema && mapExtImport(envValidationSchema),
+    envValidationSchema:
+      envValidationSchema && mapExtImport(envValidationSchema),
   }
 }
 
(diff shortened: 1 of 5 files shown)


waspc/packages/wasp-config/src/mapUserSpecToAppSpecDecls.ts
```diff
@@ -333,7 +337,8 @@ function mapServer(server: User.ServerConfig): AppSpec.Server {
   return {
     setupFn: setupFn && mapExtImport(setupFn),
     middlewareConfigFn: middlewareConfigFn && mapExtImport(middlewareConfigFn),
-    envValidationSchema: envValidationSchema && mapExtImport(envValidationSchema),
+    envValidationSchema:
+      envValidationSchema && mapExtImport(envValidationSchema),
```
FranjoMindek, line 341: I have global prettier config which works for all JS/TS project if it can't find the project specific config. Here the code probably went over the recommended line char limit.

Maybe this is something we should do team wide so we have consistent JS/TS experience. Otherwise we could try to standardize eslint/prettier rules and copy/paste them in all projects.

waspc/packages/wasp-config/src/userApi.ts
```diff
@@ -1,7 +1,7 @@
 /** This module defines the user-facing API for defining a Wasp app.
  */
-import * as AppSpec from './appSpec.js'
 import { GET_USER_SPEC } from './_private.js'
+import * as AppSpec from './appSpec.js'
```
FranjoMindek, line 4: Auto import sorting is done by VSC, but it can be done by eslint so it's standardized.

waspc/packages/wasp-config/src/mapUserSpecToAppSpecDecls.ts
infomiho: We had `.prettierrc` per package (not for all of them), maybe we can add it here as well?

web/docs/general/wasp-ts-config.md
```diff
@@ -233,7 +233,10 @@ app.api('barBaz', {
   fn: { import: 'barBaz', from: '@src/apis' },
   auth: false,
   entities: ['Task'],
-  httpRoute: ['GET', '/bar/baz']
+  httpRoute: {
```
infomiho, line 236: Do we have some example apps that we need to update?

waspc/packages/wasp-config/src/mapUserSpecToAppSpecDecls.ts
Genyus: Just a suggestion, but this feels like a job for [editorconfig](https://editorconfig.org)? You could use that to set project-wide defaults and then [apply overrides for specific packages](https://prettier.io/docs/configuration#editorconfig) when necessary.

FranjoMindek: We will open up a issue for this probably, thanks for the suggestion, I'll be sure to add it as an option.

web/docs/general/wasp-ts-config.md
FranjoMindek: This is the only instance where we showcase it as far as I've found.

infomiho (approved):

Issue #1859, opened 2024-03-07
Fix dependency version checking


sodic:
This is how how Wasp consolidates user dependencies with framework dependencies as of Wasp 0.12.0 (taken from original discussion in https://github.com/wasp-lang/wasp/pull/1602#discussion_r1431438802):

For each dependency, decide which version we allow.
   - If the user doesn't specify that dependency in their `package.json`, Wasp lists it in `{sever,web-app}/package.json`.
   - If the user specifies that dependency with the correct version in their `package.json`, Wasp installs it in the top-level `node_modules` and doesn't list it in `{sever,web-app}/package.json`.
   - If the user specifies that dependency with the wrong version in their `package.json`, Wasp reports a conflict and refuses to build (this is what we already do in the old structure, this PR just keeps the behavior).

Since our version conflict check uses single string comparisons (instead of semantically comparing version ranges), it sometimes reports false positives.

For example:
 - The user wants a dependency in range `^1.4.1` 
 - Wasp wants a dependency in range `^1.4.0`
 - The overlapping range is `^1.4.1`, but Wasp reports a conflict.

Martinsos:
This might be become redundant if we go with aliasing wasp deps, since then there is no conflict anymore and we don't need the check anymore.

infomiho:
In order to make this happen, we would need to write a parser for semver so we can semantically compare Wasp dep version with user's dep version.

We decided to shelf this task because we feel it's a part of a bigger picture (npm workspaces, aliasing npm deps etc.) and this issue might disappear all together in a different future setup.

cprecioso:
The logic for avoiding dependency shadowing does not exist anymore since #3130 (although Wasp will still report an error if the user has a different version range than we want in their package.json)

Martinsos:
Ok closing this one then @cprecioso @sodic ?

cprecioso:
@Martinsos I'm not sure, I think we need to do:
- #3129

There are still some open questions in this issue that we'd have to agree on (as part of this other one I just mentioned):
- Whether to update our dependency checking logic to allow ranges that are compatible with ours
- Whether we are supplying another dependency overriding method altogether and we keep the strict range checking
- Making sure that we declare the actual dependencies we need on each package and not rely on implicitly being able to access another package's deps

infomiho:
Removed the `bug` label since it's not a bug really, we are talking about improving the current dependency situations.

PR #2608, opened 2025-04-03
Fix problems in 0.16.3-rc1

sodic:
What I did:
- Fixed an incompatible Vite version in websockets voting app
- Fixed a Tailwind issue in `waspc/examples/todoApp`
- Updated `package-lock.json` files
- Fixed Tailwind docs
- Emphasized breaking changes in the changelog


waspc/examples/todoApp/package.json
```diff
@@ -23,10 +21,13 @@
     "wasp": "file:.wasp/out/sdk/wasp"
   },
   "devDependencies": {
+    "@tailwindcss/forms": "^0.5.10",
+    "@tailwindcss/typography": "^0.5.16",
     "@types/express": "^4.17.13",
     "@types/react": "^18.0.37",
     "@types/uuid": "^9.0.8",
     "prisma": "5.19.1",
+    "tailwindcss": "^3.4.17",
```
infomiho, line 30: Please use `^3.2.7` in the `package.json` (I know it's the same by semver, but our version checking code will give you an error for this in the future)

web/docs/project/css-frameworks.md
```diff
@@ -91,6 +91,9 @@ Make sure to use the `.cjs` extension for these config files, if you name them w
 To add Tailwind plugins, install them as npm development [dependencies](../project/dependencies) and add them to the plugins list in your `tailwind.config.cjs` file:
 
 ```shell
+# Since Wasp currently supports Tailwind ^3.2.7, you must
+# explicitly install a compatible version.
+npm install -D tailwindcss@^3.2.7
```
infomiho, line 96: To get the `^3.2.7` version in the `package.json` you need to write it like this:
```suggestion
npm install -D tailwindcss@3.2.7
```

waspc/examples/todoApp/package.json
sodic: Shieet, I need to change the instructions too then because `npm install tailwindcss@^3.2.7` installs this one (which makes sense).

Btw, can we change the validation code to check version compatibility instead of comparing exact strings? We have that logic somewhere in our codebase, shouldn't be too difficult.

That seems like a better solution because, even if I change the version here, the current approach will give users validation errors for perfectly valid `package.json`s

infomiho: Talked IRL, we still to implement a semver parser https://github.com/wasp-lang/wasp/issues/1859

infomiho (approved):

PR #2620, opened 2025-04-04
Update Wasp version in Mage

cprecioso:


cprecioso:
@infomiho If I understand correctly, this can't be deployed yet, as the 0.16.3 version is not released, so it will fail when build, right?

Locally I tried with 0.16.3-rc1 and worked well

infomiho:
Yep, I'll deploy the Mage change after the release 👍

infomiho (approved):

cprecioso:
Oki, I'll set this to draft so we don't merge it mistakenly

PR #2684, opened 2025-04-17
Fix OAuth login race condition

### LINKED ISSUE #2570: OAuth login race condition
Sometimes, when users try to login with an OAuth provider, the OAuth provider (e.g. Google) redirects to the app successfully, but the session is not created successfully - and the login is completed successfully.

https://github.com/user-attachments/assets/6738da5c-214b-4d81-bc52-0ed992826e2f

**What you can see in the video:**
The token received from the provider is exchanged successfully using `http://localhost:3001/auth/exchange-code` for a session ID. But, there is a race condition _somewhere_ which results in a call to the `http://localhost:3001/auth/me` endpoint without the `Authorization: Bearer <session id>` header. This then logs out the user and the user is never redirected to the homepage.

infomiho:
We had a race condition in the OAuth callback page which exchanges a one time token for a session ID. At the same time we had `useAuth` hook trying to fetch `/auth/me` with an old session ID and the OAuth callback page setting the new session ID. When the `/auth/me` call failed with `401` - it would clear the session ID _just as it was set_ by the OAuth callback page.

shapes at 25-04-17 15 21 24

I've added a check before clearing the session ID if the session ID used for the failed request is indeed the same as the current session ID. Basically, don't blame the wrong session ID.

Closes #2570


waspc/data/Generator/templates/sdk/wasp/api/index.ts
```diff
@@ -38,12 +38,15 @@ api.interceptors.request.use((request) => {
   const sessionId = getSessionId()
   if (sessionId) {
     request.headers['Authorization'] = `Bearer ${sessionId}`
+    request.headers['X-Session-Id'] = sessionId
```
infomiho, line 41: If needed we can use the `Authorization` header and then:
- check if it's `undefined`
- if not, split by `' '` and get the session ID that way

Adding this new header felt more precise.

cprecioso (approved):
LGTM, I also checked that this wouldn't affect CORS (since the server can reject a request based on the names of the headers that will be sent, but it doesn't seem we're filtering on that, only origin.)

cprecioso:
Wait actually, this will add a CORS preflight request for every request, as it will make all of them "complex" due to having custom headers. I'll run some tests.

infomiho:
@cprecioso I investigated if there is a way to "mark" the requests with something other then headers, but I couldn't find anything :/ thank you for the research about CORS!

waspc/data/Generator/templates/sdk/wasp/api/index.ts
```diff
@@ -34,16 +34,19 @@ export function removeLocalUserData(): void {
   apiEventsEmitter.emit('sessionId.clear')
 }
 
-api.interceptors.request.use((request) => {
+api.interceptors.request.use((config) => {
```
infomiho, line 37: Renamed to `config` because it is indeed the config and not a request (in line with the docs: https://axios-http.com/docs/interceptors)

waspc/data/Generator/templates/sdk/wasp/api/axios.d.ts
```diff
@@ -0,0 +1,7 @@
+import 'axios';
+
+declare module 'axios' {
+  export interface AxiosRequestConfig {
```
infomiho, line 4: Typing the extra field on the config

cprecioso (approved):
I'd say to add a comment to the two interceptors saying that they are related and the fact that they are checking if the session that went out is the same that arrived, and good to go! Leaving it approved 👌🏻

PR #2686, opened 2025-04-18
Remove duplicates and sort `wasp-config` types

FranjoMindek:
Deletes `EmailSender` and `EmailProvider` types which are just **unused** copies of `EmailSenderConfig`.

Reorders types to be in order they appear in the `App` class, meaning `UserSpec` first and the we go by `UserSpec` fields types.

### DIFF
--- a/waspc/packages/wasp-config/src/userApi.ts
+++ b/waspc/packages/wasp-config/src/userApi.ts
@@ -92,55 +92,29 @@ export class App {
   }
 }
 
+export type UserSpec = {
+  app: { name: string; config: AppConfig }
+  actions: Map<string, ActionConfig>
+  apiNamespaces: Map<string, ApiNamespaceConfig>
+  apis: Map<string, ApiConfig>
+  auth?: AuthConfig
+  client?: ClientConfig
+  cruds: Map<string, Crud>
+  db?: DbConfig
+  emailSender?: EmailSenderConfig
+  jobs: Map<string, JobConfig>
+  pages: Map<string, PageConfig>
+  queries: Map<string, QueryConfig>
+  routes: Map<string, RouteConfig>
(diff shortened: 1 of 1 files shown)


waspc/packages/wasp-config/src/userApi.ts
```diff
@@ -208,91 +246,46 @@ type ExecutorOptions = {
...
-  fromField: EmailFromField
-  emailVerification: EmailVerificationConfig
-  passwordReset: PasswordResetConfig
+export type ServerConfig = {
+  setupFn?: ExtImport
+  middlewareConfigFn?: ExtImport
+  envValidationSchema?: ExtImport
 }
 
-export type EmailSender = {
```
FranjoMindek, line 260: Probably went unnoticed because it's exported, so there is no unused type warning.

waspc/packages/wasp-config/src/userApi.ts
```diff
@@ -208,91 +246,46 @@ type ExecutorOptions = {
...
-  getEmailContentFn?: ExtImport
-  clientRoute: string
-}
-
-export type PasswordResetConfig = {
-  getEmailContentFn?: ExtImport
-  clientRoute: string
-}
-
-export type UserSpec = {
```
FranjoMindek, line 282: It felt bad that the most important type is at the bottom of the file.

cprecioso (approved):

waspc/packages/wasp-config/src/userApi.ts
FranjoMindek: I did the sorting manually. Just noticed that I have two of them during the process.

sodic (approved):
Nice work, I like the reordering!

PR #2688, opened 2025-04-22
Update snapshots

cprecioso:
Updates the snapshots that drifted


FranjoMindek (approved):
Those look like all 6 fails fixed. LGTM

PR #2693, opened 2025-04-23
Adds Prisma setup fn

### LINKED ISSUE #2116: Prisma setup hook
Prisma supports client extensions: https://www.prisma.io/docs/orm/prisma-client/client-extensions which for:
> You can use Prisma Client extensions to add functionality to your models, result objects, and queries, or to add client-level methods.

> You can create an extension with one or more of the following component types:

> model: [add custom methods or fields to your models](https://www.prisma.io/docs/orm/prisma-client/client-extensions/model)
> client: [add client-level methods to Prisma Client](https://www.prisma.io/docs/orm/prisma-client/client-extensions/client)
> query: [create custom Prisma Client queries](https://www.prisma.io/docs/orm/prisma-client/client-extensions/query)
> result: [add custom fields to your query results](https://www.prisma.io/docs/orm/prisma-client/client-extensions/result)

Users can't do this with the existing Wasp setup since there is no way for them to _give Wasp a new Prisma client to use_ throughout the app. 

That's why having a Prisma setup hook that could influence which Prisma client instance Wasp uses is useful.

### Implementation

It would probably be a `db.prismaSetup` hook whose implementation could look like this:
```ts
export function setupPrisma(prisma: PrismaClient): PrismaClient {
  // An example from Prisma docs
  return prisma.$extends({
    query: {
      user: {
        async findMany({ model, operation, args, query }) {
          // take incoming `where` and set `age`
          args.where = { ...args.where, age: { gt: 18 } }
  
          return query(args)
        },
      },
    },
  })
}

// Somewhere in the codebase
await prisma.user.findMany() // returns users whose age is greater than 18
```

infomiho:
This PR enables users to override the default `PrismaClient` init and setup e.g. logging of SQL that Prisma is sending to the DB. It's also useful to add [Prisma client extensions](https://www.prisma.io/docs/orm/prisma-client/client-extensions/client).

Left to do:
- [x] Investigate if we need to adjust how we use the `PrismaClient` _type_ (maybe we will need to infer the type from the user returned `PrismaClient`)
- [x] Docs
- [x] Add a test (probably headless test)
- [x] Update the Typescript Spec

Closes #2116


waspc/data/Generator/templates/sdk/wasp/server/dbClient.ts
```diff
@@ -1,9 +1,22 @@
...
+import { PrismaClient } from '@prisma/client'
 
-function createDbClient(): Prisma.PrismaClient {
-  return new Prisma.PrismaClient()
+{=# prismaSetupFn.isDefined =}
+const _waspSetupPrisma = {= prismaSetupFn.importIdentifier =}
+{=/ prismaSetupFn.isDefined =}
+{=^ prismaSetupFn.isDefined =}
+const _waspSetupPrisma = () => new PrismaClient()
+{=/ prismaSetupFn.isDefined =}
```
infomiho, line 13: Either we use the default init, or the user defined init.

waspc/data/Generator/templates/sdk/wasp/server/index.ts
```diff
@@ -14,4 +16,10 @@ export { MiddlewareConfigFn } from './middleware/index.js'
...
 
 // PUBLIC API
-export type DbSeedFn = (prisma: PrismaClient) => Promise<void>
+
+// PUBLIC API
+export type DbSeedFn = (prisma: ResolvedPrismaClient) => Promise<void>
+
+// PUBLIC API
+// TODO: We can't use this type becuase prisma.$extends() doesn't return a PrismaClient?
+export type PrismaSetupFn = () => PrismaClient
```
infomiho, line 25: I'm on the fence if it makes sense to provide this type:
1. we usually provided it and it enabled users to have _some_ type safety
2. the problem is that if the user extends the client with client extensions _for some reason_ the returned type is incompatible with `PrismaClient`

```ts
const prisma = new PrismaClient({
  log: ['query'],
}).$extends({
  query: {
    task: {
      async findMany({ args, query }) {
        args.where = {
          ...args.where,
          description: { not: { contains: 'hidden by setupPrisma' } },
        }
        return query(args)
      },
    },
  },
})

// This is an type error ❌
const someOtherObj: PrismaClient = prisma
```

waspc/examples/todoApp/headless-tests/tests/prisma-setup-fn.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+      email,
+      password,
+    })
+
+    await expect(page).toHaveURL('/profile')
+
+    await page.goto('/')
+
+    // Create a new task that will be hidden by the Prisma setup function
+    const specificTask = `hidden by setupPrisma`
```
infomiho, line 29: We filter out all the tasks that contain this string. This enables us to test out the Prisma setup fn with headless tests. I couldn't come up with something better 🤷

waspc/data/Generator/templates/sdk/wasp/server/index.ts
cprecioso: Maybe something like this?

```ts
export type PrismaSetupFn<T extends PrismaClient> = () => T
```

waspc/data/Generator/templates/sdk/wasp/server/dbClient.ts
```diff
@@ -1,9 +1,22 @@
...
 {=# areThereAnyEntitiesDefined =}
-import Prisma from '@prisma/client'
+import { PrismaClient } from '@prisma/client'
 
-function createDbClient(): Prisma.PrismaClient {
-  return new Prisma.PrismaClient()
+{=# prismaSetupFn.isDefined =}
+const _waspSetupPrisma = {= prismaSetupFn.importIdentifier =}
+{=/ prismaSetupFn.isDefined =}
```
cprecioso, line 10: Is it a requirement of the templates that we use two different variables?

Would it make sense to just check if there is an `importIdentifier` or not and skip the `isDefined`?

cprecioso (approved):
Looks good, snapshots need updating

waspc/data/Generator/templates/sdk/wasp/server/dbClient.ts
infomiho: Oh yes, unfortunately we have to provide an explicit boolean variable because a missing `importIdentifier`  is not considered false (something like that, I have faced this quite a few times, but I don't have _an exact error in mind now_)

waspc/data/Generator/templates/sdk/wasp/server/index.ts
infomiho: I'll try that and report back!

infomiho: The problem here is that the `DynamicClientExtensionThis` type which is the result of extending the `PrismaClient` doesn't extend the `PrismaClient`.

It looks like a known issue:
- https://github.com/prisma/prisma/issues/24966
- https://github.com/prisma/prisma/issues/20738#issuecomment-1807917019

I think we either acknowledge that the users might return an extended client and account for that with the type e.g. `() => PrismaClient | DynamicClientExtensionThis` or not provide a helper type. 

cc: @sodic

infomiho: The next problem seems to be that the `DynamicClientExtensionThis` type is not accessible to import from `@prisma/client`.

Damn, I guess we are not having that type helper, it's an advanced use case - I'd expect those users to know what they are doing anyways.

waspc/examples/todoApp/src/prisma.ts
```diff
@@ -0,0 +1,22 @@
+import { PrismaClient } from '@prisma/client'
+
+export const setupPrisma = () => {
```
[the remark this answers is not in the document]
sodic: But, more importantnly, what about the type? We don't provide it for this one?

waspc/src/Wasp/Generator/SdkGenerator.hs
```diff
@@ -359,12 +362,18 @@ genDbClient :: AppSpec -> Generator FileDraft
 genDbClient spec = do
   areThereAnyEntitiesDefined <- not . null <$> getEntitiesForPrismaSchema spec
 
-  let tmplData = object ["areThereAnyEntitiesDefined" .= areThereAnyEntitiesDefined]
+  let tmplData =
+        object
+          [ "areThereAnyEntitiesDefined" .= areThereAnyEntitiesDefined,
+            "prismaSetupFn" .= GJI.jsImportToImportJson (extImportToJsImport <$> maybePrismaSetupFn)
```
sodic, line 368: Franjo recently extracted these two calls into its own function. So keep that in mind after you merge main.

waspc/data/Generator/templates/sdk/wasp/server/dbClient.ts
sodic: Why the underscore? It's not exported or anything. I saw Franjo doing it too. Is that like the new cool thing?

The comment about `waspSetupPrisma` vs `waspPrismaSetup` applies here too.

waspc/data/Generator/templates/sdk/wasp/server/dbClient.ts
```diff
@@ -1,9 +1,22 @@
 {{={= =}=}}
+{=# prismaSetupFn.isDefined =}
+{=& prismaSetupFn.importStatement =}
+{=/ prismaSetupFn.isDefined =}
 {=# areThereAnyEntitiesDefined =}
```
sodic, line 5: This here smells like an edge case. What happens if then don't define any entities, can they even define a prisma setup function then?

web/docs/data-model/databases.md
```diff
@@ -349,6 +431,23 @@ You'll often want to call `wasp db seed` right after you run `wasp db reset`, as
   Defines the seed functions you can use with the `wasp db seed` command to seed your database with initial data.
   Read the [Seeding section](#seeding-the-database) for more details.
 
+- `prismaSetupFn: ExtImport`
+
+  Defines a function that sets up the Prisma Client instance. Wasp expects you to return a Prisma Client instance from this function.
```
sodic, line 436:
```suggestion
  Defines a function that sets up the Prisma Client instance. Wasp expects it to return a Prisma Client instance.
```

waspc/examples/todoApp/headless-tests/tests/prisma-setup-fn.spec.ts
sodic: This tests a lot of stuff and can fail in many ways unrelated to the thing it tests. Together with the test for SuperJSON serialization in #2701 make me feel like we're really abusing headless tests at this point :smile: 

Perhaps it's time we introduced a different kind of test. What do you think?

If so, we should create an issue and somehow list all the tests that shouldn't be headless (mark them with todos as well).

sodic: Also, I see backticks around `hidden by setupPrisma`? Probably an accident.

sodic (commented):
Nice work!

I have some thoughts and suggestions :)

waspc/data/Generator/templates/sdk/wasp/server/index.ts
sodic: Hm, what do they say? Do they plan to fix it (I haven't read the issue)?

In any case, I think we should provide a type wrapper (even if it returns any). That way, when Prisma finally fixes it, we can easily update our API to be more type safe.

waspc/data/Generator/templates/sdk/wasp/server/dbClient.ts
```diff
@@ -19,5 +32,8 @@ function createDbClient(): null {
 
 const dbClient = createDbClient()
 
+// PUBLIC API 
+export type ResolvedPrismaClient = typeof dbClient
```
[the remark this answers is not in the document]
sodic: Oh, and don't forget the public API table :)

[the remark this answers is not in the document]
sodic: Oh, and don't forget the public API table for all new symbols :)

waspc/data/Generator/templates/sdk/wasp/server/dbClient.ts
infomiho: Since we never got to user import name mangling, that's how I make it "unique to wasp" by using something obviously weird e.g. `_wasp` prefix.

waspc/data/Generator/templates/sdk/wasp/server/dbClient.ts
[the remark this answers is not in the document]
infomiho: `@prisma/client` will contain the types of the regular Prisma client and the extended ones won't affect those types. So, for the power users that will use this feature, there is an expectation of them to know it IMHO. I can rename that symbol to make it more ergonomic for sure 👍

waspc/examples/todoApp/tsconfig.json
```diff
@@ -19,16 +19,10 @@
     "esModuleInterop": true,
     "isolatedModules": true,
     "moduleDetection": "force",
-    "lib": [
-      "dom",
-      "dom.iterable",
-      "esnext"
-    ],
+    "lib": ["dom", "dom.iterable", "esnext"],
```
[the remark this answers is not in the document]
infomiho: I'm using VS Code with Prettier, try running `npx prettier ./tsconfig.json --write` in the `todoApp` dir. There is a PR right now that will run `prettier` in the CI https://github.com/wasp-lang/wasp/pull/2736/files#diff-1cd343e53bb36a6168c13c8643bb6eb9706978a08f0c7057e23daf838865395aR27

waspc/data/Generator/templates/sdk/wasp/server/dbClient.ts
infomiho: The `areThereAnyEntitiesDefined` logic check is old code and it's needed because Prisma client can't be initialized if there are no entities. This causes an error if we don't check it. 

The new code is added inside of the logic branch where entities exist. Am I missing something?

waspc/data/Generator/templates/sdk/wasp/server/dbClient.ts
```diff
@@ -1,9 +1,22 @@
...
+{=/ prismaSetupFn.isDefined =}
+{=^ prismaSetupFn.isDefined =}
+const _waspSetupPrisma = () => new PrismaClient()
+{=/ prismaSetupFn.isDefined =}
+
+// We are not typing the return value because 
+// we want to infer the type from the (potentially)
+// user-defined function _waspSetupPrisma
+function createDbClient() {
+  return _waspSetupPrisma()
```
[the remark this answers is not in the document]
infomiho: Over time I started to prefer these small indirections to make the templates easier to work with. I prefer to use templating on "top level" and then do some runtime orchestration as I did in this PR. What do you think?

web/docs/data-model/databases.md
```diff
@@ -310,6 +310,86 @@ Check the [API Reference](#cli-commands-for-seeding-the-database) for more detai
...
+This allows you to configure features like [logging](https://www.prisma.io/docs/orm/prisma-client/observability-and-logging/logging) or [client extensions](https://www.prisma.io/docs/orm/prisma-client/client-extensions):
+
+<Tabs groupId="js-ts">
+  <TabItem value="js" label="JavaScript">
+    ```wasp title=main.wasp
+    app MyApp {
+      title: "My app",
+      // ...
+      db: {
+        prismaSetupFn: import { setupPrisma } from "@src/prisma"
```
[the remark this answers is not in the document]
infomiho: I was trying to stay consistent with all other setupFn APIs, to be honest. In hindsight, having a function gives us the control when to run it. That's one benefit that I can think of right now :)

waspc/data/Generator/templates/sdk/wasp/server/index.ts
infomiho: As I understood it, it's expected and Prisma won't fix it. 

Type wrapper that returns `any` will probably kill type inference when we use `ReturnType` to extract the "user defined Prisma Client" since I don't expect most of our users to use the helper with `satisfies` so I'm against it.

waspc/data/Generator/templates/sdk/wasp/server/dbClient.ts
sodic: Everything seems alright now, not sure what I was worried about.

waspc/data/Generator/templates/sdk/wasp/server/dbClient.ts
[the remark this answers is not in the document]
sodic: I prefer it without the indirection. My thinking is:
1. From what I see, once we start separating the templates, the code with indirection won't make our jobs all that easier. It might even make them more difficult.
2. The generated code without indirections is simpler. And, since the generated code is technically "the product," it feels wrong sacrificing its simplicity to make our implementation easier. Users shouldn't see weird code because the templates for that version were easier to write (I admit this one is up to debate and probably not that important).
3. Most of our codebase does everything directly, and consistency is better than a small improvement that breaks it.
4. Finally, and it looks like this is subjective, I find the normal templates easier to read:
    - Without the indirection, I only deal with the templates. The rest of the code looks as I would write it if templates weren't a thing
    - With the indirection, I deal with the templates, but also with a more complicated "runtime pattern." I wouldn't write the code like that if templates weren't a thing so it's more difficult to reason about it (similar to point 2). Plus, I have to jump around the file while reading it.

waspc/data/Generator/templates/sdk/wasp/server/dbClient.ts
sodic: But we do have it (or I don't understand what you mean).

Here's the generated code, the user input is aliased into `name_ext`:
```ts
import { setUpPrisma as setUpPrisma_ext } from 'wasp/src/prisma'
import { PrismaClient as InternalPrismaClient } from '@prisma/client'

const _waspSetUpPrisma = setUpPrisma_ext

// We are not typing the return value because 
// we want to infer the type from the (potentially)
// user-defined function _waspSetUpPrisma
function createDbClient() {
  return _waspSetUpPrisma()
}

const dbClient = createDbClient()

// PUBLIC API 
export type PrismaClient = typeof dbClient

// PUBLIC API
export default dbClient
```

waspc/data/Generator/templates/sdk/wasp/server/dbClient.ts
```diff
@@ -1,9 +1,22 @@
 {{={= =}=}}
+{=# prismaSetupFn.isDefined =}
+{=& prismaSetupFn.importStatement =}
+{=/ prismaSetupFn.isDefined =}
 {=# areThereAnyEntitiesDefined =}
-import Prisma from '@prisma/client'
+import { PrismaClient as InternalPrismaClient } from '@prisma/client'
```
sodic: This import is only necessary when the setup function isn't defined.

waspc/data/Generator/templates/sdk/wasp/server/index.ts
```diff
@@ -13,5 +14,7 @@ export { MiddlewareConfigFn } from './middleware/index.js'
 // PUBLIC API
 export { env } from './env.js'
 
+// PUBLIC API
+
```
sodic: Something weird happening here.

waspc/data/Generator/templates/sdk/wasp/server/index.ts
```diff
@@ -1,9 +1,10 @@
-import type { PrismaClient } from '@prisma/client'
+import type { PrismaClient } from './dbClient.js'
+
```
sodic, line 2: Extra empty line.

waspc/packages/wasp-config/src/mapUserSpecToAppSpecDecls.ts
```diff
@@ -316,9 +316,10 @@ export function mapPasswordReset(
 }
 
 export function mapDb(db: User.DbConfig): AppSpec.Db {
-  const { seeds } = db;
+  const { seeds, prismaSetupFn } = db;
   return {
     seeds: seeds?.map(mapExtImport),
+    prismaSetupFn: prismaSetupFn && mapExtImport(prismaSetupFn),
```
sodic, line 322: Let's add this guy to `AppSpec/FromJSONTest.hs`.

sodic (commented):
Good stuff, a couple of unresolved things left. Should be quick :)

waspc/examples/todoApp/headless-tests/tests/prisma-setup-fn.spec.ts
sodic: Hm, thinking about it again, this one is actually fine for now.

The serialization one was more of a hack. Anyway, we can resolve this. I may create an issue or wait for it to pop up again.

waspc/data/Generator/templates/sdk/wasp/server/dbClient.ts
infomiho: I must have been operating on old assumptions, sorry about that! There really isn't any need for this `_wasp` weirdness anymore 👍

waspc/data/Generator/templates/sdk/wasp/server/dbClient.ts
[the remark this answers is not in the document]
infomiho: I guess this was too simple of a place to use the indirect pattern to see the benefits. Sometimes, I had to make a decision between "inline templating" vs. "templating blocks of code" and that's where I picked up the indirection preference. 

Inline templating:
```
type {=# condition =}Some{=/ condition =}{=^ condition =}Other{=/ condition =}Stuff = {}
```
Templating blocks of code:
```
{=# condition =}
type SomeStuff = {}
{=/ condition =}
{=^ condition =}
type OtherStuff = {}
{=/ condition =}
```

Sometimes for readability it made sense. But here we have the extra runtime component, which I admit, is affecting the end code that users will maybe read.

I've modified this template to be without indirection and organised the imports better.

waspc/data/Generator/templates/sdk/wasp/server/index.ts
infomiho: It's just me being careless, sorry about that!

waspc/packages/wasp-config/src/mapUserSpecToAppSpecDecls.ts
infomiho: I see that we already have a test for `ExtImport` and this is using that same primitive. I don't see that we have tests in that file for the `Db` or `App`, so I've went with `Db` and not the whole `App` declaration, does that make sense? [`484f0f6` (#2693)](https://github.com/wasp-lang/wasp/pull/2693/commits/484f0f64ade89b2de2ea0f937b804b49ce544054)

Should we add a note to update these tests when editing the TS Spec because I wasn't even aware of them.

infomiho: I've asked about this on Discord: https://discord.com/channels/686873244791210014/920312576133443634/1379757010178015263

waspc/data/Generator/templates/sdk/wasp/server/dbClient.ts
```diff
@@ -1,10 +1,17 @@
 {{={= =}=}}
 {=# areThereAnyEntitiesDefined =}
-import Prisma from '@prisma/client'
-
-function createDbClient(): Prisma.PrismaClient {
-  return new Prisma.PrismaClient()
+{=# prismaSetupFn.isDefined =}
+{=& prismaSetupFn.importStatement =}
+function createDbClient() {
```
sodic: I think there's an empty line missing between the import and the function.

waspc/data/Generator/templates/sdk/wasp/server/dbClient.ts
```diff
@@ -1,10 +1,17 @@
...
+  return {= prismaSetupFn.importIdentifier =}()
+}
+{=/ prismaSetupFn.isDefined =}
+{=^ prismaSetupFn.isDefined =}
+import { PrismaClient as InternalPrismaClient } from '@prisma/client'
+function createDbClient() {
```
sodic: Another empty line mising here. No big deal, but it's easy to keep tidy so why not.

waspc/test/AppSpec/FromJSONTest.hs
```diff
@@ -293,6 +295,20 @@ spec_AppSpecFromJSON = do
                       }
               }
           )
+  describe "Db" $ do
```
sodic, line 298: Whoa, I was missing the whole DB... Thanks for adding it in!

sodic (approved):
Ready for merge, just left a couple of small cosmetic suggestions.

Issue #2700, opened 2025-04-25
Serialize `Prisma.Bytes` data type


cprecioso:
Same as `Prisma.Decimal` (#2503), we can add support for [`Prisma.Bytes`](https://www.prisma.io/docs/orm/prisma-client/special-fields-and-types#working-with-bytes) to our serialization so that we cover all the Prisma custom types.

Blocked by upgrading to Prisma v6 (#2504) because `Prisma.Bytes` in v5 is converted to a Node.js `Buffer`, which does not exist in the browser. In Prisma v6, it is converted to `Uint8Array`, which does exist in the browser.

cprecioso:
Working on top of #2701, we just need to add a new case at this file:

https://github.com/wasp-lang/wasp/blob/cprecioso/2503-support-prismas-decimal-field-in-our-serialisation/waspc/data/Generator/templates/sdk/wasp/core/serialization.ts

PR #2701, opened 2025-04-25
Add serialization for `Prisma.Decimal`

### LINKED ISSUE #2503: Support Prisma's Decimal field in our serialisation
Prisma has a Decimal type which SuperJSON can't handle, we need to implement custom handling for it. 

> We should probably implement a custom serializer for the Prisma.Decimal type: https://github.com/flightcontrolhq/superjson?tab=readme-ov-file#decimaljs--prismadecimal

Source: https://ptb.discord.com/channels/686873244791210014/1337972417171165296/1338461339898740757

cprecioso:
### Description

Fixes #2503

> I also wanted to add support for [`Prisma.Bytes`](https://www.prisma.io/docs/orm/prisma-client/special-fields-and-types#working-with-bytes) but it needs to wait until we upgrade to Prisma v6 (#2700)

This centralizes our SuperJSON types and adds a custom serialization for `Prisma.Decimal`.
It is based on [their recipe](https://github.com/flightcontrolhq/superjson#decimaljs--prismadecimal), but I updated it to work with `Prisma.Decimal` instead of needing to add another dependency.

1. [x] I updated [`ChangeLog.md`](https://github.com/wasp-lang/wasp/blob/main/waspc/ChangeLog.md) with description of the change this PR introduces.

### Add a regression test if needed

If you did a **bug fix**, make sure you satisfy the following:

1. [ ]  I added a regression test that reproduces the bug and verifies the fix.

If you're unable to add a regression test, please explain why. 
This likely indicates that our current testing setup needs improvement.

If you did code changes and **added a new feature**, make sure you satisfy the following:

1. [ ] I updated [`waspc/examples/todoApp`](https://github.com/wasp-lang/wasp/tree/main/waspc/examples/todoApp) and its e2e tests as needed and manually checked it works correctly.

If you did code changes and **updated an existing feature**, make sure you satisfy the following:

1. [ ] I updated [`waspc/examples/todoApp`](https://github.com/wasp-lang/wasp/tree/main/waspc/examples/todoApp) and its e2e tests as needed and manually checked it works correctly.
2. [ ] I updated [`waspc/examples/crud-testing`](https://github.com/wasp-lang/wasp/tree/main/waspc/examples/crud-testing) as needed and manually checked it works correctly.
3. [ ] I updated [`waspc/examples/todo-typescript`](https://github.com/wasp-lang/wasp/tree/main/waspc/examples/todo-typescript) as needed and manually checked it works correctly.
4. [ ] I updated [`waspc/examples/pg-vector-example`](https://github.com/wasp-lang/wasp/tree/main/waspc/examples/pg-vector-example) as needed and manually checked it works correctly.

### Update starter apps if needed

If you did code changes and **updated an existing feature**, make sure you satisfy the following:

1. [ ] I updated [starter skeleton](https://github.com/wasp-lang/wasp/tree/main/waspc/data/Cli/templates/skeleton) as needed and manually checked it works correctly.
2. [ ] I updated [`basic` starter](https://github.com/wasp-lang/wasp/tree/main/waspc/data/Cli/templates/basic) as needed and manually checked it works correctly.
3. [ ] I updated [`todo-ts` starter](https://github.com/wasp-lang/starters/tree/dev/todo-ts) as needed and manually checked it works correctly.
4. [ ] I updated [`embeddings` starter](https://github.com/wasp-lang/starters/tree/dev/embeddings) as needed and manually checked it works correctly.
5. [ ] I updated [`saas` starter](https://github.com/wasp-lang/open-saas/tree/main/template) as needed and manually checked it works correctly.


cprecioso:
I think this should be in the `todoApp` so we test it somehow, but I don't see it really fitting anywhere, what to do?

waspc/data/Generator/templates/sdk/wasp/core/serialization.ts
```diff
@@ -1,3 +1,31 @@
+import { Prisma } from '@prisma/client'
+import type { Decimal as DecimalType } from "@prisma/client/runtime/library"
+import SuperJSON from 'superjson'
```
sodic, line 3: Can we import just the stuff we need?

```typescript
import { SuperJSON, serialize, deserialize } from 'superjson'
```

We usually try to avoid default exports and improts, reasoning here: #546

Plus, if we do it like this, we won't have to deal with bindings and `this` shennanigans - Superjson already did that for us.

waspc/data/Generator/templates/sdk/wasp/core/serialization.ts
```diff
@@ -41,3 +70,7 @@ interface SuperJSONArray extends Array<SuperJSONValue> {}
 interface SuperJSONObject {
   [key: string]: SuperJSONValue
 }
+
+
+export const serialize = SuperJSON.serialize.bind(SuperJSON)
+export const deserialize = SuperJSON.deserialize.bind(SuperJSON)
```
sodic, line 76: If we must do this, I recommend an arrow function + `typeof` to ensure the type is preserved, something like:
```typescript
export const serialize: typeof SuperJSON.serialize = (...args) => SuperJSON.serialize(...args)
```
But I think there's a simpler solution:
1. Import `serialize` and `deserialize` functions. They're independent of the class, and SuperJSON already does the binding internally.
2. Reexport them.

waspc/data/Generator/templates/sdk/wasp/core/serialization.ts
```diff
@@ -41,3 +70,7 @@ interface SuperJSONArray extends Array<SuperJSONValue> {}
 interface SuperJSONObject {
   [key: string]: SuperJSONValue
 }
+
```
sodic, line 73: Extra empty line here... man I miss having a linter :crying_cat_face:

waspc/data/Generator/templates/sdk/wasp/client/operations/internal/index.ts
```diff
@@ -1,19 +1,16 @@
...
+import { serialize, deserialize } from 'wasp/core/serialization'
 
 // PRIVATE API
 export type OperationRoute = { method: HttpMethod.Post, path: string }
 
 // PRIVATE API
 export async function callOperation(operationRoute: OperationRoute, args: any) {
   try {
-    const superjsonArgs = superjsonSerialize(args)
+    const superjsonArgs = serialize(args)
```
sodic, line 11: Let's change the variable name too (to completely decouple the code of SuperJson).

sodic (changes requested):
Great work!

I love how you've limited SuperJson into a single module of our codebase (`serialization.ts`). That's an excellent improvement! It will help us decorate and/or swap it out in the future. 

Left some improvement suggestions, but the core thing is solid.

## Adding a test for it
> I think this should be in the todoApp so we test it somehow, but I don't see it really fitting anywhere, what to do?

Nothing in the `todoApp` meaningfully fits together :smile: 
It's just a big feature dump. So, don't concern yourself too much with trying to make it make sense. Just put it somewhere and add a test for it.

For example, you could turn `getDate` into an action that receives and returns an object with a bunch of hardcoded serializable types (`Set`, `Date`, `Decimal`, etc), render them, and then write a UI test that checks whether they were transported successfully.

waspc/src/Wasp/Generator/SdkGenerator.hs
```diff
@@ -162,9 +162,6 @@ genEntitiesAndServerTypesDirs spec =
         [relfile|server/_types/taggedEntities.ts|]
         [relfile|server/_types/taggedEntities.ts|]
         (Just $ object ["entities" .= allEntities])
-    serializationFileDraft =
-      C.mkTmplFd
-        [relfile|server/_types/serialization.ts|]
```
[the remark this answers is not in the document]
cprecioso: I did a search but restricted it to `*.hs` files 😂

waspc/data/Generator/templates/sdk/wasp/package.json
```diff
@@ -16,6 +16,8 @@
       "./core/stitches.config": "./dist/core/stitches.config.js",
       {=!  Used by our code, uncodumented (but accessible) for users. =}
       "./core/storage": "./dist/core/storage.js",
+      {=!  Used by our code, uncodumented (but accessible) for users. =}
+      "./core/serialization": "./dist/core/serialization.js",
```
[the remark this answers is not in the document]
cprecioso: @sodic should I add it now or when the PR is merged?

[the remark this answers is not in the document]
sodic: If we're definitely merging the PR, you can add it right away.

waspc/data/Generator/templates/sdk/wasp/core/serialization.ts
cprecioso: This and the re-binding stuff was because I was afraid that the custom class registry was per-instance and would not apply to the default thing, but I saw now that inside their code they always use a default instace, so we can indeed be simpler.

sodic: Wow, nice thoroughness! 

I admit that didn't occur to me at all. Glad it turned out to be possible :)

waspc/examples/todoApp/main.wasp
```diff
@@ -239,6 +239,14 @@ query getDate {
...
 }
 
+query getSerializedObjects {
+  fn: import { getSerializedObjects } from "@src/testTypes/operations/definitions"
+}
+
+route SerializationRoute { path: "/serialization", to: SerializationPage }
+page SerializationPage {
+  component: import { SerializationPage } from "@src/pages/Serialization"
+}
```
sodic, line 249: Can we add a link to this from the user's homepage, for easier manual troubleshooting (if we keep it, more on this later)?

waspc/examples/todoApp/src/pages/Serialization.tsx
```diff
@@ -0,0 +1,35 @@
...
+              case 'map':
+                return value instanceof Map
+              case 'decimal':
+                return value instanceof Prisma.Decimal
+              case 'regexp':
+                return value instanceof RegExp
+              default:
+                return false
+            }
+          }
```
sodic, line 28: Instead of just checking for instances, we should do a robust check for full equality and answer the question _"Is the delievered object the same as if it were created on the client?"_

We can do that by:
- Extracting the value `serializaedObjects` into its own file (e.g., `fixtures.ts`).
- Returning this value from the operation, and importing it on the client.
- Setting an exact `Date` with `new Date(...)` to avoid slight differences in instantiation times. 

This file could then look something like this:
```typescript
import { getSerializedObjects, useQuery } from 'wasp/client/operations'
import { serializableObjects as expectedSerializableObjects } from '../testTypes/operations/fixtures'

export const SerializationPage = () => {
  const { data: receivedSerializedObjects } = useQuery(getSerializedObjects)

  if (!receivedSerializedObjects) {
    return null
  }
  
  // I assume one of our libraries has an `isEqual` method.
  // I'm very rusty with normal development, so don't how actually do a proper 
  // object comparison these days.
  //  I'm assuming you do :)
  return isEqual(receivedSerializedObjects, expectedSerializableObjects)
    ? 'Superjson serialization working.' : 'Superjson serialization error.'
}
```

But this doesn't even need to be a headless "e2e" test. It can just be a regular test testing our RPC: import function, import expected value, compare them. We could then remove the `SerializationPage` altogether.

Again, I'm not sure how to best do normal tests in our current setup, but I'm assuming either you or @infomiho do. 

P.S. I know I was the one who suggested a UI test, sorry for the red herring :grimacing:
We can keep this as a UI test if it's more practical, but it should still test values.

waspc/examples/todoApp/src/testTypes/operations/definitions.ts
```diff
@@ -103,6 +105,22 @@ export const getDate: GetDate<void, Date> = async () => {
   return new Date()
 }
 
+const makeSerializableObjects = () => ({
+  date: new Date(),
+  set: new Set(['foo']),
+  map: new Map([['foo', 'bar']]),
+  decimal: new Prisma.Decimal(10),
+  regexp: /foo/,
+})
```
sodic, line 114: How come this is a function? Why not just `const serializableObjects = ...`?

waspc/examples/todoApp/src/pages/Serialization.tsx
sodic, line 1: Our directory structure is pretty messy so we won't be cleaning all that up know.

Still, since we're now testing values (not only types), let's rename the folder `testTypes` to `rpcTests` and maybe put the `SerializationPage` inside (if we end up keeping it). What do you think?

sodic (commented):
Nice work, I love the explanations!

The code is fully ready for shipping. I'd only like us to improve the tests.

waspc/examples/todoApp/src/pages/Serialization.tsx
```diff
@@ -0,0 +1,35 @@
+import { Prisma } from '@prisma/client'
+import { getSerializedObjects, useQuery } from 'wasp/client/operations'
+
+export const SerializationPage = () => {
+  const { data: serializedObjects } = useQuery(getSerializedObjects)
+  console.log({ serializedObjects })
```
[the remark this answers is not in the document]
cprecioso: Yes, in case we want to inspect it in the console, given this is really a test page and nothing else 😅

waspc/examples/todoApp/src/pages/Serialization.tsx
cprecioso: I wouldn't rename the `testTypes` folder as there is a file actually testing the types: https://github.com/wasp-lang/wasp/blob/82a6e6ad8885eec4e4255c6ed7fdcd6a519d7843/waspc/examples/todoApp/src/testTypes/operations/client.ts

But I can write the new one in `rpcTests`

cprecioso:
@sodic ready for next review

waspc/examples/todoApp/headless-tests/tests/crud.spec.ts
```diff
@@ -42,6 +42,16 @@ test.describe('CRUD test', () => {
...
+
+  test('serialization of superjson-supported types works', async ({ page }) => {
+    await page.goto('/serialization')
+
+    await page.waitForSelector('#serializedObjects')
+
+    await expect(page.locator('#serializedObjects')).toContainText(
+      'All serialized objects are of the expected types.'
+    )
+  })
```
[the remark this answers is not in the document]
sodic: But from what I see, these tests the specific declaration, `crud`, whose values are rendered on the `CrudTestPage`.
"Crud" signifies the specific wasp feature of the same name, not a general create-read-update-delete pattern

waspc/examples/todoApp/src/pages/Serialization.tsx
sodic: > I wouldn't rename the testTypes folder as there is a file actually testing the types:

There is, true. Even more than one file - `testTypes/operations/{client,server},ts` also test types. But they're all testing RPC types.

RPC is a feature, types and values are implementation details.
So both `testTypes` and `rpcTests` can be true at the same time (and they are). I find it better to name tests after a feature they test than after a part of the language (e.g., we wouldn't call them "function tests" ). Which is pretty ironic since I'm the clown who came up with the `testTypes` name :upside_down_face: 

If the files in `testTypes` tested something other than RPC types, then we'd have a problem. But the good news is they don't :smile:. If we still wanted to somehow say "Hey, stuff in here only tests types," we could keep the `testTypes` as a subfolder of `rpcTests`. But I don't think that's necessary.

Hope my reasoning makes sense. I'll let you choose what to do and resolve when you are ready.

sodic (approved):
Great work, Carlos!

I'm approving so you feel free to merge whenever you're happy.

I'd still like us to take care of these two comments but I won't to block the PR because of them:
- https://github.com/wasp-lang/wasp/pull/2701#discussion_r2081146085
- https://github.com/wasp-lang/wasp/pull/2701#discussion_r2081161290

cprecioso:
Addressed. I'll wait until the CI passes and merge. Thanks @sodic!

waspc/examples/todoApp/src/rpcTests/operations/fixtures.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const SERIALIZABLE_OBJECTS_FIXTURE = {
+  date: new Date(1746538539344),
+  set: new Set(['foo']),
+  map: new Map([['foo', 'bar']]),
+  regexp: /foo/,
+  nan: NaN,
+  inf: Infinity,
+  negInf: -Infinity,
+  // we ensure that it'd be too big to be represented as a `number`
```
sodic, line 11: Only seeing this now, nice!

Issue #1883, opened 2024-03-12
Implement `wasp build start` command that runs output of `wasp build`


Martinsos:

sodic:
# How to test whether a Wasp app builds and runs correctly
I'll paste my current setup here (as discussed on [Discord](https://discord.com/channels/686873244791210014/1231942588588752947/1231951225306746910)).

## Usage

After the setup, testing a local deployment becomes as simple as:

1. Run `wasp build` in the root of your project.
2. Run `srun <instanceName>` from the project root to build and start the server in a Docker container. Choose anything you want for hte `instanceName`
3. Open a new terminal and run `crun` from the project root to build and start the client.

## Setup
### Step 1: The database
I recommend you keep a PostgreSQL db running in the backgorund at all times. It's easier than `wasp start db` because:
- It's project-agnostic.
- You can name your own databases (which saves us from ever having to worry about the `DATABASE_URL`).

You set it once and forget it (until you shut down your computer, but Docker can probably take care of that too). Setting it up is pretty easy:
1. Make sure you have docker running.
2. Free up port `5433` (The port is different than `5432` used by `wasp start db` on purpose so you can keep using that when you want).
3. Execute the following command in a terminal somewhere and leave it running (you want need `sudo` if you're on a Mac):
```bash
sudo docker run --rm --publish 5433:5432 -v my-app-data:/var/lib/postgresql/data --env POSTGRES_PASSWORD=devpass1234 postgres
```

### Step 2: The `srun` script

This is the command we use to build and run the server using Docker. Again, you can remove all the `sudo`s if you're on a Mac.

Put this script somewhere in your path and name it `srun` (I have all my personal scripts in `~/bin`, which I make sure is in my path):
```bash
#!/bin/bash
app_name="$1"
sudo docker build -t "$app_name" .wasp/build/ && sudo docker run \
  --env DATABASE_URL="postgresql://postgres:devpass1234@localhost:5433/${app_name}-srun-db" \
  --env WASP_WEB_CLIENT_URL='http://localhost:3000' \
  --env JWT_SECRET='your-goto-jwt-secret' \
  --env GOOGLE_CLIENT_SECRET="your-goto-google-client-secret" \
  --env GOOGLE_CLIENT_ID="your-goto-google-client-id" \
  --env KEYCLOAK_CLIENT_ID="your-keycloak-client-id" \
  --env KEYCLOAK_CLIENT_SECRET="your-keycloak-client-secret" \
[...]

PR #2706, opened 2025-04-28
Run `wasp build` output with `wasp-app-runner`

### LINKED ISSUE #2663: Run Playwright tests for wasp example apps when run from output of `wasp build`
Related (and possibly a prerequisite): https://github.com/wasp-lang/wasp/issues/1883 .

infomiho:
Closes #2663

It does the following:
- adds a concept of `mode` to the `wasp-app-runner`:
  1. `mode=dev`: runs the app using `wasp start` (we had this before)
  2. `mode=build`: runs the `wasp build` artefacts (new)
     - builds the server with `Docker` and runs it
     - builds the client using `vite` and serves it with `npx serve`
     - it provides the required env vars (client url, server url, database url) + the `.env.server` env vars to make it easy to run the server locally

In a separate PR, I've enabled the headless tests on the `todoApp` with both modes: https://github.com/wasp-lang/wasp/pull/2707


wasp-app-runner/package.json
```diff
@@ -12,12 +12,11 @@
     "prepublishOnly": "npm ci && npm run build"
   },
   "dependencies": {
-    "@inquirer/prompts": "^7.4.1",
-    "typescript": "^5.7.3",
-    "yargs": "^17.7.2"
+    "@commander-js/extra-typings": "^13.1.0",
+    "commander": "^13.1.0",
```
infomiho, line 16: Using `commander` since we use it in our deploy scripts to reduce the number of different libs any team member needs to know.

wasp-app-runner/src/build/server.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  runServerAppContainer({
+    pathToApp,
+    imageName,
+    containerName,
+    extraEnv,
+  });
+}
+
+function buildServerAppContainer({
```
infomiho, line 46: Based on #1883 comments by @sodic

wasp-app-runner/src/process.ts
```diff
@@ -31,6 +31,9 @@ class ChildProcessManager {
   }
 
   private cleanExit(reason: string) {
+    if (this.children.length === 0) {
```
infomiho, line 34: We'd get a cleanup warning when running e.g. `wasp-app-runner --help` which looked silly. It also makes sense not to clean up if there isn't anything to clean up.

wasp-app-runner/src/build/client.ts
```diff
@@ -0,0 +1,33 @@
+import * as path from "path";
+import { spawnWithLog } from "../process.js";
+
+export function buildClientApp({
+  pathToApp,
+}: {
+  pathToApp: string;
```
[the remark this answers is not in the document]
FranjoMindek: Well, no need if we unify this, see `build/index.ts`.

wasp-app-runner/src/build/server.ts
```diff
@@ -0,0 +1,138 @@
...
+}
+
+function getServerAppEnvArgs({
+  pathToApp,
+  extraEnv,
+}: {
+  pathToApp: string;
+  extraEnv: EnvVars;
+}): string[] {
+  const requiredEnv: EnvVars = {
```
FranjoMindek, line 103: If those can be overwritten, maybe `defaultRequiredEnv`?

wasp-app-runner/src/build/server.ts
```diff
@@ -0,0 +1,138 @@
...
+      "--network",
+      "host",
+      "--name",
+      containerName,
+      imageName,
+    ],
+  });
+}
+
+function getServerAppEnvArgs({
```
FranjoMindek, line 96: Maybe `getServerAppContainerEnvArgs`? Or anything that signals it works with docker.

wasp-app-runner/src/build/server.ts
```diff
@@ -0,0 +1,138 @@
...
+    const stats = fs.statSync(envFilePath);
+    if (stats.isFile()) {
+      return ["--env-file", envFilePath];
+    }
+  } catch (err) {}
+
+  return [];
+}
+
+function envVarsToArgs(envVars: EnvVars): string[] {
```
FranjoMindek, line 134: Also uses docker env option syntax, should be reflected in the name.

wasp-app-runner/src/index.ts
```diff
@@ -37,53 +43,75 @@ async function runWaspApp({
...
   return {
-    pathToApp: argv["path-to-app"],
-    waspCliCmd: argv["wasp-cli-cmd"],
+    mode: args[0],
+    pathToApp: options.pathToApp,
+    waspCliCmd: options.waspCliCmd,
   };
 }
+
+function isModeArg(args: string[]): args is [Mode] {
```
FranjoMindek, line 115: `isModeArg` seems weird, since "args" is plural and it's converted to singular "arg".

I would make this function handle `wasp-app-runner` args in general, and make it's name reflect that (even though there is only a single argument for now). Which we kinda already do (`.length === 1`).

I don't know how much the library does for us here.
E.g. is `isModeArg` here only for types sake because `Commander` already validates, or is it necessary?

FranjoMindek: Now I see: https://www.npmjs.com/package/commander#more-configuration-1 and "Custom argument processing" below it.
We could make it handle validation for ourselves, do it declaratively instead of imperatively.

FranjoMindek (changes requested):
I really like that we are getting this change.
I've got mostly naming issues. A bit of CLI parsing suggestion too.

wasp-app-runner/README.md
```diff
@@ -24,19 +29,30 @@ wasp-app-runner [--path-to-app <path>] [--wasp-cli-cmd <command>]
...
 
+### Env variables
+
+When using the `dev` mode:
+
+- If the app you are trying to run has `.env.server` or `.env.client` env files defined, the `wasp start` will use them as expected.
+
+When using the `build` mode:
+
+- `wasp-app-runner` will use the `.env.server` file when running the server container with Docker locally (which is not done usually by Wasp).
```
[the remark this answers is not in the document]
infomiho: I guess there is no reason why not to include the `.env.client` env vars for the client build. I'll add it :)

wasp-app-runner/src/build/index.ts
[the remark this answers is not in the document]
infomiho: Yep, this was a mess, I've updated it to be db -> server (single fn) -> client (single fn).

wasp-app-runner/src/build/client.ts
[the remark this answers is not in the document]
infomiho: I'll refactor this a bit to use branded types for the `PathToApp` string in a separate PR, but I don't think we benefit from extracting this into an interface (used locally, simple interface).

wasp-app-runner/src/index.ts
infomiho: Perfect, I couldn't find that nice argument construction option in the docs. I'll do that, that's ideal.

wasp-app-runner/src/build/client.ts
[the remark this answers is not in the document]
FranjoMindek: I wrote comments as I've passed, as I've read more of PR I've also opted out of interfacing this.

wasp-app-runner/src/build/server.ts
```diff
@@ -0,0 +1,143 @@
...
+  };
+  return [
+    ...getEnvVarsDockerArgs({ ...defaultRequiredEnv, ...extraEnv }),
+    ...getDevEnvFileDockerArg({
+      pathToApp,
+    }),
+  ];
+}
+
+function getEnvVarsDockerArgs(envVars: EnvVars): string[] {
```
FranjoMindek, line 124: Get is a bit weird as it maps/transforms?

FranjoMindek (changes requested):
Nice improvements. Just a small naming change.

FranjoMindek (changes requested):
Nice improvements. Just a small naming change.

wasp-app-runner/src/build/client.ts
```diff
@@ -0,0 +1,82 @@
...
+}: {
+  pathToApp: PathToApp;
+}): Promise<{
+  exitCode: number | null;
+}> {
+  return spawnWithLog({
+    name: "client-start-app",
+    cmd: "npx",
+    args: ["serve", "--single", "-p", "3000"],
+    cwd: path.join(pathToApp, ".wasp/build/web-app/build"),
```
sodic, line 66: We're 100% going to forget to change this (and other hardcoded strings) at one point, but I see no better solution :smile: 


One thing we could improve at least: `.wasp/build/web-app` is mentioned in two places. I recommend extracting that into a constant (inside this file).

wasp-app-runner/src/build/client.ts
```diff
@@ -0,0 +1,33 @@
+import * as path from "path";
```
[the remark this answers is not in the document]
sodic: I recommend linking this issue in a code comment too. So the reader can know where the procedure comes from and how to emulate it outside of `wasp-app-runner`.

wasp-app-runner/src/build/server.ts
```diff
@@ -0,0 +1,145 @@
...
+  });
+
+  const { exitCode: buildExitCode } = await buildServerAppContainer({
+    pathToApp,
+    imageName,
+  });
+
+  if (buildExitCode !== 0) {
+    log(
+      "server-build-app",
```
sodic, line 36: About the log strings (e.g., `server-build-app`, `client-start-app`)...

I have a feeling these logs should be inside their respective functions (e.g., `server-build-app` failure inside `buildServerAppContainer`). The prefix string defined inside that function and then duplicated here is a good indicator for this. I believe log prefixes shouldn't span across levels (the prefix should generally tell us which function it comes from) so, if this log really belongs here, it shuld probably have a different prefix.

If we really do need the duplication, we should probably extract a constant (although the necessary importing/exporting of said constant across hierarchies would another indicator that something is off).

wasp-app-runner/src/build/server.ts
sodic: Nice comment, Franjo.

Additional thought, the array of strings is an unfortunate quirk of passing command line args, so I advise keeping it as local as possible.

Specifically, instead of letting quirk spread through our logic, we should use the best possible type for describing the arguments (which is an object) and transform them into an array of strings as late as possible.

In this case:
```ts
const devEnvFile = getDevEnvFilePath(...)
return [
  ...getEnvVarsDockerArgs({ ...defaultRequiredEnv, ...extraEnv }),
  devEnvFile && getDevEnvFileDockerArg({
  pathToApp,
}),

getDevEnvFilePath(...): string | null

**Review by sodic — COMMENTED** [review]
Looks great, I love how it's organized.

Left some commments, let me know what you think!

FYI: I didn't run anything

File: wasp-app-runner/README.md [hunk 14]
```diff
@@ -10,12 +13,14 @@ npm install
...
+wasp-app-runner run <mode> [--path-to-app <path>] [--wasp-cli-cmd <command>]
 ```
 
-### Arguments Table
+### Options
+
+You must pass the `<mode>` as an argument, which can be either `dev` or `build`.
 
-| Argument         | Description                                            | Example         |
+| Option           | Description                                            | Example         |
```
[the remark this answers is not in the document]
FranjoMindek: I've actually deeply dived into this topic because I also thought so.
But options seems to be prevalent in modern times, especially with CLI libs.

examples/todo-typescript/headless-tests/playwright.config.ts
```diff
@@ -44,7 +44,7 @@ export default defineConfig({
 
   /* Run your local dev server before starting the tests */
   webServer: {
-    command: "wasp-app-runner --path-to-app=../",
+    command: "wasp-app-runner run dev --path-to-app=../",
```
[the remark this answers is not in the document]
infomiho: Hm, I guess I was going for `npm run dev` and `npm run prod` style. You pose a good question if we should remove it.

wasp-app-runner/src/files.ts
```diff
@@ -0,0 +1,6 @@
+import * as fs from "fs";
```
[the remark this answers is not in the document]
infomiho: I'd say by convention, I'm used to importing it like this. Interestingly enough, you didn't ask me about the `import * as path from "path";` import.

infomiho:
@sodic ready for another look, implemented all the suggestions - the API is the only open question: https://github.com/wasp-lang/wasp/pull/2706#discussion_r2074952897

examples/todo-typescript/headless-tests/playwright.config.ts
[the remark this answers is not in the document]
sodic: First on's my favorite too.

wasp-app-runner/src/build/client.ts
sodic: I'd put them at the bottom of the file, but not that important. Resolve when happy.

wasp-app-runner/src/build/client.ts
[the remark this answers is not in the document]
sodic: I saw the two places you added them to, nice.

I'd also recommend one final place (arguably the most important one): above `startAppInBuildMode`.

infomiho:
@sodic updated the command to look like `run-wasp-app dev` and `run-wasp-app build`

wasp-app-runner/src/build/server.ts
sodic: All the duplication is now localized and easy to spot, and we can extract the prefixes into constants.

Most function just have two repetition, but the biggest offender I've found is `wasp-info`:
```ts
export async function waspInfo(...) {
  const logPrefix = 'wasp-info'
  // ...
}
```

sodic (approved):
Great stuff!

I'd just like us to clean up those repetitions and it's good to go. I'm approving, no need to check back with me before merge.

wasp-app-runner/src/build/server.ts
infomiho: I'll sort this out in a separate PR 👍

infomiho: Done in #2765

PR #2707, opened 2025-04-28
Run headless tests on `todoApp` production build

### LINKED ISSUE #884: Create e2e test that tries to run Dockerfile Wasp creates via `wasp build`
Right now we have an e2e test for Wasp CLI that tries building Dockerfile that is created via `wasp build`. However, we don't have a test that then tries running that Dockerfile!
And we should have it, it happened already that it builds but it doesn't run.

This is a bit more work because then we also need to set up the database (postgre) and run it, otherwise we can't run Dockerfile since it requires it. So we should write a test that spins up postgre db and then runs Dockerfile and confirms that server successfully started. Which can also be a bit tricky because it is a bit hard to say when server is done, since it keeps running forever, but we could try curling it maybe and if it gives response in some reasonable time, we conclude it succeeded or smth like that.

This really comes down to actually writing a test that is capable of running wasp app generated by Wasp and checking that it works as it should. In some way this is basis for having e2e tests not just for Wasp CLI, but for the app that Wasp CLI generates. The only difference is if that app is spun up from the Dockerfile, or via `wasp start` -> so from "build" artifacts, or in dev mode. But for both we need to set up a postgre db (to do proper testing) and we need a way to detect that it successfully started running.

infomiho:
Updates the `todoApp` to be able to run headless tests in the CI.

Closes #884


waspc/examples/todoApp/headless-tests/tests/helpers.ts
FranjoMindek, line 1: Should we run this test only in production code then?

waspc/examples/todoApp/main.wasp
FranjoMindek, line 1: I wanted to say should we compile this part of the application only when running production code headless tests, but little hard to do without `main.wasp.ts`. Looking forward to when we transition.

waspc/examples/todoApp/src/auth/headless.ts
```diff
@@ -0,0 +1,51 @@
...
+  getProviderDataWithPassword,
+  updateAuthIdentityProviderData,
+} from 'wasp/auth/utils'
+import { VerifyAllUserEmails } from 'wasp/server/operations'
+
+/*
+  When we run headless tests on production code that doesn't auto verify emails,
+  we need to manually verify all user emails.
+*/
+export const verifyAllUserEmails: VerifyAllUserEmails = async (
```
FranjoMindek, line 12: Possible to do in a single query, but I suppose we don't need it for tests, since there won't be many users.


e.g. 
```ts
await prisma.$executeRaw`
  UPDATE auth_identity
  SET provider_data = jsonb_set(provider_data, '{isEmailVerified}', 'true')
  WHERE provider_name = 'email';
  ```

Comment (FranjoMindek) [reply waspc/examples/todoApp/src/auth/headless.ts:12]: Not sure how prisma handles JSONB, if we wanted to do it the entities way.

**Review by FranjoMindek — CHANGES_REQUESTED** [review]
Small nudges.

File: waspc/examples/todoApp/src/auth/headless.ts (hunk 1 above)
Comment (infomiho) [reply waspc/examples/todoApp/src/auth/headless.ts:12]: That would be great way to do it if the `providerData` wasn't a `String` (due to Prisma not supporting JSON files for SQlite at the time when we implemented the auth schema).

File: waspc/examples/todoApp/headless-tests/tests/helpers.ts
Comment (infomiho) [reply waspc/examples/todoApp/headless-tests/tests/helpers.ts:1]: Are you saying that we shouldn't run this on production code? 

We can't really run any tests without logging in first. And I think we should be running tests on the production builds because we do it manually. 

The problem here is the email verification flow, so either we "brute force" it like we did here or we come up with a way to have a local SMTP server to get the verification email and click on the verify link. Now that I said it - let me see if I can get that up and running.

Comment (infomiho) [reply waspc/examples/todoApp/headless-tests/tests/helpers.ts:1]: I've added a local SMTP server to avoid modifying the `todoApp` to be able to verify the email: [`cc4de83` (#2707)](https://github.com/wasp-lang/wasp/pull/2707/commits/cc4de838c4fb96aecbd05edd815e20ef68619fd1)

File: wasp-app-runner/src/smtp.ts [hunk 2]
```diff
@@ -0,0 +1,24 @@
+import { exitCode } from "process";
+import { spawnWithLog } from "./process.js";
+import { log } from "./logging.js";
+
+export async function startLocalSmtpServer(): Promise<void> {
```
infomiho, line 5: Do we want to have this in the `wasp-app-runner`? 

Alternatively, we could run the Mailcrab SMTP server just in the Wasp CI. 

Having it in the `wasp-app-runner` gives a unified way to test all the example apps that require email verification.

waspc/examples/todoApp/headless-tests/tests/helpers.ts
FranjoMindek: I've meant in this way, which I see you've added. My bad if I had bad wording.
```ts
  if (process.env.HEADLESS_TEST_MODE === 'dev') {
    // In dev mode, the emails are verified automatically
    return
  }
  ```
  But the SMTP change is nice.

File: wasp-app-runner/src/build/index.ts [hunk 3]
```diff
@@ -29,6 +30,8 @@ export async function startAppInBuildMode({
     pathToApp,
   });
 
+  await startLocalSmtpServer();
```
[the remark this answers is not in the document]
FranjoMindek: I don't really see a reason to do it in `dev` mode.
In a way this further simulates `build` environment, which we wanted to do. So I think it fits `build` only.

wasp-app-runner/src/smtp.ts
FranjoMindek: I'd rather it not depends on an external service.
I think Filip can give more input here when he reviews it.
For me this is okay for testing purposes.

FranjoMindek (approved):
I have nothing of value to add anymore

wasp-app-runner/src/smtp.ts
sodic: I guess if we're starting the database, we can start this too.

A good rule of thumb is: if the dev mode can simulate the feature (e.g., database with `sqlite` or `wasp start db`, email verification with CLI output), then so should the build mode.

wasp-app-runner/src/smtp.ts
```diff
@@ -0,0 +1,24 @@
...
+      "--rm",
+      "-p",
+      "1080:1080",
+      "-p",
+      "1025:1025",
+      "marlonb/mailcrab:latest",
+    ],
+  }).then(({ exitCode }) => {
+    if (exitCode !== 0) {
+      log("smtp-server", "error", `SMTP server exited with code ${exitCode}`);
```
sodic, line 20: This is a nice example of localized logs and their prefixes (I talked about that in the other PR).

Usually, I'd recommend all `process.exit` calls to be at the top level (i.e., bubble up the error and exit at the top). But I think we can let it slide in `wasp-app-runner`.

waspc/examples/todoApp/headless-tests/tests/auth-hooks.spec.ts
```diff
@@ -21,7 +22,7 @@ test.describe('auth hooks', () => {
     })
 
     await expect(page.locator('body')).toContainText(
-      'On Before Signup Hook disallows this email.',
+      'On Before Signup Hook disallows this email.'
```
sodic, line 25: Aw man, I liked the trailing comma rule :(

Why'd we lose it?

waspc/examples/todoApp/headless-tests/tests/helpers.ts
```diff
@@ -14,6 +14,59 @@ export async function performSignup(
...
+    * `wasp-app-runner` starts a Mailcrab server on port 1080 by default.
+*/
+export async function performEmailVerification(
+  page: Page,
+  sentToEmail: string
+) {
+  if (process.env.HEADLESS_TEST_MODE === 'dev') {
+    // In dev mode, the emails are verified automatically
+    return
+  }
```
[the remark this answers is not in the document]
sodic: Please add this comment to the code too

sodic (approved):
Awesome, looks good to me!

waspc/examples/todoApp/headless-tests/tests/auth-hooks.spec.ts
FranjoMindek: `all` vs `es5`, comment on universal styles PR if you want to influence this

sodic: Fine, I'll fight this battle again.

Issue #734, opened 2022-09-26
Rethinking Wasp code organization


Martinsos:
Related to https://github.com/wasp-lang/wasp/issues/710 .

So far we have made decisions in Wasp like:
1. Code generated by Wasp is put in .wasp dir.
2. Code generated by Wasp is imported in JS files as "@wasp/...".
3. Generated code is structured in a certain manner (queries are under queries/, actions under actions/, ...).
4. User code is organized under ext/ directory and then just copied into the generated project.
5. And many more decisions!

We are now trying to add IDE support for user code in ext/ (specifically for JS/TS code) and one big thing that makes this somewhat tricky to do is the way our code is organized -> generated project is organized differently than what we have in the source dir, specifically in ext/, so we have to "cheat" editor/IDE into thinking stuff in ext/ is part of the project that is not really there (because IDEs, or their plugins for JS/TS, expect some standard JS project setup, like node_modules in specific place, maybe package.json to be there, tsconfig.json, ... -> and we have none of those in source project in ext/ dir, we only have it once you go to generated project).

So the idea is -> what if we rethink stuff and reorganize Wasp in such way that we maybe avoid some of these issues? Maybe make source code more similar in setup to generated project? Or at least those parts that are needed for IDE support for JS/TS?

So let's rethink stuff @shayneczyzewski @sodic @matijaSos ! Let's let our minds wander and come up with crazy ideas. Doesn't matter if they are unbaked or a bit impossible, let's just try to come up with some new directions and see where they take us.
**I suggest taking an hour or two specifically for this** and just trying to come up with smth.

While thinking about ideas, here are some requirements to keep in mind:
1. In the future we will want to have Wasp modules -> standalone, reusable packages of .wasp files + other JS/html/... files. Likely we will also want to mix .wasp files with other files, so they can be next to each other.
2. We might want to support some other languages next to JS/TS, for example Python, or Go:
  2.1. As languages used on backend, instead of JS/TS backend or next to JS/TS backend.
  2.2. As languages to write jobs in.
3. We want to support deploying our web apps as different architecture and on different providers.
4. We want to support multiple IDEs (even though we are now focusing on vscode).
5. Support stuff like prettier, linter, ... .
6. We want to allow organizing project by feature.

Some rough ideas that were mentioned so far:
1. What if we generated projects directly in the source project, next to files in ext/. Maybe organized as neatly as possible, so they are not too much in the way, but still they would be there.
[...]

PR #2720, opened 2025-05-05
Update link

cprecioso:
Update docs link as per https://github.com/wasp-lang/wasp/issues/734#issuecomment-2848569710


sodic (approved):
Thanks!

PR #2727, opened 2025-05-06
Extract session ID from Authorization header

infomiho:
@sodic pointed out that it's better to have one source of truth for the used session ID. He commented on the #2570 PR after the merge so this is follow-up PR to refactor the implementation. 

The PR is just **one file** change, the rest is e2e tests.


waspc/data/Generator/templates/sdk/wasp/api/index.ts
```diff
@@ -127,3 +123,12 @@ class WaspHttpError extends Error {
     this.data = data
   }
 }
+
+function getSessionIdFromAuthorizationHeader(header: string | undefined): string | undefined {
```
sodic, line 127: One request from me: can we use `null` instead of `undefined`? It feels more explicit (I believe most treat `undefined` as "not explicitly initialized", and `null` "explicitly initialized to an empty value").

Seeing `undefined` always feels weird to me.

This is a pet peeve of mine so I'm of course approving regardless.

sodic (approved):
Noice!

waspc/data/Generator/templates/sdk/wasp/api/index.ts
sodic: Oh, and one more thing, this functions should only accept `header: string`. It shouldn't deal with the header being undefined, that's the caller's job.

infomiho: The previous implementation looked like this
```ts
config._waspSessionId = sessionId
```
which meant we can compare the result of `getSessionId` with the `_waspSessionId` directly. The `getSessionId` returns `undefined` if session ID is not defined. 

That's the reason why `getSessionIdFromAuthorizationHeader` also returns `string | undefined`.  This way, we can compare `undefined` and `undefined` directly vs. doing extra work to compare `undefined` and `null`.

We could change the `getSessionId` to return `null` when we don't have a session. But it'll be a breaking change for our users if they use `getSession` since it's part of the public API. 

What do you say? Do we keep the `string | undefined`?

sodic: > We could change the getSessionId to return null when we don't have a session. 

I vote yes even if it is a breaking change.

But from what I see, `getSessionId` is part of our visible API, but not our official public API:
- It's marked as such in the [table](https://docs.google.com/spreadsheets/d/1QW-_16KRGTOaKXx9NYUtjk6m2TQ0nUMOA74hBthTH3g/edit?pli=1&gid=2143046114#gid=2143046114).
- The comment above it says it's private.
- Our docs don't mention it.

So break away :)

infomiho: Did it 👍  I'll merge when the CI passes

PR #2736, opened 2025-05-12
Check formatting in CI

cprecioso:
1. Adds a **root** [`.editorconfig`](https://github.com/wasp-lang/wasp/blob/cprecioso/2583-universal-code-style-formatting-and-linting-rules/.editorconfig) to the repo
	- Just some basic options (will be taken into account for prettier)
1. Adds a **root** [`.prettierrc`](https://github.com/wasp-lang/wasp/blob/cprecioso/2583-universal-code-style-formatting-and-linting-rules/.prettierrc) and [`.prettierignore`](https://github.com/wasp-lang/wasp/blob/cprecioso/2583-universal-code-style-formatting-and-linting-rules/.prettierignore) formatting to the repo
	- Default options, plus the editorconfig, plus:
	  ```jsonc
	  {
		// From the assorted configs already in our repo
	  	"semi": false,
	  	"singleQuote": true,
	  	"trailingComma": "es5",
	  	"jsxSingleQuote": true,
		// Specifically requested by @martinsos
	  	"experimentalOperatorPosition": "start"
	  }
	  ```
	- Plugins:
		- `prettier-plugin-organize-imports`: to deterministically order imports
		  - This is the same action that you can run in VS Code ("Organize imports")
		  - It still respects clusters of imports broken up by double newlines
		- `prettier-plugin-pkg`: deterministically orders package.json
		- `prettier-plugin-sh`: adds support for shell scripts
		- `prettier-plugin-tailwindcss`: formats tailwind class strings
	- Ignored:
		- Templates: prettier gets confused with the mustaches
		- Markdown: prettier is only compatible with MDXv1 for now, not v3
1. Removes all the other prettier configurations throughout the repo, we don't need them anymore
1. Adds a CI step that runs ormolu and prettier and fails if any file is not correctly formatted
1. Formatted everything prettier complained about (ormolu was already passing) (`npx prettier --write .`)

> [!TIP]
> Remove the last "Format" commit from the review pane so that you can easily review the changes. The last commit is just `npx prettier --write .`.


examples/thoughts/src/client/LoginPage.jsx
```diff
@@ -1,17 +1,16 @@
-import { LoginForm } from "wasp/client/auth";
-import React from 'react'
```
cprecioso, line 2: The `React` import is only necessary for old versions of the JSX transform. We use Vite so it should not be necessary, that's why prettier is automatically removing it.

examples/thoughts/src/client/MainPage.jsx
```diff
@@ -105,55 +105,64 @@ const Thought = () => {
   }
 
   return (
-    <div className="thought">
+    <div className='thought'>
```
cprecioso, line 108: I don't especially like the single quotes in JSX but it was already set in some config so I kept it

FranjoMindek: +1 to change this to double quotes

FranjoMindek:
@sodic wanted to fight `trailingComma` `es5` vs `all` setting, but he is away currently.

.github/actions/setup-haskell/action.yml
cprecioso, line 1: I extracted the common steps of setting up Haskell + the Cabal cache to a common action with the intention of using it later for `waspc-ci.yaml` as well

sodic:
> @sodic wanted to fight trailingComma es5 vs all setting, but he is away currently.

The mods are asleep, merge `trailingComma: false` :)

Yeah, I like trailing commas. I'm guessing we all know the arguments for them so I won't go into details. I won't die defending them if everybody is against it.

Does someone not like them? Why not, I'm curious.

cprecioso:
I'm a +1 for trailing commas everywhere

examples/todo-typescript/src/MainPage.tsx
```diff
@@ -1,29 +1,29 @@
...
-import React, { FormEventHandler, FormEvent } from "react";
-import waspLogo from "./waspLogo.png";
+  updateTask,
+  useQuery,
+} from 'wasp/client/operations'
+import { type Task } from 'wasp/entities'
+import waspLogo from './waspLogo.png'
 
-import "./Main.css";
+import './Main.css'
```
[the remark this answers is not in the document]
cprecioso: Yep, I've disabled the formatting in the templates

examples/thoughts/src/client/MainPage.jsx
infomiho: +1 to keep double quotes pls

.github/actions/setup-haskell/action.yml
infomiho: Great stuff

.github/workflows/check-formatting.yaml
```diff
@@ -0,0 +1,48 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  prettier:
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          git config --global core.autocrlf input
+          git config --global core.eol lf
```
infomiho, line 19: Should team members do something like this as well?

.prettierignore
```diff
@@ -0,0 +1,10 @@
+.wasp/
+dist/
+dist-newstyle/
+waspc/data/packages/
```
infomiho, line 4: Why not enforce the same style in the `packages`?

infomiho (commented):
Left some questions. The rules look good to me, but I'd like to keep double quote for JSX, I'm more used to it and I think it's more common?

One thing to add, we have all the actions team members might need to do in the `run` script: https://github.com/wasp-lang/wasp/blob/main/waspc/run#L114-L117 Let's add the prettier commands there.

.prettierignore
cprecioso: `waspc/packages` is the source for the packages
`waspc/data/packages` is only where they are being copied to

.github/workflows/check-formatting.yaml
cprecioso: Mmmm yes, if they were on Windows, it's kind of the first step to setup Git there. This is just copied from `wasp-ci.yaml` for good measure, but I'm not sure how needed it is.

examples/thoughts/src/client/MainPage.jsx
Martinsos: +1 for double quotes also.

Martinsos:
Quick resume of my opinions:
- I love having semicolons at the end, had my no-semicolon phase for years but am now back in semicolon camp. That said, if majority doesn't want semicolons, I am ok with that.
- I don't care about trailing colons.
- I prefer double quotes over single quotes, but it seems others also do so great.
- experimentalOperatorPosition -> I don't have strong opinion, but I do like it being in front a bit better due to diffing and reading. But I don't care that much.

@cprecioso I am curious about the experience for us, since I am not that knowledgeable about these configs -> how much setup is needed on our side? I am guessing not much, config is shared, and correct version of prettier will be installed per each project because it is specific in package.json, right? There is no need to use globally installer version of prettier I hope?
What about ormolu, I think for it it is a bit trickier how to ensure we all have the same version -> have you been playing with that? If not, we can also look together into that.

Martinsos:
> @Martinsos
> 
> > There is no need to use globally installer version of prettier I hope?
> 
> Nope, `npm install` on the local project is enough.
> 
> > how much setup is needed on our side?
> 
> Just `npm install` and `npm run format`. (I'l look into getting it into the `./run` script) Best experience IMO is by configuring format on save in your editor ([VSCode](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode), [emacs](https://github.com/prettier/prettier-emacs)). The editor should pick up the installed prettier version.
> 
> > What about ormolu, I think for it it is a bit trickier how to ensure we all have the same version
> 
> I didn't change anything in ormolu's setup. AFAICT we pin its version through `dev-tool.project`, and that's what `./run ormolu:format` uses.

Ok sounds great! One thing we don't do for `ormolu` right now is set up our editors to use the locally installad version, which is installed with `./run ormolu:format`, we all mostly instead use globally installed `ormolu`. But that is more on each of us individually to set up I guess, I think @FranjoMindek already does it correctly. The main reason why we didn't do it is because the way it is installed locally is a bit unortodox, it is not as smooth as with installing prettier locally, but it does work. Ok, I will leave this comment here for posterity, although I said everything really.

FranjoMindek:
My preferences:

trailing commas always
double quotes on jsx
semicolons
100 width (but 80 is okay)
2 spaces tab width

.github/workflows/check-formatting.yaml
cprecioso: Removed it as this only runs on ubuntu

infomiho:
My approve still stands 👍

.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
```
infomiho, line 1: Do we maybe want to list out all the defaults we expect to use, what if the Prettier default changes in the meantime. If we can be explicit about our choices we made, why not make it clear.

cprecioso: Hmmm, I've not seen that in a lot of places I think. I think it's not really needed. 

We will have opportunities to change this if we ever need to, since the prettier dependencies are pinned (no `^version`) [as they recommend](https://prettier.io/docs/install#summary:~:text=Install%20an%20exact%20version%20of%20Prettier%20locally%20in%20your%20project):
https://github.com/wasp-lang/wasp/blob/3fd700d56f86a383154ecd3e6aced2e1a5f3323f/package.json#L8-L11

And any changes they make will affect us anyway, since they have explicitly communicated that they are [against options](https://prettier.io/docs/option-philosophy).

infomiho: I can get behind their philosophy 👍 I guess our options are "whatever Prettier does".

cprecioso: Ahh I think I misunderstood. I agree with you that it is a good idea to add the explicit choices. Let me cook.

cprecioso: Done, will merge

PR #2750, opened 2025-05-13
Quote all titles

cprecioso:
In code blocks, Docusaurus needs the title of the file to be quoted, otherwise it doesn't show up.

I did a find-and-replace.

~~~diff
-```wasp title=main.wasp
+```wasp title="main.wasp"
~~~


infomiho (approved):
Simple enough

Issue #2020, opened 2024-05-08
Revisit types exported from the SDK


sodic:
Item 47 of _Effective Typescript_ says "Export all types that appear in public APIs. The reasoning is solid and goes something like this:
> Users will eventually want to reference a type your library uses (e.g., it's part of a function's type signature) only to find out it isn't exported. Since TypeScript has many powerful utility types, they'll find a way to extract and use it anyway (e.g., with `ReturnType`, `Parameters`, `typeof`, etc.). 
>
> If a type appears in a public function declaration, it's effectively exported. Therefore, you should do our users a favor and just export them explicitly.

Our SDK doesn't currently do this, so let's make it happen :)

PR #2751, opened 2025-05-14
Fix `FullConfig` propagation in `wasp-config` and other small fixes

FranjoMindek:
Fix `FullConfig` to propagate to it's children.
Better configs jsdocs.

Since we fixed the type, remove optional access where unnecessary.

Remove export of `main` in `run.ts`.

Use local vars instead of `Fixtures` where applicable.

Uses new `Branded` type for branded types.

Shortens default values on optional fields.

### DIFF
--- a/waspc/packages/wasp-config/__tests__/testFixtures.ts
+++ b/waspc/packages/wasp-config/__tests__/testFixtures.ts
@@ -7,15 +7,6 @@ import * as AppSpec from '../src/appSpec.js'
 import * as UserApi from '../src/userApi.js'
 
 /**
- * This type removes all properties from T that are optional.
- *
- * The empty object type - {} - doesn't behave how you expect in TypeScript.
- * It represents any value except null and undefined.
- * To represent empty objects, we use `Record<string, never>` instead.
- * @see https://www.totaltypescript.com/the-empty-object-type-in-typescript
- */
-
-/**
  * Creates a type containing only the required properties from T.
  *
  * This utility:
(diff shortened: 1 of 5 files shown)


waspc/packages/wasp-config/__tests__/testFixtures.ts
```diff
@@ -23,10 +26,21 @@ type MinimalConfig<T> = keyof T extends never
...
+ * Creates a type with all properties and nested properties required.
+ *
+ * This utility:
+ * - Makes all properties required recursively (removes optional flags)
+ * - Stops from unwrapping branded types fully
+ *
+ * @template T - The type to make fully required
  */
-type FullConfig<T> = Required<T>
+type FullConfig<T> = T extends { _brand: string }
```
infomiho, line 39: What if we created a `Branded` type which we can check against here? Something like: https://github.com/wasp-lang/wasp/blob/main/wasp-app-runner/src/types.ts#L1

infomiho (commented):
Looking good, but had one improvement in mind

waspc/packages/wasp-config/__tests__/testFixtures.ts
FranjoMindek: Wow that looks good.
Will be easier to track branded logic with this for sure.

waspc/packages/wasp-config/__tests__/testFixtures.ts
```diff
@@ -363,7 +379,7 @@ export function getRoute(routeType: PageType): NamedConfig<UserApi.RouteConfig>
       name: 'MinimalRoute',
       config: {
         path: '/foo/bar',
-        to: getPage(routeType).name as string & { _brand: 'Page' },
+        to: getPage(routeType).name as Branded<string, 'PageName'>,
```
infomiho, line 382: Is there a reason why aren't you using the branded type defined here: https://github.com/wasp-lang/wasp/pull/2751/files#diff-0c71293e1feeb02208bdb6deaae56349707ef8bef40a83eb75072cbd76b25136R267

infomiho (approved):
Approving since there is only one small suggestion (which might not be applicable)

waspc/packages/wasp-config/__tests__/testFixtures.ts
FranjoMindek: I'm still undecided here:

I can export `PageName` and them use it in fixtures, but then `PageName` is exported and anyone can use the type. This way, only way to get it is through exported functions.

FranjoMindek:
I will fix it :)

waspc/packages/wasp-config/__tests__/testFixtures.ts
sodic: _Effective TypeScript_, 2nd edition, item 67 should clear this up :)

If you're in a rush: https://github.com/wasp-lang/wasp/issues/2020

Issue #2718, opened 2025-05-03
Reduce the size of Wasp binaries


sodic:
Wasp `.tar.gz` files are much larger than we intended them to be:
- Intended size: ~15 MB compressed, 92 MB uncompressed
- Actual size we've been shipping: ~110 Mb compressed, ~500 MB uncompressed

The size difference is thanks to our build accidentally including several `node_modules` directories, specifically:
- `data/packages/studio/client/node_modules` - Since [v0.11.5-studio1](https://github.com/wasp-lang/wasp/releases/tag/v0.11.5-studio-1). 
This folder got in because the script `install_packages_to_data_dir` only deletes the top-level `node_modules` folder, and the `studio` package also has `node_modules` in the `client` subdir.

- `data/packages/prisma/node_modules` - Since [v0.15.0](https://github.com/wasp-lang/wasp/releases/tag/v0.15.0).
This folder got in because:
  1. Our CI runs e2e tests before packaging the binary.
  2. The generator runs `prismaFormat` which uses  `prisma` package.
  3. Using the `prisma` package installs its packages and creates the `node_modules` folder.
  4. The CI packages the created `node_modules` folder into the binary.

cprecioso:
Another possibility:

Our **current** approach now is basically:

1. Build:
	1. Install dependencies and build each package
	2. Remove the `node_modules` folder
	3. Copy to `data/packages`
2. We ship the full package folders in the `data`
3. Runtime:
	1. Install dependencies for the package (inside the data dir)
	2. Run the script

---

Maybe we can use `npm pack` instead of copying the folders around. [`npm pack`](https://docs.npmjs.com/cli/v8/commands/npm-pack) is the npm command that produces the tarballs that later get updated to the registry. It doesn't include the `node_modules`, and in fact only includes the files declared in [`package.json#files`](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#files). We could probably only have our `dist` folders there.

Later, `npm install` can just be pointed to the tarball, and will install it as if it was any other package from the registry. This is not an esoteric use-case, it is [documented](https://docs.npmjs.com/cli/v8/commands/npm-install#:~:text=a%20gzipped%20tarball%20containing) and supported.

1. Build:
	1. Install dependencies and build each package
	2. **`npm pack -o wasp-deploy.tgz`** (e.g. for the deploy package)
	3. Copy **the tarball** to `data/packages`
2. We ship the the tarballs in the `data`
3. Runtime:
[...]

PR #2759, opened 2025-05-15
Draft: Use Hpack

cprecioso:
Related to #2718

HPack has a saner glob syntax than regular Cabal. Makes it simpler in general, and also better able to remove `node_modules` from the data folder.


Martinsos:
@cprecioso interesting! I always liked package.yaml from Stack more than the .cabal format, due to more stuff being done automatically, but I remember last time I was considering this there was a reason why I decided it doesn't make sense to go for it. I can't remember what it is at the moment -> maybe that changed, whatever it was, or maybe I overestimated some complexity that is not there. Or maybe the reason is still there. I will take a look and give this a thought for sure (when I catch some time).

Issue #2737, opened 2025-05-12
Rename `wasp-config` `userSpec` to `tsAppSpec` for clarity and consistency


FranjoMindek:
We've had an discussion around the current naming of `userSpec`, https://github.com/wasp-lang/wasp/pull/2690#discussion_r2073661240. The end conses was that the name is misleading and potentially confusing. 

### Problems with `userSpec`:

* The word “user” is ambiguous in the context of Wasp. In most places, “user” refers to the **auth user** or the **app user**, not the **Wasp developer**.
* There’s nothing else in the codebase using “user” to refer to Wasp developers, so this usage is inconsistent.
* The name doesn’t clearly convey the role of the structure — it’s not intuitive that this is a TypeScript-friendly representation of `AppSpec`.

### Agreed direction:

* Rename `userSpec` → `tsAppSpec` (short for **TypeScript App Spec**), as that is what it effectively is: a representation of `AppSpec` tailored for usage in TypeScript.
* Keep `appSpec` as-is. It is the canonical form and should not be renamed to something like `hsAppSpec`, since its connection to Haskell is an implementation detail.

sodic:
> It might be worthwhile to look into renaming the UserApi.App

The only thing we should ask, without considering anything else, is _**What do we want the user to see?**_

After we decide on that, we should then adapt our internal code to work with it. Our code should pay the price for good DX.

It's generally never desirable do design the user's to make it easier on ourselves. We sometimes make exceptions when the approaches greatly differ in complexity but that's not the case here.
Btw the price for good DX in this case is pretty cheap, we can easily reexport the same class under a different name.

### What I want the user to see

I prefer `new App`.

FranjoMindek:
Should be user DX first, I agree.
I know that @Martinsos got really confused around having `AppSpec.Api`, `UserApi.Api` and `UserApi.ApiConfig`, hence we brought it up.

If our DX is suffering here we can maybe try different module system:
1) `appSpec.ts` -> `AppSpec.Api`
2) `userApi.ts` -> `UserApi.Api`
3) `tsAppSpec.ts` -> `TSAppSpec.ApiConfig`

In other words, extract ts app spec stuff out of the `userApi.ts` into its own module.
Let `userApi.ts` handle user facing exports only.

sodic:
[...]

PR #2762, opened 2025-05-16
`wasp-config` userApi renaming

FranjoMindek:
Fixes: https://github.com/wasp-lang/wasp/issues/2737

First draft for renaming.
`userApi.ts` split into `publicApi/App.ts` and `publicApi/tsAppSpec.ts`
Anything named `...userSpec...` renamed to `...tsAppSpec...`.
Anything named `...userApp...` renamed to `...App...` (this referred to `publicApi.App.ts`)

Might be controversial:
added config suffix to `testFixtures.ts` config getters (so we have `getAppConfig` and `createApp`)


waspc/packages/wasp-config/__tests__/mapUserSpecToAppSpecDecls.integration.test.ts
Martinsos, line 1: Name of the file needs updating. Same goes for the other file.

Martinsos (commented):
I will let @sodic take a proper look, but I like the change in the naming, "user" is gone and names make sense! Nice stuff.

waspc/packages/wasp-config/__tests__/mapUserSpecToAppSpecDecls.integration.test.ts
FranjoMindek: Ah forgot those on test files, good catch.

FranjoMindek:
I recommend opening https://github.com/wasp-lang/wasp/pull/2762/files/0a624611038d5f42df918c3839295bd2780635d7 which shows state before merging Carlos's formatting changes.

@sodic

waspc/packages/wasp-config/src/publicApi/App.ts
```diff
@@ -0,0 +1,111 @@
...
+  EmailSenderConfig,
+  JobConfig,
+  PageConfig,
+  PageName,
+  QueryConfig,
+  RouteConfig,
+  ServerConfig,
+  TsAppSpec,
+  WebsocketConfig,
+} from './tsAppSpec.js'
```
sodic: You can use a namespace import here if you want.

waspc/packages/wasp-config/src/publicApi/App.ts
```diff
@@ -0,0 +1,111 @@
...
+
+export class App {
+  #tsAppSpec: TsAppSpec;
+
+  // NOTE: Using a non-public symbol gives us a package-private property.
+  // It's not that important to hide it from the users, but we still don't want
+  // user's IDE to suggest it during autocompletion.
+  [GET_TS_APP_SPEC]() {
```
sodic: Let's annotate the return type of this method - since it's "public."

waspc/packages/wasp-config/src/index.ts
```diff
@@ -1 +1,2 @@
-export * from './userApi.js'
+export * from './publicApi/App.js'
+export * from './publicApi/tsAppSpec.js'
```
sodic: We're exporting `TsAppSpec` which the user shouldn't see.

Also, the `TsAppSpec` object is a part of the `App` class's logic. Seems like it should be in in the other file. I don't what to do with the naming in that case though.

waspc/packages/wasp-config/__tests__/testFixtures.ts
```diff
@@ -67,53 +68,53 @@ type PageType = (typeof PAGE_TYPES)[number]
...
+  appConfigName: string
+  app: App
 } {
   if (scope === 'minimal') {
-    const { name: appName, config: appConfig } = getApp(scope)
-    const userApp = new UserApi.App(appName, appConfig)
-    return { appName, userApp }
+    const { name: appName, config: appConfig } = getAppConfig(scope)
+    const userApp = new App(appName, appConfig)
+    return { appConfigName: appName, app: userApp }
```
sodic: We can call it `app` right away, no need for `userApp` any more.

Same thing applies to other places.

waspc/packages/wasp-config/__tests__/testFixtures.ts
```diff
@@ -529,44 +538,48 @@ export function getSchedule(scope: ConfigType): Config<UserApi.ScheduleConfig> {
...
       pgBoss: { jobOptions: { attempts: 3 } },
     },
-  } satisfies FullConfig<UserApi.ScheduleConfig>
+  } satisfies FullConfig<TsAppSpec.ScheduleConfig>
 }
 
-export function getPerform(scope: 'minimal'): MinimalConfig<UserApi.Perform>
-export function getPerform(scope: 'full'): FullConfig<UserApi.Perform>
+export function getPerform(scope: 'minimal'): MinimalConfig<TsAppSpec.Perform>
+export function getPerform(scope: 'full'): FullConfig<TsAppSpec.Perform>
```
sodic: Should this one be called `getPerformConfig` and the type `Perform`?

We possibly forgot other names too.

waspc/packages/wasp-config/__tests__/testFixtures.ts
```diff
@@ -575,25 +588,25 @@ export function getApiNamespace(scope: ConfigType): NamedConfig<UserApi.ApiNames
       middlewareConfigFn: getExtImport(scope, 'named'),
       path: '/foo',
     },
-  } satisfies FullNamedConfig<UserApi.ApiNamespaceConfig>
+  } satisfies FullNamedConfig<TsAppSpec.ApiNamespaceConfig>
 }
 
-export function getApis(): NamedConfig<UserApi.ApiConfig>[] {
+export function getApiConfigs(): NamedConfig<TsAppSpec.ApiConfig>[] {
   // @ts-expect-error we don't need to know the specifc function overload in this case
```
sodic: Old code, but prefer `as any` to `@ts-expect-error`. `@ts-expect-error` doesn't tell what you expect and applies to the whole line. The type assertion applies just to the one variable that is not of the correct type.

Example where this matters...

This doesn't catch the typo:
```ts
// @ts-expect-error we don't need to know the specifc function overload in this case
return CONFIG_TYPESSSSSS.map((scope) => getApiConfig(scope))
```

This does:
```ts
return CONFIG_TYPESSSSSS.map((scope) => getApiConfig(scope as any))
```

In this case though, you don't need any of those. Just add another type signature to the overloaded function (it's the correct thing to do anyway):
```ts
export function getApiNamespaceConfig(scope: "full" | "minimal"): NamedConfig<TsAppSpec.ApiNamespaceConfig>;
```
The same comment applies to other places.

waspc/packages/wasp-config/__tests__/testFixtures.ts
```diff
@@ -575,25 +588,25 @@ export function getApiNamespace(scope: ConfigType): NamedConfig<UserApi.ApiNames
...
+export function getApiConfig(scope: 'full'): FullNamedConfig<TsAppSpec.ApiConfig>
+export function getApiConfig(scope: ConfigType): NamedConfig<TsAppSpec.ApiConfig> {
   if (scope === 'minimal') {
     return {
       name: 'MinimalApi',
       config: {
         fn: getExtImport(scope, 'named'),
         httpRoute: getHttpRoute(scope),
       },
-    } satisfies MinimalNamedConfig<UserApi.ApiConfig>
```
sodic: EDIT: I realized it doesn't work for enforcing minimal configs.

~~Also old code, but the `satisfies` aren't necessary in most of the file. Excess property checking takes care of it (unless I'm missing something with the types):~~

image

### If else
Also, when writing short `if-else` function bodies, it's best to go the [structured programming](https://en.wikipedia.org/wiki/Structured_programming) route and be explicit:
```ts
if (something) {
  return ...
} else {
  return ...
} 
```
Or you could even use a ternary.

`if` without the `else` is best used only when eliminating preconditions in early returns.

waspc/packages/wasp-config/src/appAnalyzer.ts
```diff
@@ -1,29 +1,29 @@
...
+import { mapTsAppSpecToAppSpecDecls } from './mapTsAppSpecToAppSpecDecls.js'
+import { App } from './publicApi/App.js'
 
-export async function analyzeUserApp(
+export async function analyzeApp(
   waspTsSpecPath: string,
   entityNames: string[]
 ): Promise<Result<AppSpec.Decl[], string>> {
-  const userAppResult = await getUserApp(waspTsSpecPath)
+  const userAppResult = await getApp(waspTsSpecPath)
```
sodic: `userAppResult` leftover here :)

sodic (commented):
Left some comments.

The new code is mostly ready, but we could improve the old stuff if you're up for it :)

waspc/packages/wasp-config/__tests__/testFixtures.ts
FranjoMindek: `Perform` is more like a subset `Type` of some other config.
I refactored it to follow TS app spec names. e.g. if TS app spec call it a config, the getter will have config in name.

We have:
```ts
export type Perform = {
  fn: ExtImport;
  executorOptions?: ExecutorOptions;
};
```

waspc/packages/wasp-config/__tests__/testFixtures.ts
FranjoMindek: It looks kinda funny, but it's the right way. Thanks.

Didn't know that typescript couldn't deduce that type without explicit mention. 
More correctly, I thought that the function implementation type also counted as a type option.

I was weirded out when I got this error, but I didn't pursue enough.

waspc/packages/wasp-config/src/index.ts
FranjoMindek: I remember that we talked about this already.

Somebody said something along the lines "that types are also part of the public API because users use them through the `App` API" and that it would "allow people to do abstractions like in our TS app spec example". I think it was you but it might've been @Martinsos .

I think I would still export this.

FranjoMindek: Moving `TsAppSpec` is related to https://github.com/wasp-lang/wasp/pull/2762#discussion_r2131886907

waspc/packages/wasp-config/src/publicApi/App.ts
FranjoMindek: We could in sense to:

1) Clean up the imports
2) Have type import consistency (`import * as TsAppSpec from...`)

Only problem is name collision if we also move `TsAppSpec` type from `tsAppSpec.ts` to `App.ts`.

We probably want to rename `TsAppSpec` type.
~~I've picked `AppData` for now just to avoid collision, but we would need to brainstorm this.~~
EDIT: reverted it back to `TsAppSpec` and moved back to `tsAppSpec.ts`. Kinda funny we have `TsAppSpec.TsAppSpec` but I will keep it as is till we figure it out.

Note: I put 🚀 since this is more connected to comment below.

waspc/packages/wasp-config/__tests__/testFixtures.ts
```diff
@@ -71,731 +72,847 @@ type PageType = (typeof PAGE_TYPES)[number];
...
+      methods: getAuthMethods(scope),
+      externalAuthEntity: getEntity("social-user"),
+      onAuthSucceededRedirectTo: "/profile",
+      onBeforeSignup: getExtImport(scope, "named"),
+      onAfterSignup: getExtImport(scope, "named"),
+      onAfterEmailVerified: getExtImport(scope, "named"),
+      onBeforeOAuthRedirect: getExtImport(scope, "named"),
+      onBeforeLogin: getExtImport(scope, "named"),
+      onAfterLogin: getExtImport(scope, "named"),
+    } satisfies FullConfig<TsAppSpec.AuthConfig>;
```
sodic: We talked about `FullConfig` and `MinimalConfig` in person. 

Conclusion:
- When writing complex types, we should add "test cases" in their JSDoc, just a couple of examples illustrating how they should behave. This helps someone else improve the types and spot bugs.
- `MinimalConfig` doesn't isn't recursive and is a little more complicated than it needs to be.

## Improving `MinimalConfig`
Based on my understanding of `MinimalConfig`, here's how I'd write it:
```ts
/** Document this with examples */
type MinimalConfig<Obj extends object> = keyof Obj extends never
  ? { [key: string]: never }
  : GetRequiredFieldsOf<Obj>

/** Document this with examples */
type GetRequiredFieldsOf<Obj> = {
  [K in keyof Obj as Obj[K] extends Required<Obj>[K] ? K : never]: GetRequiredFieldsOf<Obj[K]>
};

type C = MinimalConfig<{ a: number, b: { c?: number, e: number }, d?: number}>
const test: C = { a: 12, b: { c: 12, e: 14}, d: 15 }
```
[Playground link](https://www.typescriptlang.org/play/?#code/PQKhAIBEHsGMFcC2BTAdgF3OgFgSwM7gDuuO4yAHgIaIAOANsoSMALABQ6AnrcuALK5UuRFXoBhaKgBmuAOYAeAPIAjAFbkK6NABNC0dcljoAfOAC84ANbIu0aeFUbK21HvCpkAN2QAnDuDgAPzgAN7gANo2XABc4PjovkJyALpxnj6+4AC+AeBxAOLI6ABKyACO8Li+yDoAYrjI9HpK0srqJhwcoBAwCCgYWHiEJGSUNAxM4Cwc3LzgRaUVVTX1jc34re1qZpaheREA0uBC1rb2jurgVIRORymaru5lldW12yb3weDH6d5+aQWxReK1qDSaLTad0OKU67GyAG4upweHxxBYBEIRGJJDJ5ApwlR0kgVH4ADTgFRxcKwILExCk3wU5D0xk5Ck6OkeEl+bJw2BSBJYJjoOLovbXOIARgATBSqWFwLBpXLyNKACzZDnSgCsOSAA).

## Improving `FullConfig`
Let's start by adding some examples. I'm a little fuzzy on what "unwrapping branded types" means. After I understand that, I might be able to suggest an improvement.

FranjoMindek: > unwrapping branded types

Because type branding is done through a object, it makes all of the `Branded` types be recognized as objects. If we let them be caught by `keyof` we introduce non-existing fields.

image


That's why we catch them early and return them as is.

FranjoMindek: Working on this type made me realize we have more issues with our current approach:
1) We don't handle array types properly 
  -> currently they get handled by `extends object` which iterates on array's functions
  
image

2) We must handle `T` as a `any` type rather than as  an `object` to make it available on possibly non-`object` elements. Currently we assume everything is an `object` because we `keyof` by default.

3) Branded stuff explained in the comment above

Below is an example in our TS app spec where we have an array, and we need to apply the Minimal/Full config on array items:
```ts
export type DbConfig = {
  seeds?: ExtImport[]; // we need Minimal/FullConfig<ExtImport>[]
  prismaSetupFn?: ExtImport;
};
```
With all that in mind I created the following type at the end:
```ts
// I've also changed generics to have more meaningful names.
type MinimalConfig<T> =
  T extends Branded<infer TType, infer TBrand>
    ? Branded<TType, TBrand>
    : T extends Array<infer TArrayItem>
      ? Array<MinimalConfig<TArrayItem>>
      : T extends object
        ? keyof T extends never
          ? EmptyObject
          : {
              [K in keyof T as T[K] extends Required<T>[K]
                ? K
                : never]: MinimalConfig<T[K]>;
            }
        : T;
```
Most of the logic happens on `object` since it's where optional types are possible. Before that it's just filtering non-`object` logic. 

We apply the same logic to `FullConfig`:
```ts
type FullConfig<T> =
  T extends Branded<infer TType, infer TBrand>
    ? Branded<TType, TBrand>
    : T extends Array<infer TArrayItem>
      ? Array<FullConfig<TArrayItem>>
      : T extends object
        ? {
            [K in keyof T]-?: FullConfig<T[K]>;
          }
        : T;
```

FranjoMindek: I think we need to go generic T approach instead of object to make this work?

```ts
type MinimalConfig<T> =
  T extends Branded<infer TType, infer TBrand>
    ? Branded<TType, TBrand>
    : T extends ReadonlyArray<infer TArrayItem>
      ? ReadonlyArray<MinimalConfig<TArrayItem>>
      : T extends object
        ? keyof T extends never
          ? EmptyObject
          : {
              [K in keyof T as T[K] extends Required<T>[K]
                ? K
                : never]: MinimalConfig<T[K]>;
            }
        : T;
```

waspc/packages/wasp-config/src/index.ts
Martinsos: Sounds like @sodic

waspc/packages/wasp-config/__tests__/testFixtures.ts
sodic: > More correctly, I thought that the function implementation type also counted as a type option.

Yes, it's a common point of confusion. It doesn't treat the implementation as a type signature on purpose because thre are cases where you wouldn't that (e.g., when there are more arguments and the entire union doesn't make semantic sense).

waspc/packages/wasp-config/src/index.ts
sodic: It was me, yes, but that's not the case here... Or at least I don't see it :)

> Somebody said something along the lines "that types are also part of the public API because users use them through the App API"

How could the users see get `TsAppSpec` type through the `App` class's logic? From what I can tell, it's 100% internal. We even have that private symbol thing that ensures it stays that way.

FranjoMindek: I think it was mean in this way:

We expose TS App Spec config types in the public methods of the public `App` class:
```ts
  action(this: App, name: string, config: ActionConfig): void {
    this.#userSpec.actions.set(name, config);
  }
```
Therefore users are directly exposed to e.g. `ActionConfig`.

Why not allow users to create functions which return `ActionConfig`?

sodic: Oh, I think we have a misunderstanding.

I'm talking specifically about the type `TsAppSpec`, the one with the maps used internally in `App`. That one shouldn't be public.

All the other types should be public, of course.

FranjoMindek: Oh of course then.
That should be solved by moving it to `App.ts` where it will not be exported.

waspc/packages/wasp-config/__tests__/testFixtures.ts
sodic: Hm, not quite sure I got what you mean...

> Perform is more like a subset Type of some other config.

Because `ScheduleConfig` is sitting right next to `Perform`, and it does have a `Config` suffix. `EmailVerificationConfig` and `PasswordResetConfig` are even deeper.

> I refactored it to follow TS app spec names. e.g. if TS app spec call it a config, the getter will have config in name.

I understand, that's why I suggested to rename both the getter and the type :)

FranjoMindek: Aha renaming both of them could work. I get you.
I guess we didn't have config because you copied the App Spec initially. We can do that.

sodic: I think we should pick either:
1. All objects in `TsAppSpec` have the suffix `Config` (except some exceptions like `EmailFromField` or `ExtImport`)
2. Only the top objects (`Api` `Client`, `Action`...) have the suffix `Config`

I prefer the second option because it seems more straightforward.

For example, I'm a little confused why we choose to prefix some of the deeper objects with `Config` (e.g., `ScheduleConfig`) and not others (e.g., `CrudOperations). Which is especially bad because I named them :upside_down_face:

waspc/packages/wasp-config/__tests__/testFixtures.ts
```diff
@@ -71,731 +145,844 @@ type PageType = (typeof PAGE_TYPES)[number];
...
-    return { appName, userApp };
-  }
+  const { name: appName, config: appConfig } = getAppConfig(scope);
+  const app = new App(appName, appConfig);
 
-  const { name: appName, config: appConfig } = getApp(scope);
-  const userApp = new UserApi.App(appName, appConfig);
+  if (scope === "minimal") {
+    return { appConfigName: appName, app };
+  } else {
```
sodic: The early return was the perfect fit here. This is one of those cases where it's a better choice (which I mentioned [here](https://github.com/wasp-lang/wasp/pull/2762#discussion_r2132273043)).

If the difference is unclear, I can elaborate :)

waspc/packages/wasp-config/__tests__/testFixtures.ts
sodic, line 1: Since we went into refactoring... We should structure the file in a way that the public (exported) stuff is at the top (where possible), the private stuff should follow it. Remeber the newspaper article analogy from Clean Code: detalis should follow the important stuff.

waspc/packages/wasp-config/__tests__/testFixtures.ts
```diff
@@ -71,731 +145,844 @@ type PageType = (typeof PAGE_TYPES)[number];
...
-
-  return {
-    email: "test@domain.tld",
-    name: "ToDo App",
-  } satisfies FullConfig<UserApi.EmailFromField>;
 }
 
 export function getExtImport(
   scope: "minimal",
   type: AppSpec.ExtImportKind,
```
sodic: We can improve the naming here:
- The variable is only called `type` and doesn't tell us what it is on its own.
- The is using the term `kind`, not `type`, and the variable should follow that

waspc/packages/wasp-config/__tests__/testFixtures.ts
```diff
@@ -71,731 +145,844 @@ type PageType = (typeof PAGE_TYPES)[number];
...
+        import: "namedExport",
+      } satisfies FullConfig<TsAppSpec.ExtImport>;
     }
-
-    return {
-      from: "@src/external",
-      import: "namedExport",
-    } satisfies FullConfig<UserApi.ExtImport>;
+  } else {
+    throw new Error(`Unhandled scope or type: scope=${scope}, type=${type}`);
```
sodic: Do do we need this? Does TS enforce the correct type, or do the values come from runtime?

waspc/packages/wasp-config/__tests__/testFixtures.ts
```diff
@@ -71,731 +145,844 @@ type PageType = (typeof PAGE_TYPES)[number];
...
   scope: ConfigType,
   type: AppSpec.ExtImportKind,
-): Config<UserApi.ExtImport> {
+): Config<TsAppSpec.ExtImport>;
+export function getExtImport(
+  scope: ConfigType,
+  type: AppSpec.ExtImportKind,
+): Config<TsAppSpec.ExtImport> {
   if (type === "default") {
     if (scope === "minimal") {
```
sodic: There's some duplication in this function. Flipping the checks can help:
```ts
const importObject =
  type === "default"
    ? { importDefault: "defaultExport" }
    : { import: "namedExport" };

if (scope === "minimal") {
  return {
    from: "@src/external",
    ...importObject,
  } satisfies MinimalConfig<TsAppSpec.ExtImport>;
} else if (scope === "full") {
  return {
    from: "@src/external",
    ...importObject,
  } satisfies FullConfig<TsAppSpec.ExtImport>;
} else {
  ...
}

  ```
This will become more useful if we add more properties, and especially if we add more "union" properties. Imagine what happens if we add another thing with two options: 8 execution paths and each one with duplicated logic.

P.S. If TypeScript doesn't handle invalid scopes or types (not sure about this, left another comment), we can add an import type check instead of the ternary operator.

File: waspc/packages/wasp-config/__tests__/testFixtures.ts [hunk 14]
```diff
@@ -71,731 +145,844 @@ type PageType = (typeof PAGE_TYPES)[number];
...
-    method: "GET",
-    route: "/foo/bar",
-  } satisfies FullConfig<UserApi.HttpRoute>;
 }
 
-export function getJobs(): NamedConfig<UserApi.JobConfig>[] {
-  // @ts-expect-error we don't need to know the specifc function overload in this case
-  return CONFIG_TYPES.map((scope) => getJob(scope));
+export function getJobConfigs(): NamedConfig<TsAppSpec.JobConfig>[] {
+  return CONFIG_TYPES.map((scope) => getJobConfig(scope));
```
sodic: The arrow function is redundant, we can do `CONFIG_TYPES.map(getJobConfig)`. There's another function above with the same thing.

I'm guessing you knew this and choose to do an arrow for stylistic purposes. If so, that's fine :)

waspc/packages/wasp-config/__tests__/testFixtures.ts
```diff
@@ -71,731 +145,844 @@ type PageType = (typeof PAGE_TYPES)[number];
...
 export function getCrudOperationOptions(
   scope: ConfigType,
-): Config<UserApi.CrudOperationOptions> {
+): Config<TsAppSpec.CrudOperationOptions>;
+export function getCrudOperationOptions(
+  scope: ConfigType,
+): Config<TsAppSpec.CrudOperationOptions> {
   if (scope === "minimal") {
-    return {} satisfies MinimalConfig<UserApi.CrudOperationOptions>;
+    return {} satisfies MinimalConfig<TsAppSpec.CrudOperationOptions>;
```
sodic: Seems like this type is now broken:
```ts
// TS doesn't complain here:
return { test: 'a', setupFn: 'asdf'  } satisfies MinimalConfig<TsAppSpec.ServerConfig>;
```
And truly, if you create an alias and hover:

image

To make matters worse, this propagates from child to parent objects:

image

Since these config helper types have turned out to be pretty unwieldy, we should write tests for them and clearly documented how they're supposed to work.

waspc/packages/wasp-config/__tests__/testFixtures.ts
FranjoMindek: I'm also more in favor of the 2nd option.
I would go that way.

waspc/packages/wasp-config/__tests__/testFixtures.ts
```diff
@@ -71,731 +145,844 @@ type PageType = (typeof PAGE_TYPES)[number];
 const ENTITY_TYPES = ["task", "user", "social-user"] as const;
 type EntityType = (typeof ENTITY_TYPES)[number];
 
-export function createUserApp(scope: ConfigType): {
-  appName: string;
-  userApp: UserApi.App;
+export function createApp(scope: ConfigType): {
```
sodic, line 11: I can't reach the line I want, but...

I think we should comment why `PAGE_TYPES` includes `CONFIG_TYPES`. It's my first time seeing that code and I don't really understand that.

Again, this is related to code that's not even shown in the diff, so we can skip it if you think it's too much.

waspc/packages/wasp-config/__tests__/testFixtures.ts
```diff
@@ -1,48 +1,122 @@
...
- * - Makes all properties required recursively (removes optional flags)
- * - Stops from unwrapping branded types fully
+ * - Makes all properties required recursively
+ * - Stops from unwrapping branded types
  *
  * @template T - The type to make fully required
+ *
+ * @example
+ * ```ts
+ * // Given the following type:
```
sodic: The examples are great.

I mentioned in a different comment that I'm leaning towards putting them in real tests.

waspc/packages/wasp-config/__tests__/testFixtures.ts
```diff
@@ -1,48 +1,122 @@
...
-      ? { [K in keyof T]-?: FullConfig<T[K]> }
-      : T;
+  T extends Branded<infer TType, infer TBrand>
+    ? Branded<TType, TBrand>
+    : T extends Array<infer TArrayItem>
+      ? Array<FullConfig<TArrayItem>>
+      : T extends object
+        ? {
+            [K in keyof T]-?: FullConfig<T[K]>;
+          }
```
sodic: Maybe this would benefit from being its own named type.

Since,, we should apply all the rules we use to write functions to type operators (single responsibility, naming, etc.).

Then again, this avoids a mutually recursive type, which is also desirable. So perhaps we shold leave it like this.

Btw, what happens when `T` is a union type? I don't think this behaves as expected (haven't checked). Wasp currently does't have union field values, but if we rely on that assumption, we should constrain the type signature accordingly.

waspc/packages/wasp-config/__tests__/testFixtures.ts
```diff
@@ -1,48 +1,122 @@
...
-type MinimalConfig<T> = keyof T extends never
-  ? Record<string, never>
-  : {
-        [K in keyof T as undefined extends T[K] ? never : K]: T[K];
-      } extends infer R
-    ? keyof R extends never
-      ? Record<string, never>
-      : R
-    : never;
+type EmptyObject = Record<string, never>;
```
sodic: Very very unimportant, but interesting: https://github.com/wasp-lang/wasp/pull/1235#discussion_r1300022520

waspc/packages/wasp-config/__tests__/testFixtures.ts
```diff
@@ -1,48 +1,122 @@
...
+ *   b: {
+ *     c: boolean;
+ *   };
+ *   f: {
+ *     g: boolean;
+ *     h: string;
+ *   }[];
+ * };
+ */
+type MinimalConfig<T> =
```
sodic, line 49: This guy is currently broken, more details in other comments.

waspc/packages/wasp-config/__tests__/testFixtures.ts
sodic: Nice explanations! And good thinking with the arrays.

We still seem to have a bug though, check [this comment](https://github.com/wasp-lang/wasp/pull/2762#discussion_r2142906231) for more.
I think we need tests for these guys after all.

waspc/packages/wasp-config/__tests__/testFixtures.ts
FranjoMindek: Hm, no if we do just `else` for `"named"` `kind` after `if (kind === "default")`.
Otherwise we do.

We kinda do the `else` trick on every config getter.
Even though `kind` can only be `"named" | "default"`, TS forces to handle unknown `kind` case.

waspc/packages/wasp-config/__tests__/testFixtures.ts
FranjoMindek: This one works well.

waspc/packages/wasp-config/__tests__/testFixtures.ts
FranjoMindek: Honestly, I so often inline `map` functions that I forgot I could do this.

waspc/packages/wasp-config/__tests__/testFixtures.ts
FranjoMindek: I agree, but it kinda falls apart when there is no keys? Or?
It's weird.

I could do something funny, but I don't necessarily see an advantage to index signature here:
```ts
{ [thisKeyDoesNotExist: string]:  never };
```

Do we have a mapping for every key? Well yeah, because there are no keys, or every key is `never` 🤔 ?

waspc/packages/wasp-config/__tests__/testFixtures.ts
FranjoMindek: I've extracted object logic of both types as a separate type to make it more readable.

This should fix the issue:
```ts
type MinimalConfig<T> =
  T extends Branded<infer TType, infer TBrand>
    ? Branded<TType, TBrand>
    : T extends Array<infer TArrayItem>
      ? Array<MinimalConfig<TArrayItem>>
      : T extends object
        ? keyof T extends never
          ? EmptyObject
          : MinimalConfigObject<T>
        : T;

type MinimalConfigObject<T> = {
  [K in keyof T as T[K] extends Required<T>[K] ? K : never]: MinimalConfig<
    T[K]
  >;
} extends infer TObject
  ? TObject extends EmptyObject
    ? EmptyObject
    : TObject
  : never;
```

waspc/packages/wasp-config/__tests__/testFixtures.ts
FranjoMindek: I've extracted object logic of both config types.

It applies to each member of the union, which is what we want. Should be good?

FranjoMindek: e.g.
```ts
type TestType =
  | "a"
  | { a?: string }
  | { b: number }
  | { c: { d?: string }[] };
  
type Result = MinimalConfig<TestType>;
type Result = "a" | EmptyObject | {
    b: number;
} | {
    c: EmptyObject[];
}
```

FranjoMindek:
Didn't add any type tests yet. Do we want that as another issue or here?
Would also love your comment on old `TsAppSpec` type, what should be its new name? Currently it is `AppTsAppSpec`.

sodic:
> Didn't add any type tests yet. Do we want that as another issue or here?

Good question! I'll continue the discussion in the relevant thread: https://github.com/wasp-lang/wasp/pull/2762#discussion_r2143901923

> Would also love your comment on old TsAppSpec type, what should be its new name? Currently it is AppTsAppSpec.

Same thing here

waspc/packages/wasp-config/__tests__/testFixtures.ts
sodic: It still has the first problem :)

waspc/packages/wasp-config/__tests__/testFixtures.ts
sodic: Ok, fair enough. If you're wondering why, see page 15 of Effective TS: Type operations cannot effect runtime values.

Here's how we can make it clearer
```ts
// ...
   } else {
    return assertUnreachable(scope);
  }
// ...

function assertUnreachable(value: never): never {
  throw new Error(`Unhandled case: ${value}`);
}
```
Let's apply this pattern to all "impossible default cases." See Effective TS, 2nd edition, item 59 for more details.

Finally, I'd probably change this to a switch. I personally find it a little cleaner in cases like this, but not important.

waspc/packages/wasp-config/__tests__/testFixtures.ts
sodic: Ok, I see the comment.

In that case, this constant should be called `PAGE_CONFIG_TYPES`, right? Can we change that too :pleading_face:

waspc/packages/wasp-config/__tests__/testFixtures.ts
sodic: Ah yeah, I guess you're right.

Choose whatever you like and resolve. `Record` is fine by me.

waspc/packages/wasp-config/__tests__/testFixtures.ts
sodic: Ok great, I think that's fine! Feel free to resolve.

waspc/packages/wasp-config/__tests__/testFixtures.ts
sodic: One small improvement I see is that we don't need the `infer` for the brand, we can just use `T` directly I believe:
```ts
type MinimalConfig<T> =
  T extends Branded<unknown, unknown>
    ? T 
    : ...
```
Fewer repetitions and less room for errors.

I wouldn't poke around the type too much besides that, not until we get the tests. If we're not doing them now, please test them thoroughly before merging (use the cases where we caught previous errors).

waspc/packages/wasp-config/__tests__/testFixtures.ts
sodic: Ideally we'd have the tests immediately (since types were broken before and then we accidentally broke them in a different way).

But that's not the point of this PR, so I say create another issue. If you can take care of it in a different PR right away, that would be awesome.

waspc/packages/wasp-config/src/publicApi/App.ts
sodic: Man, looking at this almost changes my mind about what I said here: https://github.com/wasp-lang/wasp/issues/2737#issuecomment-2886174097

These types naturally belong with the `App` -- they're the types of its arguments. So I'm even inclined to say "screw it, bring it all together under `App` and we'll suck up the confusion with `App` and `AppConfig`."

If not, we should kick out `AppTsAppSpec` because the rest of the code expects _that thing_ to be named `TsAppSpec`.

We should instead rename the `tsAppSpec.ts` file. Maybe to something like `appBuildingBlocks.ts` or `appConfigs.ts` or `declarations.ts`?

@Martinsos you're probably busy, but maybe you have an idea. It will take a while for you to get into this though, so choose whether you want to wisely :)

waspc/packages/wasp-config/src/index.ts
sodic: We're taking care of this entire discssuion in the other thread: https://github.com/wasp-lang/wasp/pull/2762#discussion_r2143946130.

waspc/packages/wasp-config/__tests__/testFixtures.ts
sodic: I see the rocket, but there's still many of those left to rename:

image

FYI, how I found this... I used a regex search for `:.*Config`, the idea being that no nested objects in the file should have a config suffix in the new naming scheme:

image

FranjoMindek: Hey I did change the AuthConfig subtypes to not have Config suffix, but then I noticed that AppSpec has Config suffix on them anyway. If you check my git history you will see that I reverted it.

So I wanted to stop and ask the question of what to do with them? But forgot @sodic 😅

sodic: Oh man :sweat_smile: 

I guess we leave it as is. I see no better option. Nice work being thorough.

waspc/packages/wasp-config/src/publicApi/App.ts
FranjoMindek: We moved `TsAppSpec` back to `tsAppSpec.ts`

FranjoMindek: And we selectively export in `index.ts`

waspc/packages/wasp-config/__tests__/testFixtures.ts
FranjoMindek: Solved

waspc/packages/wasp-config/__tests__/testFixtures.ts
FranjoMindek: We have tests now.

waspc/packages/wasp-config/__tests__/testFixtures.ts
sodic: Oh, sorry, just seeing this now - let's remove the `T` prefix on types please :)

waspc/packages/wasp-config/src/index.ts
```diff
@@ -1 +1,35 @@
...
+  JobConfig,
+  PageConfig,
+  PageName,
+  PasswordResetConfig,
+  Perform,
+  QueryConfig,
+  RouteConfig,
+  Schedule,
+  ServerConfig,
+  // We don't want to export this type.
```
sodic: Let's explain why

waspc/packages/wasp-config/package.json
```diff
@@ -16,17 +16,22 @@
...
     "test": "vitest --reporter=verbose",
     "test:integration": "vitest --reporter=verbose --project integration",
-    "test:unit": "vitest --reporter=verbose --project unit"
+    "test:unit": "vitest --reporter=verbose --project unit",
+    "test:type": "vitest --reporter=verbose --project type"
   },
+  "//": [
+    "We pin `vitest` version because type tests are an experimental feature,",
+    "and changes to API are not breaking changes."
+  ],
```
sodic: Can we put this comment above vitest, in the devDependencies array? Or will it break?

waspc/packages/wasp-config/.gitignore
```diff
@@ -3,3 +3,4 @@ coverage/
 dist/
 .env
 .vscode/
+tsconfig.vitest-temp.json
```
sodic, line 6: What's this?

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
sodic, line 1: Is this the convention? To name the files something.test-d.ts?

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
```diff
@@ -0,0 +1,153 @@
+import { describe, expectTypeOf, test } from "vitest";
+import { Branded } from "../src/branded.js";
+import { EmptyObject, FullConfig, MinimalConfig } from "./testFixtures.js";
+
+describe("MinimalConfig<T>", () => {
+  test("should handle primitive types", async () => {
```
sodic: The test descriptions could be more descriptive. For example:
- Instead of "should handle primitive types," we could say "should not affect primitive types"
- Instead of "should handle branded type," we could say "should not affect branded types"
- Instead of "should handle nested optional props object," we could say "should remove optional props in nested objects"
- etc.

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
```diff
@@ -0,0 +1,153 @@
...
+  anotherProp?: number;
+};
+
+type RequiredObject = Required<OptionalObject>;
+
+type OptionalNested<T> = {
+  nested?: T;
+};
+
+type RequiredNested<T> = Required<OptionalNested<T>>;
```
sodic: These names could also be more descriptive. I suggest:
- `ObjectWithRequiredPrimitiveProperties`
- `ObjectWithOptionalPrimitiveProperties`
- `ObjectWithNestedOptionalProperties`

If we're using utility types, I recommend defining required variants and then adding in the optionality, not the other way around.

Finally, I'm not sure about `NoPropertiesObject`. It does make sense in terms of the test, but looking just at its definition, the name is wrong. Like the comment says, `{}` does not represent an object without properties. For example, this code would work:
```ts
const x: NoPropertiesObject = "something"
```
Which is problematic. So I suggest inlining `{}` into the test.

waspc/packages/wasp-config/package.json
FranjoMindek: Not allowed inside of `devDependencies`, only packages are. 
Can't be above `devDependencies` because of the formatting.

sodic: Stupid json...

waspc/packages/wasp-config/.gitignore
FranjoMindek: It's a file that exists while the the vitest watch is active, which is default for all our vitest scripts.
The problem is that if the project is closed unexpectedly, this file doesn't get cleaned up by the vitest.
To avoid accidental commit in those scenarios I've added it here. 
Also won't bother you if your vitest watch is currently active while you commit.

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
FranjoMindek: That is what I found in `vitest` examples and/or randomly on internet while researching it.

waspc/packages/wasp-config/.gitignore
sodic: Is adding this standard procedure for vitest? If not, can we add a commnet. If yes, no need.

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
FranjoMindek: I've changed tests themselves, so the type names no longer make sense.

>If we're using utility types, I recommend defining required variants and then adding in the optionality, not the other way around.

For me there is no real difference, between going optional or required first, just that one is defined through the other. But I get the stance of having required object defined first. So I will change it.

For some reason, `Partial` besides making field optional also allows for fields to be defined but with value of undefined.
This is the LSP information on hover: 
```ts
type Partial<T> = { [P in keyof T]?: T[P] | undefined; }
```
While source code in `lib.es5.d` shows:
```ts
type Partial<T> = {
    [P in keyof T]?: T[P];
};
```
weird :D.

waspc/packages/wasp-config/.gitignore
FranjoMindek: From docs:

Vitest allows you to write tests for your types, using `expectTypeOf` or `assertType` syntaxes. By default all tests inside `*.test-d.ts` files are considered type tests, but you can change it with [`typecheck.include`](https://vitest.dev/config/#typecheck-include) config option.

I would say it's standard.

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
sodic, line 1: This file is called `testFixtures.test-d.ts` but it contains tests as well.

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
```diff
@@ -0,0 +1,153 @@
...
+  });
+
+  test("should handle empty object", async () => {
+    expectTypeOf<
+      MinimalConfig<NoPropertiesObject>
+    >().toEqualTypeOf<EmptyObject>();
+  });
+
+  test("should handle optional props object", async () => {
+    expectTypeOf<MinimalConfig<OptionalObject>>().toEqualTypeOf<EmptyObject>();
```
sodic: The `EmptyObject` is exporeted and used as an expected test output (I think that's the only reason it's exported).

We're refering to the implementation to write the test. If someone makes a mistake in `EmptyObject`, the tests will pass but the type will be incorrect.

Let's define test outputs independently of the implementation. You can duplicate the definition or even inline it.

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
```diff
@@ -0,0 +1,95 @@
...
+    expectTypeOf<FullConfig<RequiredObject>>().toEqualTypeOf<RequiredObject>();
+  });
+
+  test("should make optional props required", async () => {
+    expectTypeOf<FullConfig<OptionalObject>>().toEqualTypeOf<RequiredObject>();
+  });
+
+  test("should not affect required props", async () => {
+    expectTypeOf<FullConfig<RequiredObject>>().toEqualTypeOf<RequiredObject>();
+  });
```
sodic: Isn't this test identical to one of the previous ones?

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  test("should handle objects recursively", async () => {
+    expectTypeOf<FullConfig<{ nested: unknown }>>().toEqualTypeOf<{
+      nested: FullConfig<unknown>;
+    }>();
+  });
```
sodic: We're missing a test for optional nested properties. Same with `MinimalConfig`.

You can do it in the same test - one prop is optional, the other isn't.

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  test("should not affect required props", async () => {
+    expectTypeOf<FullConfig<RequiredObject>>().toEqualTypeOf<RequiredObject>();
+  });
+
+  test("should handle objects recursively", async () => {
+    expectTypeOf<FullConfig<{ nested: unknown }>>().toEqualTypeOf<{
+      nested: FullConfig<unknown>;
+    }>();
```
sodic: What happened to sending unions instead of unknown?
```ts
expectTypeOf<FullConfig<AnyPossibleConfigType>>().toEqualTypeOf<{
    nested: FullConfig<AnyPossibleConfigType>;
}>();
```
It didn't work?

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
```diff
@@ -0,0 +1,95 @@
...
+  });
+
+  test("should handle objects recursively", async () => {
```
sodic: This test has the wrong description. Same with the one in `MinimalConfig`.

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  test("should make optional props required", async () => {
+    expectTypeOf<FullConfig<OptionalObject>>().toEqualTypeOf<RequiredObject>();
+  });
+
+  test("should not affect required props", async () => {
+    expectTypeOf<FullConfig<RequiredObject>>().toEqualTypeOf<RequiredObject>();
+  });
+
+  test("should handle objects recursively", async () => {
```
sodic: A better description is `should recrusively apply to required nested properties`.

sodic (changes requested):
Spotted some mistakes in tests, let's fix those.

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
FranjoMindek: I don't understand the problem here.
`testFixtures` -> the file whose logic we are testing (`testFixtures.ts`)
`test-d.ts` -> extension used to mark type tests in `vitest`

Type tests are typed under `test` in `vitest`.

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
FranjoMindek: Is that necessary? Its expected to act the same as any other optional props which we are already testing for. It's not really related to testing the object's recursive behavior.

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
FranjoMindek: It works, but our type is made to work with any type.
`AnyPossibleConfigType` is basically `any` type.

We just need to know that our type is applied to each prop of the object or each item of the array.

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
FranjoMindek: Is there an distinction?
This is testing two different functionalities of our type.
I don't think we need to think about that if we covered it in our base cases?
Once is handle optional/required props, another one is handle objects recursively.

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
sodic: Ah yes, you're right. I forgot we were testing the types we use in tests. All good here, my bad!

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
sodic: We can (and should) remove the export on the `EmptyObject` in `testFixtures.ts`.

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
sodic: Ok, fair enough!

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
sodic: Oh, I now see what you meant with "should handle objects recursively." The object in the sentance refers to the nested object, I thought it refers to the top object.

Since it confused me, might I suggest "Should recursively apply to nested objects."

One final idea, since we can't do property tests (i.e., use generic types), what do you think about testing it on a doubly nested object, like this:
```ts
  test("should handle objects recursively", async () => {
    expectTypeOf<FullConfig<{ nested1: { nested2: unknown } }>>().toEqualTypeOf<{
      nested1: { nested2: FullConfig<unknown> };
    }>();
  });
```
And, since we're using unknown here, should we also use it in our fixtures:
```ts
type RequiredObject = {
  prop: unknown;
};
```

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
sodic: Hmm, I guess you're right. No need then.

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
sodic: > For some reason, Partial besides making field optional also allows for fields to be defined but with value of undefined.
Yeah, this always confused me too. I don't know why that's the case.

> These names could also be more descriptive

I still stand by this suggestion though, `ObjectWithRequiredPrimitiveProperties` is better than `RequiredObject`.

sodic (approved):
Nice work! Only three unresolved threads left.

I'm approving, but:
- We should definitely remove that export
- Please consider the other two threads as well

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
FranjoMindek: Did, same for `testFixtures.test-d.ts`

waspc/packages/wasp-config/__tests__/testFixtures.test-d.ts
FranjoMindek: I changed to double nested, it might not do anything functionally but it better explains what we test for.

On the upside, picking `unknown` made me discover a type bug.

I've switched:
```ts
type MinimalConfigObject<T> = {
  [K in keyof T as T[K] extends Required<T>[K] ? K : never]: MinimalConfig<
    T[K]
  >;
} extends //...
// to
type MinimalConfigObject<T> = {
  [K in keyof T as EmptyObject extends Pick<T, K> ? never : K]: MinimalConfig<
    T[K]
  >;
} extends //...
```

In the first approach:
```ts
type SomeObject = {
  prop?: unknown;
};
type Test = MinimalConfig<SomeObject>;
// would return
// { prop?: unkown }
// instead of
// EmptyObject
```
This is because `unknown | undefined` is same as `unknown`.

Rather than dealing with values, the second approach tests field optionality directly. It's not affected by field values.

sodic (approved):
I've got nothing else to add. Nice work and thanks for the patience :)

Issue #2703, opened 2025-04-28
Check Dev Container configuration


cprecioso:
We should check that our Codespace works and either fix it (and implement warnings for us if it doesn't), or remove it if it's not used much

---

> Then I tried to run it in the codespaces by following the [README](https://github.com/wasp-lang/wasp/tree/main/examples/tutorials/TodoApp), it seems like the `wasp` is pre-installed, but I'm not able to execute it:
> 
> ```
> ➜ .../wasp/examples/tutorials/TodoApp (release) $ wasp
> /home/codespace/.local/share/wasp-lang/0.16.3/wasp-bin: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /home/codespace/.local/share/wasp-lang/0.16.3/wasp-bin)
> /home/codespace/.local/share/wasp-lang/0.16.3/wasp-bin: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /home/codespace/.local/share/wasp-lang/0.16.3/wasp-bin)
> /home/codespace/.local/share/wasp-lang/0.16.3/wasp-bin: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /home/codespace/.local/share/wasp-lang/0.16.3/wasp-bin)
> ``` 

 _Originally posted by @Louis-7 in [#2702](https://github.com/wasp-lang/wasp/issues/2702#issuecomment-2834243459)_

sodic:
Ah man, here we go agian 😄 

**TL;DR** at the end. 

Corre issue and some background: https://github.com/wasp-lang/wasp/issues/2705

Why is this  happening in Codespaces:
- The LTS period for Ubuntu 20 recently expired, so GitHub forced us to bump our CI to ubuntu 22.04 ([#2539](https://github.com/wasp-lang/wasp/issues/2539)).
- We set our CI to use Ubuntu 22.04 ([#2566](https://github.com/wasp-lang/wasp/pull/2566)).
- Our codespaces OS is based on Ubuntu 20.04, and we get the exact situation I described above.

Funnily enough, on first glance, you'd think our Codespaces OS is based on Ubuntu 22.04. But that's just the kernel. Userland features are still stuck on 20.04:

![Image](https://github.com/user-attachments/assets/7d13944e-da91-4e9a-863e-761779aab4b0)

## TL;DR
[...]

Issue #2283, opened 2024-09-15
Speed up our CI


Martinsos:
Our CI got a bit slow. We should look into speeding it up, as it will help us iterate faster. Improved caching? More stuff in parallel?

Martinsos:
We will want to run tests in phases, from cheaper to more expensive/slow.
For example, the unit tests should run every time.
Then, example apps tests should run next, probably also on every PR.
Starter templates tests → those can maybe run less often, and only once example apps tests pass.
Some smart division like this can make CI lighter / faster.

Martinsos:
- caching?
- multiple workflows?
- build CLI and upload as artefact?
- run only when it makes sense?
- docker something?

Let's do an RFC!

sodic:
Some of @cprecioso's ideas from the call we just had:
- Different jobs for testing, building, publishing, releasing. They can run in parallel or sequentially (depending on blockers)
- On each commit: we test, build, and push build artifacts.
- On each release (tag): we pull the appropriate artifact and create a release.
- Keep publishing the release manually for now (check #2829 for more on this).

cprecioso:
RFC ready at https://www.notion.so/wasp-lang/RFC-CI-Architecture-The-Fast-and-the-Furious-24818a74854c8049af39f3e0b2af82a4?source=copy_link

Issue #2003, opened 2024-05-03
Investigate building Wasp for more Linux architectures


sodic:
Currently (as of 0.13.2), we're releasing Wasp Linux builds only for Ubuntu x86. It works on most of the Linux distribution, but not all of them (e.g., Raspbian).

For users that don't run a compatible distribution, the only alternative is building Wasp from source source.

We should consider some of the other more popular targets (e.g., ARM) if these reports keep popping up. 

Related: #1100

Users running into this:
 - https://discord.com/channels/686873244791210014/1216022996020625458/1216022996020625458

Martinsos:
I coudl bet we had an issue about making our compilation more static, but I can't find it -> that is also related. Right now it is not completely static, but quite static (with glibc being the main culprit).

Martinsos:
For ARM, there was also a quite big discussion here: https://discord.com/channels/686873244791210014/943092165964349500/1149268438888501278 .

Some excerpts:

> OK, so if wasp CLI binary runs on ARM, then that is solved already -> the only thing left is fixing Dockerfile so it can build for ARM, right? And @shayne you said it is enough to just move python installation up, while @Zeko369 you suggested using node-slim.
What is our conclusion then? Maybe it is easiest to do node-slim so we ensure we have basic stuff installed and that is it?

> So TLDR, install LLVM 13 from homebrew and export it and then you can build Wasp on Arm just fine 👍 Ping if that does not work for you @Zeko369 and sorry I didn't document, but I was previously the only one on M1 and I forgot that one tiny extra step that was required. Additionally, I am unsure if newer LLVM versions work or it has to be 13, but you can use that just to be safe if you want (which I would recommend).

> TODOs:
> 1. node-alpine -> node-slim => faster build (40s -> 20s) + works on ARM
> 2. setup arm CI so we can deliver native ARM CLI binaries (or at least test that)

But I recommend reading the whole discussion!

Martinsos:
Related convo: https://discord.com/channels/686873244791210014/1271559114446602301/1271559114446602301

-> wasp cli doesn't run on Raspberry Pi

cprecioso:
Related #2804

PR #2768, opened 2025-05-19
CI: Build more variants of the executable (static libc, macos arm, macos universal)

cprecioso:
### Description 

Context: [RFC](https://www.notion.so/wasp-lang/RFC-Building-binaries-with-wide-libc-support-1f818a74854c809e93f8eddfd2d618b8?pvs=4)

This PR extracts the building and packaging logic from `wasp-ci.yaml` to its own `waspc-build.yaml`.

It tries to simplify the steps as much as possible. The intention is that the `steps` are easily readable and understandable in a single screenful.

Most complexity lays in how we define the environments we run in, as we declare there the OS-specific dependencies we need to install.

Through this abstraction, we can define in a single place _all_ our build targets, keeping compatibility workarounds to a minimum:

- `wasp-linux-x86_64`: built on a Ubuntu 20.04 Docker image so we have a low `glibc` requirement
- `wasp-linux-x86_64-static`: **(NEW)** built on an Alpine image and with [the `--enable-static-build` cabal flag](https://cabal.readthedocs.io/en/stable/cabal-project-description-file.html#cfg-flag---enable-executable-static), so musl gets embedded into the binary and we're completely independent of the user's distro
- `wasp-darwin-x86_64`: no changes, built on the GitHub macOS runner
- `wasp-darwin-amd64`: **(NEW)** we use the new GitHub macOS ARM runners to build an Apple Silicon binary
- `wasp-darwin-universal`: **(NEW)** I do an extra step at the end of picking up both macOS binaries and creating a Universal binary out of both of them.

All binaries are uploaded as artifacts of the workflow run, which means they are available to later steps. They are also accessible to us through [the workflow's summary page](https://github.com/wasp-lang/wasp/actions/runs/15134136572?pr=2768).

> [!IMPORTANT]
> This workflow is not active yet. We can run it manually [through GitHub's interface](https://github.com/wasp-lang/wasp/actions/workflows/waspc-build.yaml) when it comes to the default branch. I first sent the approach so we can discuss it by itself. The automation is complete and available at #2773.

### FAQs

<details>
<summary>Why do we need to manually specify so many dependencies when we weren't doing it before?</summary>

Before, we were building on the GitHub runners, which come with a lot of dependencies already installed (even GHC!). Now we're building inside Docker images, and they are usually bare-bones, so they are missing many basic dependencies (e.g. Ubuntu is missing `curl` and Alpine is missing `bash`).

</details>

<details>
<summary>Why build inside of Docker?</summary>

So we don't depend on GitHub Action runners anymore. They keep updating them, changing the available software, and deprecating old runners. If we want to keep building on an old Glibc, we need pin the environment through Docker.

</details>

<details>
<summary>Why do we build an ARM image for macOS and not Linux?</summary>

GHC 8.10.7 has support for ARM on macOS but not on Linux. Updating to a newer version (#1446) will enable us to do ARM binaries for Linux.

</details>

<details>
<summary>Why are we now able to build executables for macOS on ARM?</summary>

New images from GitHub became available.

</details>

<details>
<summary>Why have two binaries for Linux and three for macOS?</summary>

These are implemented in order for us to continue building in different environments and testing before release. We don't need to publish all of them.

I'd suggest to publish the Linux static and the macOS universal only. If we agree on that we can remove the others.

</details>

<details>
<summary>Can we statically link the macOS binaries?</summary>

No. The syscall interface is an internal API in macOS, so we must dynamically link to `libSystem.dylib`, which is the public API. Fortunately, libSystem is much lighter than libc, and quite stable between releases, so it is not a big problem there.

</details>

<details>
<summary>Why extract this into a new workflow?</summary>

`wasp-ci` is already quite heavy to read through, and full of workarounds. Having the build there would have been more messy than helpful. We can still call to this build from other workflows, so we don't lose functionality.

</details>

---

- Related to #2773 
- Related to #2703
- Related to #2003
- Related to #2283


.github/workflows/check-formatting.yaml
```diff
@@ -37,8 +37,6 @@ jobs:
       - uses: actions/checkout@v4
 
       - uses: ./.github/actions/setup-haskell
-        with:
-          cabal-project: dev-tool.project
```
cprecioso, line 41: Leftover from previous version of the action

.github/actions/setup-haskell/action.yaml
```diff
@@ -19,18 +25,25 @@ runs:
...
 
     # Based on the official recipe for Cabal caching:
-    # https://github.com/actions/cache/blob/5a3ec84eff668545956fd18022155c47e93e2684/examples.md#haskell---cabal
+    # https://github.com/actions/cache/blob/v4.2.3/examples.md#haskell---cabal
     - name: Cache Cabal dependencies
       uses: actions/cache@v4
       with:
         path: |
           ${{ steps.setup-haskell.outputs.cabal-store }}
-          waspc/dist-newstyle
```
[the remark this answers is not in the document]
Martinsos: I guess because it produces artifacts based on the system it is building on, yeah.
One solution though is probably to add the needed info to the cache key, the one that will differentiate between different systems (arch info or what is it, since we already have os info), then we will have one cache for each arch. We might end up with too many caches though hm.

cabal-store doesn't cause any issues, being on different arches? Stuff is only downloaded there, not built?

What do we lose by not caching dist-newstyle -> how slower is the CI?

.github/actions/setup-haskell/action.yaml
```diff
@@ -19,18 +25,25 @@ runs:
...
         # Caches are immutable, so we use the official recipe for simulating an updatable cache:
-        # https://github.com/actions/cache/blob/5a3ec84eff668545956fd18022155c47e93e2684/tips-and-workarounds.md#update-a-cache
-        key: ${{ runner.os }}-cabal-${{ inputs.ghc-version }}-${{ hashFiles('waspc/*.cabal', 'waspc/*.project', 'waspc/*.project.freeze') }}-${{ github.run_id }}
+        # https://github.com/actions/cache/blob/v4.2.3/tips-and-workarounds.md#update-a-cache
```
Martinsos, line 44: I remember we used to have a pretty rich comment here before, but it is gone now. Would it make sense to return some of that? Because I am reading this now, and even though I went and read the link, I am not able to quickly grasp what is this about, even though it was me who wrote both this code and that comment some longer time ago. Since this is a tricky thing here, I think explaining it more in our words would be beneficial.

.github/actions/setup-haskell/action.yaml
```diff
@@ -19,18 +25,25 @@ runs:
...
+        key: |
+          ${{ runner.os }}-cabal-${{ inputs.ghc-version }}-${{ hashFiles('waspc/*.cabal', 'waspc/*.project', 'waspc/*.project.freeze') }}-${{ github.run_id }}
         restore-keys: |
           ${{ runner.os }}-cabal-${{ inputs.ghc-version }}-${{ hashFiles('waspc/*.cabal', 'waspc/*.project', 'waspc/*.project.freeze') }}-
```
Martinsos, line 48: I think we don't use cabal.project.freeze if I am correct, in Wasp project.

Martinsos (commented):
Left a bit of comments! From what I get this is only part of the big change, so there isn't that much to review actually.

> No. The syscall interface is an internal API in macOS, so we must dynamically link to libSystem.dylib, which is the public API. Fortunately, libSystem is much lighter than libc, and quite stable between releases, so it is not a big problem there.

-> let's add this to the code somewhere, as a comment!

> So we don't depend on GitHub Action runners anymore. They keep updating them, changing the available software, and deprecating old runners. If we want to keep building on an old Glibc, we need pin the environment through Docker.

-> this should also be explained in the comment in the code!

.github/workflows/waspc-build.yaml
```diff
@@ -0,0 +1,155 @@
...
+      matrix:
+        # == Why such a big, heterogeneous list? ==
+        # This is a bit of mish-mash of different platforms and architectures,
+        # so we need to build some of these directly in the runners, and some in
+        # containers. Each environment is a different OS and needs different
+        # dependencies.
+        env:
+          - name: linux-x86_64
+            runner: ubuntu-latest
+            container: ubuntu:20.04 # Old Ubuntu version so we have a low glibc requirement
```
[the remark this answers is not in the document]
Martinsos: Maybe also super shortly explain the interplay between the github runner and docker container? I have to admit I am not sure exactly how taht goes. I understand we have a docker container ubuntu:20.04 running on the github runner ubuntu-latest, but I am not sure what is the usual dynamic -> for example `install-deps` below, are those still executing on the runner, or in the container? If on the runner, why not in the container? To how much of the host system does container have access to?

.github/workflows/waspc-build.yaml
```diff
@@ -0,0 +1,155 @@
...
+            install-deps: |
+              export DEBIAN_FRONTEND=noninteractive
+              apt-get update -y
+              # GHCup dependencies (https://www.haskell.org/ghcup/install/#version-2004-2010)
+              apt-get install -y build-essential curl libffi-dev libffi7 libgmp-dev libgmp10 libncurses-dev libncurses5 libtinfo5
+              # Cabal dependencies
+              apt-get install -y zlib1g-dev
+
+          # TODO: Add a Linux ARM64 build once we update the GHC version (#1446)
```
Martinsos, line 52: Nice, that will be cool!

.github/workflows/waspc-build.yaml
```diff
@@ -0,0 +1,155 @@
...
+          # GHC 8.10.7 does not support ARM64 on Linux yet
+
+          - name: linux-x86_64-static
+            runner: ubuntu-latest
+            # actions/setup-node does not work in alpine to work around this, we
```
Martinsos, line 57: How is that? I guess that is because actions run on the runner, not in the container, or is it sometihng else?

.github/workflows/waspc-build.yaml
```diff
@@ -0,0 +1,155 @@
...
+            # use the alpine variant of the official node image, which already
+            # has it installed.
+            container: node:${{ inputs.node-version }}-alpine
+            skip-node-install: true
+            static: true
+            install-deps: |
+              apk update
+              # GHCup dependencies (https://www.haskell.org/ghcup/install/#linux-alpine)
+              apk add binutils-gold curl gcc g++ gmp-dev libc-dev libffi-dev make musl-dev ncurses-dev perl pkgconfig tar xz
+              # Build script dependencies
```
Martinsos, line 67: Maybe stupid question, but which build script?

.github/workflows/waspc-build.yaml
```diff
@@ -0,0 +1,155 @@
...
+              # Cabal dependencies
+              apk add zlib-dev zlib-static
+
+          - name: darwin-x86_64
+            runner: macos-13 # Latest image still based on Intel architecture that can be used for free
+
+          - name: darwin-aarch64
+            runner: macos-latest # Latest macOS images are already Apple Silicon-based
+            install-deps: |
+              # We need to install llvm@13 for building on Apple Silicon (prebuilt libraries
```
Martinsos, line 78: I believe this problem will be avoided once we update to the new GHC?

.github/workflows/waspc-build.yaml
```diff
@@ -0,0 +1,155 @@
...
+
+          # TODO: Add a Linux ARM64 build once we update the GHC version (#1446)
+          # GHC 8.10.7 does not support ARM64 on Linux yet
+
+          - name: linux-x86_64-static
+            runner: ubuntu-latest
+            # actions/setup-node does not work in alpine to work around this, we
+            # use the alpine variant of the official node image, which already
+            # has it installed.
+            container: node:${{ inputs.node-version }}-alpine
```
Martinsos, line 60: I see we are not fixing the alpine version. From experience, that can result in trouble, due to latest version doing something different. We can protect from that by pinning down alpine version here, but risk that tehre won't be such version in combo with the inputs.node-version, or just leave it as it is and hope it won't cause issues. I am ok with leaving it as it is and seeing what happens, just wanted to be explicit about it.

.github/workflows/waspc-build.yaml
Martinsos: Ah, you meant our `run` script (I saw it being used below)? In that case, let's make that clear!

waspc/run
```diff
@@ -78,6 +79,8 @@ print_usage() {
     "Builds the Haskell project."
   print_usage_cmd "build:all" \
     "Builds the Haskell project + all sub-projects (i.e. TS packages)."
+  print_usage_cmd "build:all:static" \
```
Martinsos, line 82: Let's maybe explain here that this won't work unless musl is present instead of glibc, and also works only on Linux (if that is correct?). Basically let's explain the requirements/conditions, if there are some.

Martinsos (commented):
@cprecioso looking great all together! This is going to be very nice. I like a lot the whole deal with this being a separate workflow, it is much simpler to comprehend, and uploading artifacts is sweet. I don't think there is so much dependencies being installed as you mentioned, it isn't that much really.

Let's work through the comments, but there is nothing major!

Oh and if I got it right, you plan to then delete this logic from wasp-ci.yaml and of course the artifacts generated by this one, but you are leaving that step for some next PR?

.github/workflows/waspc-build.yaml
```diff
@@ -0,0 +1,155 @@
+name: Build binaries
```
Martinsos, line 1:
```suggestion
name: Build wasp CLI binary (for different archs)
```

.github/actions/setup-haskell/action.yaml
cprecioso: You're right, we're already getting hashing the cabal files so we don't need to update the cache.

.github/workflows/waspc-build.yaml
cprecioso: https://github.com/actions/setup-node/issues/1293

Node.js does not publish `musl` binaries oficially, so node is installed but doesn't run. They do publish `alpine`-based docker images, so we use that.

cprecioso: Improved the comment

.github/workflows/waspc-build.yaml
cprecioso: Kinda, when we update GHC we can just use a newer LLVM version that is not `disable!`d

.github/workflows/waspc-build.yaml
cprecioso: Yep, I am aware of this too.

.github/actions/setup-haskell/action.yaml
```diff
@@ -0,0 +1,56 @@
...
+    # permissions that lead to mismatches.
+    - name: Cache Cabal dependencies
+      uses: actions/cache@v4
+      with:
+        path: |
+          ${{ steps.setup-haskell.outputs.cabal-store }}
+        key: |
+          cabal-${{ runner.os }}-${{ runner.arch }}-${{ inputs.ghc-version }}-${{ hashFiles('${{ inputs.working-directory }}/*.cabal', '${{ inputs.working-directory }}/*.project', '${{ inputs.working-directory }}/*.project.freeze') }}
+        restore-keys: |
+          cabal-${{ runner.os }}-${{ runner.arch }}-${{ inputs.ghc-version }}-${{ hashFiles('${{ inputs.working-directory }}/*.cabal', '${{ inputs.working-directory }}/*.project', '${{ inputs.working-directory }}/*.project.freeze') }}
```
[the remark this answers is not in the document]
cprecioso: I though that we did need to specify it explictly but I re-read the documentation and indeed it seems we don't.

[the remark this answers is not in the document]
Martinsos: Yup, that is easy to think so. I was dealing with this quite a bit before so I remember some details from then, but if not, I would probably make the same mistake.

RAW

cprecioso:
All taken care of, will merge this and #2773 when the CI is green

.github/workflows/waspc-build.yaml
```diff
@@ -0,0 +1,155 @@
...
+
+# We never trigger this workflow directly.
+# We can call it manually (workflow_dispatch) or from other workflows (workflow_call).
+on:
```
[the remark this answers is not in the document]
sodic: Martin asks that [every time](https://github.com/wasp-lang/wasp/pull/2566#discussion_r2006282820) :smile: 

I fully get it though, this yaml stuff is horrendous.

PR #2773, opened 2025-05-20
CI: Publish the new builds to releases from now on

### LINKED ISSUE #1574: Add macos arm64 binary package to our releases
Right now, when releasing new version of Wasp, we include prebuilt Linux x86_64 binary package and also macos x86_64 binary package with it. These are installed via Wasp installer script when users are installing Wasp on their machine. If not using these binaries, users can compile Wasp directly from source for their machine.

Since Mac's new personal laptops/machines come with a different CPU arch, which is arm64, that means  we don't have a prebuilt binary package for those. They can still use our macos x86_64 binary via Rosetta (which emulates x86_64), but it would be nicer if we offered binary package that works directly on arm64.

Since we are building our binary packages via Github Actions (GA), that means we need a GA runner that is macos arm64. GA currently doesn't offer such runner in normal plan (only as a quite expensive large runner that is paid by minute), but allegedly will in Q1 of 2024 (https://github.com/actions/runner-images/issues/8439), so then we should be able to easily add it.
Right now we have three runners: ubuntu-20, macos-latest (which is x86_64) and win. If they add new arm64 runner, we should have four runners: ubuntu-20, macos-x86_84 (whatever is the correct label for it), macos-arm64 (whatever is the correct label for it), and win.

Then we can produce additional binary package on that new runner and we can make sure that our installer script installs it when appropriate!

Discussion about this on Discord that resulted in this issue: https://discord.com/channels/686873244791210014/1130493830647533568/1171852601655033914

### LINKED ISSUE #2705: Explore statically linking GLIBC or find a way to build on older systems
## Background

This issue was motivated by https://github.com/wasp-lang/wasp/issues/2703.

Quick background for anyone unfamiliar:
- GLIBC is the "GNU C Library", a core part of the Linux operating system. It's a bunch of wrappers around Linux kernel system calls (so it's essential for all programs).
- Wasp's executable depends on GLIBC.
- GLIBC is a *dynamically linked* library, meaning that it's not packaged into the executable during build. Instead, the executable expects to find it on the host system and "import" it dynamically.
- Problems occur when you run a program on a system with a GLIBC version that's older than the one you had during the program's compilation (because your program relies on features that don't exist on the host system).

So, what sometimes happens with Wasp:
1. We compile `wasp` on a new Linux system with a recent GLIBC version (e.g., Ubuntu 22.04 with GLIBC 2.34).
1. Users try to run `wasp` on an older Linux system with an older GLIBC version (e.g., Ubuntu 20.04 with GLIBC 2.31).
1. Wasp fails because it can't find the newer GLIBC functions it relies on. 

We usually fix the problem by telling our CI to use the lowest possible version for building Linux executables. This maximizes compatibility (explained [here](https://github.com/wasp-lang/wasp/blob/10cba640cdb51b1130db396c2b1b449c1ee43aff/.github/workflows/waspc-ci.yaml#L44)).

## The current problem
Wasp will probably always break for some portion of users:
- GitHub only allows us to build on Ubuntu versions still within their LTS period (the oldest being 22.04 at the time of writing).
- Users are possibly still using older versions despite the LTS expire (20.04 at the time of writing).

So there's always a gap between what Github allows us to build on and what some users are still running. This gap is the largest immediately after each LTS expiry.

The main question is:
> Does the problem affect a significant portion of our users and is it something we want to fix?

If so, I see two possible fixes:
- Figuring out how to build Wasp with an even older LTS-expired version.
- Exploring how to link GLIBC statically  (see [here](https://github.com/wasp-lang/wasp/issues/650#issuecomment-1180488040))

### LINKED ISSUE #2722: RFC: Build on older versions of Ubuntu for GLIBC compatibility
The Docker container for `ubuntu:20.04` does include the old `glibc` version. So in theory we could just run the compilation steps for the CI in that image (GitHub Actions allows you to set a Docker container as an environment for a step: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idcontainer

Workaround until we fix https://github.com/wasp-lang/wasp/issues/2705

cprecioso:
> [!NOTE]
> This PR goes on top of (#2768). The meat of the new build is there.

Modifies our GitHub actions so that we use the new build workflow. We leave the `wasp-ci` workflow just for CI checks, and create a new release workflow that gets triggered when we create a tag. This new workflow calls the build and the ci checks workflows, and if both are correct, publishes the new release with the build binaries.

Untitled-2025-05-09-1247 excalidraw

When we start pushing these binaries to our releases, we will have to modify the installer to pick them up.

---

- Fixes #2722
- Fixes #2705
- Fixes #1574
- Related to #2703
- Related to #2003
- Related to #2283


.github/workflows/create-release.yaml
```diff
@@ -0,0 +1,34 @@
...
+    needs: [ci, build]
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          pattern: wasp-*
```
Martinsos, line 25: I am curious, what if for example there are some old aftifacts up there, and they are also starting with `wasp-`? Would it pick up those also, even though we probably didn't want that to happen?

Not sure if this is real problem, but some quick ideas to possibly avoid this if it is (they might not be doable, I am making them up):
1. Pass version tag as input to build job, so that artifacts are marked with that someohw (in the name, or they are in such dir?), and then we reference them that way also.
2. Or maybe have build job return exact names of the artifacts it created (I don't konw if that is possible, if jobs can return outputs) and then supply those here.

.github/workflows/create-release.yaml
Martinsos, line 1: Let's maybe rename it to `wasp-release` or `wasp-create-release` or `wasp-create-gh-release` to stay consistent in the naming with `wasp-ci` and `wasp-build`?

.github/workflows/waspc-ci.yaml
Martinsos, line 1: Is `wasp-ci` still a good name? Is there a better name now, that this workflow does less? Maybe not, just checking. I never was sure what exactly Continuous Integration encompasses to be honest :D. But it does sound like it might be exactly the right name.

Martinsos, line 1: Should we update this file more? I see you haven't used the new setup-haskell action that you defined, so here we have duplicate logic for dealing with cache and stuff. Or mabye that happens in the later PR so that is why?

We also mention glibc here, talk about how we pick the specific ubuntu version, ... -> we should update that.

Let's do a pass through this whole file and make sure it is updated.

Martinsos (commented):
@cprecioso looking good! Exctied about this same as the other PR. NOthing super major, but left some comments, let's take care of those.

.github/workflows/waspc-ci.yaml
Martinsos: That makes sense, but is this really all about speeding up the CI (as 2283 aims for)? It is more about using the new things we added, if I am correct.

Ok let's go about it this way: it makes sense not to try to rework that file right now, but we should at least make it use the new haskell action + whatever was affected by the changes you did in this and that other PR, right? Since those are connected?

If you still feel that is too much for this PR / that other PR, I get that also, but let's make a separate issue for that to track it clearly, as I don't think it falls cleanly under the domain of #2283 .

Martinsos (approved):
Great, approved! But do please check that one standing comment before merging (about most likely creating another issue to track further changes in the wasp-ci.yaml).

cprecioso:
@Martinsos I updated `waspc-ci` to use the new action, not a lot of work
I will wait to merge this until the previous one is merged (https://github.com/wasp-lang/wasp/pull/2768)

PR #2709, opened 2025-04-29
Typecheck server code in build and CI

cprecioso:
Based on https://github.com/wasp-lang/wasp/pull/2672#pullrequestreview-2800969918

### Description

Fixes #2384

Moves the `ensure_app_compiles_and_builds.sh` script to the `waspc` folder, and in CI it will run it against `todoApp` and `crud-testing`.

It also adds `tsc --build` to the `npm run bundle` of the generated server.

### Select what type of change this PR introduces:

1. [ ] **Just code/docs improvement** (no functional change).
2. [ ] **Bug fix** (non-breaking change which fixes an issue).
3. [x] **New feature** (non-breaking change which adds functionality).
4. [ ] **Breaking change** (fix or feature that would cause existing functionality to not work as expected).

### Update Waspc ChangeLog and version if needed

If you did a **bug fix, new feature, or breaking change**, that affects `waspc`, make sure you satisfy the following:

1. [x] I updated [`ChangeLog.md`](https://github.com/wasp-lang/wasp/blob/main/waspc/ChangeLog.md) with description of the change this PR introduces.
2. [ ] I bumped `waspc` version in [`waspc.cabal`](https://github.com/wasp-lang/wasp/blob/main/waspc/waspc.cabal) to reflect changes I introduced, with regards to the version of the latest wasp release, if the bump was needed.

PR #2778, opened 2025-05-21
Type-check generated server code when building

### LINKED ISSUE #2458: Type-check server code when building
We haven't been type-checking server code (i.e., calling `tsc`) since #1714.

After implementing project references (#2247) and fixing all the errors, we should bring back the type checking.

cprecioso:
Fixes #2458

This was approved in #2709, but forgot to keep that when that PR was closed. We bring it back here.


cprecioso:
@infomiho this worked before merging the app runner for the headless tests, can you take a look? I think it's trying to run `npm run bundle` inside of the Docker container but it should be already bundled by then.

infomiho:
@cprecioso It's related to the building the server container, I believe we should maybe also copy the `tsconfig.json` i.e. modify the Dockerfile to allow `tsc` to run in the Docker.

cprecioso:
@infomiho

waspc/cli/src/Wasp/Cli/Command/Build.hs
```diff
@@ -115,6 +117,11 @@ build = do
           (waspProjectDir </> packageLockJsonInWaspProjectDir)
           packageLockJsonInBuildDir
 
+      liftIO $
```
infomiho, line 122: I'd add a comment that explains that we need this `tsconfig.json` file since the `server` config references the user TS config file. It took me a bit to figure this out when I was reviewing this.

PR #2780, opened 2025-05-21
Fix GHC warnings

cprecioso:
GHC was complaining:

```
src/Wasp/Project/Analyze.hs:19:1: warning: [-Wunused-imports]
    The qualified import of ‘Wasp.AppSpec.ConfigFile’ is redundant
      except perhaps to import instances from ‘Wasp.AppSpec.ConfigFile’
    To import instances alone, use: import Wasp.AppSpec.ConfigFile()
   |
19 | import qualified Wasp.AppSpec.ConfigFile as CF
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### DIFF
--- a/waspc/src/Wasp/Project/Analyze.hs
+++ b/waspc/src/Wasp/Project/Analyze.hs
@@ -16,7 +16,6 @@ import StrongPath
     fromAbsDir,
   )
 import qualified Wasp.AppSpec as AS
-import qualified Wasp.AppSpec.ConfigFile as CF
 import Wasp.AppSpec.Core.Decl.JSON ()
 import qualified Wasp.AppSpec.Valid as ASV
 import Wasp.CompileOptions (CompileOptions)


Martinsos:
> I've also seen this warning:
> 
> ```
> Defined but not used: ‘debouncerRunningEvents’
> ```
> 
> Referring to:
> 
> https://github.com/wasp-lang/wasp/blob/830d12cb570039587891c67a0fc7237d5aae73d6/waspc/waspls/src/Wasp/LSP/Debouncer.hs#L85-L95
> 
> I tried removing the `debouncerRunningEvents` definition, but it complains that `newtype`s must have at least one field. I don't know enough Haskell yet to know what to do, if someone could help out that'd be great.

I am guessing that `Debouncer` is used somewhere in the code, or you could just delete the whole type. But if `debouncerRunningEvents` is not used, that means `Debouncer` is constructed with positional arguments, not named ones. So somewhere we have `Debouncer foo`, and not `Debouncer { debouncerRunningEvents = x }`. So one soluition could be to rewrite the first form to the second one. That is anyway recommended with records, to not use positional arguments when constructing them. Although I am guessing this one has just one field so it was kinda ok.
If not that, then the solution would be to supress the error with some kind of comment. But I would go for the rewrite.

waspc/waspls/src/Wasp/LSP/Debouncer.hs
```diff
@@ -86,13 +86,13 @@ debounce (Debouncer running) waitMicros event fire = do
...
 --
 -- See 'debounce' for how to use it.
-newtype Debouncer k = Debouncer
-  { -- | A thread-safe map of event labels to async actions. This stores
-    -- actions for each event that are waiting for their debounce timers to end.
-    -- This is needed so that new calls to 'debounce' can cancel the currently
-    -- running actions before they can finish running.
-    debouncerRunningEvents :: STM.Map k (Async ())
-  }
+newtype Debouncer k = Debouncer (DebouncerRunningEvents k)
```
Martinsos, line 89: Ok that is one solution! I was suggesting the opposite though, calling it differently, not rewriting its definition. Field probably existed for a purpose, which was that it is a nice way to document what it is, that value. And this weird with the extra `type` is a bit unusual, not sure that it makes sense.

cprecioso: Ahh re-read the original comment and got it now, one sec

cprecioso: @Martinsos what about now

waspc/waspls/src/Wasp/LSP/Debouncer.hs
```diff
@@ -95,4 +95,6 @@ newtype Debouncer k = Debouncer
   }
 
 newDebouncerIO :: IO (Debouncer k)
-newDebouncerIO = Debouncer <$> STM.newIO
+newDebouncerIO = do
```
Martinsos, line 98: Boo we loose the nice short code ah :D. But yeah, this is it!

Martinsos (approved):
LGTM!

waspc/waspls/src/Wasp/LSP/Debouncer.hs
Martinsos: Oooh actually while at it, why don't we name that `m` variable something meaningful, to improve the code readibility? Woudl that make sense?

cprecioso: I don't know what it is, this is just the variable name Copilot suggested

cprecioso: I guess it's an STM.Map used for storing debouncing states

cprecioso: I'll leave it like this because the most appropriate name would be `debouncerRunningEvents` and that is what we immediately store it to.

Martinsos: > I don't know what it is, this is just the variable name Copilot suggested

-> not what I want to hear :D

But I agree with the final choice, due to context from the field name.

cprecioso: > not what I want to hear

Can't a guy get a little help when learning? 😅😂

cprecioso: Don't worry I did understand everything in the end 💁‍♀️

Martinsos: I am ok with `Copilot suggested` part, I am less ok with `"I don't know what it is"` :D. All good.

cprecioso: yep yep I got it 🙇‍♂️

PR #2781, opened 2025-05-21
Migrate Waspello to TS spec

### LINKED ISSUE #2748: Use Wasp TS spec in one example app
Given that the `thoughts` app was the only one using the TS spec and it was deleted: https://github.com/wasp-lang/wasp/pull/2746 we should convert one of the remaining example apps to use it. 

Some recommend doing it with Waspello, which makes sense since it's a complex app. One downside of using Waspello is that we no longer have that complex app with the DSL.

sodic:
Closes #2748.


examples/waspello/package.json
```diff
@@ -14,8 +14,9 @@
   "devDependencies": {
     "@types/react": "^18.0.37",
     "prisma": "5.19.1",
+    "tailwindcss": "^3.2.7",
     "typescript": "^5.1.0",
     "vite": "^4.3.9",
-    "tailwindcss": "^3.2.7"
+    "wasp-config": "file:../../../../../.local/share/wasp-lang/0.16.4/data/packages/wasp-config"
```
sodic, line 20: Currently there's no better alternative than leaving this here.

examples/waspello/main.wasp.ts
```diff
@@ -0,0 +1,86 @@
...
+
+app.action('createCard', {
+  fn: { import: 'createCard', from: '@src/cards/cards.js' },
+  entities: ['Card']
+});
+
+app.action('updateCard', {
+  fn: { import: 'updateCard', from: '@src/cards/cards.js' },
+  entities: ['Card']
+});
```
sodic, line 84: Should I do something clever here just to show what's possible?

examples/waspello/main.wasp.ts
sodic, line 1: Should I maybe add something to the README? Maybe point them to the docs on Wasp TS spec?

Martinsos: I think you should!

examples/waspello/main.wasp.ts
Martinsos: I think you should -> I would do what I did in the thoughts app, I think you can just copy that code, maybe tweak it a bit as you like.

Martinsos (approved):
LGTM, but I would act on two those two comments as you suggested yourself.

examples/waspello/main.wasp.ts
infomiho: My thoughts exactly, let's write it as best as we can.

examples/waspello/tsconfig.wasp.json
```diff
@@ -0,0 +1,19 @@
+{
+  "compilerOptions": {
+    "skipLibCheck": true,
+    "target": "ES2022",
+    "isolatedModules": true,
+    "moduleDetection": "force",
+
+    // linting
```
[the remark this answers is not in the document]
FranjoMindek: Also could be a bit more specific e.g. "type safety", "strictness"....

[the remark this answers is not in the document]
sodic: I'm removing the comment.

infomiho:
Reminder to update the `docs` alongside `versioned_docs`

examples/waspello/package.json
FranjoMindek: Btw, this won't work for people if they don't hold the project at same folder path as you.
Should we in README mention for them to run `wasp ts-setup` before  trying to run the project?

sodic: We should, nice catch

PR #2787, opened 2025-05-21
Remove Node 18 polyfill

cprecioso:
This polyfill was only needed for Node 18. Our minimum target is now Node 20.

```sh
$ mise x node@20 -- node -p '"crypto" in globalThis'
true
```


infomiho (approved):
Nice catch

sodic (approved):
Noice!

PR #2792, opened 2025-05-22
Remove prettier plugins from auto-js

cprecioso:
The `auto-js` plugin doesn't really need any plugins added to prettier to work correctly (it's just regular JS/TS code), and it was [tripping up our Cloudflare build](https://dash.cloudflare.com/75c37602a6b75c790961219411cb1b92/pages/view/wasp-docs-on-main/bc567ab6-e706-43a3-a25a-d8a0b022bcf6) (we don't have npm workspaces so the plugins, which are declared at the root level, weren't being installed).

### DIFF
--- a/web/src/remark/auto-js-code.ts
+++ b/web/src/remark/auto-js-code.ts
@@ -174,16 +174,27 @@ function convertToJs(code: string, { isJsx }: { isJsx: boolean }) {
   return blankSourceFile(sourceFile);
 }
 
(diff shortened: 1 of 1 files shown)


cprecioso:
@infomiho requested a build and works fine https://dash.cloudflare.com/75c37602a6b75c790961219411cb1b92/pages/view/wasp-docs/b0bd056a-a0c0-4be7-9873-7c7b73763e5f

infomiho (approved):

PR #2799, opened 2025-05-27
Stop lowercasing OAuth provider IDs

infomiho:
This PR addresses the issue of Wasp's OAuth implementation not being in line with the OpenID spec where the `sub` property (what we use as user ID) is case sensitive. In practice, most OAuth providers use either all lowercase or all uppercase IDs, but it's not guaranteed.


infomiho:
Need to branch from `release` instead of `main`.

PR #2800, opened 2025-05-27
Stop lowercasing IDs received by OAuth providers

infomiho:
This PR addresses the issue of Wasp's OAuth implementation not being in line with the OpenID spec where the sub property (what we use as user ID) is case sensitive. In practice, most OAuth providers use either all lowercase or all uppercase IDs, but it's not guaranteed.


waspc/data/Generator/templates/sdk/wasp/auth/utils.ts
```diff
@@ -81,7 +81,18 @@ export type ProviderId = {
...
+
+// PRIVATE API
+function normalizeProviderUserId(providerName: ProviderName, providerUserId: string): string {
+  switch (providerName) {
+    case 'email':
+    case 'username':
+      return providerUserId.toLowerCase()
+    default:
+      return providerUserId
   }
```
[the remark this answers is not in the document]
infomiho: On a second thought, I think we can actually include the exhaustiveness check as users won't be able to call this function with a `string` anyways, they'd have to use a type assertion e.g. `"spotify" as ProviderName`. 

We'll get type errors in Typescript if we forget to handle some built-in provider explicitly and the `default` case will handle all the user defined providers.

waspc/ChangeLog.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.16.5
+
+### 🐞 Bug fixes
+
+- IDs received by OAuth providers are no longer lowercased to be inline with OAuth specification. ([#2800](https://github.com/wasp-lang/wasp/pull/2800))
```
cprecioso, line 7:
```suggestion
- IDs received by OAuth providers are no longer lowercased, to be in line with OAuth specification. ([#2800](https://github.com/wasp-lang/wasp/pull/2800))
```

waspc/data/Generator/templates/sdk/wasp/auth/utils.ts
```diff
@@ -81,7 +81,18 @@ export type ProviderId = {
...
+    providerUserId: normalizeProviderUserId(providerName, providerUserId)
+  }
+}
+
+// PRIVATE API
+function normalizeProviderUserId(providerName: ProviderName, providerUserId: string): string {
+  switch (providerName) {
+    case 'email':
+    case 'username':
+      return providerUserId.toLowerCase()
```
cprecioso, line 93: I won't block on this, but it feels for me that it would make more sense to lowercase this inside the email/username code, not here. And then this code should not normalize at all, the provider is the one in charge.

cprecioso (approved):

cprecioso:
Are we missing the migration guide?

PR #2806, opened 2025-05-29
Fixes for CI

cprecioso:
Three changes that were not caught earlier:
- Add extra caching key to not use glibc builds in musl. Haskell tries to execute some binaries and it fails.
- Correctly format the caching key. In GH Actions, nested `${{ ${{ }} }}` are not permitted (but it doesn't warn if you do it 🤦‍♂️) so you need to use the `${{ format() }}` function
- The `darwin-universal` build had not had the names of the artifacts changed. Now it does.

Check a full complete run here: https://github.com/wasp-lang/wasp/actions/runs/15342687609

### DIFF
--- a/.github/actions/setup-haskell/action.yaml
+++ b/.github/actions/setup-haskell/action.yaml
@@ -23,6 +23,13 @@ inputs:
     required: false
     default: "waspc"
 
+  extra-cache-key:
+    description: |
+      An extra key to use for the cache. Mainly used to keep glibc and musl
+      caches separate, as they are not compatible.
+    required: false
+    default: "default"
+
 runs:
   using: composite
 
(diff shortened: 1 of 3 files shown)


.github/actions/setup-haskell/action.yaml
```diff
@@ -23,6 +23,13 @@ inputs:
     required: false
     default: "waspc"
 
+  extra-cache-key:
```
Martinsos, line 26: Makes sense! Me with naming again, but what confused me a bit is that this is not the key though, this is only part of the key. So maybe let's rename it to `extra-cache-key-segment` or `unique-cache-key-segment` or something like that?

Martinsos (approved):
LGTM! Please check that one comment before merging though.

cprecioso:
@Martinsos I also pinned the Alpine version like you wanted, because, indeed, it had updated and broken

Martinsos:
> @Martinsos I also pinned the Alpine version like you wanted, because, indeed, it had updated and broken

Oh wow, already! Ok good, now we will likely have an issue at some point that there is no alpine3.21 with the version of node we are using, but that is easy to detect and fix, and won't surprise us out of nowhere but only happen once we change the version of the node, so that is a better failure scenario, one we can manage quite normally.

Would it make sense adding a comment on top of that line, explaining why we used fixed alpine version? Now it sounds obvious, but it obviously wasn't obvious to as a little while ago, and I wouldn't want somebody having to guess why it was pinned, and also potentially revmoing the specific version, not being aware of why it was important. Do as you think is best, just an idea.

cprecioso:
@Martinsos done here https://github.com/wasp-lang/wasp/pull/2816

PR #2807, opened 2025-05-29
Move `wasp-app-runner` to `wasp-lang/runner-action`

### LINKED ISSUE #2840: Make e2e `wasp-app-runner` reusable
Currently the `wasp-app-runner` is located in the root of the `wasp` repository.

This makes it hard to reuse outside of `wasp` repository, and makes it hard for wasp users to use it as well.

The base case scenario would be to make it a `npm` package, so it's easily available both to users and us (e.g. in CLI or locally). But currently we holding out of `npm`.

Second best scenario is to make it a Github action. But to do that we must move it to a separate repo, as actions can only be defined in the root of the repository. The additional bonus is that the `wasp-app-runner` is easier to clone now before we sort out the `npm` situation.

FranjoMindek:
Fixed #2840

As we moved the `wasp-app-runner` to https://github.com/wasp-lang/runner-action, this PR refactors the `wasp` repository to use the new `wasp-app-runner` instead of the one in the root of this repository.

Since `wasp-app-runner` is not a `npm` package, we have to clone it, build it and install it globally by ourselves. This is done by the new `[install_wasp-app-runner.sh](https://github.com/wasp-lang/wasp/pull/2807/files#diff-5f349a8588b6965e6f2532648b14f7d3783e9f1d29ffcd842a23e998feef174b)` script inside of `tools`.

It also makes sure the currently installed version is the latest.


waspc/run
FranjoMindek, line 1: Where to place these functions inside of the `run` script?

waspc/run
```diff
@@ -30,10 +30,43 @@ TEST_CLI_CMD="cabal test cli-test"
...
+  fi
+}
+
+function install_wasp_app_runner() {
+  local TMP_DIR="$PROJECT_ROOT/.tmp"
+  local REPO_DIR="$TMP_DIR/runner-action"
+
+  echo_bold "Cloning wasp-app-runner to $REPO_DIR..."
+  mkdir -p "$TMP_DIR"
+  git clone --depth 1 https://github.com/wasp-lang/runner-action.git "$REPO_DIR"
```
FranjoMindek: `--depth 1` - Shallow clone, only latest commit.

waspc/run
```diff
@@ -30,10 +30,43 @@ TEST_CLI_CMD="cabal test cli-test"
...
+
+  echo_bold "Building and packing Wasp App Runner..."
+  cd "$REPO_DIR" || exit 1
+  npm ci --silent
```
FranjoMindek: Which of these steps do we log?
Note that `npm pack` must be silent so it returns only the path.

waspc/run
```diff
@@ -30,10 +30,43 @@ TEST_CLI_CMD="cabal test cli-test"
...
+function install_wasp_app_runner() {
+  local TMP_DIR="$PROJECT_ROOT/.tmp"
+  local REPO_DIR="$TMP_DIR/runner-action"
+
+  echo_bold "Cloning wasp-app-runner to $REPO_DIR..."
+  mkdir -p "$TMP_DIR"
+  git clone --depth 1 https://github.com/wasp-lang/runner-action.git "$REPO_DIR"
+
+  echo_bold "Building and packing Wasp App Runner..."
+  cd "$REPO_DIR" || exit 1
```
FranjoMindek: Copilot wanted to me to some stack shenanigans but I don't think it's necessary.

waspc/run
```diff
@@ -30,10 +30,43 @@ TEST_CLI_CMD="cabal test cli-test"
...
 TEST_E2E_ACCEPT_ALL_CMD="$TEST_E2E_DELETE_GOLDENS && $TEST_E2E_CMD"
+# TODO: figure this out
 TEST_INSTALL_WASP_APP_RUNNER="cd $PROJECT_ROOT/../wasp-app-runner && npm install && npm run install:global"
-TEST_HEADLESS_CMD="$TEST_INSTALL_WASP_APP_RUNNER && cd $PROJECT_ROOT/examples/todoApp && npm install && DEBUG=pw:webserver npx playwright test --config headless-tests"
+TEST_HEADLESS_CMD="ensure_wasp_app_runner_installed && cd $PROJECT_ROOT/examples/todoApp && npm install && DEBUG=pw:webserver npx playwright test --config headless-tests"
 TEST_CMD="cabal test && echo 'Running headless tests' && $TEST_HEADLESS_CMD && echo 'ALL TESTS PASSED' || { echo 'SOME TESTS FAILED'; exit 1; }"
 
+function ensure_wasp_app_runner_installed() {
+  if ! command -v run-wasp-app > /dev/null 2>&1; then
+    echo -e "${YELLOW}run-wasp-app is not installed. Installing...${RESET}"
```
infomiho: How do we update the runner installation? We could install it on each run or a dedicated install command could come in handy.

FranjoMindek: Currently we would need to `npm uninstall -g wasp-app-runner` before running it again.
But that is a good issue you're pointing out.

Should we include proper version check in `wasp-app-runner`? Too much investment?
Then the alternative of re-installing it each run is possible. But I feel weird about it.

waspc/run
sodic: What stack shennanigans?

waspc/run
sodic: Hm, not sure what you mean here. What should we log and why?

waspc/run
```diff
@@ -30,10 +30,43 @@ TEST_CLI_CMD="cabal test cli-test"
 TEST_E2E_CMD="cabal test e2e-test"
 TEST_E2E_DELETE_GOLDENS="rm -rf $PROJECT_ROOT/e2e-test/test-outputs/*-golden"
 TEST_E2E_ACCEPT_ALL_CMD="$TEST_E2E_DELETE_GOLDENS && $TEST_E2E_CMD"
+# TODO: figure this out
```
sodic: What are we figuring out? :)

waspc/run
```diff
@@ -30,10 +30,43 @@ TEST_CLI_CMD="cabal test cli-test"
...
 TEST_INSTALL_WASP_APP_RUNNER="cd $PROJECT_ROOT/../wasp-app-runner && npm install && npm run install:global"
```
sodic: I this still used somewhere?

waspc/run
FranjoMindek: `npm ci` will log the installation of packages, `--silent` flag stops that.
How much info do we want to show to the user as to not to overload the terminal?

waspc/run
FranjoMindek: I suppose it doesn't matter.
But it did some push and pop stuff to save the location of dir before `cd`? But that shouldn't matter here in function?

waspc/run
FranjoMindek: Ah I committed this before I made the function, because I didn't yet know how I will implement `TEST_INSTALL_WASP_APP_RUNNER` because I still wasn't sure what `wasp-app-runner` will look like after changes.

I will remove TODO with `TEST_INSTALL_WASP_APP_RUNNER`

waspc/run
FranjoMindek: Removed together with `# TODO`,

waspc/run
FranjoMindek: I solved it and moved all logic to `tools/install_wasp-app-runner.sh`.
We only need to merge the `--version` flag change to `wasp-app-runner` first.

waspc/run
FranjoMindek: Moved the script logic outside of `run`

waspc/run
sodic: I think this is OK. After all, "the user" is only us for now :)

waspc/run
```diff
@@ -30,8 +30,7 @@ TEST_CLI_CMD="cabal test cli-test"
 TEST_E2E_CMD="cabal test e2e-test"
 TEST_E2E_DELETE_GOLDENS="rm -rf $PROJECT_ROOT/e2e-test/test-outputs/*-golden"
 TEST_E2E_ACCEPT_ALL_CMD="$TEST_E2E_DELETE_GOLDENS && $TEST_E2E_CMD"
-TEST_INSTALL_WASP_APP_RUNNER="cd $PROJECT_ROOT/../wasp-app-runner && npm install && npm run install:global"
-TEST_HEADLESS_CMD="$TEST_INSTALL_WASP_APP_RUNNER && cd $PROJECT_ROOT/examples/todoApp && npm install && DEBUG=pw:webserver npx playwright test --config headless-tests"
+TEST_HEADLESS_CMD="${SCRIPT_DIR}/tools/install_wasp-app-runner.sh && cd $PROJECT_ROOT/examples/todoApp && npm install && DEBUG=pw:webserver npx playwright test --config headless-tests"
```
sodic:
```suggestion
TEST_HEADLESS_CMD="${SCRIPT_DIR}/tools/install_wasp_app_runner.sh && cd $PROJECT_ROOT/examples/todoApp && npm install && DEBUG=pw:webserver npx playwright test --config headless-tests"
```

waspc/tools/install_wasp-app-runner.sh
```diff
@@ -0,0 +1,125 @@
...
+  fi
+  exit 0
+}
+
+is_wasp_app_runner_installed() {
+  command -v run-wasp-app > /dev/null 2>&1
+}
+
+get_installed_wasp_app_runner_version() {
+  version=$(run-wasp-app --version 2> /dev/null | tr -d '[:space:]')
```
sodic: We could extract the executable name into a constant

waspc/tools/install_wasp-app-runner.sh
```diff
@@ -0,0 +1,125 @@
...
+SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+PROJECT_ROOT=$(cd "$SCRIPT_DIR/.." && pwd)
+
+RESET="\033[0m"
+YELLOW="\033[33m"
+GREEN="\033[32m"
+RED="\033[31m"
+
+main_flow() {
+  echo -e "Checking wasp-app-runner status..."
```
sodic: I don't think we need `echo -e` anywhere.

waspc/tools/install_wasp-app-runner.sh
```diff
@@ -0,0 +1,125 @@
...
+  echo -e "Checking for wasp-app-runner updates..."
+
+  local INSTALLED_VERSION
+  INSTALLED_VERSION=$(get_installed_wasp_app_runner_version)
+
+  local LATEST_VERSION
+  LATEST_VERSION=$(get_latest_wasp_app_runner_version_from_github)
+
+  if [[ "$INSTALLED_VERSION" == "$LATEST_VERSION" ]]; then
+    echo -e "${GREEN}wasp-app-runner (version $INSTALLED_VERSION) is up to date.${RESET}"
```
sodic: Perhaps abstract the colored echoing into a function.

waspc/tools/install_wasp-app-runner.sh
```diff
@@ -0,0 +1,125 @@
...
+      echo -e "${RED}Failed to update wasp-app-runner. Please check errors above.${RESET}"
+      exit 1
+    fi
+    echo -e "${GREEN}wasp-app-runner has been updated to version ${LATEST_VERSION}.${RESET}"
+  fi
+  exit 0
+}
+
+is_wasp_app_runner_installed() {
+  command -v run-wasp-app > /dev/null 2>&1
```
sodic: Any way we could remove the duplication and refer to the executable name as defined in `package.json`?

waspc/tools/install_wasp-app-runner.sh
```diff
@@ -0,0 +1,125 @@
...
+
+  if [[ "$INSTALLED_VERSION" == "$LATEST_VERSION" ]]; then
+    echo -e "${GREEN}wasp-app-runner (version $INSTALLED_VERSION) is up to date.${RESET}"
+  else
+    echo -e "${YELLOW}A new version ($LATEST_VERSION) of wasp-app-runner is available (installed: $INSTALLED_VERSION).${RESET}"
+    if ! install_latest_wasp_app_runner; then
+      echo -e "${RED}Failed to update wasp-app-runner. Please check errors above.${RESET}"
+      exit 1
+    fi
+    echo -e "${GREEN}wasp-app-runner has been updated to version ${LATEST_VERSION}.${RESET}"
```
sodic: As we discussed in person, there are perhaps too many `echo`s.

waspc/tools/install_wasp-app-runner.sh
```diff
@@ -0,0 +1,125 @@
...
+  else
+    echo "unknown"
+  fi
+}
+
+install_latest_wasp_app_runner() {
+  echo -e "Installing the latest version of wasp-app-runner..."
+
+  local TMP_INSTALL_DIR
+  TMP_INSTALL_DIR=$(mktemp -d "${PROJECT_ROOT}/tmp.XXXXXXXXXX")
```
sodic: Perhaps a better place to put this is `/tmp`. Or not, not sure what the convention is.

Also, we should come up with a better name for the temp directory, something like `wasp-app-runner.XXXX`

waspc/tools/install_wasp-app-runner.sh
```diff
@@ -0,0 +1,125 @@
...
+
+  if [[ -z "$TMP_INSTALL_DIR" || ! -d "$TMP_INSTALL_DIR" ]]; then
```
sodic: Perhaps it's better to check the commands exit status. I think you may even be able to do this in a single line.

waspc/tools/install_wasp-app-runner.sh
```diff
@@ -0,0 +1,125 @@
...
+    echo -e "${RED}Failed to create temporary directory for runner installation.${RESET}"
+    return 1
+  fi
+
+  # Ensure cleanup of the temporary directory on exit, error, or interrupt
+  trap 'rm -rf "$TMP_INSTALL_DIR" >/dev/null 2>&1' EXIT TERM INT
```
sodic: We need double quotes for the interpolation to work.

waspc/tools/install_wasp-app-runner.sh
```diff
@@ -0,0 +1,125 @@
...
+
+  local REPO_DIR="$TMP_INSTALL_DIR/runner-action"
```
sodic: Can we kick out the `runner-action`?

waspc/tools/install_wasp-app-runner.sh
```diff
@@ -0,0 +1,125 @@
...
+  echo -e "Building Wasp App Runner from $REPO_DIR..."
+  if ! npm ci --silent; then
+    echo -e "${RED}npm ci failed in $REPO_DIR${RESET}"
+    return 1
+  fi
+
+  echo -e "Packing Wasp App Runner from $REPO_DIR..."
+  local TAR_FILE
+  TAR_FILE=$(npm pack --silent)
+  if [[ -z "$TAR_FILE" || ! -f "$TAR_FILE" ]]; then
```
sodic: Let's do the same thing we end up doing above.

waspc/tools/install_wasp-app-runner.sh
```diff
@@ -0,0 +1,125 @@
...
+    echo -e "${RED}npm pack failed or produced no tarball in $REPO_DIR${RESET}"
+    return 1
+  fi
+  echo -e "Packed to $TAR_FILE."
```
sodic: We can probably kick out most if not all of the success messages. Even if it said nothing on success, it would be alright.

waspc/tools/install_wasp-app-runner.sh
```diff
@@ -0,0 +1,125 @@
...
+
+  echo -e "Installing Wasp App Runner globally from $REPO_DIR/$TAR_FILE..."
+  if ! npm install -g "$REPO_DIR/$TAR_FILE"; then
+    echo -e "${RED}npm install -g failed for $REPO_DIR/$TAR_FILE${RESET}"
+    return 1
+  fi
+
+  trap - EXIT TERM INT # Clear the trap as we are done
+  rm -rf "$TMP_INSTALL_DIR" > /dev/null 2>&1
```
sodic: We can just leave the trap. I'm not sure whether that's the best practice, please investigate :)

waspc/tools/install_wasp-app-runner.sh
```diff
@@ -0,0 +1,125 @@
...
+  if [[ -z "$version" ]]; then
+    echo "unknown"
+  else
+    echo "$version"
+  fi
+}
+
+get_latest_wasp_app_runner_version_from_github() {
+  local latest_version
+  latest_version=$(curl -sL "https://raw.githubusercontent.com/wasp-lang/runner-action/main/package.json" | grep '"version":' | sed -n 's/.*"version": *"\([^"]*\)".*/\1/p' 2> /dev/null)
```
sodic: Try using the `-E` flag in `sed`.

waspc/tools/install_wasp-app-runner.sh
FranjoMindek: That is maybe too much complexity to find the right `pacakge.json`?
We can do it, don't know if it's worth it.

And parsing will also be tricky.
We need to find the `bin` object and then what?
Take the first key as executable?

waspc/tools/install_wasp-app-runner.sh
FranjoMindek: I think the benefit of `/tmp` would be that it would get auto deleted after the default tmp timeout.
Benefit of having it locally is that we keep it local.

For now I kept it local.

waspc/tools/install_wasp-app-runner.sh
sodic: I'm not sure whether this is the best practice though. @Martinsos what do you think?

waspc/tools/install_wasp_app_runner.sh
```diff
@@ -0,0 +1,118 @@
...
+
+install_latest_wasp_app_runner() {
+  local tmp_installation_dir
+  if ! tmp_installation_dir=$(mktemp -d "${PROJECT_ROOT}/wasp-app-runner.XXXXXXXXXX"); then
+    echo_error "Failed to create temporary directory for runner installation."
+    return 1
+  fi
+
+  # Ensure cleanup of the temporary directory on exit, error, or interrupt
+  # shellcheck disable=SC2064
```
sodic: Can we explain this in the comment?

waspc/tools/install_wasp_app_runner.sh
sodic, line 1: The shell script looks alright to me, but I'd like to ask @Martinsos to take a look too - he wrote most of our bash scripts and will be a better reviewer here.

The context for this change is in the PR description, but I think you can ignore it if you want. All I'm wondering is:
- Assuming we have to download Wasp app runner by cloning it through GitHub, what do you think of this script?

sodic (commented):
I tagged @Martinsos to take a look at the script.

I know we discussed this at some point, but I forgot: why aren't we publishing this to NPM? @FranjoMindek can you add the reasoning to the PR or issue description. The last thing I see is [this](https://github.com/wasp-lang/wasp/issues/2840#issuecomment-3008733546) which makes it seem like we wanted to publish it.

FranjoMindek:
The initial idea for `npm` package started here:
https://discord.com/channels/686873244791210014/1351908377533354025/1374357332401852527

After the text discussion we had an on-call discussion where we decided how to continue.
IIRC we did think that npm packages is a great solution, but I think that @cprecioso pointed out we didn't have our npm packages situation ready yet. We weren't sure under which name we would publish this package. Or something else along these lines.

waspc/tools/install_wasp_app_runner.sh
```diff
@@ -0,0 +1,120 @@
...
+
+  local installed_wasp_app_runner_version
+  installed_wasp_app_runner_version=$(get_installed_wasp_app_runner_version)
+
+  local latest_wasp_app_runner_version
+  latest_wasp_app_runner_version=$(get_latest_wasp_app_runner_version_from_github)
+
+  if ! [[ "$installed_wasp_app_runner_version" == "$latest_wasp_app_runner_version" ]]; then
+    if ! install_latest_wasp_app_runner; then
+      echo_error "Failed to update wasp-app-runner. Please check errors above."
```
Martinsos, line 38: How can we get to this state? If installation failed, we already exited. If we succeeded, how can we end up with not the latest version of app runner, when we ran the command that installs the latest version?

waspc/tools/install_wasp_app_runner.sh
```diff
@@ -0,0 +1,120 @@
...
+
+main_flow() {
+  if ! is_wasp_app_runner_installed; then
+    echo_warning "wasp-app-runner is not installed."
+    if ! install_latest_wasp_app_runner; then
+      echo_error "Failed to install wasp-app-runner. Please check errors above."
+      exit 1
+    fi
+    echo_success "wasp-app-runner is now installed."
+    exit 0
```
Martinsos: A bit hard to follow, took me a bit to realize install_latest_wasp_app_runner actually does the whole installation. Probalby better to to `install_latest_wasp_app_runner` in its own line, and then comes `if` that checks the exit code. And also this if, might be nicer if it is if-else than if with exit and then exit after it.

waspc/tools/install_wasp_app_runner.sh
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env bash
+
+# This script ensures wasp-app-runner is installed and up-to-date.
+
+SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+PROJECT_ROOT=$(cd "$SCRIPT_DIR/.." && pwd)
+
+WASP_APP_RUNNER_COMMAND="run-wasp-app"
+
+main_flow() {
```
Martinsos: main flow sounds very meta name :D. How about we name it by what it actually does? or at least just `main`?

waspc/tools/install_wasp_app_runner.sh
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env bash
+
+# This script ensures wasp-app-runner is installed and up-to-date.
+
+SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+PROJECT_ROOT=$(cd "$SCRIPT_DIR/.." && pwd)
```
Martinsos: I get what is SCRIPT_DIR, but what is PROJECT_ROOT?

waspc/tools/install_wasp_app_runner.sh
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env bash
```
Martinsos, line 1: If you want it to run everywhere, better go for working for sh.

#!/bin/sh

Ubuntu doesn't come with bash by default any more, for example.

Thats how we do it for installer.

waspc/tools/install_wasp_app_runner.sh
```diff
@@ -0,0 +1,120 @@
...
+
+is_wasp_app_runner_installed() {
+  command -v "$WASP_APP_RUNNER_COMMAND" > /dev/null 2>&1
+}
+
+get_installed_wasp_app_runner_version() {
+  local installed_version
+  installed_version=$("$WASP_APP_RUNNER_COMMAND" --version 2> /dev/null | tr -d '[:space:]')
+  if [[ -z "$installed_version" ]]; then
+    echo "unknown"
```
Martinsos, line 52: I am curious, when can this happen? And if it does happen, are we i those situations ok with returning unknown?

waspc/tools/install_wasp_app_runner.sh
```diff
@@ -0,0 +1,120 @@
...
+    exit 0
+  fi
+
+  local installed_wasp_app_runner_version
+  installed_wasp_app_runner_version=$(get_installed_wasp_app_runner_version)
+
+  local latest_wasp_app_runner_version
+  latest_wasp_app_runner_version=$(get_latest_wasp_app_runner_version_from_github)
+
+  if ! [[ "$installed_wasp_app_runner_version" == "$latest_wasp_app_runner_version" ]]; then
```
Martinsos: Fun fact: if both return "unknown", you will think we have the latest version! Not great.

waspc/tools/install_wasp_app_runner.sh
```diff
@@ -0,0 +1,120 @@
...
+  if [[ -z "$latest_version" ]]; then
+    echo "unknown"
+  else
+    echo "$latest_version"
+  fi
+}
+
+install_latest_wasp_app_runner() {
+  local tmp_installation_dir
+  if ! tmp_installation_dir=$(mktemp -d "${PROJECT_ROOT}/wasp-app-runner.XXXXXXXXXX"); then
```
Martinsos: What is with all the XXXXXX? Is that some magic property of mktemp?

waspc/tools/install_wasp_app_runner.sh
```diff
@@ -0,0 +1,120 @@
...
+    echo_error "Failed to create temporary directory for runner installation."
+    return 1
+  fi
+
+  # Ensure cleanup of the temporary directory on exit, error, or interruption.
+  # shellcheck disable=SC2064 # Shellcheck wants us to use single quotes here.
+  # We intentionally use double quotes to capture the variable's current value,
+  # instead of deferring the expansion until the trap runs (which single quotes would do).
+  trap "rm -rf $tmp_installation_dir >/dev/null 2>&1" EXIT TERM INT
```
Martinsos: It's always scary to see rm -rf like this.
Is there any chance this can go wrong? That $tmp_installation_dir var ends up being empty or `.` or `*` or something like that? Any way we can protect against this?

waspc/tools/install_wasp_app_runner.sh
```diff
@@ -0,0 +1,120 @@
...
+
+  if ! git clone --depth 1 --quiet https://github.com/wasp-lang/runner-action.git "$tmp_installation_dir"; then
```
Martinsos, line 82: Duplicatino from above, this is second time you have repo url. Let's extract it to env var and we can even put it at the top of the file.

waspc/tools/install_wasp_app_runner.sh
```diff
@@ -0,0 +1,120 @@
...
+    echo_error "Failed to clone wasp-app-runner repository into $tmp_installation_dir!"
+    return 1
+  fi
+
+  cd "$tmp_installation_dir" || {
+    echo_error "Failed to cd into $tmp_installation_dir"
```
Martinsos, line 88: Oh wild, can this happen? I probably wouldn't think of checking this.
I see you check every single step below also. That is interesting. Is this usual practice? What would happen if you don't do it?

waspc/tools/install_wasp_app_runner.sh
```diff
@@ -0,0 +1,120 @@
...
+    return 1
+  fi
+
+  local tarball_filename
+  if ! tarball_filename=$(npm pack --silent); then
+    echo_error "npm pack failed in $tmp_installation_dir"
+    return 1
+  fi
+
+  if ! npm install -g --silent "$tmp_installation_dir/$tarball_filename"; then
```
Martinsos, line 103: I don't know if it makes sense to silence evertyhing, from my epxerience that can hide too much. WHy not just let most of the things output their stuff?

waspc/tools/install_wasp_app_runner.sh
FranjoMindek: If we have older version of `wasp-app-runner` installed we end up here.
Imagine we update the `wasp-app-runner` since last time running the script.

First we check if `wasp-app-runner` is installed, if not we install and exit.
If installed we check for the latest version, if not we update.

Martinsos: Oh my bad! I seem to have forgot what happen in the lines above. Ok, makes sense. Maybe it would be more obvious if there `else` following up on that `if` hm, not sure.

Although, why do we have to be so "smart"? Why don' we just run installation and that is it? We could even run if it is already installed. They did run installation script at the end. Not saying we should do what I am suggesting here, just wondering on the reasoning.

waspc/tools/install_wasp_app_runner.sh
FranjoMindek: It's `waspc` dir but since we only use it but to make temp dirs, we can just do it directly at script location instead.

waspc/tools/install_wasp_app_runner.sh
FranjoMindek: That would require some bigger changes, e.g. `local`, `[[ ]]`...
I'll look into it.

waspc/tools/install_wasp_app_runner.sh
FranjoMindek: It shouldn't be possible, unless somebody has a version which before we added `--version` flag to the `wasp-app-runner`. Which is possible for current wasp team members.

If we return `unknown` here, it will "update" the current version aka reinstall.

waspc/tools/install_wasp_app_runner.sh
FranjoMindek: Yes, but shouldn't ever happen with `get_latest_wasp_app_runner_version_from_github`.
I think it might be better to exit if `get_latest_wasp_app_runner_version_from_github` fails.

waspc/tools/install_wasp_app_runner.sh
FranjoMindek: That many random characters are added to the folder as the suffix. It's like a template.
Without it it has some built in defaults. This was more to make sure it's the same on all OS.

`mktemp` makes sure to return an unique name anyhow.

waspc/tools/install_wasp_app_runner.sh
FranjoMindek: Not sure besides moving the logic to a function and doing the checks manually?
From what I see we usually don't check anything and at 1 instance only check for empty (`-n`).
I've added the `-n` check for now, but I'll see if there is anything more through.

waspc/tools/install_wasp_app_runner.sh
FranjoMindek: Not the same:
`https://raw.githubusercontent.com/wasp-lang/runner-action/main/package.json`
vs
`https://github.com/wasp-lang/runner-action.git`

waspc/tools/install_wasp_app_runner.sh
FranjoMindek: edit: I described what would happen if it didn't `cd`, but it also didn't exit because of it. Which is wrong because it will exit. We can reduce number of checks.

What would happen:
`npm` would fail because no `package.json`, `Failed to install wasp-app-runner. Please check errors above.` prints, exits with 1?
```bash
echo_error "Failed to update wasp-app-runner. Please check errors above."
exit 1
```
The bonus here is that we know where the error happened and why exactly, rater than debugging.
Don't know if its the usual practice. I don't really do `bash`.

waspc/tools/install_wasp_app_runner.sh
FranjoMindek: Pack must be `--silent` so it returns only the tarball filename it created.
Rest don't have to. But echo messages already say everything. I didn't want to pollute the console.

waspc/tools/install_wasp_app_runner.sh
FranjoMindek: Ah an artifact from when I wrote what flows I wanted to have in pseudocode.
Renamed to `main`

waspc/tools/install_wasp_app_runner.sh
FranjoMindek: I just wanted to avoid unnecessary clone, install, build, install. Nothing more.
It takes a few seconds to do that every time.

waspc/tools/install_wasp_app_runner.sh
sodic: FYI, checking the exit code separately is considered an antipattern in shell scripting. I think shellcheck even warns about it.

The argument is similar to the one using braces on one-line if statements.

Doing so decouples the command from its exit code. Adding a command in between (even a simple echo for debugging) or shuffling the commands breaks the `if` logic, so it's always advised to avoid relying on implicitly set variables if possible.

There's more to it, but that's the gist: https://stackoverflow.com/a/36313272

waspc/tools/install_wasp_app_runner.sh
sodic: > Ubuntu doesn't come with bash by default any more, for example.

What? :D

> If you want it to run everywhere, better go for working for sh.

I noticed this, but since it's our internal tool and the script uses some bash constructs (like Franjo mentioned), I think it's not a big deal if we use bash.

After all, this script already assumes a lot more (npm, typescript, wasp...). I'd say that Bash is the least of its worries :)

waspc/tools/install_wasp_app_runner.sh
sodic: >Oh wild, can this happen? I probably wouldn't think of checking this.

I thought the same thing (one of the reasons I tagged you).
It's not usual to check every step, and I vote to skip it here. The `cd` command says enough on its own:

image

For the `--silent` commands, I'd keep it (since they don't say anything on their own).

This is my rule of thumb:
- Check what the command outputs when it fails and decide whether you'd like to see something extra on top of that. If not, just let it fail. Shell commands are mostly designed to be used like this - that's why they don't say anything when they succeed and why they say enough when they fail.

You can apply this rule to the entire file and probably remove some other checks too.

waspc/tools/install_wasp_app_runner.sh
sodic: > But echo messages already say everything. I didn't want to pollute the console.

Like I said above, it might make sense to flip this logic and use the available reporting instead of making our own. It's a case-by-case basis thing (some commands might still be to verbose), so decide yourself :)

waspc/tools/install_wasp_app_runner.sh
FranjoMindek: I did end up doing explicit else in the end:
```bash
if ! is_wasp_app_runner_installed; then
  echo_warning "wasp-app-runner is not installed."

  if install_latest_wasp_app_runner; then
    echo_success "wasp-app-runner is now installed."
    exit 0
  else
    echo_error "Failed to install wasp-app-runner. Please check errors above."
    exit 1
  fi
  ```

File: waspc/tools/install_wasp_app_runner.sh (hunk 32 above)
Comment (FranjoMindek) [reply waspc/tools/install_wasp_app_runner.sh:88]: Added edit to top comment, somehow I conjured a situation where it didn't `cd` but it also didn't exit. But yes.

Comment (FranjoMindek) [reply waspc/tools/install_wasp_app_runner.sh:88]: Ah this makes sense why we have it:
image

We can still remove the echo though.

File: waspc/tools/install_wasp-app-runner.sh (hunk 13 above)
Comment (Martinsos) [reply waspc/tools/install_wasp-app-runner.sh]: To add to that -> I would go for /tmp also. It will get cleaned up and that is good. I wouldn't be awfuly excited about installation artifacts littering my local dir (nor have I ever seen that happen, although I am sure somebody does it).

File: waspc/tools/install_wasp-app-runner.sh (hunk 14 above)
Comment (Martinsos) [reply waspc/tools/install_wasp-app-runner.sh]: Yup I agree it makes more sense to check the exit status, if it provides enough info.

File: waspc/tools/install_wasp_app_runner.sh (hunk 22 above)
Comment (Martinsos) [reply waspc/tools/install_wasp_app_runner.sh:38]: It's just a lot of logic here that can have a mistake now or in the future when it is maintained for something that when I think about it is possibly not needed.

But ok, if you think it is worth it, let's leave it.

raw

File: waspc/tools/install_wasp_app_runner.sh (hunk 23 above)
Comment (Martinsos) [reply waspc/tools/install_wasp_app_runner.sh]: Hm I get it @sodic , that makes sense, relying on reading the variable for exit code that assumes no other command executed in the meantime, that is indeed fragile.

I was coming from the other side, where in other (normal) languages, it is discouraged to run an actual logic in the condition of IF because it is confusing / easy to miss that stuff is happening there (as it was confusing for me here).

One could probably get both by doing something like
```
(exit_code=$(my_function; echo $?))
```
or something similar.

This way you can disconnect the two, but you still have it in var that you can check in IF.

But it seems it is normal/idiomatic in bash to execute functions/commands in IF condition, I just didn't know it, so let's leave it as it is then! Tnx I learned something.

raw

File: waspc/tools/install_wasp_app_runner.sh (hunk 26 above)
Comment (Martinsos) [reply waspc/tools/install_wasp_app_runner.sh:1]: Yeah that is context I was missing -> I was looking at it from the eyes of "another installer script -> we want installer scripts to work on as many machines as possible".
If indeed it is mostly used by us and we are ok with it being in bash, it can stay in bash, no worries.

raw

File: waspc/tools/install_wasp_app_runner.sh (hunk 28 above)
Comment (Martinsos) [reply waspc/tools/install_wasp_app_runner.sh]: If it will never happen, whey are we returning it :D?
Yeah let's change something, because right now it looks in code like it could be problematic, and as a reader I would rather not have to have to figure out all the possible states in order to determine that issue won't happen.

File: waspc/tools/install_wasp_app_runner.sh (hunk 29 above)
Comment (Martinsos) [reply waspc/tools/install_wasp_app_runner.sh]: Ok cool, so it is magic mktep thing, nice!

File: waspc/tools/install_wasp_app_runner.sh (hunk 30 above)
Comment (Martinsos) [reply waspc/tools/install_wasp_app_runner.sh]: Ok, empty might be enough.

File: waspc/tools/install_wasp_app_runner.sh (hunk 31 above)
Comment (Martinsos) [reply waspc/tools/install_wasp_app_runner.sh:82]: Of course two lines are not exactly the same, but duplication is there, don't you see it? If repo name or org name changes, we need to hunt it down in 2 places, risking desync.

File: waspc/tools/install_wasp_app_runner.sh (hunk 32 above)
Comment (Martinsos) [reply waspc/tools/install_wasp_app_runner.sh:88]: Yeah I would drop the echo then. If all the scripts were written like this, it would be PITA to read/maintain them.

File: waspc/tools/install_wasp_app_runner.sh (hunk 33 above)
Comment (Martinsos) [reply waspc/tools/install_wasp_app_runner.sh:103]: I am going with Filip here -> let's just go with what commands output. Too many times did my output hide too much, and I anyway had to go an unsilent it. And it is less code on our side.

File: waspc/tools/install_wasp_app_runner.sh (hunk 27 above)
Comment (Martinsos) [reply waspc/tools/install_wasp_app_runner.sh:52]: What I don't like about returning string like this is that it behaves / looks same as normal result (version), so you get issues like that one with comparing two unknowns and claiming it is the same version and similar.

I know it is bash so our options are limited though, but I would try to find another way to handle this. If none, ok, but look into it.

File: waspc/tools/install_wasp_app_runner.sh (hunk 28 above)
Comment (FranjoMindek) [reply waspc/tools/install_wasp_app_runner.sh]: I did change to `exit` forgot to 🚀 .

File: waspc/tools/install_wasp_app_runner.sh (hunk 23 above)
Comment (sodic) [reply waspc/tools/install_wasp_app_runner.sh]: @Martinsos Perhaps you'd like this convention. I've also seen people use that:
```bash
command || echo "Error message"
```

Also, @Martinsos, why do your recent PR comments have "raw" at the end? I just know it's emacs-related :)

waspc/tools/install_wasp_app_runner.sh
sodic: Additional context for Franjo:
- Ubuntu does come with bash, always has
- My "What" was meant to communicate "that's not true." Not my greatest work

waspc/tools/install_wasp_app_runner.sh
Martinsos: @sodic it means "resolve at will"! Not emacs related

waspc/tools/install_wasp_app_runner.sh
Martinsos: Explanation/correction on my side:
- Ubuntu comes with `dash` as default non-interactive shell, while it does also come with `bash` and that is default interactive shell.
- So there is no problem here with you using `bash`, Ubuntu has it (and most systems).
- What I did was mixed this up with time ago, writing Wasp installer, when I had an issue of using `sh` shebang but also using bash features in the code, which actually worked on lot of systems because `sh` was alias for `bash`, but then would fail on Ubuntu. This is not the case here.

FranjoMindek:
I'm closing this PR in favor of https://github.com/wasp-lang/wasp/pull/2939.
We decided to go for the `npm` package solution instead.

PR #2812, opened 2025-06-03
Update the `todoApp` to be a proper kitchen sink app

### LINKED ISSUE #2740: Improve the `waspc/example/todoApp` to be a proper kitchen sink app
Today, the `waspc/example/todoApp` example is our main dev app which we use while developing Wasp. It contains most of Wasp's feature because over time as people built features, they first added them to that app. 

Because it grew organically, it's not really organized or well designed. We want to upgrade the `todoApp` to be easier to navigate around with specific organization that highlight which Wasp feature is used at a certain page of the app. For example, there could be a hierarchy of pages that highlight different features:
1. Actions & queries - page where we showcase how they are used e.g. what we already have, the TODO app
2. Jobs - page where we allow users to trigger async jobs
3. CRUD - page where we showcase the CRUD feature
4. Websockets - page where we showcase the Websockets feature e.g. upgrade our chat example to look nicer 

Part of this task should be enumerating which features the `todoApp` has, a which features it's missing. The feature matrix should help: https://docs.google.com/spreadsheets/d/1xV86yKJ_VKpcZO1QpxL_WZ7DPoyhK7rnCdcO3BHbzX0/edit?gid=0#gid=0

The organization should allow for:
1. Finding a Wasp feature
2. Testing a Wasp feature

infomiho:
Closes #2740

This PR touches only the `examples/todoApp` and nothing in the compiler. The idea was to make it easy to see all the different demos we have in the app and make it a bit nicer visually so we can deploy it and share it with users to test Wasp features.

<details>
<summary>Some screenshots</summary>

Screenshot 2025-06-04 at 12 13 36
Screenshot 2025-06-04 at 12 13 53
Screenshot 2025-06-04 at 14 59 00

</details>

The way to think about this PR:
- First check out the _headless tests_ changes to see the actual functional changes.  Ideally, you'll notice that there weren't many functional changes. There are some new tests for the new APIs and Jobs demo pages.
- Next, look at the new folder structure, most of the features live in the `features` dir. I've used `pages` to group pages at different dir levels and `components` to group components. 
- Next, there are two new components `Button` and `Input` that are used in most demos. I've generated them with LLMs and then minimized the code to still be useful to us. 
- Lastly, there are lot of styles added to make the kitchen sink app more modern and consistent. I mostly went for modern and clean look while using LLMs. I believe the style shouldn't bother anyone, but please let me know if there is something we can improve.


waspc/examples/todoApp/src/features/auth/pages/CustomSignupPage.tsx
```diff
@@ -50,9 +52,9 @@ export const SignupPage = () => {
   });
 
   return (
-    <>
-      <h1>Custom Signup Action</h1>
-      <form onSubmit={onSubmit}>
+    <FeatureContainer>
+      <h1 className="text-2xl font-medium mb-4">Custom Signup Form</h1>
+      <form onSubmit={onSubmit} className="card">
```
FranjoMindek: I've did some question answering on #questions channel, and went to todo to give myself an example of custom auth.

I've noticed we also export `<Form/>` component, so why not use that?

FranjoMindek: Maybe then go all the way and also label via `<FormLabel/>`

waspc/examples/todoApp/headless-tests/tests/auth-hooks.spec.ts
```diff
@@ -45,15 +46,26 @@ test.describe("auth hooks", () => {
...
-    await expect(page.locator("body")).toContainText(
-      "Value of user.isOnAfterSignupHookCalled is true.",
-    );
+    await page.goto("/profile");
 
-    await expect(page.locator("body")).toContainText(
-      "Value of user.isOnAfterLoginHookCalled is true.",
-    );
+    await expect(
+      page.getByTestId("hook-status-onAfterSignup").getByTestId("status"),
```
infomiho, line 52: I've found out about `data-testid` helpers which enable us to be really precise when writing selectors in headless tests.

waspc/examples/todoApp/src/components/Button.tsx
```diff
@@ -0,0 +1,98 @@
...
+  outline:
+    "bg-transparent hover:bg-primary-50 text-primary-600 border-primary-300 hover:border-primary-400",
+  ghost: "bg-transparent hover:bg-gray-100 text-gray-700 border-transparent",
+  danger:
+    "bg-red-600 hover:bg-red-700 text-white border-transparent shadow-sm hover:shadow-md",
+  success:
+    "bg-green-600 hover:bg-green-700 text-white border-transparent shadow-sm hover:shadow-md",
+  warning:
+    "bg-amber-600 hover:bg-amber-700 text-white border-transparent shadow-sm hover:shadow-md",
+};
```
infomiho: Not all variants are used now - but they should be useful for the future.

waspc/examples/todoApp/src/components/FeatureContainer.tsx
```diff
@@ -0,0 +1,114 @@
+import { useState } from "react";
+import { useLocation } from "react-router-dom";
+import { Link, type Routes } from "wasp/client/router";
+import { cn } from "../cn";
+
+export function FeatureContainer({ children }: React.PropsWithChildren<{}>) {
```
infomiho, line 6: This is wrapper for most demo pages which renders a sidebar with the list of the features on the left and the demo content on the right.

waspc/examples/todoApp/src/features/apis/apis.ts
```diff
@@ -7,6 +7,11 @@ import {
 } from "wasp/server/api";
 
 export const fooBar: FooBar = (_req, res, context) => {
+  if (!context.user) {
```
infomiho, line 10: I found it better to do this (and easier to test in headless tests) than to return `Hello, undefined`

waspc/examples/todoApp/src/features/auth/pages/EmailVerification.tsx
```diff
@@ -0,0 +1,23 @@
+import { Link } from "wasp/client/router";
```
infomiho, line 1: Code stayed the same mostly, the difference is the import of `customisationProps` that are applied to all Auth UI components.

waspc/examples/todoApp/src/features/auth/pages/Login.tsx
```diff
@@ -0,0 +1,35 @@
+import { Link } from "wasp/client/router";
```
infomiho, line 1: Code stayed the same mostly, the difference is the import of `customisationProps` that are applied to all Auth UI components.

waspc/examples/todoApp/src/features/auth/pages/PasswordReset.tsx
```diff
@@ -0,0 +1,23 @@
+import { Link } from "wasp/client/router";
```
infomiho, line 1: Code stayed the same mostly, the difference is the import of `customisationProps` that are applied to all Auth UI components.

waspc/examples/todoApp/src/features/auth/pages/RequestPasswordReset.tsx
```diff
@@ -0,0 +1,14 @@
+import { ForgotPasswordForm } from "wasp/client/auth";
```
infomiho, line 1: Code stayed the same mostly, the difference is the import of `customisationProps` that are applied to all Auth UI components.

waspc/examples/todoApp/src/features/auth/pages/Signup.tsx
```diff
@@ -0,0 +1,45 @@
+import { FormItemGroup, SignupForm } from "wasp/client/auth";
```
infomiho, line 1: Code stayed the same mostly, the difference is the import of `customisationProps` that are applied to all Auth UI components.

waspc/examples/todoApp/src/features/chat/ChatPage.tsx
```diff
@@ -0,0 +1,99 @@
+import React, { useRef, useState } from "react";
```
infomiho: Code is mostly the same, just style updates and uses the new `Button` and `Input` compnents.

waspc/examples/todoApp/headless-tests/tests/async-jobs.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+  generateRandomCredentials,
+  performEmailVerification,
+  performLogin,
+  performSignup,
+} from "./helpers";
+
+test.describe("async jobs", () => {
+  const { email, password } = generateRandomCredentials();
+
+  test.describe.configure({ mode: "serial" });
```
FranjoMindek: Nitpick: i find it weird that test configuration is after test constants

waspc/examples/todoApp/headless-tests/tests/async-jobs.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+    await performEmailVerification(page, email);
+  });
+
+  test("submits a job and gets result", async ({ page }) => {
+    await performLogin(page, {
+      email,
+      password,
+    });
+
+    await expect(page).toHaveURL("/");
```
FranjoMindek: We repeat
```ts
await performLogin(page, {
  email,
  password,
});

await expect(page).toHaveURL("/");
```
a lot, 8 times?

We have
```ts
test("can log in after signed up with custom action", async ({ page }) => {
  await performLogin(page, {
    email,
    password,
  });

  await expect(page).toHaveURL("/");
});
```
which handles this check for us, no need to do it repeat it elsewhere.
I think we should focus on "submit a job and gets result" here instead.

Not that I take a better look I see we repeat this setup pattern `new page -> signup -> email verification -> login` a lot. Why not extract it out to a single function? Maybe not include login? Login flag? I think you have better context of what to extract here.

waspc/examples/todoApp/headless-tests/tests/custom-apis.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    ).not.toBeVisible();
+
+    // Unauthenticated API should return data
+    await expect(
+      page.getByTestId("unauthenticated-api").getByTestId("error"),
+    ).not.toBeVisible();
+    await expect(
+      page.getByTestId("unauthenticated-api").getByTestId("data"),
+    ).toContainText("Hello, stranger!");
+  });
```
FranjoMindek: Can be helpers to remove comments:
```ts
expectApiToError("authenticated-api");
expectApiToSucceed("unauthenticated-api", "Hello, stranger!");

// or

expectApiToError({ testId: "authenticated-api" });
expectApiToSucceed({ testId: "authenticated-api", result: "Hello, stranger!" });
```

waspc/examples/todoApp/headless-tests/tests/custom-apis.spec.ts
FranjoMindek, line 1: Same points as in async jobs.

waspc/examples/todoApp/headless-tests/tests/custom-apis.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    await page.locator("input[type='password']").fill(password);
+    await page.getByRole("button", { name: "Log in" }).click();
+
+    await expect(page).toHaveURL("/");
+
+    await page.goto("/apis");
+
+    await expect(page).toHaveURL("/apis");
+
+    // Authenticated API should return data
```
FranjoMindek: Helper functions above.

waspc/examples/todoApp/headless-tests/tests/custom-signup.spec.ts
```diff
@@ -9,7 +9,7 @@ test.describe("custom signup", () => {
   test("can sign up", async ({ page }) => {
     await page.goto("/custom-signup");
 
-    await page.waitForSelector("text=Custom Signup Action");
+    await page.waitForSelector("text=Custom Signup Form");
```
FranjoMindek: Test ids would stop us from having to do this? If you want to go all-in on them.

waspc/examples/todoApp/src/App.tsx
```diff
@@ -15,50 +18,157 @@ export function App() {
...
-              <button className="btn btn-primary" onClick={logout}>
-                Logout
-              </button>
+
+            <div className="flex items-center space-x-4">
+              {user ? (
+                <UserMenu user={user} />
+              ) : (
+                <div className="flex items-center space-x-3">
+                  <Button to="/login" variant="ghost">
```
FranjoMindek: Should be `<Link/>`'s? Ah no, `to` makes `<Button/>` a `<Link/>`. 
That might be confusing for people?

I guess you wanted to reuse the styles, which is understandable.
I'm not sure which is better, describing same styles twice or having this.

waspc/examples/todoApp/headless-tests/tests/helpers.ts
```diff
@@ -11,7 +11,7 @@ export async function performSignup(
   await page.locator("input[type='email']").fill(email);
   await page.locator("input[type='password']").fill(password);
   await page.locator("input[name='address']").fill("Dummy address");
-  await page.locator("button").click();
+  await page.getByRole("button", { name: "Sign up" }).click();
```
FranjoMindek, line 41: ~~Isn't this incorrect?
Yes we use `<Button/>` component, but it renders `<Link/>` instead? Which should be an `<a/>`?
How come this passed the test?~~

Aha, this is our auth UI `Sign up` button, not the `Sign Up` (uppercase u) I've found on `App.tsx`.
Is maybe looking for `type="submit"` better? I'm not sure.

waspc/examples/todoApp/headless-tests/tests/helpers.ts
```diff
@@ -68,6 +68,12 @@ export async function performEmailVerification(
   await page.waitForSelector("text=Your email has been verified");
 }
 
+export function isRunningInDevMode() {
+  return (
+    !process.env.HEADLESS_TEST_MODE || process.env.HEADLESS_TEST_MODE === "dev"
```
FranjoMindek: Why `!process.env.HEADLESS_TEST_MODE`?
I can't figure it out.

waspc/examples/todoApp/headless-tests/tests/simple.spec.ts
```diff
@@ -52,27 +52,34 @@ test.describe("signup and login", () => {
       password,
     });
 
-    await expect(page).toHaveURL("/profile");
+    await expect(page).toHaveURL("/");
 
-    await page.goto("/");
+    await page.goto("/tasks");
 
     // Create a new task
```
FranjoMindek: See if this works as functions better or not?
Could easily be too much functions / not really helping.
I just love using function hoisting for better readability.

```ts
const task = createANewTask(page);
navigateToTaskPage(page, task);
```

waspc/examples/todoApp/package.json
```diff
@@ -8,10 +8,12 @@
   "dependencies": {
     "@playwright/test": "^1.50.0",
     "chai": "^5.2.0",
+    "clsx": "^2.1.1",
     "linebyline": "^1.3.0",
```
FranjoMindek, line 13: Do we need both?
I usually only have `tailwind-merge` in my projects, and use `twMerge` and `twJoin` as necessary.
Do we have any benefits of both?

Tailwind merge itself supports conditional logic, arrays...

waspc/examples/todoApp/tailwind.config.cjs
```diff
@@ -4,7 +4,23 @@ const { resolveProjectPath } = require("wasp/dev");
...
+          100: "#fffec1",
+          200: "#fff886",
+          300: "#ffec41",
+          400: "#ffdb0d",
+          500: "#ffcc00",
+          600: "#d19500",
+          700: "#a66a02",
+          800: "#89530a",
+          900: "#74430f",
+          950: "#442304",
```
FranjoMindek, line 20: I'm interested how you created this? 😄 
I once used a VSC plugin for that.

waspc/examples/todoApp/src/App.tsx
```diff
@@ -15,50 +18,157 @@ export function App() {
...
-          <option disabled value="">
-            Select one
-          </option>
-          <option value="/">Main page</option>
-          <option value="/serialization">Serialization Test</option>
-        </select>
-      </label>
+      <Outlet />
+
+      <footer className="bg-white border-t border-gray-200">
```
FranjoMindek: If it was me I would probably extract `<Footer/>` and `<Header/>` into separate components.
Resolve as you want.

waspc/examples/todoApp/package.json
```diff
@@ -8,10 +8,12 @@
...
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-router-dom": "^6.26.2",
+    "tailwind-merge": "^3.3.0",
```
FranjoMindek: This version of `tailwind-merge` is for tailwind v4

waspc/examples/todoApp/src/components/Button.tsx
```diff
@@ -0,0 +1,98 @@
...
+export function Button(props: ButtonProps) {
+  const { disabled, children, variant = "primary", className = "" } = props;
+
+  const classes = cn(
+    "inline-flex items-center justify-center",
+    "font-medium rounded-lg border",
+    "transition-all duration-200 ease-in-out",
+    "focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-primary-500",
+    "select-none",
+    "px-4 py-2 text-sm",
```
FranjoMindek: I woud like `font-medium text-sm` together, as well as `px-4 py-2 rounded-lg border`

waspc/examples/todoApp/src/components/Button.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  if (isLinkButton(props)) {
+    const { to, replace } = props;
+
+    return (
+      <Link
+        to={to}
+        replace={replace}
+        className={classes}
+        style={disabled ? { pointerEvents: "none" } : undefined}
```
FranjoMindek: Why use `style` instead of `className`? `pointer-events-none` works?
It's not a dynamically created string so it's necessary to use `style`.

Kinda s symptom of button being a link.

waspc/examples/todoApp/src/components/Button.tsx
```diff
@@ -0,0 +1,98 @@
...
+    disabled && disabledStyles,
+    className,
+  );
+
+  if (isLinkButton(props)) {
+    const { to, replace } = props;
+
+    return (
+      <Link
+        to={to}
```
FranjoMindek: I'd prefer approach of e.g.

```tsx
<Link {...props} className={classes} />
//same with 
<button {...props} className={classes} />
```

Since it's cleaner to me, but you may resolve as you want.


I usually do it like 
```tsx
export function Button({
	className, 
	variant =  "primary",
	...props}: ButtonProps) {
}
```

But it's same as doing it inside of function.
`className` doesn't need to be initialized to `""`, `undefined` is fine

waspc/examples/todoApp/src/components/FeatureContainer.tsx
```diff
@@ -0,0 +1,154 @@
+import { useState } from "react";
+import { useLocation } from "react-router-dom";
+import { Link, type Routes } from "wasp/client/router";
+import { cn } from "../cn";
+
+const features: Feature[] = [
```
FranjoMindek: I like having public API first, e.g. exported stuff, that's why I move the constants down usually.
But a nitpick preference.

waspc/examples/todoApp/src/components/FeatureContainer.tsx
```diff
@@ -0,0 +1,154 @@
...
+      className={cn(
+        "group relative block p-3 rounded-lg transition-all duration-200 hover:bg-gray-50 border border-transparent",
+        isActive
+          ? "bg-gray-50 border-gray-200 text-gray-900 hover:bg-gray-50"
+          : "text-gray-700 hover:text-gray-900",
+      )}
+    >
+      <div className="flex items-center gap-2">
+        <h3 className="font-medium text-sm leading-tight">{title}</h3>
+        {isPublic && <PublicFeature />}
```
FranjoMindek: I'd rather inline this one `<span>` element, and extract mobile and desktop `FeatureListMenu` instead.

waspc/examples/todoApp/src/components/Input.tsx
```diff
@@ -0,0 +1,60 @@
+import React, { forwardRef } from "react";
+import { cn } from "../cn";
+
+interface InputProps extends React.InputHTMLAttributes<HTMLInputElement> {
+  label?: string;
+  required?: boolean;
+  disabled?: boolean;
```
FranjoMindek: `required` and `disabled` are part of `React.InputHTMLAttributes<HTMLInputElement>`

waspc/examples/todoApp/src/components/Input.tsx
```diff
@@ -0,0 +1,60 @@
...
+  inputClassName?: string;
```
FranjoMindek, line 10: But you didn't disable `className`, now somebody can send `className`, `inputClassName`, and `containerClassName`. `className` will overwrite `inputClassName` since `{...textProps}`

FranjoMindek: We can use `Omit` to remove `className` from type.

waspc/examples/todoApp/src/components/Input.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const inputClasses = cn([
+    "block w-full px-3 py-2 rounded-lg border transition-colors duration-200 ease-in-out",
+    "focus:outline-none focus:ring-2 focus:ring-offset-0",
+    "placeholder-gray-400",
+    "disabled:opacity-50 disabled:cursor-not-allowed disabled:bg-gray-50",
+    "bg-white border-gray-300 focus:border-primary-500 focus:ring-primary-500",
+    inputClassName,
+  ]);
+
+  const inputId = React.useId();
```
FranjoMindek, line 34: TIL `React.useId();`

waspc/examples/todoApp/src/components/Input.tsx
FranjoMindek: Same for `inputId`, which is overwritten by `id`.

waspc/examples/todoApp/src/features/apis/pages/ApisPage.tsx
```diff
@@ -0,0 +1,78 @@
...
+          {unauthenticatedApi.data && (
+            <div data-testid="data">{unauthenticatedApi.data.msg}</div>
+          )}
+        </div>
+      </div>
+    </FeatureContainer>
+  );
+}
+
+function useCustomApi<Data = unknown>(
```
FranjoMindek, line 51: Oh damn, we don't provide access to `tanstack-query`?

I usually pefer:
```tsx
const [state, setState] = useState<{
  isLoading: boolean,
  data: Data | null,
  error: AxiosError | null,
}>({
  isLoading: true,
  data: null,
  error: null,
})

//with...
.then(({ data }) => {
  setState({
    data,
    error: null,
    isLoading: false
  })
})
```

Less accident prone, and reflects what actually happens, state changes together.

waspc/examples/todoApp/src/features/auth/pages/Login.tsx
```diff
@@ -0,0 +1,34 @@
...
+const Login = () => {
+  return (
+    <FeatureContainer>
+      <div className="max-w-sm mx-auto">
+        <LoginForm {...customisationProps} />
+        <br />
+        <span className="text-sm font-medium text-gray-900">
+          Don't have an account yet?{" "}
+          <Link to="/signup" className="link">
+            go to signup
```
FranjoMindek: I did this on my `basic` starter: `go to signup` -> `Go to signup`

Same for all else.

waspc/examples/todoApp/src/features/auth/pages/ProfilePage.tsx
```diff
@@ -0,0 +1,109 @@
...
+              isCalled={user.isOnAfterLoginHookCalled}
+            />
+          </div>
+        </div>
+      </div>
+    </FeatureContainer>
+  );
+};
+
+const SimpleCard = ({ children }: React.PropsWithChildren<{}>) => {
```
FranjoMindek: `<SimpleCard/>` below `<HookStatus/>`?
Since it's used by both.

waspc/examples/todoApp/src/features/auth/pages/ProfilePage.tsx
```diff
@@ -0,0 +1,109 @@
...
+          </div>
+        </div>
+
+        <div className="card">
+          <h2 className="text-xl font-semibold text-gray-900 mb-4 flex items-center">
+            Hook Status
+          </h2>
+          <div className="space-y-3">
+            <HookStatus
+              hookName="onAfterSignup"
```
FranjoMindek: Not sure if we should enumerate this?

waspc/examples/todoApp/src/features/auth/pages/auth-ui.ts
```diff
@@ -0,0 +1,10 @@
+export const customisationProps = {
+  appearance: {
+    colors: {
+      brand: "#ffcc00",
+      brandAccent: "#ffdb0d",
+      submitButtonText: "#1f2937",
+    },
+  },
+  socialLayout: "horizontal",
+} as const;
```
FranjoMindek: Can we get type for this?

waspc/examples/todoApp/src/features/crud/pages/ListPage.tsx
FranjoMindek, line 1: I usually like splitting stuff like this into multiple files, but I don't know if it just me and too much.
It's easier when each component handles it's state, under the condition that they are orthogonal and not dumb components.

waspc/examples/todoApp/src/features/jobs/pages/JobsPage.tsx
```diff
@@ -0,0 +1,154 @@
...
+import { Input } from "../../../components/Input";
+
+export function JobsPage() {
+  const { data: requests } = useQuery(getTextUppercaseRequests, null, {
+    refetchInterval: 1000,
+  });
+  const [jobPayload, setJobPayload] = useState("");
+  const [isSubmitting, setIsSubmitting] = useState(false);
+
+  async function onSubmitJob(e: React.FormEvent<HTMLFormElement>) {
```
FranjoMindek, line 46: Same as `useQuery`, here `useMutation` would help not re-implement this stuff.

waspc/examples/todoApp/src/features/jobs/pages/JobsPage.tsx
```diff
@@ -0,0 +1,154 @@
...
+    </FeatureContainer>
+  );
+}
+
+function StatusBadge({ state }: { state: UppercaseTextRequestState }) {
+  const getClasses = (state: UppercaseTextRequestState) => {
+    const baseClasses =
+      "inline-flex items-center px-3 py-1.5 rounded-lg text-xs font-semibold border shadow-sm";
+
+    switch (state) {
```
FranjoMindek: Why create `cn` utility, but not use it and revert back to `switch`?
This can be done as in any other big `className` we had.

e.g.
```ts
cn(
	baseClasses,
	state === UppercaseTextRequestState.PENDING && "classes",
	// ...
)
```

waspc/examples/todoApp/src/features/operations/pages/TaskDetailPage.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { Button } from "../../../components/Button";
+import { FeatureContainer } from "../../../components/FeatureContainer";
+
+type TaskPayload = Pick<Task, "id" | "isDone">;
+
+const VISIBILITY_EXPLANATION = {
+  [TaskVisibility.PRIVATE]: "only you",
+  [TaskVisibility.LINK_ONLY]: "people with a link",
+  [TaskVisibility.PUBLIC]: "everyone",
+} as const satisfies Record<TaskVisibility, string>;
```
FranjoMindek: IIRC @sodic said somewhere if it's a complete object (we handle each key, and we know all keys upfront),
that just:
```ts
const VISIBILITY_EXPLANATION: Record<TaskVisibility, string> = {
  [TaskVisibility.PRIVATE]: "only you",
  [TaskVisibility.LINK_ONLY]: "people with a link",
  [TaskVisibility.PUBLIC]: "everyone",
}
```
would be fine. 

I might link it a bit later if I can find it.

FranjoMindek (changes requested):
Whoops, accidentally submitted before writing out a comment.

I like the direction we are going with.
I have some changes, a lot of nitpicks and maybe stylistic/preference "resolve as you want" changes.

I lost concentration a bit towards the end, so I might have missed some stuff, but rather published the review now so you have something to work on.

Good job!

waspc/examples/todoApp/headless-tests/tests/async-jobs.spec.ts
infomiho: The `expect` part is here mainly to wait for navigation to complete:
```ts
await expect(page).toHaveURL("/");
```
Maybe there is a cleaner way of doing it, I'll investigate and try to reduce the repetition.

waspc/examples/todoApp/headless-tests/tests/custom-apis.spec.ts
infomiho: Great stuff!

waspc/examples/todoApp/headless-tests/tests/async-jobs.spec.ts
infomiho: Screenshot 2025-06-06 at 12 38 30

Thanks for nudging me to simplify this!

waspc/examples/todoApp/tailwind.config.cjs
infomiho: I've used this website to do it :) https://uicolors.app/generate/ffcc00

waspc/examples/todoApp/package.json
infomiho: I've used it like without much thought since this is how ShadCN does it. I see that `tailwind-merge` supports conditional classes, but I wanted to stick to using something that's "standard" even though it might be an overkill for our use case.

waspc/examples/todoApp/package.json
infomiho: Good catch! I didn't realize that there will be a difference but I now see in the package description which version is right for Tailwind v3.

waspc/examples/todoApp/src/components/Button.tsx
infomiho: Doing it inside of the function let's keep all the `props` in one object. We are missing some neat syntax like in Haskell `TypeName { name }@allProps`.

waspc/examples/todoApp/headless-tests/tests/async-jobs.spec.ts
```diff
@@ -1,38 +1,11 @@
...
-  performEmailVerification,
-  performLogin,
-  performSignup,
-} from "./helpers";
+import { performLogin, setupTestUser } from "./helpers";
 
 test.describe("async jobs", () => {
-  const { email, password } = generateRandomCredentials();
-
-  test.describe.configure({ mode: "serial" });
```
FranjoMindek: We lost `test.describe.configure` on a lot of places, this is intended?

waspc/examples/todoApp/package.json
FranjoMindek: Resolve as you like. I'm fine with either, just wanted to point out.

waspc/examples/todoApp/src/App.tsx
FranjoMindek: ~~I'm not sure what do you mean with 👍🏻, as in I will move it to separate components, or okay I will resolve as I want and won't separate. I don't see separation but I see 🚀 too so I'm confused.~~

EDIT:  the commit came shortly after

waspc/examples/todoApp/src/components/Button.tsx
FranjoMindek: I don't see this change?

waspc/examples/todoApp/src/components/Button.tsx
FranjoMindek: ~~But we don't need to keep all of the props properties.
I thought of something like this when I wrote it, only downside if you have to change the `isLinkButton` a bit then.~~

Nevermind I was too hasty with reviewing.

```tsx
export function Button({ variant = "primary", className, children, ...props}: ButtonProps) {
  // const { disabled } = props; // if you dislike props.disabled
  const classes = cn(
    "inline-flex items-center justify-center",
    "font-medium rounded-lg border",
    "transition-all duration-200 ease-in-out",
    "focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-primary-500",
    "select-none",
    "px-4 py-2 text-sm",
    variantStyles[variant],
    props.disabled && disabledStyles,
    className,
  );

  if ("to" in props) {
    return (
      <Link {...props} className={classes}>
        {children}
      </Link>
    );
  }

  return (
    <button {...props} className={classes}>
      {children}
    </button>
  );
}

// or

export function Button({variant = "primary", className,  ...props}: ButtonProps) {
  // const { disabled } = props; // if you dislike props.disabled
  const classes = cn(
    "inline-flex items-center justify-center",
    "font-medium rounded-lg border",
    "transition-all duration-200 ease-in-out",
    "focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-primary-500",
    "select-none",
    "px-4 py-2 text-sm",
    variantStyles[variant],
    props.disabled && disabledStyles,
    className,
  );

  if ("to" in props) {
    return (
      <Link {...props} className={classes}/>
    );
  }

  return (
    <button {...props} className={classes}/>
  );
}
```

waspc/examples/todoApp/src/features/auth/pages/ProfilePage.tsx
infomiho: Yep, let's do it 👍

waspc/examples/todoApp/src/features/auth/pages/CustomSignupPage.tsx
infomiho: We don't export `Form` unfortunately but we are using `FormLabel` for labels in this custom signup form :) 

Auth UI components need a rework: better styles, better style customization etc. so it's really perfect, but we'll improve once we improve Auth UI.

waspc/examples/todoApp/src/App.tsx
infomiho: I went with explicit `Button` and `ButtonLink` components to make the code simpler :)

waspc/examples/todoApp/headless-tests/tests/helpers.ts
infomiho: Dev was the implicit default value, I've now made it explicit :)

waspc/examples/todoApp/headless-tests/tests/simple.spec.ts
infomiho: I think it's okay to keep the step explicit like this. I found that it makes updating headless test easier. Going to page, entering some values, expecting to see something. I don't find much value in abstracting that away.

waspc/examples/todoApp/src/features/crud/pages/ListPage.tsx
infomiho: I'd like to keep it in one file in the name of simplicity, but I did rewrite the state and moved it into the component themselves. You are right, it's much nicer!

waspc/examples/todoApp/headless-tests/tests/async-jobs.spec.ts
infomiho: Yes sir, it wasn't necessary for most tests. We needed the serial execution only for auth tests (where I kept the logic). For most tests, parallel execution is fine since we log in at the start of each test.

waspc/examples/todoApp/src/features/apis/pages/ApisPage.tsx
infomiho: Woah, you are right, this is so much better with Tanstack Query! I haven't thought of using it... users need to install it in their `package.json` and then they can use it without trouble since we already installed it in the background for them.

infomiho:
@FranjoMindek I've tried to address all of your comments, which were great and made the app much better IMHO

waspc/examples/todoApp/headless-tests/tests/async-jobs.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+    await page.getByRole("button", { name: "Submit Job" }).click();
+
+    await expect(
+      page.getByTestId("job-request").getByTestId("input"),
+    ).toContainText(text);
+    await expect(
+      page.getByTestId("job-request").getByTestId("status"),
+    ).toContainText("pending");
+
+    await page.waitForTimeout(2000);
```
Martinsos: Any other way to know how ong we need to wait here? Or to wait for something else? Waiting for fixed period of time is always fragile.

waspc/examples/todoApp/headless-tests/tests/auth-hooks.spec.ts
```diff
@@ -45,15 +45,24 @@ test.describe("auth hooks", () => {
...
+    ).toContainText("Called");
+
+    const expectedEmailVerificationStatus = isRunningInDevMode()
+      ? "Not Called"
+      : "Called";
+    await expect(
+      page
+        .getByTestId("hook-status-onAfterEmailVerified")
+        .getByTestId("status"),
+    ).toContainText(expectedEmailVerificationStatus);
```
FranjoMindek: Using `toContainText` might make this test invalid, since `"Called"` is contained in both `"Called"` and `"Not Called"`.
A better alternative would be a strict string equality.

waspc/examples/todoApp/headless-tests/tests/auth.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+import {
+  generateRandomCredentials,
+  performEmailVerification,
+  submitLoginForm,
+  submitSignupForm,
+} from "./helpers";
+
+test.describe("auth", () => {
+  // We need the login test to run after the signup test
+  test.describe.configure({ mode: "serial" });
```
FranjoMindek: I would maybe move this `test.describe.configure({ mode: "serial" });` to part of the test which is really serial.
e.g. in this test the signup->login flow and social button renders can be tested parallelly.

It can probably be done by moving signup->login flow into another nested `describe`.

waspc/examples/todoApp/headless-tests/tests/auth.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  const { email, password } = generateRandomCredentials();
```
FranjoMindek: `generateRandomCredentials` generates email credentials specifically. Can we reflect this in the function name?

waspc/examples/todoApp/headless-tests/tests/auth.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  test("social button renders", async ({ page }) => {
+    await page.goto("/signup");
+
+    await page.waitForSelector("text=Create a new account");
```
FranjoMindek: This `await page.waitForSelector` is actually part of the setup, waiting for page to finish loading.
This newline makes it seems like something else.

waspc/examples/todoApp/headless-tests/tests/auth.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    });
+
+    await expect(page.locator("body")).toContainText(
+      `You've signed up successfully! Check your email for the confirmation link.`,
+    );
+
+    await performEmailVerification(page, email);
+  });
+
+  test("can log in and create a task", async ({ page }) => {
```
FranjoMindek: The test is named `"can log in and create a task"` but it actually test both the successful and invalid log in at once.
Maybe two tests?  Or change the name? Also `"and create a task"` part is wrong?

waspc/examples/todoApp/headless-tests/tests/auth.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    await submitSignupForm(page, {
+      email,
+      password,
+    });
+
+    await expect(page.locator("body")).toContainText(
+      `You've signed up successfully! Check your email for the confirmation link.`,
+    );
+
+    await performEmailVerification(page, email);
```
FranjoMindek: This feels odd one out. Having additional functionality after testing everything. Any better way? RAW

waspc/examples/todoApp/headless-tests/tests/auth.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { expect, test } from "@playwright/test";
+import {
+  generateRandomCredentials,
+  performEmailVerification,
+  submitLoginForm,
+  submitSignupForm,
+} from "./helpers";
+
+test.describe("auth", () => {
+  // We need the login test to run after the signup test
```
Martinsos:
```suggestion
  // We need the login test to run after the signup test.
```

waspc/examples/todoApp/headless-tests/tests/auth.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+  test.describe.configure({ mode: "serial" });
+
+  const { email, password } = generateRandomCredentials();
+
+  test("social button renders", async ({ page }) => {
+    await page.goto("/signup");
```
Martinsos, line 13: This test is named "social button renders". But where, in which context?
I see in the test that you go to `/signup` page. So you actually assume that as a context. Let's make that explicit somehow.

Also, I see below that other tests then assume they are already on the signup page. Which makes them dependent on this test (which kind of makes sense, it is running serial, but due to stuff above it is not very clear all together).

Maybe having one big describe that says we are on the signup page (and does the setup by moving there) would be enough? Or maybe each test wants to go to signup page to ensure that part? Not sure but we we should do something.

waspc/examples/todoApp/headless-tests/tests/auth.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+      `You've signed up successfully! Check your email for the confirmation link.`,
+    );
+
+    await performEmailVerification(page, email);
+  });
+
+  test("can log in and create a task", async ({ page }) => {
+    await submitLoginForm(page, {
+      email,
+      password: "12345678xxx",
```
Martinsos: There is a super small chance that this password will actually be valid password! Hah. And we can avoid that easily (e.g. just take original password and add one character to it or modify one character (while not making it be as it was)).

Nitpick: Maybe also extracting it and naming it `invalidPassword` will help a bit with readibility / intention.

waspc/examples/todoApp/headless-tests/tests/auth.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  const { email, password } = generateRandomCredentials();
+
+  test("social button renders", async ({ page }) => {
+    await page.goto("/signup");
+
+    await page.waitForSelector("text=Create a new account");
+
+    await expect(
+      page.locator("a[href='http://localhost:3001/auth/google/login']"),
```
Martinsos: Are we ok with hardcoding `localhost:3001` here? Is that a part of how the tests are set up -> if so, should that be obtained from somewhere, or at least we should centralize that in one place (I guess the port especially)?

FranjoMindek (changes requested):
Peer code review with @Martinsos.

waspc/examples/todoApp/src/App.tsx
FranjoMindek: I like the approach.

waspc/examples/todoApp/src/components/Button.tsx
```diff
@@ -0,0 +1,99 @@
...
+    </Link>
+  );
+}
+
+function getButtonClasses({
+  variant,
+  disabled,
+  className,
+}: {
+  variant: ButtonVariant;
```
FranjoMindek: This is just a subset of `BaseButtonProps` if you want you can use e.g. `Omit<BaseButtonProps, "children">` or something similar. RAW

waspc/examples/todoApp/src/components/Button.tsx
```diff
@@ -0,0 +1,99 @@
...
+  onClick?: (e: React.MouseEvent<HTMLButtonElement>) => void;
+  type?: "button" | "submit" | "reset";
+}
+
+interface LinkButtonProps extends BaseButtonProps {
+  to: string;
+  replace?: boolean;
+}
+
+const variantStyles: Record<ButtonVariant, string> = {
```
FranjoMindek: I would push `variantStyles` below `getButtonClasses`

waspc/examples/todoApp/src/features/crud/pages/ListPage.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+  const [error, setError] = useState("");
+  const [isEditing, setIsEditing] = useState<number | null>(null);
+
+  async function handleCreateTask(data: { description: string }) {
+    setError("");
+    try {
+      await createTask(data);
+    } catch (err: unknown) {
+      setError("Error creating task.");
```
FranjoMindek, line 28: Btw `String(err)` works even on `unknown`.
You don't have to show a error to user, just a trick.

Edit: `undefined` -> `unknown`

waspc/examples/todoApp/src/features/jobs/pages/JobsPage.tsx
FranjoMindek: @infomiho missed this one?

waspc/examples/todoApp/src/features/operations/pages/TaskDetailPage.tsx
```diff
@@ -63,29 +59,39 @@ const Todo = () => {
...
-      <Link to="/">Go to dashboard</Link>
-    </>
+    <FeatureContainer>
+      <div className="space-y-4">
+        <div className="card" data-testid="task-detail">
+          {isLoading ? (
+            <div className="text-center py-8 text-gray-500">
+              Fetching task...
+            </div>
+          ) : (
```
FranjoMindek: I usually like to be more explicit with `tanstack-query` since they allow for it.
e.g.

```tsx
return (
  <>
    {isLoading && (<>Loading...</>)}
    {isSuccess && (<>Actual content</>)}
    {isError && (<>Handle error</>)}
  </>
)
```

This is very easy on eyes to quickly scan through what is rendered when.
Ternary operators sometimes make this hard.

Bonus point: inside of `isSuccess` the `data` can't be undefined. Don't know if our version of `tanstack-query` affects this comment.

waspc/examples/todoApp/src/features/jobs/pages/JobsPage.tsx
infomiho: It would seem so, yes! I'll check it out 👍

waspc/examples/todoApp/headless-tests/tests/async-jobs.spec.ts
infomiho: Playwright tests usually have internal timeout and they retry the test until it passes. I put here extra 2000ms to be extra safe - but I can simply remove it and it would still work.

waspc/examples/todoApp/headless-tests/tests/auth.spec.ts
infomiho: > There is a super small chance that this password will actually be valid password! 

Actually, no, becuase it's a hard-coded `123456`. Nice try though :)

I'll do it what you suggested anyways 👍

waspc/examples/todoApp/headless-tests/tests/auth.spec.ts
infomiho: By convention, our backend port is `3001` and our Playwright config also relies on that convention. This is the only place where the port is used, so I don't see any value in extracting it somewhere.

waspc/examples/todoApp/headless-tests/tests/async-jobs.spec.ts
Martinsos: But will it wait long enough? If so, sure, let's remove it. If not sure, then how do we know how long do we need to wait? How did you come up with 2s?

waspc/examples/todoApp/headless-tests/tests/auth.spec.ts
Martinsos: But you do use the function called `generateRandomCredentials` to generate the password you use to signup. You are telling me that function doesn't really genreate random credentials, but always returns `123456`? If that is so, then you have two problems, not one, additional one being that the function `generateRandomCredentials` doesn't do what it says it does.

waspc/examples/todoApp/headless-tests/tests/auth.spec.ts
infomiho: You are quite right, I've now added a proper email verification test 👍

waspc/examples/todoApp/src/features/crud/pages/ListPage.tsx
FranjoMindek: @infomiho tagging you because I misstyped `undefined` instead of `unknown`. Whoops.

waspc/examples/todoApp/headless-tests/tests/auth.spec.ts
Martinsos: Likely, with time, there will be more places in the tests where it is used. Extracting it right now might be a good move. It would be even better if you could somehow pull that information from the playwright config itself.
But do as you like, you have a better feeling for the situation here.
RAW

waspc/examples/todoApp/src/rpcTests/operations/SerializationPage.tsx
cprecioso, line 1: I'd say this can go to `features/operations` since the serialization only works on those

cprecioso (commented):
Tested it on my local machine, gave a general look, works fine! 💪 Also, awesome job on look-and-feel, and organization 🔥
Just two things:
- the `.env.server.example` doesn't have the SLACK_* env vars, so the app will not run if we do `npm run env:pull`
- (see comment below)

infomiho:
@cprecioso the `.env.server.example` has the Slack dummy variables, but we don't have them in the remote https://www.dotenv.org/ storage, that's true. I'll do that in separate PR since it's out of scope here: https://github.com/wasp-lang/wasp/pull/2842

I'll put some dummy Slack credentials because sharing a real client ID and secret is problematic. Slack requires HTTPS even when working locally. The easiest way for me to get HTTPS is to use something like Localtunnel which gives me a way to have HTTPS for the server easily but then that URL is random (or at least you'd need to use a specific subdomain) which is then hard to share amongst team members. Having the dummy values will not crash the app, but to get functional Slack auth you'll need to get your own client ID and secret.

waspc/examples/todoApp/headless-tests/tests/auth.spec.ts
infomiho: It generates a random email and a fixed password, for the purposes of headless tests - this was good enough 👍

Martinsos: I think you commented on the wrong comment thread here!

As for the `generateRandomCredentials` function -> that is ok, if what you need is fixed password and random email, and that works, for you, great, but what I am saying is that name is misleading -> when I was reading the code, and I see `generateRandomCredentials`, I don't expect it to be non-random password. Which means that I might easily call it from somewhere else with wrong expectations. This is a naming issue. You should change the name. Maybe rewrite it to `generateRandomUsername` and have that fixed password as a fixed thing. Or rename it to `generateSomewhatRandomizedCredentials`. Not sure what is the best solution, you will know what you want.

Martinsos:
@infomiho just a ping that there are some unresolved comments of mine, in case you missed them!

waspc/examples/todoApp/headless-tests/tests/auth.spec.ts
infomiho: I've reworked the function to be `generateRandomEmail()`

waspc/examples/todoApp/headless-tests/tests/auth.spec.ts
infomiho: I've extracted the port into a `WASP_SERVER_PORT` constant that used in both the config and the helpers

waspc/examples/todoApp/headless-tests/tests/async-jobs.spec.ts
infomiho: I've now removed the `waitForTimeout` and it will work since Playwright keeps retrying for some time (5 seconds I believe, which is plenty). I've come up with 2s since that's the fake delay I added in the job to simulate longer processing.

infomiho:
@Martinsos Please take another look

waspc/examples/todoApp/headless-tests/tests/async-jobs.spec.ts
Martinsos: Ok great!

If we decided to stay with 2s, then it would be good if we at least documented why it is 2s (because of the delay in job being 2 seconds) or even better, connecting the two via code if not too hard.

This way we avoid it completely so even better.

RAW

Martinsos:
> @Martinsos Please take another look

Took another look! Two main comments are resolved, great (I left comment on one for you to read and then resolve). I think one you maybe missed still: https://github.com/wasp-lang/wasp/pull/2812#discussion_r2135270437 since there is no reaction (I didn't check the code though). But it is a tiny one so I won't be taking another look. 

I will approve now but with a "comment", not the "approve", since I reviewed only a tiny bit of a PR.

PR #2818, opened 2025-06-03
Simplified auth pills to single color

### LINKED ISSUE #2775: Simplity auth pills in the docs
Image

We have these pills in the docs that are supposed to link to different auth method and signify which Auth UI component is used with which auth method. While useful, it's a pain to come up with new distinct colors (light and dark mode) each time. We should simplify the design to be e.g. just while pill with a black border and maybe a smaller font.

zkmrtn8:
### Description
Simplified the auth pills to be a single color. Works on both light and dark mode with no needed tweaks.

Example:

image
image

All instances of the pills across all versions were updated, and I've verified that the actual content of the pages are not changed other than the pill order in a few spots.

No longer needs separate components for each auth provider. Just use `<AuthPills providers={['github','username','email','keycloak']} />`. The component automatically inserts commas and "and"

Fixes #2775


Scorpil:
I think you want main as a base branch - there's one more pill there (slack).

web/docs/auth/Pills.jsx
```diff
@@ -1,99 +1,57 @@
...
-        textDecoration: 'none',
-        display: 'inline-block',
-        ...style,
+        padding: "0.1rem 0.5rem",
+        borderRadius: "0.375rem",
+        color: "white",
+        backgroundColor: "#444",
+        textDecoration: "none",
+        display: "inline-block",
+        marginRight: "0.25rem",
```
cprecioso, line 28: I know this wasn't in the original code, but can we move this to a be a class in a  CSS file, and have here only the `className`? (non-blocking)

web/docs/auth/Pills.jsx
```diff
@@ -1,99 +1,57 @@
...
+  username: {
+    label: "Username & Password",
+    link: "/docs/auth/username-and-pass",
+  },
+  slack: { label: "Slack", link: "/docs/auth/social-auth/slack" },
+  discord: { label: "Discord", link: "/docs/auth/social-auth/discord" },
+  github: { label: "Github", link: "/docs/auth/social-auth/github" },
+  google: { label: "Google", link: "/docs/auth/social-auth/google" },
+  keycloak: { label: "Keycloak", link: "/docs/auth/social-auth/keycloak" },
+};
```
cprecioso, line 15: Nice!

web/docs/auth/Pills.jsx
```diff
@@ -1,99 +1,57 @@
...
-      style={{
-        backgroundColor: 'var(--auth-pills-google)',
-      }}
-      linkToPage="/docs/auth/social-auth/google"
-    >
-      Google
-    </Pill>
-  )
-}
+    <span style={{ display: "inline" }}>
```
cprecioso, line 42: Same with this inline style

web/docs/auth/Pills.jsx
cprecioso, line 1: Since this file's main export is now `AuthPills`, we should rename it to match

web/docs/auth/Pills.jsx
```diff
@@ -1,99 +1,57 @@
...
-      linkToPage="/docs/auth/social-auth/github"
-    >
-      Github
-    </Pill>
-  )
-}
+export function AuthPills({ providers }) {
+  const validPills = providers
+    .map((provider) => PILL_INFO[provider])
+    .filter(Boolean);
```
cprecioso, line 39: Let's make this a bit more robust and resistant to mis-use: instead of silently ignoring bad provider names, throw an error if the provider name doesn't exist!

web/docs/auth/Pills.jsx
cprecioso, line 1: If you're feeling up to it, you can also translate this file to TS (it will probably be very easy!). I won't block on it though.

web/docs/auth/Pills.jsx
cprecioso, line 28: Why do we add the `marginRight` here? It makes it look like there's a space between the pill and the comma:

Before
image

After
image

web/docs/auth/Pills.jsx
```diff
@@ -1,99 +1,57 @@
...
-      linkToPage="/docs/auth/social-auth/keycloak"
-    >
-      Keycloak
-    </Pill>
-  )
+        return (
+          <React.Fragment key={pill.link}>
+            {isLast && validPills.length > 1 && "and "}
+            <Pill label={pill.label} link={pill.link} />
+            {!isLast && (isSecondLast ? ", " : ", ")}
```
cprecioso, line 51: Do we need the `isSecondLast` variable? It's not being used.

cprecioso (changes requested):
Hi, thanks for the contribution! And good job! I like the comma handling and that the links are centrally defined too.

I'd like to see some changes to this PR, all quite small, don't worry. After that, its good to go!

- Can we rebase to `main` branch? We don't need to make a hotfix for this
- Can we undo the changes in the `versioned_docs`? Those are considered frozen and we should only modify them for critical issues

web/docs/auth/auth-hooks.md
```diff
@@ -108,7 +108,7 @@ Wasp calls the `onBeforeSignup` hook before the user is created.
 
 The `onBeforeSignup` hook can be useful if you want to reject a user based on some criteria before they sign up.
 
-Works with <EmailPill /> <UsernameAndPasswordPill /> <DiscordPill /> <GithubPill /> <GooglePill /> <KeycloakPill />
+Works with <AuthPills providers={['username','email','github','google','keycloak','discord']}/>.
```
cprecioso, line 111: > I've verified that the actual content of the pages are not changed other than the pill order in a few spots

Not a super big issue, but why reorder the pills?

developerzeke: I copy/pasted them between pages, and ui.md and auth-hooks.md both had the same pills but in a different order. This way they match.

cprecioso: Makes sense!

developerzeke:
@cprecioso 
> * Can we rebase to `main` branch? We don't need to make a hotfix for this
> * Can we undo the changes in the `versioned_docs`? Those are considered frozen and we should only modify them for critical issues
image
I had based it on `main`, but then I saw this and moved it over to `release`.

developerzeke:
@cprecioso there were too many edits to continue on this branch easily, so I've moved it to #2821, which merges a different branch of my fork.

web/docs/auth/Pills.jsx
developerzeke: There were errors about me using JSX in the file, even though the filename was .tsx, so I'm not going to change this at the moment.

web/docs/auth/Pills.jsx
developerzeke: added in new pr, #2821

web/docs/auth/Pills.jsx
developerzeke: renamed in new pr, #2821

web/docs/auth/Pills.jsx
developerzeke: changed in new pr, #2821

web/docs/auth/Pills.jsx
developerzeke: changed in new pr, #2821

developerzeke: fixed in new pr, #2821. it needs a bit of margin so the comma is separated from the pill, but I've made it a more reasonable number now.

PR #2796, opened 2025-05-26
Create `wasp build start` command

cprecioso:
Resolves #1883

Implements a new `wasp build start` command.
It will take the output from `wasp build`, build the client, the server, and run them both.

The logic is consistent with `wasp-app-runner`, except we don't run the SMTP server or the DB.

Nothing is excessively surprising, except maybe the creation of a `JobExcept` type to make working with the jobs easier. This might later grow into a full orchestration utility if we want.

PR #661, opened 2022-07-01
DRAFT: Fix versioned package npm bug

shayneczyzewski:
# Description

This attempts to fix an `npm` bug which is present in the latest version of `node` (issue here: https://github.com/npm/cli/issues/5018).

Previously, we attempted to issue Prisma commands like so: `npx --no-install prisma@3.15.2 studio` This was because we wanted to ensure we did not use a globally installed one (possibly with a different version), nor ask users to install in the console if missing. But we also needed to be sure it would _somehow_ get installed on a clean project, so we had a fallback mechanism that if the command failed, it tried to `npm install` and then retried the failed command. All of this worked until that bug.

However, there was some progress in Wasp that makes some of this easier. We no longer need to worry about Prisma (or any NPM dep) not being installed, since we now `npm install` during compilation. Compilation happens before every command, so we can be sure we will have a local Prisma installation before we run any Prisma commands. Therefore, it feels pretty safe to use the local `node_modules` version. Thoughts? Risks?

## Type of change

Please select the option(s) that is more relevant.

- [ ] Code cleanup
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

PR #2820, opened 2025-06-04
Improve the DBConnectionEstablished Command Requirement

cprecioso:
Part of #2796

- Prisma commands now run as `npm exec prisma` instead of `node_modules/.bin/prisma`, so we don't need to hardcode binary paths

- Added an option to call a prisma command with the `.env.server` environment, even when the user hasn't run `wasp start` (e.g. a CI that is only running `wasp build`)

- The `DBConnectionEstablished` requirement now prints the Prisma command output if it received an unknown error, helping the user pinpoint the error
  - Specifically, in #2796 I was getting errors for not having `DATABASE_URL` env set, but I couldn't see the prisma output that says so. We can't scan the output for it specifically because it is reported as a `P1012 Invalid schema file`, which could be conflated with actual errors we would like to show to our users.

image

- The `DBConnectionEstablished` is not a requirement anymore, by itself. Instead, we need to wrap it in a `InBuildDir` or `InOutDir` type that will run this requirement in the correct context.

### DIFF
--- a/waspc/cli/src/Wasp/Cli/Command.hs
+++ b/waspc/cli/src/Wasp/Cli/Command.hs
@@ -11,6 +11,7 @@ module Wasp.Cli.Command
     -- See "Wasp.Cli.Command.Requires" for documentation.
     require,
     Requirable (checkRequirement),
+    InDirRequirable (checkRequirementInDir),
   )
 where
 
@@ -20,8 +21,10 @@ import Control.Monad.IO.Class (MonadIO)
 import Control.Monad.State.Strict (StateT, evalStateT, gets, modify)
 import Data.Data (Typeable, cast)
 import Data.Maybe (mapMaybe)
+import qualified StrongPath as SP
 import System.Exit (exitFailure)
 import Wasp.Cli.Message (cliSendMessage)
+import Wasp.Generator (ProjectRootDir)
 import qualified Wasp.Message as Msg
 
 newtype Command a = Command {_runCommand :: StateT [Requirement] (ExceptT CommandError IO) a}
@@ -68,3 +71,6 @@ require =
       req <- checkRequirement
       Command $ modify (Requirement req :)
       return req
+
+class InDirRequirable a where
+  checkRequirementInDir :: SP.Path' SP.Abs (SP.Dir ProjectRootDir) -> Command a
--- a/waspc/cli/src/Wasp/Cli/Command/Require.hs
+++ b/waspc/cli/src/Wasp/Cli/Command/Require.hs
@@ -24,6 +24,8 @@ module Wasp.Cli.Command.Require
     -- * Requirables
     Requirable (checkRequirement),
     DbConnectionEstablished (DbConnectionEstablished),
+    InBuildDir (InBuildDir),
+    InOutDir (InOutDir),
     InWaspProject (InWaspProject),
   )
 where
@@ -36,7 +38,7 @@ import Data.Maybe (fromJust)
 import qualified StrongPath as SP
 import System.Directory (doesFileExist, doesPathExist, getCurrentDirectory)
 import qualified System.FilePath as FP
-import Wasp.Cli.Command (CommandError (CommandError), Requirable (checkRequirement), require)
+import Wasp.Cli.Command (CommandError (CommandError), InDirRequirable (checkRequirementInDir), Requirable (checkRequirement), require)
 import Wasp.Generator.DbGenerator.Operations (isDbConnectionPossible, testDbConnection)
 import Wasp.Job.IO (readJobMessagesAndPrintThemPrefixed)
 import Wasp.Project.Common (WaspProjectDir)
@@ -44,23 +46,16 @@ import qualified Wasp.Project.Common as Project.Common
 
 data DbConnectionEstablished = DbConnectionEstablished deriving (Typeable)
 
-instance Requirable DbConnectionEstablished where
-  checkRequirement = do
-    -- NOTE: 'InWaspProject' does not depend on this requirement, so this
-    -- call to 'require' will not result in an infinite loop.
-    InWaspProject waspProjectDir <- require
-    let outDir =
-          waspProjectDir
-            SP.</> Project.Common.dotWaspDirInWaspProjectDir
-            SP.</> Project.Common.generatedCodeDirInDotWaspDir
-    dbIsRunning <- liftIO $ isDbConnectionPossible <$> testDbConnection outDir
+instance InDirRequirable DbConnectionEstablished where
+  checkRequirementInDir projectDir = do
+    dbIsRunning <- liftIO $ isDbConnectionPossible <$> testDbConnection projectDir
 
     either
       ( \chan -> do
           liftIO (readJobMessagesAndPrintThemPrefixed chan)
           throwError noDbError
       )
-      (\_ -> return DbConnectionEstablished)
+      (const $ return DbConnectionEstablished)
       dbIsRunning
     where
       noDbError =
@@ -105,3 +100,25 @@ instance Requirable InWaspProject where
           ( "Couldn't find wasp project root - make sure"
               ++ " you are running this command from a Wasp project."
           )
+
+data InOutDir req = InOutDir req deriving (Typeable)
+
+instance (Typeable req, InDirRequirable req) => Requirable (InOutDir req) where
+  checkRequirement = do
+    InWaspProject waspProjectDir <- require
+    let buildDir =
+          waspProjectDir
+            SP.</> Project.Common.dotWaspDirInWaspProjectDir
+            SP.</> Project.Common.generatedCodeDirInDotWaspDir
+    InOutDir <$> checkRequirementInDir buildDir
+
+data InBuildDir req = InBuildDir req deriving (Typeable)
+
+instance (Typeable req, InDirRequirable req) => Requirable (InBuildDir req) where
+  checkRequirement = do
+    InWaspProject waspProjectDir <- require
+    let buildDir =
+          waspProjectDir
+            SP.</> Project.Common.dotWaspDirInWaspProjectDir
+            SP.</> Project.Common.buildDirInDotWaspDir
+    InBuildDir <$> checkRequirementInDir buildDir
--- a/waspc/cli/src/Wasp/Cli/Command/Require.hs
+++ b/waspc/cli/src/Wasp/Cli/Command/Require.hs
@@ -38,6 +38,7 @@ import System.Directory (doesFileExist, doesPathExist, getCurrentDirectory)
 import qualified System.FilePath as FP
 import Wasp.Cli.Command (CommandError (CommandError), Requirable (checkRequirement), require)
 import Wasp.Generator.DbGenerator.Operations (isDbConnectionPossible, testDbConnection)
+import Wasp.Job.IO (readJobMessagesAndPrintThemPrefixed)
 import Wasp.Project.Common (WaspProjectDir)
 import qualified Wasp.Project.Common as Project.Common
 
@@ -54,9 +55,13 @@ instance Requirable DbConnectionEstablished where
             SP.</> Project.Common.generatedCodeDirInDotWaspDir
     dbIsRunning <- liftIO $ isDbConnectionPossible <$> testDbConnection outDir
 
-    if dbIsRunning
-      then return DbConnectionEstablished
-      else throwError noDbError
+    either
+      ( \chan -> do
+          liftIO (readJobMessagesAndPrintThemPrefixed chan)
+          throwError noDbError
+      )
+      (\_ -> return DbConnectionEstablished)
+      dbIsRunning
     where
       noDbError =
         CommandError
--- a/waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
+++ b/waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
@@ -24,6 +24,7 @@ import Wasp.Generator.ServerGenerator.Db.Seed (dbSeedNameEnvVarName)
 import qualified Wasp.Job as J
 import Wasp.Job.Process (runNodeCommandAsJob, runNodeCommandAsJobWithExtraEnv)
 import Wasp.Project.Common (WaspProjectDir, waspProjectDirFromProjectRootDir)
+import Wasp.Project.Env (readDotEnvServer)
 
 migrateDev :: Path' Abs (Dir ProjectRootDir) -> MigrateArgs -> J.Job
 migrateDev projectRootDir migrateArgs =
@@ -131,7 +132,6 @@ seed projectRootDir seedName =
   runPrismaCommandAsJobWithExtraEnv
     serverDir
     [(dbSeedNameEnvVarName, seedName)]
-    projectRootDir
     ["db", "seed"]
   where
     serverDir = projectRootDir </> serverRootDirInProjectRootDir
@@ -144,8 +144,9 @@ seed projectRootDir seedName =
 -- SQL command, which works perfectly for checking if the database is running.
 dbExecuteTest :: Path' Abs (Dir ProjectRootDir) -> J.Job
 dbExecuteTest projectRootDir =
-  runPrismaCommandAsJobFromWaspServerDir projectRootDir ["db", "execute", "--stdin", "--schema", SP.fromAbsFile schema]
+  runPrismaCommandAsJobWithServerEnv waspProjectDir projectRootDir ["db", "execute", "--stdin", "--schema", SP.fromAbsFile schema]
   where
+    waspProjectDir = projectRootDir </> waspProjectDirFromProjectRootDir
     schema = projectRootDir </> dbSchemaFileInProjectRootDir
 
 -- | Runs `prisma studio` - Prisma's db inspector.
@@ -174,24 +175,23 @@ generatePrismaClient projectRootDir =
 
 runPrismaCommandAsJobFromWaspServerDir :: Path' Abs (Dir ProjectRootDir) -> [String] -> J.Job
 runPrismaCommandAsJobFromWaspServerDir projectRootDir cmdArgs =
-  runPrismaCommandAsJobWithExtraEnv serverDir [] projectRootDir cmdArgs
+  runPrismaCommandAsJobWithServerEnv waspProjectDir serverDir cmdArgs
   where
-    -- We must run our Prisma commands from the server dir for Prisma
-    -- to pick up our .env file there like before.  In the future, we might want
-    -- to reconsider how Prisma and the server env vars interact and change
-    -- this. Text copied from: https://github.com/wasp-lang/wasp/pull/1662
     serverDir = projectRootDir </> serverRootDirInProjectRootDir
+    waspProjectDir = projectRootDir </> waspProjectDirFromProjectRootDir
+
+runPrismaCommandAsJobWithServerEnv :: Path' Abs (Dir WaspProjectDir) -> Path' Abs (Dir a) -> [String] -> J.Job
+runPrismaCommandAsJobWithServerEnv waspProjectDir fromDir cmdArgs chan = do
+  extraEnv <- readDotEnvServer waspProjectDir
+  runPrismaCommandAsJobWithExtraEnv fromDir extraEnv cmdArgs chan
 
 runPrismaCommandAsJobWithExtraEnv ::
   Path' Abs (Dir a) ->
   [(String, String)] ->
-  Path' Abs (Dir ProjectRootDir) ->
   [String] ->
   J.Job
-runPrismaCommandAsJobWithExtraEnv fromDir envVars projectRootDir cmdArgs =
-  runNodeCommandAsJobWithExtraEnv envVars fromDir (absPrismaExecutableFp waspProjectDir) cmdArgs J.Db
-  where
-    waspProjectDir = projectRootDir </> waspProjectDirFromProjectRootDir
+runPrismaCommandAsJobWithExtraEnv fromDir envVars cmdArgs =
+  runNodeCommandAsJobWithExtraEnv (envVars ++ [("PRISMA_HIDE_UPDATE_MESSAGE", "1")]) fromDir "npm" (["exec", "prisma", "--"] ++ cmdArgs) J.Db
 
 -- | NOTE: The expectation is that `npm install` was already executed
 -- such that we can use the locally installed package.
--- a/waspc/src/Wasp/Generator/DbGenerator/Operations.hs
+++ b/waspc/src/Wasp/Generator/DbGenerator/Operations.hs
@@ -176,10 +176,10 @@ testDbConnection genProjectDir = do
 prismaErrorContainsDbNotCreatedError :: T.Text -> Bool
 prismaErrorContainsDbNotCreatedError text = text TR.=~ ("\\bP1003\\b" :: String)
 
-isDbConnectionPossible :: DbConnectionTestResult -> Bool
-isDbConnectionPossible DbConnectionSuccess = True
-isDbConnectionPossible DbNotCreated = True
-isDbConnectionPossible _ = False
+isDbConnectionPossible :: DbConnectionTestResult -> Either (Chan J.JobMessage) ()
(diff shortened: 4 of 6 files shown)


waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
```diff
@@ -174,24 +175,23 @@ generatePrismaClient projectRootDir =
...
+  runPrismaCommandAsJobWithServerEnv waspProjectDir serverDir cmdArgs
   where
-    -- We must run our Prisma commands from the server dir for Prisma
-    -- to pick up our .env file there like before.  In the future, we might want
-    -- to reconsider how Prisma and the server env vars interact and change
-    -- this. Text copied from: https://github.com/wasp-lang/wasp/pull/1662
     serverDir = projectRootDir </> serverRootDirInProjectRootDir
+    waspProjectDir = projectRootDir </> waspProjectDirFromProjectRootDir
+
+runPrismaCommandAsJobWithServerEnv :: Path' Abs (Dir WaspProjectDir) -> Path' Abs (Dir a) -> [String] -> J.Job
```
Martinsos, line 163: Are we 100% sure this is the same behaviour as before?

Before, we would run prisma command from the .wasp/out/server dir, so it would pick up env vars from the .env file in that directory. That .env file was created by copying .env.server from the wasp project root into the .wasp/out/server.

Now, we run it from wherever (`fromDir`) and we pass it env vars that we read from the .env.server file ourselves.

When dealing with (dot) env vars, this is how precedence goes:
.env files < shell env vars < inline/cmd env vars.

But now you turned the values from .env.server into inline/cmd env vars, so they just jumped from the lowest precedence to the highest precedence.

Why does this matter? Because this way, we potentially get different behaviour than in some other situations (e.g. in `wasp start`). Also, is it not an expected behaviour. I am still building understanding of why you need this, the use case, but I would assume this could be problematic for that use case?

That is why we on purpose try to run Prisma in its "natural" environment, with an actual .env file next to it, so we "reimplement" as little of what Prisma does on its own.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
```diff
@@ -174,24 +175,23 @@ generatePrismaClient projectRootDir =
...
   [(String, String)] ->
-  Path' Abs (Dir ProjectRootDir) ->
   [String] ->
   J.Job
-runPrismaCommandAsJobWithExtraEnv fromDir envVars projectRootDir cmdArgs =
-  runNodeCommandAsJobWithExtraEnv envVars fromDir (absPrismaExecutableFp waspProjectDir) cmdArgs J.Db
-  where
-    waspProjectDir = projectRootDir </> waspProjectDirFromProjectRootDir
+runPrismaCommandAsJobWithExtraEnv fromDir envVars cmdArgs =
+  runNodeCommandAsJobWithExtraEnv (envVars ++ [("PRISMA_HIDE_UPDATE_MESSAGE", "1")]) fromDir "npm" (["exec", "prisma", "--"] ++ cmdArgs) J.Db
```
Martinsos: Why did you do this change (using `npm exec`) only here, why do you continue to use direct path to the prisma binary in other places in this file?

Besides that, while I like the idea of using `npm exec` instead of us directly specifying absolute path, I am not 100% if that is OK in our case. Because we knew about `npm exec` (ok, well certainly we knew about `npx` at least), but we didn't use it. Why? (https://en.wiktionary.org/wiki/Chesterton%27s_fence). I can't remember if there was a specific reason. There are no comments on that in the code? Maybe at the place of the definition of those abs paths? Aha I see it below, it says we assume `npm install` happened already. One thing that comes to my mind is that we had this problem before where if they didn't have prisma installed corerctly, but there was prisma installed for some reason in a higher level dir, somewhere above the project itself, maybe even globally on the machine, Wasp would pick up that prisma, and users would end up with errors quite hard to debug (usually due to version mismatch and what not). Maybe that was one reason why we want with such an absolute path.

@sodic do you remember the motivation behind this?

Martinsos: How is it that you had to re-add this PRISMA_HIDE_UPDATE_MESSAGE env var here? I remember we put it to 1 some time ago, but I think we do it in a central place somewhere. Yeah, I think we do it for this whole process we are running in Haskell, so it gets inherited to any other subprocesses of it. So we didn't have to specify it again in other places. But you had to do it here? Somehow this is not a subprocess and doesn't inherit that env var from this process?

If we have to do it separately here, I would be tempted a bit to look into potentially de-duping this with that other usage we have. To some degree maybe, if that makes sense at all.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
```diff
@@ -174,24 +175,23 @@ generatePrismaClient projectRootDir =
...
-    -- to pick up our .env file there like before.  In the future, we might want
-    -- to reconsider how Prisma and the server env vars interact and change
-    -- this. Text copied from: https://github.com/wasp-lang/wasp/pull/1662
     serverDir = projectRootDir </> serverRootDirInProjectRootDir
+    waspProjectDir = projectRootDir </> waspProjectDirFromProjectRootDir
+
+runPrismaCommandAsJobWithServerEnv :: Path' Abs (Dir WaspProjectDir) -> Path' Abs (Dir a) -> [String] -> J.Job
+runPrismaCommandAsJobWithServerEnv waspProjectDir fromDir cmdArgs chan = do
+  extraEnv <- readDotEnvServer waspProjectDir
+  runPrismaCommandAsJobWithExtraEnv fromDir extraEnv cmdArgs chan
```
Martinsos:
```suggestion
  serverEnv <- readDotEnvServer waspProjectDir
  runPrismaCommandAsJobWithExtraEnv fromDir serverEnv cmdArgs chan
```

Small thing, but in cases like this, it makes more sense to me call stuff what it is, not what it is being passed as.

waspc/src/Wasp/Generator/DbGenerator/Operations.hs
```diff
@@ -155,28 +156,30 @@ testDbConnection ::
   Path' Abs (Dir ProjectRootDir) ->
   IO DbConnectionTestResult
 testDbConnection genProjectDir = do
-  chan <- newChan
-  exitCode <- DbJobs.dbExecuteTest genProjectDir chan
+  chanForInspecting <- newChan
+  chanForPrinting <- dupChan chanForInspecting
```
Martinsos: Ok, so the idea is that we enrich the DbConnectionFailure with the output from the "test" process.

In that case, I wouldn't neccessarily call this `chanForPrinting`, as you don't really know what the consumer of `DbConnectionFailure` is going to do with it, not from here. So we should name it more generally. Maybe just `chanCopy` is good enough, it is very clear what it is about.

Sucks that we have to copy it here, and can't do the copying below, closer to place where it is used. But I just read that `dupChan` starts empty, so yeah we have to duplicate it here.

Btw, there are stronger concurrency primitives in Haskell, as part of the STM package, which bring stuff like TChan and TVar, which are like more advanced versions of `Chan` and `MVar` and similar. Those allow duplicating the channel with the content in it already and so on. But when I was implementing this code I didn't know enough about it to be brave enough to use it (these days maybe I would). Just a thought for the future, that we might want to switch to TChans at some point.

Ok one more thought: Do you need at all to create a duplicate channel? Why don't you just put `outputLines` into `DbConnectionFailure`, isn't that simpler? If the answer is becuase you can then use `readJobMessagesAndPrintThemPrefixed`, I would ask isn't there a simpler function there that can print text prefixed if it is just lines? If not, we should possibly extract one from `readJobMessagesAndPrintThemPrefixed` or at least make it next to it with extracted important pieces of logic.

waspc/cli/src/Wasp/Cli/Command.hs
```diff
@@ -68,3 +71,6 @@ require =
       req <- checkRequirement
       Command $ modify (Requirement req :)
       return req
+
+class InDirRequirable a where
```
Martinsos: Just looking at this class here, it is a bit hard to figure out what it is about.
The name and location in this file imply connection with the `Requireable` above, but that is it. No comments or types or anything further provide that connection.

I wonder if it makes sense to do two things:
1. Also add `Typeable` restriction to it. You do need it anyway later, you just put it more below, and this way it is clearer it is similar to the `Requirable`.
2. Add comment to it explaining what this method it has is supposed to do. That is normal for typeclasses, you usually want to explain the expected beahviour because it is often too hard to capture it at the type level. And this will also make it look much more similar to `Requirable`.

Also, maybe this definition should be moved down there, next to `instance (Typeable req, InDirRequirable req) => Requirable (InOutDir req) where` and similar, as there is more context provided there, and it kind of starts making sense there once you see it's instance of `Requirable`.

waspc/cli/src/Wasp/Cli/Command/Require.hs
```diff
@@ -36,27 +38,25 @@ import Data.Maybe (fromJust)
...
+import Wasp.Job.IO (readJobMessagesAndPrintThemPrefixed)
 import Wasp.Project.Common (WaspProjectDir)
 import qualified Wasp.Project.Common as Project.Common
 
 data DbConnectionEstablished = DbConnectionEstablished deriving (Typeable)
 
-instance Requirable DbConnectionEstablished where
-  checkRequirement = do
-    -- NOTE: 'InWaspProject' does not depend on this requirement, so this
-    -- call to 'require' will not result in an infinite loop.
```
Martinsos, line 49: Did you drop the comment on purpose? I guess we can do without it, but it is not so bad, maybe I would keep it.

waspc/cli/src/Wasp/Cli/Command/Require.hs
```diff
@@ -100,3 +100,25 @@ instance Requirable InWaspProject where
...
+data InOutDir req = InOutDir req deriving (Typeable)
+
+instance (Typeable req, InDirRequirable req) => Requirable (InOutDir req) where
+  checkRequirement = do
+    InWaspProject waspProjectDir <- require
+    let buildDir =
+          waspProjectDir
+            SP.</> Project.Common.dotWaspDirInWaspProjectDir
+            SP.</> Project.Common.generatedCodeDirInDotWaspDir
+    InOutDir <$> checkRequirementInDir buildDir
```
Martinsos: I think this is copy/paste mistake, it is all about build dir here, but you probably want it to be about out dir?

Also, I would pop out that `buildDir` variable into `where`, so it doesn't break the reading flow, give it a try. I see it you just kept how it was before, but it is nice opporutnity to improve that.

Martinsos (commented):
Commented! Was a bit complex because it mixes multiple changes at the same time (although connected), but the fact that you split it into commits helped a bit.

One general comment on the Requirable types that you added: that is quite cool, and nice job pulling of that in Haskell! It is quite complex though, it took me a bit to figure out how it works / what is happening. I think it is legit, it makes sense (left some comments), but I am also wondering if there is a simple way to go about it.

One very simple idea is that you simply could have replaced `data DbConnectionEstablished` with two data types: `data DbConnectionEstablishedFromOutDir` and `DbConnectionEstablishedFromBuildDir` and each one is instance of `Requirable` and that is it, they would probably share big part of their implementation via a shared helper function. This way you would avoid all the extra (type) machinery.

Another idea is that you can turn what you did currently a bit around. Right now, you made a class for the "argument", where "Dir" is that argument. If you needed more arguments in the future, for these or other "requirables", you would need more of this classes per "argument" and I can imagine that getting messy quite fast.
But if you turned it around to focus on the "command", and not the "argument", you could end up with something like this, which I think would scale better / is a naturally better fit for the problem here (and is also quite simpler):
```hs
data OutDir
data BuildDir

data DbConnectionEstablishedFrom fromDir = DbConnectionEstablished

instance Requireable (DbConnectionEstablishedFrom OutDir) where
  checkRequirement = ...

instance Requireable (DbConnectionEstablishedFrom BuildDir) where
  checkRequirement = ...
  
...

  DbConnectionEstablishedFrom OutDir <- require
  DbConnectionEstablishedFrom BuildDir <- require
```
NOTE: I haven't actually tried coding this last approach, I merely sketched it here, so there is a chance I am missing some reason why this wouldn't work.
TIP: Also, you can make this a bit nicer by using DataKinds and having `data GeneratedDir = OutDir | BuildDir` (not sure about the naming though. Also, I am wondering now, don't we already have these types for OutDir and BuildDir in the codebase somewhere, so we can just use them instead of redefining them here again?). But this is an extra thing, we can play with that after this works out.

Yeah, I would actually vote for trying out this last approach, if it works, I think it is the nicest solution.

cprecioso:
@Martinsos 

> One very simple idea is that you simply could have replaced data DbConnectionEstablished with two data types

Yes, that was my initial approach, but I felt that I was missing something, Haskell should allow me to parametrize this check. That's why I went on this typed adventure.

---

> ```
> data OutDir
> data BuildDir
> data DbConnectionEstablishedFrom fromDir = DbConnectionEstablished
> instance Requireable (DbConnectionEstablishedFrom OutDir) where
> ```

I had already tried it, but it doesn't work:

```
• Illegal instance declaration for
    ‘Requirable (DbConnectionEstablished OutDir)’
    (All instance types must be of the form (T a1 ... an)
     where a1 ... an are *distinct type variables*,
     and each type variable appears at most once in the instance head.
     Use FlexibleInstances if you want to disable this.)
• In the instance declaration for
    ‘Requirable (DbConnectionEstablished OutDir)’typecheck
```

It asks me to enable `FlexibleInstances`, but I don't know the ramifications of that so I didn't do it.

---

> TIP: Also, you can make this a bit nicer by using DataKinds and having `data GeneratedDir = OutDir | BuildDir`

Mmm I didn't know about this. I am now reading about DataKinds and I'm trying to wrap my head around it, it's a bit confusing. I will try it.

Martinsos:
> @Martinsos
> 
> > One very simple idea is that you simply could have replaced data DbConnectionEstablished with two data types
> 
> Yes, that was my initial approach, but I felt that I was missing something, Haskell should allow me to parametrize this check. That's why I went on this typed adventure.
> 
> > ```
> > data OutDir
> > data BuildDir
> > data DbConnectionEstablishedFrom fromDir = DbConnectionEstablished
> > instance Requireable (DbConnectionEstablishedFrom OutDir) where
> > ```
> 
> I had already tried it, but it doesn't work:
> 
> ```
> • Illegal instance declaration for
>     ‘Requirable (DbConnectionEstablished OutDir)’
>     (All instance types must be of the form (T a1 ... an)
>      where a1 ... an are *distinct type variables*,
>      and each type variable appears at most once in the instance head.
>      Use FlexibleInstances if you want to disable this.)
> • In the instance declaration for
>     ‘Requirable (DbConnectionEstablished OutDir)’typecheck
> ```
> 
> It asks me to enable `FlexibleInstances`, but I don't know the ramifications of that so I didn't do it.
> 
> > TIP: Also, you can make this a bit nicer by using DataKinds and having `data GeneratedDir = OutDir | BuildDir`
> 
> Mmm I didn't know about this. I am now reading about DataKinds and I'm trying to wrap my head around it, it's a bit confusing. I will try it.

So my knowledge of type shenannigans in Haskell is actually not super rich, for example I never used type families and more advanced type level programming. But, FlexibleInstances, if you google it, is not a big deal, it just allows defining instances for stuff like `Maybe String` instead of only `Maybe a`. Allows you to make these more specific instances. From what I read, it is quite common to use. There are some scenarios where it makes life harder for GHC, if you are doing some more advanced stuff, but nothing we should worry about. So I would say go for it, it will be a nice solution, simpler and on point.

DataKinds are a bit of a mindfuck, even though they are actually quite simple (just weird design). But they are not essential at all for this improvement here, I would say don't go into them till you have it working without, to not get entangled in the wrong direction.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
sodic: I don't remember exactly, but I dug something out.

It turns out we did use `npx` at one point, and we then changed it to this. We never used `npm exec`, but I think it would suffer from the same problem.

Anyway, we changed `npx` to the exact path because of this bug: https://github.com/npm/cli/issues/5018
Shayne explains it in this thread: https://github.com/wasp-lang/wasp/pull/661#discussion_r915255446.
Here's the [exact change](https://github.com/wasp-lang/wasp/commit/d5a90397229754c15d0ea037eb6b1ebff51b7ab1#diff-b24716d343af17daf33f7e342c7d16fc269f712a03efff342a583f52fd3b3ddfR23) (same PR).

Judging by their bug report, [Npm fixed the problem](https://github.com/npm/cli/issues/5018#issuecomment-1204209178) after we merged our fix. So we probably kept it by accident.

That said, it's still possible that this change breaks something. We've had a lot of iterations on where we keep prisma, the command's working dir, and how does two interact. So proceed with caution :)

### FYI for finding stuff like this

When I see something weird and the offending code has been moved around a lot, this is what I sometimes do (and what I did now):
1. I find a specific string I'd expect to see in the first occurence of this change (e.g., `The expectation is`)
2. I find the first commit that mentions the string with:
    ```bash
    git log --reverse -S 'The expectation is' -- src/Wasp/Generator/DbGenerator/Jobs.hs
    ```
3. I check that [commit on GitHub](https://github.com/wasp-lang/wasp/commit/d5a90397229754c15d0ea037eb6b1ebff51b7ab1) to see whether it's the one I want
4. Check the changes and the PR comments

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
cprecioso: Main problem is that I want to run this command from `WaspProjectDir: .wasp/out` instead of `WaspProjectDir: .wasp/build`, and `.wasp/out` doesn't have the `.env.server` -> `.env` copy (indeed, since we might not want to embed those values into the Docker build).
So I do need to parse the .env file myself and pass it over to Prisma.

> .env files < shell env vars < inline/cmd env vars
Yes. Though Shell + Inline is already consolidated by the shell but the time it reaches us. But I agree that .env should have lower priority. I had it like that but then refactored and lost it 🤦‍♂️

It does seem though that I'll end up making many different functions to deal with really small differences in calling 🫠

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
cprecioso: I think `npm exec --no-install` should work okay for our purposes, and I feel better using the standard interface to get to the npm binaries than something not guaranteed (Yarn PnP doesn't install stuff on `node_modules/.bin`, but that's also not a setup I expect we'll ever support). **BUT** I agree that it is probably not needed, I'll see if I can make it work without changing that.

> why do you continue to use direct path to the prisma binary in other places in this file?

I didn't want to touch the messy `migrateDev` job, with the quoting and the `script` utility, but I did end up working on it: https://github.com/wasp-lang/wasp/pull/2841 And now both calls of Prisma executable can be consistent.

cprecioso: You are right, I didn't see that it had already been declared at https://github.com/wasp-lang/wasp/blob/c326246a4fe2174c7bd29cd5c2a16b4ee6d71d1e/waspc/cli/exe/Main.hs#L148 I removed it from here

waspc/cli/src/Wasp/Cli/Command/Require.hs
cprecioso: I just didn't see that it was really needed anymore, but I'll put it back

waspc/src/Wasp/Generator/DbGenerator/Operations.hs
cprecioso: I applied the naming changes 

---

> Ok one more thought: Do you need at all to create a duplicate channel? Why don't you just put outputLines into DbConnectionFailure, isn't that simpler?

Yeah, indeed I wanted to use `readJobMessagesAndPrintThemPrefixed`

> we should possibly extract

It seems like the `PrefixedWriter` is a stateful io monad. I will not for the moment work on extracting that, otherwise this task will keep expanding in time 😅

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
cprecioso: I did get it to work without changing to `npm exec`

waspc/cli/src/Wasp/Cli/Command.hs
cprecioso: This code is all gone in the new version

waspc/cli/src/Wasp/Cli/Command/Require.hs
cprecioso: Fixed the copy/paste

---

> Also, I would pop out that buildDir variable into where 

Can't do, it needs the `waspProjectDir` from the line before

cprecioso:
I know that we talked about DataKinds and advanced stuff, but to be honest I don't think the extra complexity is worth it to just save us from a bit of verbosity, so in the end I just did two different `Requirable`s: `DbConnectionEstablishedFromBuildDir` and `DbConnectionEstablishedFromOutDir`, with the common part extracted. I guess this is "Boring Haskell" now? 😛

---

I will fix the CI errors and you can review then

Martinsos:
> I know that we talked about DataKinds and advanced stuff, but to be honest I don't think the extra complexity is worth it to just save us from a bit of verbosity, so in the end I just did two different `Requirable`s: `DbConnectionEstablishedFromBuildDir` and `DbConnectionEstablishedFromOutDir`, with the common part extracted. I guess this is "Boring Haskell" now? 😛
> 
> I will fix the CI errors and you can review then



> I know that we talked about DataKinds and advanced stuff, but to be honest I don't think the extra complexity is worth it to just save us from a bit of verbosity, so in the end I just did two different `Requirable`s: `DbConnectionEstablishedFromBuildDir` and `DbConnectionEstablishedFromOutDir`, with the common part extracted. I guess this is "Boring Haskell" now? 😛
> 
> I will fix the CI errors and you can review then

It is indeed :D. And sometimes that is the best. I would probably not bother with DataKinds, but would probably bother with FlexibleInstances. But if this works, great, that change is minor anyway, we can play with that later possibly.

OK, I am waiting for you re-request my reivew, then I will review it again! Btw is this going to stay in draft forever :D?

cprecioso:
I'm closing this PR, since right now it is getting very complicated to connect all the missing pieces. `.wasp/build` doesn't have the .env or the built Prisma schema, and I think this change needs more effort than what is worth. When I fixed one thing other appeared :/

Martinsos:
> I'm closing this PR, since right now it is getting very complicated to connect all the missing pieces. `.wasp/build` doesn't have the .env or the built Prisma schema, and I think this change needs more effort than what is worth. When I fixed one thing other appeared :/



> I'm closing this PR, since right now it is getting very complicated to connect all the missing pieces. `.wasp/build` doesn't have the .env or the built Prisma schema, and I think this change needs more effort than what is worth. When I fixed one thing other appeared :/

Oh no!
What do you mean by being very complicated to connect all the missing pieces?
From what I got, this PR was complex because it was solving multiple things at the same time, so it became hard to track it all, figure out what is connected and what not.
But all of the changes seemed to me like they are making a real improvement.
Would it make sense to instead create smaller PRs that solve those individual parts? And if one of those parts is not worth solving right we can skip that one (I think you said something with .env in build dir).

cprecioso:
The only real improvement this specific PR brought was, in the end, being able to run the DBConnectionEstablished check from the build dir in addition to the out dir. Only in order to show a specific error message that gets anyway shown through Prisma immediately after.

But to gain that (apart from the Haskell wrangling, which was not a problem), we would have to output the `.env.server` into `.wasp/build/.env`, which would then get embedded into every build, even if the envs where different. We would also have to install Prisma and generate the client from the command, while that gets actually done inside the docker build. 

That's what I refer to many missing pieces to connect that might not be worth it. Thanks to closing this I'm about done with https://github.com/wasp-lang/wasp/pull/2796, about to send actually.

I don't think there's anything important in this PR worth rescuing now.

Martinsos:
> I don't think there's anything important in this PR worth rescuing now.

Ok got it, makes sense then!

PR #2821, opened 2025-06-04
Update auth pills new

### LINKED ISSUE #2775: Simplity auth pills in the docs

zkmrtn8:
### Description
Simplified the auth pills to be a single color. Works on both light and dark mode with the same styling.

Examples:

image

image

All instances of the pills on the `next` version were updated, and I've verified that the actual content of the pages are not changed other than the pill order in a few spots.

No longer needs separate components for each auth provider. Just use `<AuthPills providers={['github','username','email','keycloak']} />`. The component automatically inserts commas and "and"

Fixes #2775

**New version of #2818 with problems fixed. It was easier to basically start from scratch than it was to continue that branch as there were a lot of requested changes.**


web/docs/auth/AuthPills.css
cprecioso, line 1: todo: newline at the end of file

web/docs/auth/AuthPills.css
```diff
@@ -0,0 +1,13 @@
+.display-inline {
```
cprecioso, line 1: Maybe something more semantic, like `.pill-list`

web/docs/auth/AuthPills.css
```diff
@@ -0,0 +1,13 @@
...
+  display: inline;
+}
+
+.pills {
```
cprecioso, line 5: This would be a single `pill`

```suggestion
.pill {
```

web/docs/auth/AuthPills.jsx
```diff
@@ -0,0 +1,50 @@
...
+    >
+      {label}
+    </Link>
+  );
+}
+
+export function AuthPills({ providers }) {
+  const validPills = providers
+    .map((provider) => PILL_INFO[provider])
+    .filter(Boolean);
```
cprecioso, line 32: From #2818

> Let's make this a bit more robust and resistant to mis-use: instead of silently ignoring bad provider names, throw an error if the provider name doesn't exist!

web/docs/auth/AuthPills.css
```diff
@@ -0,0 +1,13 @@
...
+  padding: 0.1rem 0.5rem;
+  border-radius: 0.375rem;
+  color: white;
+  background-color: #444;
+  text-decoration: none;
+  display: inline-block;
+  margin-right: 0.25rem;
```
cprecioso, line 12: From #2818:

> Why do we add the `marginRight` here? It makes it look like there's a space between the pill and the comma:
> 
> Before
> image
> 
> After
> image

---

> fixed in new pr, https://github.com/wasp-lang/wasp/pull/2821. it needs a bit of margin so the comma is separated from the pill, but I've made it a more reasonable number now.

This is still outstanding though

web/docs/auth/AuthPills.jsx
```diff
@@ -0,0 +1,50 @@
...
+    <span className={'display-inline'}>
+      {validPills.map((pill, index) => {
+        const isLast = index === validPills.length - 1;
+        const isSecondLast = index === validPills.length - 2;
+
+        return (
+          <React.Fragment key={pill.link}>
+            {isLast && validPills.length > 1 && "and "}
+            <Pill label={pill.label} link={pill.link} />
+            {!isLast && (isSecondLast ? ", " : ", ")}
```
cprecioso, line 44: From #2818 

> Do we need the isSecondLast variable? It's not being used.

web/docs/auth/auth-hooks.md
```diff
@@ -103,7 +103,7 @@ Wasp calls the `onBeforeSignup` hook before the user is created.
 
 The `onBeforeSignup` hook can be useful if you want to reject a user based on some criteria before they sign up.
 
-Works with <EmailPill /> <UsernameAndPasswordPill /> <SlackPill /> <DiscordPill /> <GithubPill /> <GooglePill /> <KeycloakPill />
+Used with <AuthPills providers={['email','username','slack','discord','github','google','keycloak']} /> authentication.
```
cprecioso, line 106: I'd prefer the "Works with" back (here and everywhere else in this file)

web/docs/auth/auth-hooks.md
```diff
@@ -284,7 +284,7 @@ The `onAfterEmailVerified` hook is useful for triggering actions in response to
 
 The `onAfterEmailVerified` hook receives an `email` string and `user` object, this makes it easy to perform personalized actions upon email verification.
 
-Works with <EmailPill />
+Works with Used with <AuthPills providers={['email']} /> authentication.
```
cprecioso, line 287:
```suggestion
Works with <AuthPills providers={['email']} />
```

cprecioso (changes requested):
Okay, thanks for rebasing! I'm leaving here some comments that weren't resolved from the previous PR

infomiho:
Quick drive-by style idea I had when I wrote the task, just use the `currentColor` as the text and border color:

### Screenshots
Screenshot 2025-06-05 at 14 46 31
Screenshot 2025-06-05 at 14 46 34
Screenshot 2025-06-05 at 14 46 51
Screenshot 2025-06-05 at 14 46 53

### Styles

```css
.pills {
  padding: 0.1rem 0.3rem;
  border-radius: 0.375rem;
  color: currentColor;
  border: 1px solid currentColor;
  font-size: 0.9rem;
  text-decoration: none;
  display: inline-block;
}
```

This way - it makes the pills less dominant like when they have a black background and they adopt to the context (like in the info collpase).

developerzeke:
Some of those issues (the ones that I marked resolved) were solved. Not sure how the changes got lost. 

This was supposed to be a quick pr to knock out another issue, but evidently it's not going to be. I personally don't see how a line at the end of the file or using inline vs separate styles is that important—and if it is, it should be clearly documented somewhere so that I could've known before submitting the pr.

Someone else can resolve the issue. I'm not going to.

infomiho:
@zkmrtn8 I'm glad you took a stab at this issue, it's great to see community members trying to solve issues and contribute, we appreciate your time :)

I understand you wanted to make this quick, and that is all right! The thing is, we approach all changes in the Wasp codebase with the same care, it doesn't matter if it's the compiler or the docs. There are certain code standards we follow and if we were to accept a contribution, code can't be in a state that would require some Wasp team member to clean up afterwards. We have to keep maintaining the code, which is why we care so much.

PR #2836, opened 2025-06-11
Improve our e2e tests DX

### LINKED ISSUE #2731: Make easy improvements to our e2e test process
This issue covers several e2e test-related improvements:
- We recently documented how a PR author should run e2e tests after getting the approval on the PR, but we should add more emphasis to this. One argument we forgot to mention is that, when you run e2e tests before the PR finishes all its iterations, you're less likely to check whether their final form is OK.
- We should make it clear that the PR author is the one responsible for checking e2e test outputs (not just updating them without looking). Enforcing this will help with the first point.
- We should stop checking some generated files, mostly those in `dist` directories. It creates a lot of noise and only tests the TS compiler. For example, one `foo.ts` file change changes 7 files in a single e2e test (`files.manifest`, `.waspchecksums`, `foo.ts`, `dist/foo.js`, `dist/foo.js.map`, `dist/foo.d.ts`, `dist/foo.d.ts.map`), and we don't care about most of them.

FranjoMindek:
Fixes #2731

Gives more explanation about e2e tests in our `waspc`'s `README.md`.
Reminds about e2e tests in PR template.
Removed `dist` folders from golden tests.

The changes can actually be seen in the CI pipeline.


waspc/e2e-test/GoldenTest.hs
FranjoMindek, line 1: I've only added `"dist"`, but I reordered the rest:
system
db
project
typescript

FranjoMindek:
~~TODO: explore if I can remove more stuff other than `"dist"`~~

EDIT: There are a lot of config, checksum and dotfiles, but they all seem like important pieces we want to track and notice when they change. I don't think there is anything more.

waspc/README.md
```diff
@@ -156,7 +156,7 @@ When done, new tab in your browser should open and you will see a Todo App!
    - NOTE: You will need to install `ghcid` globally first. You can do it with `cabal install ghcid`.
 3. Do a change in the codebase (most often in `src/` or `cli/src/` or `data/`), and update tests if that makes sense (see [Test](#tests)).
    Fix any errors shown by HLS/`ghcid`.
-   Rinse and repeat. If you're an internal team member, posptpone updating e2e tests until approval (see [here](#note-for-team-members)).
+   Rinse and repeat. If you're an internal team member, postpone updating e2e tests until approval (see [here](#note-for-team-members)).
```
sodic, line 159: How do you know I didn't mean to say "posptpone?"

waspc/README.md
```diff
@@ -168,7 +168,7 @@ When done, new tab in your browser should open and you will see a Todo App!
 
 ### Note for team members
 
-Don't update e2e tests until someone approves the PR. They slow down the UI and create noise for the reviewers.
+Don't update e2e tests until your PR is approved. Running them and accepting changes prematurely can slow down the review process, create noise for reviewers, and makes it less likely that you'll thoroughly check the final diffs if changes are made during review iterations. **Carefully review the diffs before accepting them** (see [End-to-end (e2e) tests](#end-to-end-e2e-tests) for more details).
```
sodic: Let's make it more authoritative and add spacing for emphasis.
I also corrected a minor mistake (e.g., review process -> UI)

```suggestion
Don not update e2e tests until your PR is approved.

Running the e2e tests and accepting them prematurely:
- Slows down the UI.
- Creates noise for the reviewers.
- Makes it less likely that you'll thoroughly check the final diffs after all review iterations.

**Carefully review e2e test diffs before accepting them** (see [End-to-end (e2e) tests](#end-to-end-e2e-tests) for more details).
```

waspc/README.md
```diff
@@ -277,7 +277,7 @@ To run Playwright/headless tests on the example apps, you can do `/run test:head
 Besides unit tests, we have e2e tests that run `waspc` on a couple of prepared projects, check that they successfully run, and also compare generated code with the expected generated code (golden output).
 
 This means that when you make a change in your code that modifies the generated code, e2e tests will fail while showing a diff between the new generated code and the previous (golden) one.
-This gives you an opportunity to observe these differences and ensure that they are intentional and that you are satisfied with them. If you notice something unexpected or weird, you have an opportunity to fix it.
+This gives you an opportunity to observe these differences and ensure that they are intentional and that you are satisfied with them. **It is the PR author's (or reviewers if its an outside contribution) responsibility to carefully review these diffs.** Do not blindly accept changes, ensure they align with your intended modifications. If you notice something unexpected or weird, you have an opportunity to fix it.
```
sodic:
```suggestion
This gives you an opportunity to observe these differences and ensure that they are intentional and that you are satisfied with them. **It is the PR author's (or the reviewers for outside contributions) responsibility to carefully review these diffs.** Do not blindly accept changes, ensure they align with your intended modifications. If you notice something unexpected or weird, you have an opportunity to fix it.
```

waspc/e2e-test/GoldenTest.hs
sodic: Ok, nice!

But now I'm wondering whether we should keep `dist`. For example, what if we break the compiler in a way that doesn't execute the TS build. Would we know it without the `dist`?

sodic (commented):
Great stuff!

I suggested some improvements and am having some second thoughts about the `dist`. We can discuss it in that thread.

waspc/README.md
FranjoMindek: 😆 , if you really want to say it can always revert 😉

waspc/e2e-test/GoldenTest.hs
FranjoMindek: Write in dist files in file manifest but don't check files content?

FranjoMindek: If applicable it could be applied to more files, where we just want to check existence.

sodic: Hmm, this is a good idea. Can we do it here?

waspc/e2e-test/GoldenTest.hs
```diff
@@ -79,27 +80,52 @@ runGoldenTest goldenTest = do
...
+    listFilesForManifest dirToFilterAbsFp =
+      getDirFiltered (return <$> shouldIncludeInManifest) dirToFilterAbsFp >>= filterM doesFileExist
+
+    -- Lists files that should be compared in golden tests.
+    listFilesForGoldenTest :: FilePath -> IO [FilePath]
+    listFilesForGoldenTest dirToFilterAbsFp =
+      getDirFiltered (return <$> shouldIncludeInGoldenTest) dirToFilterAbsFp >>= filterM doesFileExist
+
+    -- Determines if a file should be included in the `files.manifest`.
+    shouldIncludeInManifest :: FilePath -> Bool
```
FranjoMindek: @sodic 
`shouldIncludeInGoldenTest` and `shouldIncludeInManifest` now have a lot of duplication, but I don't want to extract anything out yet. Since we may want to change those two now.

We should discuss if there is anything else that we want to include in `files.manifest` but not actually compare in golden tests. In my opinion anything not OS-dependent can be tracked for existence. We should also see if we currently compare anything in golden tests, that we actually only want to test existence of. 

@Martinsos You might also know more about this part of `waspc`. I've split up the `isTestOutputFileTestable` into two separate categories - those which go to `files.manifest` and those which are tested in golden tests.

waspc/e2e-test/GoldenTest.hs
```diff
@@ -79,27 +80,52 @@ runGoldenTest goldenTest = do
...
-    listRelevantTestOutputFiles dirToFilterAbsFp =
-      getDirFiltered (return <$> isTestOutputFileTestable) dirToFilterAbsFp >>= filterM doesFileExist
-
-    isTestOutputFileTestable :: FilePath -> Bool
-    isTestOutputFileTestable fp =
+    fileManifestFileName :: String
+    fileManifestFileName = "files.manifest"
+
+    -- Lists files that should be included in the `files.manifest`.
+    listFilesForManifest :: FilePath -> IO [FilePath]
```
FranjoMindek: I mention `files.manifest` as `Manifest` only, I don't know if that is clear enough for other people. Or should we suffix it with `FilesManifest`?

Martinsos: I vote for filesManifest

waspc/e2e-test/GoldenTest.hs
Martinsos: @FranjoMindek I was involved into this code time ago, but I would need to do some context loading (heh) to be able to offer any insights at this moment. I can take a closer look at the PR if need be, but it will take me time, so up to you, if you need it let me know.
If there are more specific questions you think I can be helpful with, you can help me by providing me with more (a lot of) context, that will help me answer them much faster without having to dig on my own.

FranjoMindek: We do 2 different actions in our golden tests:
1) We enumerate all the files we visited in the `files.manifest` file
2) We `diff` the "golden files" with the "current files"

Before all of the files that were `diff`ed were also enumerated in the `files.manifest`
Now those are two separate actions which can be applied to different files.

For the files where content doesn't matter, we want to use 1), otherwise we want to use both 1) and 2).

Now we are trying to find out which exactly files fall under:
1) only `files.manifest`
2) only `diff`
2) both the `files.manifest` and `diff`

Previously we had these files in our "global ignore list":
```
".waspinfo",
"node_modules",
"dev.db",
"dev.db-journal",
"package-lock.json",
".gitignore",
".DS_Store",
```

But in my opinion all non-OS dependent files (and not external dependencies like `node_modules`)  should be in the 1), meaning I would remove all of these from the new "`files.manifest` ignore list":
```
".waspinfo",
"dev.db",
"dev.db-journal",
"package-lock.json",
".gitignore",
```
Leaving only this ignored: 
```
".DS_Store",
"node_modules",
```
For the 2) I would leave the ignore list the same + add the new `"dist"` folder. Which is checked by `files.manifest`.
Since we don't want to check content of those files.

sodic: Nice analysis @FranjoMindek!

Your idea makes sense to me. It makes our tests both more robust and easier to manage.

waspc/e2e-test/GoldenTest.hs
sodic: I'm resolving this since we're discussing it here: https://github.com/wasp-lang/wasp/pull/2836/files#r2149168012

waspc/e2e-test/GoldenTest.hs
sodic: If the other parts of code are calling it `manifestFile`, we can keep that here: `listFilesForManifestFile`. I think I prefer that.

But I prefer a more fundamental name change, see other comments for details.

waspc/e2e-test/GoldenTest.hs
```diff
@@ -79,27 +80,52 @@ runGoldenTest goldenTest = do
...
-                    ".DS_Store",
+                    ".waspinfo",
+                    "package-lock.json",
+                    "node_modules",
                     "tsconfig.tsbuildinfo"
                   ]
 
+    -- Determines if a file should be included in golden test comparisons.
+    shouldIncludeInGoldenTest :: FilePath -> Bool
+    shouldIncludeInGoldenTest fp =
```
sodic: These names come from the implementation (solution domain), not the intent (problem domain). I'm sure there's a section in Clean Code that talks about this, but I'm not sure where :)

Anyway, I think the names should reflect what we want to do: check for existance and compare contents. So something like:
```
shouldCheckFileContents fp = ...
shouldCheckIfFileExists fp = ...
``` 
Or something along those lines.

waspc/e2e-test/GoldenTest.hs
```diff
@@ -79,27 +80,52 @@ runGoldenTest goldenTest = do
...
+    manifestFileName = "files.manifest"
+
+    -- Lists files that should be included in the `files.manifest`.
+    listFilesForManifest :: FilePath -> IO [FilePath]
+    listFilesForManifest dirToFilterAbsFp =
+      getDirFiltered (return <$> shouldIncludeInManifest) dirToFilterAbsFp >>= filterM doesFileExist
+
+    -- Lists files that should be compared in golden tests.
+    listFilesForGoldenTest :: FilePath -> IO [FilePath]
+    listFilesForGoldenTest dirToFilterAbsFp =
```
sodic: Similar to my other comment: golden tests are a non-standard implementation name. A better name for this variable is `getFilesForCheckingContents` or something like that.

Same thing about `listFilesForManifest`.

I think we should adopt this naming scheme to the rest of the code too.

sodic (commented):
I love the idea and the approach, left some comments mostly on naming.

waspc/e2e-test/GoldenTest.hs
FranjoMindek: I think `FilesManifest` makes more sense, since `manifestFile` implies any manifest file, while `FilesManifest` implies the manifest we have for files. 
This file is only ever mentioned in `GoldenTest.hs`, but I'll use the full name since we already have confusion.
e.g. `listFilesForFilesManifest`

waspc/e2e-test/GoldenTest.hs
FranjoMindek: `getFilesForCheckingExistence` and `getFilesForCheckingContent` work.
I've also renamed all `filesManifest` mentions, since files manifest itself carries no meaning other than it holds some metadata about files. Golden tests make sense since it is a term used in testing, so I'm not sure where to draw the line when renaming to e.g. expected?

waspc/e2e-test/GoldenTest.hs
```diff
@@ -128,9 +144,9 @@ runGoldenTest goldenTest = do
             }
 
         reformatJson :: FilePath -> IO ()
-        reformatJson jsonFilePath =
-          BSL.writeFile jsonFilePath . AesonPretty.encodePretty' aesonPrettyConfig . unsafeDecodeAnyJson
-            =<< B.readFile jsonFilePath
+        reformatJson jsonFp =
+          BSL.writeFile jsonFp . AesonPretty.encodePretty' aesonPrettyConfig . unsafeDecodeAnyJson
+            =<< B.readFile jsonFp
```
sodic: I'm not a fan of these changes (`filePath` -> `fp`) because I dislike abbreviations.

I prefer the old code, but I won't insist on it. Keep what you like. But how come you changed it anyway?

waspc/e2e-test/GoldenTest.hs
```diff
@@ -36,20 +36,21 @@ makeGoldenTest :: String -> ShellCommandBuilder [ShellCommand] -> GoldenTest
...
 -- | This runs a golden test by creating a Wasp project (via `wasp-cli new`), running commands,
--- and then comparing all file outputs to the corresponding golden test output directory.
+-- and then comparing all file outputs to the corresponding expected test output directory.
 runGoldenTest :: GoldenTest -> IO TestTree
 runGoldenTest goldenTest = do
   testOutputsDirAbsSp <- getTestOutputsDir
   let testOutputsDirAbsFp = SP.fromAbsDir testOutputsDirAbsSp
   let currentOutputDirAbsFp = testOutputsDirAbsFp FP.</> (_goldenTestName goldenTest ++ "-current")
-  let goldenOutputDirAbsFp = testOutputsDirAbsFp FP.</> (_goldenTestName goldenTest ++ "-golden")
+  let expectedOutputDirAbsFp = testOutputsDirAbsFp FP.</> (_goldenTestName goldenTest ++ "-expected")
```
sodic: I think it's alright keeping the "golden" name here:
- This part of the code belongs to the solution domain so it's fine.
- We still use it in the rest of the code (e.g., `runGoldenTests`, `_goldenTestName`), so we should keep it consistent.
- Turns out it is kind of standard (I googled it, and sure enough, it's a thing).

So, my suggestion is to keep all the renaming except changing "golden" to "expected" (and possibly those abbreviations, but that's a different discussion).

waspc/e2e-test/GoldenTest.hs
sodic: > Golden tests make sense since it is a term used in testing, so I'm not sure where to draw the line when renaming to e.g. expected?

This was a correct concern, I've written more about it in a different comment. When talking about the files and folders, we can use "golden." When talking about intent (e.g., checking contents), we should express it.

waspc/e2e-test/GoldenTest.hs
FranjoMindek: Most of the code used `fp`/`fps` while only a smaller portion used `filePath`/`filePaths`.
I wanted it to be more consistent. 

Alternative is to go full verbose everywhere - `filePath`/`filePaths`.

But I don't like the any those two really.
Maybe to just leave it as it is.

waspc/e2e-test/GoldenTest.hs
FranjoMindek: Makes sense

sodic (approved):
This looks good to me! Just make sure to test it and we can merge. I haven't run it

waspc/e2e-test/GoldenTest.hs
Martinsos: Also makes sense to me, and thanks for the nice explanation!

OK, so basically before we were either ignoring files, or we were checking them for changes in existence and content.
Now, we can be more granular and can check either for changes in existence or content or both.
So you are asking if some of those we ignored before, we can now check for just changes in existence (and no content).

I think that makes sense to me, yeah.

.waspinfo makes sense, we actaully want to confirm this one exists.
"dev.db" -> I guess hm! It is an artefact kind of, so not 100% sure.
"dev.db-journal", -> same
"package-lock.json", -> same
".gitignore", -> we actually want to confirm this one exists

Yeah ok we are not checking here if something exists, just changes in existence. Ok I guess it makes sense (although it is a tiny bit foggy for me on the dev.db and package-lock.json side. But better to add it and we can remove it than the opposite.

waspc/e2e-test/GoldenTest.hs
Martinsos: > I think `FilesManifest` makes more sense, since `manifestFile` implies any manifest file, while `FilesManifest` implies the manifest we have for files. This file is only ever mentioned in `GoldenTest.hs`, but I'll use the full name since we already have confusion. e.g. `listFilesForFilesManifest`

Yes, exactly

PR #2841, opened 2025-06-16
Remove workaround for interactive Prisma

cprecioso:
- Part of #2820

Prisma needs shell interactivity for the `prisma migrate dev` command, so it will check if the terminal is a TTY and die if it's not.

Prisma used to check if _stdout_ is interactive, instead of _stdin_. We had a workaround for this, by calling the Prisma executable through the `script` utility.

This has been fixed since Prisma v4.3.0 (https://github.com/prisma/prisma/pull/14649), they now check for a TTY on stdin, which we have always bequeathed to it.

We can remove our workaround code, and close our issue (https://github.com/prisma/prisma/issues/7113).

### DIFF
--- a/waspc/src/Wasp/Generator/SdkGenerator.hs
+++ b/waspc/src/Wasp/Generator/SdkGenerator.hs
@@ -37,7 +37,6 @@ import qualified Wasp.Generator.DbGenerator.Auth as DbAuth
 import Wasp.Generator.DepVersions (prismaVersion, superjsonVersion)
 import Wasp.Generator.FileDraft (FileDraft)
 import qualified Wasp.Generator.FileDraft as FD
-import qualified Wasp.Generator.JsImport as GJI
 import Wasp.Generator.Monad (Generator)
 import qualified Wasp.Generator.NpmDependencies as N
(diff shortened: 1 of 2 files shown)


waspc/src/Wasp/Generator/SdkGenerator.hs
```diff
@@ -37,7 +37,6 @@ import qualified Wasp.Generator.DbGenerator.Auth as DbAuth
 import Wasp.Generator.DepVersions (prismaVersion, superjsonVersion)
 import Wasp.Generator.FileDraft (FileDraft)
 import qualified Wasp.Generator.FileDraft as FD
-import qualified Wasp.Generator.JsImport as GJI
```
cprecioso, line 40: This is unrelated, someone left an unused import here and the compiler was complaining.

cprecioso:
I did try it before sending the PR 😁
Screen Recording 2025-06-16 at 17 50 17

Martinsos:
Awesome! I was sure you did but I remember it not always asking for name so wanted to make sure. Nice stuff!

PR #2848, opened 2025-06-18
Update Wasp Studio (deps, Express, tsx)

### LINKED ISSUE #2774: `wasp studio` is broken on `main` in Firefox
The socket connection returns 404, we should check what's going on and fix it.

Image

infomiho:
This PR upgrade Wasp Studio:
- [x] Replaces `ts-node` with `tsx` (dev mode was broken, so I went with `tsx` which has a watch mode so I can remove `nodemon`)
- [x] Removes `nodemon` since it's no longer neded
- [x] Replaces Fastify with Express (we use Express in Wasp, so this reduces the amount of libs team members need to be aware of)
- [x] ~~Deletes the `studio/client/node_modules` after building to keep the binary size down~~ No longer the case
- [x] Bumps the dep versions

As a side-effect, all this fixed the issues with Wasp Studio not working in Firefox.    

Closes #2774


waspc/packages/studio/client/index.html
```diff
@@ -2,7 +2,6 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
-    <link rel="icon" type="image/svg+xml" href="/vite.svg" />
```
infomiho, line 5: `vite.svg` didn't exist, so this just ended up being a 404 in the network tab

waspc/packages/studio/package.json
```diff
@@ -8,28 +8,25 @@
   "main": "index.js",
   "keywords": [],
   "scripts": {
-    "build": "npm run build:client && rm -rf dist && tsc && cp -r ./public ./dist/public",
+    "build": "npm run build:client && rm -rf ./client/node_modules && rm -rf dist && tsc && cp -r ./public ./dist/public",
```
infomiho: This removes the client's `node_modules` (this should help with keeping the build size smaller) @sodic 

Related to https://github.com/wasp-lang/wasp/issues/2718

waspc/packages/studio/package.json
```diff
@@ -8,28 +8,25 @@
...
-    "dev": "nodemon ./src/index.ts",
-    "start": "tsc --noEmit && ts-node -r dotenv/config ./src/index.ts"
+    "dev": "tsx --watch ./src/index.ts",
+    "start": "node ./dist/index.js"
   },
   "dependencies": {
-    "@fastify/cors": "^8.3.0",
-    "@fastify/static": "^6.11.2",
-    "commander": "^11.0.0",
-    "dotenv": "^16.0.3",
```
infomiho, line 20: We didn't use `.env` files so I've removed `dotenv` as dep.

waspc/packages/studio/src/index.ts
```diff
@@ -19,21 +20,24 @@ const options = program.opts<{
...
   },
 });
-fastify.register(cors, {
-  origin: true,
-});
-fastify.register(FastifyStatic, {
-  root: new URL("./public", import.meta.url).pathname,
-});
+
+app.use(morgan("dev"));
```
infomiho: Logging requests

waspc/packages/studio/package.json
```diff
@@ -8,28 +8,25 @@
   "main": "index.js",
   "keywords": [],
   "scripts": {
-    "build": "npm run build:client && rm -rf dist && tsc && cp -r ./public ./dist/public",
+    "build": "npm run build:client && rm -rf ./client/node_modules && rm -rf dist && tsc && cp -r ./public ./dist/public",
     "build:client": "npm --prefix ./client install && npm --prefix ./client run copy",
-    "dev": "nodemon ./src/index.ts",
-    "start": "tsc --noEmit && ts-node -r dotenv/config ./src/index.ts"
+    "dev": "tsx --watch ./src/index.ts",
```
cprecioso, line 13: I'm fantasizing with updating to node24 and just do `node --watch ./src/index.ts` directly

waspc/packages/studio/package.json
```diff
@@ -8,28 +8,25 @@
...
   "dependencies": {
-    "@fastify/cors": "^8.3.0",
-    "@fastify/static": "^6.11.2",
-    "commander": "^11.0.0",
-    "dotenv": "^16.0.3",
-    "fastify": "^4.23.2",
-    "fastify-socket.io": "^4.0.0",
-    "socket.io": "^4.7.2"
+    "@tsconfig/node20": "^20.1.6",
+    "@types/node": "^24.0.3",
```
cprecioso: Since we're using node 20...

```suggestion
    "@types/node": "^20",
```

waspc/packages/studio/src/index.ts
```diff
@@ -19,21 +20,24 @@ const options = program.opts<{
...
-  logger: true,
-});
-
-fastify.register(FastifySocketIO, {
+const app = express();
+const server = createServer(app);
+const io = new Server(server, {
   cors: {
-    origin: "*",
+    origin: true,
```
cprecioso: Just curious, what's the difference?

waspc/packages/studio/src/index.ts
```diff
@@ -43,24 +47,23 @@ function readFile() {
...
 try {
-  await fastify.listen({ port: 4000 });
+  server.listen(4000, () => {
+    console.log("Server listening on port 4000");
+  });
 } catch (err) {
-  fastify.log.error(err);
+  console.error(err);
   process.exit(1);
 }
```
cprecioso: Why don't we just let it throw instead of `process.exit`? It's going to exit anyway

waspc/packages/studio/README.md
```diff
@@ -18,13 +18,13 @@ npm install
...
 ```bash
-npm run dev -- -- -d <path_to_data_file>
+npm run dev -- -d <path_to_data_file>
 ```
 
 For example, running the server with the data file from the `examples` directory:
 
 ```bash
-npm run dev -- -- -d ../../examples/todoApp/.wasp/out/.wasp-studio-data.json
+npm run dev -- -d ../../examples/todoApp/.wasp/out/.wasp-studio-data.json
```
cprecioso, line 27: Slightly out of scope but can we add here how to get this json file created?

waspc/packages/studio/src/index.ts
infomiho: There isn't any AFAIK, it's the same meaning. I've reverted this change to make it look less weird :)

waspc/packages/studio/package.json
sodic: If I got everything right, I don't think this is the way to go.

We already have a script that is supposed to remove `node_modules` in packages, but it's broken for this specific case. Still, that script is what's in charge of removing all other `node_modules` for all other packages, so I think we should fix it instead of putting that logic here.

waspc/packages/studio/tsconfig.json
```diff
@@ -1,14 +1,8 @@
 {
-  "extends": "@tsconfig/node18/tsconfig.json",
+  "extends": "@tsconfig/node20/tsconfig.json",
```
sodic: I believe we've recently decided that our new packages won't inherit from `@tsconfig/nodeXX` (e.g., wasp app runner). Should we do the same here in this PR?

sodic (commented):
I'm not really familiar with the code and I haven't run it, but the changes look reasonable.

I have two comments though :)

waspc/packages/studio/tsconfig.json
infomiho: I was not aware of that 😄 It's always good to write that kind of conclusions in Discord.

Lemme check what's up in the Wasp App Runner :)

waspc/packages/studio/package.json
infomiho: Makes sense, I'll remove the change then 👍

waspc/packages/studio/tsconfig.json
infomiho: I've inlined the values from `@tsconfig/node20/tsconfig.json` now 👍

sodic (approved):
Good to go!

PR #2852, opened 2025-06-20
Add security incident blog post

infomiho:


web/blog/2025-06-20-oauth-security-incident.md
```diff
@@ -0,0 +1,101 @@
+---
+title: 'Incident Report: Case insensitive OAuth IDs  vulnerability in Wasp'
```
Martinsos:
```suggestion
title: 'Incident Report: Case insensitive OAuth IDs vulnerability in Wasp'
```

web/blog/2025-06-20-oauth-security-incident.md
```diff
@@ -0,0 +1,101 @@
...
+authors: [miho]
+image: /img/security-incident/security-incident.png
+tags: [security, oauth, keycloak, wasp]
+---
+
+## Introduction
+
+We received a report on a security vulnerability in Wasp auth related to our OAuth support in Wasp `0.16.5` and earlier.
```
Martinsos: I like how we did in the last one:

> On Feb 12th, 2025, we learned about a security vulnerability in the updateCurrentUser function of our Open SaaS template. Users of apps built with Open SaaS can exploit this vulnerability to edit any field in their own User database record, including fields they weren’t supposed to have write permissions for, like subscriptionPlan and isAdmin.

Maybe rewrite it a bit with that as inspiration. I think the very start is a nice format: "On <date>, we learned about ...".

web/blog/2025-06-20-oauth-security-incident.md
```diff
@@ -0,0 +1,101 @@
...
+    providerUserId: providerUserId.toLowerCase(), // <--- incorrect behavior
+  }
+}
+```
+
+## Applied fix
+
+Wasp has released version `0.16.6` which includes a fix for the vulnerability - Wasp no longer lowercases user IDs received from OAuth providers, only the `email` and `username` user IDs.
+
+All users should upgrade to Wasp `0.16.6`, especially if they use Keycloak auth with a configuration that makes the user IDs case sensitive.
```
Martinsos:
```suggestion
All users should upgrade to Wasp `0.16.6`, especially if they use Keycloak auth with a configuration that makes the user IDs case sensitive. 
```
All users should upgrade, especially .. -> makes it sound like everybody it affected to some degree, but they are not, maybe rephrase somehow a bit?

web/blog/2025-06-20-oauth-security-incident.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Introduction
+
+We received a report on a security vulnerability in Wasp auth related to our OAuth support in Wasp `0.16.5` and earlier.
+
+**This issue only affects users using Keycloak authentication with non-default case-sensitive IDs.**
+
+All other configurations (Google, GitHub, Discord, or Keycloak with the default case-insensitive configuration) are **not affected**. Email and username auth providers are also **not affected**.
+
+Users should upgrade to Wasp `0.16.6` which contains the fix ASAP.
```
Martinsos: They will also need to migrate their data in the database, correct? Since provider ids were saved lowercased so far? Or hm they can't even migrate it, because there is information loss -> they can't "unlowercase it". What should they do then? What is with the currently saved accounts in the database, that are using Keycloak with case sensitive identifiers, what should they doa bout them? I would hint about it here, and give some more pointers down in the solution.

PR #2853, opened 2025-06-23
Rename `DbConnectionEstablished` to `DbConnectionEstablishedFromOutDir`

cprecioso:
Based on https://github.com/wasp-lang/wasp/pull/2796#discussion_r2154569434

- Moved the `DbConnectionEstablished` requirement to its own module (this was mainly done to accommodate the large comment)
  ```diff
  -Wasp.Cli.Command.Require
  +Wasp.Cli.Command.Require.Db
  ```

- Renamed it
  ```diff
  -DbConnectionEstablished
  +DbConnectionEstablishedFromOutDir
  ```

- Added a comment explaining why no `DbConnectionEstablishedFromBuildDir`


waspc/cli/src/Wasp/Cli/Command/Require/Db.hs
sodic, line 1: If we're keeping this file, I believe the convention dictates that it should go in a subdirectory (`Command/Require/Db.hs`).

But that's a little to fragmented for my taste, especially because we'd either also have to do that for `InWaspProject`, or sacrifice consistency.

My favorite option is keeping everything in `Command/Require.hs` regardless of the long comment. Is the comment the only reason this is in its own module?

waspc/cli/src/Wasp/Cli/Command/Require/Db.hs
```diff
@@ -0,0 +1,53 @@
...
+import qualified Wasp.Project.Common as Project.Common
+
+data DbConnectionEstablishedFromOutDir = DbConnectionEstablishedFromOutDir deriving (Typeable)
+
+-- TODO(carlos): Modify this requirement to work in `.wasp/build` as well
+--
+-- Why is this requirement specific to `.wasp/out`? For the time being, most of
+-- our DB are designed to work in the context of `wasp start`, and the functions
+-- end up depending on two specific outputs:
+-- 1. the `.env.server` file, as copied to `<WaspProjectDir>/server/.env`
```
sodic: I don't think this is the right path. `WaspProjectDir` is the root directory of the Wasp project, not the `.wasp/out` dir. (which I'm guessing this is supposed to be). Same thing with the `schema.prisma` bullet point.

waspc/cli/src/Wasp/Cli/Command/Require/Db.hs
```diff
@@ -0,0 +1,53 @@
...
+import Data.Data (Typeable)
+import qualified StrongPath as SP
+import Wasp.Cli.Command (CommandError (CommandError), Requirable (checkRequirement), require)
+import Wasp.Cli.Command.Require (InWaspProject (InWaspProject))
+import Wasp.Generator.DbGenerator.Operations (isDbConnectionPossible, testDbConnection)
+import qualified Wasp.Project.Common as Project.Common
+
+data DbConnectionEstablishedFromOutDir = DbConnectionEstablishedFromOutDir deriving (Typeable)
+
+-- TODO(carlos): Modify this requirement to work in `.wasp/build` as well
```
sodic: Can we create an issue for this?

sodic (commented):
Nice change, I left some suggestions :)

waspc/cli/src/Wasp/Cli/Command/Require/Db.hs
cprecioso: > Is the comment the only reason this is in its own module?

Yes, undid it haha

waspc/cli/src/Wasp/Cli/Command/Require/Db.hs
cprecioso: #2858

cprecioso:
@sodic ready

waspc/cli/src/Wasp/Cli/Command/Require.hs
```diff
@@ -100,3 +76,43 @@ instance Requirable InWaspProject where
           ( "Couldn't find wasp project root - make sure"
               ++ " you are running this command from a Wasp project."
           )
+
+data DbConnectionEstablishedFromOutDir = DbConnectionEstablishedFromOutDir deriving (Typeable)
+
+-- TODO(carlos): Modify this requirement to work in `.wasp/build` as well (#2858)
```
Martinsos: FYI this comment is a standalone comment -> it is not attached to any code, in a way.

If you wanted to make it reference the code below (which I thikn you did), then you should make it start with `-- |` -> this makes it a "declaration comment". Check out Haddock (like jsdoc but for Haskell) docs here: https://haskell-haddock.readthedocs.io/latest/markup.html#documenting-a-top-level-declaration .

waspc/cli/src/Wasp/Cli/Command/Require.hs
```diff
@@ -100,3 +76,43 @@ instance Requirable InWaspProject where
...
+--
+-- Why is this requirement specific to `.wasp/out`? For the time being, most of
+-- our DB are designed to work in the context of `wasp start`, and the functions
```
Martinsos:
```suggestion
-- our DB functions are designed to work in the context of `wasp start`, and the functions
```

I guess?

waspc/cli/src/Wasp/Cli/Command/Require.hs
```diff
@@ -100,3 +76,43 @@ instance Requirable InWaspProject where
...
+-- 1. the `.env.server` file, as copied to `<ProjectRootDir>/server/.env`
+-- 2. the `schema.prisma` file, processed by the Wasp compiler (to add Auth
+--   entities), and copied to `<ProjectRootDir>/db/schema.prisma`
+--
+-- The `wasp start` command generates these files, but `wasp build` does not, as
+-- it expects the environment variables to be given to the resulting Docker
+-- container. Moreover, the processing of the `schema.prisma` file is done
+-- inside the Docker image building process, so the user's local filesystem
+-- doesn't see the final schema file, making it unavailable for the compiler
+-- running normally in the user's environment.
```
Martinsos: This whole comment was a bit hard to read, it goes quite much into the implementation details and I think it requires a lot of konwledge already to understand it. It is not easy to read nor understand, which means most people won't get it on its own, which doesn't make it all that useful.

I would ask what is the purpose of this comment at all, if there is a github issue for this? I guess the point is that if somebody looks at it here and wonders why tehre is no support for BuildDir, they don't just go about adding it and rediscovering the reasons behind why we don't have it yet, but instead they learn about it here.

If so, I think we can have a short comment that warns that there is a reason why no support for BUIld dir, links to that github issue, and says very shortly that build dir is on purpose missing some stuff like .env file and prisma.schema, by design, which is why this is more delicate to implement, and that is it.

waspc/cli/src/Wasp/Cli/Command/Require.hs
```diff
@@ -100,3 +76,43 @@ instance Requirable InWaspProject where
           ( "Couldn't find wasp project root - make sure"
               ++ " you are running this command from a Wasp project."
           )
+
+data DbConnectionEstablishedFromOutDir = DbConnectionEstablishedFromOutDir deriving (Typeable)
```
Martinsos: And now what I was planning to ask once you make this PR :D -> let's push it a small step further and turn this into `data DbConnectionEstablishedFrom fromDir = DbConnectionEstablishedFrom fromDir ...` right now. And implement just one instance of it for now, the `instance Requirable (DbConnectionEstablishedFrom OutDir)`. And add FlexibleInstances as needed. It is small change, will be interesting practice for you, and it lays the ground immediately for later adding BuildDir.

If you would rather do this later when adding BuildDir -> let's not and let's do it now, we do already know we will be adding it for sure, and this communicates better how this part of our codebase should work, that we intend to support multiple dirs. So it is nicer code + allows for easier extension in the future.

Martinsos (changes requested):

waspc/cli/src/Wasp/Cli/Command/Require.hs
cprecioso: I don't think it makes sense for this specific comment to be attached but good to know for the future!

sodic:
I'm removing myself from this one. Tag me if you need me :)

cprecioso:
Done @Martinsos

waspc/cli/src/Wasp/Cli/Command/Require.hs
```diff
@@ -100,3 +79,36 @@ instance Requirable InWaspProject where
...
           )
+
+data FromOutDir = FromOutDir deriving (Typeable)
+
+-- TODO(carlos): Refactor this requirement to work in `.wasp/build` as well
+--
+-- Check GitHub issue #2858 for more details. In short, `.wasp/build` is missing
+-- the `server/.env` and `db/schema.prisma` files. `wasp start` creates these
+-- files for `.wasp/out`, which our current DB functions expect; but `wasp
+-- build` doesn't do the same for `.wasp/build`.
```
Martinsos:
```suggestion
-- TODO(carlos): Refactor this requirement to work for `.wasp/build` as well, by implementing `instance Requirable (DbConnectionEstablished FromBuildDir) where`.
--
-- Check GitHub issue #2858 for more details. In short, it is not trivial to do because `.wasp/build` is missing
-- the `server/.env` and `db/schema.prisma` files. `wasp start` creates these
-- files for `.wasp/out`, which our current DB functions expect; but `wasp
-- build` doesn't do the same for `.wasp/build`.
```

Martinsos: What confuses me is that you say "this requirement" but since this comment block is standalone, I don't get what "this requirement" is referring to, it doesn't seem clear.

waspc/cli/src/Wasp/Cli/Command/Require.hs
```diff
@@ -100,3 +79,36 @@ instance Requirable InWaspProject where
...
+
+data DbConnectionEstablished fromDir = DbConnectionEstablished fromDir deriving (Typeable)
```
Martinsos, line 94: Would `DbConnectionEstablishedFrom fromDir` and therefore `DbConnectionEstablishedFrom OutDir <- require` be better? I on purpose suggested that as naming, because that allows you to have `OutDir` and `BuildDir` which make more sense as concepts than `FromOutDir` and `FromBuildDir`, in the sense that they are a standalone concept also, so that is nicer. We even have them already somewhere else in the codebase and they can be reused if I am correct. Ok I did a bit of digging, seems we don't although we almost do :D (if we did it would probably go into Wasp.Project.Common).

Martinsos (approved):
Ok this is looking great now! Cool, it works :D.

I left some comments, check them out, but I also approved, so feel free to merge once you are happy with them.

waspc/cli/src/Wasp/Cli/Command/Require.hs
cprecioso: I purposefully didn't use `OutDir`/`BuildDir` because it seemed too general. Indeed, it would seem like we would have to have it all throughout our codebase.

I named them `From[blah]Dir` so, together with the fact that they come from `Cli.Command.Require`, it makes sense that their scope is limited. I can change it if you feel differently though.

cprecioso:
@Martinsos I know you approved but I'd like your go-ahead again on the fixes

waspc/cli/src/Wasp/Cli/Command/Require.hs
Martinsos: I get why you did it, and it makes sense. I am still drawn to that other approach, but I am not 100% confident about it without diving into it deeper so maybe this is the best indeed! Yeah, might be half-solution if we add it but then not use it in other places. Although maybe it was just never needed. Ok, sounds good!
raw

waspc/cli/src/Wasp/Cli/Command/Require.hs
```diff
@@ -100,3 +79,37 @@ instance Requirable InWaspProject where
...
+
+data FromOutDir = FromOutDir deriving (Typeable)
+
+-- TODO(carlos): Create a `FromBuildDir` instance of `DbConnectionEstablished`
+-- as well.
+--
+-- Check GitHub issue #2858 for more details. In short, `.wasp/build` is missing
+-- the `server/.env` and `db/schema.prisma` files. `wasp start` creates these
+-- files for `.wasp/out`, which our current DB functions expect; but `wasp
+-- build` doesn't do the same for `.wasp/build`.
```
Martinsos: It is not clear why are you explaining the lacking of these files in .wasp/build, context is missing. You are because without them it is hard to do this check. So I would just mention that super shortly -> basically that the reason why we didn't implement it yet is because it is a bit tricky because of those files. That is it. I mentioned that in my previous comment already I think.

waspc/cli/src/Wasp/Cli/Command/Require.hs
```diff
@@ -83,12 +83,14 @@ instance Requirable InWaspProject where
...
--- Check GitHub issue #2858 for more details. In short, `.wasp/build` is missing
--- the `server/.env` and `db/schema.prisma` files. `wasp start` creates these
--- files for `.wasp/out`, which our current DB functions expect; but `wasp
--- build` doesn't do the same for `.wasp/build`.
+-- Currently, our functions that interface with Prisma expect the `server/.env`
+-- and `db/schema.prisma` files. The `wasp start` command runs from `.wasp/out`,
+-- and creates these files as part of its function. `wasp build` runs from
+-- `.wasp/build`, but it doesn't create these files (check #2858 for
+-- explanation). We should refactor our Prisma functions to not depend on these
+-- files.
```
Martinsos: So I feel bad writing about this again, but if I read this whole comment from start to bottom, it is not clear to me what is this second paragraph (part I am commenting on) about, how it is connected to the thing.

You say that we should create a FromBuildDir instance. Ok, makes sense. You don't say why, but that is kind of obvious (I think?).

But then the second part immediately goes into details: that server/.env and db/schema.prisma is needed, and they are not genreated for wasp build. And we should refactor stuff. But how is that connected to the paragraph above? That is not clear. I am sure about that because I have context and yet I have to think twice to remember why it is important. So somebody without the context also has to think more than they should + they might get it at the end.

I provided suggestion how to go about this but it seems you dropped it out. How is that? Maybe because you tried to do what I suggested without owning it on your own? Would it help if you focused on less what I told you to write, and focus more on what you think would fix the problem I described?
For you, what is the purpose of this whole comment? What do you want to communicate, that is not already in that github issue -> why do we even have, when we have already that GH issue? I think we should have it, but how do you see it?

Btw, additional thing:
> We should refactor our Prisma functions to not depend on these
-- files.
Hmmm, maybe yes, maybe not, I am not yet sure that is the right solution. Maybe solution is generating those files for wasp build. So I would skip this part where you offer the solution here, we can discuss that in the issue.

cprecioso: > I provided suggestion how to go about this but it seems you dropped it out.

I did follow your suggestion, but maybe there was a misunderstanding. The suggestion was to `Have a short comment that warns that there is a reason why no support for Build dir, links to that github issue, and says very shortly that build dir is on purpose missing some stuff like .env file and prisma.schema, by design, which is why this is more delicate to implement, and that is it.` and then to `mention that super shortly -> basically that the reason why we didn't implement it yet is because it is a bit tricky because of those files`.

I don't think I dropped any of your suggestions, I rewrote the comment to basically say quite directly what you suggested. (or at least what I understood that you suggested from your comments 🤔)

> Maybe because you tried to do what I suggested without owning it on your own? Would it help if you focused on less what I told you to write, and focus more on what you think would fix the problem I described?

I don't see how this is related to me not following your suggestions? IIUC you're telling me the opposite thing here(?)

> For you, what is the purpose of this whole comment? What do you want to communicate, that is not already in that github issue?

None now that we have the issue. I'd just have the first line. But you explcitly asked me to also explain that the build dir is missing the files and why that affects us.

> So I would skip this part where you offer the solution here

Fair, makes sense

Martinsos: Ok, so first things first, if you don't see the purpose in big part of this comment (or in some other change I suggest, in general), you shouldn't have written it! You shouldn't be writing stuff just because I told you to. You should discuss it with me till it makes sense to both of us, or we explicitly agree to disagree and go with specific approach (which should happen rarely).

When I said you dropped out part of my suggestion, I meant the part where you connect the second paragraph with the first, which is mentioned twice in my quote that you just quoted above:

> says very shortly that build dir is on purpose missing some stuff like .env file and prisma.schema, by design, which is why this is more delicate to implement,

> the reason why we didn't implement it yet is because it is a bit tricky because of those files

So **the reason** why we didn't implement FromBuildDir so far, that is the point. You explained that reason, but quite technically + you didn't say it is the reason, and I don't think it is clear on its own.

I don't know if I am communicating this well or not, but basically, I was imagining comment to go something like this:

```
-- TODO: Implement a `FromBuildDir` instance of `DbConnectionEstablished`
-- as well. (#2858)
-- The reason why we haven't implemented it already is because `.wasp/build` dir by design does not have some files like `.env` or `prisma.schema`, which makes it more tricky.
```

I just checked the 2858, I see it covers this well. I don't mind having a comment like the one I listed above because it gives  me a lot of info without having to open the actual issue. But it could be even shorter, like

```
-- TODO: Implement a `FromBuildDir` instance of `DbConnectionEstablished`
-- as well. Check #2858 for more details and why we haven't implement it yet.
```

Issue #2825, opened 2025-06-09
Automatically test example apps before release


sodic:
On each release, we manually run and check all example apps. This can take a while, especially when there are fixes involved. We have recently reorganized and cleaned up our example apps in https://github.com/wasp-lang/wasp/issues/2713, so let's see whether we can automate everything we check in the remaining apps.

What we did in #2713:
- Ensured that all example apps combined cover most of Wasp's feature surface.
- Added some headless tests for some of those features.

What's left to do:
1. Ensure those tests truly test all the specific features we care about (covered by #2812)
2. Ensure we run the tests both in dev and "in production" (on a built version of the app). We already got this capability with https://github.com/wasp-lang/wasp/pull/2663
3. Add those tests to the CI for each example app

This task sould take care of points 2 and 3.

This is more about testing Wasp than about testing specific examples.
Closing this issue means we no longer need to manually test the example apps as a part of our release checklist.

infomiho:
Related PR https://github.com/wasp-lang/wasp/pull/2791

But it's slow until we parallelize our CI I'm not sure if we want to run this on each PR.

Also, #2812 updates should help with closing the that gap of not testing all Wasp features.

sodic:
@infomiho Do we have an issue for that parallelization? I can't find it. I'll create one if not.

If that draft PR is ready, can we only run it on releases for now?

@cprecioso @infomiho The three of us will sync (or the two of you without me) when you get back to see what remains to be done to make this happen.

From what I got, most of the groundwork has already been laid out.

Martinsos:
TLDR: This likely comes down to "just" creating a workflow for this in Github Actions and then running it at the right moment (on tags?).

FranjoMindek:
Considering that we already run the `todoApp` headless tests  on all PRs to `waspc/**`, I'll assume that this task is about running headless tests for leftover example apps in `wasp/examples`.

Those are `TodoApp`, `TodoAppTs`, `waspello`, `waspleau`, `websockets-realtime-voting`.
I'll explore how we will call this workflow.

PR #2791, opened 2025-05-22
Run all example apps' headless tests in the CI

infomiho:

PR #2854, opened 2025-06-26
Run example apps headless tests in CI 

FranjoMindek:
Fixes https://github.com/wasp-lang/wasp/issues/2825

### DIFF
--- a/.github/workflows/root-example-apps.yaml
+++ b/.github/workflows/root-example-apps.yaml
@@ -1,0 +1,158 @@
+name: Run headless tests on root example apps
+
+on:
+  workflow_call:
+  workflow_dispatch:
+  pull_request:
+
+env:
+  WASP_TELEMETRY_DISABLE: 1
+
+defaults:
+  run:
+    shell: bash
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  setup-and-cache:
+    name: Setup Global Dependencies and Cache Playwright
+    runs-on: ubuntu-22.04
+    outputs:
+      playwright-version: ${{ steps.get-playwright-version.outputs.version }}
+    
+    steps:
+      - uses: "actions/checkout@v4"
+        with:
+          # Workaround for a Github Checkout action bug
+          # https://github.com/actions/checkout/issues/1359#issuecomment-1567902034
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: "lts/*"
+
+      - name: Cache Wasp Installation
+        id: cache-wasp
+        uses: actions/cache@v4
+        with:
+          path: ~/.local/share/wasp-lang
+          key: wasp-${{ runner.os }}-latest
+          restore-keys: |
+            wasp-${{ runner.os }}-
+
+      - name: Set up Wasp
+        id: setup-wasp
+        if: steps.cache-wasp.outputs.cache-hit != 'true'
+        run: |
+          curl -sSL https://get.wasp.sh/installer.sh | sh
+
+      - name: Add Wasp to PATH
+        run: |
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+
+      - name: Cache Wasp App Runner
+        id: cache-wasp-app-runner
+        uses: actions/cache@v4
+        with:
+          path: ~/.npm
+          key: wasp-app-runner-${{ runner.os }}-latest
+          restore-keys: |
+            wasp-app-runner-${{ runner.os }}-
+
+      - name: Set up Wasp App Runner
+        if: steps.cache-wasp-app-runner.outputs.cache-hit != 'true'
+        uses: wasp-lang/runner-action@main
+
+      - name: Get Playwright Version from Example App
+        id: get-playwright-version
+        working-directory: examples/waspello
+        run: |
+          npm ci
+          PLAYWRIGHT_VERSION=$(npm ls @playwright/test | grep @playwright | sed 's/.*@//')
+          echo "Playwright's Version: $PLAYWRIGHT_VERSION"
+          echo "version=$PLAYWRIGHT_VERSION" >> $GITHUB_OUTPUT
+
+      - name: Cache Playwright Browsers for Playwright's Version
+        id: cache-playwright-browsers
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-browsers-${{ steps.get-playwright-version.outputs.version }}
+
+      - name: Setup Playwright
+        id: setup-playwright
+        if: steps.cache-playwright-browsers.outputs.cache-hit != 'true'
+        run: |
+          npx playwright install --with-deps
+
+  test-example-apps:
+    name: Test Example Apps
+    runs-on: ubuntu-22.04
+    needs: setup-and-cache
+    
+    strategy:
+      fail-fast: false
+      matrix:
+        example:
+          - tutorials/TodoApp
+          - tutorials/TodoAppTs
+          - waspello
+          - waspleau
+          - websockets-realtime-voting
+
+    steps:
+      - uses: "actions/checkout@v4"
+        with:
+          # Workaround for a Github Checkout action bug
+          # https://github.com/actions/checkout/issues/1359#issuecomment-1567902034
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: "lts/*"
+
+      - name: Restore Wasp Installation Cache
+        uses: actions/cache/restore@v4
+        with:
+          path: ~/.local/share/wasp-lang
+          key: wasp-${{ runner.os }}-latest
+          restore-keys: |
+            wasp-${{ runner.os }}-
+      - name: Add Wasp to PATH
+        run: |
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+
+      - name: Restore Wasp App Runner Cache
+        uses: actions/cache/restore@v4
+        with:
+          path: ~/.npm
+          key: wasp-app-runner-${{ runner.os }}-latest
+          restore-keys: |
+            wasp-app-runner-${{ runner.os }}-
+
+      - name: Set up Wasp App Runner
+        uses: wasp-lang/runner-action@main
+
+      - name: Restore Playwright Browsers Cache
+        uses: actions/cache/restore@v4
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-browsers-${{ needs.setup-and-cache.outputs.playwright-version }}
+
+      - name: Install Dependencies for ${{ matrix.example }}
+        working-directory: examples/${{ matrix.example }}
+        run: |
+          npm ci
+
+      - name: Run Headless Tests for ${{ matrix.example }}
+        working-directory: examples/${{ matrix.example }}/headless-tests
+        # Useful for debugging when the playwright webserver fails.
+        # env:
+        #   DEBUG: pw:webserver
+        run: |
+          npx playwright test


infomiho:
@FranjoMindek just a heads up, there was this PR that tried to tackle the same thing - but I never finished it since running all the tests took too long: https://github.com/wasp-lang/wasp/pull/2791

.github/workflows/examples-headless-tests.yaml
```diff
@@ -0,0 +1,157 @@
...
+          fi
+
+      - name: Extract Database Provider
+        id: extract-db-provider
+        working-directory: examples/${{ matrix.example }}
+        run: |
+          DATABASE_PROVIDER=$(wasp-cli info \
+            | grep "Database system" \
+            | sed 's/.*: //' \
+            | sed -e 's/\x1b\[[0-9;]*m//g' \
```
FranjoMindek: Any way to add comments here?
Any variant I've tried, it got inferred as part of "bash string" and errored.

At first I read the `schema.prisma` file but then remembered `wasp info`.
Seems more robust, the only problem is that we use ANSI colors in the output.

This resulted is some debugging since `'sqlite' == 'sqlite'` was false because one `sqlite` was colored, and this can't be seen in the logs.

I want to leave a comment to explain `sed -e 's/\x1b\[[0-9;]*m//g' \` which removes ANSI coloring chars.
But can't find out how.

.github/workflows/examples-headless-tests.yaml
```diff
@@ -0,0 +1,157 @@
...
+  workflow_call:
+  workflow_dispatch:
+  push:
+    paths:
+      - "examples/**"
+    branches:
+      - main
+      - release
+  pull_request:
+    # paths:
```
FranjoMindek: Temp, or workflow won't trigger here, because the initial PR commits didn't include changes to `examples/**`

.github/workflows/examples-headless-tests.yaml
```diff
@@ -0,0 +1,157 @@
...
+      - name: Restore Playwright Browsers Cache
+        uses: actions/cache/restore@v4
+        with:
+          key: playwright-browsers-${{ needs.setup-and-cache-playwright.outputs.playwright-version }}
+          path: ~/.cache/ms-playwright
+
+      - name: Set up Wasp App runner
+        run: npm i -g @wasp.sh/wasp-app-runner
+
+      - name: Install Dependencies for ${{ matrix.example }}
```
FranjoMindek, line 51: I'm following `waspc/examples/todoApp` convention, so if we detect `.env.server.headless` we use that env for running the headless tests.

This is because some apps require env vars to run properly.

examples/tutorials/TodoApp/headless-tests/playwright.config.ts
```diff
@@ -1,5 +1,8 @@
 import { defineConfig, devices } from "@playwright/test";
 
+const HEADLESS_TEST_MODE = process.env.HEADLESS_TEST_MODE ?? "dev";
+const WASP_CLI_CMD = process.env.WASP_CLI_CMD ?? "wasp-cli";
+
```
FranjoMindek, line 5: I'm doubling down on "dev" here, so that the default is that we run headless tests in `dev` mode with "dev" wasp (`wasp-cli`).

I think this is more suitable for us.

.github/workflows/examples-headless-tests.yaml
```diff
@@ -0,0 +1,152 @@
...
+env:
+  WASP_TELEMETRY_DISABLE: 1
+
+defaults:
+  run:
+    shell: bash
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
sodic: Can we remove this?

.github/workflows/examples-headless-tests.yaml
```diff
@@ -0,0 +1,101 @@
...
+
+      - name: Install Dependencies for ${{ matrix.example }}
+        working-directory: examples/${{ matrix.example }}
+        run: |
+          npm ci
+          npx playwright install --with-deps
+
+          if [ -f ".env.server.headless" ]; then
+            cp .env.server.headless .env.server
+          fi
```
sodic, line 59: What if it doesn't exist, is that a valid scenario?

.github/workflows/examples-headless-tests.yaml
sodic: If the normal comments don't work (and I remember you telling me they don't), then we can either put one above or you can extract this to the `get-database-provider.sh` script and comment it there.

Also, I remember us talking about creating an issue that deals with Wasp's ANSI color output in non-interactive mode. I was supposed to create it, right? I forgot :upside_down_face:

.github/workflows/examples-headless-tests.yaml
```diff
@@ -0,0 +1,101 @@
...
+          DATABASE_PROVIDER=$(wasp-cli info \
+            | grep "Database system" \
+            | sed 's/.*: //' \
+            | sed -e 's/\x1b\[[0-9;]*m//g' \
+            | tr '[:upper:]' '[:lower:]')
+
+          if [ -z "$DATABASE_PROVIDER" ]; then
+            echo "ERROR: Could not determine database system from wasp-cli info"
+            exit 1
+          fi
```
sodic: If you decide to extract a script, this could go into it as well.

.github/workflows/examples-headless-tests.yaml
```diff
@@ -0,0 +1,101 @@
...
+
+          echo "Database provider: $DATABASE_PROVIDER"
```
sodic: Is this just for debugging or do we plan to keep it?

.github/workflows/examples-headless-tests.yaml
```diff
@@ -0,0 +1,101 @@
...
+      - name: Run Headless DEV Tests for ${{ matrix.example }}
+        working-directory: examples/${{ matrix.example }}
+        env:
+          WASP_CLI_CMD: wasp-cli
+          HEADLESS_TEST_MODE: dev
+        run: |
+          npm run test
+
+      - name: Run Headless BUILD Tests for ${{ matrix.example }}
+        if: steps.extract-db-provider.outputs.provider != 'sqlite'
```
sodic, line 78: Normally I'd say, let's do `isProjectBuildable = sptes.extract...provider != 'sqlite'`. But since I don't think we can do that, let's add a comment explaining what this is about.

.github/workflows/examples-headless-tests.yaml
```diff
@@ -0,0 +1,101 @@
...
+
+env:
+  WASP_TELEMETRY_DISABLE: 1
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  test-example-apps:
```
sodic, line 19: Should we separate this into multiple jobs? I'm not sure what the standard is (a question for @infomiho or @cprecioso).

sodic (commented):
Good stuff!

Left some comments/questions.

I don't know enough about GH actions so I'd like to see @infomiho or @cprecioso review it too.

.github/workflows/examples-headless-tests.yaml
FranjoMindek: Hm, this should've been removed already?
Edit: oh so this comment is very old, just hasn't been posted in review yet

.github/workflows/examples-headless-tests.yaml
FranjoMindek: Yes, some examples don't require env vars at all.

.github/workflows/examples-headless-tests.yaml
FranjoMindek: We can remove.

.github/workflows/examples-headless-tests.yaml
FranjoMindek: I have the following in the description:

> The biggest benefit here is that we don't stop testing other example apps when one of them fails.

This also leaves a good baseline for when we optimize the CI.

.github/workflows/examples-headless-tests.yaml
sodic: Sorry sorry.

.github/workflows/examples-headless-tests.yaml
sodic: Ok, let's do it :)

.github/workflows/examples-headless-tests.yaml
FranjoMindek: Some background:

I did experiment which much more "optimized" and cached approach:
https://github.com/wasp-lang/wasp/pull/2963/files

But Filip rather wanted to focus on just getting it to work now, and we will optimize later.
That's is why this one is much more simple and straightforward (also no playwright caching).

infomiho: > Should we separate this into multiple jobs?

I believe it's already executed as multiple jobs because of the `matrix` option.

Image

Could you clarify a bit @sodic what you meant exactly should be the change?

.github/workflows/examples-headless-tests.yaml
```diff
@@ -0,0 +1,91 @@
...
+  pull_request:
+    # paths:
+    #   - "examples/**"
+
+env:
+  WASP_TELEMETRY_DISABLE: 1
+
+defaults:
+  run:
+    shell: bash
```
cprecioso: Not needed

.github/workflows/examples-headless-tests.yaml
cprecioso: I would only split it into different jobs if the outcome of each job was easily packageable (and it made sense). I don't think it's really necessary here. If we ever get a Wasp CLI installation action, then I would use it here, that's the only refactor I'd suggest.

sodic: I think you all answered my questions. Ok, fine by me then :)

.github/workflows/examples-headless-tests.yaml
sodic: Can we move this coment above the condition?

scripts/get-wasp-database-provider.sh
```diff
@@ -0,0 +1,40 @@
...
+  echo "       Where <command> is either 'wasp' or 'wasp-cli'" >&2
+  exit 1
+fi
+
+WASP_COMMAND="$1"
+
+if [ "$WASP_COMMAND" != "wasp" ] && [ "$WASP_COMMAND" != "wasp-cli" ]; then
+  echo "ERROR: Command must be either 'wasp' or 'wasp-cli', got: $WASP_COMMAND" >&2
+  exit 1
+fi
```
sodic, line 23: I'm not sure whether we need all this.

What's the best practice for writing these CI/CD scripts? Should they be robust, should they be inline? Is this small enough to fit in the action, was I overdoing it when I told @FranjoMindek to take it out?

Because, when this was inline in the `yaml`, it was much smaller.
@infomiho @cprecioso another question for you two.

sodic (commented):
Just one more thing I'm not 100% on.

scripts/get-wasp-database-provider.sh
FranjoMindek: I've added arg parsing logic which takes as much space as the logic itself.
Also the comments explaining the pipeline.

.github/workflows/examples-headless-tests.yaml
```diff
@@ -0,0 +1,87 @@
+name: Run headless tests on root example apps
+
+on:
+  workflow_call:
```
FranjoMindek, line 4: `workflow_call` and `workflow_dispatch` can also be removed

.github/workflows/examples-headless-tests.yaml
```diff
@@ -0,0 +1,87 @@
...
+        example:
+          - tutorials/TodoApp
+          - tutorials/TodoAppTs
+          - waspello
+          - waspleau
+          - websockets-realtime-voting
+    steps:
+      - uses: "actions/checkout@v4"
+        with:
+          # Workaround for a Github Checkout action bug
```
FranjoMindek: I'd maybe remove this ref thing

.github/workflows/examples-headless-tests.yaml
```diff
@@ -0,0 +1,83 @@
+name: Run headless tests on root example apps
+
+on:
+  push:
+    paths:
+      - "examples/**"
+    branches:
+      - main
+      - release
```
infomiho, line 10: Does it makes sense to also run this for the in the workflow that runs when we tag a new Wasp version? https://github.com/wasp-lang/wasp/blob/main/.github/workflows/waspc-release.yaml

cc: @cprecioso

scripts/get-wasp-database-provider.sh
cprecioso: Matter of taste IMO, in then end the more things that can be run and debugged outside of the yaml without being overkill, the better.

.github/workflows/examples-headless-tests.yaml
cprecioso: Sure, why not 🤷‍♂️

scripts/get-wasp-database-provider.sh
infomiho: I'm all for these kind of extractions. I think this will make it easier to reason about the logic in the future when somebody needs to change something in the CI.

Let me go one step further and propose one extra NPM package where we could extract all the shared testing logic:
- extracting type of database (wasp app runner could maybe use it)
- common parts of the Playwright configs
- common utils related to testing Wasp apps e.g. how we verify the email or smth like that

It feels like we are repeating these kind of utils in many places so I feel like we could benefit from such a package. `wasp-testing-utils`

Issue #2844, opened 2025-06-17
Move Fly deployments from MAD region to some other (it is being deprecated)


Martinsos:
![Image](https://github.com/user-attachments/assets/ee987eb4-4090-47d2-9e60-9fb091d04c3d)

cprecioso:
All moved to CDG

Issue #2831, opened 2025-06-10
Automatically deploy all example apps on release


sodic:
We deploy some of our example apps and use them for showcasing wasp (e.g., OpenSaas, web sockets voting, etc.).

We should deploy all apps we deem useful and we should do it automatically on each release. This is a distinct issue from #2665. That issue is all about testing the Wasp deployment feature for different providers, and this one is all about deploying specific example apps (as if we were regular users).

PR #2855, opened 2025-06-26
Move all example apps to CDG region

cprecioso:
- Part of #2831
- Part of #2844

--- 

Move all of our example apps to CDG (Paris)

- Some were on MAD (Madrid), which is getting deprecated as per #2844.
- Some were on MIA (Miami), but it makes sense to move them to CDG as well for consistency, and because all the team is in Europe.

The CDG region was chosen base on recommendation by Fly themselves (see the email at #2844) 

**Please note**:
- This PR is being done against `release`.
- The new manifest have not been deployed yet, will do it once we merge.
- Thoughts app has been removed in `main` (and taken offline) but is still in the `release` branch. I did not modify it.

### DIFF
--- a/examples/waspleau/fly-client.toml
+++ b/examples/waspleau/fly-client.toml
@@ -4,7 +4,7 @@
 #
 
 app = "waspleau-app-client"
-primary_region = "mad"
+primary_region = "cdg"
 
 [build]
 
--- a/examples/waspleau/fly-server.toml
+++ b/examples/waspleau/fly-server.toml
@@ -4,7 +4,7 @@
 #
 
 app = "waspleau-app-server"
-primary_region = "mad"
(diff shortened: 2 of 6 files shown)


cprecioso:
> - Can we redeploy them to ensure everything's OK?

I was waiting until we approved the change to do it. It seems it's a matter of scaling up in a new region and scaling down in the other: https://fly.io/docs/launch/scale-count/#add-a-new-region

> Are we sure we want to keep them in Europe?

@Martinsos @matijaSos any ideas?

Martinsos:
Europe vs US -> I don't think it matters much, both are ok with me. Let's go with Paris.

cprecioso:
Updated these two deployments, now running at CDG

PR #2857, opened 2025-06-26
WIP: Automatically deploy example apps

### LINKED ISSUE #2831: Automatically deploy all example apps on release

cprecioso:
WIP

- Resolves #2831

### Description

> Describe your PR! If this PR closes an issue, use “Fixes #(issue_number)" syntax so GitHub will auto-close it when merged.

1. [ ] I updated [`ChangeLog.md`](https://github.com/wasp-lang/wasp/blob/main/waspc/ChangeLog.md) with description of the change this PR introduces.

### Add a regression test if needed

If you did a **bug fix**, make sure you satisfy the following:

1. [ ] I added a regression test that reproduces the bug and verifies the fix.

If you're unable to add a regression test, please explain why.
This likely indicates that our current testing setup needs improvement.

If you did code changes and **added a new feature**, make sure you satisfy the following:

1. [ ] I updated [`waspc/examples/todoApp`](https://github.com/wasp-lang/wasp/tree/main/waspc/examples/todoApp) and its e2e tests as needed and manually checked it works correctly.

If you did code changes and **updated an existing feature**, make sure you satisfy the following:

1. [ ] I updated [`waspc/examples/todoApp`](https://github.com/wasp-lang/wasp/tree/main/waspc/examples/todoApp) and its e2e tests as needed and manually checked it works correctly.

### Update starter apps if needed

If you did code changes and **updated an existing feature**, make sure you satisfy the following:

1. [ ] I updated [starter skeleton](https://github.com/wasp-lang/wasp/tree/main/waspc/data/Cli/templates/skeleton) as needed and manually checked it works correctly.
2. [ ] I updated [`basic` starter](https://github.com/wasp-lang/wasp/tree/main/waspc/data/Cli/templates/basic) as needed and manually checked it works correctly.
3. [ ] I updated [`todo-ts` starter](https://github.com/wasp-lang/starters/tree/dev/todo-ts) as needed and manually checked it works correctly.
4. [ ] I updated [`embeddings` starter](https://github.com/wasp-lang/starters/tree/dev/embeddings) as needed and manually checked it works correctly.
5. [ ] I updated [`saas` starter](https://github.com/wasp-lang/open-saas/tree/main/template) as needed and manually checked it works correctly.


.github/workflows/deploy-examples.yml
```diff
@@ -0,0 +1,68 @@
...
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Wasp version to use for deployment (default to latest)"
+        required: false
+
+  # TODO: REMOVE THIS BEFORE MERGING, JUST FOR TESTING
+  push:
+    branches:
+      - 2831-automatically-deploy-all-example-apps-on-release
```
cprecioso: I will remove this before merging, but I need it in order to trigger the workflow (I can't use `workflow_dispatch` because it is only allowed if it's present in the default branch too)

.github/workflows/deploy-examples.yml
```diff
@@ -0,0 +1,68 @@
...
+            dir: examples/websockets-realtime-voting
+            url: https://websockets-voting-client.fly.dev/
+
+    name: Deploy ${{ matrix.projects.name }}
+
+    runs-on: ubuntu-latest
+
+    environment:
+      name: ${{ matrix.projects.name }}-production
+      url: ${{ matrix.projects.url }}
```
cprecioso: This allows us to have a history of deployments at the dpeloyments page:
- https://github.com/wasp-lang/wasp/deployments/waspleau-production
- https://github.com/wasp-lang/wasp/deployments/websockets-realtime-voting-production

.github/workflows/deploy-examples.yml
```diff
@@ -0,0 +1,68 @@
...
+
+concurrency:
+  group: "deploy-examples"
+  cancel-in-progress: true
+
+jobs:
+  deploy:
+    strategy:
+      matrix:
+        projects:
```
cprecioso: `thoughts` is also present in the `release` branch, but it has been removed from `main` and not deployed anymore

.github/workflows/deploy-examples.yml
sodic: Ah yes, sounds familiar...

.github/workflows/deploy-examples.yml
sodic: Nice!

.github/workflows/deploy-examples.yml
```diff
@@ -0,0 +1,68 @@
...
+    secrets:
+      FLY_API_TOKEN:
+        description: "Fly API token for deploying apps"
+        required: true
+
+  # Allow calling this workflow manually from the GitHub UI
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Wasp version to use for deployment (default to latest)"
```
sodic, line 20: The picture mentions GHC and Node versions, but this mentions the Wasp version.

How come?

Also, should we figure out the Wasp version from the `version` field in the app, have a single source of truth?

.github/workflows/deploy-examples.yml
```diff
@@ -0,0 +1,68 @@
...
+    strategy:
+      matrix:
+        projects:
+          - name: waspleau
+            dir: examples/waspleau
+            url: https://waspleau-app-client.fly.dev/
+
+          - name: websockets-realtime-voting
+            dir: examples/websockets-realtime-voting
+            url: https://websockets-voting-client.fly.dev/
```
sodic: There's some duplication here - we have the same info in our project and the secrets. Any chance we can get rid of it?

.github/workflows/deploy-examples.yml
```diff
@@ -0,0 +1,68 @@
...
+    steps:
+      - name: Install Wasp
+        run:
+          curl -sSfL https://get.wasp.sh/installer.sh | sh -- ${{ inputs.version
+          && format('-v {0}', inputs.version) || '' }}
+
+      - name: Install Fly CLI
+        uses: superfly/flyctl-actions/setup-flyctl@v1
+
+      - uses: actions/checkout@v4
```
sodic: What's this?

.github/workflows/deploy-examples.yml
```diff
@@ -0,0 +1,68 @@
...
+        required: false
+
+  # TODO: REMOVE THIS BEFORE MERGING, JUST FOR TESTING
+  push:
+    branches:
+      - 2831-automatically-deploy-all-example-apps-on-release
+
+concurrency:
+  group: "deploy-examples"
+  cancel-in-progress: true
```
sodic, line 26: I'm guessing this tells github to run all jobs in the group in parallel.

But where do we tell it what's in the group. In the file name, or is the `concurrency` dictionary enough?

sodic (commented):
I like where this is going!

Left some legit review questions and some n00b questions to better educate myself on GH actions.

.github/workflows/deploy-examples.yml
cprecioso: > The picture mentions GHC and Node versions

The picture is from another workflow. As I mentioned, I am not able yet to get the workflow dispatch trigger to work until the workflow hits `main`

---

> Also, should we figure out the Wasp version from the version field in the app, have a single source of truth?

I don't think so. If we want to e.g. release these projects with a pre-release version, we need a way of saying so. Otherwise, these examples that live in the wasp repo are using the repo's version, so, latest.

.github/workflows/deploy-examples.yml
cprecioso: Can you be more specific?

.github/workflows/deploy-examples.yml
cprecioso: Checking out the repo code

.github/workflows/deploy-examples.yml
cprecioso: > this tells github to run all jobs in the group in parallel

No. This is about _workflow runs_, and not _jobs_ inside those runs. What this means is that, if a new run of this workflow is queued, this run (and both jobs inside it) will get immediately cancelled and run the newer one.

> what's in the group

Everything with the same `concurrency.group` value is in the same group.

.github/workflows/deploy-examples.yml
sodic: Sure, here goes.

Keep in mind that if something sounds like it doesn't make sense, it's likely my lack of knowledge :)

What looks like duplication:
1. We list our example app here, and we list them in the `examples` folder.
2. The app URL is explicitly listed here, but it's also implicitly defined in our `fly-client.toml` and `fly-server.toml`

Is there a way we can avoid changing this action when we change the `fly-*.toml` files or the example's name.

Ideally, this is what I'd like (not sure if possible):
1. Each app has its deployment workflow that triggers when that app is, and is defined in the respective app's folder (I'm guessing this would make it harder to trigger all of them, but it would also give us the option to trigger only one).
2. There's no duplication of app names (e.g., `waspleau`) or Fly URLs (https://websockets-voting-client.fly.dev/).

.github/workflows/deploy-examples.yml
sodic: What about `waspello`?

.github/workflows/deploy-examples.yml
```diff
@@ -0,0 +1,68 @@
+name: Deploy example apps to Fly
+
+on:
+  # Allow calling this workflow from other workflows
+  workflow_call:
+    inputs:
+      version:
+        type: string
```
sodic, line 8: Are there any other input types (e.g., multiple choice based on releases) or is this just wishful thinking?

sodic (commented):
Did another pass

.github/workflows/deploy-examples.yml
sodic: Makes sense.

One last question: why not merge this to `main` directly? Why start from `release`?

.github/workflows/deploy-examples.yml
cprecioso: I could not find it in Fly so I left it for later and forgot about it. Found out it's currently deployed on Heroku + Netlify (and the Heroku is down). Should we move it to Fly?

.github/workflows/deploy-examples.yml
cprecioso: My reasoning is this is using the released version of wasp, so it should also use the release version of the apps, not the one in `main`.

.github/workflows/deploy-examples.yml
cprecioso: Only string, number, and boolean

.github/workflows/deploy-examples.yml
cprecioso: > Each app has its deployment workflow that triggers when that app is, and is defined in the respective app's folder (I'm guessing this would make it harder to trigger all of them, but it would also give us the option to trigger only one).

Mmmm having the deployment workflow in each folder is not possible, they all need to be in `<root>/.github/workflows`. We could create a new workflow per app, or require all of them have a `deploy.sh` script, but tbh it does look like overkill. We just need to run `wasp deploy fly deploy` in the folders.

> There's no duplication of app names (e.g., waspleau) or Fly URLs 

Done

cprecioso:
@sodic addressed a change and responded to comments. We can merge this and see what we do with waspello once there's a reply here https://discord.com/channels/686873244791210014/1351908377533354025/1388152911862829056 (that will be just creating the fly files, and adding the app to the list in this workflow)

.github/workflows/deploy-examples.yml
sodic: But it will still end up on the main branch at some point. A week from now, this action will be both in `main` and on `release`, so why not start from main and follow the normal flow immediately?

.github/workflows/deploy-examples.yml
sodic: I think we should, yes!

.github/workflows/deploy-examples.yml
sodic: Classic :(

.github/workflows/deploy-examples.yml
```diff
@@ -0,0 +1,72 @@
...
+        run: |
+          wasp deploy fly deploy --org wasp
+        env:
+          FLY_API_TOKEN: ${{ secrets.FLY_API_TOKEN }}
+
+      - name: Get deployed app info
+        id: app-info
+        working-directory: ${{ matrix.app-dir }}
+        run: |
+          echo "url=https://$(flyctl status -c fly-client.toml --json | jq -r .Hostname)" >> $GITHUB_OUTPUT
```
sodic, line 67: Will `jq` be available in the GitHub action?

waspc/README.md
```diff
@@ -415,6 +415,21 @@ During CI, we build and test Wasp code on Linux, MacOS and Windows.
 
 If commit is tagged with tag starting with `v`, github draft release is created from it containing binary packages.
 
+We also have a workflow for deploying example apps to Fly.io (`deploy-examples.yml`). This workflow can be run manually from the GitHub UI and should typically be run from the `release` branch after publishing a new release to ensure the deployed examples are using the latest stable version of Wasp.
```
sodic, line 398: Only team members dan do this, right?

.github/workflows/deploy-examples.yml
sodic: I'm wondering whether we should also deploy the tutorials? Wdyt?

.github/workflows/deploy-examples.yml
cprecioso: Yes! It's preinstalled, they've had it since the beginning. https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#:~:text=Haveged%201.9.14-,jq,-1.7

waspc/README.md
cprecioso: Yep, from the [docs](https://docs.github.com/en/actions/how-tos/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow)

> Write access to the repository is required to perform these steps

.github/workflows/deploy-examples.yml
cprecioso: > I think we should, yes!

On it

> I'm wondering whether we should also deploy the tutorials? Wdyt?
 
No strong opinion, but leaning towards no. Examples are deployed so people can play with them without doing it yourself. The value in the tutorials is that people _do_ do them themselves. And the end state isn't something really worth showcasing.

.github/workflows/deploy-examples.yml
cprecioso: Okay. I will keep it on `release` for now while doing tests, but change to `main` before merging.

cprecioso:
@sodic ready

.github/workflows/deploy-examples.yml
```diff
@@ -0,0 +1,73 @@
...
+    runs-on: ubuntu-latest
+
+    environment:
+      name: ${{ matrix.app-dir }}-production
+      url: ${{ steps.app-info.outputs.url }}
+
+    steps:
+      - name: Install Wasp
+        run: |
+          curl -sSfL https://get.wasp.sh/installer.sh | sh -- ${{ inputs.version && format('-v {0}', inputs.version) || '' }}
```
sodic: This is the only place where we're handling the default system for Wasp version, right?

Any chance we could do this and the formatting in a more obvious place (e.g., when specifying the versoin argument at the top of the file)?

FYI, this works too:

image

web/blog/2021-12-02-waspello.md
```diff
@@ -13,7 +13,7 @@ import ImgWithCaption from './components/ImgWithCaption'
 ![Enter Waspello](../static/img/waspello-screenshot.png)
 
 <p align="center">
-    <Link to={'https://waspello.netlify.app/'}>Try Waspello here!</Link> | <Link to={'https://github.com/wasp-lang/wasp/blob/release/examples/waspello/main.wasp'}>See the code</Link>
+    <Link to={'https://waspello-app-client.fly.dev/'}>Try Waspello here!</Link> | <Link to={'https://github.com/wasp-lang/wasp/blob/release/examples/waspello/main.wasp'}>See the code</Link>
```
sodic: Wow, props for the thoroughness here!

.github/workflows/deploy-examples.yml
cprecioso: Unresolving this conversation so I remember

.github/workflows/deploy-examples.yml
cprecioso: Nope, there's not really a set place for GitHub workflows to set variables or stuff like that. I don't see how this might be confusing, what are you worried about?

sodic: > I don't see how this might be confusing, what are you worried about?

The input definition says `"Wasp version to use for deployment (default to latest)"`, but that's not right. The default value is not specified, and `inputs.version` doesn't default to latest (I'm guessing it defaults to `null`).

What truly defaults to latest is this one specific piece of logic that uses `inputs.version`. 

This makes the "defaults to latest" comment at the top potentially misleading:
- It's perfectly fine from the user's perspective
- From the perspective of someone changing this script, it sounds like a promise but is actually a request saying "make sure you default to latest when this is `null`."

Say we add a new step in the workflow that uses `inputs.version`. We'd have to remember to handle the "default to latest case."

I understand that the "defaults to latest" message is primarily meant for the user, so I'd still keep it there (and ideally handle the default case as soon as possible).

> Nope, there's not really a set place for GitHub workflows to set variables or stuff like that.

Here's my idea, not sure if it's better or if it works. Let me know what you think.

```yaml
  # Allow calling this workflow manually from the GitHub UI
  workflow_dispatch:
    inputs:
      version:
        description: "Wasp version to use for deployment (default to latest, e.g., 0.16.7)"
        required: "" # We are relying on the install script treating an empty version string as the latest version 
        
  steps:
      - name: Install Wasp
        run: |
          curl -sSfL https://get.wasp.sh/installer.sh | sh -- ${{ format('-v {0}', inputs.version)}}
```
Changes:
- The bash script no longer deals with picking the default version.
- The user knows the format of the string they need to input (e.g., `v0.16.7` vs `0.16.7`)
- The programmer explicitly sees the default value for `inputs.version` and its assumptions. They may still need to make some changes depending on what they want to accomplish, but it's difficult to miss.

cprecioso: Makes sense, on it!

Issue #2859, opened 2025-06-27
Extract `wasp-cli` building logic into a seperate action


FranjoMindek:
Similarly how we have [`setup-haskell`](https://github.com/wasp-lang/wasp/blob/main/.github/actions/setup-haskell/action.yaml) action.
We should extract out the logic for building the whole `wasp-cli` into a seperate action.

The reason is that with headless tests being added as separate workflows, both to `wasp-lang/wasp` and `wasp-lang/starters`, 
we will be reusing the `wasp-cli` logic a lot. 

We want to to provide a reproducible `wasp-cli` environment for all workflows which depend on it.

We should also look if there any cache possibilities as to speed up the process on all workflows.
Though this can be done as a separate PR.

FranjoMindek:
For `wasp` repo we can just use `run` script.

PR #2860, opened 2025-06-27
Extract `setup-wasp-cli` action from `waspc-ci.yaml`

FranjoMindek:
Fixes https://github.com/wasp-lang/wasp/issues/2859

I'm concerned about the lack of control we got on individual steps of the process, but that matters only if we need to access them.

In `waspc-ci.yaml`, we do access them to limit which steps run on unix and which only on `ubuntu`.

But do we really don't want to build the `wasp-cli` on windows and macos?
Do we have problem with those?

Related files:
[`setup-haskell` action](https://github.com/wasp-lang/wasp/blob/main/.github/actions/setup-haskell/action.yaml) 
[`waspc-ci.yaml`](https://github.com/wasp-lang/wasp/blob/main/.github/workflows/waspc-ci.yaml)


.github/actions/setup-wasp-cli/action.yaml
```diff
@@ -0,0 +1,77 @@
...
+  3. Compiles TypeScript packages and integrates them into the Cabal build system
+  4. Builds all Wasp dependencies and the Wasp CLI itself
+  5. Installs the Wasp CLI binary, making it available for subsequent workflow steps
+
+inputs:
+  node-version:
+    description: |
+      The version of Node to install.
+    required: false
+    default: "lts/*"
```
FranjoMindek, line 17: Latest LTS version, I like this more than our `latest` approach in `waspc-ci.yaml`.

.github/actions/setup-wasp-cli/action.yaml
```diff
@@ -0,0 +1,77 @@
...
+    description: |
+      The version of Cabal to install.
+    required: false
+
+  cabal-project-dir:
+    description: |
+      The working directory for the action.
+    required: false
+
+  extra-cache-key-segment:
```
FranjoMindek: Do we prefix this anyway to signal it's used by `setup-haskell`?

.github/actions/setup-wasp-cli/action.yaml
```diff
@@ -0,0 +1,77 @@
...
+        cabal-version: ${{ inputs.cabal-version }}
+        cabal-project-dir: ${{ inputs.cabal-project-dir }}
+        extra-cache-key-segment: ${{ inputs.extra-cache-key-segment }}
+
+    - name: Set up Node
+      uses: actions/setup-node@v4
+      with:
+        node-version: ${{ inputs.node-version }}
+
+    - name: Compile TS packages and move it into the Cabal data dir
```
FranjoMindek, line 62: In `waspc-ci.yaml`, we call this only on mac and linux.

.github/actions/setup-wasp-cli/action.yaml
```diff
@@ -0,0 +1,77 @@
...
+      shell: bash
+      working-directory: waspc
+      run: cabal build --enable-tests --enable-benchmarks --only-dependencies
+
+    - name: Build wasp code
+      shell: bash
+      working-directory: waspc
+      run: cabal build all
+
+    - name: Install Wasp CLI
```
FranjoMindek, line 77: In `waspc-ci.yaml`, we call this only on linux.

.github/actions/setup-wasp-cli/action.yaml
```diff
@@ -0,0 +1,77 @@
...
+  extra-cache-key-segment:
+    description: |
+      An extra key to use for the Haskell cache. Mainly used to keep glibc and musl
+      caches separate, as they are not compatible.
+    required: false
+
+runs:
+  using: composite
+
+  steps:
```
FranjoMindek, line 44: Like in `setup-haskell` we assume the caller used the checkout action prior to calling this.

cprecioso:
@FranjoMindek before I go and review, can you expand a bit on what is the difference between this and https://github.com/wasp-lang/wasp/blob/main/.github/workflows/waspc-build.yaml ?

FranjoMindek:
> before I go and review, can you expand a bit on what is the difference between this and [main/.github/workflows/waspc-build.yaml](https://github.com/wasp-lang/wasp/blob/main/.github/workflows/waspc-build.yaml?rgh-link-date=2025-07-01T14%3A06%3A43Z) ?

My core idea was just to abstract the steps up to the point of installing `wasp-cli` (to enable us to do actions with the development version of the wasp) on the machine. 
This is because multiple workflows would use the same logic above.

When we want to test if the application will build and when we want to run unit, `todoApp` headless and e2e tests we need it (`waspc-ci-yaml`).
When we want to run `starters` headless tests we need it. <- a issue I have on me
When we want to run `examples` headless tests we need it. <- a issue I have on me

I missed `waspc-build.yaml` because I followed `wasc-ci.yaml` and though If anything like that existed, that workflow would use it.

I would still have to make sure this action is runnable on every OS.

.github/actions/setup-wasp-cli/action.yaml
```diff
@@ -0,0 +1,77 @@
...
+    - name: Build wasp code
+      shell: bash
+      working-directory: waspc
+      run: cabal build all
+
+    - name: Install Wasp CLI
+      shell: bash
+      working-directory: waspc
+      id: headless-install-wasp-cli
+      run: cabal install
```
cprecioso, line 81: Why do all these in separate steps? Won't `cabal install` just build whatever it needs?

.github/actions/setup-wasp-cli/action.yaml
cprecioso: Safe bet! It's required in order to call an action in the same repo: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-an-action-in-the-same-repository-as-the-workflow

.github/actions/setup-wasp-cli/action.yaml
cprecioso: Yeah pls

.github/actions/setup-wasp-cli/action.yaml
cprecioso: Ah yeah it makes sense! Care to update it too in the `wasp-ci`?

cprecioso (commented):
LGTM, I'd like to see this running before merging, can we get that?

What do we want to do about some steps only being called in Linux or Mac?

cprecioso:
> I missed waspc-build.yaml because I followed wasc-ci.yaml and though If anything like that existed, that workflow would use it.

Does it make sense to update `wasp-ci` to use this? And `wasp-build`? If so, I'd do it in this same PR.

cprecioso:
Idea: This action performs the steps up to (and incl.) `cabal build`. Then you can just call `cabal install` when we need it installed in your workflows, `cabal test` in `wasp-ci`, or `./generate-packages.sh` (i don't remember the name) in `wasp-build`. What do you think? Does it make sense?

.github/actions/setup-wasp-cli/action.yaml
FranjoMindek: I think we do it so it's easier to debug where it failed exactly, but what you say is true.
I asked martin about the two separate build commands, but he doesn't fully remember too.

.github/actions/setup-wasp-cli/action.yaml
FranjoMindek: ~~So you would rather include it inside of the action, rather than let the caller checkout before using this action?~~
edit: I think I misunderstood you, you wanted to say you need to checkout anyways to have access to the action?

.github/actions/setup-wasp-cli/action.yaml
FranjoMindek: Do we want to update it everywhere then? `check-formatting.yaml`, `waspc-build.yaml`. `waspc-ci.yaml`?
Hm too fragile for `waspc-build.yaml`?

.github/workflows/test-setup-wasp-cli.yaml
```diff
@@ -0,0 +1,86 @@
+name: Test Setup Wasp CLI
```
FranjoMindek, line 1: This file will be deleted before merging.

.github/actions/setup-haskell/action.yaml
```diff
@@ -23,7 +23,7 @@ inputs:
     required: false
     default: "waspc"
 
-  extra-cache-key-segment:
```
FranjoMindek, line 26: I went ahead and renamed this in `setup-haskell` action too.
I think it's much more clear now.

FranjoMindek:
I've added a temporary workflow to test out the action.
It seems that everything passes even without OS specific steps, except that `windows` fails on `npm`?
This is weird considering we have to build `Typescript` packages using `npm` beforehand and that passes.
Most likely an issue on how windows handles PATH when we execute `npm --version` through wasp.
https://github.com/wasp-lang/wasp/actions/runs/16019963985

Other than that, a most probable reason why we skip stuff on `macos` and `windows` would be the duration.
We only install on `ubuntu` since we only ever run tests on `ubuntu`.

`ubuntu`: 16 min
`macos`: 28 min
`windows`: 35 min

That is for installation only, no tests.

FranjoMindek:
We could also refactor this to be `build-wasp-cli` action instead as you've said.
That would make it more reusable across current workflows.

FranjoMindek:
Additional question, is this PR necessary if we have `run` script. I overlooked this when creating it.
Should transferring `waspc-ci.yaml` to use `./run build:all` instead of manual commands be the solution?

One downside I see is it relies on `bash`, which would make windows support harder? 
I think windows supports bash because it ships with `git`. But what subset?

Another thing is that we can't have precise OS-specific control per command.
But I don't know why we avoided certain commands on e.g. `windows` before?

Edit: 
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#defaultsrunshell
image

image

.github/actions/setup-wasp-cli/action.yaml
cprecioso: Yeah do it everywhere I think, better than `latest`

.github/actions/setup-wasp-cli/action.yaml
cprecioso: yep! i'm saying that you are correct, the assumption you said in the first comment is right, as they will have to checkout the repo to use the action locally. RAW

.github/actions/setup-wasp-cli/action.yaml
```diff
@@ -0,0 +1,81 @@
...
+
+    - name: Build wasp code
+      shell: bash
+      working-directory: waspc
+      run: cabal build all
+
+    - name: Install Wasp CLI
+      shell: bash
+      working-directory: waspc
+      id: headless-install-wasp-cli
```
cprecioso, line 80: I don't think this step needs an `id`, right? We're not exposing its output or anything

.github/actions/setup-wasp-cli/action.yaml
cprecioso: Personal opinion is that, I'd try to do it in one step. But I understand if you want to skip trying it in case it brings trouble. Not blocking -- RAW

.github/actions/setup-wasp-cli/action.yaml
cprecioso: Wait this is also expected to run on other repos right? Then I'd actually check out the repo inside this action.

Martinsos:
> * Using `./run` instead of `cabal install`: I don't care. I like to use standard interfaces more than custom ones, so I prefer `cabal install`, but don't really mind either.
> * Wasp on Windows: do you want to work on that? If no, I'd add a message to this action that it doesn't run on Windows (or better yet, fail early if it is detected)
> 
> I think probably @Martinsos has more wisdom on the last two points

I am commenting only on the last two things, as I haven't read enough yet about this issue to be able to comment on the rest (and I might even be missing context to answer these two things, but I will give it a try):
1. ./run vs cabal -> I also prefer using standard tools, but since Wasp is more than just a simple cabal project, with TypeScript parts and what not, then ./run captures info how to actually correctly build the project. So yeah, use ./run, otherwise you will be duplicating the knowledge of how wasp CLI is built. Side note: maybe we could cram all that into cabal, I don't know if it is flexible enough, to also build those TypeScript projects, but I am not a fan of that approach and we are not doing it for now.
2. Windows -> not sure. Wasp should compile on Windows and if env is correctly set, it should even work in the biggest part. But we haven't compiled it for Win for some time so probably this needs some updating. You could fail it explicitly, sure, for now.

FranjoMindek:
From the current direction we can see that for internal use there is no real benefit to this action.
Either we want to run specialized workflow commands, or using the `run` script is enough.

For external wasp repos, we are currently moving towards monorepo, so there is not much benefit to gain.
Also most likely the will depend on `wasp` rather than `wasp-cli`, which doesn't  require any difficult set up.

I will probably take some useful things we learned and apply it to existing scripts.
For now we will close the PR and the issue.

Martinsos:
Good call on recognizing this!

Issue #1838, opened 2024-02-29
Investigate NPM workspaces and our dependency setup


sodic:
During the restructuring and the release of Wasp 0.12.0, we've ran into issues caused by our project's node_modules hierarchy: #1640.

Our setup is documented in https://github.com/wasp-lang/wasp/issues/1595 and #1429.

Investigate whether we could improve our setup (read: make it more standard) with NPM workspaces (or anything else the industry might be using).

@infomiho already attempted this in #1653.

Other resources:
  - https://iws.io/2022/invalid-hook-multiple-react-instances

cprecioso:
Done in #3130

PR #2865, opened 2025-06-30
Prevent Wasp from installing Vite 7 in web app

sodic:
Fixes an issue of Vite 7 clashing with older versions of Node:
- We pin down our Vite version in the user's (root) `package.json`.
- `@vitejs/plugin-react`, a dependency of web app, installs Vite as a transient dependency.
- This installation of Vite bypasses the version specified in the root `package.json` and ends up resolving to Vite 7
- Vite 7 does not work with Node older than 20.19.0, while Wasp officially claims to support Node > 20.0.0

image


FranjoMindek (approved):
If the problem is caused solely because of web app vite, this looks okay. I assume by your first point that all other possibilities of this problem already have fixed versions?

cprecioso (commented):
I am not satisfied with this solution, but I won't block if we want to have it as a temporary workaround.

**Why is Vite 7 installed?**
`@vitejs/plugin-react` asks for it as a `peerDependency`, not a `dependency`. `npm` realizes we don't have `vite` installed and does it for us, using the latest allowed version (this is `npm` trading correctness for convenience). **BUT, even in the latest version `@vitejs/plugin-react` asks for `^4.2.0 || ^5.0.0 || ^6.0.0 || ^7.0.0-beta.0`.** This means that if we install by ourselves an older version, `npm` will use that one. Remember, it is a _peer dependency_, so **we** are supposed to provide that as a dependency injection of sorts.

This means the correct solution here would be not to pin `@vitejs/plugin-react`, but to _add `vite`_ with the version we want it. It is more correct, as that is the dependency we actually want to pin.

**First solution: add `vite <7` to manifest**
This doesn't work as `wasp-cli` realizes that Vite's versions are not consistent.

```
- Error: Dependency conflict for user dependency (vite, ^4.3.9): Version must be set to the exactly the same version as the one wasp is using: <7
```

And in here, I think the actual, _correct_ solution is quite clear:

**We should add `vite ^4.3.9` to the web-app package.json**

This is already the version we make user's projects have, so this is actually the version we should have been using all along inside the web-app, and anywhere else.

Until we have real dependency sharing between subprojects (#1838), that'd be the workaround I'd be happy with.

cprecioso:
From an internal message by @infomiho:

>  We have vite as a dep for WebApp, but we also have a system for deduping deps in Wasp between user's package.json and WebApp deps. The vite dep gets dropped by that system from WebApp's package.json. 
>
> What you suggested was considered on a call Filip and I had (we realised that we'd have to mess with the dedupe system which felt more trouble than it's worth), but we dropped it in favor of the simpler fix for now because we ultimately want npm workspaces 🙂

PR #2866, opened 2025-07-01
Add tailwind plugin to root prettier

FranjoMindek:


FranjoMindek:
We don't want to remove sub project's prettier files since we want those to be included in the starter template.

~~There is no diff to the current code base by running the format currently, because there is no tailwind code in the code base yet.~~ 
Edit: we have tailwindcss in `todoApp` and many more, I forgot to include the plugin at all.

FranjoMindek:
We maybe want to add `--config` flag to force the root styles everywhere.

```ts
    "prettier:check": "prettier --ignore-unknown --check --config .prettierrc .",
    "prettier:format": "prettier --ignore-unknown --write --config .prettierrc ."
```

Because root prettier encounters prettier settings in `basic` starter template it tries to use those, but it can't because its not really a fully valid project.

https://github.com/wasp-lang/wasp/actions/runs/15995436523/job/45117676864
( In that PR I tried to add this plugin to root beforehand, while it fixes its search for plugin, it doesn't fix the improper project issues. )

Adding the flag above fixes the problem that PR has.

This means that our base prettier files are the source of the truth for the prettier formatting in the whole project.

FranjoMindek:
Pre-format commit: [`ffb2391` (#2866)](https://github.com/wasp-lang/wasp/pull/2866/commits/ffb2391e4eaaa0771db59fdf79d67f9e7a23e60e)

mage/src/client/pages/ResultPage.jsx
```diff
@@ -779,18 +759,11 @@ function Feedback({ projectId }) {
...
-                                active
-                                  ? "ring-2 ring-white ring-opacity-60 ring-offset-2 ring-offset-sky-300"
-                                  : ""
-                              }
-
-
-                              ${checked ? "bg-sky-900 bg-opacity-75 text-white" : ""}
-                              cursor-pointer px-3 py-2 shadow-md focus:outline-none
-                              rounded-md
-                            `}
```
cprecioso, line 793: Is the loss of whitespace here significant?

cprecioso:
Looks good!

Good job on the `--config` option, that's some good thinking and will make runs faster!

Just left one question, non blocking, approving meanwhile.

**How I would merge this:**

> These big repo-wide changes are a bit bothersome for people with open branches. I already made the mistake with the initial Prettier PR, so I'm leaving here the workflow that would have avoided it.

1. Squash commits until you have only two:

    ```mermaid
    gitGraph
      commit
      branch this-pr
      commit id: "Config"
      commit id: "Format"
    ```

    - **First** Installing the plugin, configuring it, adding `--config` flag
    - **Second** Formatting

2. When merging, **rebase don't squash** so both commits are left in the history

3. Everyone updates their branches

	```mermaid
	gitGraph
		commit
		branch other-pr
		commit
		commit
		checkout main
		commit id: "Config"
        checkout other-pr
		merge main
        commit id: "Run format myself"
		checkout main
        commit id: "Format"
        checkout other-pr
        merge main tag: "Use `-s ours` to ignore all conflicts"
        commit
	```

mage/src/client/pages/ResultPage.jsx
FranjoMindek: It seems fine

cprecioso:
For updating your branches while avoiding conflicts:

```sh
config_commit=4fbb0ef
format_commit=8141302

# from your branch, root dir of the repo
git fetch origin main
git merge $config_commit 
# >> solve conflicts <<
npm i
npm run prettier:format
git commit -am Format
git merge $format_commit -s ours 
```

(`--strategy ours` makes it so we take our current version of the tree and disregard the incoming one. it should be fine as it's only formatting that we just did, but the merged will be "solved" in the git history)

Issue #2833, opened 2025-06-11
Create headless tests for starter templates


sodic:
We should have proper headless tests for our starter tempaltes. See parent issue for more details.

Martinsos:
Open saas already has them.
For others, we should add them, but they don't have to be super deep -> we are not testign Wasp features here, just that the starters themselves are looking ok / working.

PR #2868, opened 2025-07-01
Add e2e tests to starter templates

FranjoMindek:
Fixes: https://github.com/wasp-lang/wasp/issues/2833

Since starter templates are not a valid Wasp project by themselves,
we need to use Wasp CLI to generate the projects to test them out.
 
There should be two possible executions of e2e tests:
1) `StarterWithoutHeadlessE2ETests`
    - this means that we run tests under `waspc/starters-headless-e2e-tests/tests` folder for this starter, for each starter we run tests without tags, and tests tagged for that specific starter
    - e.g. following tests will only execute for the `basic` starter
    ```ts
	test.describe(
	  "basic only",
	  {
	    tag: "@basic",
	  },
	  () => {
	    // tests here...
	  },
	);
	```

2) `StarterWithHeadlessE2ETests`
    - this means that we run the tests which are shipped together with the starter template, example being `saas` starter
    - in that case we will provide information to the script on how to run the shipped e2e tests, which will call it instead of calling the local ones

Currently there is no possibility of running the both test variations for a single starter.

In this PR I won't implement `StarterWithHeadlessE2ETests` route fully.
This is because current headless e2e tests in `saas` starter are not made to be run with `wasp-app-starter` which is the current standard. So I will delay the implementation until we refactor the `saas` headless e2e tests, and we have an example upon which we can experiment to make the best solution for the problem.

### DIFF
--- a/waspc/starters-e2e-tests/src/main.ts
+++ b/waspc/starters-e2e-tests/src/main.ts
@@ -17,9 +17,7 @@ async function runAllStartersE2ETests(): Promise<void> {
 
   for (const starterE2ETests of STARTERS_E2E_TESTS) {
     console.log(
-      chalk.bold(
-        `\nRunning tests for ${starterE2ETests.starterName} starter`,
-      ),
+      chalk.bold(`\nRunning tests for ${starterE2ETests.starterName} starter`),
     );
     await runStarterE2ETests({
       waspCliCommand,
--- a/waspc/starters-e2e-tests/src/starter-e2e-test-runner.ts
+++ b/waspc/starters-e2e-tests/src/starter-e2e-test-runner.ts
@@ -127,9 +127,7 @@ async function runBuildE2ETests(
     return;
   }
 
-  console.log(
-    `Running BUILD e2e tests for ${templateName} starter...`,
-  );
+  console.log(`Running BUILD e2e tests for ${templateName} starter...`);
   const waspAppRunnerBuildEnv = {
     ...process.env,
     WASP_APP_PATH: tempWaspProjectPath,
--- a/waspc/starters-e2e-tests/src/starter-e2e-tests.ts
+++ b/waspc/starters-e2e-tests/src/starter-e2e-tests.ts
@@ -3,9 +3,7 @@ type StarterTemplateName = "minimal" | "basic" | "saas";
 type WaspProjectRelativePath = `.${string}`;
 type IncludedTestsRelativePath = `.${string}`;
 
-export type StarterE2ETests =
-  | StarterWithoutE2ETests
-  | StarterWithE2ETests;
+export type StarterE2ETests = StarterWithoutE2ETests | StarterWithE2ETests;
 
 export type StarterWithoutE2ETests = {
   starterName: StarterTemplateName;


waspc/starters-headless-tests/run-starter-headless-tests.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
+
+# Script to run E2E tests for a given Wasp template.
```
FranjoMindek: We need a script because we need to generate projects from the starter templates.
Starter templates are not valid wasp projects:

1. Missing files which are part of starters skeleton
2. `main.wasp` has interpolation placeholder values for e.g. version, which is invalid wasp syntax

waspc/starters-headless-tests/tests/health-check.spec.ts
```diff
@@ -0,0 +1,6 @@
+import { expect, test } from "@playwright/test";
+
+test("health check", async ({ page }) => {
```
FranjoMindek: "Text the app runs successfully and works."

waspc/starters-headless-tests/playwright.config.ts
```diff
@@ -0,0 +1,75 @@
...
+  },
+
+  /* Configure projects for major browsers */
+  projects: [
+    {
+      name: "chromium",
+      use: { ...devices["Desktop Chrome"] },
+    },
+    // {
+    //   name: "firefox",
```
FranjoMindek: Since we are only testing that the app works, mobile is not really needed.
Though it might be useful to try it with firefox and webkit since we did have a firefox only bug?

waspc/starters-headless-tests/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "headless-tests",
+  "version": "1.0.0",
+  "description": "",
+  "author": "",
+  "license": "ISC",
+  "main": "index.js",
+  "keywords": [],
+  "scripts": {
+    "test": "for starter in basic minimal; do ./run-starter-headless-tests.sh $starter wasp || exit 1; done",
```
FranjoMindek: Is there a better way for this?

waspc/starters-headless-tests/playwright.config.ts
FranjoMindek: After we decide I will of course remove all the commented code.

waspc/starters-headless-tests/run-starter-headless-tests.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
```
infomiho: Maybe the script can run `npm install` before starting to ensure Playwright is installed.

We could run `npx -y playwright test ` to avoid asking the user if they want to install the Playwright CLI.

waspc/starters-headless-tests/run-starter-headless-tests.sh
```diff
@@ -0,0 +1,70 @@
...
+  if template_uses_sqlite; then
+    echo "Skipping BUILD tests for ${TEMPLATE_NAME} project (sqlite detected in schema.prisma)."
+    return
+  fi
+
+  echo "Running BUILD headless tests for ${TEMPLATE_NAME} project..."
+  DEBUG=pw:webserver E2E_APP_PATH="./${TEMP_PROJECT_NAME}" WASP_CLI_CMD="${WASP_CLI_CMD}" HEADLESS_TEST_MODE=build npx playwright test
+}
+
+template_uses_sqlite() {
```
infomiho: I feel like the `sqlite` detection is kinda fragile, can we do better? 

Do we want to upgrade from bash to Node.js and use `zx` here - we can reuse same logic for detecting SQlite from the Wasp App Runner: https://github.com/wasp-lang/wasp/blob/main/wasp-app-runner/src/waspCli.ts#L94

waspc/starters-headless-tests/package.json
```diff
@@ -0,0 +1,18 @@
...
+    "test:dev": "for starter in basic minimal; do ./run-starter-headless-tests.sh $starter wasp-cli || exit 1; done"
+  },
+  "devDependencies": {
+    "@playwright/test": "^1.53.0",
```
infomiho: I'd suggest we pin the Playwright version so browsers can be reused (Playwright versions and browser version are tied together).

We use the same version across kitchen sink app and example apps, I believe it's `"@playwright/test": "1.51.1"`

waspc/starters-headless-tests/tests/health-check.spec.ts
infomiho: I'd be much more comfortable with not testing starters manually we did something like added a new task with the `basic` template.

waspc/starters-headless-tests/run-starter-headless-tests.sh
infomiho: Can we test `saas` starter with this as well? 
1. There is an issue that the `saas` template has the app in the `/app` folder. 
2. Also the `saas` starter needs some `.env` vars set up to be able to run the app.
3. We can't test the `saas` starter in production mode with email sender set as `Dummy`.

infomiho: We could be declarative around which tests we are doing, and define an object:
- `minimal` / [`sqlite`] / use `.` as dir
- `basic` / [`sqlite`] / use `.` as dir
- `saas` / [`postgres`] / use `app` as dir

This way we don't bother with parsing the db type and we can declare stuff like app dir.

Also, do we see some value in testing all the starters with both database types?

waspc/starters-headless-tests/run-starter-headless-tests.sh
infomiho: Do we want to use `zx` regardless of reusing the SQLite detection logic, I feel like it's a quite common tool we use (Wasp App Runner, deploy package).

waspc/starters-headless-tests/run-starter-headless-tests.sh
FranjoMindek: I'd rather keep that out. Same of `wasp-app-runner`.
These are the project dependencies. I don't want them to be handled by in-project script. Does that make sense? Another thing would be if it was an script outside of the project. Don't know how to explain it the best.

FranjoMindek: I thought `open-saas` had it's own tests, I always looked at `open-saas` and `ai-generated` as bit of special cases compared to `basic` and `minimal`.

Hm, in my personal opinion, we could try to have a single script test ALL of the starters, but I wouldn't mix with this script. This script should work for current and future non-special wasp projects. Exceptions should be handled exceptionally.

waspc/starters-headless-tests/run-starter-headless-tests.sh
FranjoMindek: I can switch to `wasp info`, did it same for examples CI tests, didn't realize that we had this command when I wrote this code.

waspc/starters-headless-tests/run-starter-headless-tests.sh
FranjoMindek: It's a nice idea, but I feel weird about mixing `open-saas` logic here when it has it's own repo with it's own headless tests. I would probably also want to run those tests instead of the ones in here, that sone additional info. The more exceptions that we introduce the worse.

waspc/starters-headless-tests/tests/health-check.spec.ts
FranjoMindek: We can do that, it feels a bit weird to copy the whole `helpers.ts` from `todoApp` to do this, but we will need to do auth to test anything in `basic`.

Still I'm not 100% sure we need to do that.
We just want to test that our starters are not broken. What is testing starters and what is testing already tested wasp functionalities.

waspc/starters-headless-tests/run-starter-headless-tests.sh
FranjoMindek: For now I've added logic in bash for this.
We can think if node is better suited for this job.
It's mostly calling commands expect for the "check sqlite part", that is why I leaned into bash.

waspc/starters-headless-tests/tests/basic.spec.ts
FranjoMindek, line 1: This was a test to check if we could tag tests based on starter template, and we can.

waspc/starters-headless-tests/tests/health-check.spec.ts
FranjoMindek: Maybe that should be a separate issue. So we get going with at least a health check headless test before we go more vertical into specific starters.

I did a POC, and we can run starter specific tests in addition to the base health check test.
(it's actually commited to the PR currently).
If we do go with another issue, I will revert those and give them as example on the issue.

waspc/starters-headless-tests/run-starter-headless-tests.sh
infomiho: I feel like it's a part of the script to ensure the deps are installed, I don't see the clear boundary between the script that's using the project to do something and the script installing the deps without which it would fail. 

What about `npx -y playwright test` change? Adding the `-y`

FranjoMindek: I could create another script.
`run-starter-headless-tests.sh` does what it names says, runs headless tests for a single starter.

I could do something like: `setup-and-run-starters-headless-tests.sh`.
It would:
1) take the iteration logic out of `package.json`, which does bother me
2) take care of the setup needed to run headless tests before calling `run-starter-headless-tests.sh`.

Alternative is to do all of above in a single file:
1) we move current `run-starter-headless-tests.sh` logic into a function
2) we add above setup and iteration logic
3) we rename the file to reflect new behavior

infomiho: > take the iteration logic out of package.json, which does bother me

Just saw this - I missed it before, yep, it's kinda weird to see that much logic in a npm script.

I vote for a single script with functions inside of it. It feels cleaner.

infomiho: Even though Open Saas has it's own e2e tests - I still had the urge to create a new project with the `saas` template and click around. I guess I wanted to make sure what the users will get is working regardless if the e2e tests pass in the Open Saas repo.

If we use the rule of thumb "let's automate all the steps we do manually", I'd like to use us testing the `saas` template as well here. But I'm not going to insist on it in this PR, we can make another issue that will explore the e2e tests situation in Open Saas.

If we don't do it automatically here, I'd keep it as a manual step in the release checklist. It's just based on some of our previous releases where we shipped something to the users and we quickly found out it was broken, so I tend to test everything I can before the release 😄

infomiho: What about the rest of the comment:
> We could be declarative around which tests we are doing

waspc/starters-headless-tests/run-starter-headless-tests.sh
infomiho: >  We can think if node is better suited for this job.

I'll let you be the judge of that after you implement all the changes, how you feel about maintaining bash vs. Node.js variant of the code. I'm a bit biased towards Node.js because I often had that path of going first with bash then moving to Node.js because it got too complex for my taste (the script gets very long and I don't have any types to model the domain).

waspc/starters-headless-tests/tests/health-check.spec.ts
infomiho: > Maybe that should be a separate issue. So we get going with at least a health check headless test before we go more vertical into specific starters.

I'm also fine with this 👍 

What I said in the Open Saas comment: the rule of thump should be to automate all the stuff we now do manually. And I know I want to click around each starter because sometimes even if the client loads, there might be a CORS error which we might not see if we only do the basic health check. 

I'd be much more confident in the release process knowing all the apps are tested as I would test them manually. Maybe that's an overkill, but again, I'm going on my intuition form all previous releases.

waspc/starters-e2e-tests/playwright.config.ts
```diff
@@ -0,0 +1,55 @@
...
+    // },
+  ],
+
+  /* Run your local dev server before starting the tests */
+  webServer: {
+    command: `run-wasp-app ${process.env.HEADLESS_TEST_MODE} --path-to-app=${process.env.E2E_APP_PATH} --wasp-cli-cmd=${process.env.WASP_CLI_CMD}`,
+    // Wait for the backend to start
+    url: "http://localhost:3001",
+    reuseExistingServer: false,
+    timeout: 240 * 1000,
```
FranjoMindek, line 52: The timeout is bumped from usual the 120 seconds because of `open-saas`.

waspc/starters-e2e-tests/run-starter-e2e-tests.sh
```diff
@@ -0,0 +1,122 @@
...
+
+  ${WASP_CLI_CMD} new "${TEMP_PROJECT_NAME}" -t "${TEMPLATE_NAME}"
+
+  # Starters don't have .env.server.headless (like todoApp)
+  # as we don't want to generate those files for end users.
+
+  if [ -f "$TEMP_WASP_PROJECT_PATH/.env.server.example" ]; then
+    cp "$TEMP_WASP_PROJECT_PATH/.env.server.example" "$TEMP_WASP_PROJECT_PATH/.env.server"
+  else
+    touch "$TEMP_WASP_PROJECT_PATH/.env.server"
```
FranjoMindek: We always create `.env.server` so we can inject env vars.
There is no `.env.server.headless` because we don't ship those with starters.

waspc/starters-e2e-tests/run-starter-e2e-tests.sh
```diff
@@ -0,0 +1,122 @@
...
+  fi
+
+  # Add email verification skip for dev mode
+  cat >> "$TEMP_WASP_PROJECT_PATH/.env.server" << EOF
+
+SKIP_EMAIL_VERIFICATION_IN_DEV=true
+EOF
+
+  # Replace email provider with SMTP so it works with `wasp-app-runner`
+  sed -i '' 's/provider: [A-Za-z0-9_][A-Za-z0-9_]*/provider: SMTP/g' "$TEMP_WASP_PROJECT_PATH/main.wasp"
```
FranjoMindek: This was a bit problematic to get it to work since macOS `sed` is not same as GNU.

waspc/starters-e2e-tests/run-starter-e2e-tests.sh
```diff
@@ -0,0 +1,122 @@
...
+  rm -rf "${TEMP_PROJECT_NAME}"
+}
+
+run_dev_e2e_tests() {
+  echo "Running DEV e2e tests for ${TEMPLATE_NAME} project..."
+  # export DEBUG=pw:webserver
+  export E2E_APP_PATH="$TEMP_WASP_PROJECT_PATH"
+  export WASP_CLI_CMD="${WASP_CLI_CMD}"
+  export HEADLESS_TEST_MODE=dev
+  npx playwright test --grep "(@${TEMPLATE_NAME}|^(?!.*@).*)"
```
FranjoMindek: `--grep` limits the tests we run to either:
1) tests tagged with `@{template_name}`
2) anything without `@` in tag, and since `@something` is the convention, we run all untagged tests

waspc/starters-e2e-tests/tests/auth.ts
FranjoMindek, line 1: Taken from `todoApp`, and improved a bit with types, naming, reordering etc.

waspc/starters-e2e-tests/src/utils.ts
```diff
@@ -0,0 +1,17 @@
+import { existsSync } from "fs";
+import { dirname, join } from "path";
+
+export function findNodeProjectRootDirectory(
```
FranjoMindek: We do this because when we build, we get `dist/src/main.js` which is 1 folder deeper than `src/main.ts`.
We can't hardcode values to get to the project root, so we do it recursively.

waspc/starters-e2e-tests/run-starter-e2e-tests.sh
FranjoMindek: This actually doesn't work on Linux, ill need to fix this.

waspc/starters-e2e-tests/run-starter-e2e-tests.sh
```diff
@@ -0,0 +1,122 @@
...
+
+  echo "Running BUILD e2e tests for ${TEMPLATE_NAME} project..."
+  # export DEBUG=pw:webserver
+  export E2E_APP_PATH="$TEMP_WASP_PROJECT_PATH"
+  export WASP_CLI_CMD="${WASP_CLI_CMD}"
+  export HEADLESS_TEST_MODE=build
+  npx playwright test --grep "(@${TEMPLATE_NAME}|^(?!.*@).*)"
+}
+
+template_uses_sqlite() {
```
FranjoMindek: Now that I merged with `main`, I might reuse `scripts/get-wasp-database-provider.sh` instead.

FranjoMindek:
I'll rename `HEADLESS_TEST_MODE` in accordance to https://github.com/wasp-lang/wasp/pull/3018

waspc/starters-headless-tests/run-starter-headless-tests.sh
FranjoMindek: I've now reused existing script we have for detecting wasp database provider.

waspc/starters-headless-tests/run-starter-headless-tests.sh
FranjoMindek: We handle test organization in ts file now, with simple `package.json` scripts.
I still avoid doing deps management inside of scripts.

waspc/starters-headless-tests/tests/health-check.spec.ts
FranjoMindek: This is now implemented, we have `basic` specific tests.

waspc/starters-headless-tests/run-starter-headless-tests.sh
FranjoMindek: I want to use `open-saas`'s e2e tests, which makes sense to me rather than creating new ones.
However we will only do this after we update `open-saas` e2e tests.

For now we run the basic health check.

FranjoMindek: I've implemented something similar to this in typescript.

waspc/starters-e2e-tests/tsconfig.json
```diff
@@ -0,0 +1,23 @@
...
+    "esModuleInterop": true,
+    "skipLibCheck": true,
+    "target": "es2022",
+    "allowJs": true,
+    "resolveJsonModule": true,
+    "moduleResolution": "nodenext",
+    "moduleDetection": "force",
+    "isolatedModules": true,
+
+    // linting
```
infomiho: I'd say type-checking not linting 😄 we probably don't need this comment anyways

waspc/starters-e2e-tests/src/cli.ts
```diff
@@ -0,0 +1,23 @@
+export type WaspCliCommand = "wasp" | "wasp-cli";
```
infomiho: I think it can just be a `string` because someone might run this in some weird setup and the CLI can be named anything. I don't see the benefit of limiting it to these two values only. Am I missing something?

waspc/starters-e2e-tests/src/main.ts
```diff
@@ -0,0 +1,111 @@
...
+  RemoteTestsRelativePath,
+  STARTER_TESTS,
+  WaspProjectRelativePath,
+  type StarterTemplateName,
+  type StarterTests,
+} from "./starter-tests.js";
+import { findNodeProjectRootDirectory } from "./utils.js";
+
+const SCRIPT_DIRECTORY = dirname(fileURLToPath(import.meta.url));
+const STARTERS_E2E_TEST_RUNNER_SCRIPT_NAME = "run-starter-e2e-tests.sh";
```
infomiho: I'd move these two in the `runStarterE2ETests` function since they are not really used anywhere else.

waspc/starters-e2e-tests/src/cli.ts
```diff
@@ -0,0 +1,23 @@
...
+
+type StartersHeadlessTestsArgs = {
+  waspCliCommand: WaspCliCommand;
+};
+
+export function parseArgs(args: string[]): StartersHeadlessTestsArgs {
```
infomiho: We need to run `npm install` anyways so why not use `commander` for parsing the args? Also, why not use `zx` to execute the commands you need. I get we want to make it "just a script" but are already beyond that point since we need the `typescript` compiler installed.

waspc/starters-e2e-tests/src/starter-tests.ts
```diff
@@ -0,0 +1,38 @@
+export type StarterTemplateName = "minimal" | "basic" | "saas";
+
+export type WaspProjectRelativePath = `/${string}`;
```
infomiho: Shouldn't relative paths start with `./` and absolute paths with `/`? I don't feel like there is a need to enforce this on the type level. Am I missing something?

waspc/starters-e2e-tests/src/starter-tests.ts
```diff
@@ -0,0 +1,38 @@
...
+export type RemoteTestsRelativePath = `/${string}`;
+
+export type StarterTests = LocalStarterTests | RemoteStarterTests;
```
infomiho: `StarterTests` implies some actual tests are contained inside, but this is more like a starter description or just starter.

```ts
type Starter = StarterWithoutTests | StarterWithTests
```

I'm not 100% sure about the naming, but the difference is that one category comes with tests and the other doesn't.

waspc/starters-e2e-tests/src/starter-tests.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const basicStarterTests: LocalStarterTests = {
+  templateName: "basic",
+  waspProjectRelativePath: "/",
+};
+
+const openSaasStarterTests: RemoteStarterTests = {
+  templateName: "saas",
+  waspProjectRelativePath: "/app",
+  remoteTestsRelativePath: "/e2e-tests",
```
infomiho: I'd maybe go for `includedTestsRelativePath`

waspc/starters-e2e-tests/tests/auth.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  await expect(page.locator("body")).toContainText(
+    `You've signed up successfully! Check your email for the confirmation link.`,
+  );
+}
+
+async function submitSignupForm(page: Page, credentials: Credentials) {
+  for (const [key, value] of Object.entries(credentials)) {
+    await page.locator(`input[name='${key}']`).fill(String(value));
+  }
```
infomiho: Maybe you can extract this into `fillForm(objectWithValues)` which you can use here and in the login below. It feels inputs with names and values. That's generic enough that it can be extracted I think.

waspc/starters-e2e-tests/tests/auth.ts
```diff
@@ -0,0 +1,139 @@
...
+interface MailcrabMessage {
+  id: string;
+  to: { email: string }[];
+}
+
+interface MailcrabMessageDetails {
+  text: string;
+}
+
+async function getEmailVerificationLink(
```
infomiho: Maybe extract the Mailcrab specific code into `mailcrab.ts` to keep it easier to follow along.

waspc/starters-e2e-tests/tests/auth.ts
infomiho: `wasp-testing-utils` wen 😄

waspc/starters-e2e-tests/tests/basic.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+ */
+
+import { expect, type Locator, type Page, test } from "@playwright/test";
+import { randomUUID } from "crypto";
+import { performLogin, setupTestUser } from "./auth.js";
+
+test.describe(
+  "basic only",
+  {
+    tag: "@basic",
```
infomiho: TIL about tags, cool!

waspc/starters-e2e-tests/tests/basic.spec.ts
```diff
@@ -0,0 +1,131 @@
+/**
+ * This file describes additional tests for the `basic` starter.
+ */
```
infomiho: The name `basic` is not doing justice to this special case IMHO, someone might mistake this for some "basic" tests that we run. 

Maybe we can improve this by having `starterSpecific` folder so it would read `tests/starterSpecific/basic.spec.ts` or something like `tests/basicStarter.spec.ts` would be enough?

waspc/starters-e2e-tests/tests/basic.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    function getTagDialog(page: Page): Locator {
+      return page.getByRole("dialog");
+    }
+
+    test.describe("tasks", () => {
+      test.describe.configure({ mode: "serial" });
+
+      const credentials = setupTestUser({
+        username: "Test User",
```
infomiho: Username with spaces 😩

waspc/starters-e2e-tests/tests/basic.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+      });
+      // we use random UUIDs to ensure that the tasks are unique
```
infomiho: Kinda redundant comment

waspc/starters-e2e-tests/tests/basic.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+      const taskWithoutTagDescription = `Test task without tags: ${randomUUID()}`;
```
infomiho: Used only inside of the test, maybe move it inside?

waspc/starters-e2e-tests/tests/basic.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+      const tagName = `Test tag: ${randomUUID()}`;
+      const taskWithTagDescription = `Test task with tag: ${randomUUID()}`;
```
infomiho: Also, used only in the test, maybe move it inside. I understand that you have to have `tagName` or maybe `exampleTagName` on the top level since you used it across tests so you wanted to group all these variables. 

IMHO it's easier to follow along if local stuff stays local so I don't wonder where it's used besides in that one place.

waspc/starters-e2e-tests/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "starters-e2e-tests",
+  "version": "1.0.0",
+  "type": "module",
+  "description": "",
+  "author": "",
+  "license": "ISC",
+  "main": "index.js",
+  "keywords": [],
```
infomiho: I'd remove all the stuff we don't need

waspc/starters-e2e-tests/src/utils.ts
infomiho: This prompts me to ask if you maybe want to move the bash script in JS as well? This way you'd avoid having to search for the script like this and have one JS function you call where you need to. You can install `zx` and easily rewrite the script into JS.

infomiho: Here's an example of how the script can look like:

```ts
import { $, fs, path } from 'zx';

interface E2ETestParams {
  waspCliCommand: string;
  templateName: string;
  waspProjectRelativePath: string;
}

export async function runE2ETests(params: E2ETestParams): Promise<void> {
  validateParams(params);

  const { waspCliCommand, templateName, waspProjectRelativePath } = params;
  const tempProjectName = `temp-project-${templateName}`;
  const tempWaspProjectPath = `${tempProjectName}${waspProjectRelativePath}`;

  const cleanup = async (): Promise<void> => {
    try {
      await $`rm -rf ${tempProjectName}`;
    } catch {
      // Ignore cleanup errors
    }
  };

  // Setup cleanup handlers
  process.on('exit', cleanup);
  process.on('SIGINT', cleanup);
  process.on('SIGTERM', cleanup);
  process.on('SIGQUIT', cleanup);

  try {
    await initializeTestEnvironment(waspCliCommand, templateName, tempProjectName, tempWaspProjectPath);
    await runDevE2ETests(templateName, tempWaspProjectPath, waspCliCommand);
    await runBuildE2ETests(templateName, tempWaspProjectPath, waspCliCommand);
  } finally {
    await cleanup();
  }
}

function validateParams(params: E2ETestParams): void {
  if (!params.waspCliCommand) {
    throw new Error("Wasp CLI command is required");
  }
  if (!params.templateName) {
    throw new Error("Template name is required");
  }
  if (!params.waspProjectRelativePath) {
    throw new Error("Wasp project relative path is required");
  }
}

async function initializeTestEnvironment(
  waspCliCommand: string,
  templateName: string,
  tempProjectName: string,
  tempWaspProjectPath: string
): Promise<void> {
  // Clean up any existing project
  await $`rm -rf ${tempProjectName}`;

  // Create new project from template
  await $`${waspCliCommand} new ${tempProjectName} -t ${templateName}`;

  await setupEnvironmentFiles(tempWaspProjectPath);
  await configureEmailProvider(tempWaspProjectPath);
}

async function setupEnvironmentFiles(tempWaspProjectPath: string): Promise<void> {
  await setupServerEnvironment(tempWaspProjectPath);
  await setupClientEnvironment(tempWaspProjectPath);
}

async function setupServerEnvironment(tempWaspProjectPath: string): Promise<void> {
  const serverExamplePath = `${tempWaspProjectPath}/.env.server.example`;
  const serverEnvPath = `${tempWaspProjectPath}/.env.server`;
  
  if (await fs.pathExists(serverExamplePath)) {
    await fs.copy(serverExamplePath, serverEnvPath);
  } else {
    await fs.ensureFile(serverEnvPath);
  }

  const serverConfig = `
SKIP_EMAIL_VERIFICATION_IN_DEV=true
SMTP_HOST=localhost
SMTP_USERNAME=any
SMTP_PASSWORD=any
SMTP_PORT=1025
`;
  await fs.appendFile(serverEnvPath, serverConfig);
}

async function setupClientEnvironment(tempWaspProjectPath: string): Promise<void> {
  const clientExamplePath = `${tempWaspProjectPath}/.env.client.example`;
  const clientEnvPath = `${tempWaspProjectPath}/.env.client`;
  
  if (await fs.pathExists(clientExamplePath)) {
    await fs.copy(clientExamplePath, clientEnvPath);
  }
}

async function configureEmailProvider(tempWaspProjectPath: string): Promise<void> {
  const mainWaspPath = `${tempWaspProjectPath}/main.wasp`;
  
  try {
    const content = await fs.readFile(mainWaspPath, 'utf8');
    const updatedContent = content.replace(
      /provider:\s+[A-Za-z0-9_][A-Za-z0-9_]*/g,
      'provider: SMTP'
    );
    await fs.writeFile(mainWaspPath, updatedContent);
  } catch (error) {
    console.warn(`Could not update email provider in ${mainWaspPath}:`, error);
  }
}

async function runDevE2ETests(
  templateName: string,
  tempWaspProjectPath: string,
  waspCliCommand: string
): Promise<void> {
  console.log(`Running DEV e2e tests for ${templateName} project...`);
  
  const originalEnv = { ...process.env };
  try {
    process.env.WASP_APP_PATH = tempWaspProjectPath;
    process.env.WASP_CLI_CMD = waspCliCommand;
    process.env.WASP_RUN_MODE = 'dev';
    
    await $`npx playwright test --grep "(@${templateName}|^(?!.*@).*)"`;
  } finally {
    Object.assign(process.env, originalEnv);
  }
}

async function runBuildE2ETests(
  templateName: string,
  tempWaspProjectPath: string,
  waspCliCommand: string
): Promise<void> {
  if (await templateUsesSqlite(tempWaspProjectPath, waspCliCommand)) {
    console.log(`Skipping BUILD tests for ${templateName} project - uses SQLite`);
    return;
  }

  console.log(`Running BUILD e2e tests for ${templateName} project...`);
  
  const originalEnv = { ...process.env };
  try {
    process.env.WASP_APP_PATH = tempWaspProjectPath;
    process.env.WASP_CLI_CMD = waspCliCommand;
    process.env.WASP_RUN_MODE = 'build';
    
    await $`npx playwright test --grep "(@${templateName}|^(?!.*@).*)"`;
  } finally {
    Object.assign(process.env, originalEnv);
  }
}

async function templateUsesSqlite(tempWaspProjectPath: string, waspCliCommand: string): Promise<boolean> {
  const scriptDirectory = path.dirname(new URL(import.meta.url).pathname);
  const databaseProviderScriptPath = path.join(
    scriptDirectory, 
    '../../scripts/get-wasp-database-provider.sh'
  );

  const originalCwd = process.cwd();
  try {
    process.chdir(tempWaspProjectPath);
    const result = await $`${databaseProviderScriptPath} ${waspCliCommand}`;
    return result.stdout.trim() === 'sqlite';
  } catch (error) {
    console.warn('Could not determine database provider, assuming not SQLite');
    return false;
  } finally {
    process.chdir(originalCwd);
  }
}
```

infomiho (commented):
I think it's going in the right direction 👍 you are handling a lot of edge cases which is the reality with our different starters, good job on that!

### Biggest suggestion

I'd maybe move to using `zx` we already have a proper JS project now. It would save you from searching the JS project root and passing the data between JS and the script. I've GPTed some version that might help you imagine how it could look like.

I won't block on this, but let's consider it 😄 

### Locally testing `saas`

I've tried running locally and didn't manage to run `saas` tests:
```
Error: Process from config.webServer was not able to start. Exit code: 1

Failed to run tests for saas: Error: Starter tests failed with exit code 1
    at ChildProcess.<anonymous> (file:///Users/ilakovac/dev/wasp/waspc/starters-e2e-tests/dist/src/main.js:48:24)
    at ChildProcess.emit (node:events:518:28)
    at maybeClose (node:internal/child_process:1101:16)
    at ChildProcess._handle.onexit (node:internal/child_process:304:5)
```

I've added `DEBUG=pw:webserver` to the command and I saw that I didn't have Docker running. Maybe we want to add `DEBUG=pw:webserver` by default?

waspc/starters-e2e-tests/tsconfig.json
FranjoMindek: Which is done by linter :P.
`tsconfig.json` was copied from `wasp-config`, I'll remove the comment though, both here and in `wasp-config`.

waspc/starters-e2e-tests/tests/auth.ts
FranjoMindek: I also need to provide a way to locate form inputs e.g. `Page`
So it would be `fillForm(page, objectWithValues)`

I'm not 100% what would be the best approach exactly because of that locator.
`Page` can have multiple forms, so that seems weird.

We know it has only 1 form in this specific case.

I would delay it for now.

I've changed key value pairs to be better named:
```ts
for (const [formFieldName, formFieldValue] of Object.entries(credentials)) {
  await page
    .locator(`input[name='${formFieldName}']`)
    .fill(String(formFieldValue));
}
```

waspc/starters-e2e-tests/tests/basic.spec.ts
FranjoMindek: Gotta test everything, right :D

waspc/starters-e2e-tests/src/starter-tests.ts
FranjoMindek: This is because they were appended to the end of some absolute path.
I will need to re-think this whole path approach a bit.

FranjoMindek: Since I can easily use `path.join` now, I made them actually relative.

FranjoMindek: Starts with `.`, because `../` can happen too

waspc/starters-e2e-tests/src/utils.ts
FranjoMindek: I started from this but really changed it up in the end.

FranjoMindek:
I'm never sure if I want to add `DEBUG=pw:webserver` or not by default.
No info vs too much logs.

We should pick one and do so for all of the headless e2e tests.


I've changed the naming a bit to reflect our new idea of full test names -> `headless e2e`.
Project is now fully typescript.

waspc/starters-headless-e2e-tests/src/mailcrab.ts
```diff
@@ -0,0 +1,42 @@
+import { fs } from "zx";
+
+/**
+ * Configures the Wasp for MailCrab SMTP email server which is used by `wasp-app-runner`.
+ *
+ * Assumes the Wasp project uses the `main.wasp` as the Wasp configuration file.
```
FranjoMindek: Also assumes that `.env.server` exists? Should I add that? Didn't seem that important.

waspc/starters-headless-e2e-tests/package.json
```diff
@@ -0,0 +1,22 @@
...
+  },
+  "dependencies": {
+    "@commander-js/extra-typings": "^14.0.0",
+    "commander": "^14.0.0",
+    "zx": "^8.8.0"
+  },
+  "devDependencies": {
+    "@playwright/test": "1.51.1",
+    "@types/node": "^24.0.1",
+    "@wasp.sh/wasp-app-runner": "^0.0.7"
```
infomiho: Are Playwright and Wasp App Runner really dev deps?

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
```diff
@@ -0,0 +1,163 @@
...
+  starterHeadlessE2ETests: StarterHeadlessE2ETests;
+}
+
+export async function runStarterHeadlessE2ETests(
+  execution: StarterTestsExecution,
+): Promise<void> {
+  const { waspCliCommand, starterHeadlessE2ETests } = execution;
+  const { starterName } = starterHeadlessE2ETests;
+
+  const tempWaspProjectPath = await spinner(
```
infomiho: TIL about `spinner` pretty sweet

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
```diff
@@ -0,0 +1,163 @@
...
+  process.on("SIGINT", () => cleanup(tempDirectoryPath));
+  process.on("SIGTERM", () => cleanup(tempDirectoryPath));
+  process.on("SIGQUIT", () => cleanup(tempDirectoryPath));
+
+  const waspStarterProjectName = `wasp-starter-${starterName}`;
+
+  await within(async () => {
+    $.cwd = tempDirectoryPath;
+    await $`${waspCliCommand} new ${waspStarterProjectName} -t ${starterName}`;
+  });
```
infomiho:
```suggestion
  await $({ cwd: tempDirectoryPath })`${waspCliCommand} new ${waspStarterProjectName} -t ${starterName}`;
```

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  const tempStarterProjectPath = path.join(
```
infomiho:
```suggestion
  const starterProjectPath = path.join(
```

I don't think we care about this being a temp thing or not. 

The naming here confused me a bit, I had to take a look around to figure out that you have:
- tmp dir
- folder where the starter was created
- folder where the wasp project lives inside of starter folder

So I'd maybe try to encode that better with some naming e.g.

- `tempDirectoryPath` - stays the same
- `starterDirectoryPath`
- `waspProjectPath`


Also... since you don't need `tempStarterProjectPath` you could do something like:
```ts
const waspProjectPath = path.join(tempDirectoryPath, starterName, waspProjectRelativePath)
```

waspc/starters-headless-e2e-tests/src/mailcrab.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  const waspAppSpecPath = `${waspProjectPath}/main.wasp`;
+  setupMailCrabWaspAppSpec(waspAppSpecPath);
+}
+
+async function setupMailCrabEnvVariables(
+  waspServerEnvFilePath: string,
+): Promise<void> {
+  const mailCrabSMTPEnvVars = `
+SKIP_EMAIL_VERIFICATION_IN_DEV=true
```
infomiho: Unrelated env var to Mailcrab

waspc/starters-headless-e2e-tests/src/mailcrab.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  await fs.appendFile(waspServerEnvFilePath, mailCrabSMTPEnvVars);
+}
+
+async function setupMailCrabWaspAppSpec(
+  waspAppSpecPath: string,
+): Promise<void> {
+  const waspAppSpec = await fs.readFile(waspAppSpecPath, "utf8");
+  const waspSMTPAppSpec = waspAppSpec.replace(
+    /provider:\s+[A-Za-z0-9_][A-Za-z0-9_]*/g,
```
infomiho: This looks fragile on the account that `provider` is a generic name. But it'll work for now.

waspc/starters-headless-e2e-tests/src/mailcrab.ts
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Configures the Wasp for MailCrab SMTP email server which is used by `wasp-app-runner`.
+ *
+ * Assumes the Wasp project uses the `main.wasp` as the Wasp configuration file.
+ */
+export async function setupWaspMailCrabConfiguration(
+  waspProjectPath: string,
+): Promise<void> {
+  const waspServerEnvFilePath = `${waspProjectPath}/.env.server`;
```
infomiho: We can use `path.join` here as well.

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+async function cleanup(tempDirectoryPath: string): Promise<void> {
+  await $({ nothrow: true })`rm -rf ${tempDirectoryPath}`;
+}
+
+async function initializeServerEnvironment(
+  tempWaspProjectPath: string,
+): Promise<void> {
+  const serverEnvFileExamplePath = `${tempWaspProjectPath}/.env.server.example`;
```
infomiho: `path.join`

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
```diff
@@ -0,0 +1,163 @@
...
+  const serverEnvFilePath = `${tempWaspProjectPath}/.env.server`;
+
+  if (await fs.pathExists(serverEnvFileExamplePath)) {
+    await fs.copy(serverEnvFileExamplePath, serverEnvFilePath);
+  } else {
+    await fs.ensureFile(serverEnvFilePath);
```
infomiho: What if a starter doesn't require env vars?

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
```diff
@@ -0,0 +1,163 @@
...
+async function initializeClientEnvironment(
+  tempWaspProjectPath: string,
+): Promise<void> {
+  const clientEnvFileExamplePath = `${tempWaspProjectPath}/.env.client.example`;
+  const clientEnvFilePath = `${tempWaspProjectPath}/.env.client`;
+
+  if (await fs.pathExists(clientEnvFileExamplePath)) {
+    await fs.copy(clientEnvFileExamplePath, clientEnvFilePath);
+  } else {
+    await fs.ensureFile(clientEnvFilePath);
```
infomiho: Same question as for `.env.server`

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
```diff
@@ -0,0 +1,163 @@
...
+    await fs.copy(serverEnvFileExamplePath, serverEnvFilePath);
+  } else {
+    await fs.ensureFile(serverEnvFilePath);
+  }
+}
+
+async function initializeClientEnvironment(
+  tempWaspProjectPath: string,
+): Promise<void> {
+  const clientEnvFileExamplePath = `${tempWaspProjectPath}/.env.client.example`;
```
infomiho: `path.join`

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
```diff
@@ -0,0 +1,163 @@
...
+    scriptDirectory,
+    "../../../../scripts/get-wasp-database-provider.sh",
+  );
+
+  return await within(async () => {
+    $.cwd = tempWaspProjectPath;
+    const result = await $`${databaseProviderScriptPath} ${waspCliCommand}`;
+    return result.stdout.trim() === "sqlite";
+  });
+}
```
infomiho: Not for this PR, but this kinda prompts me to think that we could benefit from a `run-wasp-app get-database-provider` kinda of command in the Wasp App Runner.

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
```diff
@@ -0,0 +1,163 @@
...
+): Promise<void> {
+  const { waspCliCommand, starterHeadlessE2ETests } = execution;
+  const { starterName } = starterHeadlessE2ETests;
+
+  const tempWaspProjectPath = await spinner(
+    "Initializing test environment...",
+    () => initializeTestEnvironment(execution),
+  );
+
+  // TODO: implement branching logic based on included tests or not
```
infomiho: Will we handle this TODO in this PR or in the future?

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-tests.ts
```diff
@@ -0,0 +1,39 @@
...
+  starterName: "saas",
+  waspProjectRelativePath: "./app",
+  includedTestsRelativePath: "./e2e-tests",
+};
+
+export const STARTERS_HEADLESS_E2E_TESTS: StarterHeadlessE2ETests[] = [
+  minimalStarterHeadlessE2ETests,
+  basicStarterHeadlessE2ETests,
+  openSaasStarterHeadlessE2ETests,
+];
```
infomiho: What if we inlined the tests, it'll be less verbose and easier to add new tests:

```ts
export const STARTERS_HEADLESS_E2E_TESTS: StarterHeadlessE2ETests[] = [
  {
    starterName: 'minimal',
    waspProjectRelativePath: './',
  },
  {
    starterName: 'basic',
    waspProjectRelativePath: './',
  },
  {
    starterName: 'saas',
    waspProjectRelativePath: './app',
    includedTestsRelativePath: './e2e-tests',
  },
];

```

waspc/starters-headless-e2e-tests/tests/auth.ts
```diff
@@ -0,0 +1,99 @@
...
+    `You've signed up successfully! Check your email for the confirmation link.`,
+  );
+}
+
+async function submitSignupForm(page: Page, credentials: Credentials) {
+  for (const [formFieldName, formFieldValue] of Object.entries(credentials)) {
+    await page
+      .locator(`input[name='${formFieldName}']`)
+      .fill(String(formFieldValue));
+  }
```
infomiho: What about the old comment about extracting this into `fillForm` helper that you can use for signup and login.

waspc/starters-headless-e2e-tests/src/mailcrab.ts
infomiho: I'd add that assumption as well 👍 You could even validate it in the function not to depend on outside code doing it.

waspc/starters-headless-e2e-tests/src/cli.ts
```diff
@@ -0,0 +1,28 @@
...
+    .name("starters-headless-e2e-tests")
+    .description(
+      "Run headless end-to-end tests for Wasp starter templates (except ai)",
+    )
+    .version(packageJson.version)
+    .addArgument(
+      new Argument(
+        "wasp-cli-command",
+        "Path or command name for the Wasp CLI executable to test against",
+      ).argRequired(),
```
infomiho: Nitpick: this feels more like a required option to me:
```bash
cli wasp-cli
# vs
cli --wasp-cli-command wasp-cli
```

If you did it like an option you could just return `command.opts()`.

waspc/starters-headless-e2e-tests/package.json
FranjoMindek: Usually yes, since they are not a part of the actual application, but rather used only for testing purposes.

Well the purpose of this project is to test stuff, so it kinda muddles it out. 

I would personally leave them there unless you insist.

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
FranjoMindek: Ooo forgot about that option, nice!

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
FranjoMindek: Multiple paths inside of `path.join`, forgot about that.
Yeah `temp` is a detail, we can go with `waspProjectPath`.
Rest of code doesn't care it's temp because we set up the cleanup here.

waspc/starters-headless-e2e-tests/src/mailcrab.ts
FranjoMindek: Will be moved to `initializeServerEnvironment`.

waspc/starters-headless-e2e-tests/src/mailcrab.ts
FranjoMindek: I agree, I had the same thought, we actually mention `provider` multiple times in `saas`'s `main.wasp`.
But `provider:` is only once.

waspc/starters-headless-e2e-tests/src/mailcrab.ts
FranjoMindek: Good for cross-platform support 👍🏻

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
FranjoMindek: We will always at least want the `SKIP_EMAIL_VERIFICATION_IN_DEV` in server env vars.

But it's good to ensure those files exist, even if empty. 
Mostly so we don't have to think about it in further code which would touch env vars.

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
FranjoMindek: I've been thinking about how it's messy too.
This will be much easier if we would follow https://github.com/wasp-lang/wasp/issues/3031

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
FranjoMindek: Future, after we fix `saas` e2e tests to use `wasp-app-runner`

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-tests.ts
FranjoMindek:
```ts
export const STARTERS_HEADLESS_E2E_TESTS: StarterHeadlessE2ETests[] = [
  {
    starterName: "minimal",
    waspProjectRelativePath: "./",
  } satisfies StarterWithoutHeadlessE2ETests,
  {
    starterName: "basic",
    waspProjectRelativePath: "./",
  } satisfies StarterWithoutHeadlessE2ETests,
  {
    starterName: "saas",
    waspProjectRelativePath: "./app",
    includedTestsRelativePath: "./e2e-tests",
  } satisfies StarterWithHeadlessE2ETests,
];
```

waspc/starters-headless-e2e-tests/tests/auth.ts
FranjoMindek: I've made a reply on that comment https://github.com/wasp-lang/wasp/pull/2868#discussion_r2256837007

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
FranjoMindek: We will always at least want the `SKIP_EMAIL_VERIFICATION_IN_DEV` in server env vars.

But it's good to ensure those files exist, even if empty.
Mostly so we don't have to think about it in further code which would touch env vars.

waspc/starters-headless-e2e-tests/src/mailcrab.ts
FranjoMindek: Added ensure that file exists.

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
FranjoMindek: Ir we don't want to assume that, then we can remove ensure from the client side.

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
infomiho: Ok, so there is an expectation that the user of this CLI tool will copy the env files in the right place. That makes sense and we should be validating it like this 👍

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
infomiho: For the client side, we don't really need the file if I'm not mistaken. I'm okay with us requiring it here. It's a condition that we set and we validate it. 

The question that bugs me know - did we do it when e2e headless test the example apps as well? Do we want to adjust the logic there as well? 

RAW

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
infomiho: If we think it's worth it, let's make an issue to tackle it later on.

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
infomiho: Ok, let's create an issue for this then if you haven't already and link it in the comment.

waspc/starters-headless-e2e-tests/package.json
infomiho: I think you said it yourself:
> Well the purpose of this project is to test stuff

I'd move it, but I won't block on this.

RAW

infomiho:
@FranjoMindek food for thought - do we maybe want to move this out of `waspc` and into the top level? We have the `wasp-app-runner` there as well (I thought it's not a part of the Wasp compiler since it was a tool for testing). What do you think?

infomiho:
Another thing, I wanted to test it out locally and realized a `README.md` is missing with some basic instructions. Let's add it. You can ask e.g. Copilot to generate the draft based on project contents, it's usually a good starting point.

infomiho:
Also, I'd def add the DEBUG env var by default, I think we should see the output if something goes wrong - we do it for example apps testing as well I believe.

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
FranjoMindek: I'm not sure what do you mean here.

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
FranjoMindek: I've removed the ensure on client side, since we don't necessarily need that file.
We don't do it for example apps (most likely since we didn't have `.env.client.example`), but it would be a good practice to add it there too.

There is also the difference that we use `.env.server.headless` there.

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
FranjoMindek: But it kinda feels weird to me, especially if we want `wasp-app-runner` to only run the apps.
I think that solving the core issue above is the way to go.

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
FranjoMindek: https://github.com/wasp-lang/open-saas/issues/469

Created it, seems like Carlos self-assigned to it.

FranjoMindek:
> @FranjoMindek food for thought - do we maybe want to move this out of `waspc` and into the top level? We have the `wasp-app-runner` there as well (I thought it's not a part of the Wasp compiler since it was a tool for testing). What do you think?

It's testing that the Wasp compiler (through Wasp CLI) is generating starter projects correctly, so `waspc` may be okay.

waspc/starters-headless-e2e-tests/tests/starter/basic.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+    function getTaskForm(page: Page): Locator {
+      return page.locator("#create-task");
+    }
+
+    function getTasksSection(page: Page): Locator {
+      return page
+        .locator(
+          "section.card.flex.w-full.max-w-3xl.flex-col.gap-6.p-4.lg\\:p-6",
+        )
+        .nth(1);
```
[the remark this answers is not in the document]
FranjoMindek: We can't use test id in starter templates

FranjoMindek:
I've added the debug flag to `package.json` so it doesn't interfere with testing logic itself.
If you want to run without it you can easily do it via `npm run start` 👍🏻 .

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
infomiho: I was just say "this makes sense" sorry for trying to be fancy 😂

waspc/starters-headless-e2e-tests/src/starter-headless-e2e-test-runner.ts
infomiho: The above issue talks about the CLI output and I talked about having a tool for getting the database type e.g. in `wasp-app-runner`. I'm not sure if I follow how sorting the `NO_COLOR` will help us with the thing I asked about. Maybe I'm not getting something from the conversation, sorry 😄

FranjoMindek: It would just make it easier to parse from `wasp info`.
One of the things we should do is also provide JSON flag for `wasp info` which would output it in JSON format.

https://clig.dev/#output
> Display output as formatted JSON if `--json` is passed. JSON allows for more structure than plain text, so it makes it much easier to output and handle complex data structures. [jq](https://stedolan.github.io/jq/) is a common tool for working with JSON on the command-line, and there is now a [whole ecosystem of tools](https://ilya-sher.org/2018/04/10/list-of-json-tools-for-command-line/) that output and manipulate JSON. 

With those two combined we have no need for helper script.

infomiho (approved):

Issue #2867, opened 2025-07-01
Upgrade to latest Vite


cprecioso:
We need to upgrade to the latest Vite version (Vite 7 at time of writing)

Placeholder issue to centralize discussions around it and to have somewhere to tag.

cprecioso:
The upgrade to Vite v7 is straightforward, only really needing to warn our users about Node support https://v7.vite.dev/guide/migration.html#node-js-support
We'll need to upgrade our minimum Node version to 20.19

cprecioso:
We've been using latest Vite inadvertently up until https://github.com/wasp-lang/wasp/pull/2865, to basically no problems except the minimum Node version, so this is quite a safe upgrade.

Issue #2504, opened 2025-02-12
Update to Prisma 6


sodic:
@infomiho on Discord:
> Updating to Prisma 6 doesn't seem that problematic, we should do it sooner than later: https://www.prisma.io/docs/orm/more/upgrade-guides/upgrading-versions/upgrading-to-prisma-6

cprecioso:
@vincanger discovered that Prisma <6.1 is not compatible with newer Alpine images

Discord support thread: https://discord.com/channels/686873244791210014/1370209746266755154/1387735087642185809
Prisma issue: https://github.com/prisma/prisma/issues/25817#issuecomment-2538544254

cprecioso:
After doing this task, we could evaluate the following ones:
- #2676 
- #2700 
- #2967

PR #2902, opened 2025-07-03
Add custom Prisma serialization only when the schema is not empty

### LINKED ISSUE #2798: Frontend doesn't render when there's an empty Prisma schema
**Describe the bug**

> Write-up: https://www.notion.so/wasp-lang/Error-writeup-Browser-error-on-blank-project-20018a74854c80249e83ec893c7d6d91?pvs=4

`wasp/core/serialization` is calling `@prisma/client` as part of its logic. When the Prisma schema doesn't have any models, `@prisma/client` is an empty placeholder in the CJS format instead of ESM, which trips up Vite.

OTOH, it seems like this workaround is not needed anymore:
https://github.com/wasp-lang/wasp/blob/014f661a27f829bddf2290f7cdf1cd7c38f3387c/waspc/data/Generator/templates/react-app/vite.config.ts#L44-L54

**To Reproduce**

1. `wasp new`
2. `wasp start`
3. Browser is blank

**Expected behavior**

The template page shows up

**Screenshots**

Image

**Possible solutions**
- Do not import `@prisma/client` from `wasp/core/serialization` if there aren't any Prisma models
- Wrap that import in a `try..catch`
- Create a Vite plugin that correctly resolves whether Prisma is being a CJS or ESM file
- Update Prisma
- Something else (?)

cprecioso:
Resolves #2798

`@prisma/client`, as installed from `npm` is just a placeholder empty client. When calling `prisma generate`, that placeholder gets overwritten with the actual client.

The problem is the placeholder is CJS and the actual built client is ESM. Together with [our workaround](https://github.com/wasp-lang/wasp/blob/a771c50e5cadc589d7fbd54b3fed0c0ec38545b4/waspc/data/Generator/templates/react-app/vite.config.ts#L45-L53) for correct Vite resolution of Prisma in our old version; Vite gets confused.

I made the Prisma importing conditional on there being actual entities in our AppSpec, so we will never import an empty client.

We can remove the workarounds when we:
- #2504 
- #2867


cprecioso:
- Format is failing because of #2903
- Build is failing because of #2900

waspc/data/Generator/templates/sdk/wasp/core/serialization.ts
```diff
@@ -47,6 +54,13 @@ export interface SuperJSONObject {
   [key: string]: SuperJSONValue
 }
 
+{=# entitiesExist =}
+/*
+  This section is only included if there are (Prisma) entities in the app spec,
+  as it needs a non-empty Prisma client to work.
+  TODO: Check if we can remove this  conditional when we upgrade Prisma (#2504)
+*/
```
infomiho: You can replace this comment with Mustache comment:
```
  {=! This section is only included if there are (Prisma) entities in the app spec, =}
```

This way this comment stays "internal" and doesn't get in the user's code.

waspc/src/Wasp/Generator/SdkGenerator.hs
```diff
@@ -243,6 +243,16 @@ depsRequiredForTesting =
...
 
+genCoreSerializationFile :: AppSpec -> Generator FileDraft
+genCoreSerializationFile spec = return $ C.mkTmplFdWithData relCoreSerializationPath tmplData
+  where
+    relCoreSerializationPath = [relfile|core/serialization.ts|]
+    tmplData =
+      object
+        [ "entitiesExist" .= entitiesExist
+        ]
+    entitiesExist = not . null $ getEntities spec
```
infomiho: I'm tempted to ask you to extract this into a helper function that is named `arePrismaModelsDefined` or `isPrismaUsed`. We have the same check repeated 3, 4 times now. 

Places where we do the same check:
- `src/Wasp/Generator/DbGenerator.hs`
- `src/Wasp/Generator/DockerGenerator.hs`
- `src/Wasp/Generator/ServerGenerator.hs`

cprecioso:
@infomiho ready

waspc/data/Generator/templates/sdk/wasp/core/serialization.ts
```diff
@@ -1,4 +1,9 @@
+{{={= =}=}}
```
infomiho: I just realized that it would be nice if we extracted all the Prisma specific stuff into `prismaSerialization.ts` and then imported it here and plugged in here. This way all the Prisma stuff is in one file and it's easier to reason which code bits are Prisma specific.

It's out of scope for this PR, but maybe we can encode the idea if it makes sense into an issue?

waspc/data/Generator/templates/sdk/wasp/core/serialization.ts
```diff
@@ -1,4 +1,9 @@
...
+
+{=# entitiesExist =}
 import { Prisma } from '@prisma/client'
+{=/ entitiesExist =}
+
 import { deserialize, registerCustom, serialize } from 'superjson'
```
infomiho: Nitpick: `registerCustom` is unused if no Prisma entities are defined

waspc/data/Generator/templates/sdk/wasp/core/serialization.ts
cprecioso: I can do it, I had the same idea and was already thinking about this

cprecioso:
@infomiho ready

I split `serialization` into three files `base`, `prisma`, and `index`.

I moved all Prisma-specific things into the `prisma` file (only written if there are entities), and kept the rest in the `index`. `index` only has a conditional import of `./prisma`.

Given that registering custom types in superjson is a side-effect, I wanted to do it also for types. I copied a pattern from [TanStack Router](https://github.com/TanStack/router/blob/v1.124.1/docs/router/framework/react/guide/creating-a-router.md#router-type-safety) where files can register themselves into an interface. This way we keep the types working well either way.

**With Prisma file**
image

**Without Prisma file**
image

This pattern might be useful to us in the Templates > Libraries project

waspc/src/Wasp/Generator/SdkGenerator.hs
```diff
@@ -243,6 +244,24 @@ depsRequiredForTesting =
...
 
+genCoreSerializationDir :: AppSpec -> Generator [FileDraft]
+genCoreSerializationDir spec =
+  return $
+    [ C.mkTmplFd [relfile|core/serialization/base.ts|],
+      C.mkTmplFdWithData [relfile|core/serialization/index.ts|] tmplData
+    ]
+      ++ maybeToList prismaSerializationFile
+  where
+    entitiesExist = hasEntities spec
```
infomiho: Nitpick: I'd move it to last place since it's used by `tmplData` and `prismaSerializationFile`

waspc/data/Generator/templates/sdk/wasp/core/serialization/base.ts
```diff
@@ -0,0 +1,19 @@
+import "superjson"
```
infomiho: This looks pretty hardcore to me and the `base.ts` name doesn't do the concept justice. I'd like to signal somehow that is the "custom serialization init" or "custom serialization base" concept in the filename itself.

waspc/data/Generator/templates/sdk/wasp/core/serialization/index.ts
```diff
@@ -0,0 +1,60 @@
...
+// Removed `Symbol` since we don't explicitly register any symbol to be serialized
+// Added custom types from `WaspInternal_CustomSerialization_Register`
+export type SerializableJSONValue =
+  | Set<SuperJSONValue>
+  | Map<SuperJSONValue, SuperJSONValue>
+  | undefined
+  | bigint
+  | Date
+  | RegExp
+  | WaspInternal_CustomSerialization_Register[keyof WaspInternal_CustomSerialization_Register];
```
infomiho: Will it still work if we moved this to `base.ts` and imported it here again:
```ts
export type CustomSerializableJSONValues = WaspInternal_CustomSerialization_Register[keyof WaspInternal_CustomSerialization_Register];
```

waspc/data/Generator/templates/sdk/wasp/core/serialization/index.ts
```diff
@@ -0,0 +1,60 @@
+{{={= =}=}}
+import {
+  WaspInternal_CustomSerialization_Register,
+  deserialize,
+  serialize,
+} from "superjson"
+import "./base"
```
infomiho: Yeah, `base` really doesn't give me anything as a name here, we need to be more precise.

waspc/data/Generator/templates/sdk/wasp/core/serialization/base.ts
```diff
@@ -0,0 +1,19 @@
...
+  This declaration creates an empty interface. Implementations of custom
+  serialization types can extend this interface through augmentation
+  (https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces).
+  We can later inspect this interface to see which custom types are registered.
+
+  This approach is copied from TanStack Router's `Register` interface
+  https://github.com/TanStack/router/blob/v1.124.1/docs/router/framework/react/guide/creating-a-router.md#router-type-safety
+*/
+declare module "superjson" {
+  interface WaspInternal_CustomSerialization_Register {}
```
infomiho:
```suggestion
  interface WaspInternal_CustomSerializableJSONValue_Register {}
```

To maybe make it more in line with existing naming?

cprecioso:
Fixed comments, renamed `base` to `custom-register` @infomiho

infomiho:
I'm okay with merging 👍

PR #2904, opened 2025-07-03
Migrate `llms.txt` script to Typescript

infomiho:
The old script depended on `sidebars.js` and we needed migrate it to Typescript.

I've used `tsx` to run the script without much hassle and added some types and asserts.


web/scripts/generate-llm-files.ts
```diff
@@ -192,133 +214,146 @@ async function populateDocInfoMap(orderedDocIds, docIdToPathMap) {
...
+    const absoluteFilePath = path.join(BLOG_DIR, file);
     try {
-      const rawContent = await fs.readFile(absoluteFilePath, 'utf8')
-      const { attributes } = fm(rawContent)
+      const rawContent = await fs.readFile(absoluteFilePath, "utf8");
+      const { attributes } = fm(rawContent);
 
-      const title = extractBlogPostTitle(attributes, file)
-      const linkPath = constructBlogUrl(file)
+      const title = extractBlogPostTitle(attributes as { title: string }, file);
```
infomiho: TODO: add an assertion for this bit

web/scripts/generate-llm-files.ts
```diff
@@ -192,133 +214,146 @@ async function populateDocInfoMap(orderedDocIds, docIdToPathMap) {
...
-  for (const item of items) {
-    if (typeof item === 'string') {
-      paths.push(item)
-    } else if (item.type === 'category' && item.items) {
-      paths = paths.concat(flattenSidebarItemsToDocIds(item.items))
-    } else if (item.type === 'autogenerated') {
+function flattenSidebarItemsToDocIds(sidebarConfig: SidebarConfig): string[] {
+  let paths = [];
+  if (!sidebarConfig) return paths;
+  // We need to iterate over the sidebarConfig, but it can be an array or an object.
```
infomiho: TODO: write a better comment

web/scripts/generate-llm-files.ts
```diff
@@ -192,133 +214,146 @@ async function populateDocInfoMap(orderedDocIds, docIdToPathMap) {
...
     }
   }
-  return structuredOverview
+  return structuredOverview;
 }
 
 /**
  * Recursively traverses the sidebar configuration to produce a flat, ordered list of document IDs.
- * @param {object[]} items - An array of sidebar items (strings or objects).
+ * @param {SidebarItemConfig[]} items - An array of sidebar items (strings or objects).
```
infomiho: TODO: fix type mention here, do we just drop the JS doc types now?

web/scripts/generate-llm-files.ts
```diff
@@ -192,133 +214,146 @@ async function populateDocInfoMap(orderedDocIds, docIdToPathMap) {
...
+ * @returns {StructuredOverview[]} A structured array of objects, each containing a categoryLabel and its array of docIds.
  */
-function getDocsSidebarCategoryStructure(sidebarTopLevelItems) {
-  const structuredOverview = []
+function getDocsSidebarCategoryStructure(
+  sidebarConfigDocs: SidebarItemConfig[],
+): StructuredOverview[] {
+  const structuredOverview: StructuredOverview[] = [];
 
-  for (const topItem of sidebarTopLevelItems) {
```
infomiho: TODO: I've worsened the name now

web/scripts/generate-llm-files.ts
```diff
@@ -1,92 +1,104 @@
...
+`;
 const OVERVIEW_MISC_SECTION_CONTENT = `
 ## Miscellaneous
 - [Wasp Developer Discord](https://discord.com/invite/rzdnErX)
 - [Open SaaS -- Wasp's free, open-source SaaS boilerplate starter](https://opensaas.sh)
-`
+`;
 
-await generateFiles()
+(async () => await generateFiles())();
```
cprecioso: Huh, this wasn't working? Maybe we can use the `.mts` extension?

At any rate, This is basically the same as

```suggestion
generateFiles();
```

web/scripts/generate-llm-files.ts
```diff
@@ -1,92 +1,104 @@
...
-const BLOG_DIR = path.join(SITE_ROOT, 'blog/')
+import {
+  SidebarConfig,
+  SidebarItemConfig,
+} from "@docusaurus/plugin-content-docs/src/sidebars/types.js";
+import fm from "front-matter";
+import fs from "fs/promises";
+import { globSync } from "glob";
+import path from "path";
+import docsSidebarConfig from "../sidebars.js";
```
cprecioso, line 10: Should this now be `.ts`? (or no extension)

web/scripts/generate-llm-files.ts
infomiho: TODO: TSX wasn't too happy with top level await, maybe I can pass some flag to it to enable support

web/scripts/generate-llm-files.ts
```diff
@@ -1,92 +1,104 @@
...
+  console.log("🎉 LLM file generation complete.");
 }
 
 /**
  * Processes all documentation files based on the sidebar configuration's order.
  * It builds a map of document IDs to file paths, processes each document,
  * and generates both a summarized overview and a full concatenated string of the content.
- * @param {object[]} sidebarConfig - The `docs` array from the sidebars.js configuration.
+ * @param {object[]} sidebarConfigDocs - The `docs` array from the sidebars.js configuration.
  * @returns {{overviewDocsSection: string, fullConcatContent: string}} - An object containing the generated overview section and the full concatenated content.
```
cprecioso: We can remove the JSDoc typings throughout the file, TS will ignore them in ts files

web/scripts/generate-llm-files.ts
```diff
@@ -328,24 +363,24 @@ async function processBlogFiles() {
  * @param {string} file - The filename to validate.
  * @returns {RegExpMatchArray | null} The result of the match, or null if it doesn't match.
  */
-function validateBlogFileDate(file) {
-  const dateMatch = file.match(/^(\d{4}-\d{2}-\d{2})-.*\.(mdx|md)$/)
+function validateBlogFileDate(file: string): RegExpMatchArray | null {
```
cprecioso: I'd not set the type to `RegExpMatchArray` because the functions outside don't know in which position the info we want is, they have no knowledge of the regex we're using.

You can return that properly, or if wanting to do minimal changes, do something like:
```suggestion
function validateBlogFileDate(file: string): [fullMatch: string, date: string] | null {
```

web/scripts/generate-llm-files.ts
cprecioso: zod all the things hehe

web/scripts/generate-llm-files.ts
infomiho: The `.mts` extension didn't help 🤷 

I've applied the simplification you suggested 👍

web/scripts/generate-llm-files.ts
infomiho: Typescript doesn't rewrite `.ts` extensions to `.js` ([more info](https://www.typescriptlang.org/docs/handbook/modules/theory.html#module-resolution)) and since we are not bundling this script, I have to keep the `.js` extension since the resulting Node.js code won't be able to find the file imported with `.ts`.

web/scripts/generate-llm-files.ts
infomiho: I've went with `string | null` since the outside is interested "is there a valid date inside".

web/scripts/generate-llm-files.ts
cprecioso: TypeScript doesn't but tsx does 😁
Not blocking anyway

web/scripts/generate-llm-files.ts
cprecioso: Huh, it worked for me, but not blocking anyway since it will start giving some ESM compatibilities in other places.

infomiho: When I changed it to `mts` just now, it imports the `sidebars.ts` default import as 
```ts
{ default: [Getter] }
```

I'll just keep it simple since this is not that important right now :)

web/scripts/generate-llm-files.ts
infomiho: Oh yes, `esbuild` is in the background, than I can do it, cool!

 I'll keep it like this for the mere fact I'm used to it :D extrapolating that to believing it will raise less eyebrows for others as well

PR #2909, opened 2025-07-04
Prettier's printWidth to 100

Martinsos:
I was revieiwng Miho's Railway PR and I found it is too aggressive at the moment, look at this image I attached (first image is hat I would like it to be, second is how prettier formats it now), it is ridicoulos -> I can easily fit this line on my screen on my laptop while I have two files open at the same time, with extra space, and yet it brakes it into 4 lines, making this code just harder to read.

image

It seems to break them at 80 chars right now -> I thought we said that that doesn't mean 80 chars stristly, but just some way it makes decisions and it can easily end up being up to 100 in reality. But doesn't seem to be happening. So I would say let's bump this thing to 100? If not, then at least 90, although I thinkn 100 could be just fine.

Because I see tehre are quite a few functions in this PR that fall under 80-95 range that really don't benefit from being broken into multiple lines.

Discussion on Discord: https://discord.com/channels/686873244791210014/1351908377533354025/1390096824261542111 .

Btw Prettier's default is 80, and they recommend to stick with it.
But since I can see right now that I don't like this, I would rather give it a try, go for more, and then revert if we learn that is not good, than stay forever at 80 that seems silly just because it is default.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -8,5 +8,6 @@
   "semi": true,
   "singleQuote": false,
   "jsxSingleQuote": false,
-  "trailingComma": "all"
+  "trailingComma": "all",
+  "printWidth": 100
 }


Martinsos:
@cprecioso is there any other place where I should have added this config?

cprecioso (changes requested):
Nope, that's all, but you'll have to also re-run the format.

Here's how you should do it so everyone can fix conflicts easily:
https://github.com/wasp-lang/wasp/pull/2866#issuecomment-3023345243

And here's what people should do afterwards:
https://github.com/wasp-lang/wasp/pull/2866#issuecomment-3023826186

matijaSos:
nice

Martinsos:
Tnx @cprecioso ! I should do it as part of this PR?

I am also still waiting for others to OK this change first though.

cprecioso:
> I should do it as part of this PR?

@Martinsos IMO yes, the output of this is part of how we evaluate it. Also, otherwise the CI won't go green.

infomiho:
I don't mind the change 👍 Prettier had ease of reading as an argument, but I'd like to try it and see, it's not a problem to reformat everything again if we dislike the change.

It would make sense @Martinsos to maybe run `wrun prettier:format` and then show some before/after here to showcase the effect of the change

Martinsos:
> I don't mind the change 👍 Prettier had ease of reading as an argument, but I'd like to try it and see, it's not a problem to reformat everything again if we dislike the change.
> 
> It would make sense @Martinsos to maybe run `wrun prettier:format` and then show some before/after here to showcase the effect of the change

Ok great! Yeah I will do as @cprecioso said, just wanted to get some buy in before I go into that.

sodic:
> it is ridicoulos -> I can easily fit this line on my screen on my laptop while I have two files open at the same time, with extra space, and yet it brakes it into 4 lines, **making this code just harder to read.**

I disagree with the bold part. I find the second screenshot in the description much nicer than the first one (I like function arguments in multiple lines.

I care more about the arguments on multiple lines than I care about the `printWidth` (Miho and Martin discussed this the other day but I can't find the comment now).

Anyway, I'd probably keep `printWidth` as is but won't complain either way. Like you guys said, t's easily reversible.

cprecioso:
@Martinsos are you still doing this? otherwise should we close this PR?

Martinsos:
Still haven't given up just didn't find the time, let's keep it open!

.prettierrc
```diff
@@ -8,5 +8,6 @@
   "semi": true,
   "singleQuote": false,
   "jsxSingleQuote": false,
```
Martinsos, line 10: This is just a test comment

Martinsos:
Closing since nobody got super excited, I will open again if I feel the problem again (haven't coded much recently in this codebase).

PR #2910, opened 2025-07-04
Fix golden tests failing when accepting new golden tests

FranjoMindek:


sodic:
Please add some context (issue, resoning, etc.) :)

FranjoMindek:
I've added all that I know, but I don't think I have enough background to fully figure this one out in a reasonable time.
Any ideas? @sodic

sodic:
It's probably something to do with Haskell's IO lazyness, but this is plenty to go on. Thanks!

PR #2911, opened 2025-07-04
Refactor env variables

cprecioso:
Extracts and refactors env variables based on https://github.com/wasp-lang/wasp/pull/2796#discussion_r2182454511

### DIFF
--- a/waspc/src/Wasp/Generator/SdkGenerator/EnvValidation.hs
+++ b/waspc/src/Wasp/Generator/SdkGenerator/EnvValidation.hs
@@ -19,7 +19,7 @@ import Wasp.Generator.FileDraft (FileDraft)
 import Wasp.Generator.Monad (Generator)
 import qualified Wasp.Generator.SdkGenerator.Common as C
 import Wasp.Generator.SdkGenerator.JsImport (extImportToImportJson)
-import Wasp.Generator.ServerGenerator.AuthG (jwtSecretEnvVarName)
+import qualified Wasp.Generator.ServerGenerator.AuthG as AuthG
 import qualified Wasp.Generator.ServerGenerator.Common as Server
 import qualified Wasp.Generator.WebAppGenerator.Common as WebApp
 import qualified Wasp.Project.Db as Db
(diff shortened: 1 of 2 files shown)


Martinsos:
All right @cprecioso! I will go to vacation in a day or so, so please have somebody else review it, I thikn @infomiho might be a good candidate. It might also benefit from fresh pair of eyes.

waspc/src/Wasp/Generator/ServerGenerator/Common.hs
```diff
@@ -127,6 +129,12 @@ toESModulesImportPath = changeExtensionTo "js"
   where
     changeExtensionTo ext = (++ '.' : ext) . fst . splitExtension
 
+clientUrlFromServerEnvVarName :: String
```
infomiho: Nitpick: `clientUrlFromServerEnvVarName` feels a bit off as a name because of the "from" (I understand that you wanted to say "from the perspective of") so I'd maybe go with just `clientUrlEnvVarName` since we have the context of this being in the server generator? And I see that you qualified the imports so it reads like `Server.clientUrlEnvVarName`.

waspc/src/Wasp/AI/GenerateNewProject/InitialFiles.hs
```diff
@@ -115,12 +116,13 @@ generateBaseWaspFile newProjectDetails = ((path, content), planRules)
...
   where
+    databaseUrlEnvVarNameText = T.pack databaseUrlEnvVarName
     content =
       [trimming|
       datasource db {
         provider = "sqlite"
-        // Wasp requires that the url is set to the DATABASE_URL environment variable.
-        url      = env("DATABASE_URL")
+        // Wasp requires that the url is set to the ${databaseUrlEnvVarNameText} environment variable.
+        url      = env("${databaseUrlEnvVarNameText}")
```
infomiho: Given that we can't provide the `DATABASE_URL` variable name in the other starters via Haskell, should we just keep it hard-coded here as well? I'm think for consistency sake and maybe some sort of search and replace in the future will put this place in the same results as other starters in the future.

waspc/src/Wasp/Generator/SdkGenerator/EnvValidation.hs
```diff
@@ -58,6 +59,9 @@ genServerEnv spec = return $ C.mkTmplFdWithData tmplPath tmplData
     tmplData =
       object
         [ "isAuthEnabled" .= isJust maybeAuth,
+          "clientUrlEnvVarName" .= Server.clientUrlFromServerEnvVarName,
+          "serverUrlEnvVarName" .= Server.serverUrlFromServerEnvVarName,
+          "jwtSecretEnvVarName" .= jwtSecretEnvVarName,
```
infomiho: I'd either move `jwtSecretEnvVarName` to `ServerGenerator.Common` so we get nice context when using the env var name `Server.jwtSecretEnvVarName` or... if we believe that `AuthG` is a different context, maybe then better to import it using a qualified import so we get `AuthG.jwtSecretEnvVarName`.

infomiho (commented):
Left some suggestions

waspc/src/Wasp/Generator/SdkGenerator/EnvValidation.hs
cprecioso: I left it as `AuthG.jwtSecretEnvVarName`

cprecioso:
@infomiho ready

PR #2915, opened 2025-07-04
Upgrade node requirement to v22.12

cprecioso:
Upgrade Node version to support latest Vite. 

> [!NOTE]
> Vite v7 needs `node ^20.19 || >=22.12`. We'd usually move to Node 22 when it moves out of "Active LTS" and into "Maintenance LTS", but that will happen on October, so it's fair enough that we just do the upgrade now so we don't have to update to Node 20.19 now and 22 in three months


FranjoMindek:
FYI: last time I forgot to update the docs on `open-saas`

waspc/packages/prisma/package.json
```diff
@@ -12,15 +12,13 @@
   },
   "dependencies": {
     "json5": "^2.2.3",
-    "prisma": "Prisma version here needs to be kept in sync with prisma version used in wasp-cli.",
```
infomiho: Why remove the comments?

waspc/ChangeLog.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### ⚠️ Breaking Changes
+
+- We've raised the Node.js requirement to >=v22.12. [#2915](https://github.com/wasp-lang/wasp/pull/2915)
```
infomiho: Nipick:
```suggestion
- Wasp now requires Node.js version to be >=v22.12. [#2915](https://github.com/wasp-lang/wasp/pull/2915)
```

infomiho (approved):
Left some comments, non-blocking

waspc/packages/prisma/package.json
cprecioso: I didn't do it on purpose, `npm install` did due to these not being actual versions. I'll restore them.

PR #2922, opened 2025-07-07
Fix wasp-config package

cprecioso:
There's a bug in Wasp v0.17.0-rc1:

In `wasp-config`, the built TypeScript files are output to `dist/src` instead of just `dist`. This is because we added the one more folder to the TypeScript config [(commit)](https://github.com/wasp-lang/wasp/commit/11909a902999c7d93e12e1f63478e65c8066efae#diff-60eb05a951ff79c3caaf55a3bc4667489fb3635cfdea6f0909ab637428774ee9L23), so TypeScript decides to sub-folder each of them.

My solution is to move the tests inside of `src`, so it will still end up being just one single folder in the TS project.

I tried to have two TS subprojects (`tsconfig.app.json` and `tsconfig.test.json`), but TS was not liking that the test files import app files.

I also explicitly added the types to the exports in the package.json, and removed the `main` reference to the executable file (which should only be `bin`).

### DIFF
--- a/waspc/packages/wasp-config/package-lock.json
+++ b/waspc/packages/wasp-config/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.0.0",
       "license": "MIT",
       "bin": {
-        "wasp-config": "dist/run.js"
+        "wasp-config": "dist/src/run.js"
       },
       "devDependencies": {
         "@eslint/js": "^9.9.0",
--- a/waspc/packages/wasp-config/package.json
+++ b/waspc/packages/wasp-config/package.json
@@ -5,10 +5,12 @@
   "description": "Wasp TS SDK",
   "author": "wasp",
   "license": "MIT",
-  "bin": "./dist/run.js",
-  "main": "dist/run.js",
+  "bin": "./dist/src/run.js",
   "exports": {
-    ".": "./dist/index.js"
+    ".": {
+      "types": "./dist/src/index.d.ts",
+      "default": "./dist/src/index.js"
+    }
   },
   "scripts": {
     "build": "tsc",


sodic:
> In wasp-config, the built TypeScript files are output to dist/src instead of just dist. This is because we added the one more folder to the TypeScript config [(commit)](https://github.com/wasp-lang/wasp/commit/11909a902999c7d93e12e1f63478e65c8066efae#diff-60eb05a951ff79c3caaf55a3bc4667489fb3635cfdea6f0909ab637428774ee9L23), so TypeScript decides to sub-folder each of them.

@cprecioso Why was this a problem? Because of the path for `bin` or something else?

cprecioso:
@sodic it was either changing the paths in package.json or moving the __tests__ into the src. I chose the one which looked most "correct" to me (I don't like having `src` inside `dist`). But no strong opinion and can change to the other option if you think it's better.

sodic:
I'd probably go for changing the path in `package.json` :smile: 
But I'm also not strongly in favor of anything.

Some reasoning:
- It's a simpler change (one line if I'm not mistaken).
- @FranjoMindek probably picked this structure for a reason (probably followed some convention, the name `__tests__` implies so as well), so I'd wait for him before changing it.
- I don't think we should organize our source code based on how the compiled code looks like.

The third one is the most important point, so let's ask ourselves: _"If the dist file wasn't an issue, how would we prefer to structure this?"_ and pick that, regardless of what ends up in the `dist` folder. @cprecioso what do you think?

cprecioso:
@sodic for me the correct solution would be to have `__tests__` in a different ts project, but given that was annoying, I did the other one. But yeah, I agree with your points. Switched to the other solution, take a look.

sodic (approved):
Ok, yeah, let's leave it like this until we properly set up two projects.

@cprecioso Do you mind creating an issue for that?

cprecioso:
yep but i'll wait until @FranjoMindek is back, maybe he has already explored it

PR #2927, opened 2025-07-08
Adds Railway deployment docs

### LINKED ISSUE #2248: Update Railway deployment docs
There have been a few UI changes in the Railway app and we should update our docs to match.

Mostly, users need to "Deploy apps" to apply the variable changes and the CLI `railway link` works a bit differently. Related Discord thread: https://ptb.discord.com/channels/686873244791210014/1276282468088938568/1276301522237657120

infomiho:
This PR:
- updates the Railway deployment instructions 
- adds info about one-line Railway deployment via Wasp CLI

Closes #2248


web/docs/deployment/deployment-methods/paas.md
```diff
@@ -284,132 +285,60 @@ railway link
...
 
-   Copy the `build` folder within the `web-app` directory to `dist`:
+    We used the `--ci` flag to only stream the build logs and not the logs that happen after the client is deployed.
+    </small>
 
-   ```shell
-   cp -r build dist
-   ```
 
-   We'll need to create the following files:
```
infomiho, line 327: This whole bit is redundant because:

> Railway will detect the `index.html` file and deploy the client as a static site using [Railpack](https://railpack.com/languages/staticfile#root-directory-resolution).

waspc/cli/src/Wasp/Cli/Command/Telemetry/Project.hs
```diff
@@ -209,7 +209,7 @@ getProjectTelemetryData userSignature projectHash cmdCall context =
 -- NOTE: If you update the list of keywords here, make sure to also update them in the official docs
 --   on our webpage, under Telemetry.
 extractKeyDeployArgs :: [String] -> [String]
-extractKeyDeployArgs = intersect ["fly", "setup", "create-db", "deploy", "cmd"]
+extractKeyDeployArgs = intersect ["fly", "railway", "setup", "create-db", "deploy", "cmd", "launch"]
```
infomiho, line 212: Added `railway` because we added Railway and `launch` command since it was missing.

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -34,27 +34,25 @@ Using the Wasp CLI, you can easily deploy a new app to [Fly.io](https://fly.io)
 wasp deploy fly launch my-wasp-app mia
 ```
 
-:::caution Unique Name
```
infomiho, line 37: Removing admonition since they were distracting. Content kept as it was before.

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -40,19 +40,19 @@ wasp deploy fly launch my-wasp-app mia
...
 Two things to keep in mind:
 
-1. Your app name (e.g. `my-wasp-app`) must be **unique** across all of Fly or deployment will fail.
+1. Your app name (for example `my-wasp-app`) must be **unique** across all of Fly or deployment will fail.
 
-1. If your account is a member of **more than one organization** on Fly.io, you will need to specify under which one you want to execute the command. To do that, provide an additional `--org <org-slug>` option. You can find out the names(slugs) of your organizations by running `fly orgs list`.
+1. If your account is a member of **more than one organization** on Fly.io, you will need to specify under which one you want to execute the command. To do that, provide an additional `--org <org-slug>` option. You can find out the names (slugs) of your organizations by running `fly orgs list`.
 
-Under the covers, `wasp deploy fly launch` is same as running:
+Under the covers, `wasp deploy fly launch` is the same as running:
```
sodic: Old text, but I believe we wanted to say "under the hood."

"Under the covers" means... something else :smile:

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -116,11 +115,11 @@ wasp deploy fly cmd --context server secrets set WASP_WEB_CLIENT_URL=https://myc
   We need to do this to keep our CORS configuration up to date.
 </small>
 
-That's it, your app should be available at `https://mycoolapp.com`! 🎉
```
sodic, line 116: Oh boy, don't let Martin see this.

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -215,32 +215,31 @@ When the `fly.toml` file exists in .wasp/build/ dir, you do not need to specify
 
 We will show how to deploy the client, the server, and provision a database on Railway.
 
-Railway is a simple and great way to host your server and database. It's also possible to deploy your entire app: database, server, and client. You can use the platform for free for a limited time, or if you meet certain eligibility requirements. See their [plans page](https://docs.railway.app/reference/pricing/plans) for more info.
+:::tip We automated this process for you
```
sodic: I have a feeling this callout is easy to miss. Perhaps we should turn it into normal text or an "important" callout? The same goes for fly.

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -215,32 +215,31 @@ When the `fly.toml` file exists in .wasp/build/ dir, you do not need to specify
 
 We will show how to deploy the client, the server, and provision a database on Railway.
 
-Railway is a simple and great way to host your server and database. It's also possible to deploy your entire app: database, server, and client. You can use the platform for free for a limited time, or if you meet certain eligibility requirements. See their [plans page](https://docs.railway.app/reference/pricing/plans) for more info.
```
sodic, line 218: Looks like we lost this text. It's no longer here and hasn't been moved. Was that on purpose or do we want to keep it?

web/docs/deployment/deployment-methods/paas.md
sodic, line 1: Great job cleaning this up!

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -69,7 +67,11 @@ You'll notice that Wasp creates two new files in your project root directory:
 
 You should include these files in your version control so that you can deploy your app with a single command in the future.
 
-### Using a Custom Domain For Your App
+Wasp CLI sets some [required environment variables](../../project/env-vars.md#wasp-server-env-vars) in the server app automatically, such as `DATABASE_URL`, `WASP_WEB_CLIENT_URL`, and `WASP_SERVER_URL`.
```
sodic: This seems a little random.

I understand what we set and why, but perhaps we can explain it better.

web/docs/deployment/deployment-methods/paas.md
infomiho: I wanted to avoid talking about Railway platform pricing specifics as they get outdated and then we have to maintain that information. Railway doesn't have a free offering anymore, but they offer a 30 day free trial which is not really useful. Their cheapest plan is 5$/month which is pretty reasonable, but I didn't want to put those specifics here.

web/docs/deployment/deployment-methods/paas.md
infomiho: Image

I thought that the admonition is quite prominent, big green box just in the beginning of the provider text.

infomiho: We talked offline and this is the alternative 

Image

web/docs/deployment/deployment-methods/cli.md
infomiho: I've adjusted it to flow better:

> When you run the `launch` command, Wasp CLI sets some ...

infomiho:
@sodic ready for another look

waspc/ChangeLog.md
```diff
@@ -32,6 +32,7 @@ Follow the [the official migration guide](https://wasp.sh/docs/migration-guides/
 - You can now return Prisma `Decimal`s from your Queries and Actions. ([#2701](https://github.com/wasp-lang/wasp/pull/2701))
 - Added build-time client env variable validation. ([#2392](https://github.com/wasp-lang/wasp/pull/2392))
 - Added `prismaSetupFn` hook to database config to allow users to run custom Prisma setup code. ([#2693](https://github.com/wasp-lang/wasp/pull/2693))
+- Added support for one-line deployment with Railway to `wasp deploy` command. ([#2578](https://github.com/wasp-lang/wasp/pull/2578))
```
cprecioso: I'd actually add this a bit higher on the list (first or second)

web/docs/deployment/deployment-methods/cli.md
cprecioso, line 1: I think we could split this file into two separate pages, so the sidebar would turn into something like

```
Deployment Methods > Overview
Deployment Methods > Wasp Deploy > Fly
Deployment Methods > Wasp Deploy > Railway
Deployment Methods > Other platforms
Deployment Methods > Self-hosted
```

because of two reasons:
- We're mixing in the same page, very long instructions that don't need to be read one after the other, nor refer to each other
- It's difficult to find out that Wasp has 1st-class support for Fly and Railway if you're trying to find it in the docs and don't know what "Wasp CLI" means (e.g. here's a user who had been following the PaaS tutorial because the Fly support didn't register: https://discord.com/channels/686873244791210014/1390334847175753738/1390351860812218388)

web/docs/deployment/extras.md
```diff
@@ -20,7 +20,8 @@ It's usually a two-step process, and it's the same for both the client and the s
 
 :::note Using Wasp CLI?
```
cprecioso:
```suggestion
:::note Using `wasp deploy`?
```

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -87,24 +87,19 @@ We'll cover a few different deployment providers below:
 
 ## Fly.io <Server /> <Database /> {#flyio}
 
-We will show how to deploy the server and provision a database for it on Fly.io.
+We recommend that you use [Wasp CLI](./cli.md#flyio) to deploy your Wasp app to Fly.io. Wasp CLI automates deploying the client, the server and the database with one command.
```
cprecioso: I don't think this paragraph registers _at all_.

And I don't know what I would do with this manual instructions. I don't think we're nudging users enough to `wasp deploy`, and I'd like to see them very de-emphasized. Some options:
- At the very least, make this into a big callout with a background color, or better yet its own top-level header, and a link to the `wasp cli > fly` page
- Move this documentation to the "Wasp Deploy > Fly" page I suggested in the comment above, so the structure of the page is something like

	```
	Deploy to Fly
	├── Automatically with `wasp deploy`
	└── Manually
	```

- Actually remove this from the docs proper, move them into a gist and link it from the docs, so it is properly de-emphasized. Another option is to move them to their own page that is only linked from here and not visible in the sidebar.

Same for Railway

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -87,24 +87,19 @@ We'll cover a few different deployment providers below:
...
-Fly.io offers support for both locally built Docker containers and remotely built ones. However, for simplicity and reproducibility, we will default to the use of a remote Fly.io builder.
+1. Create a [Fly.io](https://fly.io/) account,
+1. Install the [`flyctl` CLI](https://fly.io/docs/flyctl/install/),
+1. Log in with the `flyctl` CLI.
 
-Additionally, `fly` is a symlink for `flyctl` on most systems and they can be used interchangeably.
-:::
-
-Make sure you are logged in with `flyctl` CLI. You can check if you are logged in with `flyctl auth whoami`, and if you are not, you can log in with `flyctl auth login`.
+You can check if you are logged in with `flyctl auth whoami`, and if you are not, you can log in with `flyctl auth login`.
```
cprecioso: Fly's docs refer to "Installing flyctl" but then all of the examples they are using the `fly` command instead of `flyctl`. AFAICS Homebrew [installs both](https://github.com/Homebrew/homebrew-core/blob/e0b424cb9aa05ecbf2ce4c7f8f23cedc65875792/Formula/f/flyctl.rb#L42), their [own installer too](https://github.com/Homebrew/homebrew-core/blob/e0b424cb9aa05ecbf2ce4c7f8f23cedc65875792/Formula/f/flyctl.rb#L42), and [most other registries](https://github.com/aquaproj/aqua-registry/blob/764189135f14b73c8f54cf51429e7eb04d48c5f0/pkgs/superfly/flyctl/registry.yaml#L7-L10).

Therefore maybe I'd switch to using `fly` instead of `flyctl` as well

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -249,175 +238,113 @@ Let's create our Railway project:
...
 
-<!-- TOPIC: env vars -->
+    Select `server` when prompted with `Select Service`.
+
+  <!-- TOPIC: env vars -->
 
 3. Go into the Railway dashboard and set up the required env variables:
 
-   Open the `Settings` and go to the `Variables` tab:
+   Click on the `server` serice go to the `Variables` tab:
```
cprecioso:
```suggestion
   Click on the `server` service go to the `Variables` tab:
```

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -213,34 +208,28 @@ When the `fly.toml` file exists in .wasp/build/ dir, you do not need to specify
...
-1. Go to your [Railway dashboard](https://railway.app/dashboard), click on **New Project**, and select `Provision PostgreSQL` from the dropdown menu.
-2. Once it initializes, right-click on the **New** button in the top right corner and select **Empty Service**.
-3. Once it initializes, click on it, go to **Settings > General** and change the name to `server`
-4. Go ahead and create another empty service and name it `client`
-
-![Changing the name](/img/deploying/railway-rename.png)
+1. Go to your [Railway dashboard](https://railway.com/dashboard), click on **New Project**, and select `Deploy PostgreSQL` from the dropdown menu.
+1. Once the project is created, left-click on the **Create** button in the top right corner and select **Empty Service**.
+1. Click on the new service, and change the name to `server`.
+1. Go ahead and create another empty service and name it `client`.
```
cprecioso:
```suggestion
1. Create another empty service and name it `client`.
```

Yes I am in the process of going ahead indeed 😂

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -249,175 +238,113 @@ Let's create our Railway project:
 We'll need the domains for both the `server` and `client` services:
 
 1. Go to the `server` instance's `Settings` tab, and click `Generate Domain`.
-2. Do the same under the `client`'s `Settings`.
-
-Copy the domains as we will need them later.
+1. Enter `8080` as the port and click **Generate Domain**.
+1. Do the same under the `client`'s `Settings`.
+1. Copy the domains as we will need them later.
```
cprecioso: Total nitpick

```suggestion
1. Copy both domains, as we will need them later.
```

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -249,175 +238,113 @@ Let's create our Railway project:
...
 
-```shell
-cd .wasp/build
-```
+    ```shell
+    cd .wasp/build
+    ```
 
-2. Link your app build to your newly created Railway project:
+2. Link your the `build` directory to your newly created Railway project:
```
cprecioso:
```suggestion
2. Link the `.wasp/build` directory to your newly created Railway project:
```

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -249,175 +238,113 @@ Let's create our Railway project:
...
-   - click **Variable reference** and select `DATABASE_URL` (it will populate it with the correct value)
-   - add `WASP_WEB_CLIENT_URL` - enter the `client` domain (e.g. `https://client-production-XXXX.up.railway.app`). `https://` prefix is required!
-   - add `WASP_SERVER_URL` - enter the `server` domain (e.g. `https://server-production-XXXX.up.railway.app`). `https://` prefix is required!
-   - add `JWT_SECRET` - enter a random string at least 32 characters long (use an [online generator](https://djecrety.ir/))
+   1. Click **Variable reference** and select `DATABASE_URL` (it will populate it with the correct value)
+   1. Add `WASP_WEB_CLIENT_URL` - enter the `client` domain (e.g. `https://client-production-XXXX.up.railway.app`). `https://` prefix is required!
+   1. Add `WASP_SERVER_URL` - enter the `server` domain (e.g. `https://server-production-XXXX.up.railway.app`). `https://` prefix is required!
+   1. Add `JWT_SECRET` - enter a random string at least 32 characters long (use an [online generator](https://djecrety.ir/))
 
      <AddExternalAuthEnvVarsReminder />
```
cprecioso, line 288: Can we add some more spacing here somehow? It looks a bit bad on the built page.
image

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -249,175 +238,113 @@ Let's create our Railway project:
...
 3. Go into the Railway dashboard and set up the required env variables:
 
-   Open the `Settings` and go to the `Variables` tab:
+   Click on the `server` serice go to the `Variables` tab:
 
-   - click **Variable reference** and select `DATABASE_URL` (it will populate it with the correct value)
-   - add `WASP_WEB_CLIENT_URL` - enter the `client` domain (e.g. `https://client-production-XXXX.up.railway.app`). `https://` prefix is required!
-   - add `WASP_SERVER_URL` - enter the `server` domain (e.g. `https://server-production-XXXX.up.railway.app`). `https://` prefix is required!
-   - add `JWT_SECRET` - enter a random string at least 32 characters long (use an [online generator](https://djecrety.ir/))
+   1. Click **Variable reference** and select `DATABASE_URL` (it will populate it with the correct value)
```
cprecioso, line 280: I don't think it's very important, when calling out the text in the buttons we're sometimes using **bold**, `code` or "quotes" with no apparent consistency. Very minor, but to take into account for the future.

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -249,175 +238,113 @@ Let's create our Railway project:
...
+   1. Add `WASP_WEB_CLIENT_URL` - enter the `client` domain (e.g. `https://client-production-XXXX.up.railway.app`). `https://` prefix is required!
+   1. Add `WASP_SERVER_URL` - enter the `server` domain (e.g. `https://server-production-XXXX.up.railway.app`). `https://` prefix is required!
+   1. Add `JWT_SECRET` - enter a random string at least 32 characters long (use an [online generator](https://djecrety.ir/))
```
cprecioso: _For future work:_ I'm not in love with linking to an outside "random" website for generating secrets. We can make a button here that does `Array.from(crypto.getRandomValues(new Uint8Array(32)), n => n.toString(16).padStart(2,"0")).join("")` and puts it on the screen.

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -249,175 +238,113 @@ Let's create our Railway project:
...
-4. We need to configure Railway's static hosting for our client.
+    ```shell
+    railway up --ci
+    ```
 
-   :::info Setting Up Static Hosting
+    <small>
 
-   Copy the `build` folder within the `web-app` directory to `dist`:
+    We used the `--ci` flag to only stream the build logs and not the logs that happen after the server is deployed.
```
cprecioso: This annotation is a bit confusing

```suggestion
    Adding the `--ci` flag, we can see the build process happening in real time in our console, and detect any errors easily.
```

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -249,175 +238,113 @@ Let's create our Railway project:
...
+    </small>
 
-   ```shell
-   cp -r build dist
-   ```
+    Railway will now locate the `Dockerfile` and deploy your server.
```
cprecioso:
```suggestion
    Railway will now locate Wasp's `Dockerfile` and deploy your server.
```

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -249,175 +238,113 @@ Let's create our Railway project:
...
 ### Updates & Redeploying
 
 When you make updates and need to redeploy:
 
-- run `wasp build` to rebuild your app
-- run `railway up` in the `.wasp/build` directory (server)
-- repeat all the steps in the `.wasp/build/web-app` directory (client)
+- Run `wasp build` to rebuild your app.
+- Run `railway up --ci` in the `.wasp/build` directory to redeploy the server.
+- Rebuild the client using `npm` and run `railway up --ci` in the `.wasp/build/web-app/build` directory.
```
cprecioso: These instructions assume you know what "rebuild the client using `npm`" means, and that you'll remember to pass `REACT_APP_API_URL` again.

I think this is too off-hand, and is easy to get lost as it assumes you understood what you just did when doing the initial deploy (or that you will remember when doing it again in a week, a month, a year...).

I think it's worth it to be concrete (maybe then this section gets way longer), or to link back to the first deploy tutorial and just add annotations such as `(Only if you're deploying for the first time)` so people know to skip steps

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -34,27 +34,25 @@ Using the Wasp CLI, you can easily deploy a new app to [Fly.io](https://fly.io)
...
 </small>
 
-Under the covers, this runs the equivalent of the following commands:
+Two things to keep in mind:
+
+1. Your app name (for example `my-wasp-app`) must be **unique** across all of Fly or deployment will fail.
+
+1. If your account is a member of **more than one organization** on Fly.io, you will need to specify under which one you want to execute the command. To do that, provide an additional `--org <org-slug>` option. You can find out the names (slugs) of your organizations by running `fly orgs list`.
+
+Under the hood, `wasp deploy fly launch` is the same as running:
```
cprecioso:
```suggestion
Under the hood, `wasp deploy fly launch` is running the same `fly` commands you'd run manually, but with some smart handling for naming and environment variables.

For example, for an app called `my-wasp-app` deployed to the `mia` (Miami, US) region, the CLI would internally call:
```

cprecioso: Actually why do we need to showcase these commands here? Maybe people get confused and run them themselves too. _Maybe they don't even change the name and region 😱_

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -140,9 +142,93 @@ Using the `www` and `non-www` domains at the same time will require you to updat
...
+
+### Prerequisites
+
+:::caution Railpack required
+
+Wasp CLI requires that Railpack is set as the default deployment builder **for client routing to work correctly**.
+
+Go to your [Railway account settings](https://railway.com/account/feature-flags) and enable "Default to Railpack".
+
+:::
```
cprecioso: I'd move this + the "keep in mind that" items to the prerequisites list, just as regular requirements. We don't need to scare users with `:::caution Railpack required`, that's why we have a requirements header! 😊

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -140,9 +142,93 @@ Using the `www` and `non-www` domains at the same time will require you to updat
...
+
+<small>
+  Please do not CTRL-C or exit your terminal while the commands are running.
+</small>
+
+Keep in mind that:
+
+1. Your project name (for example `my-wasp-app`) must be unique across all your Railway projects or deployment will fail. This is a current limitation of the Wasp CLI and Railway integration ([#2926](https://github.com/wasp-lang/wasp/issues/2926)).
+
+1. If you are a member of multiple Railway organizations, the CLI will prompt you to select the organization under which you want to deploy your app.
```
cprecioso: This is what I was referring that we could move to the requirements heading.

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -140,9 +142,93 @@ Using the `www` and `non-www` domains at the same time will require you to updat
...
+wasp deploy railway launch my-wasp-app
+```
+
+<small>
+  Please do not CTRL-C or exit your terminal while the commands are running.
+</small>
+
+Keep in mind that:
+
+1. Your project name (for example `my-wasp-app`) must be unique across all your Railway projects or deployment will fail. This is a current limitation of the Wasp CLI and Railway integration ([#2926](https://github.com/wasp-lang/wasp/issues/2926)).
```
cprecioso: I wouldn't link this here, it doesn't provide actionable information and distracts from the task at hand. We'll just update this when we make the change.

```suggestion
1. Your project name (for example `my-wasp-app`) must be unique across all your Railway projects or deployment will fail.
```

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -140,9 +142,93 @@ Using the `www` and `non-www` domains at the same time will require you to updat
...
+    - Go into the [Railway dashboard](https://railway.app/dashboard).
+    - Select your project (for example `my-wasp-app`).
+    - Click on the client service (for example `my-wasp-app-client`).
+    - Go to the `Settings` tab and click `Custom Domain`.
+    - Enter your domain name (for example `mycoolapp.com`) and port `8080`.
+    - Click `Add Domain`.
+
+2. You need to add the DNS records for your domain:
+
+   _This will depend on your domain provider, but it should be a matter of adding a CNAME record with the values provided by Railway._
```
cprecioso: A bit more explicative

```suggestion
    - Copy the `CNAME` domain you're given

2. Update the DNS records for your domain, adding a CNAME record at the domain or subdomain you want, pointing to the address you've been given in the previous step. _This step depends on your domain provider, consult their documention in case of doubt._
```

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -140,9 +142,93 @@ Using the `www` and `non-www` domains at the same time will require you to updat
...
+
+3. You need to set your new client URL as the `WASP_WEB_CLIENT_URL` environment variable (for example `https://mycoolapp.com`) for your **server service** in the Railway dashboard.
+
+    - Go into the [Railway dashboard](https://railway.app/dashboard).
+    - Select your project (for example `my-wasp-app`).
+    - Click on the server service (for example `my-wasp-app-server`).
+    - Go to the `Variables` tab.
+
+    We need to update the `WASP_WEB_CLIENT_URL` env variable to keep our CORS configuration up to date.
```
cprecioso:
```suggestion
    - Update the `WASP_WEB_CLIENT_URL` variable with the new domain for your client.
```

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -140,9 +142,93 @@ Using the `www` and `non-www` domains at the same time will require you to updat
...
+    - Click on the client service (for example `my-wasp-app-client`).
+    - Go to the `Settings` tab and click `Custom Domain`.
+    - Enter your domain name (for example `mycoolapp.com`) and port `8080`.
+    - Click `Add Domain`.
+
+2. You need to add the DNS records for your domain:
+
+   _This will depend on your domain provider, but it should be a matter of adding a CNAME record with the values provided by Railway._
+
+3. You need to set your new client URL as the `WASP_WEB_CLIENT_URL` environment variable (for example `https://mycoolapp.com`) for your **server service** in the Railway dashboard.
```
cprecioso:
```suggestion
3. To avoid [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS) errors, you need to set your new client URL as the `WASP_WEB_CLIENT_URL` environment variable (for example `https://mycoolapp.com`) for your **server service** in the Railway dashboard.
```

web/docs/deployment/deployment-methods/cli.md
cprecioso: I liked the 🎉, it's a easy-to-see visual signifier that you're done with the tutorial and have accomplished the task you set out to do.

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -165,27 +251,27 @@ wasp deploy fly create-db <region>
...
 
 ```shell
 REACT_APP_ANOTHER_VAR=somevalue wasp deploy fly launch my-wasp-app mia
 ```
 
-### `setup`
+#### `setup` command
 
-`setup` will create your client and server apps on Fly, and add some secrets, but does _not_ deploy them.
+`setup` will create your client and server apps on Fly, and add some secrets. It does _not_ deploy the client or server apps.
```
cprecioso:
```suggestion
`setup` will register your client and server apps on Fly, and setup needed environment variables. It only needs to be run once, when initially creating the app. It does _not_ trigger a deploy for the client or server apps.
```

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -227,53 +313,53 @@ It accepts the following arguments:
...
 
-### `deploy`
+#### `deploy` command
 
 ```shell
 wasp deploy fly deploy
 ```
 
-`deploy` pushes your client and server live.
+The `deploy` command pushes your client and server live.
```
cprecioso:
```suggestion
The `deploy` command pushes your built client and server live.
```

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -311,16 +397,159 @@ You can find the list of all available Fly regions by running:
...
+wasp deploy railway deploy <project-name>
+```
+
+##### Explicitly providing the Railway project ID
+
+Wasp CLI tries to create a new Railway project named `<project-name>`, but you can use an existing project you have created on Railway by passing the `--existing-project-id [projectId]` option.
+
+##### Environment Variables {#railway-launch-environment-variables}
+
+###### Server
```
cprecioso, line 432: IMO Such deep nesting is reflecting the fact that we might need to split this page.

web/docs/deployment/deployment-methods/paas.md
sodic: I agree, although it seemed ok yesterday, I now don't find the nudge obvious enough.

@cprecioso, did you see this thread: https://github.com/wasp-lang/wasp/pull/2927#discussion_r2200268121. We used to have a callout, and then I asked Miho to remove it.

I prefer the other two ideas to the callout. We'll probably be handling this separately though.

web/docs/deployment/deployment-methods/cli.md
sodic: No no, I meant, the list of environment variables seems random and arbitrary to someone who doesn't know Wasp internals.

web/docs/deployment/deployment-methods/cli.md
sodic: I agree, nice catch!

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -140,9 +142,93 @@ Using the `www` and `non-www` domains at the same time will require you to updat
...
+
+When you run the `launch` command, Wasp CLI sets some [required environment variables](../../project/env-vars.md#wasp-server-env-vars) in the server app automatically.
+
+If you have any additional environment variables that your app needs, read how to set them in the [API Reference](#railway-environment-variables).
+
+### Using a Custom Domain For Your App {#railway-custom-domain}
+
+Setting up a custom domain is a three-step process:
+
+1. You need to add your domain to your Railway client service:
```
sodic:
```suggestion
1. Add your domain to the Railway client service:
```

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -285,7 +371,7 @@ or
 REACT_APP_ANOTHER_VAR=somevalue wasp deploy fly deploy
 ```
 
-Please note, you should do this for **every deployment**, not just the first time you set up the variables. One way to make sure you don't forget to add them is to create a `deploy` script in your `package.json` file:
+Please note, that you should do this for **every deployment**, not just the first time you set up the variables. One way to make sure you don't forget to add them is to create a `deploy` script in your `package.json` file:
```
sodic:
```suggestion
Please note that you should do this for **every deployment**, not just the first time you set up the variables. One way to make sure you don't forget to add them is to create a `deploy` script in your `package.json` file:
```

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -311,16 +397,159 @@ You can find the list of all available Fly regions by running:
...
+
+#### Environment Variables {#railway-environment-variables}
+
+##### Server Secrets
+
+If your app requires any other server-side environment variables (like social auth secrets), you can set them:
+
+1. initially in the `launch` or `setup` commands with the [`--server-secret` option](#railway-launch-environment-variables),\
+   or
+2. after the app has already been deployed go into the Railway dashboard and set them in the `Variables` tab of your server service.
```
sodic: This is a little weird: a numbered list as a continuation of a sentence with an "or" in between.

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -311,16 +397,159 @@ You can find the list of all available Fly regions by running:
...
+REACT_APP_ANOTHER_VAR=somevalue wasp deploy railway launch my-wasp-app
+```
+
+or
+
+```shell
+REACT_APP_ANOTHER_VAR=somevalue wasp deploy railway deploy
+```
+
+Please note, that you should do this for **every deployment**, not just the first time you set up the variables. One way to make sure you don't forget to add them is to create a `deploy` script in your `package.json` file:
```
sodic:
```suggestion
Please note that you should do this for **every deployment**, not just the first time you set up the variables. One way to make sure you don't forget to add them is to create a `deploy` script in your `package.json` file:
```

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -311,16 +397,159 @@ You can find the list of all available Fly regions by running:
...
+- `--skip-client` - do not deploy the web client
+- `--skip-server` - do not deploy the server
+
+If you've added any [client-side environment variables](../../project/env-vars#client-env-vars) to your app, make sure to pass them to the terminal session before running the `deploy` command, for example:
+
+```shell
+REACT_APP_ANOTHER_VAR=somevalue wasp deploy railway deploy my-wasp-app
+```
+
+Make sure to add your client-side environment variables every time you redeploy with the above command [to ensure they are included in the build process](../../project/env-vars#client-env-vars-1).
```
sodic:
```suggestion
You must specify your client-side environment variables every time you redeploy with the above command [to ensure they are included in the build process](../../project/env-vars#client-env-vars-1).
```

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -311,16 +397,159 @@ You can find the list of all available Fly regions by running:
...
+When you run the `deploy` command, Wasp CLI will use the Railway project that's linked to the Wasp project directory. If no Railway project is linked, the command will fail asking you to run the `setup` command first.
+
+If you are deploying your Railway app in the CI, you can pass the `--existing-project-id [projectId]` option to tell Wasp CLI the Railway project ID to use for the deployment.
+
+##### Other Available Options
+
+- `--skip-client` - do not deploy the web client
+- `--skip-server` - do not deploy the server
+
+If you've added any [client-side environment variables](../../project/env-vars#client-env-vars) to your app, make sure to pass them to the terminal session before running the `deploy` command, for example:
```
sodic:
```suggestion
If you've added any [client-side environment variables](../../project/env-vars#client-env-vars) to your app, pass them to the terminal session before running the `deploy` command, for example:
```
There are some more "make sure"s we can probably get rid of.

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -311,16 +397,159 @@ You can find the list of all available Fly regions by running:
...
+
+`launch` is a convenience command that runs `setup` and `deploy` in sequence.
+
+```shell
+wasp deploy railway launch <project-name>
+```
+
+It accepts the following arguments:
+
+- `<project-name>` - the name of your project <Required />
```
sodic: I think we usually put the required label before the description. Also, the stuff after the dash is capitalized.

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -311,16 +397,159 @@ You can find the list of all available Fly regions by running:
...
+
+It gives you the same result as running the following commands:
```
sodic:
```suggestion
Running `wasp deploy railway launch` you the same result as running the following commands:
```

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -311,16 +397,159 @@ You can find the list of all available Fly regions by running:
...
-### Building Locally
+#### Building Locally
 
 Fly.io offers support for both **locally** built Docker containers and **remotely** built ones. However, for simplicity and reproducibility, the CLI defaults to the use of a remote Fly.io builder.
 
 If you want to build locally, supply the `--build-locally` option to `wasp deploy fly launch` or `wasp deploy fly deploy`.
+
+### Railway CLI Commands {#railway-cli}
+
+#### `launch` command
```
sodic:
```suggestion
#### The `launch` command
```
Same applies to other similar titles

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -311,16 +397,159 @@ You can find the list of all available Fly regions by running:
...
+It gives you the same result as running the following commands:
+
+```shell
+wasp deploy railway setup <project-name>
+wasp deploy railway deploy <project-name>
+```
+
+##### Explicitly providing the Railway project ID
+
+Wasp CLI tries to create a new Railway project named `<project-name>`, but you can use an existing project you have created on Railway by passing the `--existing-project-id [projectId]` option.
```
sodic:
```suggestion
By default, Wasp CLI tries to create a new Railway project named `<project-name>`. If you want to use an existing Railway project, pass its ID with `--existing-project-id [projectId]`.
```
Or something along those lines. Perhaps we should also add a codebox below to show how it's used (and then possibly remove the inline argument from the sentence).

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -311,16 +397,159 @@ You can find the list of all available Fly regions by running:
...
+
+If you are deploying an app that requires any other environment variables (like social auth secrets), you can set them with the `--server-secret` option:
+
+```
+wasp deploy railway launch my-wasp-app --server-secret GOOGLE_CLIENT_ID=<...> --server-secret GOOGLE_CLIENT_SECRET=<...>
+```
+
+###### Client
+
+If you've added any [client-side environment variables](../../project/env-vars#client-env-vars) to your app, make sure to pass them to the terminal session before running the `launch` command, for example:
```
sodic:
```suggestion
If you've added any [client-side environment variables](../../project/env-vars#client-env-vars) to your app, pass them to the terminal session before running the `launch` command, for example:
```

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -311,16 +397,159 @@ You can find the list of all available Fly regions by running:
...
+
+```shell
+wasp deploy railway deploy my-wasp-app
+```
+
+##### Explicitly providing the Railway project ID
+
+When you run the `deploy` command, Wasp CLI will use the Railway project that's linked to the Wasp project directory. If no Railway project is linked, the command will fail asking you to run the `setup` command first.
+
+If you are deploying your Railway app in the CI, you can pass the `--existing-project-id [projectId]` option to tell Wasp CLI the Railway project ID to use for the deployment.
```
sodic: Seems like we're repeating this part twice.

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -311,16 +397,159 @@ You can find the list of all available Fly regions by running:
...
+
+```shell
+REACT_APP_ANOTHER_VAR=somevalue wasp deploy railway deploy my-wasp-app
+```
+
+Make sure to add your client-side environment variables every time you redeploy with the above command [to ensure they are included in the build process](../../project/env-vars#client-env-vars-1).
+
+#### `setup` command
+
+`setup` will create your client, server, and database services on Railway, and configure environment variables. It does _not_ deploy the client or server services.
```
sodic:
```suggestion
The `setup` command creates your client, server, and database services on Railway, and configures environment variables. It does _not_ deploy the client or server services.
```
Avoids mixing tenses.

I recommend sticking to the present tense and avoiding "will" in the entire text whenever possible.

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -311,16 +397,159 @@ You can find the list of all available Fly regions by running:
...
+The project name is used as a base for your server and client service names on Railway:
+
+- `<project-name>-client`
+- `<project-name>-server`
+
+Railway will also create a PostgreSQL database service named `Postgres`.
+
+##### Explicitly providing the Railway project ID
+
+You can use an existing project you have created on Railway by passing the `--existing-project-id [projectId]` option.
```
sodic: I think this part is repeated 3 times in the API reference.

web/docs/deployment/deployment-methods/paas.md
sodic: Very cool!

Care to create an issue @cprecioso. This is also a great task for outside contributors.

web/docs/deployment/deployment-methods/paas.md
sodic: Yes, I agree, we shouuld make it less noisy. I seem to remember this happens automatically if you add a space somewhere, Docosaurus does some weird formatting.

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -249,175 +238,113 @@ Let's create our Railway project:
...
-     ```bash title="deploy_client.sh"
-     #!/usr/bin/env bash
+#### Deploying the Client
 
-     if [ -z "$REACT_APP_API_URL" ]
-     then
-       echo "REACT_APP_API_URL is not set"
-       exit 1
-     fi
+:::caution Make sure to enable Railpack
```
sodic: We shold probably move this into the intro/prerequisites.

web/docs/deployment/deployment-methods/paas.md
cprecioso: I saw it after I had written this, but still thought I wanted to give my opinion 😉

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -249,175 +238,113 @@ Let's create our Railway project:
...
-     node_modules/
-     EOF
-     )
+    ```shell
+    cd web-app
+    ```
 
-     echo "$dockerfile_contents" > Dockerfile
-     echo "$dockerignore_contents" > .dockerignore
+2. Create the production build, using the `server` domain as the `REACT_APP_API_URL`:
```
sodic, line 311: Should we mention other possible client environment variables here?

web/docs/deployment/deployment-methods/cli.md
infomiho: Yes, but our writing guides [discourages use of emojis](https://wasp.sh/docs/writingguide#grammar).

web/docs/deployment/deployment-methods/cli.md
infomiho: That's an interesting idea! I'll try it out after I finish all the other comments 👍

web/docs/deployment/deployment-methods/paas.md
infomiho: Yes, but you see the logs even without the `--ci` flag, but you see both _build logs_ and the logs that stream _after the deployment_. So I wanted to say "hey, this flag will help us only see the logs up until the point the app deploys, not after". I'll try to reword it 👍

web/docs/deployment/deployment-methods/cli.md
infomiho: We had the practice of linking to issues about things users might ask us about. For example, I'm pointing out an artificial limitation that we introduced, maybe the user wants to say "Hey make this happens pls" and that's why I'm giving them a way to do it. 

What if I put the sentence in parentheses?

web/docs/deployment/deployment-methods/cli.md
infomiho: You are 100% right, I wanted to do it, but I thought it might be too out of scope for this PR. But hey, why not do it right, right away? 😄

web/docs/deployment/deployment-methods/cli.md
infomiho: > I think we usually put the required label before the description

SCR-20250714-ocnx

This is how currently deployed docs look like.

> Also, the stuff after the dash is capitalized.

👍

web/docs/deployment/deployment-methods/cli.md
infomiho: Yes, for different commands, but with a little of twist. For `launch` it will try to create a new project if there is no linked project, for `deploy` it will fail if there is no linked project.

web/docs/deployment/deployment-methods/cli.md
infomiho: Yes, for different commands, I want to provide the info in context at cost of being repetitive.

web/docs/deployment/deployment-methods/paas.md
infomiho: We don't specify any other variables, you mean some variables that the user might have defined?
SCR-20250714-odsc

web/docs/deployment/deployment-methods/cli.md
cprecioso: the the writing guide must change 😡😡😡😡😂

web/docs/deployment/deployment-methods/paas.md
cprecioso: #2956

web/docs/deployment/deployment-methods/cli.md
infomiho: I'll jump on this after we polish the "smaller" stuff here: https://github.com/wasp-lang/wasp/pull/2957

web/docs/deployment/deployment-methods/paas.md
infomiho: I've reworded this line a bit 👍

web/docs/deployment/deployment-methods/cli.md
infomiho: I've moved the Railpack warning to the prerequisites list, but I've kept the "Keep in mind" below because it's not really a prerequisite for running `wasp deploy railway`.

web/docs/deployment/deployment-methods/cli.md
infomiho: For me it doesn't fall under the prerequisites to use `wasp deploy railway` - it's a different kind of thing, a limitation, a tip etc. not really a blocker for using the Wasp CLI to deploy to Railway.

web/docs/deployment/deployment-methods/cli.md
infomiho: Removed it 👍

web/docs/deployment/deployment-methods/cli.md
infomiho: I've put in parentheses now 👍

web/docs/deployment/deployment-methods/cli.md
infomiho: Settle down past Mihovil, we decided to do it in a separate PR now: #2957

web/docs/deployment/deployment-methods/cli.md
infomiho: Removed all `make sure to` to be more direct.

web/docs/deployment/deployment-methods/cli.md
infomiho: SCR-20250715-lkwg

Updated it to look like this 👍

infomiho:
@sodic @cprecioso I think I've address most of the comments now, please take a look when you can.

I've not addressed the question of restructuring the Wasp Deploy docs (I'll do it here #2957) so this is our current callout:

image

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -203,221 +203,164 @@ While we will improve this process in the future, in the meantime, you have a fe
...
 
-1. Move into your app's `.wasp/build/` directory:
+1. Move into the `.wasp/build` directory:
 
-```shell
-cd .wasp/build
-```
+    ```shell
+    cd .wasp/build
+    ```
```
cprecioso, line 264: Nice catch!

cprecioso (approved):

web/docs/deployment/deployment-methods/cli.md
sodic: I think we still don't get each other :smile: 

So, the problem is...
It's weird to say "we'll set some of the variables for you, such as..." without explicitly listing all of them and explaining why they are special. Removing the list of variables does not improve the situation, I'd argue it makes it worse.

Imagine being a user and reading this:
> When you run the launch command, Wasp CLI sets some [required environment variables](http://localhost:3002/docs/project/env-vars#wasp-server-env-vars) in the server app automatically. If your app requires any additional environment variables, use the wasp deploy fly cmd secrets set command. Read more in the [API Reference](http://localhost:3002/docs/deployment/deployment-methods/cli#flyio-cli-environment-variables).

You don't know what it sets or why it sets it. You can go to the API reference or the linked env variable page, but neither will answer those questions - Until you try it out, you have no idea what needs to be set and what doesn't. There's virtually no way your deployment will work on the first try without some trial and error.

The original version had a list of the things the command will set, that's good. The problems were:
- The "such as" part - We should list all of them and make it claer that those aren't examples and are in fact everything Wasp sets. I actually think we did list all of them but the wording wasn't clear on that.
- No explanation on why those environment variables are special.

So, I imagined something like:

> Since Wasp deployes all parts of your application and decies on their names, it also sets all environment variables needed to connect them, specifically: ...
You still have to set all other environment (e.g., ...) variables yourself

This is just a rough idea, but I trust you get what I'm going for.

web/docs/deployment/deployment-methods/cli.md
sodic: This is ok, but if we want to emphasize that these are mutually exclusive, we can do something like:
> If your app requires ...., there are two ways you can set them:

Or
> If your app requires ...., you can set them either:

But this is fine too, so resolve when your happy.

web/docs/deployment/deployment-methods/cli.md
sodic: Great!

Yeah, I have to admit I was looking at other pages, not this one - forgot parts of it were already online.

In any case, I'm happy with the change :)

web/docs/deployment/deployment-methods/cli.md
sodic: Fair enough!

web/docs/deployment/deployment-methods/cli.md
sodic: There's a missing s in "configures."

Also, "It also configures environment variables" could be a new sentance, since we're already listing stuff.

web/docs/deployment/deployment-methods/paas.md
sodic: I thought we should say something like "and any other client env variables you might have defined," but I see we don't do it for server either.

I'm not sure whether to add that and where

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -140,9 +133,76 @@ Using the `www` and `non-www` domains at the same time will require you to updat
...
+
+### Prerequisites
+
+To deploy to Railway using Wasp CLI:
+
+1. Create a [Railway](https://railway.com/) account,
+
+1. Install the [`railway` CLI](https://docs.railway.com/guides/cli#installing-the-cli) on your machine.
+
+1. Wasp CLI requires that Railpack is set as the default deployment builder **for client routing to work correctly**. Go to your [Railway account settings](https://railway.com/account/feature-flags) and enable "Default to Railpack".
```
sodic, line 147: I'd flip the two sentences. First tell them what to do, then why. I think the flow is better

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -140,9 +133,76 @@ Using the `www` and `non-www` domains at the same time will require you to updat
...
+wasp deploy railway launch my-wasp-app
+```
+
+<small>
+  Please do not CTRL-C or exit your terminal while the commands are running.
+</small>
+
+Keep in mind that:
+
+1. Your project name (for example `my-wasp-app`) must be unique across all your Railway projects or deployment will fail. (This is a current limitation of the Wasp CLI and Railway integration [#2926](https://github.com/wasp-lang/wasp/issues/2926))
```
sodic:
```suggestion
1. Your project name (for example `my-wasp-app`) must be unique across all your Railway projects or deployment will fail (this is a current limitation of the Wasp CLI and Railway integration [#2926](https://github.com/wasp-lang/wasp/issues/2926))
```

web/docs/deployment/deployment-methods/cli.md
```diff
@@ -140,9 +133,76 @@ Using the `www` and `non-www` domains at the same time will require you to updat
...
+
+1. If you are a member of multiple Railway organizations, the CLI will prompt you to select the organization under which you want to deploy your app.
+
+The project name is used as a base for your server and client service names on Railway:
+
+- `my-wasp-app-client`
+- `my-wasp-app-server`
+
+Railway doesn't allow setting the database service name using the Railway CLI, so it will always be named `Postgres`.
```
sodic:
```suggestion
Railway doesn't allow setting the database service name using the Railway CLI. It will always name it "Postgres".
```

web/docs/deployment/deployment-methods/cli.md
infomiho: Now I understand the issue, you want to be more precise what it is doing and why. I'll reword it based on your suggestions 👍

infomiho: Updated [`05fc713` (#2927)](https://github.com/wasp-lang/wasp/pull/2927/commits/05fc71396d9584e8d8f8505b4e83248543dad3df)

Issue #1769, opened 2024-02-14
Improve how we generate code (especially SDK) during `wasp build`


Martinsos:
### Update: Read [the comment below](https://github.com/wasp-lang/wasp/issues/1769#issuecomment-1952938100) for the latest info

-------

## Original description
Since after restructuring we have the line `"wasp": "file:.wasp/out/sdk/wasp",` in user's `package.json`, this meant that we also need to change something about how Wasp project is built.

Why? Because `wasp build` generates code in `.wasp/build`, while `wasp start` generates wasp code in `.wasp/out`. This was done on purpose, because we didn't want those two to mix, since the generated code can be (and is, although not a lot) different.

So `wasp start` generates framework code + SDK code in `.wasp/out`, and that code imports and uses user's code from the root of the project.
`wasp build` works the same way, but writes generated framework code + SDK code into `./wasp/build`.

The problem, as I said above, is that now, with restructuring, this won't work well for code generated with `wasp build` because user's code, via its package.json, depends on SDK in `.wasp/out/sdk`, while it should depend on SDK in `./wasp/build/sdk`.

We discussed how to best fix this, and what we did for the moment, is had `wasp build` write its SDK code into `./wasp/out/sdk` instead of `./wasp/build/sdk`, overwriting `./wasp/out/sdk`.
This is a hack, but it is not problematic because currently SDK is exactly the same code be it for `wasp start` or `wasp build` -> code is different for generated framework code, but not for SDK.

Also, we did this in a bit hacky fashion in the code: search for `../out/sdk/wasp` in the codebase to see what was done.

We should however look into a nicer solution, because this one is, as mentioned, somewhat hacky, and also could be causing some issues if in the future there would appear some differences between the SDK code generated for the `wasp start` and the one generated for the `wasp build`.

What seems to be the best idea for the future is to also copy all the user's code into the `.wasp/build` directory, instead of leaving the user's code as it is. Then, when copying that code, we can rewrite its package.json to have a different path for importing SDK, one that imports the SDK generated by `wasp build`. One way to do it is to put user's code somewhere next to generated and framework code, update imports of user's code in the generated code as needed, and that should work. Or, another idea is to copy the user's code into `.wasp/build`, but then not generate the framework and SDK code directly in `.wasp/build`, but put it two levels lower, potentially in `.wasp/build/.wasp/out` -> this way we replicate the relative paths among generated code and user's code, and don't have to modify imports of user's code. Maybe there are some better ideas also.
[...]

PR #2929, opened 2025-07-08
Fix docker build for apps with Wasp TS Spec

cprecioso:
- Bug introduced in #2458 

In `wasp build` we copied just `tsconfig.json` because I didn't realize in the TS Spec we were using `tsconfig.src.json`. When you try to `docker build .wasp/build` in a TS Spec app, the TypeScript checking step currently fails because the main `tsconfig.json` just references `tsconfig.src.json`.

We now inspect the AppSpec and correctly copy the specific file we're expecting.

### DIFF
--- a/waspc/cli/src/Wasp/Cli/Command/Build.hs
+++ b/waspc/cli/src/Wasp/Cli/Command/Build.hs
@@ -24,7 +24,18 @@ import Wasp.Generator.Common (ProjectRootDir)
 import Wasp.Generator.Monad (GeneratorWarning (GeneratorNeedsMigrationWarning))
 import Wasp.Generator.SdkGenerator.Common (sdkRootDirInGeneratedCodeDir, sdkRootDirInProjectRootDir)
 import qualified Wasp.Message as Msg
-import Wasp.Project.Common (CompileError, CompileWarning, WaspProjectDir, buildDirInDotWaspDir, dotWaspDirInWaspProjectDir, generatedCodeDirInDotWaspDir, getSrcTsConfigInWaspProjectDir, packageJsonInWaspProjectDir, packageLockJsonInWaspProjectDir, srcDirInWaspProjectDir)
+import Wasp.Project.Common
+  ( CompileError,
+    CompileWarning,
+    WaspProjectDir,
+    buildDirInDotWaspDir,
+    dotWaspDirInWaspProjectDir,
+    generatedCodeDirInDotWaspDir,
+    getSrcTsConfigInWaspProjectDir,
+    packageJsonInWaspProjectDir,
+    packageLockJsonInWaspProjectDir,
+    srcDirInWaspProjectDir,
+  )
 import Wasp.Project.WaspFile (findWaspFile)
 import Wasp.Util.IO (copyDirectory, copyFile, doesDirectoryExist, removeDirectory)
 import Wasp.Util.Json (updateJsonFile)


waspc/cli/src/Wasp/Cli/Command/Build.hs
```diff
@@ -49,6 +49,8 @@ import Wasp.Util.Json (updateJsonFile)
 build :: Command ()
 build = do
   InWaspProject waspProjectDir <- require
+  appSpec <- analyze waspProjectDir
```
sodic: I can't take a detailed look at this now, but this is very likely not the way to fix this.

We call analyze here and deal with the `appSpec` waay too high in terms of abstraction. In fact, this function now calls the analysis twice - both here and down the call stack of `buildIO`.

@infomiho @cprecioso let me know if you guys need help figuring this out.

cprecioso: > We call analyze here and deal with the appSpec waay too high in terms of abstraction.  In fact, this function now calls the analysis twice - both here and down the call stack of buildIO.

tbh i don't think this is very high in the abstraction ladder, it stand to reason that we need a parsed AppSpec before doing stuff with the app. i would go further and turn the analyze into a requirement (so it is deduplicated) and have something like `ParsedAppSpec appSpec <- require`

infomiho: We do a similar thing in multiple different commands e.g.:
```hs
info :: Command ()
info = do
  InWaspProject waspDir <- require

  compileInfo <- liftIO $ readWaspCompileInfo waspDir
  projectSize <- liftIO $ readDirectorySizeMB waspDir

  appSpec <- analyze waspDir
```

The difference here is definitely that `build` calls `compileWithOptions`  which calls `analyzeWithOptions` which means we call `analyzeWithOptions` twice (`analyze` is just `analyzeWithOptions` with default options).

_One potentially dangerous thing could be that `build` is not using default options, so the app spec you get the first time (`isBuild: False`). could be different than the app spec the second time (`isBuild: True`)._

Requirements system might be the solution or it might be an overkill? If we used the requirements system to cache the `AppSpec` that would make passing `appSpec` in a lot of places as arguments unnecessary which would be a nice win. 

I'm not sure about this one, great catch @sodic

infomiho: Build options:
Image

Default options:
Image

infomiho: If the `buildIO` returned the `appSpec` somehow - that could work as well, is it weird to return errors, warnings and app spec?

Is the location of the `tsconfig.src.json` wrong, should it even be in the app spec? Is it okay if we get it outside of the app spec _again_ just for the purpose of this fix?

Is copying files needed for Docker build too high in the program, should it be inside somewhere where it has access to app spec e.g. generator?

cprecioso: > Is it okay if we get it outside of the app spec again just for the purpose of this fix?

I tried to see that but it seems the details on how to get this are all worked out inside `analyze`, I'm not sure it'd make sense to duplicate that.

> Is copying files needed for Docker build too high in the program, should it be inside somewhere where it has access to app spec e.g. generator?

With my limited knowledge, that sounds like something

infomiho: We already have `FileDraftCopyDirFd` and `FileDraftCopyAndModifyTextFd` file drafts so in theory we might be able to do the changes we need to the Docker files by using the generator and not IO directly. 

Hacky solution might be going into `generateCode` where we have app spec but we are also in IO:
```hs
generateCode ::
  AS.AppSpec ->
  Path' Abs (Dir Generator.ProjectRootDir) ->
  CompileOptions ->
  IO ([CompileWarning], [CompileError])
generateCode appSpec outDir options = do
  (generatorWarnings, generatorErrors) <- Generator.writeWebAppCode appSpec outDir (sendMessage options)
  let filteredWarnings = generatorWarningsFilter options 

  when (CO.isBuild options) $ do
    prepareFilesNecessaryForDockerBuild (CO.waspProjectDirPath options) outDir >>= \case
      -- TODO: throw a better error here
      Left err -> throwIO $ userError $ "Failed to prepare files necessary for Docker build: " ++ show err
      Right () -> return ()

  return (show <$> filteredWarnings, show <$> generatorErrors)
```

waspc/cli/src/Wasp/Cli/Command/Build.hs
```diff
@@ -121,13 +123,15 @@ build = do
...
-          (waspProjectDir </> srcTsConfigInWaspLangProject)
+          (waspProjectDir </> srcTsConfigPath)
           tsconfigJsonInBuildDir
 
       -- A hacky quick fix for https://github.com/wasp-lang/wasp/issues/2368
       -- We should remove this code once we implement a proper solution.
       ExceptT $ updateJsonFile removeWaspConfigFromDevDependenciesArray packageJsonInBuildDir
       ExceptT $ updateJsonFile removeAllMentionsOfWaspConfigInPackageLockJson packageLockJsonInBuildDir
+      where
+        srcTsConfigPath = AS.srcTsConfigPath appSpec
```
sodic: We should always avoid nested where blocks, they're a common Haskell antipattern.

The reason is that they make the control flow difficult to follow and allow too much mixing of monadic and non-nomadic code. In short, whenever there's a nested `where`, the code can be improved to not have it. 

In this case, you can do it by immediately getting what you want out of the `appSpec`:
```haskell
srcTsConfigPath <- AS.srcTsConfigPath <$> analyze waspProjectDir
```
We'll be changing this code anyway, but this knowledge will still come in handy :)

waspc/cli/src/Wasp/Cli/Command/Build.hs
sodic: @infomiho and I had a discussion on this and the consensus was:
- We want to do a small non-risky change to unblock the release, so no changes in approach.
- We'll prioritize removing this hack (which means solving [1769](https://github.com/wasp-lang/wasp/issues/1769)) ASAP.

@cprecioso With that in mind, the best solution is determining the proper TS config file from the `waspProjectRoot` which is already available. To see how to do that, check this function: https://github.com/wasp-lang/wasp/blob/769cef2fca3b66d904a8150f910b3b49b47e846b/waspc/src/Wasp/Project/Analyze.hs#L56

I've highlighted the important parts:

image

This will still have some duplication, but it's better than running the entire analysis step twice with different arguments.

sodic (commented):
Left a suggestion for the fix

waspc/cli/src/Wasp/Cli/Command/Build.hs
infomiho: Added the way to do it with generator to the linked issue: https://github.com/wasp-lang/wasp/issues/1769#issuecomment-3052204904

Let's keep the hack going for a bit longer, not to get stuck here and let's fix it the proper way ASAP.

waspc/cli/src/Wasp/Cli/Command/Build.hs
```diff
@@ -105,7 +99,7 @@ build = do
 
       let packageJsonInBuildDir = buildDir </> castRel packageJsonInWaspProjectDir
       let packageLockJsonInBuildDir = buildDir </> castRel packageLockJsonInWaspProjectDir
-      let tsconfigJsonInBuildDir = buildDir </> castRel srcTsConfigInWaspLangProject
+      let tsconfigJsonInBuildDir = buildDir </> basename srcTsConfigPath
```
sodic: Why do we need `basename` here? I think that's a mistake that works because `tsconfig.json` happens to be at the top level of Wasp's project dir. `castRel` is the correct option, I believe.

sodic (commented):
One more question

cprecioso:
Thanks for working it out and for the guidance!
@sodic @infomiho ready

waspc/cli/src/Wasp/Cli/Command/Build.hs
cprecioso: Yep, sorry, in a previous version it was only working with `basename` and when I changed it I didn't think to check. Fixed!

cprecioso:
@sodic @infomiho ready!

waspc/cli/src/Wasp/Cli/Command/Build.hs
```diff
@@ -24,18 +24,8 @@ import Wasp.Generator.Common (ProjectRootDir)
 import Wasp.Generator.Monad (GeneratorWarning (GeneratorNeedsMigrationWarning))
 import Wasp.Generator.SdkGenerator.Common (sdkRootDirInGeneratedCodeDir, sdkRootDirInProjectRootDir)
 import qualified Wasp.Message as Msg
-import Wasp.Project.Common
```
infomiho: I'd revert his formatting change, it's more in line what we have in the rest of the codebase

waspc/cli/src/Wasp/Cli/Command/Build.hs
```diff
@@ -12,7 +12,7 @@ import Data.Aeson.Lens
 import qualified Data.HashMap.Strict as HM
 import Data.List (isSuffixOf)
 import Data.Text (Text, unpack)
-import StrongPath (Abs, Dir, Path', castRel, (</>))
+import StrongPath (Abs, Dir, Path', basename, castRel, (</>))
```
infomiho: Unused import

waspc/cli/src/Wasp/Cli/Command/Build.hs
cprecioso: I didn't do it manually, this was ormolu

infomiho: Ormolu takes hints from your code (it's opposite of Prettier which always generates the same output), so you'd have to add a newline and it will format it in a new way. Anyho, I'd format it like it was before to keep it consistent 😄

PR #2930, opened 2025-07-08
Avoid TypeScript confusion with Decimal type

cprecioso:
@infomiho got this error while building `ask-the-documents`:

```
core/serialization/prisma.ts:30:47 - error TS2448: Block-scoped variable 'Decimal' used before its declaration.

30 const Decimal = (Prisma as { Decimal?: typeof Decimal }).Decimal;
                                                 ~~~~~~~

  core/serialization/prisma.ts:30:7
    30 const Decimal = (Prisma as { Decimal?: typeof Decimal }).Decimal;
             ~~~~~~~
    'Decimal' is declared here.


Found 1 error in core/serialization/prisma.ts:30
```

It seems TS is getting confused with the `Decimal` type being the same name of the `Decimal` variable. We'll just rename everything so it doesn't get confused.

### DIFF
--- a/waspc/data/Generator/templates/sdk/wasp/core/serialization/prisma.ts
+++ b/waspc/data/Generator/templates/sdk/wasp/core/serialization/prisma.ts
@@ -26,8 +26,9 @@ import "./custom-register"
   property, and give it the `Decimal` type that we take from option (b). Importantly, we only
   do a type import from (b) so we don't trigger the runtime error when importing from the client.
 */
-type Decimal = import("@prisma/client/runtime/library").Decimal;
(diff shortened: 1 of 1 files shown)


infomiho:
Why it happened for `ask-the-documents` example app and not in our CI? Because `ask-the-documents` had Typescript version 5.2.2 in its `package-lock.json` and the kitchen sink app we run headless tests on in the CI had Typescript version 5.8.2. 

Downgrading the Typescript version for the kitchen sink app cause the same issues 👍

infomiho (approved):
It fixes the issue 👍 

I'd still like us to create an issue for exploring how we want to limit the Typescript versions that are installed in users' projects.

PR #2938, opened 2025-07-10
Always build Wasp project before deployment

infomiho:
Our current logic assumed that if the `.wasp/build` directory exists - there is no need to run `wasp build`. This is not a really good assumption because the user might have changed the source code and they didn't rerun the `wasp build` command manually. This means that the deployment process will take the old build artefacts and deployment them!

I've added a change to always rebuild the Wasp project before deployment + made sure we only run `wasp build` once.

### DIFF
--- a/waspc/packages/deploy/src/common/waspBuild.ts
+++ b/waspc/packages/deploy/src/common/waspBuild.ts
@@ -1,0 +1,31 @@
+import { WaspCliExe, WaspProjectDir } from "./brandedTypes.js";
+import { waspSays } from "./terminal.js";
+import { createCommandWithCwd } from "./zx.js";
+
+export const ensureWaspProjectIsBuilt = createEnsureWaspProjectIsBuilt();
+
+function createEnsureWaspProjectIsBuilt() {
+  // We want to build the Wasp project only once per CLI invocation.
+  // Sometimes `ensureWaspProjectIsBuilt` is called multiple times
+  // (e.g. in `setup` command, we call it and then again when deploying).
+  let isWaspBuildExecuted = false;
+
+  async function ensureWaspProjectIsBuilt({
+    waspProjectDir,
+    waspExe,
+  }: {
+    waspProjectDir: WaspProjectDir;
+    waspExe: WaspCliExe;
+  }): Promise<void> {
+    if (isWaspBuildExecuted) {
+      return;
+    }
+
+    waspSays("Building your Wasp app...");
+    const waspCli = createCommandWithCwd(waspExe, waspProjectDir);
+    await waspCli(["build"]);
+    isWaspBuildExecuted = true;
+  }
+
+  return ensureWaspProjectIsBuilt;
+}
--- a/waspc/packages/deploy/src/common/waspProject.ts
+++ b/waspc/packages/deploy/src/common/waspProject.ts
@@ -3,7 +3,6 @@ import path from "node:path";
 
 import { cd } from "zx";
 import { WaspCliExe, WaspProjectDir } from "./brandedTypes.js";
-import { waspSays } from "./terminal.js";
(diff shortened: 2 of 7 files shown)


waspc/packages/deploy/src/common/waspBuild.ts
```diff
@@ -0,0 +1,31 @@
...
+import { waspSays } from "./terminal.js";
+import { createCommandWithCwd } from "./zx.js";
+
+export const ensureWaspProjectIsBuilt = createEnsureWaspProjectIsBuilt();
+
+function createEnsureWaspProjectIsBuilt() {
+  // We want to build the Wasp project only once per CLI invocation.
+  // Sometimes `ensureWaspProjectIsBuilt` is called multiple times
+  // (e.g. in `setup` command, we call it and then again when deploying).
+  let isWaspBuildExecuted = false;
```
infomiho: This code is old code with the addition of `isWaspBuildExecuted`

waspc/packages/deploy/src/common/waspBuild.ts
```diff
@@ -0,0 +1,31 @@
...
+  async function ensureWaspProjectIsBuilt({
+    waspProjectDir,
+    waspExe,
+  }: {
+    waspProjectDir: WaspProjectDir;
+    waspExe: WaspCliExe;
+  }): Promise<void> {
+    if (isWaspBuildExecuted) {
+      return;
+    }
```
cprecioso: This has a race condition:

```js
const taskA = async () => {
  await ensureWaspProjectIsBuilt()
  await dependsOnBuildA()
}

const taskB = async () => {
  await ensureWaspProjectIsBuilt()
  await dependsOnBuildB()
}

await Promise.all([taskA(), taskB()])
```

Both `taskA` and `taskB` will interfere build and interfere between each other, with one of the tasks possibly going ahead while the other task is in the midst of a re-build again.

Proposed solution: instead of storing a boolean, immediately store the actual result promise and always return that.

infomiho: Nice catch! Even though we always run commands sequentially in our codebase, it's better to be safe than sorry. I'll try implementing your solution and report back 👍

waspc/packages/deploy/src/common/waspBuild.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    waspSays("Building your Wasp app...");
+    const waspCli = createCommandWithCwd(waspExe, waspProjectDir);
+    waspBuildResult = waspCli(["build"]);
+    return waspBuildResult;
+  }
+
+  return ensureWaspProjectIsBuilt;
+}
```
cprecioso: Ok, this works! It's a bit weird following the logic, I'd do it like this, nicer separation between caching and logic:

```suggestion
function createEnsureWaspProjectIsBuilt() {
  async function ensureWaspProjectIsBuilt({
    waspProjectDir,
    waspExe,
  }: {
    waspProjectDir: WaspProjectDir;
    waspExe: WaspCliExe;
  }): Promise<void> {
    waspSays("Building your Wasp app...");
    const waspCli = createCommandWithCwd(waspExe, waspProjectDir);
    await waspCli(["build"]);
  }

  let cached: ReturnType<typeof ensureWaspProjectIsBuilt> | undefined = undefined

  // We want to build the Wasp project only once per CLI invocation.
  // Sometimes `ensureWaspProjectIsBuilt` is called multiple times
  // (e.g. in `setup` command, we call it and then again when deploying).
  return (...args: Parameters<typeof ensureWaspProjectIsBuilt>) => {
    cached ??= ensureWaspProjectIsBuilt(...args)
    return cached
  };
}
```

Not blocking on this as it's a style preference and we're short on time.

cprecioso (approved):

PR #2939, opened 2025-07-10
Convert `wasp-app-runner` to a `npm` package

### LINKED ISSUE #2840: Make e2e `wasp-app-runner` reusable

FranjoMindek:
Changes:
- Moves the current `wasp-app-runner` from `/` to `/packages` directory.
- Adds `npm run test` script to all examples.
- Converts all of the headless tests to use `@wasp.sh/wasp-app-runner` instead
- Updated `wasp-app-runner` readme, added publishing section
- Moved `@playwright/test` to dev deps
- Fixed `package.json` names of most examples
- Changed `waspc-ci.yaml` and `run` to reflect new changes

Discussible: 
- I've added `DEBUG=pw:webserver` to every `test` script. We might not want that.
- `/packages/wasp-app-runner` location
  - Personally I don't want it in `/waspc`, but I also find it weird in `/`. However, as we currently don't really have more packages to group together in `/packages`, it might be best to just leave it in `/`.

NOTE: Changes to `wasp-app-runner` logic are fixes we already merged to `wasp-lang/runner-action`. There is no need to look at them again. Only notable changes are `package.json` and `README.md`.

NPM link: https://www.npmjs.com/package/@wasp.sh/wasp-app-runner

Fixes #2840


.github/workflows/waspc-ci.yaml
```diff
@@ -145,15 +145,6 @@ jobs:
         if: matrix.os == 'ubuntu-22.04'
         run: cabal install
 
-      - name: Headless - Setup the Wasp app runner
```
FranjoMindek, line 148: Now installed during `todoApp`'s `npm ci`

packages/wasp-app-runner/README.md
```diff
@@ -5,24 +5,49 @@ A robust script to run your Wasp application, including automated database setup
...
-### Options
+### Local dependency:
+
+```bash
+npm i -D @wasp.sh/wasp-app-runner
+
+npx run-wasp-app dev
+```
+
+You can use it without the `npx` prefix inside of the `package.json`'s `scripts` field:
```
FranjoMindek: Is there a better way to explain this?
`npm` adds `node_module/.bin` to the `PATH` automatically.

packages/wasp-app-runner/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "type": "module",
+  "description": "CLI tool for running Wasp applications for headless testing",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/wasp-lang/wasp.git"
+  },
+  "homepage": "https://github.com/wasp-lang/wasp#readme",
+  "bugs": "https://github.com/wasp-lang/wasp/issues",
+  "author": "Wasp Team <info@wasp-lang.dev> (https://wasp.sh)",
+  "license": "MIT",
```
FranjoMindek: `npm init` stuff

sodic:
I vote for keeping the `wasp-app-runner` in the root unless there's a compelling reason to move it - it's simpler to both find and work with this way.

examples/tutorials/TodoApp/package.json
```diff
@@ -1,15 +1,19 @@
...
     "react-dom": "^18.2.0",
     "react-router-dom": "^6.26.2",
     "wasp": "file:.wasp/out/sdk/wasp"
   },
   "devDependencies": {
+    "@playwright/test": "1.51.1",
     "@types/react": "^18.0.37",
+    "@wasp.sh/wasp-app-runner": "^0.0.5",
     "prisma": "5.19.1",
     "typescript": "^5.1.0",
```
sodic: We'll have to do this again since TS has been updated on `main`.

examples/tutorials/TodoAppTs/package.json
```diff
@@ -1,15 +1,19 @@
...
+  },
   "dependencies": {
-    "@playwright/test": "1.51.1",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-router-dom": "^6.26.2",
     "wasp": "file:.wasp/out/sdk/wasp"
   },
   "devDependencies": {
+    "@playwright/test": "1.51.1",
```
sodic, line 14: Nice catch with the dev dependency!

.github/workflows/waspc-ci.yaml
sodic: I'm wondering whether the current approach (`wasp-app-runner` as a `devDependency`) is better than us explicitly installing the Wasp app runner when we need it.

On one hand, this is simpler. On the other hand, one could argue that `wasp-app-runner` isn't really a dev dependency of the project. It's just something an outside system might use to start the app. Also, if I had the app runner installed globally, I'd expect to use it on any of my apps (regardless of whether they list it in their `devDependencies` field).

@FranjoMindek any thoughts here?

FranjoMindek: One thing i like about `devDependency` approach is that it is much more transparent where we get it from.
e.g. consider onboarding future developers, this is much clearer, we don't even need to explain it 
I doubt we will have `wasp-app-runner` version issues, so I don't know if I would count that as an another point, but it's also worth considering.

Another alternative is to use `npx`, in that way you don't have to install it yourself beforehand, though the name gets butchered then -> `npx @wasp.sh/wasp-app-runner dev` instead of `npx run-wasp-app dev`.

Personally, I debated this before doing this PR, and I did opt in for `devDependency`.

>one could argue that `wasp-app-runner` isn't really a dev dependency of the project. It's just something an outside system might use to start the app.

I would look at it as part of the testing ecosystem, `wasp-app-runner` currently has only 1 purpose, and that is headless testing.

FranjoMindek:
I've moved `wasp-app-runner` back to root, and also included the change from https://github.com/wasp-lang/runner-action/pull/4#discussion_r2171184428

.github/workflows/waspc-ci.yaml
sodic: Good points! It stays a dev dep then :)

wasp-app-runner/README.md
```diff
@@ -9,20 +9,46 @@ When installed, it provides a `run-wasp-app` command that can be used to run you
...
+### Local dependency:
 
 ```bash
-run-wasp-app < mode > [--path-to-app < path > ] [--wasp-cli-cmd < command > ]
+npm i -D @wasp.sh/wasp-app-runner
+
+npx run-wasp-app dev
+```
+
+You can also use it in `npm` scripts without the `npx` prefix, since `node_modules/.bin` is automatically added to the PATH:
```
sodic:
```suggestion
You can also use it in `npm` scripts without the `npx` prefix, since `node_modules/.bin` is automatically added to the `PATH` inside `scripts`:
```

wasp-app-runner/src/args.ts
```diff
@@ -13,24 +17,21 @@ export function parseArgs(): {
...
-
-  program.parse();
-
   const options = parsedProgram.opts();
-  const args = parsedProgram.processedArgs;
+  const [mode] = parsedProgram.processedArgs;
 
   return {
-    mode: args[0],
+    mode: mode,
```
sodic:
```suggestion
    mode,
```

wasp-app-runner/tsconfig.json
```diff
@@ -12,7 +12,9 @@
     "noUncheckedIndexedAccess": true,
     "moduleResolution": "NodeNext",
     "module": "NodeNext",
+    "rootDir": "src",
     "outDir": "dist",
+    "declaration": true,
     "sourceMap": true,
```
sodic: Seems like we don't need these two. You can comment them out and explain why if you want.

sodic (approved):
It's basically ready!

Left some small suggestions, take care of them please and merge :)

Issue #2932, opened 2025-07-08
Explore a stricter `typescript` version in the users' projects


infomiho:
Related to https://github.com/wasp-lang/wasp/pull/2930#issuecomment-3049285401

Some of our users have version `5.2.2` in their `package-lock.json` and some have version `5.8.2`. The version of Typescript which we use in our headless tests is probably not the same as what our users have. This means that our tests are useless - they don't catch regression that might happen with a different Typescript version.

We should consider using a stricter range `~5.1.0` vs. `^5.1.0` or downright pin the version to `5.1.0` (I'm using this version as an example here).

@cprecioso would like to be able to use any version he wants as a user. I agree with that sentiment, but I'm not sure how to pull that off, that's something @sodic thought about more.

sodic:
We decided to go with a tilde (~) range, but that it was harder to support than expected.

Here's an issue: https://github.com/wasp-lang/wasp/issues/2941#issue-3219057067

sodic:
In the meantime, we set TypeScript to a fixed version (`5.8.2`) in #2940.

PR #2940, opened 2025-07-10
Set TypeScript version to 5.8.2. and update examples

sodic:
Bumps our TypeScript version to the latest one and fixes it for now (see #2932 for details).

We had to bump it to at least `5.4.0` because we use `module: preserve`. And, since we're already updating it, we decided to set it to the latest available version.


examples/waspello/package.json
```diff
@@ -16,8 +16,8 @@
     "@types/react": "^18.0.37",
     "prisma": "5.19.1",
     "tailwindcss": "^3.2.7",
-    "typescript": "^5.1.0",
+    "typescript": "5.8.2",
     "vite": "^4.3.9",
-    "wasp-config": "file:../../../../.local/share/wasp-lang/0.17.0-rc4/data/packages/wasp-config"
+    "wasp-config": "file:../../waspc/data/packages/wasp-config"
```
sodic, line 21: I gave up on keeping this tidy. They'll have to run `wasp ts-setup` anyway.

FranjoMindek: Yeah, till we improve this it's impossible

FranjoMindek:
I would also look into `generatePackageJson` of `waspc/src/Wasp/AI/GenerateNewProject/InitialFiles.hs` and starters `package.json` files.

cprecioso (approved):
Shallow review, LGTM. Do the snapshots need updating?

PR #2944, opened 2025-07-10
Add cache warm and cache evict workflows

### LINKED ISSUE #2919: Ensure warm cache for CI
We've seen CI runs of up to 30 mins due to evicted caches:
https://github.com/wasp-lang/wasp/actions/runs/15710311603?pr=2812

We should ensure the caches are warm so this doesn't happen

cprecioso:
Resolves #2919 

**Context:**
- Caches can be shared **from a base branch to its parent branch**. For example, a PR can pick up the cache for `main`. They can't be shared the opposite direction, nor between any other branches. [(docs)](https://docs.github.com/en/actions/reference/dependency-caching-reference#restrictions-for-accessing-a-cache)
- Caches are evicted if either: (a) haven't been accessed in 7 days, or (b) the repo goes over the 10GB limit for caches (LRU evicted first). [(docs)](https://docs.github.com/en/actions/reference/dependency-caching-reference#usage-limits-and-eviction-policy)

**Therefore:**
- We want to ensure the cache for our `main` branch is never stale and never at risk of being evicted. An evicted `main` cache means un-cached runs for all the PRs.
- Caches for branches that have been merged are **immediately** no longer useful, so we can remove them.
- This lowers the risk of evicting other actually important caches (such as `main`'s)

**This PR:**
Adds two new workflows: 
1. Runs every 6 days (so we don't go over the 7 days limit) and makes sure that our `main` caches are warm, by running `waspc-ci` and `waspc-build`
2. Every time a PR is merged, goes through the caches and evicts the ones linked to no-longer-existing branches.


.github/workflows/deploy-examples.yml
```diff
@@ -25,6 +25,9 @@ concurrency:
   group: "deploy-examples"
   cancel-in-progress: true
 
+env:
+  WASP_TELEMETRY_DISABLE: 1
```
infomiho, line 29: Nice catch!

.github/workflows/cache-warm.yml
```diff
@@ -0,0 +1,18 @@
...
+concurrency:
+  group: "cache-warm"
+  cancel-in-progress: false
+
+jobs:
+  warm-cache-build:
+    uses: ./.github/workflows/waspc-build.yaml
+
+  warm-cache-ci:
+    uses: ./.github/workflows/waspc-ci.yaml
```
infomiho, line 18: What about this schedule in the `wasp-ci` workflow: https://github.com/wasp-lang/wasp/blob/main/.github/workflows/waspc-ci.yaml#L18

Do we want to remove it now?

.github/workflows/cache-evict.yml
```diff
@@ -0,0 +1,57 @@
...
+        # This script lists all caches for the repository, lists all remote refs
+        # for the repository, and then finds the caches that are not present in
+        # the remote refs. It then deletes those caches.
+        #
+        # For that, it uses the `gh` CLI to list and delete caches, and `git` to
+        # list remote refs. It also uses `jq` to manipulate the data as JSON.
+        # (It's surprisingly difficult to find a shell command to check if a
+        # string is inside a given array, that's why we use `jq`.)
+        run: |
+          (
```
infomiho: It would be really useful to have some example data displayed for each stage. I've gone and ran the commands locally to better understand what's going on.

cprecioso: mmm i know, it's kinda messy... maybe i can write a small vanilla `.js` script so the logic is easier to follow

.github/workflows/cache-evict.yml
```diff
@@ -0,0 +1,57 @@
...
+            gh cache list --limit 100 --json key,ref
+
+            # Input 1: List all remote refs for the repository, do some string
+            # manipulation to get the ref names (second column) as a JSON array
+            git ls-remote "https://github.com/$GITHUB_REPOSITORY.git" |
+              awk '{print $2}' |
+              jq --raw-input --compact-output '.' |
+              jq --slurp '.'
+          ) |
+            # Find cache refs not present in the remote refs, and extract their keys
```
infomiho: Now that I look at complexity here, I'd might suggest extracting this into a bash script with named functions to make it more readable. It can be super simple, but names would really help.

Here's a GPTed script as an idea:
```shell
#!/bin/bash

# Delete GitHub Actions caches for branches that no longer exist

set -euo pipefail

get_cache_data() {
    gh cache list --limit 100 --json key,ref
}

get_remote_refs() {
    git ls-remote "https://github.com/$GITHUB_REPOSITORY.git" | \
        awk '{print $2}' | \
        jq --raw-input --compact-output '.' | \
        jq --slurp '.'
}

find_stale_cache_keys() {
    local cache_data="$1"
    local remote_refs="$2"
    
    echo "$cache_data" "$remote_refs" | \
        jq --slurp --raw-output \
            '.[0] as $caches | .[1] as $remote_refs | 
             $caches[] | 
             select([.ref] | inside($remote_refs) | not) | 
             .key'
}

delete_cache_key() {
    local key="$1"
    gh cache delete "$key" || echo "::warning::Failed to delete cache key $key"
}

main() {
    local cache_data remote_refs stale_keys
    cache_data=$(get_cache_data)
    remote_refs=$(get_remote_refs)
    stale_keys=$(find_stale_cache_keys "$cache_data" "$remote_refs")
    
    if [[ -n "$stale_keys" ]]; then
        echo "$stale_keys" | while read -r key; do
            [[ -n "$key" ]] && delete_cache_key "$key"
        done
    fi
}

main "$@"
```

infomiho (commented):
Left some comments!

.github/workflows/cache-warm.yml
cprecioso: Yes!

cprecioso:
I felt the shell script inside the yaml was way too obtuse, so I moved it to a node.js script. (Made it with no dependencies and using the preinstalled Node version in `ubuntu-latest` so that it doesn't need to install node or run npm)

I made a dry-run at https://github.com/wasp-lang/wasp/actions/runs/16296955021/job/46021046403?pr=2944

@infomiho recheck pls

.github/workflows/cache-evict.yml
cprecioso: I re-made everything as js script, I don't think the shell language is very explicate wrt what data we have and in which shape, no matter if we organize in different functions.

scripts/cache-evict.mjs
```diff
@@ -0,0 +1,105 @@
...
+
+const { repository } = assertGithubActionsEnv();
+deleteOrphanedCaches(repository);
+
+function deleteOrphanedCaches(/** @type {string} */ githubRepository) {
+  const ghCachesOutput = listGitHubCaches();
+  const lsRemoteOutput = listRemoteRefs(githubRepository);
+
+  // Find refs that have caches but do not exist in the remote repository anymore.
+  const remoteRefs = new Set(lsRemoteOutput.map(([, ref]) => ref));
```
infomiho: Why not do the mapping in the `listRemoteRefs`?

scripts/cache-evict.mjs
```diff
@@ -0,0 +1,105 @@
...
+  const cachesToDelete = ghCachesOutput.filter(
+    (cache) => !remoteRefs.has(cache.ref),
+  );
+  console.log(`Found ${cachesToDelete.length} caches to delete`);
+
+  // Delete those cache keys
```
infomiho: Redundant comment

scripts/cache-evict.mjs
```diff
@@ -0,0 +1,105 @@
...
+import { execFileSync } from "node:child_process";
+
+const { repository } = assertGithubActionsEnv();
+deleteOrphanedCaches(repository);
+
+function deleteOrphanedCaches(/** @type {string} */ githubRepository) {
+  const ghCachesOutput = listGitHubCaches();
+  const lsRemoteOutput = listRemoteRefs(githubRepository);
+
+  // Find refs that have caches but do not exist in the remote repository anymore.
```
infomiho: This comment is probably above the wrong line, and maybe it's even redundant since it's clear from the code that you are finding caches that are not in the remote refs.

scripts/cache-evict.mjs
```diff
@@ -0,0 +1,105 @@
...
+  const cachesToDelete = ghCachesOutput.filter(
+    (cache) => !remoteRefs.has(cache.ref),
+  );
+  console.log(`Found ${cachesToDelete.length} caches to delete`);
+
+  // Delete those cache keys
+  for (const { key, ref } of cachesToDelete) {
+    try {
+      console.group(`Deleting cache "${key}" for ref "${ref}"`);
+      runCmd("gh", ["cache", "delete", key], { collectStdout: false });
```
infomiho: I'd maybe even extract this into a function e.g. `deleteGithubCache` to make it cleaner.

scripts/cache-evict.mjs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+function listRemoteRefs(githubRepository) {
+  const lsRemoteOutput = /** @type {[objectId: string, ref: string][]} */ (
+    parseTabSeparatedLines(
+      runCmd(
+        // We use `git ls-remote` so we also receive refs such as `refs/pull/123/head`
+        // which are not downloaded by `git fetch` or `git pull`.
+        //
+        // According to https://git-scm.com/docs/git-ls-remote, the output is:
```
infomiho: This format comment could go into `parseTabSeparatedLines` function if we rename the function to handle the remote refs format specifically. See the comment: https://github.com/wasp-lang/wasp/pull/2944/files#r2212842477

scripts/cache-evict.mjs
```diff
@@ -0,0 +1,105 @@
...
+        ["cache", "list", "--limit", "100", "--json", "key,ref"],
+      ),
+    )
+  );
+  console.log(`Found ${ghCachesOutput.length} cache keys`);
+  return ghCachesOutput;
+}
+
+function listRemoteRefs(githubRepository) {
+  const lsRemoteOutput = /** @type {[objectId: string, ref: string][]} */ (
```
infomiho: Technically this is not the `lsRemoteOutput` but the parsed output, maybe a better name would be `remoteRefs` since that's what you function says it lists.

scripts/cache-evict.mjs
```diff
@@ -0,0 +1,105 @@
...
+) {
+  return execFileSync(cmd, args, {
+    encoding: "utf-8",
+    stdio:
+      // stdin, stdout, stderr
+      ["ignore", collectStdout ? "pipe" : "inherit", "inherit"],
+  });
+}
+
+function parseTabSeparatedLines(/** @type {string} */ str) {
```
infomiho: I'd be specific here since the script is quite specific, name this helper `parse git remotes output` and put the explainer comment on top.

You can even make the output not be `[str1, str2]` but provide names for the values you already have names for in the format `<oid> TAB <ref> LF` -> `{ oid, ref }`. I'm not sure what `oid` is in this case, maybe the name can reflect that.

cprecioso:
@infomiho ready

PR #2947, opened 2025-07-11
Basic starter template fixes

infomiho:
- Adds Wasp app name to `package.json`.
- Fixes design when there are no tasks.
- Updates sort to be by `createdAt`


cprecioso (approved):
LGTM

Issue #2824, opened 2025-06-09
Automatically restart Algolia crawler on new docs releases


sodic:
Whenever we publish a new version of our docs, we need to manually log into the Algolia Crawler and update its search index. The search doesn't work properly until we do. This is annoying and easy to forget, so we should automate it.

Investigate whether we can have Algolia automatically scrape our docs whenever we do a release (either with a hook or an explicit API call).

PR #2948, opened 2025-07-11
Start Algolia crawler automatically on web deploy

FranjoMindek:
Fixes: https://github.com/wasp-lang/wasp/issues/2824

### DIFF
--- a/.github/workflows/restart-algolia-crawler.yaml
+++ b/.github/workflows/restart-algolia-crawler.yaml
@@ -1,0 +1,21 @@
+name: Restart Algolia Crawler
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - deploy-web
+
+jobs:
+  restart-crawler:
+    name: Restart Algolia Crawler
+    runs-on: ubuntu-latest
+    steps:
+      - name: Restart Algolia Crawler
+        uses: algolia/algoliasearch-crawler-github-actions@v1
+        id: crawler_push
+        with:
+          crawler-user-id: ${{ secrets.ALGOLIA_CRAWLER_USER_ID }}
+          crawler-api-key: ${{ secrets.ALGOLIA_CRAWLER_API_KEY }}
+          algolia-app-id: ${{ secrets.ALGOLIA_APP_ID }}
+          crawler-name: wasp-lang
--- a/.github/workflows/restart-algolia-crawler.yaml
+++ b/.github/workflows/restart-algolia-crawler.yaml
@@ -18,4 +18,5 @@ jobs:
           crawler-user-id: ${{ secrets.ALGOLIA_CRAWLER_USER_ID }}
           crawler-api-key: ${{ secrets.ALGOLIA_CRAWLER_API_KEY }}
           algolia-app-id: ${{ secrets.ALGOLIA_APP_ID }}
-          crawler-name: wasp-lang
+          algolia-api-key: ${{ secrets.ALGOLIA_API_KEY }}
+          site-url: 'https://wasp.sh'


.github/workflows/restart-algolia-crawler.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - deploy-web
+
+jobs:
+  restart-crawler:
+    name: Restart Algolia Crawler
+    runs-on: ubuntu-latest
+    steps:
+      - name: Restart Algolia Crawler
+        run: |
+          algolia_authentication_token=$(echo -n "${{ secrets.ALGOLIA_CRAWLER_USER_ID }}:${{ secrets.ALGOLIA_CRAWLER_API_KEY }}" | base64)
```
infomiho: I'd add a link to the Algolia API reference in the comment above: https://www.algolia.com/doc/rest-api/crawler/#tag/actions/operation/startReindex

infomiho:
There is one thing related to timing here, we start publishing the docs when `deploy-web` gets merged, but that doesn't mean that the docs are published yet.

Some minutes pass before the _latest docs_ are live so if we start crawling _right away_ - the crawler will still see the old version of docs. 

We should either delay this somehow or find a better trigger e.g. run a [deploy hook](https://developers.cloudflare.com/pages/configuration/deploy-hooks/) on Cloudflare when the deployment finishes?

FranjoMindek:
> There is one thing related to timing here, we start publishing the docs when `deploy-web` gets merged, but that doesn't mean that the docs are published yet.

Good point 

> We should either delay this somehow or find a better trigger e.g. run a [deploy hook](https://developers.cloudflare.com/pages/configuration/deploy-hooks/) on Cloudflare when the deployment finishes?

We still would need to wait for the `deploy hook` to finish.
Can we get such information? We would most likely pool for the build to finish.
In that case we can also pool here without using deploy hook?
Ideally Cloudflare would have on-deploy webhook, but I don't see that.

Edit: https://github.com/WalshyDev/cf-pages-await seems suitable

.github/workflows/restart-algolia-crawler.yaml
```diff
@@ -0,0 +1,51 @@
+name: Restart Algolia crawler
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - deploy-web
```
cprecioso, line 8: What's this branch?

.github/workflows/restart-algolia-crawler.yaml
```diff
@@ -0,0 +1,51 @@
...
+          done
+
+      - name: Restart Algolia crawler
+        run: |
+          # Based on Algolia API: https://www.algolia.com/doc/rest-api/crawler/#tag/actions/operation/startReindex
+          algolia_authentication_token=$(echo -n "${{ secrets.ALGOLIA_CRAWLER_USER_ID }}:${{ secrets.ALGOLIA_CRAWLER_API_KEY }}" | base64 -w 0)
+          curl -X POST \
+            -H "Authorization: Basic $algolia_authentication_token" \
+            -H "Content-Type: application/json" \
+            -H "User-Agent: algolia_crawler_github_actions/1.0.0" \
```
cprecioso: Do we need this? Is it an internal value? Are we misrepresenting ourselves?
Maybe we can use the official Algolia action? https://github.com/algolia/algoliasearch-crawler-github-actions

.github/workflows/restart-algolia-crawler.yaml
```diff
@@ -0,0 +1,51 @@
...
+              exit 1
+            fi
+          done
+
+      - name: Restart Algolia crawler
+        run: |
+          # Based on Algolia API: https://www.algolia.com/doc/rest-api/crawler/#tag/actions/operation/startReindex
+          algolia_authentication_token=$(echo -n "${{ secrets.ALGOLIA_CRAWLER_USER_ID }}:${{ secrets.ALGOLIA_CRAWLER_API_KEY }}" | base64 -w 0)
+          curl -X POST \
+            -H "Authorization: Basic $algolia_authentication_token" \
```
cprecioso:
```suggestion
          curl -X POST \
            --user "${{ secrets.ALGOLIA_CRAWLER_USER_ID }}:${{ secrets.ALGOLIA_CRAWLER_API_KEY }}" \
```

.github/workflows/restart-algolia-crawler.yaml
```diff
@@ -0,0 +1,51 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Wait for Wasp web to finish deploying
+        env:
+          CF_API_TOKEN: ${{ secrets.CF_API_TOKEN }}
+        run: |
+          set -e
+
+          while true; do
+            response=$(curl -s -X GET "https://api.cloudflare.com/client/v4/accounts/75c37602a6b75c790961219411cb1b92/pages/projects/wasp-docs/deployments?per_page=1&page=1&env=production" \
```
cprecioso: Suggestion: you can tag this job with [an environment](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idenvironment), add the account id to the environment variables, and use that variable here

cprecioso:
```suggestion
            response=$(curl -s "https://api.cloudflare.com/client/v4/accounts/75c37602a6b75c790961219411cb1b92/pages/projects/wasp-docs/deployments?per_page=1&page=1&env=production" \
```

.github/workflows/restart-algolia-crawler.yaml
FranjoMindek: Some background I forgot to mention:

Algolia action requires some additional permission which we don't have rights to (`addObject`).
This is because that action can also create new crawlers if there is no existing one.
That requires admin API key, which we don't have access to in Algolia (we are actually not the owners of the wasp team/organization on Algolia, some Algolia account is).
We've tried with the API write key, but it lacked `addObject` permission.

Because of the we call the REST API directly for restarting crawlers.

As for `User-Agent` I can leave it out. We stole that from their action.

.github/workflows/restart-algolia-crawler.yaml
FranjoMindek: This is branch that we track for our production web deployment on Cloudflare pages.
Do we want to add comments here?

.github/workflows/restart-algolia-crawler.yaml
FranjoMindek: But we need in base64 as per docs:
https://www.algolia.com/doc/rest-api/crawler/#section/Availability-and-authentication

FranjoMindek: Oh, thats a good one, didn't know about it.

.github/workflows/restart-algolia-crawler.yaml
FranjoMindek: Removed `User-Agent`

.github/workflows/restart-algolia-crawler.yaml
cprecioso: This is not what I was referring to 😅 Check the `environment` key, not the `env` one.
Feel free to drop it though if you see it'd be much of a hassle.

FranjoMindek: Aha, though `CLOUDFLARE_ACCOUNT_ID` is the same for all the environments, others could be parameterized per environment:
- `CLOUDFLARE_PROJECT_NAME`
- `ALGOLIA_CRAWLER_ID`
- `ALGOLIA_CRAWLER_USER_ID`
- `ALGOLIA_CRAWLER_API_KEY`

But I don't see too much need.
We only ever want to crawl production, so there is no real need for that?
Unless we wanted to reduce pollution of global repo secrets.

I hope I got right this time.
One another issue is I don't have access to secrets/envs/vars, so I would need somebody else to set them too.

cprecioso: Okay, the feel free to drop it, thanks for looking into it!

cprecioso:
Looks fine by me, will let @infomiho approve

.github/workflows/restart-algolia-crawler.yaml
infomiho: I'd maybe link to the `web` README.md https://github.com/wasp-lang/wasp/tree/main/web#deployment

infomiho (approved):
Good to go

PR #2949, opened 2025-07-11
Add support for Prisma triple-slash comments

### LINKED ISSUE #2132: Support parsing triple slash comments in `schema.prisma`
While reading [some blogs](https://medium.com/@kz-d/multi-tenancy-with-prisma-a-new-approach-to-making-where-required-1e93a3783d9d) about Prisma I found out that there is a special type of comment in Prisma:

```prisma
/// that looks like this
```

As [explained](https://www.prisma.io/docs/orm/prisma-schema/overview#comments) in the Prisma docs:
> /// comment: These comments will show up in the abstract syntax tree (AST) of the schema as descriptions to AST nodes. Tools can then use these comments to provide additional information. All comments are attached to the next available node - [free-floating comments](https://github.com/prisma/prisma/issues/3544) are not supported and are not included in the AST.

These comment need to _survive parsing_ and should be in our AST so we can also output them in the final Prisma schema. 

The triple slash comments can be:
- above a block
- one a standalone line in the block body
- trailing after some block line

cprecioso:
- Resolves #2132
- On top of #2985 

Adds support for triple-slash comments in the schemas, adding them to the AST:
- We create `data WithCtx node` that wraps an existing node with a `NodeContext` (which only holds comments for now)
- We create a higher-order parser that handles the `WithCtx` business
This solution design is based on @Martinsos's tips at https://discord.com/channels/686873244791210014/776460802131099680/1394365158159290569

The main important changes are in `waspc/src/Wasp/Psl`, all the other changes are mostly either wrapping or unwrapping AST structures to match the changes.


waspc/e2e-test/test-outputs/waspComplexTest-golden/waspComplexTest/.wasp/out/db/schema.prisma
cprecioso, line 1: I am not sure why this changed format :/

infomiho:
Reviewed this (but didn't comment anything yet), I'll go through it once more :D looking good, but I'd like to test it with some plugin using the triple slash syntax e.g. https://github.com/CarterGrimmeisen/zod-prisma

cprecioso:
@infomiho hold off, i found some errors when trying it out with that library

waspc/src/Wasp/Psl/Parser/WithCtx.hs
```diff
@@ -0,0 +1,35 @@
...
+  )
+import Wasp.Psl.Parser.Common (lexeme)
+
+withCtx :: Parser node -> Parser (WithCtx node)
+withCtx nodeParser = do
+  leadingComments <- many (try documentationComment)
+  node <- nodeParser
+  -- TODO: Handle trailing comments properly, right now we fail on them.
+
+  return $ WithCtx node (NodeContext leadingComments)
```
cprecioso: No matter how I tried, I couldn't get the parser to both get leading and trailing comments.

This is the closest I got:

```hs
withCtx :: Parser node -> Parser (WithCtx node)
withCtx nodeParser = do
  leadingComments <- many (try documentationComment)

  (node, trailingComment) <-
    choice
      [ -- When trying for a trailing comment, we forcefully bind the parser to
        -- a single line, as these comments must appear on the same line as the
        -- node.
        singleLine $ liftTuple (nodeParser, Just <$> documentationComment),
        -- Otherwise, we parse the node without a trailing comment.
        liftTuple (nodeParser, pure Nothing)
      ]

  let allComments = leadingComments ++ maybeToList trailingComment
  return $ WithCtx node (NodeContext allComments)
  where
    singleLine = between empty (char '\n')
    liftTuple (a, b) = liftA2 (,) a b
```

but it doesn't end up working, and I'm at my wit's end on how to make it work.

AFAIU, this is an intrinsically difficult task for Parsec, as every token, by design consumes and ignores any amount of whitespace after it, making it really hard/impossible for the doc-comment parser to find out if there was a newline between the end of the inner parser or not, which is the difference between a trailing comment on the current node, or a leading comment on the next one.

cprecioso:
@infomiho I dropped the trailing comment support, I don't know how to do it effectively (more details at https://github.com/wasp-lang/wasp/pull/2949#discussion_r2237409344). You can review now.
If this shortcoming is accepted, I'll create an issue for it.

waspc/e2e-test/test-outputs/waspComplexTest-golden/waspComplexTest/.wasp/out/db/schema.prisma
Martinsos, line 1: How is it that formatting changed here? Are we ok with it?

waspc/src/Wasp/Psl/Ast/WithCtx.hs
```diff
@@ -0,0 +1,50 @@
...
+    DocumentationComment,
+    DocumentationComments,
+    commentedNode,
+    getNode,
+  )
+where
+
+import Data.Data (Data)
+
+-- | A PSL node that defines an output in the generated code. As such, it might
```
Martinsos: This comment is outdated: it still talks about "output in the generated code".
Btw, maybe you don't even need any comment, maybe it is clear enough without? Not sure.

waspc/src/Wasp/Psl/Ast/WithCtx.hs
```diff
@@ -0,0 +1,50 @@
...
+  deriving (Show, Eq, Data)
+
+data NodeContext = NodeContext
+  { documentationComments :: DocumentationComments
+  }
+  deriving (Show, Eq, Data)
+
+type DocumentationComments = [DocumentationComment]
+
+type DocumentationComment = String
```
Martinsos: Is this a Prisma concept: "documentation comment"? I wonder if this should be defined outside of this file then maybe?

waspc/src/Wasp/Psl/Ast/WithCtx.hs
```diff
@@ -0,0 +1,50 @@
...
+
+commentedNode :: [String] -> node -> WithCtx node
+commentedNode comments node = WithCtx node (NodeContext comments)
```
Martinsos: You are using positional arguments here for the `NodeContext` record -> that is anti pattern. Imagine that somebody adds new field to `data NodeContext`, called `tags`, which is list of strings, and puts it as the first field in the record. You might easily end up passing comments under tags here.

So better to go with `NodeContext { documentationComments = comments }`.

waspc/src/Wasp/Psl/Ast/WithCtx.hs
```diff
@@ -0,0 +1,50 @@
...
+data NodeContext = NodeContext
+  { documentationComments :: DocumentationComments
+  }
+  deriving (Show, Eq, Data)
+
+type DocumentationComments = [DocumentationComment]
+
+type DocumentationComment = String
+
+commentedNode :: [String] -> node -> WithCtx node
```
Martinsos, line 24:
```suggestion
commentedNode :: [String] -> node -> WithCtx node
```

Should `String` here be `String` indeed, or should it be DocumentaionComment? To help with answering: if implemnetation of DocumentaionComment changes to Text, would you change this to Text or leave it as String?

waspc/src/Wasp/Psl/Ast/WithCtx.hs
```diff
@@ -0,0 +1,50 @@
...
+  pure node = WithCtx node (NodeContext [])
+  WithCtx f context <*> WithCtx node context' =
+    WithCtx (f node) (context <> context')
+
+instance Semigroup NodeContext where
+  (NodeContext comments1) <> (NodeContext comments2) =
+    NodeContext (comments1 <> comments2)
+
+instance Monoid NodeContext where
+  mempty = NodeContext []
```
Martinsos: Ha nice stuff :D!

waspc/src/Wasp/Generator/DbGenerator.hs
```diff
@@ -110,7 +110,7 @@ genPrismaSchema spec = do
     entityToPslModelSchema :: (String, AS.Entity.Entity) -> String
     entityToPslModelSchema (entityName, entity) =
       Psl.Generator.Schema.generateSchemaBlock $
-        Psl.Schema.ModelBlock $ Psl.Model.Model entityName (AS.Entity.getPslModelBody entity)
+        Psl.Schema.ModelBlock $ pure $ Psl.Model.Model entityName (AS.Entity.getPslModelBody entity)
```
Martinsos: One consideration: imagine that in the future we add another field to `NodeContext`, so it has something more than just documentationComments.

What if you do want to set that extra field to something here, and not just leave it empty (as is happening now with `pure`)? You won't know you should look into setting it, because compiler won't let you know that there is usage here that should be updated, since it will be happy producing an empty one for you.

So what you might want to do is explicitly call `WithCtx node { documentationComments = [] }` here, to have compiler let you know to update this part if NodeContext is modified.

Unless you are 100% sure you will never want to set anything into node context here, then `pure` is great.

Same goes for other places where you used `pure`.

waspc/src/Wasp/Psl/Generator/WithCtx.hs
```diff
@@ -0,0 +1,16 @@
...
+import Wasp.Psl.Ast.WithCtx (NodeContext (NodeContext), WithCtx (WithCtx))
+import Wasp.Psl.Generator.Common (PslSource)
+
+generateWithCtx :: (node -> PslSource) -> WithCtx node -> PslSource
+generateWithCtx generateNode (WithCtx node context) =
+  generateNodeContext context (generateNode node)
+
+generateNodeContext :: NodeContext -> PslSource -> PslSource
+generateNodeContext (NodeContext comments) content =
+  unlines (map ("///" ++) comments ++ [content])
```
Martinsos: I like `generateWithCtx` as a name, and signature also makes sense to me.

`generateNodeContext` confuses me though. Its name says it generates node context, but it does more, it also takes already generated node and adds that into the game. I guess you might need this, taking generated node as a parameter, because you might have some comments to print aftet it, not only before, but than function can be named `generateNodeWithCtx`. That sounds awfully a lot like the function above though, `generateWithCtx`. And the question was all the time, why do you really  need this extra function? It could easily be done with a `where` in `generateWithCtx`. So I would remove this second function, `generateNodeContext` and just implemented this logic directly in the `generateWithCtx` while putting helper logic in `where`.

waspc/src/Wasp/Psl/Generator/WithCtx.hs
```diff
@@ -0,0 +1,16 @@
...
+
+import Wasp.Psl.Ast.WithCtx (NodeContext (NodeContext), WithCtx (WithCtx))
+import Wasp.Psl.Generator.Common (PslSource)
+
+generateWithCtx :: (node -> PslSource) -> WithCtx node -> PslSource
+generateWithCtx generateNode (WithCtx node context) =
+  generateNodeContext context (generateNode node)
+
+generateNodeContext :: NodeContext -> PslSource -> PslSource
+generateNodeContext (NodeContext comments) content =
```
Martinsos: Same as I commented above, unwrapping NodeContext positionaly here is anti-pattern, it is better to unwrap it as `NodeContext { documentationComments = comments }` (or whatever is correct syntax I didn't check).

waspc/src/Wasp/Psl/Generator/WithCtx.hs
Martinsos: This `///` here -> this must be duplication, because somewhere in the parser you must be defining the same special character sequence.

This is not a huge thing, since it probably won't change often, but still, I woudl consider removing this duplication.

Martinsos: Just an alternative idea, take what you like (that first $ is clear win I think).
```suggestion
  unlines $ (("///" ++) <$> comments) ++ [content]
```

waspc/test/AppSpec/ValidTest.hs
```diff
@@ -279,18 +279,19 @@ spec_AppSpecValid = do
...
-                            Psl.Model._type = Psl.Model.String,
-                            Psl.Model._typeModifiers = [],
-                            Psl.Model._attrs =
-                              [ Psl.Attribute.Attribute
-                                  { Psl.Attribute._attrName = "id",
-                                    Psl.Attribute._attrArgs = []
-                                  }
-                              ]
-                          }
+                    [ pure $
```
Martinsos: Another thought here: it can be hard to read code when one sees `pure` here but has no idea what it stands for. Maybe having a function `WithCtx.empty` would be nice, and implementation of `pure` can merely call it. Then you can choose if you want to use `pure` or `WithCtx.empty` in different situations. And here I would probably go with `WithCtx.empty` then.

Some Haskell libraries do this similar thing, e.g. Map (or HashMap or what is it).

waspc/test/Psl/Common/ModelTest.hs
```diff
@@ -24,127 +24,134 @@ sampleBodySchema =
 sampleBodyAst :: Psl.Model.Body
 sampleBodyAst =
   Psl.Model.Body
-    [ Psl.Model.ElementField
```
Martinsos, line 27: Maybe you could have just fmaped `pure` on the whole list? But ok, this is also ok.

waspc/test/Psl/Generator/SchemaTest.hs
```diff
@@ -46,9 +46,9 @@ instance Arbitrary Psl.Type.Type where
 
 instance Arbitrary Psl.Model.Body where
   arbitrary = do
-    fieldElement <- Psl.Model.ElementField <$> arbitrary
-    elementsBefore <- scale (const 5) arbitrary
-    elementsAfter <- scale (const 5) arbitrary
+    fieldElement <- pure . Psl.Model.ElementField <$> arbitrary
+    elementsBefore <- map pure <$> scale (const 5) arbitrary
```
Martinsos: What I love about Haskell is that I don't need to understand what is really happening here, I just understand that some kind of application is happening and that compiler took care of it being valid :D.

waspc/test/Psl/Generator/WithCtxTest.hs
```diff
@@ -0,0 +1,134 @@
...
+import qualified Wasp.Psl.Ast.Attribute as Psl.Attribute
+import qualified Wasp.Psl.Ast.Model as Psl.Model
+import qualified Wasp.Psl.Ast.Schema as Psl.Schema
+import Wasp.Psl.Ast.WithCtx (commentedNode)
+import qualified Wasp.Psl.Generator.Schema as Psl.Generator
+
+spec_generatePslNode :: Spec
+spec_generatePslNode = do
+  describe "Triple slash documentation comments" $ do
+    it "Prisma file is correctly generated" $ do
```
Martinsos:
```suggestion
    it "Prisma model with leading triple slash comments is correctly generated" $ do
```

waspc/test/Psl/Generator/WithCtxTest.hs
```diff
@@ -0,0 +1,134 @@
...
+                model MyModel {
+                  /// Simple comment attached to `prop1`
+                  prop1 Int @id @default(autoincrement())
+
+                  /// Simple comment attached to `prop2`
+                  prop2 String @unique
+
+                  /// Multiline leading comments
+                  /// For prop3
+                  /// And also trailing!
```
Martinsos: I believe this is left remaining after you removed the trailing comments support.

waspc/test/Psl/Generator/WithCtxTest.hs
```diff
@@ -0,0 +1,134 @@
...
+                            " And also trailing!"
+                          ]
+                          $ Psl.Model.ElementField $
+                            Psl.Model.Field
+                              "prop3"
+                              Psl.Model.String
+                              []
+                              [Psl.Attribute.Attribute "unique" []]
+                      ]
+                  )
```
Martinsos: I wish this could be more readable! So it is easier to maintain.

One way would be to not define AST here, but to parse actual string. This changes the test somewhat though -> now it also involves parsing, so it is more of an integration test and less of a unit test. But you get easier way to define a test. Hm maybe best to leave it as it is, but it is tempting.

Another way would be to define some kind of helper functions that make this piece of code more readable, by putting emphpasis on important stuff. Helper functions for constructing Prisma PSL. E.g. imagine
```hs
commentedNode [" Simple comment attached to `prop2`"] $
  Psl.Model.ElementField $
    Psl.Model.Field
      "prop2"
      Psl.Model.String
      []
      [Psl.Attribute.Attribute "unique" []],
```

being
```hs
withComments [" Simple comment attached to `prop2`"] $
  stringField "prop2" [] [attr "unique"]
```

This might not be as simple in practice as I imagine, due to naming conflicts potentially and similar, but might be worth trying.

This whole comment is suggestion: you don't have to do it, but it might be worth playing with the idea if it makes tests easier to maintain.

waspc/test/Psl/Generator/SchemaTest.hs
Martinsos, line 1: You went with pure everywhere here, but this is property testing with Quickcheck -> don't we want to sometimes generate a node context with something in it?

waspc/src/Wasp/Psl/Ast/WithCtx.hs
Martinsos, line 24: This will work as long as WithCtx has nothing but comments, but once it does, you will have to refactor all the places where you called this.

I wonder if you might rather want to have function `withComments :: WithCtx node -> [String] -> WithCtx node` that sets (or adds?) comments to the context? Combine this with fuction trhat creates empty context and than this adds comments adn you hvae the same effect. Or maybe you don't need this at all, and can just set the field directly? Not sure hm, I am ideating here somewhat.

Feel free to not follow up on this suggestion if you think it is not worth the trouble.

raw

waspc/test/Psl/Generator/WithCtxTest.hs
```diff
@@ -0,0 +1,134 @@
...
+                  /// Multiline leading comments
+                  /// For prop3
+                  /// And also trailing!
+                  prop3 String @unique
+                }
+              |]
+
+      Psl.Generator.generateSchemaBlock inputAst `shouldBe_ignoringWhitespace` expectedPrismaSchema
+
+    it "Prisma-zod example" $ do
```
Martinsos: How is this test example special? Why do we have it compared to the test above? I think it is worth making that clear.

waspc/test/Psl/Generator/WithCtxTest.hs
```diff
@@ -0,0 +1,134 @@
...
+                }
+              |]
+
+      Psl.Generator.generateSchemaBlock inputAst `shouldBe_ignoringWhitespace` expectedPrismaSchema
+
+shouldBe_ignoringWhitespace :: String -> String -> Expectation
+shouldBe_ignoringWhitespace = shouldBe `on` removeWhitespace
+
+-- | Removes leading indentation and empty lines from the string.
+removeWhitespace :: String -> String
```
Martinsos: Name removeWhitespace is misleading somewhat. Which is why you also need that comment.
You could just give it a longer name that says what comment does. 
Or you could name it `normalizeWhitespace`.
Or you could split it into multiple functions like `removeEmptyLines` and `removeLeadingIndentation` and combine those together in `shouldBe_...`.

I would probably go with `normalizeWhitespace`.

waspc/src/Wasp/Psl/Parser/WithCtx.hs
Martinsos: Same as in other places, make sure to use names of fields when constructing a record like this, not positional args.

Martinsos: Interesting!

Ok yeah seems Parsec is somewhat limited here. I get the problem -> `nodeParser` is a lexeme parser, consuming whitespace after it, and tehre is no way in Parsec to customize that.

What if we looked for the first newline, and then tried parsing text before it with nodeParser + doc-comment parser, and if that works that is it, if not, we just parse nodeParser and continue?

Well I guess taht is what you tried above. Why doesn't that work? Probably because `between` runs `nodeParser` first, and only then it checks for `\n`. So `nodeParser` could easily be consuming some `\n`s regardless of you using `between`.

There is `lookAhead` in Parsec, I wonder if that might help.
Maybe some playing with `try`, `lookAhead` and similar can bring us where we need to be.

That said, it none of this works, one solution might be to use Megaparsec instead of Parsec. It is a more serious library for parsing, serious projects use Megaparsec instead of Parsec, we just never switched because we didn't need it yet, but it would be great if we used it from the start. I have no idea how hard the switch is, I think they should be veyr similar / use same API, but I am not 100% sure. It might be out of scope of this PR.

In the meantime I agree with solving this in the separate issue if we can't get it working here soon, to not stop this PR. But I believe there is a chance we can get this working without some crazy stuff.

Martinsos:
Nice stuff! I like it all together, looks solid. Left some comments.

waspc/src/Wasp/Psl/Ast/WithCtx.hs
cprecioso: In this instance I explicitly thought if I wanted `[String]` or `DocumentationComments`. I created this for being a short-hand function to quickly create `WithCtx` at the callsites, so if we ever change `DocumentationComments`' definition, I'd like to keep this as `String` and make it easier for the callsites even if we have to do some wrangling here.

waspc/src/Wasp/Psl/Ast/WithCtx.hs
cprecioso: At first I wrote functions like `mapOutputNode` and `justNode` until I realized this 😂

waspc/src/Wasp/Psl/Ast/WithCtx.hs
Martinsos: Ok perfect, that is what I was asking about -> in that case, it makes sense to keep a `String` here!

raw

waspc/src/Wasp/Psl/Parser/WithCtx.hs
cprecioso: I tried multiple combinations of `try`, `lookAhead`, `choice` and `<|>`, the one in the comment is the one that got me furthest ahead, but yeah.

I saw that Megaparsec can solve the issue at #2985, didn't check for this one. Also haven't checked if it would be difficult to swap, I think I don't have enough of a grip to calibrate that until I try. @infomiho told me that we also use Parsec for the wasp file, so if we migrate to Megaparsec here we should migrate there too. I'm definitely not migrating in this PR 😅

waspc/e2e-test/test-outputs/waspComplexTest-golden/waspComplexTest/.wasp/out/db/schema.prisma
cprecioso: I pointed it out here https://github.com/wasp-lang/wasp/pull/2949#discussion_r2221659732, but yeah I'm not sure why. I can investigate it further, but wanted to sort out the main functionality first.

waspc/src/Wasp/Psl/Ast/WithCtx.hs
cprecioso: Yep, no comment needed, my brain was fried and I forgot to take it out 😅

waspc/src/Wasp/Psl/Ast/WithCtx.hs
cprecioso: Yep, it's a concept from their AST. We don't use it anywhere else in the code so I don't think there's a need to extract it somewhere else.

waspc/src/Wasp/Psl/Ast/WithCtx.hs
cprecioso: Makes sense

waspc/src/Wasp/Generator/DbGenerator.hs
cprecioso: My view is that in the places where I've used `pure` it's because I don't care for the context of the node, we're just generating one in code, indeed context-less. So it checks out semantically.

Plus, IIUC if we add more fields to `WithCtx`, we'd have to update the `Applicative` instance so we'd need to either handle that there, or remove the instance altogether if the context is needed in the future.

waspc/src/Wasp/Psl/Generator/WithCtx.hs
cprecioso: Created the `Wasp.Psl.Common` module and moved it there, wdyt?

waspc/src/Wasp/Psl/Parser/WithCtx.hs
Martinsos: I think it can solve this one also because I saw it being mentioned in this context.
But we would need to check.

You could keep using Parsec for Wasp file, I don't thikn they would conflict in any way, so that shouldn't be an issue. Not to mention we might be soon deprecating the parsing of Wasp files anyway.

As for estimating how much trouble it will be -> Megaparsec is a fork of Parsec, so I am pretty confident the switch would be easy.

This PR is less of an issue, although you are stuck here with not being able to implement trailing comments. But that other PR with whole two modules being copied over from Parsec, that one is a bigger offender in my opinion.

What I would do is take three hours and try to switch Parsec with Megaparsec for the PSL parsing, see where I get. I think there is a fair chance it actually works mostly out of the box, and solves both of the issues we are having in these PRs. If not, I would leave things as they are, and write down the learning in the issue for replacing Parsec with Megaparsec.

waspc/e2e-test/test-outputs/waspComplexTest-golden/waspComplexTest/.wasp/out/db/schema.prisma
Martinsos: Makes sense, and let's investigate it further for sure.

waspc/src/Wasp/Psl/Ast/WithCtx.hs
Martinsos: Let's define it outside of here, because it is confusing right now: it feels as if it is a concept of WithCtx, but it is not, it is a PSL concept.

waspc/src/Wasp/Generator/DbGenerator.hs
Martinsos: You would update Applicative instance to fill it with empty stuff, the problem is, you still won't be notified that you don't actually want all empty stuff potentially on these call sites, so that doesn't help / change anything.

It's kind of like saying that you want all the ingredients in your future kebabs forever: whenever they add new ingredient or remove an ingredient in their offering, you don't care, just put it in (in this case just leave it out hah). But that it assuming you indeed like all the ingredients. Maybe they will add something weird, or you will want only half of the amount of that ingredient, who knows. So maybe you would rather be notified in that case to make a decision for that specific ingredient.

The question is, when you say you don't care about the context in these places: do you not care based on what context is now, or do you not care even in the case of context in the future being whatever it could be? If you are thinking far enough ahead, then fine, but if not, you are missing the opportunity to utilize the compiler for future refactorings. Obviously this depends on what we guess node context might become, but I wanted to hear your toughts on that.

Let's say we add position from the source file to the context. Ok, I guess it makes sense that the code we generated/inserted doesn't have the position, to is being pure in that case is fine.
What if we want to add extra information about the syntax, so we can later, when printing the AST, use that info to generate code exactly as it was before? Ok, in that case we also might be ok with either no info in the node context, or we will anyway have to end up with that field being required and not being able to set it to empty.
All right, when I thikn about it in the future, `pure` sounds ok for this use case (I haven't checked every single one in the PR though but I assume most are similar).

waspc/src/Wasp/Psl/Generator/WithCtx.hs
Martinsos: This is now the second thing we are capturing about the comments -> the fact they start with `///` (at least the doc comments), and above we had that thing where we define them to be String. So I would create Wasp.Psl.Comments and put that there.

waspc/src/Wasp/Generator/DbGenerator.hs
cprecioso: No, I understand the reasoning, but I just can't think of a property that we could sensibly add into the `NodeContext` record that would trigger this condition. I'm resistant to implement this change because we have a lot of callsites of `pure` (now `WithCtx.empty` based on your other comment), and in basically all of them it's pure boilerplate, so I don't want to add even more boilerplate and more verbose. That's the trade-off I went with.
But if you think it's better to go with the other trade-off I'll change it.

waspc/test/Psl/Common/ModelTest.hs
cprecioso: Yeah it makes sense! I left this boring refactor to Claude Code and it chose to do the boring thing 😁

waspc/src/Wasp/Generator/DbGenerator.hs
Martinsos: No no that's it, I agreed above, I also actually can't think of the likely thing that would clash with us using `pure`, it just wasn't clear to me from the start and I didn't know if you put the thought into that.

Ok, sounds good!

raw

cprecioso:
@Martinsos I ended up doing the Megaparsec migration, a bit tedious but straightforward, nothing crazy (#3015). I re-built this code on top of that. I haven't yet addressed all the comments, I still need to do a pass and check which ones are still applicable.

Martinsos:
> @Martinsos I ended up doing the Megaparsec migration, a bit tedious but straightforward, nothing crazy (#3015). I re-built this code on top of that. I haven't yet addressed all the comments, I still need to do a pass and check which ones are still applicable.

Awesome! When you do let me konw and I will take a look again.

infomiho:
I'll wait until you address @Martinsos's comments before I review the PR.

waspc/src/Wasp/Psl/Parser/WithCtx.hs
cprecioso: Megaparsec still gave me a lot of difficulties with this

In the end, I decided to drop this feature even though we migrated to Megaparsec, as it's taking way more time from me than it is worth:

- Non-functional PR: https://github.com/wasp-lang/wasp/pull/3016
- Issue: https://github.com/wasp-lang/wasp/issues/3041

waspc/e2e-test/test-outputs/waspComplexTest-golden/waspComplexTest/.wasp/out/db/schema.prisma
cprecioso: Fixed that formatting change, but now there's a different change where properties inside a block are not separated by a blank line. This is expected, due to my changes in simplifying indentation handling (necessary because we now might return more than one line of text for each member of a block). IMO it looks better.

waspc/src/Wasp/Psl/Ast/WithCtx.hs
cprecioso: defined at https://github.com/wasp-lang/wasp/blob/468dde483a648b637eac035387d60573ab921ab2/waspc/src/Wasp/Psl/Comments.hs

waspc/test/Psl/Generator/WithCtxTest.hs
cprecioso: I understand, and I agree, but I'd rather not add that to the scope of this PR, as this would add one whole new vertical of functionality to the Wasp.Psl module (we'd have the AST definitions lib, the Parser lib, the Generator lib, plus now a "Builders" lib).

waspc/test/Psl/Generator/WithCtxTest.hs
cprecioso: You're right, it's already covered by the example above. Removing.

waspc/test/Psl/Generator/SchemaTest.hs
cprecioso: You're right, I looked more into QuickCheck, it's quite nice! I added `Arbitrary` for `WithCtx` and it helped me find bugs 🎉 (and had me do one small rearch of this solution 😅)

waspc/test/Psl/Generator/WithCtxTest.hs
cprecioso: No needed anymore in the latest version of this PR! 🎉

cprecioso:
@Martinsos @infomiho ready!

waspc/test/Psl/Generator/WithCtxTest.hs
Martinsos: You wouldn't add this to the normal code though, you would add it to the test code. And you would make it only enough to support it. This followed the idea of "build test-only DSLs to make tests more readable". Good thing then is it doesn't have to be fully encompassing / right, it has to serve only the tests and we can do some trade offs in that direction.

Do we have to do it in scope of this PR? No, I guess not, it is not trivial. But when then? Either in some other PR similar to this one, or as a stnadalone effort (which is not likely to happen hah). But then if in similar PR, why not this one?

Maybe let's create a special issue on GH for this and let's handle it as one of "kaizen" tasks in the next sprint(s)? It should be fun task. Or we can skip it if we think we will drop Prisma parsin anyway in the future. Ok yeah that is fair. But if not I would say let's handle it sooner than later.

waspc/test/Psl/Generator/SchemaTest.hs
Martinsos: Awesome!

raw

waspc/test/Psl/Generator/WithCtxTest.hs
Martinsos: Oh wow didn't excpect that! I thought it had its purpose but it just wasn't clear. Great!

raw

waspc/examples/todoApp/schema.prisma
```diff
@@ -16,9 +16,9 @@ model User {
   isOnAfterLoginHookCalled         Boolean @default(false)
   isOnAfterEmailVerifiedHookCalled Boolean @default(false)
 
-  tasks                Task[]
-  address              String?
-  votes                TaskVote[]
+  tasks                 Task[]
```
infomiho, line 19: What happened here? 😄

waspc/src/Wasp/Psl/Generator/Model.hs
```diff
@@ -6,9 +6,10 @@ where
 import qualified Wasp.Psl.Ast.Model as Psl.Model
 import Wasp.Psl.Generator.Attribute (generateAttribute)
 import Wasp.Psl.Generator.Common (PslSource)
+import Wasp.Psl.Generator.WithCtx (generateWithCtx)
 
 generateModelBody :: Psl.Model.Body -> PslSource
-generateModelBody (Psl.Model.Body elements) = unlines $ map (("  " ++) . generateModelElement) elements
```
infomiho, line 11: I guess these spaces aren't really needed since we format the output Prisma file?

waspc/src/Wasp/Psl/Generator/Schema.hs
```diff
@@ -13,12 +14,44 @@ import Wasp.Psl.Generator.Common (PslSource)
...
+      Psl.ConfigBlock.Datasource -> "datasource"
+      Psl.ConfigBlock.Generator -> "generator"
+
+-- | Common structure for all top-level blocks in the schema.
+generateBlock :: String -> String -> PslSource -> PslSource
+generateBlock blockType name body =
+  unlines $
+    [blockType ++ " " ++ name ++ " {"]
+      ++ [indent 2 $ trim body]
+      ++ ["}"]
```
infomiho, line 57: Not sure if applicable here, but you could rewrite it as:
```hs
[trimming|
${blockType} ${name} {
  ${formattedBody}
}
|]
```

Not 100% sure if `trimming` will indent all lines from `formattedBody` or not.

Just a small suggestion, the current code is fine as well.

waspc/src/Wasp/Psl/Generator/Schema.hs
```diff
@@ -13,12 +14,44 @@ import Wasp.Psl.Generator.Common (PslSource)
...
-  Psl.Schema.ViewBlock (Psl.View.View name body) -> "view " ++ name ++ " {\n" ++ generateModelBody body ++ "}"
-  Psl.Schema.TypeBlock (Psl.Type.Type name body) -> "type " ++ name ++ " {\n" ++ generateModelBody body ++ "}"
-  Psl.Schema.EnumBlock (Psl.Enum.Enum name values) -> "enum " ++ name ++ " {\n" ++ generateEnumBody values ++ "}"
-  Psl.Schema.ConfigBlock (Psl.ConfigBlock.ConfigBlock Psl.ConfigBlock.Datasource name content) -> "datasource " ++ name ++ " {\n" ++ generateConfigBlockKeyValuePairs content ++ "}"
-  Psl.Schema.ConfigBlock (Psl.ConfigBlock.ConfigBlock Psl.ConfigBlock.Generator name content) -> "generator " ++ name ++ " {\n" ++ generateConfigBlockKeyValuePairs content ++ "}"
+  Psl.Schema.ModelBlock modelBlock -> generateModel modelBlock
+  Psl.Schema.ViewBlock viewBlock -> generateView viewBlock
+  Psl.Schema.TypeBlock typeBlock -> generateType typeBlock
+  Psl.Schema.EnumBlock enumBlock -> generateEnum enumBlock
+  Psl.Schema.ConfigBlock configBlock -> generateConfigBlock configBlock
```
infomiho, line 29: Nice!

waspc/src/Wasp/Psl/Parser/Common.hs
```diff
@@ -104,6 +105,7 @@ whiteSpace =
 
 lineComment :: Parser String
 lineComment =
-  C.string "//"
-    >> notFollowedBy (C.char '/')
+  try doubleSlashSymbolParser
```
infomiho: Nitpick: the built in parser don't usually have the `Parser` suffix

waspc/src/Wasp/Psl/Parser/Schema.hs
```diff
@@ -15,28 +15,31 @@ import Wasp.Psl.Parser.Enum (enum)
...
-  elements <-
-    many $
-      choice
-        [ Psl.Schema.ModelBlock <$> model,
-          Psl.Schema.ViewBlock <$> view,
-          Psl.Schema.TypeBlock <$> typeBlock,
-          Psl.Schema.EnumBlock <$> enum,
-          Psl.Schema.ConfigBlock <$> configBlock
-        ]
+  elements <- many $ withCtx block
```
infomiho, line 30: Nice simplification

waspc/src/Wasp/Psl/Parser/WithCtx.hs
```diff
@@ -0,0 +1,34 @@
...
+    WithCtx (WithCtx),
+  )
+import Wasp.Psl.Comments (DocumentationComment, documentationCommentSymbol)
+import Wasp.Psl.Parser.Common (Parser, lexeme)
+
+withCtx :: Parser node -> Parser (WithCtx node)
+withCtx nodeParser = do
+  leadingComments <- many documentationComment
+  node <- nodeParser
+  -- TODO: Handle trailing comments properly, right now we fail on them.
```
infomiho: I'd link to the issue here

waspc/test/Psl/Generator/SchemaTest.hs
infomiho: What a good use case of QuickCheck! 😄

web/docs/data-model/prisma-file.md
```diff
@@ -255,8 +255,26 @@ You can also access them from your client code:
...
+  id Int @id @default(autoincrement())
+}
+```
+
+However, Wasp does not support `///` comments in the _trailing_ position:
+
+```prisma title="schema.prisma"
+model User {
+  // highlight-next-line
+  id Int @id @default(autoincrement()) /// The unique identifier for the user.
```
infomiho:
```suggestion
  id Int @id @default(autoincrement()) /// This is not supported
```

I've seen this kind of style to point out that the thing we you reading is not supported.

waspc/test/Psl/Parser/WithCtxTest.hs
```diff
@@ -0,0 +1,132 @@
...
+import Test.Tasty.Hspec
+import qualified Wasp.Psl.Ast.Argument as Psl.Argument
+import qualified Wasp.Psl.Ast.Attribute as Psl.Attribute
+import qualified Wasp.Psl.Ast.Model as Psl.Model
+import qualified Wasp.Psl.Ast.Schema as Psl.Schema
+import Wasp.Psl.Ast.WithCtx (commentedNode)
+import qualified Wasp.Psl.Ast.WithCtx as Psl.WithCtx
+import qualified Wasp.Psl.Parser.Schema as Psl.Parser
+
+spec_parsePslWithCtx :: Spec
```
infomiho, line 17: Do we maybe want to encode the expected failure for trailing comments? It could serve to document the scenarios fully. If it turns out too complex to write the test, I'm okay with not having it.

RAW

waspc/src/Wasp/Psl/Ast/WithCtx.hs
```diff
@@ -0,0 +1,46 @@
...
+  pure = empty
+  WithCtx f context <*> WithCtx node context' =
+    WithCtx (f node) (context <> context')
+
+instance Semigroup NodeContext where
+  NodeContext {documentationComments = comments1} <> NodeContext {documentationComments = comments2} =
+    NodeContext {documentationComments = comments1 <> comments2}
+
+instance Monoid NodeContext where
+  mempty = NodeContext {documentationComments = []}
```
infomiho, line 46: This was a fun Haskell onboarding task is seems 😄

waspc/examples/todoApp/schema.prisma
cprecioso: It was mis-formatted (check mis-alignment on `uppercaseTextRequests`), now it isn't

waspc/src/Wasp/Psl/Generator/Model.hs
cprecioso: I'm indenting everything in a single place:

https://github.com/wasp-lang/wasp/blob/857bfed94b99c0297fbcf3dc850e63005ec350d4/waspc/src/Wasp/Psl/Generator/Schema.hs#L53

The dance of indents plus `lines`/`unlines` calls was making it hard to follow, since each element can now output more than a single line. So now we only indent when creating a block. The rationale is that if you're just generating the model body (inline in Wasp DSL), you don't want/care about the indentation.

waspc/src/Wasp/Psl/Generator/Schema.hs
cprecioso: With all the `T.pack` and `T.unpack` calls, I don't think it makes more readable necessarily, going ahead with the current soulution

waspc/src/Wasp/Psl/Ast/Schema.hs
```diff
@@ -29,23 +31,39 @@ data Block
...
-getGenerators schema = [generator | generator@((Psl.ConfigBlock.ConfigBlock Psl.ConfigBlock.Generator _ _)) <- getConfigBlocks schema]
+getGenerators :: Schema -> [WithCtx ConfigBlock]
+getGenerators = selectBlocks $ \case
+  ConfigBlock configBlock@(Psl.ConfigBlock.ConfigBlock Psl.ConfigBlock.Generator _ _) -> Just configBlock
+  _ -> Nothing
 
-getConfigBlocks :: Schema -> [ConfigBlock]
-getConfigBlocks (Schema blocks) = [configBlock | ConfigBlock configBlock <- blocks]
+selectBlocks :: (Block -> Maybe a) -> Schema -> [WithCtx a]
+selectBlocks filterFn (Schema blocks) = mapMaybe (liftMaybe . fmap filterFn) blocks
```
Martinsos, line 65: I found this quite hard to read, although I got the idea at the end.
Things that made it hard to read:
- It is not just filtering/selecting blocks, it is also unpacking them a bit. That filter function is not a normal filter function, it also maps. The problem is name `filter` because in Haskell that usually means one specific thing, so it threw me off.
- It is a bit confusing to track where is block wrapped into Ctxt and where it is not wrapped into context.

Solution migth be to rename `filterFn` to `extractBlock`, and to rename `(Schema blocks)` to `(Schema blocksWithCtxt)`. `extract` explains correctl ywhat is happening, better than `get` or `filter`.

I would maybe even go further and drop the `liftMaybe` helper function and just do it as one thing: function that takes block wrapped into WithCtx, applies `extractBlock` to it, and then returns appropriate thing wrapped in Maybe. `mapMaybe` at the end is fine.

Idea (I wrote it without trying to compile it so might be wrong, sorry if so):
```hs
extractSpecificBlocks extractSpecificBlock (Schema blocksWithCtxt) =
  mapMaybe $ blocksWithCtxt <&>
    \(WithCtx block ctx) -> extractSpecificBlock block <&> (`WithCtx` ctx)
```

waspc/src/Wasp/Psl/Comments.hs
```diff
@@ -0,0 +1,13 @@
+module Wasp.Psl.Comments
+  ( DocumentationComments,
+    DocumentationComment,
+    documentationCommentSymbol,
+  )
+where
+
+type DocumentationComments = [DocumentationComment]
```
Martinsos, line 8: Nitpick: this is maybe redundant, not sure if you get anything with it except for obscuring the fact it is list.

waspc/test/Psl/Generator/WithCtxTest.hs
```diff
@@ -0,0 +1,74 @@
...
+import qualified Wasp.Psl.Ast.Argument as Psl.Argument
+import qualified Wasp.Psl.Ast.Attribute as Psl.Attribute
+import qualified Wasp.Psl.Ast.Model as Psl.Model
+import qualified Wasp.Psl.Ast.Schema as Psl.Schema
+import Wasp.Psl.Ast.WithCtx (commentedNode)
+import qualified Wasp.Psl.Generator.Schema as Psl.Generator
+import Wasp.Util (trim)
+
+spec_generatePslWithCtx :: Spec
+spec_generatePslWithCtx = do
```
Martinsos, line 17: Consider also adding negative test for trailing comments. Basically a test taht confirms parsing a trailling comment will fail. Such tests are nice because they clearly document that we are aware of trhis behaviour + serve as a nice starting ground once somebody wants to implement it.

raw

cprecioso: miho said the same thing, 🚀 already 💁‍♀️

waspc/src/Wasp/Psl/Comments.hs
cprecioso: Yeah... I see it... In the beginning I didn't know how I was going to represent this (a single string, a list of them, a record like `{leading, trailing}`)... given there are multiple possible ways of representing it, I thought it was nice to explicitly specify the method I chose.

waspc/src/Wasp/Psl/Ast/Schema.hs
cprecioso: Fixed at https://github.com/wasp-lang/wasp/pull/3076, wdyt?

Issue #2951, opened 2025-07-14
Make Wasp AI work with binary files


sodic:
Version: 0.17.0

Wasp AI currently sets up a project by:
1. Reading starter files and interpreting them as `Text`.
2. Writing this `Text` data into the same relative files in the new project.

This is a problem for binary files, such as `favicon.ico` which we added to the skeleton with https://github.com/wasp-lang/wasp/pull/2625. Since `favicon.ico` is in a binary format, Wasp fails when it tries to read the file as `Text`.
```
wasp-bin: /home/filip/.local/share/wasp-lang/0.17.0-rc5/data/Cli/starters/skeleton/public/favicon.ico: hGetContents: invalid argument (invalid byte sequence)
```

todo: finish

cprecioso:
Wasp AI is not part of Wasp anymore

PR #2625, opened 2025-04-04
Generate a default `favicon.ico` in `public` folder, don't generate one on compile

FranjoMindek:
New projects start with `favicon.ico` inside of `public` directory.
If there is no `favicon.ico` we still exclude the favicon `<link>` tag from the `<head>` of `index.html`.

Fixes https://github.com/wasp-lang/wasp/issues/1979

Question for reviewers: Do we, and where do we notify users of new behavior in docs?

TODO: golden tests

PR #2952, opened 2025-07-14
Fix favicon issue with Wasp AI

sodic:
Temporary fix until we implement https://github.com/wasp-lang/wasp/issues/2951


infomiho:
@sodic removing of the `favicon.ico` from `react-app` is probably a leftover after we stopped using it here: https://github.com/wasp-lang/wasp/pull/2625/files#diff-5c849f0c343eac448a1ab375e436b53c29104a39f58d664f9334372f5ed6788aL169?

I'd also mention in the PR description why we added `.gitkeep`.

PR #2958, opened 2025-07-14
Pin Rollup version that Vite uses in web-app

infomiho:
`vite build` fails for https://github.com/wasp-lang/ask-the-documents with the latest RC

Connected to: https://github.com/rollup/rollup/issues/6012

We are overriding the `rolldown` version Vite uses to the latest version that works and it's recommended by some commentators in the linked issue.

Ideally, we'd remove this override ASAP since we assume that Rollup will be functional for us in a few days.

### DIFF
--- a/waspc/data/Generator/templates/react-app/package.json
+++ b/waspc/data/Generator/templates/react-app/package.json
@@ -6,6 +6,7 @@
   "type": "module",
   {=& depsChunk =},
   {=& devDepsChunk =},
+  {=& overridesChunk =},
   "scripts": {
     "start": "vite",
     "build": "tsc --build && vite build"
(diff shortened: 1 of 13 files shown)


cprecioso:
Why override? Wouldn't us pinning it as a regular dependency make vite use it too?

infomiho:
@cprecioso I thought this is the recommended way to sort issues with deps of your deps (we don't use `rollup` directly). Would it be a more common thing to put `rollup` as one of our deps?

cprecioso:
I thought about it more, setting it as override is nicer, you're right. We're not depending on it directly so it wouldn't make sense to add it as a dep. And we'd be at npm's mercy of wanting to resolve our dependencies like we want.

PR #2972, opened 2025-07-16
Automatically test `wasp deploy`

### LINKED ISSUE #2665: Test `wasp deploy` for all providers (Fly, ...)
Currently it is only Fly, soon also Railway.

Maybe we test this already when we deploy example apps? In any case we want to be explicit about us testing this intentionally, to ensure these tests don’t go away by accident, even if they are not standalone.

sodic:
Closes #2665


sodic:
Closing this in favor of #2990

PR #2974, opened 2025-07-17
Fix tailwind ES modules warning in `basic` starter

FranjoMindek:
As reported by Filip (internal message): https://discord.com/channels/686873244791210014/1351908377533354025/1395147983347257436

We received following warning message when using `wasp start` on new `basic` starter project.

image

This is because the `tailwind.config.cjs` was forced to use `import` instead of `require` by `eslint` rule.

We fix it by ignoring the rule.

NOTE1:
There is an option to enable `require` imports in `eslint`, but IMHO, I would rather make it error as to promote `import` syntax.
https://typescript-eslint.io/rules/no-require-imports/#allowasimport

NOTE2:
This does not remove the `eslint` error:
```ts
// @typescript-eslint/no-require-imports
const { resolveProjectPath } = require('wasp/dev')
```
This is enough by itself, but I want to leave the `@typescript-eslint/no-require-imports` to make it clear what the reason is for this.
```ts
// eslint-disable-next-line
const { resolveProjectPath } = require('wasp/dev')
```

### DIFF
--- a/waspc/data/Cli/starters/basic/eslint.config.js
+++ b/waspc/data/Cli/starters/basic/eslint.config.js
@@ -13,9 +13,9 @@ export default defineConfig([
   eslintReact.configs.flat['jsx-runtime'],
   eslintPrettier,
   // Overrides:
-  // `@typescript-eslint/no-require-imports` is enabled by default in `typescript-eslint/recommended` config.
-  // This allows us to use `require` syntax in CJS files.
   {
+    // `@typescript-eslint/no-require-imports` is enabled by default in `typescript-eslint/recommended` config.
+    // This allows us to use `require` syntax in CJS files.
     files: ["**/*.{cjs,cts}"],
     rules: {
       "@typescript-eslint/no-require-imports": ["off"]


waspc/data/Cli/starters/basic/tailwind.config.cjs
```diff
@@ -1,4 +1,5 @@
-import { resolveProjectPath } from "wasp/dev";
+// eslint-disable-next-line @typescript-eslint/no-require-imports
```
FranjoMindek: We can add additional comment as explanation for this.
Is this necessary? This will be in each `basic` starter.

sodic (changes requested):
I think the core issue here is that our ESLint config is wrong. It enforces the `import` syntax in CJS modules, which is contradictory. So let's fix that instead.

Or verify that I'm wrong :)

FranjoMindek:
I'm aligned on the idea that we should promote ESM when possible, and make CJS an exception.
In that case disallowing it, and ignoring it in this specific case is the way to go

But that might not be the best for the framework.
I can do:
```ts
export default defineConfig([
  // ...old
  {
    "files": ["*.cjs"],
    "rules": {
      "@typescript-eslint/no-require-imports": ["off"]
    }
  }
]);
```
Where I turn off this rule for `.cjs` files.

It's also possible is to use: `"@typescript-eslint/no-require-imports": ["error", { "allowAsImport": true }]` instead of `"off"`.
That would allows only for `require` imports in `.cjs` files, and no other `require` use cases.
I don't know how much freedom do you want to give.

sodic:
Like I said, I've read about this stuff recently but forgot the details. In any case, we should set the lint rules to enforce what's correct for our setup instead of setting incorrect rules and overriding them to allow correct code.

> I'm aligned on the idea that we should promote ESM when possible, and make CJS an exception.
**In that case disallowing it, and ignoring it in this specific case is the way to go**

I wouldn't say this is the case at all, and I believe we're talking about two different things here. One is ESM vs CJS, another one is the import syntax. Depending on the setup, those two choices can be independent of each other.

We not only allow CJS, we explicitly use it and rely on it - `tailwind.config.cjs` is our official file. And, since we already use CJS, our lint rules should allow us and the users to use them correctly.

I'm guessing the bold part means _"We don't really want Wasp projects to use CJS, but we have to use it for this one thing, so it's best to craete a special case just for us._

If that's the case, I vote against haphazardly putting any extra limitations on users using CJS:
- This is a breaking change and we don't know what it will break.
- We need CJS for Tailwind, so users might need it for something similar.

We should properly reasearch the implications of disallowing CJS and decide accordingly.

The course of action should therefore be:
- Understand why our setup works the way it works (e.g., which import syntax is allowed where and why) and set up lint rules that model it correctly.
- Create an issue for discussing CJS module support in Wasp

FranjoMindek:
> I'm guessing the bold part means _"We don't really want Wasp projects to use CJS, but we have to use it for this one thing, so it's best to craete a special case just for us._

That was my initial reaction.

> If that's the case, I vote against haphazardly putting any extra limitations on users using CJS:
> 
>     * This is a breaking change and we don't know what it will break.
> 
>     * We need CJS for Tailwind, so users might need it for something similar.

Makes sense.

>		* We need CJS for Tailwind, so users might need it for something similar.

We don't. We changed `.js` to `.cjs` because the module was written in CommonJS.
https://github.com/wasp-lang/wasp/pull/1019#discussion_r1114290526
We could've also re-wrote it in ESM by changing to import/export syntax.

If we re-write it with ESM and use the `.js` extensions, everything works. Of course also requires change to `TailwindConfigFile.hs`

e.g.
https://github.com/wasp-lang/wasp/pull/2987/files
Incomplete draft (didn't change examples, didn't change docs, didn't describe breaking change etc.)
but new `basic` projects work with above.



>		* Understand why our setup works the way it works (e.g., which import syntax is allowed where and why) and set up lint rules that model it correctly.

I think that eslint change above is pretty much what we want, but I'll dig a bit deeper.

>     * Create an issue for discussing CJS module support in Wasp

https://github.com/wasp-lang/wasp/issues/2986

sodic (approved):
All good!

I'd maybe put the elaborating comment inside the rule object, but that's a minor nitpick.

PR #2978, opened 2025-07-17
Issue : - Frequent Cor's Errors which Dev's building on wasp are getting

### LINKED ISSUE #2933: CORS errors in browser upon server errors (500) - what can we do about that?
We have a lot of people hitting the situation where they deploy the app and then get CORS error in the browser. And problem is not CORS at all, the problem is actually e.g. missing env var on the server, server returning 500 or something, and browser showing a CORS error.

This masks what the real problem is and people spend quite some time confused about this.

While we do now have "custom env var validations" as a Wasp feature, and that helps somewhat if employed, any other error can still cause this.

We will also have `wasp build start` soon which will make it easier to test built app locally, so that will also reduce these errors, but again they can happen.

We should look into ways to mitigate this. Ideal would be if we can show an actual error, not a CORS error. I am not sure if that is possible, or to which degree, but if not, then we should look into other ways to other avoid these errors or to inform the user sooner than later.

0xTaneja:
### Description

closes #2933 

### Description

Many users deploy a Wasp app, hit the client, and see CORS error in the browser console even though the real cause is a missing env-var or other 5xx on the server.
This happens because the response that bubbles up from a crashed server has no Access-Control-Allow-Origin header, so the browser rewrites the failure as a CORS issue and hides the actual JSON body.

I maybe wrong on the issue but the only problem regarding the CORS which i found out in our codebase was it,

### DIFF
--- a/waspc/data/Generator/templates/server/src/app.js
+++ b/waspc/data/Generator/templates/server/src/app.js
@@ -11,7 +11,10 @@ const app = express()
 // NOTE: Middleware are installed on a per-router or per-route basis.
 
 app.use('/', indexRouter)
-
+app.use((_req, res, next) => {
+  res.setHeader('Access-Control-Allow-Origin', '*')
+  next()
+})
 // Custom error handler.
 app.use((err, _req, res, next) => {
   // As by expressjs documentation, when the headers have already
--- a/waspc/e2e-test/test-outputs/waspBuild-golden/waspBuild/.wasp/build/.waspchecksums
+++ b/waspc/e2e-test/test-outputs/waspBuild-golden/waspBuild/.wasp/build/.waspchecksums
@@ -487,7 +487,7 @@
             "file",
             "server/src/app.js"
         ],
-        "a10bb6f3daab886b0685994368711fd7718a2a3e50a9329cd7da6394b9c59302"
+        "289d8d09cd6d2446ef0ad8dc443afed9aaed33b689cc432dc6246101f67cf281"
     ],
     [
         [
--- a/waspc/e2e-test/test-outputs/waspBuild-golden/waspBuild/.wasp/build/server/src/app.js
+++ b/waspc/e2e-test/test-outputs/waspBuild-golden/waspBuild/.wasp/build/server/src/app.js
@@ -11,7 +11,10 @@ const app = express()
 // NOTE: Middleware are installed on a per-router or per-route basis.
 
 app.use('/', indexRouter)
-
+app.use((_req,res,next)=>{
+  res.setHeader('Access-Control-Allow-Origin','*')
+  next()
+})
 // Custom error handler.
 app.use((err, _req, res, next) => {
   // As by expressjs documentation, when the headers have already
--- a/waspc/e2e-test/test-outputs/waspCompile-golden/waspCompile/.wasp/out/.waspchecksums
+++ b/waspc/e2e-test/test-outputs/waspCompile-golden/waspCompile/.wasp/out/.waspchecksums
@@ -494,7 +494,7 @@
             "file",
             "server/src/app.js"
         ],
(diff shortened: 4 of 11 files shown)


infomiho:
Hey @0xTaneja thank you for the contribution. 

I think this might be going too loose with security because we are effectively disabling CORS for all endpoints for all of our users. We want to be explicit in which domains we allow to connect to our server. Setting the value to `*` allows any client to connect to our server (which is [usually fine](https://advancedweb.hu/is-access-control-allow-origin-star-insecure/) but we don't want that _by default_). Users can do this themselves by modifying the global middleware.

I'd like us to more explicitly solve #2933 by doing something like this:
- we already have `/` that returns 200 and an empty body
- we can add `Access-Control-Allow-Origin: *` for only that endpoint
-  we can ping the server at `/` when there is Network error on the client, so we can output something in the browser e.g. `Hey, it seems like the client can't connect to the server - the server is probably not running or the client is trying to connect to the wrong URL. If the URL looks correct XXX, check the server logs to debug.`
	- I'd be as precise as possible here and try to output it only Network errors if possible.

This is a rough idea, I'm not 100% about this, but I'm trying to give you an idea how we'd like the DX to look like.

0xTaneja:
yep i agree with you @infomiho i have added `Access-Control-Allow-Origin: *` globally but i can make it to that particular endpoint.
pinging could be an option
and also one more thing maybe we can do is validate things like if required env's to make the app run are missing, i know we are doing this on haskell side but maybe if we can do this on server side as well then we can throw the error !!
another thing i suggest is /health route as well just to check things are good for db and env's stuff, and in deploy documentation maybe tell users to check /health route when they get error !!

Martinsos:
> yep i agree with you @infomiho i have added `Access-Control-Allow-Origin: *` globally but i can make it to that particular endpoint. pinging could be an option and also one more thing maybe we can do is validate things like if required env's to make the app run are missing, i know we are doing this on haskell side but maybe if we can do this on server side as well then we can throw the error !! another thing i suggest is /health route as well just to check things are good for db and env's stuff, and in deploy documentation maybe tell users to check /health route when they get error !!

Validating required envs -> we are actually doing this on the server side now also since recently if I am correct, check out https://wasp.sh/docs/project/env-vars#custom-env-var-validations -> we will run those checks on server start, and if a deployed server is missing env vars, it will fail with useful error message! Right @infomiho ? That said hm, I guess that error message will be logged up there into the logs on the hosting provider, and you might still be getting CORS errors from the client perspective. So that helps if you go there to see what the issue is, but you still need to konw that CORS error might not be CORS error.

@infomiho pinging idea makes general sense to me -> by checking `/` or `/health` or sometihng we can check if server is running or not, so if it is not, we can make it clear that is the real error, not CORS. Although we have to then do that check on CORS errors hm, sounds a bit weird to be performing such extra checks in production, but if it works ... .

I wonder if there is some other way to go around all this? So first of all, why do we get CORS errors if server crashed / there is some 500? @0xTaneja said the following:

> This happens because the response that bubbles up from a crashed server has no Access-Control-Allow-Origin header, so the browser rewrites the failure as a CORS issue and hides the actual JSON body.

If that is correct, would the solution be adding Access-Control-Allow-Origin header onto errors, so that they properly get through? We still wouldn't be doing it for all respnses from our server, so no global `*`, but only for these error response so they get properly through? Am I talking nonsense or is this doable?

Additionally:
- We don't really have a standard way to do a health check. I know we discussed it multikple times but never ended up on a solution. I think we concluded users can add it on their own, but maybe indeed we should have something of our own hm. Still not sure, I can see people wanting to remove in case we add it.
- Could we provide people with better default CORS story? We know the URL of the client anyway, we could set CORS to that as a deafult setting? But ok this is story for another issue (I added it to another CORS issue right now).

cprecioso:
Hi, I will close this based on inactivity, feel free to reopen or send a new one 😊

PR #2981, opened 2025-07-18
Update Fly toml files for new deployment

sodic:
We redeployed mage after an outage. These are the new toml files.


infomiho:
What's up with the docs changes, did you check out from `release`?

mage/fly-client.toml
```diff
@@ -1,14 +1,22 @@
...
+primary_region = 'cdg'
+
+[build]
 
 [http_service]
   internal_port = 8043
   force_https = true
-  auto_stop_machines = false
-  auto_start_machines = false
-  min_machines_running = 1
```
infomiho, line 14: Change `min_machines_running` from `1` to `0` - maybe we want to keep the old value to keep the Mage more responsive?

PR #2985, opened 2025-07-21
Patch Parsec's comment parsing

cprecioso:
Part of #2949

#### Parsec...

When creating a parser with Parsec, you can create a base lexer over which to build up your high-level parser. This lexer accepts options for describing how comments work in the language, and will then handle them transparently as part of the `whiteSpace` parser.

However, the comment options only accept `String`s. Parsec will immediately turn those strings into `Parser String`s by running `Text.Parsec.string commentLine`, but it doesn't accept parsers itself.

#### ... and Prisma

The Prisma Schema Language (PSL), has two kinds of comments: regular comments that can be ignored (start with `//`), and documentation comments that have to become part of the AST (start with `///`). 

If we just tell Parsec that comments start with `//`, it will ignore the documentation comments too. But if we don't let Parsec handle the comments automatically, we will have to remember to put optional comment handling in all of our parsers, spamming it semi-randomly through the codebase, and risking forgetting it some place, or not creating a test for a specific comment placements.

### Solution

We need to be able to tell Parsec that the comment definition is
```hs
commentLine :: Parser ()
commentLine = string "//" >> notFollowedBy (char '/')
```

The patch inside of Parsec to accept that is just a couple of lines. But to do it I've needed to copy the `Text.Parsec.Language` and `Text.Parsec.Token` modules into our codebase, and modify them so they accept a `Parser` for comment symbols instead of only a string.


waspc/src/Wasp/Psl/Parser/Common/Token.hs
```diff
@@ -0,0 +1,677 @@
...
+
+    -- whiteSpace
+    whiteSpace = skipMany (simpleSpace <|> oneLineComment <?> "")
+
+    simpleSpace =
+      skipMany1 (satisfy isSpace)
+
+    oneLineComment =
+      do
+        _ <- try (commentLine languageDef)
```
cprecioso, line 675: This is the main change, instead of creating a `string` parser, we just pass a `Parser` already

```diff
-        _ <- try (string (commentLine languageDef))
+        _ <- try (commentLine languageDef)
```

cprecioso:
If we merge this solution, I will create an issue and link to here so that we know that we must remove this workaround

Martinsos:
Uff! Nice job, but it sucks that this is solution.

I have to yet check the code a bit, but before that, let me ask a couple of questions:
 - When do you think we might be able to remove this workaround?
 - As I commented in other PR, what if we switched to Megaparsec? I am guessing it might have more flexible way to define comments.
 - How did you come up with this approach -> completely on your own, or was it suggested somewhere online? I am asking because I wonder if this is what others also do, or how they solve it. Ok I found this issue https://github.com/haskell/parsec/issues/15 -> maybe you got inspiration from there? They suggest there to go use megaparsec and that was 10 years ago.

cprecioso:
Closed in favour of #3015

Martinsos:
Awesome!

PR #2760, opened 2025-05-16
Hotfix: Adapt to new flyctl version

cprecioso:
New `flyctl` version outputs `{ name: string, code: string }` instead of `{ Name: string, Code: string }`. Allow both versions.

PR #2990, opened 2025-07-22
Automatically test deployment to Fly

### LINKED ISSUE #2993: Automatically test deploy for Fly
Test the `wasp deploy fly launch` command in the CI. More details in the parent issue.

sodic:
Closes #2993


.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,126 @@
...
+on:
+  push:
+    paths:
+      - "waspc/**"
+      - ".github/**"
+    branches:
+      - main
+      - release
+      # TODO: Remove before merge
+      - filip-automatic-deploy-test-fly
```
sodic: A couple of questions:
- When do you guys think we should trigger this? I thought only on main and release.
- Does it need to be reusable?

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,126 @@
...
+
+jobs:
```
sodic: Does my job split make sense, and why or why not?

What I was going for:
- Once we expand the test to railway, we can reuse the same prefix.
- Each job is its own thing. I treated them as I'd treat normal functions, but I don't know what the best practice is.

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,126 @@
...
+            | sed -E 's/#.*//' \
+            | awk 'NF {$1=$1;print}' \
+            | sed -E 's/^/--server-secret /' \
+            | paste -sd' ')" >> $GITHUB_ENV
+
+      - name: Deploy App to Fly.io
+        working-directory: waspc/examples/todoApp
+        run: |
+          # NOTE:The `yes` command is necessary because the `fly launch` command prompts for confirmation 
+          yes | wasp-cli deploy fly launch $APP_PREFIX $FLY_REGION --org wasp-testing ${{ env.ENV_VARS }}
```
sodic: At one point during our deployment process, Wasp outputs the database URL to the user. Github somehow knows to censor this:

image

My questions:
- How does it know?
- Can we rely on this?
- Even if it was in plaintext, is it a problem? Someone could mess around with our testing apps but it doesn't sound like a big deal.

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,125 @@
...
+      APP_PREFIX: ${{ needs.generate_uniqe_app_prefix.outputs.app_prefix }}
+
+    steps:
+      - name: Install Fly CLI
+        uses: superfly/flyctl-actions/setup-flyctl@v1
+
+      - name: Clean up testing app from Fly.io
+        run: |
+          # NOTE: We are relying on Wasp's naming conventions here
+          flyctl apps destroy -y $APP_PREFIX-{server,client,db}
```
sodic: I'm open to recommendations on how to do this:
- One option is the old command `flyctl apps list -q | grep "$APP_PREFIX" | xargs flyctl apps destroy -y` with an extra check for the app prefix. This is more robust (doesn't duplicate Wasp conventions) but arguably less clear.
- There was talk of using JS instead of Bash.

.github/workflows/test-deploy.yaml
infomiho: What do you say we make this a requirement before we release like we have here? https://github.com/wasp-lang/wasp/blob/main/.github/workflows/waspc-release.yaml#L16

**One big con of doing that:** Fly or Railway go down - we can't release since the tests can't pass.

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,125 @@
...
+      - "waspc/**"
+      - ".github/**"
+    branches:
+      - main
+      - release
+      # TODO: Remove before merge
+      - filip-automatic-deploy-test-fly
+
+jobs:
+  generate_uniqe_app_prefix:
```
infomiho: Nitpick: we used dashes in all other job names in our workflows


```suggestion
  generate-unique-app-prefix:
```

Also, there was a typo in `unique`

.github/workflows/test-deploy.yaml
infomiho: Why do you need to share the prefix between Fly and Railway? I think it's okay to do it, but I'm not sure if necessary.

Normally, I wouldn't split it like this since each job spins up a VM. But, since we want to reuse the prefix, I believe this is the only way to share it among multiple jobs.

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,125 @@
...
+    name: Generate unique deployment prefix
+    runs-on: ubuntu-latest
+
+    outputs:
+      app_prefix: ${{ steps.generate_app_prefix.outputs.app_prefix }}
```
infomiho: Nitpick: I got confused by `outputs` being on top of `steps` because I'm used first reading the steps and then the outputs are usually below. Now I started looking for code on top and realized that it's the same name and then I looked below.

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,125 @@
...
+
+    steps:
+      - name: Generate unique app prefix from commit SHA
+        id: generate_app_prefix
+        run: |
+          echo "app_prefix=ci-$(echo $GITHUB_SHA | head -c 7)" >> "$GITHUB_OUTPUT"
```
infomiho: Why not:
```suggestion
          echo "APP_PREFIX=ci-$(echo $GITHUB_SHA | head -c 7)" >> "$GITHUB_OUTPUT"
```

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,125 @@
...
+
+  deploy_app_to_fly:
+    name: Deploy Wasp app
+
+    runs-on: ubuntu-latest
+    needs: generate_uniqe_app_prefix
```
infomiho: Nitpick: I found it more natural to read, that's it :D 
```suggestion
    needs: [generate_uniqe_app_prefix]
```

.github/workflows/test-deploy.yaml
infomiho: Github has some common pattern it recognizes and masks: https://www.blacksmith.sh/blog/best-practices-for-managing-secrets-in-github-actions#:~:text=Configured,keys

> Configured GitHub secrets
> OAuth, JWT, Bearer tokens, etc
> Database connection strings
> GitHub tokens (PATs and installation tokens)
> Personal access tokens
> SSH keys

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,125 @@
...
+      CLIENT_HOSTNAME: ${{ needs.deploy_app_to_fly.outputs.client_hostname }}
+
+    steps:
+      - name: Smoke test the server
+        run: |
+          curl --fail --silent -X POST https://$SERVER_HOSTNAME/operations/get-date
+
+      - name: Smoke test the client
+        run: |
+          curl --fail --silent https://$CLIENT_HOSTNAME
```
infomiho: Maybe we should expect to see some string in the output? e.g. `<title>Kitchen Sink App</title>`

.github/workflows/test-deploy.yaml
infomiho: I'm a bash noob, so I just learned while reading this that `$APP_PREFIX-{server,client,db}` expands into three different command calls. I'd argue that this is not that advanced, but maybe in the future, someone glances over that fact and something goes wrong. 

What if we dumbed it down and just wrote it three times:

```
flyctl apps destroy -y $APP_PREFIX-server
flyctl apps destroy -y $APP_PREFIX-client
flyctl apps destroy -y $APP_PREFIX-db
```

infomiho: I'm not sure I see the benefit of converting just this bit into JS (maybe I'm wrong). 

If we added it somehow to our testing utilities that would be cool - `wasp-testing-utils destroy-wasp-app-on-fly` 😄

infomiho (commented):
Nice PR, very easy to review.

Left some comments

.github/workflows/test-deploy.yaml
sodic: Oh, cool!

What about the other questions? Is it alright if we ignore this problem?

.github/workflows/test-deploy.yaml
sodic: > What if we dumbed it down and just wrote it three times:

Yes, we could do that.

@infomiho What about the first option with the pipes, where I said:
> This is more robust (doesn't duplicate Wasp conventions) but arguably less clear.

Finally:
> I'm not sure I see the benefit of converting just this bit into JS (maybe I'm wrong).

I definitely agree. I thought you wanted JS (based on the post mortem comment) and was ready to compromise :D
But if we both prefer Bash, awesome.

.github/workflows/test-deploy.yaml
sodic: I actually fixed the typo on thursday and forgot to push it...

I'll take care of the kebab case.

.github/workflows/test-deploy.yaml
sodic: The convention for outputs is lowercase I believe.

Btw, the convention for bash variables is also `lowercase_underscore`.
Posix suggests `UPPERCASE_UNDERSCORE` should only be used for env vars.

.github/workflows/test-deploy.yaml
sodic: > Why do you need to share the prefix between Fly and Railway?

Hm, maybe I don't, but how else would I name Railway apps?

.github/workflows/test-deploy.yaml
sodic: That works too. I don't know a lot here so you guys tell me what you think is best.

.github/workflows/test-deploy.yaml
sodic: > Now I started looking for code on top and realized that it's the same name and then I looked below.

I'll move them below, but I didn't get this sentence.

.github/workflows/test-deploy.yaml
cprecioso: I don't have a good answer for when to test this. On `main` and `release` sounds reasonable.

One thing that I would add is a `workflow_dispatch` trigger so that we can trigger it manually if we want.

.github/workflows/test-deploy.yaml
cprecioso: I'd say: turn the prefix into `test-app-${{ github.sha }}`, no need for extra jobs

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,128 @@
...
+      server_hostname: ${{ steps.save_hostnames.outputs.server_hostname }}
+      client_hostname: ${{ steps.save_hostnames.outputs.client_hostname }}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          # Workaround for a Github Checkout action bug
+          # https://github.com/actions/checkout/issues/1359#issuecomment-1567902034
+          ref: ${{ github.event.pull_request.head.sha }}
```
cprecioso: Are you just copying this from `waspc-ci.yaml`? This bug is a non-issue. It's on my list of things to deal with for that workflow.

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,128 @@
...
+    runs-on: ubuntu-latest
+    needs: generate_unique_app_prefix
+    env:
+      FLY_API_TOKEN: ${{ secrets.FLY_GITHUB_TESTING_TOKEN }}
+      FLY_REGION: mia
+      APP_PREFIX: ${{ needs.generate_unique_app_prefix.outputs.app_prefix }}
+
+    outputs:
+      server_hostname: ${{ steps.save_hostnames.outputs.server_hostname }}
+      client_hostname: ${{ steps.save_hostnames.outputs.client_hostname }}
```
cprecioso: You can also add

```yaml
environment:
  name: fly-deploy-test
  url: https://${{ steps.save_hostnames.outputs.client_hostname }}/
```

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,128 @@
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
cprecioso: in my personal opinion, naming self-describing actions is not needed

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,128 @@
...
+        with:
+          # Workaround for a Github Checkout action bug
+          # https://github.com/actions/checkout/issues/1359#issuecomment-1567902034
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Set up Haskell
+        uses: ./.github/actions/setup-haskell
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
```
cprecioso: I'd set `version: 22` here

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,128 @@
...
+
+      - name: Install Wasp CLI
+        working-directory: waspc
+        run: ./run install
+
+      - name: Install Fly CLI
+        uses: superfly/flyctl-actions/setup-flyctl@v1
```
cprecioso: Heads up! This action will just install the latest version of fly, whatever it is. I think this is the behaviour we want, but I'd add a comment here.

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,128 @@
...
+        run: |
+          # NOTE: This assumes env var values don't contain:
+          #   - The character `#`.
+          #   - An empty line (i.e., there are no multiline env values).
+          echo "ENV_VARS=$(
+            cat .env.server.headless \
+            | sed -E 's/#.*//' \
+            | awk 'NF {$1=$1;print}' \
+            | sed -E 's/^/--server-secret /' \
+            | paste -sd' ')" >> $GITHUB_ENV
```
cprecioso: Why do the github output song and dance? I'd do this and the `wasp deploy` call in the same script, in a single step

.github/workflows/test-deploy.yaml
cprecioso: > Can we rely on this?

I'd say, kinda, not really, but I also don't expect it to change.
There's a way of explicitly setting something to be a secret and then you can rely on it being censored, but it requires you to know the secret, verbatim, beforehand: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#masking-a-value-in-a-log
I'd keep it like this, revisit if we get any problems.

> Even if it was in plaintext, is it a problem?

I don't think so? DBs will get removed anyway. Also it is only accessible from within machines on our Fly account https://fly.io/docs/postgres/connecting/connecting-external/#:~:text=Fly%20Postgres%20databases%20can%20be,to%20the%20internet%20by%20default.

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,128 @@
...
+        run: |
+          curl --fail --silent https://$CLIENT_HOSTNAME
+
+  clean_up_deployed_fly_app:
+    runs-on: ubuntu-latest
+    name: Clean up deployed Fly app
+    needs: [generate_unique_app_prefix, deploy_app_to_fly]
+    # NOTE: Fly deployments can sometimes "fail" but still deploy the apps. We
+    # want to always clean them up.
+    if: needs.generate_unique_app_prefix.result == 'success'
```
cprecioso: You mean:

```suggestion
    if: always()
```

;)

.github/workflows/test-deploy.yaml
infomiho: Yeah, I think the downside outweighs the upside for me here. @cprecioso @FranjoMindek what do you guys think?

.github/workflows/test-deploy.yaml
cprecioso: Does this command fail or bail if the app doesn't exist? e.g. it got to deploy `db` and `server` but failed on `client`, and if the command doesn't find `client`, it will not do anything. Maybe write the command three times with a `|| true` each time?

> I'm a bash noob, so I just learned while reading this that $APP_PREFIX-{server,client,db} expands into three different command calls

nitpick: it doesn't expand into three commands, it just expands that argument into three different arguments for the same command. Fly just accepts more than one app name to destroy

> I'm not sure I see the benefit of converting just this bit into JS

yeah, meh

> What about the first option with the pipes

interesting for me, but I'd use the `--json` option and do some `jq` wrangling as it's probably more stable than the human-facing output (or maybe not, see #2760)

.github/workflows/test-deploy.yaml
infomiho: > but how else would I name Railway apps?

Reuse the logic e.g. in some script both of the jobs use? e.g. `./generate-app-name.sh` or something like that.

This is really optimizing for not starting one VM, but I feel like it's wasteful to start a VM just to generate an app name. But I won't block on this.

.github/workflows/test-deploy.yaml
infomiho: I don't think it's a big deal even if it's visible. It's a short-lived database which I don't think people can even connect to directly since the host is in the internal Fly network.

.github/workflows/test-deploy.yaml
cprecioso: Also, for the other jobs, I don't mind the job split but I'd like them to be more tersely named, so it's easier to see at a glance that they are related `fly_deploy`, `fly_test`, `fly_destroy`.

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,128 @@
...
+          echo "server_hostname=$(flyctl status -j -c fly-server.toml | jq -r '.Hostname')" >> "$GITHUB_OUTPUT"
+          echo "client_hostname=$(flyctl status -j -c fly-client.toml | jq -r '.Hostname')" >> "$GITHUB_OUTPUT"
+
+  smoke_test_app:
+    name: Run smoke tests on deployed app
+
+    runs-on: ubuntu-latest
+    needs: deploy_app_to_fly
+    env:
+      FLY_API_TOKEN: ${{ secrets.FLY_GITHUB_TESTING_TOKEN }}
```
cprecioso: You don't need this env here

.github/workflows/test-deploy.yaml
infomiho: I'm not a fan of the `xargs` approach for the reasons I mentioned in the postmortem. Even though we duplicate Wasp conventions, it's more explicit and safe. 

I'm okay with all of the solutions you proposed. 

1) Calling the command explicitly three times - super clear, but it relies on Wasp convention (who says the other commands don't rely on them as well... we can change our naming scheme to be even more wild and not even include the app name in all of the apps - I know we won't do this, but just saying)
2) Using `xargs` - clean but it can be dangerous and maybe a little bit opaque for bash noobs like me
3) Using JS - overkill for sure, but I could see it in some sort of common utils JS package

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,128 @@
...
+    # want to always clean them up.
+    if: needs.generate_unique_app_prefix.result == 'success'
+
+    env:
+      FLY_API_TOKEN: ${{ secrets.FLY_GITHUB_TESTING_TOKEN }}
+      APP_PREFIX: ${{ needs.generate_unique_app_prefix.outputs.app_prefix }}
+
+    steps:
+      - name: Install Fly CLI
+        uses: superfly/flyctl-actions/setup-flyctl@v1
```
cprecioso: We want the clean up job to be the most robust, we don't want to end up with a bunch of leftovers in our Fly account. Therefore, I'd say that we should pin the Fly CLI version here (`with: version: blah`), so that changes in the CLI don't affect our ability to destroy apps.

.github/workflows/test-deploy.yaml
infomiho: If you moved it, the point go across :D sorry for terrible writing

.github/workflows/test-deploy.yaml
infomiho: It seems like you know more about this than me, carry on 👍

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,128 @@
...
+          curl --fail --silent -X POST https://$SERVER_HOSTNAME/operations/get-date
+
+      - name: Smoke test the client
+        run: |
+          curl --fail --silent https://$CLIENT_HOSTNAME
+
+  clean_up_deployed_fly_app:
+    runs-on: ubuntu-latest
+    name: Clean up deployed Fly app
+    needs: [generate_unique_app_prefix, deploy_app_to_fly]
```
cprecioso: You need to add the `smoke_test_app` here too

.github/workflows/test-deploy.yaml
FranjoMindek: We need `workflow_call` for it to be callable in `waspc-realse.yaml`, no?

> One big con of doing that: Fly or Railway go down - we can't release since the tests can't pass.

This is a risk I would be willing to take.
Our (almost) whole infra relies on Fly, so why wouldn't our release process too.
If we don't feel comfortable having this, I wouldn't host our apps there.

All in all, I would either run it only before release, or also on main. Not more.
(most important thing is not to screw up the release)

.github/workflows/test-deploy.yaml
FranjoMindek: Hm, I usually don't mind it. Its something like a function declaration to me.
I know what we want to get out of steps, and then I can go into the steps logic knowing the end goal.

.github/workflows/test-deploy.yaml
FranjoMindek: TIL, I always used lowercase for local and uppercase for global (at least when I remembered).

.github/workflows/test-deploy.yaml
FranjoMindek: I'd remove this extra job, Carlos's suggestion is good.

.github/workflows/test-deploy.yaml
cprecioso: > We need workflow_call for it to be callable in waspc-realse.yaml, no?

yes

.github/workflows/test-deploy.yaml
sodic: I added `grep 'Todo App'` and `jq '.json'`.

Wasn't too sure about this as it couples our RPC and app title with the test, but I think it's alright.

.github/workflows/test-deploy.yaml
sodic: > fly_test

The testing job doesn't mention fly on purpose because I wanted to use it for both fly and railway. Does that make sense?

I can rename the others.

.github/workflows/test-deploy.yaml
sodic: Ok, thanks guys!

.github/workflows/test-deploy.yaml
sodic: Haha, a comedian

.github/workflows/test-deploy.yaml
sodic: > Does this command fail or bail if the app doesn't exist?

Nice catch, it fails and leaves the other alive.

.github/workflows/test-deploy.yaml
sodic: I agree @FranjoMindek, but I moved it already and it's not that important. So I think I'll just keep it at the end :)

.github/workflows/test-deploy.yaml
sodic: Yeah, thanks. I removed it. But I don't get it, why do we have it there then? It's also unnecessary?

sodic:
@cprecioso @infomiho @FranjoMindek I think I've addressed everything we discussed.

You can take another look.

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,120 @@
...
+            | sed -E 's/^/--server-secret /' \
+            | paste -sd' '
+          )
+
+          echo "Deploying with prefix: $APP_PREFIX"
+          # NOTE:
+          # - The `yes` command is necessary because the `fly launch` command prompts for confirmation.
+          # - The `eval` is forces the shell to interpret `$ENV_VARS` as multiple arguments rather 
+          #   than pass it as a single string value.
+          yes | eval wasp-cli deploy fly launch "$APP_PREFIX" "$FLY_REGION" --org wasp-testing $ENV_VARS
```
cprecioso: Are you sure about the `eval` thing? When I run shellcheck through this script it tells me that if it isn't doublequoted, it will split words [(SC2086)](https://www.shellcheck.net/wiki/SC2086), so it seems unneeded.

Anyhow, to make it more robust, I'd use arrays and forego the need for `eval`. I was able to wrangle it into this:

```suggestion
          mapfile -t ENV_VAR_ARGUMENTS < <(
            cat .env.server.headless |
              sed -E 's/#.*//' |
              awk 'NF {$1=$1;print}' |
              sed -E 's/^/--server-secret\n/'
          )

          echo "Deploying with prefix: $APP_PREFIX"
          # NOTE:
          # - The `yes` command is necessary because the `fly launch` command prompts for confirmation.
          yes | wasp-cli deploy fly launch "$APP_PREFIX" "$FLY_REGION" --org wasp-testing "${ENV_VAR_ARGUMENTS[@]}"

```

cprecioso: Also, Copilot tells me, in my suggestion, we can fold the `sed/awk/sed` into a single `awk` call. I don't think it's very maintainable, but if you want to use it here it is:

```sh
awk 'NF && $0 !~ /^#/ {gsub(/#.*/, ""); print "--server-secret\n"$0}'
```

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,120 @@
...
+    # want to always clean them up.
+    if: always()
+
+    steps:
+      - uses: superfly/flyctl-actions/setup-flyctl@v1
+        with:
+          # NOTE: We pinned the Fly because we don't want the changes in the
+          # Fly CLI to affect our cleanup procedure. `fly destroy` isn't a part
+          # of Wasp, we're just incidentally using the same tool Wasp uses.
+          version: v0.3.164
```
cprecioso, line 114: I like this comment!

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,120 @@
...
+
+      - name: Clean up testing app from Fly.io
+        run: |
+          # NOTE: We are relying on Wasp's naming conventions here
+          flyctl apps destroy -y $APP_PREFIX-server || true
+          flyctl apps destroy -y $APP_PREFIX-client || true
+          flyctl apps destroy -y $APP_PREFIX-db || true
```
cprecioso, line 121: gr8

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,120 @@
...
+
+    runs-on: ubuntu-latest
+
+    environment:
+      name: fly-deploy-test
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
cprecioso: Why? We don't need the `git` history for anything, no? It will slow down the workflow

cprecioso (commented):
Small things

.github/workflows/test-deploy.yaml
FranjoMindek: Oh damn this is something I also did.
We have multiple workflows with this I think.

I saw it somewhere and thought okay, 0 means we fetch only the latest commit, it's like 0 depth?
Now that I've read the docs I see it's completely the opposite. We should fix those.

e.g. I've introduced it to `check-formatting.yaml`

infomiho: TIL about `0` meaning all instead of none in this case.

.github/workflows/test-deploy.yaml
```diff
@@ -0,0 +1,120 @@
...
+          echo "Deploying with prefix: $APP_PREFIX"
+          # NOTE:
+          # - The `yes` command is necessary because the `fly launch` command prompts for confirmation.
+          # - The `eval` is forces the shell to interpret `$ENV_VARS` as multiple arguments rather 
+          #   than pass it as a single string value.
+          yes | eval wasp-cli deploy fly launch "$APP_PREFIX" "$FLY_REGION" --org wasp-testing $ENV_VARS
+
+      - name: Save deployed app hostnames
+        id: save_hostnames
+        working-directory: waspc/examples/todoApp
```
infomiho:
```suggestion
        working-directory: ${{ env.APP_TO_DEPLOY }}
```

.github/workflows/test-deploy.yaml
sodic: > Are you sure about the eval thing? 

Hm, I'm pretty sure, yes. I've tried it with and without eval both remote and locally multiple times. Here's the output from a local test:

image

And here's a [working remote workflow](https://github.com/wasp-lang/wasp/actions/runs/16680297009/job/47217066936), and a [non-working one](https://github.com/wasp-lang/wasp/actions/runs/16679849501/job/47215592910) I had right before it.

> When I run shellcheck through this script it tells me that if it isn't doublequoted, it will split words [(SC2086)](https://www.shellcheck.net/wiki/SC2086), so it seems unneeded.

I'm reading the rule you linked but it seems to say the opposite:

image

> Anyhow, to make it more robust, I'd use arrays and forego the need for eval

I thought about using bash arrays but they seemed to exotic compared to `eval`. That said, I've since thought of a way `eval` might be dangerours in this situation so I might change it (more on this on Discord :)).

EDIT: Done, we now use arrays.

.github/workflows/test-deploy.yaml
sodic: @FranjoMindek Yeah, that's where I copied it from :smile: 

Anyway, I'll remove it here and I'll create a new PR where I remove it from everywhere else if that's fine with you.

.github/workflows/test-deploy.yaml
sodic: Lol thanks. Good thing I got the variable out...

.github/workflows/test-deploy.yaml
sodic: > Also, Copilot tells me, in my suggestion, we can fold the sed/awk/sed into a single awk call.

Oh, and yes, it suggested the same to me as well but I decided to keep the pipeline steps as simple as possible for maintainability (just like you said).

.github/workflows/test-deploy.yaml
```diff
@@ -7,6 +7,7 @@ on:
       - ".github/**"
     branches:
       - main
+      - filip-automatic-deploy-test-fly
```
cprecioso: Remeber to take this out

cprecioso (approved):
let's go!

PR #2999, opened 2025-07-24
Fixes App declaration in Wasp TS Spec

### LINKED ISSUE #2991: When using main.wasp.ts v0.17.0 shows "You need to enable JavaScript to run this app."
**Describe the bug**
When on v0.17.0, using a TS main.wasp.ts file seems to be breaking the app. I see the `<noscript>You need to enable JavaScript to run this app.</noscript>` message in the source. 

The way I got to know this is when I was migrating from  v0.16.0 to v0.17.0 and then I migrated to using the TS version instead of Wasp DSL. The migration to v0.17.0 worked as expected but when I swapped to TS version of the file, the website seemed to have no styling and the source HTML had the noscript tag. All of the app was already written in TS. 

I then downgraded to using v0.16.7 and everything is working as expected. 

**To Reproduce**

1. Create an app on v.0.16.0
2. Upgrade to v0.17.0
3. Update to use the main.wasp.ts file following instructions [here](https://wasp.sh/docs/general/wasp-ts-config) 
4. Wasp start and check if the app is loading normally.
5. Downgrade to v.0.16.7 (also need to download the version `curl -sSL https://get.wasp.sh/installer.sh | sh -s -- -v 0.16.7` and update the TS version in package.json to `"typescript": "^5.1.0",`)
**Expected behavior**
The website should load normally even on v0.17.0 

**Screenshots**

- OS: macOs Sequoia 15.5 (24F74)
- Browser: Chrome tried even incognito
- Version: Version 138.0.7204.158 (Official Build) (arm64)

**Smartphone (please complete the following information):**

**Additional context**

infomiho:
> [!NOTE]
> Note for reviewers, review only the `Fixes App declaration in Wasp TS Spec` commit to review the fix. The other commits are preparing the 0.17.1 version.

### Context

Given this `main.wasp.ts`:
```ts
import { App } from "wasp-config";

const app = new App("minimalRootCompnent", {
  title: "minimal-root-compnent",
  wasp: { version: "^0.17.0" },
  head: ["<link rel='icon' href='/favicon.ico' />"],
});

app.client({
  rootComponent: { import: "App", from: "@src/App" },
});

const mainPage = app.page("MainPage", {
  component: { import: "MainPage", from: "@src/MainPage" },
});

app.route("RootRoute", { path: "/", to: mainPage });

export default app;
```
We get this as `decls.json` output:
```json
[
  {
    "declType": "App",
    "declName": "minimalRootCompnent",
    "declValue": {
      "wasp": { "version": "^0.17.0" },
      "title": "minimal-root-compnent",
      "head": ["<link rel='icon' href='/favicon.ico' />"],
      "server": {}
    }
  },
  {
    "declType": "Page",
    "declName": "MainPage",
    "declValue": {
      "component": {
        "kind": "named",
        "name": "MainPage",
        "path": "@src/MainPage"
      }
    }
  },
  {
    "declType": "Route",
    "declName": "RootRoute",
    "declValue": {
      "path": "/",
      "to": { "name": "MainPage", "declType": "Page" }
    }
  }
]
```

You'll notice that the `client`  is missing from the `App` declaration. This will then mess the app generation process and user's app will not have the e.g. `rootComponent` defined.

### Investigation

I started logging stuff and it seems that we have the `client` value until we enter the `mapApp` function:
```
CLIENT mapTsAppSpecToAppSpecDecls {"rootComponent":{"import":"App","from":"@src/App"}}
CLIENT mapApp undefined
CLIENT appDecl undefined
```

### Issue

We are passing both `server` and `client` to the `mapApp` as positional arguments **but we send them in the wrong order**. They are structurally the same so Typescript doesn't complain.

```diff
ig/src/mapTsAppSpecToAppSpecDecls.ts
index 17bfa269e..f53524d08 100644
--- a/waspc/packages/wasp-config/src/mapTsAppSpecToAppSpecDecls.ts
+++ b/waspc/packages/wasp-config/src/mapTsAppSpecToAppSpecDecls.ts
@@ -68,8 +68,8 @@ export function mapTsAppSpecToAppSpecDecls(
       entityRefParser,
       routeRefParser,
       auth,
-      server,
       client,
+      server,
       db,
       emailSender,
       websocket,
```

Swapping the position of the arguments fixed the issue. 

### This PR

In this PR, I did the minimal fix of swapping `client` and `server` positions. We'll refactor the code not to use positional arguments later on.

Fixes #2991


waspc/packages/wasp-config/src/mapTsAppSpecToAppSpecDecls.ts
```diff
@@ -181,18 +181,28 @@ export function mapApiNamespace(
...
-  entityRefParser: RefParser<"Entity">,
-  routeRefParser: RefParser<"Route">,
-  auth?: TsAppSpec.AuthConfig,
-  client?: TsAppSpec.ClientConfig,
-  server?: TsAppSpec.ServerConfig,
-  db?: TsAppSpec.DbConfig,
-  emailSender?: TsAppSpec.EmailSenderConfig,
-  webSocket?: TsAppSpec.WebsocketConfig,
-): AppSpec.App {
+export function mapApp({
```
infomiho: I've replaced positional arguments with one object argument.

waspc/packages/wasp-config/__tests__/mapTsAppSpecToAppSpecDecls.integration.test.ts
```diff
@@ -136,17 +136,17 @@ describe("mapTsAppSpecToAppSpecDecls", () => {
...
         routeRefParser,
-        tsAppSpec.auth,
-        tsAppSpec.server,
-        tsAppSpec.client,
-        tsAppSpec.db,
-        tsAppSpec.emailSender,
-        tsAppSpec.websocket,
-      ),
+        auth: tsAppSpec.auth,
+        server: tsAppSpec.server,
```
FranjoMindek: Nitpick, but could we order the keys in all usage locations in the same order? Something which makes logical sense.

Even though now we have named keys.

waspc/packages/wasp-config/src/mapTsAppSpecToAppSpecDecls.ts
```diff
@@ -63,17 +63,17 @@ export function mapTsAppSpecToAppSpecDecls(
...
       entityRefParser,
       routeRefParser,
       auth,
       server,
       client,
       db,
       emailSender,
-      websocket,
-    ),
+      webSocket: websocket,
```
infomiho: We should probably refactor the `tsAppSpec` to name this field `webSocket` as well, but I didn't want to do it in this PR.

waspc/packages/wasp-config/__tests__/mapTsAppSpecToAppSpecDecls.integration.test.ts
FranjoMindek: Alternative:

Eslint rule:
```ts
  "sort-keys": ["error", "asc", { "natural": true, "minKeys": 2 }]
  ```

Comment (FranjoMindek) [reply waspc/packages/wasp-config/__tests__/mapTsAppSpecToAppSpecDecls.integration.test.ts]: This will though pollute the PR with formatting.

File: waspc/packages/wasp-config/src/mapTsAppSpecToAppSpecDecls.ts (hunk 3 above)
Comment (FranjoMindek) [reply waspc/packages/wasp-config/src/mapTsAppSpecToAppSpecDecls.ts]: Not only that, but change the `WebsocketConfig` type name too.
All name instances of WebSockets should be checked.

File: waspc/packages/wasp-config/__tests__/mapTsAppSpecToAppSpecDecls.integration.test.ts (hunk 2 above)
Comment (infomiho) [reply waspc/packages/wasp-config/__tests__/mapTsAppSpecToAppSpecDecls.integration.test.ts]: I've applied the same ordering as it was before, I think this is out of scope for this PR - but we can create an issue to add that formatting rule and auto-format the code in a separate PR.

Comment (FranjoMindek) [reply waspc/packages/wasp-config/__tests__/mapTsAppSpecToAppSpecDecls.integration.test.ts]: I meant this.
We have some examples:
```ts
	server: tsAppSpec.server,
	client: tsAppSpec.client,
```
Some:
```ts
	client: tsAppSpec.client,
	server: tsAppSpec.server,
```

Either make all client first or all server first.
This is just re-ordering at 2-3 places.

sodic (changes requested):
Can we just do a simple fix for now, reorder the server and the client to be correct, and add a test that checks for it?

This fix also includes a refactor, which I like, but I'd also like to change some things (e.g., ordering of stuff, the websocket thing, not bundling app arguments with parsers in the same object, plus other stuff we possibly haven't anticipated). And I don't want those to hold off the fix.

So, I'd like us to just implement a simple one line fix, update the tests, and create an issue for the refactoring. IMO hotfixes should be as small as possible.

infomiho:
@sodic I'll modify the fix to be as simple as possible, but can you please create the issue with all the refactoring you mentioned.

FranjoMindek:
> This fix also includes a refactor, which I like, but I'd also like to change some things (e.g., ordering of stuff, the websocket thing, not bundling app arguments with parsers in the same object, plus other stuff we possibly haven't anticipated). And I don't want those to hold off the fix.

Makes sense.

I'll assume that you've tested that the new minimal approach has correct App Spec JSON outputs.
Feel free to merge once Filip approves.

Issue #2961, opened 2025-07-15
Run formatting workflows only when their respective files are changed


FranjoMindek:
Currently, we run the `check-formatting.yaml` workflow on each pull request.
We always run both `prettier` and `ormolu`.

It might be worthwhile to split those formatters and run each only when their respective files change. e.g. `waspc/**/*.hs` for `ormolu`.

Martinsos:
We use `dx` label for developer experience of people using Wasp, not our own, so I removed it from this PR.

PR #3000, opened 2025-07-24
Run ormolu only when haskell files change

FranjoMindek:
Fixes: #2961 

It's easy to detect Haskell file changes for `ormolu`.

`prettier` is a bit larger problem, because it works with A LOT of different file extensions.
I think it's easier to just always run `prettier` since it is very fast.

### DIFF
--- a/.github/workflows/check-formatting.yaml
+++ b/.github/workflows/check-formatting.yaml
@@ -15,11 +15,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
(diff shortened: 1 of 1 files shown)


infomiho (approved):
LGTM

I'd go with Option 1 (not doing anything) for Prettier since it's fast enough.

Martinsos (approved):
LGTM! And I agree to just let prettier be unless it becomes too slow for us for real.

I wish we didn't have to repeat that condition `if: steps.haskell-changes.outputs.run == 'true'` twice, but I don't know of a better way either, and extracting into separate files and stuff seems redunadnt.

PR #3007, opened 2025-07-25
fix - Improves UX for auth signin failing due to account already existing

### LINKED ISSUE #3002: Consider improving DX of Auth signin failing due to account already existing
If you try to sign in via Wasp Auth and email is already taken, we won't tell you that, for security reasons (we don't want anybody to be able to learn which accounts exist), instead we will tell you to go check your email for verification link, but email will never arrive.

Similar with login, we won't tell you if email is non-existant, but that a combo of email and pass is wrong.

While this is good security-wise, it is not great DX wise, if you for example forgot which email you are using in the app, of if you forgot that you already have an account.

We should look into ways to possibly improve DX without loosing on the security side. Maybe we could send emails in the background to the existing account to provide more info (hey you used wrong pass). Maybe we can leave it to the Wasp app to customize the error message, provide more info if they want so.
We should also look into what other popular apps are doing, maybe we are going overboard here?

naaa760:
### Description

Fixes #3002 - Improves UX for auth signin failing due to account already existing.

When users try to sign up with an email that already exists, Wasp now sends a helpful email to guide them instead of just pretending the signup succeeded. The frontend message is also updated to be more generic and user-friendly.

**Changes:**
- Backend: Send "account already exists" email when signup attempted with verified email
- Frontend: Update signup success message to "If you have an account, you'll receive an email with further instructions"

1. [x] I updated [`ChangeLog.md`](https://github.com/wasp-lang/wasp/blob/main/waspc/ChangeLog.md) with description of the change this PR introduces.
2. [x] I bumped `waspc` version in [`waspc.cabal`](https://github.com/wasp-lang/wasp/blob/main/waspc/waspc.cabal) to reflect changes I introduced, with regards to the version of the latest wasp release, if the bump was needed.

### DIFF
--- a/waspc/data/Generator/templates/sdk/wasp/auth/forms/internal/common/LoginSignupForm.tsx
+++ b/waspc/data/Generator/templates/sdk/wasp/auth/forms/internal/common/LoginSignupForm.tsx
@@ -172,7 +172,7 @@ export const LoginSignupForm = ({
     onError: onErrorHandler,
     showEmailVerificationPending() {
       hookForm.reset()
-      setSuccessMessage(`You've signed up successfully! Check your email for the confirmation link.`)
+      setSuccessMessage('If you have an account, you’ll receive an email with further instructions.')
     },
     onLoginSuccess() {
       navigate('{= onAuthSucceededRedirectTo =}')
--- a/waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
+++ b/waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
@@ -23,8 +23,10 @@ import {
   isEmailResendAllowed,
   sendEmailVerificationEmail,
 } from 'wasp/server/auth/email/utils'
-import { EmailFromField } from 'wasp/server/email/core/types'
+import { EmailFromField } from 'wasp/server/email'
 import { onAfterSignupHook, onBeforeSignupHook } from '../../hooks.js'
+import { Email } from 'wasp/server/email'
+import { emailSender } from 'wasp/server/email'
 
 export function getSignupRoute({
   userSignupFields,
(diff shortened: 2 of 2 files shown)


waspc/data/Generator/templates/sdk/wasp/auth/forms/internal/common/LoginSignupForm.tsx
```diff
@@ -107,7 +172,7 @@ export const LoginSignupForm = ({
     onError: onErrorHandler,
     showEmailVerificationPending() {
       hookForm.reset()
-      setSuccessMessage(`You've signed up successfully! Check your email for the confirmation link.`)
+      setSuccessMessage('If you have an account, you’ll receive an email with further instructions.')
```
cprecioso: Let's undo this change, as users who don't have an account might get confused

waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
```diff
@@ -82,6 +84,19 @@ export function getSignupRoute({
...
       // the email!
       if (providerData.isEmailVerified) {
+        // Send a generic 'account already exists' email to the user.
+        try {
+          await sendAccountAlreadyExistsEmail(fields.email, {
+            from: fromField,
+            to: fields.email,
+            subject: 'Account Already Exists',
+            text: 'It looks like you (or someone else) tried to sign up with this email. If you already have an account, you can log in or reset your password. If this wasn’t you, you can ignore this email.',
+            html: '<p>It looks like you (or someone else) tried to sign up with this email. If you already have an account, you can <a href="/login">log in</a> or <a href="/reset-password">reset your password</a>. If this wasn’t you, you can ignore this email.</p>'
```
cprecioso: These links need to be full links with the domain.

waspc/data/Generator/templates/server/src/auth/providers/email/signup.ts
```diff
@@ -163,3 +178,11 @@ function ensureValidArgs(args: object): void {
...
   ensureValidPassword(args)
 }
+
+async function sendAccountAlreadyExistsEmail(email: string, content: Email): Promise<void> {
+  // Use the same email sending utility as verification emails, but do not update verification metadata.
+  // This avoids leaking info about verification status.
+  emailSender.send(content).catch((e) => {
+    console.error('Failed to send account already exists email', e);
+  });
+}
```
cprecioso: Maybe we can use the same utility for email sending but adding an option for saving the metadata or not

cprecioso (changes requested):

cprecioso:
@naaa760 I sent a review, are you able to address it?

nehaaprasad:
Hello @cprecioso Please check it out, I’ve addressed the issues you raised.
and sorry for being a little late!

PR #3011, opened 2025-07-28
Remove old header

cprecioso:
We don't want to get contributors or ourselves confused when adding changes, we're adding them on top


infomiho (approved):
Oh yeah, I saw that someone added stuff on the bottom, good catch!

PR #3015, opened 2025-07-29
Move to Megaparsec for PSL

cprecioso:
- Part of #2132 
- Closes #2985 (this is another -better- approach to get the same result)

Use Megaparsec instead of Parsec for PSL. Migration is more or less straight-forward. It allows the triple-slash comment case

The only big differences are in `Wasp.Psl.Parser.Common`, as Megaparsec does not include such a shorthand lexer approach like Parsec does, so it is a bit more manual, but nothing crazy. I took a look at some resources ([Parsec's internal lexers](https://hackage.haskell.org/package/parsec-3.1.18.0/docs/src/Text.Parsec.Token.html#makeTokenParser), Megaparsec's [tutorial](https://markkarpov.com/tutorial/megaparsec.html) and [documentation](https://hackage.haskell.org/package/megaparsec-9.2.1) and [this gist](https://gist.github.com/LightAndLight/36926a4e3a7133910d9aa199da50c4fd)), and implemented the lexers without much problem.

Also errors are nicer now:

**Before**

```

- Wasp couldn't parse your schema.prisma file, please check if you have any errors in it.

(line 15, column 60):
unexpected "#"
expecting identifier, "@" or "}"
```

**After**

```

- Wasp couldn't parse your schema.prisma file, please check if you have any errors in it.

15:60:
   |
15 |   isOnAfterSignupHookCalled        Boolean @default(false) #invalid
   |                                                            ^
unexpected '#'
expecting '@', '}', or identifier
```


waspc/src/Wasp/Project/Analyze.hs
```diff
@@ -128,7 +128,7 @@ analyzePrismaSchema waspProjectDir = do
 
       case Psl.Parser.parsePrismaSchema prismaSchemaContent of
         Left err ->
-          return (Left [couldntParsePrismaSchemaMessage ++ "\n\n" ++ show err], [])
+          return (Left [couldntParsePrismaSchemaMessage ++ "\n\n" ++ err], [])
```
cprecioso, line 131: This `show` here is not needed as `err` is already always a `String`

waspc/src/Wasp/Analyzer/Evaluator/EvaluationError.hs
```diff
@@ -53,6 +53,8 @@ data EvalErrorCtx
 data EvaluationParseError
   = -- | In case when evaluation includes parsing with Parsec and it fails.
     EvaluationParseErrorParsec Text.Parsec.ParseError
+  | -- | In case when evaluation includes parsing with Megaparsec and it fails.
+    EvaluationParseErrorMegaparsec String
```
cprecioso: Errors are already [prettified by Megaparsec](https://markkarpov.com/tutorial/megaparsec.html#displaying-parse-errors) by the time they reach here, as otherwise we'd be having to pass around the internal Psl `Parser` definition.

waspc/src/Wasp/Psl/Parser/Argument.hs
```diff
@@ -70,15 +66,9 @@ arrayExpr =
   Psl.Argument.ArrayExpr
     <$> brackets (commaSep expression)
 
--- NOTE: For now we are not supporting negative numbers.
---   I couldn't figure out from Prisma docs if there could be the case
---   where these numbers could be negative.
---   Same goes for numberIntExpr below.
---   TODO: Probably we should take care of that case.
```
cprecioso, line 77: Megaparsec supports this 🎉

https://github.com/wasp-lang/wasp/blob/f264b6658f39707e3c8229058a93a2e647f0878d/waspc/src/Wasp/Psl/Parser/Common.hs#L96-L104

waspc/waspc.cabal
```diff
@@ -638,6 +639,7 @@ test-suite waspc-test
     , mtl
     , neat-interpolation
     , parsec
+    , megaparsec
```
cprecioso, line 642: AFAIU, we're not setting dependency versions here because we're already transitively depending on `megaparsec` so Cabal will just resolve to the same version. Correct?

waspc/src/Wasp/Analyzer/Evaluator/EvaluationError.hs
```diff
@@ -105,6 +107,7 @@ getErrorMsgAndErrorCtxMsgsAndParsingCtx (EvaluationError (WithCtx ctx evalError)
         ++ "Valid values: "
         ++ intercalate " | " validEnumVariants
   MissingDictField fieldName -> makeMainMsg $ "Missing dictionary field '" ++ fieldName ++ "'"
+  ParseError (EvaluationParseErrorMegaparsec e) -> makeMainMsg ("Parse error:\n" ++ indent 2 e)
```
Martinsos: Super mega nitpick, but below we use `e` for error object, or `msg` for when we already have message, so we can be consistent with that.

```suggestion
  ParseError (EvaluationParseErrorMegaparsec msg) -> makeMainMsg ("Parse error:\n" ++ indent 2 msg)
```

waspc/src/Wasp/Analyzer/Evaluator/EvaluationError.hs
Martinsos: One thought: you could have skipped having this special Megaparsec data constructor and just used `EvaluationParseError`. Becuase obviously it doesn't matter it is Megaparsec error or not when it is just a string anyway, and you could say "megaparsec" is just a detail of "general parsing".
Unless there is some reason we have for treating megaparsed error messsages in a special way.

waspc/src/Wasp/Psl/Parser/Argument.hs
Martinsos: All right, nice!

waspc/src/Wasp/Psl/Parser/Common.hs
```diff
@@ -13,66 +13,105 @@ module Wasp.Psl.Parser.Common
...
+type SourceCode = String
 
 reserved :: String -> Parser ()
-reserved = T.reserved lexer
+reserved name =
+  lexeme $
+    try $
+      do
+        _ <- C.string name
+        notFollowedBy identLetter <?> ("end of " ++ show name)
```
Martinsos: I am not 100% sure redundant parenthesses, but this should work fine.

```suggestion
reserved name =
  lexeme $ try $ C.string name >> (notFollowedBy identLetter <?> ("end of " ++ show name))
```

Martinsos: Hm intersting that we have to impelment this on our own! But you said it is so. What is "end of ", that is error message?

waspc/src/Wasp/Psl/Parser/Common.hs
```diff
@@ -13,66 +13,105 @@ module Wasp.Psl.Parser.Common
...
 identifier :: Parser String
-identifier = T.identifier lexer
+identifier = lexeme $ try ident
+  where
+    ident =
+      do
+        c <- identStart
+        cs <- many identLetter
+        return (c : cs)
+        <?> "identifier"
```
Martinsos: Why not like this (simpler)?

```suggestion
identifier =
  lexeme $ try $ do
    c <- identStart
    cs <- many identLetter
    return (c : cs) <?> "identifier"
```

Martinsos: I also put that `<?> "identifier"` in same line with return, but maybe I don't know what that does hm. Why did you put it in the last line alone?
EDIT: Ok I read that is a way to label stuff to get nicer error messages. Sounds nicer to do it as I did in suggestion above?

waspc/src/Wasp/Psl/Parser/Common.hs
Martinsos, line 1: It is cool that we have this flexibility, but it also feels a bit much having to define all these things on our own. But ok!

Are we sure tehre are no already ready definitions for some of these functions, like parsec had them? I think I read there is an accompanying library to megaparsec that brings some extra stuff, but I couldn't find it now with quick search and I can't remmeber where I read that. Maybe I got somethign wrong.

Feel free to resolve if there is no better option (and I have a feeling there isn't but wanted to check still).

waspc/src/Wasp/Psl/Parser/Model.hs
```diff
@@ -32,8 +27,8 @@ import Wasp.Psl.Parser.Common
 -- NOTE: We need to consume the leading whitespace specifically here, because we use the `body`
 -- parser directly (meaning not as part of parsing the whole schema) which means that the
 -- leading whitespace is not consumed by the `schema` parser.
-parseBody :: SourceCode -> Either Parsec.ParseError Psl.Model.Body
-parseBody = Parsec.parse (whiteSpace >> body) ""
+parseBody :: SourceCode -> Either String Psl.Model.Body
+parseBody = left errorBundlePretty . Megaparsec.parse (whiteSpace >> body) ""
```
Martinsos, line 31: Ok, this is where we prettify, makes sense.

waspc/src/Wasp/Psl/Parser/Model.hs
```diff
@@ -52,7 +47,7 @@ model = do
 -- which is everything besides model keyword, name and braces:
 --   `model User { <body> }`.
 body :: Parser Psl.Model.Body
-body = Psl.Model.Body <$> many1 element
+body = Psl.Model.Body <$> some element
```
Martinsos, line 50: Aha `some` is alternative to `many1`, ok makes sense. I think `many1` requirest at least one element, but I guess so does `some` or it would be named `any` if it didn't :D.

waspc/waspc.cabal
Martinsos: Yup, it will use the version from the `library waspc`!

Martinsos (approved):
LGTM! Some small comments, but approved, merged once you are happy with it.

waspc/src/Wasp/Psl/Parser/ConfigBlock.hs
```diff
@@ -47,6 +44,6 @@ configBlockBody = braces (many keyValuePair)
 keyValuePair :: Parser Psl.ConfigBlock.KeyValuePair
 keyValuePair = do
   key <- identifier
-  reserved "="
+  _ <- symbol "="
```
infomiho, line 47: What does this change mean? I knew before the difference between `reserved` and `symbol` but now I'm not so sure.

waspc/src/Wasp/Psl/Parser/Common.hs
```diff
@@ -13,66 +13,105 @@ module Wasp.Psl.Parser.Common
...
+
+whiteSpace :: Parser ()
+whiteSpace =
+  L.space (void C.spaceChar) (void lineComment) empty
+
+lineComment :: Parser String
+lineComment =
+  C.string "//"
+    >> notFollowedBy (C.char '/')
+    >> takeWhileP (Just "character") (/= '\n')
```
infomiho, line 109: What does this `(Just "character")` mean here?

infomiho (approved):
Left some questions, but LGTM if the unit tests are passing, that should be it

waspc/src/Wasp/Analyzer/Evaluator/EvaluationError.hs
cprecioso: Resolving as I removed the `EvaluationParseErrorMegaparsec` constructor

waspc/src/Wasp/Psl/Parser/Common.hs
cprecioso: [`takeWhileP` takes an optional name for the kind of chunk it is consuming](https://hackage.haskell.org/package/megaparsec-9.2.1/docs/Text-Megaparsec.html#v:takeWhileP). It can have no name (`Nothing`) or a name (`Just "whatever"`). This helps with error messages.
I took and adapated this code from Megaparsec's own `skipLineComment` [(source)](https://hackage.haskell.org/package/megaparsec-9.2.1/docs/src/Text.Megaparsec.Char.Lexer.html#skipLineComment) which is what they point you to use if we didn't have the `//` vs `///` issue.

waspc/src/Wasp/Psl/Parser/Common.hs
cprecioso: > What is "end of ", that is error message?

It's the name of what you're matching for, will get shown in error messages. It has an alias called `label` which I think better communicates what it does. The code is copied/adapted from Parsec which is where it gets the style from, which I guess is a bit weird.

waspc/src/Wasp/Psl/Parser/Common.hs
cprecioso: I rewrote this better IMO

```hs
identifier :: Parser String
identifier =
  lexeme $ try (ident <?> "identifier")
  where
    ident = liftA2 (:) identHead identTail
    identHead = C.letterChar
    identTail = many identLetter
```

The last line `<?>` weirdness is due to it being like that in Parsec's source

waspc/src/Wasp/Psl/Parser/Common.hs
cprecioso: Yeah I feel you, but I think it's contained enough to not be a big bother. I tried looking for something better but didn't find anything interesting.
It seems Megaparsec actually wants you to use [`alex`](https://hackage.haskell.org/package/alex) with BNF parsers, that's why it has relatively minimal lexer facilities.

waspc/src/Wasp/Psl/Parser/Model.hs
cprecioso: Yep, `many` for 0 or more and `some` for 1 or more, for symmetry with `Control.Applicative`. There's no `many1` function in Megaparsec.

waspc/src/Wasp/Psl/Parser/ConfigBlock.hs
cprecioso: `reserved` is a `symbol` but also checks that is not followed by identifier letters.
Basically so that it would match `throw` (a reserved keyword) but not e.g. `throwMyError` (an identifier name).

This is not a problem for actual symbols as they can't be prefixes of identifiers. AKA if we write (badly formatted) `a =b`, we want it to be lexed as `IDENT(a) EQUALS IDENT(b)`. But if we had it as `reserved`, it would backtrack and not parse.

waspc/src/Wasp/Psl/Parser/Common.hs
Martinsos: Nice! I would maaaybe inline the content of `ident`, but `identHead` and `identTail` are nice. But it's great.

Issue #2995, opened 2025-07-23
Make our headless test work with production apps


sodic:
Our headless tests for the kitchen sink app rely heavily on the app being deployed locally (e.g., hardcoded localhost links, couping with `wasp app runner`, reliance on `SKIP_EMAIL_VERIFICATION_IN_DEV`).

We want to test our deployments using these tests (#2665). To do that, we must adapt them to work with remote apps, not only local ones.

FranjoMindek:
I wouldn't remove `wasp-app-runner` coupling.
Doesn't seem to be issue here. Or am I missing something?

sodic:
What I meant is "The tests always attempt to start a local app with `wasp-app-runner` which doesn't make sense when testing production apps".

cprecioso:
I'd love to be able to package an app as a single Docker image (#3102) so that we could use [GitHub Action's `service` option](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idservices) to test either local or remote apps in the same way.

cprecioso:
I'm putting this back on triage for the time being. I don't feel that the trade-off of having to add escape hatches for testing (and enabling them on production code) is "correct". Especially when, online, I'm not seeing much mention of best practices for doing more than smoke testing on the production deployment.

My reasoning is: if this was a more "serious" app (imagine e.g. Microinfluencer from Cam), would be creating, modifying and deleting entities on production? Would have code running on production to e.g. skip a rate-limiter, or allow emails to bare IP addresses?

My gut says "no" so I'd like for us to re-evaluate.

Martinsos:
No more than smoke tests on the production app -> yeah, makes sense, unless there are some special requirements.

But, what you could and probably should do, when you have an app, is run tests in staging environment, which emulates the production app.

So that is what we want to do: we want to deploy the example app to a staging environment, run the e2e tests there, then tear it down.

We are already doing half of it, when we are testing the deployment scripts -> we are uploading them to some kind of staging environment, we check it works kind of, then we tear it down. I guess we just don't do all the e2e tests but just a smoke test, right? But it might make sense to run all the tests.

I am ok with postponing this task, till we get a better feeling if we really need this or not. Because right now we fully test in dev mode, we fully test the output of `wasp build`, and we smoke test the deployed thing (right? Hm do we?) (staging env). That might be enough. If it turns out it is not enough, we can add full(er) tests for staging env.

sodic:
[...]

Issue #1963, opened 2024-04-16
Find a way to test the CLI


sodic:
Figure out a way to test Wasp's CLI. We want to prevent something like #1787 from happening again.

Also, figure out why #1787 happened in the first place and try to account for it in our processes.

Martinsos:
We could start with some basic tests, like just confirming that `wasp version`, `wasp new`, `wasp reset` and similar behave as we expect them to (check that they run, some basic output, ...).

Martinsos:
Btw, this would probably fall under the machinery that we already have for CLI e2e tests, so called "golden" tests, right? Right now there we are running CLI on example projects and then we check the output in .wasp/out and do snapshots of it. But that is actually testing the CLI -> just focused on its .wasp/out output. So we could expand those tests with also checking the actual stdout and exit code. And all together, it is `CLI e2e tests`.

FranjoMindek:
Current situation:
We only have snapshot tests as part of our `waspc e2e tests`.
While they do test few CLI commands along the way (`wasp new`, `wasp compile`, `wasp migrate`, `wasp build`), their main purpose is to test "Wasp app" outputs of our Wasp CLI.

My current findings:
Besides the current snapshot tests, to fully test the Wasp CLI we need 2 additional test variants:

###  1. Ephemeral tests

Think of these as plain e2e tests without any extra actions (e.g. saving snapshots). 
You run the Wasp CLI and verify the results against some expectations. Once the tests is done, all of the outputs are cleaned up, hence the "ephemeral" part of the name.

Most of the currently not tested outputs/commands we want to test can be tested this way. 

As in snapshot tests, we interact with the Wasp CLI through the shell. There is one big difference. In snapshot tests the shell commands are a tool to generating the "current snapshot" files. Shell commands themselves don't assert anything. All of the "testing" is done during the snapshot comparison. The shell commands are essentially a set-up process. That's why we generate a single shell command which does the whole setup.

In ephemeral tests, the shell commands are the test case code itself.
They make up the whole arrange/act/assert part of testing.
There can also be multiple arrange/assert steps.
To make this work we need to simulate that behavior ourselves:
1. **Execute each step of the test case separately**. We no longer a have single shell command. Each step is a separately executed shell command now.
[...]

PR #3018, opened 2025-07-31
Fix naming problems around Wasp's tests

FranjoMindek:
Changes:
- Renamed all `headless-tests` to `e2e-tests`
	- I think that naming them either `e2e` or `headless-e2e` tests works much better than `headless` tests.
- Renamed `HEADLESS_TEST_MODE` to `WASP_RUN_MODE`
	- Conveys it's meaning much better, this just changes how `wasp-app-runner` runs the Wasp project
- Renamed `waspc/test` -> `waspc/tests`
- Renamed `waspc/e2e-test` -> `waspc/e2e-tests`
- Renamed `waspc/cli/test` -> `waspc/cli/tests`
- Renamed `waspc/waspls/test` -> `waspc/waspls/tests`
- Propagated above `test` -> `tests` changes to Haskell files as necessary
- Same for `.gitattributes` and ignore files
- Changed `waspc/README.md` to follow new names
- Changed `waspc/run`
	- Updated with new names above
	- Added command for running Wasp LS tests
	- Added command for running `starters-e2e-tests`
	- Added command for running `examples/**/e2e-tests` (not `waspc/examples/todoApp`) 
	- Made it so generic test command includes `starters-e2e-tests`, `waspls`, and `examples-e2e-tests`
- Merged `.env.server.headless` files into `.env.server.example`

### DIFF
--- a/waspc/README.md
+++ b/waspc/README.md
@@ -156,10 +156,10 @@ When done, new tab in your browser should open and you will see a Todo App!
    - NOTE: You will need to install `ghcid` globally first. You can do it with `cabal install ghcid`.
 3. Do a change in the codebase (most often in `src/` or `cli/src/` or `data/`), and update tests if that makes sense (see [Test](#tests)).
    Fix any errors shown by HLS/`ghcid`.
-   Rinse and repeat. If you're an internal team member, postpone updating snapshot tests until approval (see [here](#note-for-team-members)).
+   Rinse and repeat. If you're an internal team member, postpone updating waspc e2e tests tests until approval (see [here](#note-for-team-members)).
 4. Use `./run build` to build the Haskell/cabal project, and `./run wasp-cli` to both build and run it. If you changed code in `packages/`, you will also need to run `./run build:packages` (check [TypeScript Packages section](#typescript-packages) for more details). Alternatively, you can also run slower `./run build:all` to at the same time build Haskell, TS packages, and any other piece of the project in one command.
 5. For easier manual testing of the changes you did on a Wasp app, you have the `examples/todoApp` app, which we always keep updated. Also, if you added a new feature, add it to this app (+ tests) if needed. Check its README for more details (including how to run it).
-6. Run `./run test` to confirm that all the tests are passing. If needed, accept changes in the snapshot tests with `./run test:waspc:e2e:accept-all`. Check "Tests" for more info.
+6. Run `./run test` to confirm that all the tests are passing. If needed, accept changes in the waspc e2e tests with `./run test:waspc:e2e:accept-all`. Check "Tests" for more info.
 7. If you did a bug fix, added new feature or did a breaking change, add short info about it to `Changelog.md`. Also, bump version in `waspc.cabal` and `ChangeLog.md` if needed. If you are not sure how to decide which version to go with, check out [how we determine the next version](#determining-next-version).
 8. Create a PR. Keep an eye on CI tests -> Everything must pass. If it doesn't, look into it.
 9. If your PR changes how users(Waspers) use Wasp, make sure to also update the documentation, which is in this same repo, but under `/web/docs`.
@@ -168,15 +168,15 @@ When done, new tab in your browser should open and you will see a Todo App!
 
 ### Note for team members
 
-Do not update snapshot tests until your PR is approved.
+Do not update waspc e2e tests until your PR is approved.
 
-Accepting snapshot tests prematurely:
+Accepting waspc e2e tests prematurely:
 
 - Slows down the UI.
 - Creates noise for the reviewers.
 - Makes it less likely that you'll thoroughly check the final diffs after all review iterations.
 
-**Carefully review snapshot test diffs before accepting them** (see [Waspc e2e tests](#waspc-e2e-tests) for more details).
+**Carefully review waspc e2e tests diffs before accepting them** (see [Waspc e2e tests](#waspc-e2e-tests) for more details).
 
 ## Design docs (aka RFCs)
 


infomiho:
We already have `e2e-test` in the `waspc` dir - should we then rename them as well? @cprecioso like using the name Snapshot Tests?

cprecioso:
Yep, makes sense!

FranjoMindek:
Added changes.
Will need to test it out though.

Edit: working now, had 1 typo

waspc/run
```diff
@@ -177,11 +177,11 @@ case $COMMAND in
   test:cli)
     echo_and_eval "$TEST_CLI_CMD"
     ;;
-  test:e2e)
-    echo_and_eval "$TEST_E2E_CMD"
+  test:snapshot)
+    echo_and_eval "$TEST_SNAPSHOT_CMD"
     ;;
-  test:e2e:accept-all)
```
infomiho, line 183: End of an era :D

waspc/run
```diff
@@ -90,17 +90,17 @@ print_usage() {
     "Installs the wasp executable as 'wasp-cli' globally on the machine."
   echo ""
   print_usage_cmd "test" \
-    "Executes all tests (unit + e2e + headless). Builds the project first if needed."
+    "Executes all tests (unit + e2e snapshot + e2e headless). Builds the project first if needed."
```
infomiho:
```suggestion
    "Executes all tests (unit + snapshot + e2e). Builds the project first if needed."
```

Do you want to go all in on naming them `snapshot` vs. `e2e` here?

infomiho: Old e2e tests -> snapshot tests
Headless tests -> e2e tests

waspc/README.md
```diff
@@ -277,23 +277,25 @@ To run individual unit test, you can do `./run test:unit "Some test description
...
+To run e2e snapshot tests only, you can do `/run test:snapshot`.
 
-To run Playwright/headless tests on the example apps, you can do `/run test:headless`.
+To run Playwright e2e headless tests on the example apps, you can do `/run test:headless`.
 
-### End-to-end (e2e) tests
+### Snapshot tests
 
-Besides unit tests, we have e2e tests that run `waspc` on a couple of prepared projects, check that they successfully run, and also compare generated code with the expected generated code (golden output).
+Besides unit tests, we have e2e snapshot tests that run `waspc` on a couple of prepared projects, check that they successfully run, and also compare generated code with the expected generated code (golden output).
```
infomiho:
```suggestion
Besides unit tests, we have snapshot tests that run `waspc` on a couple of prepared projects, check that they successfully run, and also compare generated code with the expected generated code (golden output).
```

waspc/README.md
```diff
@@ -93,7 +93,7 @@ If that is the case, relax and feel free to get yourself a cup of coffee! When s
 ./run test
 ```
 
-to ensure all the tests (unit, e2e, ...) are passing.
+to ensure all the tests (unit, integration, e2e) are passing.
```
infomiho: We added `integration` here, what do we mean by that? If it's not one of the categories 

- unit
- snapshot
- e2e

I wouldn't list it here

.github/workflows/waspc-ci.yaml
```diff
@@ -188,7 +188,7 @@ jobs:
           cd examples/todoApp
           # Prepare the environment for the tests
           cp .env.server.headless .env.server
-          HEADLESS_TEST_MODE=build npm run test
+          WASP_RUN_MODE=build npm run test
 
-      - name: Run e2e tests
-        run: cabal test e2e-test
+      - name: Run snapshot tests tests
```
infomiho:
```suggestion
      - name: Run snapshot tests
```

.github/workflows/examples-e2e-tests.yaml
```diff
@@ -58,27 +58,27 @@ jobs:
...
         working-directory: examples/${{ matrix.example }}
         env:
           WASP_CLI_CMD: wasp-cli
-          HEADLESS_TEST_MODE: dev
+          WASP_RUN_MODE: dev
         run: |
           npm run test
 
-      - name: Run Headless BUILD Tests for ${{ matrix.example }}
+      - name: Run hadless E2E BUILD tests for ${{ matrix.example }}
```
infomiho:
```suggestion
      - name: Run BUILD e2e tests for ${{ matrix.example }}
```

.github/workflows/examples-e2e-tests.yaml
```diff
@@ -58,27 +58,27 @@ jobs:
...
-      - name: Extract Database Provider
+      - name: Extract database provider
         id: extract-db-provider
         working-directory: examples/${{ matrix.example }}
         run: |
           DATABASE_PROVIDER=$($GITHUB_WORKSPACE/scripts/get-wasp-database-provider.sh wasp-cli)
           echo "provider=$DATABASE_PROVIDER" >> $GITHUB_OUTPUT
 
-      - name: Run Headless DEV Tests for ${{ matrix.example }}
+      - name: Run headless E2E DEV tests for ${{ matrix.example }}
```
infomiho:
```suggestion
      - name: Run DEV e2e tests for ${{ matrix.example }}
```

infomiho (commented):
I wouldn't mix terminology like `e2e snapshot` tests or `e2e headless` tests.

I'd be really clear on the types of tests we have and have consistent names for them:
- unit
- snapshot
- e2e

Otherwise, a great change! I like the clarity it brings.

waspc/run
FranjoMindek: In our snapshot tests we are doing e2e snapshot tests.
Same as we are doing e2e headless tests in our headless tests.

It feels wrong to call headless tests just "e2e", when we also have other e2e tests.

waspc/README.md
FranjoMindek: I would call snapshot tests a different category from unit/integration/e2e.
I don't know if the parenthesis are needed here.

We do have integration tests in `wasp-config`, though they are not connected to the `run` script.

FranjoMindek:
On the contrary I think we need them.
`e2e` and `snapshot` aren't exclusive. 
Same for `e2e` and `headless`.

We could just as easily do `snapshot` testing on a lower scale e.g. `unit snapshot` tests.
It's important to capture the whole meaning of the tests in as little words as possible.
I think that `e2e snapshot` and `e2e headless` do just that.

.github/workflows/examples-e2e-tests.yaml
```diff
@@ -58,27 +58,27 @@ jobs:
...
         working-directory: examples/${{ matrix.example }}
         env:
           WASP_CLI_CMD: wasp-cli
-          HEADLESS_TEST_MODE: dev
+          WASP_RUN_MODE: dev
         run: |
           npm run test
 
-      - name: Run Headless BUILD Tests for ${{ matrix.example }}
+      - name: Run hadless e2e BUILD tests for ${{ matrix.example }}
```
infomiho:
```suggestion
      - name: Run headless e2e BUILD tests for ${{ matrix.example }}
```

infomiho:
@FranjoMindek Makes sense, I can get behind that logic, but let's then be consistent, in this PR in some places you have [snapshot](https://github.com/wasp-lang/wasp/pull/3018/files#diff-5bdd7be9bf510914ac4ec6c8f50cc6b8abc21d96c72e0e28a027e34160bb164dR193) and in others [e2e snapshot](https://github.com/wasp-lang/wasp/pull/3018/files) naming. I was just looking to define some standard that we'll apply everywhere.

Martinsos:
"snapshot tests" -> those are tests where we test Wasp CLI end to end, don't we, with golden tests
I would consider calling it "CLI e2e tests" then, since that is what they really are.
The fact we are mostly using "snapshots" in there is more of an implementation detail, same like the fact that e2e tests we run on example apps are running in "headless" mode is an implementation detail, and not really what those tests are about.

FranjoMindek:
The idea of the PR was to be more correct with test names.
We often let out part of the "adjectives" so you got the wrong picture.

We have e.g. `examples headless tests`.
This is bad because we are trying to be specific but letting out the important part that we are doing end-to-end testing. Much more important than `headless` part.
If we go being specific route we should go with `examples headless e2e tests`.
If we go general route it should be `examples e2e tests` or `examples tests`.

> "snapshot tests" -> those are tests where we test Wasp CLI end to end, don't we, with golden tests I would consider calling it "CLI e2e tests" then, since that is what they really are.

I wouldn't agree. Yes they use the CLI to generate projects, but all they do is compare the snapshots. This I would call a snapshot test. We do have an issue https://github.com/wasp-lang/wasp/issues/1963#event-14890248261 which might expand current snapshot tests to a more general CLI tests, but currently they are not. 

> The fact we are mostly using "snapshots" in there is more of an implementation detail

I would just the opposite in this case. We are doing e2e on Wasp projects, and the way to do that in Wasp is CLI. If we had something else we would use that. The core idea was to compare snapshots, not to test out the CLI.

> same like the fact that e2e tests we run on example apps are running in "headless" mode is an implementation detail, and not really what those tests are about.

Yes! And that is way I would like to change it.
That is why `examples headless tests` is bad. We are doing e2e testing, and how is an implementation detail.
Going full `examples headless e2e tests` might be too much and we can tone it down.

Martinsos:
@FranjoMindek I agree with the part of the "example e2e tests", but I don't follow the other part with the CLI tests, it seems to me like you are using one approach in the first one and the opposite approach in the second one.

I get why you drop "headless" in "example e2e tests" -> because it is a detail. To be honest I barely understand what "headless" even means in this context, it was alwasy confusing to me.

What are those tests? Those are e2e tests for our example apps. So "example apps e2e tests" is a good name. It seems we agree on that.

Then, we have CLI unit tests, those seems we agree on -> those are CLI unit tests, we didnt' even bother discussin those now.

And then we have CLI e2e tests. Those are the "golden/snapshot" tests we have in the folder waspc/e2e-test. Those are e2e tests for waspc. waspc is CLI. The fact that all or most of them are currently snapshot tests is an implementation detail. That details will probably change soon because we might add non-golden tests tehre. But it was always an impelmentation detail. So I wouldn't name them "snapshot" tests, or "golden" tests. Those are "waspc/CLI e2e tests". Then run the CLI, check/validate the output, in some manner. That is it.

> The core idea was to compare snapshots, not to test out the CLI.

I don't think so, from what I know, core idea was/is to test out the CLI end to end, not to compare snapshots (comparing snapshots is a detail).

FranjoMindek:
@Martinsos 

My main concern is about the origin of the idea.
The result (tests) is the same, but the context in which it started matters to me.

- If the goal was "let's verify the changes we made to the compiler output", then I think it’s valid to treat these as snapshot tests of the compiler. The fact that we use CLI to generate them is due to it being our only interface to `waspc`, and we need to do it e2e.

- If the goal was "let's start testing the CLI, beginning with the output of `wasp new/compiler/build/migrate`", then I’d consider these part of the CLI e2e tests.

If it is the 2nd one, I would move `e2e-tests` to `waspc/cli` instead.
So it would be `waspc/cli/e2e-tests`. 
The current situation with `waspc/e2e-test` is confusing their purpose. 

> waspc is CLI.

This I don't get.
I would view CLI as an interface of the `waspc`.
It is our only interface, and we probably won't add more.
But I wouldn't view them as one.

FranjoMindek:
We concluded to keep `waspc/e2e-tests` as is.
They are in a difficult spot which is hard to pin point exactly.

waspc/README.md
```diff
@@ -89,15 +89,15 @@ If that is the case, relax and feel free to get yourself a cup of coffee! When s
 
 ### Run tests
 
-```
+```sh
 ./run test
 ```
 
-to ensure all the tests (unit, e2e, ...) are passing.
+to ensure all the tests (`waspc` + Wasp CLI + Wasp LS + `todoApp` + examples + starters) are passing.
```
sodic: Not sure if necessary to list _all_ the tests we have here. It's just another list that can go out of date :)

waspc/README.md
```diff
@@ -156,10 +156,10 @@ When done, new tab in your browser should open and you will see a Todo App!
...
 4. Use `./run build` to build the Haskell/cabal project, and `./run wasp-cli` to both build and run it. If you changed code in `packages/`, you will also need to run `./run build:packages` (check [TypeScript Packages section](#typescript-packages) for more details). Alternatively, you can also run slower `./run build:all` to at the same time build Haskell, TS packages, and any other piece of the project in one command.
 5. For easier manual testing of the changes you did on a Wasp app, you have the `examples/todoApp` app, which we always keep updated. Also, if you added a new feature, add it to this app (+ tests) if needed. Check its README for more details (including how to run it).
-6. Run `./run test` to confirm that all the tests are passing (unit + integration + e2e). If needed, accept changes in the e2e golden/snapshot tests with `./run test:e2e:accept-all`. Check "Tests" for more info.
+6. Run `./run test` to confirm that all the tests are passing (`waspc` + Wasp CLI + Wasp LS + `todoApp` + examples + starters). If needed, accept changes in the snapshot tests with `./run test:waspc:e2e:accept-all`. Check "Tests" for more info.
```
sodic: If we decided to rename e2e tests into snapshot tests, we should apply that change everywhere.

This means that:
- The command `test:waspc:e2e:accept-all` should become something like `test:waspc:snapshots:accept-all`.
- The code should reflect those changes (in `run.sh` and anywhere else).
- The file system names should reflect those changes too.

The rule of thumb is this: If a new person started working on Wasp, they should never see any indication that snapshot tests were ever called e2e tests.

### EDIT
 I know see we may have [given up on this renaming](https://github.com/wasp-lang/wasp/pull/3018#issuecomment-3184281534). So, just to sum up: I don't particularly care what we call the tests (cli, snapshots, e2e, or otherwise).

I care about:
1. Consistent terminology
2. Simple test names instead of a technically correct hierarchy (ideally one word for each kind of test). IMO it's more important to have clear terminology than 100% correct terminology.

I'll give one possible example for the second thing.

Let's assume we have "snapshot e2e tests" and "headless e2e tests." When someone inevitably uses the term "e2e tests" to refer to only one of those and not the other, they'll learn to clarify (after someone else possibly makes a mistake). Once everyone is on the same page, we'll just start calling them "snapshot" and "headless" (without the "e2e" part) because it's shorter. So we might as well do that right away.

sodic: This is another place where we list all the tests. I think we can kick out everything in the parentheses. It's not that useful and can easily become outdated and confusing.

waspc/README.md
```diff
@@ -258,49 +258,56 @@ We can however still organize tests manually if we want in Tasty test trees, and
...
 
 We additionally use **doctest** for testing code examples in documentation.
 
-All tests go into `test/` directory.
+All tests go into `tests/` directory.
 This is convention for Haskell, opposite to mixing them with source code as in Javascript for example.
 Not only that, but Haskell build tools don't have a good support for mixing them with source files, so even if we wanted to do that it is just not worth the hassle.
 
-Tests are run with `./run test`. This will run all the tests: unit, integration, and e2e.
+Tests are run with `./run test`. This will run all the tests (`waspc` + Wasp CLI + Wasp LS + `todoApp` + examples + starters).
```
sodic: Hm, another place where we list the tests.

If we really want to list them in the readme, then we could do it once and point to that each time.

But I vote for not listing them at all

waspc/README.md
```diff
@@ -258,49 +258,56 @@ We can however still organize tests manually if we want in Tasty test trees, and
...
 
-To run unit tests only, you can do `./run test:unit`.
-To run individual unit test, you can do `./run test:unit "Some test description to match"`.
+To run `waspc` unit tests only, you can do `./run test:waspc:unit`.
+To run individual unit test, you can do `./run test:waspc:unit "Some test description to match"`.
+To run `waspc` e2e tests only, you can do `./run test:waspc:e2e`.
```
sodic: Another place where we use the `e2e` instead of `snapshot`.

waspc/README.md
```diff
@@ -258,49 +258,56 @@ We can however still organize tests manually if we want in Tasty test trees, and
...
-```
-./run test:e2e:accept-all
+This means that when you make a change in your code that modifies the generated code, snapshot tests will fail while showing a diff between the new generated code and the previous (golden) one.
+This gives you an opportunity to observe these differences and ensure that they are intentional and that you are satisfied with them.
+**It is the PR author's (or the reviewers for outside contributions) responsibility to carefully review these diffs.**
+Do not blindly accept changes, ensure they align with your intended modifications.
+If you notice something unexpected or weird, you have an opportunity to fix it.
+Once you are indeed happy with the changes in the generated code, you will want to update the golden output to the new (current) output, so that tests pass.
+Basically, you want to say "I am ok with the changes and I accept them as the new state of things.".
+Easiest way to do this is to use the convenient command from the `./run` script:
```
sodic, line 296: Nice simplification here!

waspc/run
```diff
@@ -25,20 +27,55 @@ BUILD_ALL_STATIC_CMD="$WASP_PACKAGES_COMPILE && $BUILD_HS_CMD --enable-executabl
...
-TEST_UNIT_CMD="cabal test waspc-test"
-TEST_CLI_CMD="cabal test cli-test"
-TEST_E2E_CMD="cabal test e2e-test"
-TEST_E2E_DELETE_GOLDENS="rm -rf $PROJECT_ROOT/e2e-test/test-outputs/*-golden"
-TEST_E2E_ACCEPT_ALL_CMD="$TEST_E2E_DELETE_GOLDENS && $TEST_E2E_CMD"
-TEST_HEADLESS_CMD="cd $PROJECT_ROOT/examples/todoApp && npm install && npm run test"
-TEST_CMD="cabal test && echo 'Running headless tests' && $TEST_HEADLESS_CMD && echo 'ALL TESTS PASSED' || { echo 'SOME TESTS FAILED'; exit 1; }"
+# Testing
+WASPC_UNIT_TESTS_CMD="cabal test waspc-tests"
+WASPC_E2E_TESTS_CMD="cabal test waspc-e2e-tests"
```
sodic, line 32: I talked about this in comments on the README - the terminology we communicate to users does not entirely match the terminology in in the script/command/file system.

.github/workflows/waspc-ci.yaml
```diff
@@ -188,7 +188,7 @@ jobs:
           cd examples/todoApp
           # Prepare the environment for the tests
           cp .env.server.headless .env.server
-          HEADLESS_TEST_MODE=build npm run test
+          WASP_RUN_MODE=build npm run test
 
-      - name: Run e2e tests
-        run: cabal test e2e-test
+      - name: Run `waspc` e2e tests
+        run: ./run test:waspc:e2e
```
sodic, line 195: Another place where we should use the new terminology.

.github/workflows/waspc-ci.yaml
```diff
@@ -179,7 +179,7 @@ jobs:
           cd examples/todoApp
           # Prepare the environment for the tests
           cp .env.server.headless .env.server
-          HEADLESS_TEST_MODE=dev npm run test
+          WASP_RUN_MODE=dev npm run test
```
sodic: We still have a lot of "Headless" terminology in this file.

waspc/run
sodic: Let's keep the terminology as simple as possible.
I don't see much value in saying "headless e2e" and "snapshot e2e."

I think we can drop the `e2e` part altogether. The more tests we can categorize as e2e, the less useful (and more confusing).

I therefore suggest:
- Snapshot tests
- Headless tests (or projects tests)

If we insist on keeping `e2e`, I suggest we just apply that name to headless tests and call it a day.

waspc/README.md
sodic: I wouldn't list anything :smile: 

More on this in the other threads.

sodic (commented):
Left some comments but am now reading new stuff. I'll edit that in.

waspc/run
FranjoMindek: I think that is okay in our informal speech, but I wouldn't do it as such in the project itself.
We should be explicit here, since we don't know who is reading.

waspc/README.md
FranjoMindek: Outdated.

waspc/README.md
FranjoMindek: I agree with general list changes.
This will get out of sync quickly. We should just drop it.

.github/workflows/waspc-ci.yaml
```diff
@@ -122,73 +122,74 @@ jobs:
...
 
-      - name: Headless - Cache Node Modules
-        id: headless-cache-node-modules
+      - name: Cache `todoApp` Node modules
         uses: actions/cache@v4
         if: matrix.os == 'ubuntu-22.04'
         with:
           path: ~/.npm
-          key: node-modules-${{ hashFiles('./waspc/headless-test/package-lock.json') }}-${{ matrix.os }}-${{ matrix.node-version }}
+          key: node-modules-${{ hashFiles('./waspc/examples/todoApp/package-lock.json') }}-${{ matrix.os }}-${{ matrix.node-version }}
```
FranjoMindek, line 136: This is a bug we had in CI

examples/waspello/.env.server.example
```diff
@@ -1,2 +1,2 @@
-GOOGLE_CLIENT_ID=
-GOOGLE_CLIENT_SECRET=
+GOOGLE_CLIENT_SECRET='dummy-g-client-secret'
```
FranjoMindek, line 1: I don't see purpose of having both `.env.server.example` and `.env.server.headless` if they are identical.

waspc/README.md
```diff
@@ -258,49 +258,51 @@ We can however still organize tests manually if we want in Tasty test trees, and
...
+To summarize: If you are writing new tests, just put them in a file that ends with `Test.hs` in `tests/` dir and that is it.
 
 For unit testing, we use **Hspec**.
 
 For property testing, we use **Quickcheck**.
 
 We additionally use **doctest** for testing code examples in documentation.
 
-All tests go into `test/` directory.
-This is convention for Haskell, opposite to mixing them with source code as in Javascript for example.
```
FranjoMindek, line 270: Not really, both conventions exist in JS world.
Heck, we use `__tests__` folder in `wasp-config` among others.
Somebody is selling propaganda here.

examples/waspello/.env.server.example
sodic: I think they serve different purposes but currently have the same data by accident and at the moment. There's likely a name for this which I can't remember.

Here's an analogy. I know it's perfectly understandable without one, but I couldn't resist.

You have two mailing lists: `devs@wasp.sh` and `team@wasp.sh`. For the first two years of Wasp's official existance, these two would be identical. Still, it wouldn't make sense to remove one. And sure enough, with time and as the company grew, they diverged.

That said, I'm not 100% sure that's what's going on here, but we should double check when and why we use `.env.server.headless` to ensure it's safe to remove. I also recommend asking the person who put it there as it was probably for a reason (that's [a term](https://en.wikipedia.org/wiki/Wikipedia:Chesterton%27s_fence) I do remember the name of).

If I was guessing, I'd say we'll need proper values for headless tests when we tackle https://github.com/wasp-lang/wasp/issues/2995 (maybe not here, but probably in `todoApp`).

waspc/run
```diff
@@ -161,36 +202,41 @@ case $COMMAND in
     #   around :main --color always that way and it was not working.
     ghcid -T=':main --color always' --command=cabal repl test/TastyDiscoverDriver.hs
     ;;
-
   test)
     echo_and_eval "$TEST_CMD"
     ;;
-  test:unit)
+  test:waspc:unit)
```
sodic, line 214: Do we perhaps want to support all groups with our commands. I think this only means adding `test:waspc`.

waspc/README.md
```diff
@@ -258,49 +258,51 @@ We can however still organize tests manually if we want in Tasty test trees, and
...
-To run cli tests only, you can do `./run test:cli`.
+- To run all tests, you can do `./run test`
+- To run `waspc` unit tests only, you can do `./run test:waspc:unit`.
+- To run individual unit test, you can do `./run test:waspc:unit "Some test description to match"`.
+- To run `waspc` e2e tests only, you can do `./run test:waspc:e2e`.
+- To run Wasp CLI tests only, you can do `./run test:cli`.
+- To run Wasp LS tests only, you can do `./run test:waspls`.
+- To run `todoApp` e2e tests, you can do `./run test:todoApp`.
+- To run examples e2e tests, you can do `./run test:examples`.
+- To run starter templates e2e tests, you can do `./run test:starters`.
```
sodic: If we decide to add `test:waspc` (mentioned in another comment), it might be nice to nest the list.

Other than that, I love the new commands. Very neat!

waspc/README.md
```diff
@@ -258,49 +258,51 @@ We can however still organize tests manually if we want in Tasty test trees, and
...
 
-To run end-to-end tests only, you can do `/run test:e2e`.
+### Wasp e2e snapshot tests
```
sodic: I think the terminology we use for this one is `Wasp CLI e2e tests`.

waspc/README.md
```diff
@@ -168,15 +168,15 @@ When done, new tab in your browser should open and you will see a Todo App!
 
 ### Note for team members
 
-Do not update e2e tests until your PR is approved.
+Do not update snapshot tests until your PR is approved.
```
sodic: A lot of mentions of snapshots throughout this file. It should probably be `waspc e2e tests`, right?

.github/workflows/waspc-ci.yaml
```diff
@@ -122,73 +122,74 @@ jobs:
...
+        run: cabal test wasp-cli-tests waspc-tests waspls-tests
 
-      - name: Ensure examples/todoApp compiles and builds
+      - name: Ensure `todoApp` compiles and builds
         if: matrix.os == 'ubuntu-22.04' || matrix.os == 'macos-13'
         run: ./examples/todoApp/ensure_app_compiles_and_builds.sh
 
-      - name: Headless - Cache Node Modules
-        id: headless-cache-node-modules
+      - name: Cache `todoApp` Node modules
```
sodic, line 131: I see we've removed the "headless" prefix. This loses the grouping the original author intented? Do you think this is important?

If so, perhaps we should put the (ex-)headless stuff into its own job.

.github/workflows/waspc-release.yaml
```diff
@@ -12,8 +12,8 @@ jobs:
   build:
     uses: ./.github/workflows/waspc-build.yaml
 
-  examples-headless-tests:
-    uses: ./.github/workflows/examples-headless-tests.yaml
+  examples-e2e-tests:
+    uses: ./.github/workflows/examples-e2e-tests.yaml
```
sodic, line 16: I found some remaining mentions of `headless` in the project, one of which is likely a bug.

I did this with:
```bash
rg -i --hidden headless | grep -v headlessui | nvim
```
The normal `grep` equivalent is:
```bash
grep headless -i --exclude-dir=node_modules -r . | grep -v headlessui | wc -l
```
So check those out and get back to me :)

Btw, this search is a good showcase of how much faster `rg` is than `grep`.

examples/waspello/.env.server.example
FranjoMindek: Okay, `.env.server.headless` exists because we needed to modify the old `.env.server.example` to work with `wasp-app-runner`.

Specifically we needed it to work with `MailCrab`.
That's how the first `.env.server.headless` was born on `todoApp`.

So the only difference between `.env.server.example` and `.env.server.headless` was `SMTP_*` env vars values. Example one had e.g. `smtp.dummy.com` and headless one had `localhost`. Also similar in other fields.

For the user, this difference in the means nothing.
So I thought why not just merge them together and use the values of `.env.server.headless`.

As for `waspello` I don't know why `.env.server.example` had no values.
Usually we provide "dummy strings" so the application can be at least run.
We don't really need `.env.server.headless` here, only a proper `.env.server.example` instead.

waspc/run
FranjoMindek: Thought about that one. Didn't think anyone would use it but I can add it.
Will do for completion sake.

waspc/README.md
FranjoMindek: Are we? I think we stuck with `waspc e2e tests`?
But on the other side, I wanted to address snapshot portion of `waspc e2e tests` here.

That's why I reworded it to: "Inside of `waspc` e2e tests we have snapshot tests".
So it's a title about a group of tests inside of `waspc e2e tests`.
Should I make it clearer in some way?

waspc/README.md
FranjoMindek: I wanted to differentiate snapshot portion of `waspc e2e tests` from all `waspc e2e tests`.
Currently all `waspc e2e tests` are snapshots, but we are going to change this.
So I don't want to use `waspc e2e tests` when I actually want to refer to snapshot tests exclusively.

.github/workflows/waspc-ci.yaml
FranjoMindek: I've added `todoApp` to make it pop out more.
Doing it in another job is a no-go (for now) because we would need to build `wasp-cli` again.

However, as part of CI refactoring I believe all tests will be decoupled from the "build workflow".
So we will fix this in (near) future.

.github/workflows/waspc-release.yaml
FranjoMindek: I really need to learn `rg` stuff. Thanks.

waspc/run
```diff
@@ -200,11 +203,14 @@ case $COMMAND in
     # --color always is needed for Tasty to turn on the coloring.
     # NOTE: I did not put this into variable because I was not able to put single quotes
     #   around :main --color always that way and it was not working.
-    ghcid -T=':main --color always' --command=cabal repl test/TastyDiscoverDriver.hs
+    ghcid -T=':main --color always' --command=cabal repl tests/TastyDiscoverDriver.hs
```
FranjoMindek, line 206: Found this one

waspc/run
sodic: We agreed to let Franjo exercise his inner Carl Linnaeus and ended up naming the tests according to the technically correct hierarchy.

waspc/README.md
sodic: > Are we? I think we stuck with waspc e2e tests?

Yes, I mistyped.

> Should I make it clearer in some way?

I suggest to changing it to at least "`waspc` e2e snapshot tests". The connection with the list above is not obvious in the current version.

But again, my first choice is not making this distinction at all. Every other part of the system treats waspc e2e tests and snapshot tests as one and the same:
- The list above only says "`waspc` e2e tests," does not mention snapshots
- The command for running these snapshot tests only says `test:waspc:e2e`

Using the snapshot terminology that doesn't come up at all when you're working with the text is IMO confusing. We can explain how the tests work, sure. But I wouldn't use the term "snapshot tests" when referring to them. Is there a benefit to this that I'm missing?

I won't block the PR on this though, so resolve when happy.

waspc/README.md
sodic: Feel free to close this thread since we're discussing it elsewhere

examples/waspello/.env.server.example
sodic: Ok, not quite familiar with all the details. But if the verdict is that we don't and won't need them soon, you can resolve the comment.

waspc/starters-e2e-tests/tests/auth.ts
```diff
@@ -79,7 +79,7 @@ async function performEmailVerification(page: Page, email: string) {
 }
 
 function isRunningInDevMode() {
-  const testMode = process.env.HEADLESS_TEST_MODE ?? "dev";
+  const testMode = process.env.WASP_RUN_MODE ?? "dev";
```
sodic, line 82: Make sure to test this out since a bug slipped by in the first iteration.

waspc/README.md
sodic: We had a misunderstanding here. I meant "Nest `waspc:unit` and `waspc:e2e` under `waspc`. But this is fine too.

Resolve at will.

sodic (approved):
I'm approving.

Please check the comments though and implement what you like.

waspc/README.md
FranjoMindek: Okay the title also works really great as `Waspc e2e tests`, so we can have it like that.

Only odd case is here:
```md
### Note for team members

Do not update snapshot tests until your PR is approved.

Accepting snapshot tests prematurely:

- Slows down the UI.
- Creates noise for the reviewers.
- Makes it less likely that you'll thoroughly check the final diffs after all review iterations.

**Carefully review snapshot test diffs before accepting them** (see [Waspc e2e tests](#waspc-e2e-tests) for more details).
```
Also mentioned a bit in the `## Typical development workflow` above.

Is keeping snapshot there fine for you?

waspc/starters-e2e-tests/tests/auth.ts
FranjoMindek: There is no build tests since we removed `open-saas` from starters e2e tests so this silently passed.
But it should be okay now since we set it correctly, e.g.:
```ts
  const waspAppRunnerBuildEnv = {
    ...process.env,
    WASP_APP_PATH: waspProjectPath,
    WASP_CLI_CMD: waspCliCommand,
    WASP_RUN_MODE: "build",
  };
```

waspc/README.md
FranjoMindek: Oh haha.

waspc/README.md
sodic: Haha, ok so...

I still see no reason to ever say "snapshot tests" in our current setup, ever. I would always say "Waspc e2e tests," if that's the question.

If the question is _"I'm definitely keeping the snaphost terminology, what do I do here?_" then I'd again use the whole name "Waspc e2e snapshot tests."

Why is this case odd? It looks similar to the one discussed above.

infomiho (approved):
Took a quick look and I like it

FranjoMindek:
Well here we go

PR #3020, opened 2025-07-31
fix #48

### LINKED ISSUE #48: `wasp start` failiing on windows
Explained here: https://github.com/wasp-lang/wasp/pull/47#issue-489824724 .

nodtem66:
# WIP

### Description

- “Fixes #48 "  by changing `System.Process proc to shell`
- Need to test in Linux and MacOS

### Steps to compile in Windows 11

- Install [ghcup](https://www.haskell.org/ghcup/)
- Change directory to `waspc`
- Set ghc to version 8.10.7
  ```ps
  ghcup tui
  ```
- Build `wasp-cli`
  ```ps
  cabal build all
  ```
- Test the ability to run `npm --version`
  ```ps
  cabal run wasp-cli
  ```
- Test the ability to create a new project
  ```ps
  cd examples/
  cabal run wasp-cli -- new MyApp
  ```
- Test prisma
  ```ps
  cabal run wasp-cli -- db migrate
  cabal run wasp-cli -- start
  ```

1. [ ] I updated [`ChangeLog.md`](https://github.com/wasp-lang/wasp/blob/main/waspc/ChangeLog.md) with description of the change this PR introduces.


cprecioso:
Hey @nodtem66, thanks for this contribution!

Since the PR is still a draft, please let us know on the [#wasp-dev channel on Discord](https://discord.com/channels/686873244791210014/920312576133443634) once it's ready for a full review.

To help us review this effectively, could you please update the PR description to outline the scope of your changes? It's very helpful for us to know:
- The specific problem this PR intends to solve.
- Any related issues it intentionally does not cover (i.e., what's out of scope).

After that, please run the test suite with `./run test` and share a summary of the results. Knowing which tests are failing (and whether they fall outside the scope of this PR) gives us a perfect overview of its current state and helps us chart a path to getting it merged.
Thanks for the detailed info in advance!

nodtem66:
Hi @cprecioso,

I've added the description that you suggested. The test suit did not pass and produced a long output, so I put it here for reviewing. 

[test.log](https://github.com/user-attachments/files/21545386/test.log)

### Summary
- waspc-test: FAIL `waspc-test.exe: <stdout>: commitBuffer: invalid argument (invalid character)`
- waspls-test: FAIL (4 out of 11 tests failed (0.01s))
- cli-test: PASS
- Test suite e2e-test: SKIP due to using *nix-only commands


I guess it is the problem of Windows EOF, right? Maybe we need to rewrite the test for handling `\r\n` and `\n`. What do you think?

Update: I have fixed this issue in https://github.com/wasp-lang/wasp/pull/3027

cprecioso:
Hey @nodtem66, if you don't mind we need to wait this PR as we'd like @Martinsos to review it when he comes back from holidays. So it might be a couple of weeks or so until this gets reviewed.

nodtem66:
This PR is going to patch many POSIX logics that are not compatible to Windows. I would like to PR everything in one go to save time. Currently, I'm working on the prerequisite PR https://github.com/wasp-lang/wasp/pull/3027. When it's done, I'll update this PR.

waspc/src/Wasp/Job/Process.hs
```diff
@@ -100,7 +100,7 @@ runNodeCommandAsJobWithExtraEnv extraEnvVars fromDir command args jobType chan =
     NodeVersion.VersionCheckFail errorMsg -> exitWithError (ExitFailure 1) (T.pack errorMsg)
     NodeVersion.VersionCheckSuccess -> do
       envVars <- getAllEnvVars
-      let nodeCommandProcess = (P.proc command args) {P.env = Just envVars, P.cwd = Just $ SP.fromAbsDir fromDir}
+      let nodeCommandProcess = (P.shell fullCommand) {P.env = Just envVars, P.cwd = Just $ SP.fromAbsDir fromDir}
```
Martinsos, line 103: I would add some comment here, explaining why it is important that we use `shell` and not `proc`, so future readers don't wonder why `shell` was used.

waspc/src/Wasp/Job/Process.hs
```diff
@@ -119,3 +119,4 @@ runNodeCommandAsJobWithExtraEnv extraEnvVars fromDir command args jobType chan =
             J._jobType = jobType
           }
       return exitCode
+    fullCommand = unwords $ command : args
```
Martinsos, line 122: There is one thing I am not sure about: are we missing out on any args escaping here? So before we were giving command and args to P.proc, and I guess it was potentially doing some manipulation with those args. Now it is just one big string (command + args) -> could that be an issue? Should we be wrapping args into `'` or something? What could happen if we don't? I am always a bit confused with escaping of args, so we should understand this properly.

waspc/src/Wasp/Node/Version.hs
```diff
@@ -76,7 +76,7 @@ readCommandOutput :: String -> [String] -> IO (Either ErrorMessage String)
 readCommandOutput commandName commandArgs = do
   commandResult <-
     catchIOError
-      (Right <$> readProcessWithExitCode commandName commandArgs "")
+      (Right <$> readCreateProcessWithExitCode (shell fullCommand) "")
```
Martinsos, line 79: Hm same question here as above, do we need to somehow do something with arguments? fullCommand was used only for printing so that was fine, but now we are passing it to shell.

Martinsos (commented):
Awesome @nodtem66 , thanks a lot for figuring this out and creating this PR!

So this makes sense to me, and it is great to hear a solution so simple could be the best solution.

As you said, that older PR used a different approach, where it specified process differently when platform was Win. That also works, but this seems simpler + more robust (using shell), as we are letting the system (shell) figure out how to resolve where `npm` or `npx` or whatever is, instead of us hardcoding that knowledge in Wasp.

The only thing I am worried about is potential issues that using `shell` vs `proc` could cause. As you said, potentially more can be executed than via `proc`. Wasp is a piece of software that user executes on their own on their own machine and these calls are very controlled, so it is probably fine, but still, I woudl also like to make sure we are not making any kind of regression here. I mostly asked about escaping the arguments, you will see in the comments.

Martinsos:
Btw GPT is proposing to use `proc "cmd.exe" ["/d","/s","/c", cmdLine]` instead of `shell`, I am not sure if that makes sense or not though. Allegedly should be the same thing, because `shell` just calls `cmd.exe` on Winwdows, but we have a bit more control over these flags?

It also advised not to switch to `shell` on Linux/Mac, only on Windows, since `proc` should be enough on Linux/Mac, and well it is tighter / more expected.

Yeah this is kind of where I ended up chatting with GPT on this, and it makes sense to me:
> - Linux/macOS → proc "npm" args (no quoting headaches, safe argv splitting, works with normal Node installs).
> - Windows → proc "cmd.exe" ["/d","/s","/c", "npm ...args..."] so that cmd.exe handles .cmd/.ps1 resolution and PATHEXT.
>
> The only “extra” thing on Windows is building that "npm ...args..." string safely — double quotes around anything with spaces, escape double quotes inside, and ^-escape cmd metacharacters outside quotes.

So on Linux/Mac we do it as we were doing so far, with `proc`, while on Win we also use `proc` but call "cmd.exe" instead of the actual command and the actual command becomes an argument to it (which we have to properly escape / sanitize).

waspc/src/Wasp/Job/Process.hs
nodtem66: Hi @Martinsos, Thanks for reviewing. I've checked the source code of both P.proc and P.shell. Indeed, P.proc is the modern and safe version of P.shell (see https://github.com/haskell/process/blame/98101f82543b3a28e9f5192d758a508881f8b464/System/Process/Windows.hsc#L423).

I think that we are going to reinvent the wheel by trying to escape args for P.shell. After reviewing these things, I've changed my mind to go for P.proc. (cont. below)

nodtem66:
@Martinsos 
That sounds good to me as well. But, `proc "cmd.exe" ["/d","/s","/c", "npm --version"]` is a little bit incorrect. I wrote a small program to check this and I would like to propose three solutions using P.proc:

1. `P.proc "cmd" ["/d /s /c npm --version"]`
2. `P.proc "cmd" ["/d", "/s", "/c npm --version"]`
3. `P.proc "pwsh" ["-CommandWithArgs", "npm --version"]`

All arguments in these solution will be escaped automatically by [`translateInternal`](
https://github.com/haskell/process/blame/98101f82543b3a28e9f5192d758a508881f8b464/System/Process/Windows.hsc#L423)
 
The reason why `proc "cmd.exe" ["/d","/s","/c", "npm --version"]` doesn't work is related to how cmd.exe handle the string after `/c`. As you can see below in the test case 4, cmd.exe tried to run `"npm --version` which is weird.

## Test program
```haskell
import System.IO
import System.Process
import Control.Exception

-- A minimal program to play with System.Process.proc
main :: IO ()
main = do
    putStrLn "--- Test Case 1: Simple command (ls) ---"
    testSimpleCommand
    putStrLn "\n--- Test Case 2: NPM command ---"
    testNPMCommandViaProc
    putStrLn "\n--- Test Case 3: NPM command (shell) ---"
    testNPMCommandViaShell
    putStrLn "\n--- Test Case 4: NPM command (cmd suggested by GPT) ---"
    testNPMCommandViaProcAndCmdSuggestedByGPT
    putStrLn "\n--- Test Case 5: NPM command (cmd with separate argument) ---"
    testNPMCommandViaProcAndCmdWithSeparatedArgument
    putStrLn "\n--- Test Case 6: NPM command (cmd) ---"
    testNPMCommandViaProcAndCmd
    putStrLn "\n--- Test Case 7: NPM command (cmd with escaped argument) ---"
    testNPMCommandViaProcAndCmdWithEscapedArgument
    putStrLn "\n--- Test Case 8: NPM command (PowerShell) ---"
    testNPMCommandViaProcAndPowershell

-- Running a simple command that succeeds.
testSimpleCommand :: IO ()
testSimpleCommand = do
    let p = proc "ls" ["-l"]
    tryCreateProcess p

-- Failing command there is no npm.exe in the path
testNPMCommandViaProc :: IO ()
testNPMCommandViaProc = do
    let p = proc "npm" ["--version"]
    tryCreateProcess p

-- Unsafe way to run a command
testNPMCommandViaShell :: IO ()
testNPMCommandViaShell = do
    let p = shell "npm --version"
    tryCreateProcess p

-- Failed
testNPMCommandViaProcAndCmdSuggestedByGPT :: IO ()
testNPMCommandViaProcAndCmdSuggestedByGPT = do
    -- Run the command using cmd.exe
    -- /c carries out the command specified by string and then terminates
    -- /d disables the execution of AutoRun commands
    -- /s strip the first and last quotes (") around each arguments and leave the rest of the command unchanged
    -- ref: https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/cmd
    let p = proc "cmd" ["/d", "/s", "/c", "npm --version"]
    tryCreateProcess p

-- Failed
testNPMCommandViaProcAndCmdWithSeparatedArgument :: IO ()
testNPMCommandViaProcAndCmdWithSeparatedArgument = do
    let p = proc "cmd" ["/d", "/s", "/c", "npm", "--version"]
    tryCreateProcess p

-- Solution 1: combine everything
testNPMCommandViaProcAndCmd :: IO ()
testNPMCommandViaProcAndCmd = do
    let p = proc "cmd" ["/d /s /c npm --version"]
    tryCreateProcess p

-- Solution 2: only /c command and arguments
testNPMCommandViaProcAndCmdWithEscapedArgument :: IO ()
testNPMCommandViaProcAndCmdWithEscapedArgument = do
    let p = proc "cmd" ["/d", "/s", "/c npm --version"]
    tryCreateProcess p

-- Solution 3: use Powershell
testNPMCommandViaProcAndPowershell :: IO ()
testNPMCommandViaProcAndPowershell = do
    let p = proc "pwsh" ["-CommandWithArgs", "npm --version"]
    tryCreateProcess p

tryCreateProcess :: CreateProcess -> IO ()
tryCreateProcess p = do
    result <- try (createProcess p) :: IO (Either SomeException (Maybe Handle, Maybe Handle, Maybe Handle, ProcessHandle))
    case result of
        Left err -> putStrLn $ "Error: " ++ show err
        Right (_, _, _, ph) -> do
            exitCode <- waitForProcess ph
            putStrLn $ "Exit Code: " ++ show exitCode
``` 
## Output
```bash
--- Test Case 1: Simple command (ls) ---
total 17
...
...
Exit Code: ExitSuccess

--- Test Case 2: NPM command ---
Error: npm: createProcess: does not exist (No such file or directory)

--- Test Case 3: NPM command (shell) ---
11.4.2
Exit Code: ExitSuccess

--- Test Case 4: NPM command (cmd suggested by GPT) ---
'"npm --version' is not recognized as an internal or external command,
operable program or batch file.
Exit Code: ExitFailure 1

--- Test Case 5: NPM command (cmd with separate argument) ---
node:internal/modules/cjs/loader:1228
  throw err;
  ^

Error: Cannot find module '<path>\node_modules\npm\bin\npm-prefix.js'
    at Function._resolveFilename (node:internal/modules/cjs/loader:1225:15)
    at Function._load (node:internal/modules/cjs/loader:1055:27)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:220:24)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:170:5)
    at node:internal/main/run_main_module:36:49 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v22.14.0
Exit Code: ExitFailure 1

--- Test Case 6: NPM command (cmd) ---
11.4.2
Exit Code: ExitSuccess

--- Test Case 7: NPM command (cmd with escaped argument) ---
11.4.2
Exit Code: ExitSuccess

--- Test Case 8: NPM command (PowerShell) ---
11.4.2
Exit Code: ExitSuccess
```

Martinsos:
Some very detailed work @nodtem66 ! Thanks a lot for putting time into this -> I will have to take a bit of time to properly check what you wrote, and the tricky thing is I just started a bit longer (2 weeks) vacation, so forgive me if it will take me a bit to get fully back to you. I will try to get back to you during it, but if not, you can count on me to get on it after the vacation for sure.

Martinsos:
@nodtem66 just to let you know I am working on this, I dedicated some time and my understanding is now at a much better level I think, but I still need a bit more time to check all you did here and to make some final recommendations -> coming soon.

nodtem66:
Closed because we have the proper solution for address this issue: https://github.com/wasp-lang/wasp/commit/0b1cf6199f84816b8539f09a4af6118ac81de909

Issue #3024, opened 2025-08-01
Test suits: waspc-test fail due to Windows EOF


nodtem66:
## This is an issue for the PR  https://github.com/wasp-lang/wasp/pull/3020

**Describe the bug**
waspc-test and waspls-test fail when using wasp on native Windows. 

**To Reproduce**

- Clone commit 04193ae63
- Run test on native Windows
```bash
cabal test waspc-test
```
```bash
cabal test waspls-test
```

**Expected behavior**
No errors. All tests pass.

**Environment:**

- OS: Windows 11 (Build 10.0.26100)
- Cabal 3.12.1.0
- Stack 3.3.1
- HLS 2.10.0.0
- GHC 8.10.7 (base-4.14.3.0)

PS. I'm working on this issue.

nodtem66:
## Self-note
The additional issue from printing special unicode to Window shell. 
```
waspc-test.exe: <stdout>: commitBuffer: invalid argument (invalid character)
```

nodtem66:
Solved in https://github.com/wasp-lang/wasp/commit/0b1cf6199f84816b8539f09a4af6118ac81de909

PR #3027, opened 2025-08-02
Fix Windows EOF for waspc-test and waspls-test

nodtem66:
> [!NOTE]
> This is a prerequisite PR for fixing Windows EOF problem in https://github.com/wasp-lang/wasp/pull/3020

### Description

waspc-test and waspls-test fail on native Windows  #3024. So, the test suits need some minor fixes for handling `\r\n` in native Windows. 

> waspc-test: FAIL waspc-test.exe: <stdout>: commitBuffer: invalid argument (invalid character)
> waspls-test: FAIL (4 out of 11 tests failed (0.01s))
> cli-test: PASS
> Test suite e2e-test: SKIP due to using *nix-only commands

## Solution (updated)
- Change .gitattributes: add `eol=lf` for all text files (especially the *.golden *.wasp) and `eol=crlf` for all powershell scripts (*.ps1).
- Windows users need to config their envs to handle `LF` in some test suits. This information is noted in `README.md`
```sh
## Handle unicode-encoding
chcp 65001
## Handle LF
git config core.eol lf
git config core.autocrlf input
git check-index --force --all
```
## Output
All tests pass. 🎉 
[test.log](https://github.com/user-attachments/files/21558365/test.log)

[x] **Bug fix** (non-breaking change which fixes an issue).


infomiho:
@nodtem66 you can fix the formatting issues (reported by the CI here: https://github.com/wasp-lang/wasp/actions/runs/16692996635/job/47327393566?pr=3027) by running `./run ormolu:format` locally

nodtem66:
@infomiho It's done.

nodtem66:
I did not understand the error from `ormolu:format`:
```
waspls/test/Wasp/LSP/CompletionTest.hs
  AST of input and AST of formatted code differ.
  Please, consider reporting the bug.
Error: Process completed with exit code 102.
```
https://github.com/wasp-lang/wasp/actions/runs/16743833922/job/47417955568?pr=3027

I guess I should refactor `CompletionTest.hs` so that it can pass the ormolu, right?

nodtem66:
@infomiho This is a new one that pass both `ormolu` and `prettier` locally on Windows.
https://github.com/wasp-lang/wasp/pull/3027/commits/c221c51398af87169fa0ca876ff7111a0eb807d7

I just changed the code style and moved `normalizeCRLF` to the top-level.

nodtem66:
I found a clean solution without adding `normalizeCRLF` to the tests: https://github.com/wasp-lang/wasp/pull/3027/commits/1510a3e31faf1bfe0b8fdfa4b4174b20c52b14d4 
- `.gitattributes` should have `eol=lf` for all text files (especially the *.golden *.wasp) and `eol=crlf` for powershell scripts (*.ps1).
- Stdout encoding error from invalid characters in `waspc/test/SemanticVersion/VersionBoundTest.hs` can be fixed locally using powershell command `chcp 65001` to set the active code page to utf-8.

waspc/.gitattributes
```diff
@@ -1,2 +1,5 @@
+* text eol=lf
+*.ps1 text eol=crlf
+*.png -text
```
Martinsos: I did some investigation, and what worries me here is that by default all files will be treated as text and normalized to have LF as an EOL.

And then we have that exception for *.png, but what about other binary formats? Other image formats, and what not? Should we also be listing them here?

From investigation I did so far, this might be a better setting (I added comments):
```
# For all files that git detects as textual files: always normalize end of lines to
# LF (Unix/Mac style) on commit and keep LF on checkout.
# This way we avoid possibility of Win user changing eols to CRLF by accident,
# and this is ok with Win users because modern editors on Win work ok with LF anyway.
* text=auto eol=lf
# For .ps1 scripts on Win, it is important that their lines end with CRLF to execute correctly.
*.ps1 text eol=crlf
```

This shouldn't then need exclusion of binary files because we are counting on Github to detect correctly if sometihng is text file or not. I read that it can still cause probelms when it detects incorrectly though, so I am even a bit scared of this, but well, seems to be popular default.

Or we go completely simple and go with just
```
# For .ps1 scripts on Win, it is important that their lines end with CRLF to execute correctly.
*.ps1 text eol=crlf
```

and count on textual files being handled well by people on their machiens + us picking it up in PRs if not.

Because the only thing we really need here is to ensure that .ps1 can contain CRLF, right?

nodtem66: Thanks for reviewing.

I'm happy with the last option. Indeed, it is completely simple and clean for this minor issue. I've tested on Windows 11 and all tests were passed without any issues. Also, I've added the detail for running the test on Windows in README.md. Feel free to share any feedback you may have.

Martinsos: All right, I was able to take some decent time to think this through and do research into detail!

So actually, what I wanted to suggest as my final answer is:
```
* text=auto
*.ps1 text eol=crlf
```
Sorry for changing my mind, but based on some more research I did, this seems like the best thing to me.
My reasoning:
- `* text=auto` will always convert text files (what git thinks are text files and not binary files) to have LF eols (specifically it will convert any CRLF into LF) when storing them in git. On the local machine, they will follow the rules from local git configuration (`config.autoctrl` and similar: either stay LF, or get converted to CRLF when checked out). This prevents any CRLFs from making it into the textual files in the repo and making a mess, but it does still allow Win devs to configure conversion of files to CRLF locally if they want to (because e.g. their editor needs it or whatever).
- `*.ps1 text eol=crlf` will always check out .ps1 scripts with CRLF EOLs, independent of local git settings. This is important because .ps1 scripts need CRLF endings to work correctly.

I didn't go with `* text=auto eol=lf` because that would make it so that all files are always checked out with LF, even on Win, despite any of their local git settings, so devs use that extra level of control on their machine, which I don't see the reason for.

How does this sound to you?

I found the whole thing quite complex to be honest, but the starting point you provided helped a lot, and then it was basically jumping between GPT and Stack Overflow, trying to understand these settings into more details, and the use cases.

If you are ok with this as a final thing, feel free to make a change, and I would encourage you to write rich comments, due to how confusing the domain is. I might add a bit to the comments if I think even more clarity is needed, but I think that is it then and we can merge.

waspc/README.md
```diff
@@ -33,6 +33,14 @@ Clone this repo (or its fork) to your machine.
...
 Position yourself in this directory (`waspc/`) and make sure that you are on the `main` branch.
 
+> [!NOTE]
+> On Windows, we recommend setting line endings to `LF` because some test suites are incompatible with `CRLF`:
+>
+> ```sh
+> git config core.eol lf
+> git config core.autocrlf input
+> ```
+
```
Martinsos: Ok if we go with `* text=auto`, then I think we can skip this section completely, right? Because we are ok with whatever they have set locally. We could potentially be like `we store all files with lf so if you want crlf you should do ...` but I think we are now out of the scope, they will likely not need taht anyway and if they do, it is a personal/local thing.

waspc/README.md
```diff
@@ -95,6 +103,13 @@ If that is the case, relax and feel free to get yourself a cup of coffee! When s
 
 to ensure all the tests (unit, e2e, ...) are passing.
 
+> [!NOTE]
```
Martinsos, line 98: Aha, this is local configuration needed in the powershell to run our tests.
Ok makes sense, and thanks a lot for figuring this out!

Anything we can do on our side, we can't right? I guess we could not use those special letters? But ah they are fun, I love those symbols.

Martinsos:
@nodtem66 I reviewed the PR, please check the comments!

One additional question: you said you tested on Windows, and you pasted the test logs and all.
I also need to start testing Wasp more on Windows if we are to drive the whole native Win support forward, but I haven't done any proper development on Windows ever, so I am a bit in the dark regarding the best practices: if you could share how you did it exactly, it would help a lot. I guess you used powershell for it all, and just followed the instructions from the README.md? But did ./run work for you? Anything other special you did? Any info is interesting for me really.

waspc/.gitattributes
nodtem66: Sure, it's also perfect to me. I learned a lot as well and that's why I love contributing the new project outside my field.
I've added the comment about this decision in `.gitattribute`.

waspc/README.md
nodtem66: One thing that I don't know is why the CI build on Windows did not have this issue, so it's always built successfully.
I think it happens just for my local environment. We need someone else to confirm this.

waspc/README.md
nodtem66: The reason for noting that in README.md is I tried to work with default config (`core.autocrlf = true`, `core.eol = not set`) and found that `waspc-test` failed because of CRLF.

For example, this test compares the inline strings (with CRLFs from Win EOL), and the expected strings from Wasp (which uses LF as EOL):

https://github.com/wasp-lang/wasp/blob/4a6f106ea08148cccdabcb5678ad82ff60b16103/waspc/tests/Psl/Parser/WithCtxTest.hs#L17-L73

So, the test failed when this file was checked out with CRLF.

```bash
parsePslWithCtx
    Triple slash documentation comments
      Prisma file is correctly parsed:                                                                                                                                                             FAIL
        expected: Right (Schema [WithCtx (ModelBlock (Model "MyModel" (Body [WithCtx (ElementField (Field {_name = "prop1", _type = Int, _typeModifiers = [], _attrs = [Attribute {_attrName = "id", _attrArgs = []},Attribute {_attrName = "default", _attrArgs = [ArgUnnamed (FuncExpr "autoincrement" [])]}]})) (NodeContext {documentationComments = [" Doc comment attached to `prop1`"]}),WithCtx (ElementField (Field {_name = "prop2", _type = String, _typeModifiers = [], _attrs = [Attribute {_attrName = "unique", _attrArgs = []}]})) (NodeContext {documentationComments = []}),WithCtx (ElementField (Field {_name = "prop3", _type = String, _typeModifiers = [], _attrs = [Attribute {_attrName = "unique", _attrArgs = []}]})) (NodeContext {documentationComments = [" Multiline comments"," For prop3"]})]))) (NodeContext {documentationComments = []})])
         but got: Right (Schema [WithCtx (ModelBlock (Model "MyModel" (Body [WithCtx (ElementField (Field {_name = "prop1", _type = Int, _typeModifiers = [], _attrs = [Attribute {_attrName = "id", _attrArgs = []},Attribute {_attrName = "default", _attrArgs = [ArgUnnamed (FuncExpr "autoincrement" [])]}]})) (NodeContext {documentationComments = [" Doc comment attached to `prop1`\r"]}),WithCtx (ElementField (Field {_name = "prop2", _type = String, _typeModifiers = [], _attrs = [Attribute {_attrName = "unique", _attrArgs = []}]})) (NodeContext {documentationComments = []}),WithCtx (ElementField (Field {_name = "prop3", _type = String, _typeModifiers = [], _attrs = [Attribute {_attrName = "unique", _attrArgs = []}]})) (NodeContext {documentationComments = [" Multiline comments\r"," For prop3\r"]})]))) (NodeContext {documentationComments = []})])
```

I tried to fix this EOL issue by removing `\r` from the inline strings before checking with the expected strings (https://github.com/wasp-lang/wasp/commit/c221c51398af87169fa0ca876ff7111a0eb807d7). But, it is not as clean as the solution using `core.eol = lf` and `.gitattribute`. 

Another clean solution would be rewriting `trimming` to remove both `\n` and `\r\n`

nodtem66:
Sure, I'm happy to share it with you. 

Exactly, I used powershell and followed README.md to install `ghcup`. Then I forked and cloned the project to my local repo.
But, unfortunately, `./run` doesn't work and I was so lazy to convert `./run` into either `./run.bat` or `./run.ps1` (it's not in the scope of the PR). So, I derived the raw commands from `./run` like

-  `cabal run wasp-cli` instead of `./run wasp-cli`.
- `cabal test` instead of `./run test`
- `./bin/ormolu --color always --check-idempotence ...` instead of `./run ormolu:check`
- `./node_modules/.bin/prettier ...` for `./run prettier:check`

These four commands are all I need for the PR. Later, I believe that once you have more contributors who use Windows, one of them will get tired of running these super long commands and create a `./run.bat` for you.

waspc/README.md
Martinsos: Ouf ok that is sometihng I haven't thought of. Yeah ideally this would not happen, and trimming would remove \r\n, but there is always a risk of this happening if we are always working with LF and then someobdy at some point tries to use CRLF. best to stick with LF the whole time.

Ok you know what, you convinced me hah, let's do `* text=auto eol=lf` in .gitattributes. That way we always have the same situation, modern editors on Win know how to work with LF so they should be good, and we don't depend on any local config on Windows.

Thanks for being patient with me here and iterating!

waspc/README.md
Martinsos: hm ok, it still doesn't hurt to have this note here I think!

Martinsos:
> Sure, I'm happy to share it with you.
> 
> Exactly, I used powershell and followed README.md to install `ghcup`. Then I forked and cloned the project to my local repo. But, unfortunately, `./run` doesn't work and I was so lazy to convert `./run` into either `./run.bat` or `./run.ps1` (it's not in the scope of the PR). So, I derived the raw commands from `./run` like
> 
> * `cabal run wasp-cli` instead of `./run wasp-cli`.
> * `cabal test` instead of `./run test`
> * `./bin/ormolu --color always --check-idempotence ...` instead of `./run ormolu:check`
> * `./node_modules/.bin/prettier ...` for `./run prettier:check`
> 
> These four commands are all I need for the PR. Later, I believe that once you have more contributors who use Windows, one of them will get tired of running these super long commands and create a `./run.bat` for you.

Thanks for sharing. Yeah we should probably rewrite `run` into something that is platform-agnostic, e.g. nodejs or maybe something like https://github.com/casey/just ! But great that you managed to get by in the meantime.

waspc/.gitattributes
Martinsos: That is really nice to hear -> I also learned a lot from your contributions, and you pushed us to make progress on Win support due to how seriously and methodically you approached it, which is then enjoyment to iterate upon :).

waspc/.gitattributes
```diff
@@ -1,2 +1,4 @@
+# For .ps1 scripts on Win, it is important that their lines end with CRLF to execute correctly.
+*.ps1 text eol=crlf
```
Martinsos, line 9:
```suggestion
# This will always convert text (as recognized by git) files to have LF EOLs
# (i.e. it will convert any CRLF into LF) when storing them in git, and will also
# check them out with LF. This prevents any CRLFs from making it into the textual
# files in the repo and making a mess, and also ensures that dev experience is
# consistent across the platforms. This works ok on Win because modern editors
# support LF on Win.
* text=auto eol=lf
# For .ps1 scripts on Win, it is important that their lines end with CRLF to execute correctly.
*.ps1 text eol=crlf

```

waspc/README.md
Martinsos:
```suggestion
```

Martinsos:
@nodtem66 I did the changes as discussed -> please throw one last look and let me know if that looks all right, and if so I will merge!

nodtem66:
LGTM.

waspc/.gitattributes
```diff
@@ -1,2 +1,12 @@
+# This will always convert text files (as recognized by git) to have LF EOLs
+# (i.e. it will convert any CRLF into LF) when storing them in git, and will also
+# check them out with LF. This prevents any CRLFs from making it into the textual
+# files in the repo and making a mess, and also ensures that dev experience is
+# consistent across the platforms. This works ok on Win because modern editors
```
Martinsos:
```suggestion
# consistent across the platforms. This also works ok on Win because modern editors
```

PR #3029, opened 2025-08-02
Fix websockets documentation example code

### LINKED ISSUE #2845: Websockets example code is wrong
On this page: https://wasp.sh/docs/advanced/web-sockets#websocketfn-function

We tell users to get the username like this:
```ts
import { getFirstProviderUserId } from 'wasp/auth'

// ...

const username = getFirstProviderUserId(socket.data.user) ?? 'Unknown'
```

But... `getFirstProviderUserId` expects a `UserWithAuthEntity` while `socket.data.user` is an `AuthUser` and has been since version 0.14.0. This causes a type error and it doesn't work.

The correct way to get the username (from our `waspc/examples/todoApp`:

```ts
const username = socket.data.user?.getFirstProviderUserId() ?? "Unknown";
```

ItsRoy69:
### Description

Fixes #2845 - Updates the websockets documentation to use the correct API for getting user provider ID.

[ ] **Just code/docs improvement** (no functional change).

[ ] I updated [`web-sockets.md`](https://github.com/wasp-lang/wasp/blob/main/web/docs/advanced/web-sockets.md).

### DIFF
--- a/web/docs/advanced/web-sockets.md
+++ b/web/docs/advanced/web-sockets.md
@@ -73,11 +73,10 @@ This is how we can define our `webSocketFn` function:
   <TabItem value="js" label="JavaScript">
     ```ts title="src/webSocket.js"
     import { v4 as uuidv4 } from 'uuid'
-    import { getFirstProviderUserId } from 'wasp/auth'
 
     export const webSocketFn = (io, context) => {
       io.on('connection', (socket) => {
-        const username = getFirstProviderUserId(socket.data.user) ?? 'Unknown'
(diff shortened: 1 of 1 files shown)


infomiho:
@ItsRoy69 Thank you for your contribution! Can I ask you to do the same change in the `versioned_docs` folder for versions 17, 16, 15 and 14 because that code example was wrong across multiple versions :)

JyotirmoyWorks:
Appreciate @infomiho ! Fixed in v-14,15,16,17 too as said

PR #3033, opened 2025-08-04
Add documentation on how to disable Wasp's default auth for custom auth actions

cprecioso:
Users using custom auth actions might not realize they have the default auth still enabled, so we tell them to disable it.

Actual changes are in commit [`0cd2687` (#3033)](https://github.com/wasp-lang/wasp/pull/3033/commits/0cd2687f196d7e7e0a55d5c8b49ae16d2ded8cd7), the other change is just converting the code examples to `auto-js`


infomiho (approved):
Yes, good addition!

PR #3035, opened 2025-08-04
Fix fetch depth on checkout action

FranjoMindek:
Contrary to my own belief, `fetch-depth` of 0 fetches all of the git history (commits, tags...).
The default behavior is to fetch only the latest commit, exactly what we want.

### DIFF
--- a/.github/workflows/check-formatting.yaml
+++ b/.github/workflows/check-formatting.yaml
@@ -35,6 +35,8 @@ jobs:
           git config --global core.eol lf
 
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
       - name: Check for Haskell file changes
         id: haskell-changes


infomiho (approved):
LGTM

.github/workflows/check-formatting.yaml
```diff
@@ -37,8 +35,6 @@ jobs:
           git config --global core.eol lf
 
       - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
```
cprecioso: We're using git diff below. Is that affected by fetch-depth?

cprecioso: ISTR there's a property in the `${{ github }}` context that has the changed file list, might be wrong though

cprecioso (commented):
One question, everything else ok

.github/workflows/check-formatting.yaml
FranjoMindek: Good catch, sadly no such properly in github context.

FranjoMindek: I've returned the 0 depth fetch for the diff.
We could probably try to optimize this, but I don't think it's really necessary.

PR #3036, opened 2025-08-05
Fix Waspello CI

cprecioso:
The test example app CI step is failing on Waspello due to not running `wasp ts-setup`. This PR fixes that.

Before: https://github.com/wasp-lang/wasp/actions/runs/16745347836/job/47402470404?pr=2914#step:7:9
After: https://github.com/wasp-lang/wasp/actions/runs/16746190236/job/47405193264?pr=2914#step:8:1

### DIFF
--- a/.github/workflows/examples-headless-tests.yaml
+++ b/.github/workflows/examples-headless-tests.yaml
@@ -28,6 +28,10 @@ jobs:
           - waspello
           - waspleau
           - websockets-realtime-voting
+        include:
+          - example: waspello
+            wasp-ts: true
+
     steps:
       - uses: "actions/checkout@v4"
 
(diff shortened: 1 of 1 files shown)


FranjoMindek:
This is a good workaround for the current problem.
I have two questions:
1) Do we want to open an issue about this?
	Probably not, Wasp needs to parse the `wasp.main.ts` and it can't do that if we didn't set it up for it.

2) Do we want to implement the fix inside of `get-wasp-database-provider.sh`?
	i.e. detect `main.wasp.ts` and run `wasp/wasp-cli ts-setup`
	Introduces side effects you don't expect? We currently do the same in `wasp-app-runner`.

FranjoMindek (approved):
I think this approach is good.
Only thing I'm worried about is that we might encounter it again somewhere else.

You can also think about implementing it in a way described in my second question. Up to you though.

cprecioso:
Mmm I don't think it should go into `get-wasp-database-provider.sh`, `ts-setup` is IMO a kinda prerequisite for anything to work, so it's fine for me to treat it as such.

I'll merge as is.

PR #2732, opened 2025-05-09
Tutorial app from Markdown WIP

infomiho:
Very much work in progress. Testing different strategies for encoding the tutorial app steps so they can be automatically performed to generate the final tutorial TODO app.

Creating this PR for easier reference when I'll write the RFC.

PR #3056, opened 2025-08-11
Experiment of converting patch files to code blocks

infomiho:
## Idea

- #2732

In the PR linked above, we built a tool that enables us to execute tutorial steps automatically and generate the tutorial app. The steps are written down in Markdown (human readable) and then again as patch files (machine readable). This means we are storing the same information _twice_. What if we could do it only _once_?

If we took the patch files and translated them somehow to human friendly instructions for the users in the docs, we might be able to only maintain the patches and not the manually written code blocks.

I've fired up Claude Code and tried to validate it quickly. The results are mixed bag based on the current examples. 

**Preliminary conclusion:** works great for basic cases, it falls apart in more complex cases - maybe better heuristics or better algorithm can help us to get it looking just right. I'm on the fence if this is something we want to do. 

### Basic cases

Hand crafted | Automatic translation
-- | --
image | image

**Score:** 10/10

<details>
<summary>Based on diff</summary>

```diff
diff --git a/src/MainPage.tsx b/src/MainPage.tsx
index bdb62b3..144163a 100644
--- a/src/MainPage.tsx
+++ b/src/MainPage.tsx
@@ -1,37 +1,3 @@
-import Logo from "./assets/logo.svg";
-import "./Main.css";
-
-export function MainPage() {
-  return (
-    <main className="container">
-      <img className="logo" src={Logo} alt="wasp" />
-
-      <h2 className="title">Welcome to Wasp!</h2>
-
-      <p className="content">
-        This is page <code>MainPage</code> located at route <code>/</code>.
-        <br />
-        Open <code>src/MainPage.tsx</code> to edit it.
-      </p>
-
-      <div className="buttons">
-        <a
-          className="button button-filled"
-          href="https://wasp.sh/docs/tutorial/create"
-          target="_blank"
-          rel="noreferrer noopener"
-        >
-          Take the Tutorial
-        </a>
-        <a
-          className="button button-outlined"
-          href="https://discord.com/invite/rzdnErX"
-          target="_blank"
-          rel="noreferrer noopener"
-        >
-          Chat on Discord
-        </a>
-      </div>
-    </main>
-  );
-}
+export const MainPage = () => {
+  return <div>Hello world!</div>;
+};
```
</details>

Hand crafted | Automatic translation
-- | --
SCR-20250811-kvpy | SCR-20250811-kvpc

**Score:** 9/10
We can add the `// ...` comment easily

<details>
<summary>Based on diff</summary>

```diff
diff --git a/schema.prisma b/schema.prisma
index 190e2a8..65d7d30 100644
--- a/schema.prisma
+++ b/schema.prisma
@@ -8,3 +8,9 @@ datasource db {
 generator client {
   provider = "prisma-client-js"
 }
+
+model Task {
+  id          Int     @id @default(autoincrement())
+  description String
+  isDone      Boolean @default(false)
+}
```
</details>

### Complex cases

Hand crafted | Automatic translation
-- | --
SCR-20250811-kvzf | SCR-20250811-kvyh

**Score:** 7/10

I'm missing the extra comments that explain the context of the other parts of the file

<details>
<summary>Based on diff</summary>

```diff
diff --git a/src/MainPage.tsx b/src/MainPage.tsx
index addc7ba..3f1732f 100644
--- a/src/MainPage.tsx
+++ b/src/MainPage.tsx
@@ -1,4 +1,5 @@
-import { getTasks, useQuery } from "wasp/client/operations";
+import type { FormEvent } from "react";
+import { createTask, getTasks, useQuery } from "wasp/client/operations";
 import type { Task } from "wasp/entities";
 
 export const MainPage = () => {
@@ -34,3 +35,24 @@ const TasksList = ({ tasks }: { tasks: Task[] }) => {
     </div>
   );
 };
+
+const NewTaskForm = () => {
+  const handleSubmit = async (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    try {
+      const target = event.target as HTMLFormElement;
+      const description = target.description.value;
+      target.reset();
+      await createTask({ description });
+    } catch (err: any) {
+      window.alert("Error: " + err.message);
+    }
+  };
+
+  return (
+    <form onSubmit={handleSubmit}>
+      <input name="description" type="text" defaultValue="" />
+      <input type="submit" value="Create task" />
+    </form>
+  );
+};
```
</details>

Hand crafted | Automatic translation
-- | --
SCR-20250811-kwba |  SCR-20250811-kwak

**Score:** 3/10
It's completely missing the context where the user should add the component. I think the algorithm should be adjusted to keep the surrounding code.

<details>
<summary>Based on diff</summary>

```diff
diff --git a/src/MainPage.tsx b/src/MainPage.tsx
index eb804b9..cd11965 100644
--- a/src/MainPage.tsx
+++ b/src/MainPage.tsx
@@ -8,6 +8,7 @@ export const MainPage = () => {
 
   return (
     <div>
+      <NewTaskForm />
       {tasks && <TasksList tasks={tasks} />}
 
       {isLoading && "Loading..."}
```
</details>

Hand crafted | Automatic translation
-- | --
SCR-20250811-kwfy |  SCR-20250811-kwff

**Score:** 1/10
Not only it's missing the surrounding context code, the syntax is broken and it's look off

<details>
<summary>Based on diff</summary>

```diff
diff --git a/main.wasp b/main.wasp
index f761afe..d4798a4 100644
--- a/main.wasp
+++ b/main.wasp
@@ -5,7 +5,17 @@ app TodoApp {
   title: "TodoApp",
   head: [
     "<link rel='icon' href='/favicon.ico' />",
-  ]
+  ],
+  auth: {
+    // Tells Wasp which entity to use for storing users.
+    userEntity: User,
+    methods: {
+      // Enable username and password auth.
+      usernameAndPassword: {}
+    },
+    // We'll see how this is used in a bit.
+    onAuthFailedRedirectTo: "/login"
+  }
 }
 
 route RootRoute { path: "/", to: MainPage }
```
</details>

### DIFF
--- a/web/src/remark/tutorial-action-to-codeblock.ts
+++ b/web/src/remark/tutorial-action-to-codeblock.ts
@@ -1,0 +1,72 @@
+import { visit } from 'unist-util-visit';
+import { join } from 'path';
+import type { Root } from 'mdast';
+import type { MdxJsxFlowElement } from 'mdast-util-mdx';
+import { findPatchFile, transformPatchFile } from '../utils/patch-transformer';
+
+interface TutorialActionProps {
+  id: string;
+  action: string;
+  lang?: string;
+  title?: string;
+}
+
+function extractTutorialActionProps(node: MdxJsxFlowElement): TutorialActionProps | null {
+  const props: TutorialActionProps = {
+    id: '',
+    action: 'apply-patch',
+  };
+
+  if (!node.attributes) return null;
+
+  for (const attr of node.attributes) {
+    if (attr.type === 'mdxJsxAttribute' && attr.name) {
+      const value = attr.value;
+      if (typeof value === 'string') {
+        props[attr.name as keyof TutorialActionProps] = value;
+      } else if (value && typeof value === 'object' && 'value' in value) {
+        props[attr.name as keyof TutorialActionProps] = value.value as string;
+      }
+    }
+  }
+
+  return props.id ? props : null;
+}
+
+export function tutorialActionToCodeblock() {
+  return (tree: Root, file: any) => {
+    const basePath = file.dirname || process.cwd();
+    const patchesDir = join(basePath, 'patches');
+
+    visit(tree, 'mdxJsxFlowElement', (node: MdxJsxFlowElement, index, parent) => {
+      if (node.name !== 'TutorialAction') return;
+
+      const props = extractTutorialActionProps(node);
+      if (!props || props.action !== 'apply-patch') return;
+
+      const patchPath = findPatchFile(patchesDir, props.id);
+      if (!patchPath) {
+        console.error(`Patch file not found for id: ${props.id}`);
+        return;
+      }
+
+      const transformedCode = transformPatchFile(patchPath, props.lang || 'ts');
+
+      // Create a code block node
+      const metaValues = [
+        props.title && `title="${props.title}"`,
+        "auto-js",
+      ];
+      const codeBlock = {
+        type: 'code',
+        lang: props.lang || 'tsx',
+        meta: metaValues.filter(Boolean).join(' '),
+        value: transformedCode,
+      };
+
+      if (parent && typeof index === 'number') {
+        parent.children[index] = codeBlock as any;
+      }
+    });
+  };
+}
--- a/web/docs/tutorial/03-pages.md
+++ b/web/docs/tutorial/03-pages.md
@@ -97,7 +97,7 @@ Now that you've seen how Wasp deals with Routes and Pages, it's finally time to 
 
 Start by cleaning up the starter project and removing unnecessary code and files.
 
-<TutorialAction id="prepare-project" action="apply-patch" />
+<TutorialAction id="prepare-project" action="apply-patch" lang="tsx" title="src/MainPage.tsx" />
 
 First, remove most of the code from the `MainPage` component:
 
--- a/web/docs/tutorial/04-entities.md
+++ b/web/docs/tutorial/04-entities.md
@@ -11,7 +11,7 @@ Wasp uses Prisma to talk to the database, and you define Entities by defining Pr
 
 Since our Todo app is all about tasks, we'll define a Task entity by adding a Task model in the `schema.prisma` file:
 
-<TutorialAction id="prisma-task" action="apply-patch" />
+<TutorialAction id="prisma-task" action="apply-patch" lang="prisma" title="schema.prisma" />
 
 ```prisma title="schema.prisma"
 // ...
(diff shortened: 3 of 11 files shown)


web/src/remark/tutorial-action-to-codeblock.ts
```diff
@@ -0,0 +1,72 @@
+import { visit } from 'unist-util-visit';
```
infomiho, line 1: This was Claude Code in action, no need to review it really, I didn't bother with it yet.

One failure scenario, it clashes with `auto-js` and break down if I add it before the `auto-js` plugin @cprecioso but I'll explore what's going on if we decide we want to pursue this idea.

Martinsos:
Yeah would be cool if this worked!

I guess the benefit here is not the code AI produced, since it doesn't work well enough, but that you explored different cases of this conversion.

Based on what you have shown here, I would say it is clear that main challenge is providing the context. I mean code context -> the extra code above or under the change itself. Also the comments. Sometimes you don't want to show any extra code above or below, sometimes you do, ... .

I don't think this can be determined with heuristics, nor automatically, but instead it should be determenistically encoded togetehr with the diff itself.

What if there is a mid-solution, where we add some extra info to the patches? So we don't repeate all of the info, but we also don't rely completely just on patches: instead we supplement patches with extra info that says how to display that additional context. Maybe in patches we can add lines that communicate that somehow, not sure.

If that doesn't turn out easy enough to do right now, I do agree it is ok to stick with the duplicatoin we have right now.

cprecioso:
Here "Maybe better heuristics" is sounding too much like ["A sufficiently smart compiler"](https://wiki.c2.com/?SufficientlySmartCompiler)... 😂

I'd try not to sink too much time in too smart heuristics, and I'm at first glance untrustworthy of them. Maybe we can side-step the need for semantic importance change detection altogether: Full-file code blocks that minimized, expandable, and focused on the change by default? https://x.com/devongovett/status/1953497537022959878

Otherwise, if we can't find a not-too-complex, good-enough heuristic, I wouldn't go ahead with this, as I am reluctant to leave such an important entry point to our product to automatic generatino.

infomiho:
> Otherwise, if we can't find a not-too-complex, good-enough heuristic, I wouldn't go ahead with this, as I am reluctant to leave such an important entry point to our product to automatic generation.

I believe this is the right conclusion here by @cprecioso. Given that this is an important entry point for new users, I don't want to leave things to chance.

I think we'd have to invest more time here to reach a level of quality we need:
- combining source files and the git diffs, this opens up a question of where do we store the files in the point-in-time?
- inventing our own diff format of some sort to accommodate generating code blocks better

I'm open to revisiting this topic in the future, but given that we have one tutorial, the maintenance cost of maintaining hand crafted tutorial is probably the same (or less) than of a automatic code block generation form diffs system.

PR #3059, opened 2025-08-12
Update SuperJSON note

cprecioso:
SuperJSON now exports the `SuperJSONValue` type https://github.com/flightcontrolhq/superjson/pull/322 (but we still can't use it).
This PR updates the note.


infomiho (approved):
Simple enough

Issue #2834, opened 2025-06-11
Execute starter templates e2e (playwright) tests in CI


Martinsos:
Keep in mind that open saas and also most of the other starter are in separate repos from the `wasp` repo. But there should be GA mechanism for cross-repo triggering of workflows.

FranjoMindek:
There is no real Github support for working with workflows from multiple repositories.

If we want `starters` workflow to affect our `wasp` workflow, we need to use Github API to trigger the workflow on `starters` and then get the results by simulating webhooks or by doing pooling inside of the workflows.
Another downside is that we need to install wasp all over again on the `starters` repo. 
-> this takes time

Edit: I assumed here that we want to fail the `wasp` PR if the `starters` workflow fails. If we just want to trigger `starters` with Github API the it's not that bad.

It's probably for the best that we do a similar thing we did with `wasp-app-runner` by making an action? 

Alternative to all of this is to move `starters` to `wasp` repo, and then it would be a bit easier to setup everything.

Martinsos:
So I am missing quite a bit of context here, but will do my best to comment.

Ok, so what do we actually want? Let's forget the implementation details and what can be done and what can't -> what is our goal? I assumed that we want to execute e2e tests on starters when either their code changes, or when Wasp code changes. That later maybe sounds a bit too aggressive hm / resource demanding. We should think about workflow here, when do we want to execute these e2e tests, define that. Change on `main` in Wasp repo? PRs? I don't feel like I have thought enough about our workflow here or all the parts of it to be able to answer this question myself right now, but I could probably easily comment with quite some confidence on any analysis you will do on this front if it is sufficiently thorough.

You said you would want CI in `starters` to fail CI in `wasp` if it fails -> do we want that? I didn't have that in mind. I am not saying that might not be a good thing to have, I am not sure, but I haven't thought about it so far. Comes again to what we want here.

I don't know what we did with `wasp-app-runner` as I haven't followed that so I can't comment on that unfortunately.

Moving `starters` to `wasp` repo sounds a bit compelling to me because I like keeping everything in the same repo :D, but from what I understood, there are multiple factors here to be taken into account when making this decision, so we can't make it just based on this issue, we need to take all of it into account. Maybe that is also one thing we should analyze properly and thoroughly right now, and make a decision, because it is seems we are not quite sure and are sticking with status quo while not being completely convinced what is the best option?

FranjoMindek:
[...]

PR #3070, opened 2025-08-12
Run starters e2e tests in CI

FranjoMindek:
Runs e2e headless tests in CI

Fixes https://github.com/wasp-lang/wasp/issues/2834

### DIFF
--- a/.github/workflows/starters-e2e-tests.yaml
+++ b/.github/workflows/starters-e2e-tests.yaml
@@ -1,0 +1,51 @@
+name: Run headless e2e tests on starter templates
+
+on:
+  workflow_call:
+  push:
+    paths:
+      - "waspc/data/Cli/starters/**"
+    branches:
+      - main
+      - release
+  pull_request:
+    # paths:
+    #   - "waspc/data/Cli/starters/**"
+
+env:
+  WASP_TELEMETRY_DISABLE: 1
+
+jobs:
+  test-wasp-config:
+    name: Test wasp-config
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: "actions/checkout@v4"
+        with:
+          fetch-depth: 0
+
+      - uses: ./.github/actions/setup-haskell
+        with:
+          ghc-version: "8.10.7"
+          cabal-version: "3.10.2.0"
+          cabal-project-dir: waspc
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: lts/*
+
+      - name: Install Wasp CLI
+        working-directory: waspc
+        run: |
+          ./run install
+
+      - name: Install dependencies
+        working-directory: waspc/starters-e2e-tests
+        run: |
+          npm ci
+          npx playwright install --with-deps
+
+      - name: Run headless E2E tests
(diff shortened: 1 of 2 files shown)


.github/workflows/waspc-release.yaml
```diff
@@ -15,11 +15,21 @@ jobs:
...
   create-release:
-    needs: [ci, build, examples-headless-tests, wasp-config-tests]
+    needs:
+      [
+        ci,
+        build,
+        examples-headless-tests,
+        starters-e2e-tests,
+        wasp-config-tests,
+      ]
```
cprecioso: We can just use the regular YAML syntax

```suggestion
      - ci
      - build
      - examples-headless-tests
      - starters-e2e-tests
      - wasp-config-tests
```

.github/workflows/starters-e2e-tests.yaml
```diff
@@ -0,0 +1,49 @@
...
+on:
+  workflow_call:
+  push:
+    paths:
+      - "waspc/data/Cli/starters/**"
+    branches:
+      - main
+      - release
+  pull_request:
+    # paths:
```
FranjoMindek: TEMP

.github/workflows/starters-e2e-tests.yaml
```diff
@@ -0,0 +1,49 @@
+name: Run e2e tests on starter templates
+
+on:
+  workflow_call:
+  push:
+    paths:
+      - "waspc/data/Cli/starters/**"
```
sodic: Not sure about this condition. Starter tests can fail even if we do something to `waspc` without touching them, right?

The obvious example is changing some of the Haskell files that work with starters.

sodic (commented):
Looks reasonable! I have one concern.

.github/workflows/starters-e2e-tests.yaml
FranjoMindek: Yes that is tricky.
`workflow_call` and `waspc-release.yaml` make sure we won't have bugs before release. 
Though something could exist on `main` before that.

I don't know if there is a good enough heuristics for the `paths` part we can do.
We could run it on every PR/merge but I thought that might be too aggressive.

A compromise is every merge and `paths` respecting PRs.

I don't know if you have any idea other?

FranjoMindek: Obviously I picked the might be bugged on `main` but won't be on `release` compromise here.
But we can do something else.

sodic: > A compromise is every merge and paths respecting PRs.

I think this is ok. I might even remove `paths` then, just to make it less misleading. What do you think?

FranjoMindek: In that one I meant we have `paths` on `pull_request` but not on `push`.
What do you exactly mean by remove `paths`?

FranjoMindek: To clarify, Filip wanted to remove trigger on `pull_request`, run it on every `push`.

But in the end we will remove `paths` on `push` and leave `pull_request` with `paths`.

PR #3077, opened 2025-08-13
Treat package-lock.json as generated in the PRs

Martinsos:
Added by @infomiho 

Read more about how `linguist-generated` works: https://medium.com/@clarkbw/managing-generated-files-in-github-1f1989c09dfd

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,5 @@
 *.hs linguist-detectable=false
 
 /examples/* linguist-documentation
+
+**/package-lock.json linguist-generated=true


.gitattributes
```diff
@@ -3,3 +3,5 @@
 *.hs linguist-detectable=false
 
 /examples/* linguist-documentation
+
+**/package-lock.json linguist-generated=true
```
FranjoMindek, line 7: Table: https://github.com/github-linguist/linguist/blob/main/docs/overrides.md#summary

~~Would `linguist-vendored` be more appropriate?
`package-lock.json` shows the true state of things when we use version ranges. 
It may be important for us. Not 100% sure.~~

edit: Hiding in diffs just makes it non-expanded by default and that is okay. 👍🏻 

`package-lock.json` certainly is generated though.

I don't think we need `=true`.

FranjoMindek (approved):
~~I just have a question regarding which override to use.~~
Nothing blocking.

.gitattributes
FranjoMindek: I've thought that hidden in diffs might mean we won't see it at all in PRs.
Though that may be false.
It might just mean that it's not expanded by default, I'll have to check.
edit: It is.

If it's just that then we can hide it for sure.

infomiho (approved):

.gitattributes
Martinsos: Yeah it will show it in PR but it will be marked as generated and hidden by default, it helps a bit.
Sorry, I should have explained what this change will do in the PR description!
Btw I checked docs https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github and while you are probably right that `=true` might not be needed, they show example with it, so I will leave it.

PR #3078, opened 2025-08-13
Ignore `tgz` files in e2e tests

infomiho:
This PR:
- adds logic not to check `tgz` file contents in the e2e tests -> **but it will check if the `tgz` files exist, in other words they will appear in the expected-files list**
- makes sure to build the libs before running the e2e tests (they fail otherwise)

The CI will fail because I didn't update the e2e tests - I'll do that once the parent PR is done.


waspc/run
```diff
@@ -19,16 +19,17 @@ DEFAULT_COLOR="\033[39m"
 
 WASP_PACKAGES_COMPILE="${SCRIPT_DIR}/tools/install_packages_to_data_dir.sh"
 WASP_LIBS_COMPILE="${SCRIPT_DIR}/tools/install_libs_to_data_dir.sh"
+WASP_ALL_DEPS_COMPILE="$WASP_PACKAGES_COMPILE && $WASP_LIBS_COMPILE"
```
infomiho, line 22: Group all deps installation into one name

waspc/e2e-test/GoldenTest.hs
```diff
@@ -150,3 +151,6 @@ runGoldenTest goldenTest = do
           where
             unsafeDecodeAnyJson :: B.ByteString -> Aeson.Value
             unsafeDecodeAnyJson = fromJust . Aeson.decodeStrict
+
+    isTgzFile :: FilePath -> Bool
+    isTgzFile filePath = ".tgz" `isSuffixOf` filePath
```
Martinsos:
```suggestion
    isTgzFile = (".tgz" `isSuffixOf`)
```

Martinsos (approved):

waspc/e2e-test/GoldenTest.hs
FranjoMindek: I would use `FilePath` specific methods to make it clearer.
e.g. `isExtensionOf` which perfectly describes what we are doing much better.

Maybe we can even inline then?
```hs
&& not "tgz" `isExtensionOf` filePath  -- note ".tgz" also works

File: waspc/run [hunk 3]
```diff
@@ -19,16 +19,17 @@ DEFAULT_COLOR="\033[39m"
...
+BUILD_ALL_CMD="$WASP_ALL_DEPS_COMPILE && $BUILD_HS_CMD"
+BUILD_ALL_STATIC_CMD="$WASP_ALL_DEPS_COMPILE && $BUILD_HS_CMD --enable-executable-static"
 
-INSTALL_CMD="$WASP_PACKAGES_COMPILE && $WASP_LIBS_COMPILE && cabal install --overwrite-policy=always"
+INSTALL_CMD="$WASP_ALL_DEPS_COMPILE && cabal install --overwrite-policy=always"
 
 TEST_UNIT_CMD="cabal test waspc-test"
 TEST_CLI_CMD="cabal test cli-test"
-TEST_E2E_CMD="cabal test e2e-test"
+TEST_E2E_CMD="$WASP_ALL_DEPS_COMPILE && cabal test e2e-test"
```
FranjoMindek: Do we want this?
We have `build:all` for a reason. I don't think we want to repeat this each time.

waspc/run
FranjoMindek: Nice

waspc/e2e-test/GoldenTest.hs
Martinsos: Ok nice one!

infomiho: Perfect suggestion, I quickly wrote this one without much thought.

waspc/run
infomiho: Yep, I've removed it now. I realized that I'll need to update the `run` script and the waspc README to take into account the libs. I'll do this in a separate PR.

infomiho:
@FranjoMindek ready for another look

PR #3094, opened 2025-08-18
Remove `open-saas` foundations from `starters-e2e-tests`

FranjoMindek:
Removes code which was left as foundation to running `open-saas`'s `e2e-tests` inside of this script.


waspc/starters-e2e-tests/src/starter-e2e-tests.ts
```diff
@@ -1,31 +1,19 @@
...
 };
 
-export type StarterWithE2ETests = StarterWithoutE2ETests & {
-  includedTestsRelativePath: IncludedTestsRelativePath;
-};
-
 export const STARTERS_E2E_TESTS: StarterE2ETests[] = [
   {
     starterName: "minimal",
     waspProjectRelativePath: "./",
```
FranjoMindek: `waspProjectRelativePath` is not really necessary now.
But I will leave it for now.

sodic: I think we can remove it. Choose what you like.

sodic (approved):
Looks good!

waspc/starters-e2e-tests/package.json
```diff
@@ -17,6 +17,7 @@
     "zx": "^8.8.0"
   },
   "devDependencies": {
-    "@types/node": "^24.0.1"
+    "@types/node": "^24.0.1",
+    "typescript": "^5.9.2"
```
FranjoMindek, line 21: Don't know how we missed this (I mean I know, global `typescript`).
We need it for `tsc`.

waspc/starters-e2e-tests/src/starter-e2e-tests.ts
FranjoMindek: Done. It does simplify the project a bit.

waspc/starters-e2e-tests/src/starter-e2e-test-runner.ts
```diff
@@ -2,34 +2,29 @@ import { fileURLToPath } from "url";
...
-interface StarterTestsExecution {
+interface StarterTestConfiguration {
   waspCliCommand: WaspCliCommand;
-  starterE2ETests: StarterE2ETests;
+  starterName: StarterName;
 }
 
 export async function runStarterE2ETests(
-  execution: StarterTestsExecution,
+  execution: StarterTestConfiguration,
```
sodic: Parameter name does not match type name (execution vs configuration).

Same goes for other functions.

waspc/starters-e2e-tests/src/starter-e2e-test-runner.ts
```diff
@@ -95,12 +86,16 @@ async function initializeClientEnvironment(
...
 
-async function runDevE2ETests(
-  templateName: string,
-  waspProjectPath: string,
-  waspCliCommand: string,
-): Promise<void> {
-  console.log(`Running DEV e2e tests for ${templateName} starter...`);
+interface StarterTestExecution extends StarterTestConfiguration {
+  waspProjectPath: string;
+}
```
sodic, line 90: Hm, `StarterTestConfiguration` and `waspProjectPath` don't seem semantically related. Why mix them into the same object instead of passing the configuration as one argument, and the path as the other?

If it's because we want to avoid positional arguments, then we should still separate them (i.e,. have a nested object with the configuration instead of flattening). But I vote for using positional arguments (assuming I'm right about them being semantically unrelated).

Preferring named to positional arguments is a nice tie breaker for designing APIs, but it shouldn't be a hard requirement for writing functions, especially when the API is not user-facing and when it's at the expense of natural semantic grouping.

FranjoMindek: Makes sense, I just propagated changes after removing `open-saas` withouth too much thinking here.

This makes sense to me:
```ts
async function runBuildE2ETests(
  waspProjectPath: string,
  { starterName, waspCliCommand }: StarterTestConfiguration,
): Promise<void> {
```

FranjoMindek:
> This PR mixes functional changes (removing open-saas and the now-redundant field in the configuration) with non-functional ones (renaming symbols and files).

I know.
It was kinda weird to leave it as is because removal of `open-saas` made old names no longer reflect their current state. So if I was going to remove `waspProjectRelativePath` I should also make them reflect their new state.

In the end I decided to ship it all together since it was small enough, and it seemed like it should go together.

sodic (approved):
Looks good to me!

PR #3100, opened 2025-08-19
Wasp completion - add missing `db` commands + change instructions

### LINKED ISSUE #2898: When installing Wasp completion, recommend to which directory to put it
Image

It would be helpful if the CLI already recommended something as a best practice. I wanted to do it "right", and then I had to go around and explore where to put it. GPT recommended me to put it in `.completions` so I did that, but it would be nicer if got a rec off the bat.

FranjoMindek:
Fixes #2898

This also changes how we do Wasp CLI completion commands.
Instead of generating the file for the user itself, we rather instruct them what to add to their interactive shell configuration.
This is possible because I've simplified the completion script to:
```bash
complete -o default -o nospace -C "wasp completion:list" wasp
```

## In action
Instructions:
image

Commands:
image


cprecioso:
Instead of writing a file, we can do something like:

~~~
Add this to your .bashrc (or whatever file)

```sh title=".bashrc"
eval "$(wasp completion:generate)"
```
~~~

this is also the way some other tools do it.

waspc/cli/src/Wasp/Cli/Command/BashCompletion.hs
```diff
@@ -62,14 +62,23 @@ printBashCompletionInstruction :: Command ()
 printBashCompletionInstruction =
   liftIO . putStrLn $
     unlines
-      [ "Run the following command to generate bash completion script for wasp on your machine:",
+      [ "Setting up Bash auto-completion for Wasp",
         "",
-        "wasp completion:generate > <your-chosen-directory>/wasp-completion",
+        "1. Generate the completion script by running:",
```
cprecioso: I like the numbering, good touch!

waspc/cli/src/Wasp/Cli/Command/BashCompletion.hs
```diff
@@ -62,14 +62,23 @@ printBashCompletionInstruction :: Command ()
...
+        "     wasp completion:generate > <file_path>",
+        "   (replace <file_path> with where you want to save the script, e.g. ~/.wasp_completion)",
```
cprecioso: I like it, but I'm unsure if people will get confused just copy-pasting `<file-path>` and not replacing it. Although at some point we need to stop being defensive and trust our users.

What do you think?

waspc/cli/src/Wasp/Cli/Command/BashCompletion.hs
```diff
@@ -62,14 +62,23 @@ printBashCompletionInstruction :: Command ()
...
         "",
-        "After that, depending on your system, you will need to edit your bash profile:",
-        "- on MacOS (OSX): you will normally want to edit ~/.bashrc",
-        "- on Linux: you will normally want to edit ~/.bash_profile",
-        "and add this line:",
-        "  source <your-chosen-directory>/wasp-completion",
-        "then reset your terminal session."
+        "2. Configure your shell to use the script:",
+        "   - Bash:",
+        "       * On macOS: edit ~/.bashrc",
```
cprecioso: macOS defaults to `zsh` now since v13, we should move their instructions to the zsh shell.

Ultimately some people will not know which shell they're using. Do you think we can do something about it? I'm not sure I have any good ideas :/

Maybe something like:

```
- bash (the default for most Linux and WSL distributions)
  
  blah

- zsh (the default for macOS)

  blah
```

waspc/cli/src/Wasp/Cli/Command/BashCompletion.hs
cprecioso: Ultimately `.wasp-completion` is a fine choice for people who don't care about littering their home folder. People who do care will now how to move it someplace else.

FranjoMindek: Yeah, people who understand will move it where they want.
But we should have a good default for those who don't understand.
That is the way I'm looking at it.

waspc/cli/src/Wasp/Cli/Command/BashCompletion.hs
FranjoMindek: This is what I struggled a lot with. Not really sure how to approach this.

Old instructions mentioned `bash` for macOS, so I didn't want to get rid of that.
But maybe it's good to modernize it a bit. I don't think too much macOS users use it.

waspc/cli/src/Wasp/Cli/Command/BashCompletion.hs
FranjoMindek: That is way I kinda tried to group them together, use blank lines to my favor.
I think it's okay to trust them a bit.

FranjoMindek:
> Instead of writing a file, we can do something like:
> 
> ~~~
> Add this to your .bashrc (or whatever file)
> 
> ```sh title=".bashrc"
> eval "$(wasp completion:generate)"
> ```
> ~~~
> 
> this is also the way some other tools do it.

I don't know what the usual practice is for completions.
Is there a reason to keep them out of shell configuration and just source the separate script file or not.

Maybe the reason is that usually completion scripts are pretty large.
However, for Wasp, we calculate the completion stuff inside of Wasp CLI instead.
That is why our completion script is:

```sh
function _wasp {
    wasp completion:list
}

complete -o default -o nospace -C _wasp wasp
``` 

I don't know why we have the function though. 
This should work as well:

```sh
complete -o default -o nospace -C "wasp completion:list" wasp
```

Considering that it might be okay to just ask users to copy paste that into their shell configuration / startup file.
We ask the same already, but for sourcing our created script file.

waspc/cli/src/Wasp/Cli/Command/BashCompletion.hs
```diff
@@ -46,30 +43,28 @@ bashCompletion = do
...
 -- return the bash completion instruction
 printBashCompletionInstruction :: Command ()
 printBashCompletionInstruction =
   liftIO . putStrLn $
     unlines
-      [ "Run the following command to generate bash completion script for wasp on your machine:",
+      [ "Setting up Bash auto-completion for Wasp",
+        "",
+        "1. Configure your shell to auto-complete Wasp commands:",
+        "   - Bash (default for most Linux and WSL): edit ~/.bash_profile",
```
cprecioso:
```suggestion
        "   - Bash (default for most Linux and WSL): edit ~/.bashrc",
```

The `.{shell}rc` scripts are for interactive shells, a completion script makes more sense there rather than the `.{shell}_profile` files which are for non-interactive ones.

cprecioso (approved):
I like the simplification and the explanation LGTM. I'm approving in that regard. But I'm not sure this is the approach we want (the user copying the `complete` call directly)

- Cons: One could argue we might want to keep control over generating the script we're loading in users' rc files (for if we make changes in the future) + it looks nicer

- Pros: is simpler and faster for users

I don't know if this is something we care about so I will defer final approval to someone else in the team. (My opinion is let's do it, if we change it later we just update the instructions. )

waspc/cli/src/Wasp/Cli/Command/BashCompletion.hs
```diff
@@ -46,30 +43,28 @@ bashCompletion = do
...
-      [ "Run the following command to generate bash completion script for wasp on your machine:",
+      [ "Setting up Bash auto-completion for Wasp",
+        "",
+        "1. Configure your shell to auto-complete Wasp commands:",
+        "   - Bash (default for most Linux and WSL): edit ~/.bashrc",
+        "   - Zsh (default for macOS): edit ~/.zshrc",
+        "   - Other: check your shell's documentation on how to source a script.",
+        "",
+        "   Add the following line to the end of the file:",
+        "     complete -o default -o nospace -C \"wasp completion:list\" wasp",
```
sodic: '

waspc/cli/src/Wasp/Cli/Command/BashCompletion.hs
```diff
@@ -46,30 +43,28 @@ bashCompletion = do
...
 printBashCompletionInstruction =
   liftIO . putStrLn $
     unlines
-      [ "Run the following command to generate bash completion script for wasp on your machine:",
+      [ "Setting up Bash auto-completion for Wasp",
+        "",
+        "1. Configure your shell to auto-complete Wasp commands:",
+        "   - Bash (default for most Linux and WSL): edit ~/.bashrc",
+        "   - Zsh (default for macOS): edit ~/.zshrc",
+        "   - Other: check your shell's documentation on how to source a script.",
```
sodic: We're not sourcing anything any more.

waspc/cli/src/Wasp/Cli/Command/BashCompletion.hs
```diff
@@ -46,30 +43,28 @@ bashCompletion = do
...
+        "",
+        "   Add the following line to the end of the file:",
```
sodic: Maybe put this at the beginning.

waspc/cli/src/Wasp/Cli/Command/BashCompletion.hs
```diff
@@ -46,30 +43,28 @@ bashCompletion = do
...
 printBashCompletionInstruction :: Command ()
 printBashCompletionInstruction =
   liftIO . putStrLn $
     unlines
-      [ "Run the following command to generate bash completion script for wasp on your machine:",
+      [ "Setting up Bash auto-completion for Wasp",
+        "",
+        "1. Configure your shell to auto-complete Wasp commands:",
+        "   - Bash (default for most Linux and WSL): edit ~/.bashrc",
+        "   - Zsh (default for macOS): edit ~/.zshrc",
```
sodic: Maybe emphasize that these are the default locations since an increasing number of distros have started respecting the XDG config standard.

waspc/cli/src/Wasp/Cli/Command/BashCompletion.hs
FranjoMindek: .

waspc/cli/src/Wasp/Cli/Command/BashCompletion.hs
```diff
@@ -46,30 +44,26 @@ bashCompletion = do
...
     unlines
-      [ "Run the following command to generate bash completion script for wasp on your machine:",
+      [ "Setting up Bash auto-completion for Wasp:",
+        "",
+        "1. Add the following line at the end of your shell configuration file:",
+        styleCode "     complete -o default -o nospace -C 'wasp completion:list' wasp",
+        "",
+        "   Default shell configuration file locations:",
+        "   - Bash: " ++ styleCode "~/.bashrc",
+        "   - Zsh: " ++ styleCode "~/.zshrc",
```
cprecioso, line 64: @FranjoMindek I didn't realize you took away the thing where we say Bash (default for most Linux) and Zsh (default for macOS)

FranjoMindek: Me and @sodic had a discussion about it IRL. 
I don't remember exactly why we did it but there was a reason.
Maybe he can remember.

PR #3101, opened 2025-08-20
Add Wasp version change to 0.17.X migration docs

FranjoMindek:


cprecioso (approved):
Good catch! Leaving it up to you if you want to use the Before/After tabs (I have no strong opinion either way)

FranjoMindek:
I copied the style from previous migration docs. So I'll leave them as is.

PR #3104, opened 2025-08-26
Update footer links

cprecioso:


web/src/components/Footer.jsx
```diff
@@ -44,7 +48,7 @@ const company = [
   },
   {
     text: "Company",
-    url: "https://www.notion.so/wasp-lang/Founding-Engineer-at-Wasp-88a73838f7f04ab3aee1f8e1c1bee6dd#20569f14a8af452db10ae618d764d505",
```
cprecioso, line 47: This page doesn't exist / is not public anymore, but I understand that it was going to the "Who We Are" section in an old job post. I updated the link to the "Who We Are" section in a current job post.

web/src/components/Footer.jsx
```diff
@@ -24,8 +24,12 @@ const community = [
     url: "https://discord.gg/rzdnErX",
   },
   {
-    text: "Twitter",
-    url: "https://twitter.com/WaspLang",
+    text: "X / Twitter",
```
cprecioso, line 27: Didn't want to just say "X" as it looked a bit weird

web/src/components/Footer.jsx
cprecioso: Also, maybe we should create an actual, public "Company" page in the Notion and link it here? Just editing the content from the job post a bit, and adding an up-to-date list of the team

cprecioso: People confused about who is part of the team: https://discord.com/channels/686873244791210014/1409572584273612972

cprecioso:
Done and online! https://wasp.sh/

Issue #3005, opened 2025-07-25
Improving our CI workflow efficiency


FranjoMindek:
General issue for discussing ideas about improving our CI workflows efficiency.

Ides discussed so far:

1. Moving `playwright` setup and cache into an action

	Currently in [`waspc-ci.yaml`](https://github.com/wasp-lang/wasp/blob/main/.github/workflows/waspc-ci.yaml), we have the following 3 steps:
	```yaml
	- name: Headless - Store Playwright's Version
	# ...
	- name: Headless - Cache Playwright Browsers for Playwright's Version
	# ...
	- name: Headless - Setup Playwright
	# ...
	```
	This steps of extract playwright version, try to get browsers from cache, if empty download, install and cache browsers is repeat in each workflow where we work with playwright, and there is bound to be a lot of them.

2. Moving development Wasp CLI (`wasp-cli`) setup and cache (if optimal) into an action.
 
	Previously we only ever needed `wasp-cli` in [`waspc-ci.yaml`](https://github.com/wasp-lang/wasp/blob/main/.github/workflows/waspc-ci.yaml), however not it's required in each of our testing workflows where we want to test on PR's latest commit version of Wasp. The process consists of setting up haskell, setting up node, installing all deps (Node and Haskell), building, and installing Wasp. 
		
	We currently already have `cabal` cache in our [`setup-haskell.yaml` action](https://github.com/wasp-lang/wasp/blob/main/.github/actions/setup-haskell/action.yaml). There is a question of is it worth it to cache anything additional in this process. Especially since there will be a lot of development commits, which may overload our cache storage. 

3. Idea of having between workflow communication to reduce work duplication
	
	There is an idea of coordinating between workflows rather than jobs/steps to achieves less work duplication. 
	```mermaid
	graph TD;
	P[playwright] --> Wait{AND}
	W[wasp-cli] --> Wait
	Wait --> A[waspc workflow]
	Wait --> B[examples workflow]
	Wait --> C[starter templates workflow]
	``` 
	
	However, the coordination overhead we would need to write out here might be not worth it.
	We would need to manually control when to trigger each of the workflows, and manually control all of the async work. 
	I'm not fan of the complexity that this would bring.

cprecioso:
> However, the coordination overhead we would need to write out here might be not worth it.
We would need to manually control when to trigger each of the workflows, and manually control all of the async work.

Wouldn't this just be handled with the `needs` key for the corresponding job?

e.g. this from a WIP personal project

[...]

PR #3111, opened 2025-09-04
Remove path from on push workflow in examples e2e tests

FranjoMindek:
This ensure we run the tests on every merge to `main`.
Why?
Because we can break example apps through `waspc` changes without touching the `examples` folder.

### DIFF
--- a/.github/workflows/examples-headless-tests.yaml
+++ b/.github/workflows/examples-headless-tests.yaml
@@ -3,8 +3,6 @@ name: Run headless tests on root example apps
 on:
   workflow_call:
   push:
-    paths:
-      - "examples/**"
     branches:
       - main
       - release


.github/workflows/examples-headless-tests.yaml
```diff
@@ -3,8 +3,6 @@ name: Run headless tests on root example apps
 on:
   workflow_call:
   push:
-    paths:
-      - "examples/**"
```
cprecioso: Should we remove it from PR trigger too? By the time it reaches the `push` trigger it's already in the code. By doing it in the PR we can block a merge.

FranjoMindek: Me and Filip and this discussion on starters e2e tests.
It's hard to be sure when we affect starters/examples.

We thought that running all of the tests (waspc, starters, examples) on each PR commit might be too much.

This is a compromise.
We make sure to detect when we break wasp, but it might not be done before the merge.

Martinsos: Yeah we can't know when we break those, that is true! We in theory have to run them all the time.

Idea: we could run only one example app, kitchen sink, on any PR, but run all of the example apps on main? OK maybe that is too complex? Ha would be cool if we could run all these tests just before merging, instead of after. Maybe you have to manually trigger that on Github, and it won't let you merge before that, if that is even possible to set up. Maybe that is too cumbersome though? Hm.

Btw are we possibly over optimizng here? What if we just ran them on every PR, and then see what happens, and if that becomes problematic, then we look into running them less often?

OK I just talked quickly with my buddy GPT, and he mentioned using smoke tests + merge queue. So smoke test is kind of what I said before -> smaller suite of tests that is less expensive to run so we can run it on every PR. Maybe that means running just KitchenSink app, as I suggested above. If that is also too expensive, maybe something even smaller / faster, but that is likely to trigger big portion of errors wihtout running all the tests. If such test exists, if not ok.

As for merge queue, I googled it quickly, so it is exaclty what we want -> once PR is ready to be merged, you send it to "merge queue" where more intense checks are performed, and PR is automatically merged if they pass. Cool thing is these checks then run only once we start merging, but again won't let it merge if they don't pass -> exactly what we need.
Check https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue .

So, how about we do smoke/light tests + merge queue?

Martinsos (commented):
@FranjoMindek I raised some questions on how to best handle this, it might result in us handling this in different, more advanced way, please take a look.

.github/workflows/examples-headless-tests.yaml
FranjoMindek: > Idea: we could run only one example app, kitchen sink, on any PR,

We already do as part of `waspc-ci.yaml`.
It's run on any `waspc/**` changes.

> So, how about we do smoke/light tests + merge queue?

Merge queues are a cool idea. I would be happy to explore it.
We could mention it here: https://github.com/wasp-lang/wasp/issues/3005

cprecioso: I've tried to setup merge queues in the past, and just a heads-up that they are very finicky. First point, they work on merge commits, not branch heads (so the trigger would have to be `pull_request:` and not `push:` -- which I'm in favor of). Second, they need to set some checks as required (also in favor), but the required checks are the same for both PRs and merge queues, limiting quite a lot their usefulness as a way of delaying CI work for later, as any checks have to be done at both times.

Martinsos: Oh yeah hm what is the point if you can't have more checks in merge queues than you have in the PRs? That sucks. Are you sure that is so, maybe it is possible? Or maybe they made it possible in the meantime?
We can also potentially use some other service for this, not Github. For merge queues. I know Graphite has them also, might be a way to go.

cprecioso: > Are you sure that is so, maybe it is possible? Or maybe they made it possible in the meantime?

Yes, I did check if that had changed in the meantime since I last used them, but it has stayed the same. There could be workarounds (you mark some checks as required but you check the environment to see if you're in a merge queue and if so do a no-op success); but I'd prefer to not have much weird stuff in the hardest part of our setup to debug (our CI).

Martinsos: Maybe that is the way to go though. Ok thanks for info -> I would still expor emerge queues because they sound right, we just need to find the best way how to do it (GA with "hacks", Graphite, ...).

FranjoMindek:
Added it to our general issue about improving CI efficiency: https://github.com/wasp-lang/wasp/issues/3005

PR #3112, opened 2025-09-04
Make ormolu run only on changed Haskell files in PRs

FranjoMindek:
TODO

Probably will finish later.
This is to not forget the idea.

### DIFF
--- a/.github/workflows/check-formatting.yaml
+++ b/.github/workflows/check-formatting.yaml
@@ -41,15 +41,30 @@ jobs:
       - name: Check for Haskell file changes
         id: haskell-changes
         run: |
-          if git diff --name-only origin/${{ github.base_ref }}..HEAD | grep -E 'waspc/.*\.hs$'; then
-            echo "run=true" >> $GITHUB_OUTPUT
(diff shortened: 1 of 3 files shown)


waspc/cli/exe/Main.hs
```diff
@@ -75,6 +75,7 @@ main = withUtf8 . (`E.catch` handleInternalErrors) $ do
 
   telemetryThread <- Async.async $ runCommand $ Telemetry.considerSendingData commandCall
 
+  -- ormolu cache test
```
FranjoMindek: All of the comments are temporary, they are here to test the changes.

cprecioso:
Given this action is getting a bit complex, would it make sense to migrate to the official ormolu action? https://github.com/haskell-actions/run-ormolu

FranjoMindek:
> Looks alright! Maybe we can make an issue for updating ormolu and link it here. Happy to offload this maintenance burden :D

Most likely will be done as part of updating the Haskell ecosystem.
I'll link the ormolu thing there.

PR #3117, opened 2025-09-05
Brew workaround

cprecioso:
Workaround for failing `brew install` until we do #1446


cprecioso:
Tried it out in a VM, works correctly!

sodic (approved):
Thanks!

PR #3140, opened 2025-09-08
Fix script for deploying examples

sodic:
- Set proper Node version
- Leave an example of how a version string is supposed to look like (`v0.18.0` vs `0.18.0`). Discussed here: https://github.com/wasp-lang/wasp/pull/2857#discussion_r2182872515
- Fixed the missing default input value when calling the workflow

### DIFF
--- a/.github/workflows/deploy-examples.yml
+++ b/.github/workflows/deploy-examples.yml
@@ -8,7 +8,7 @@ on:
         type: string
         description: "Wasp version to use for deployment (e.g., 0.18.0), defaults to latest"
         required: false
-        default: "0.18.0-rc1" # We are relying on the install script treating the empty string as the latest version
+        default: "" # We are relying on the install script treating the empty string as the latest version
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-examples.yml
```diff
@@ -6,8 +6,9 @@ on:
     inputs:
       version:
         type: string
-        description: "Wasp version to use for deployment (default to latest)"
+        description: "Wasp version to use for deployment (e.g., 0.18.0), defaults to latest"
         required: false
+        default: "" # We are relying on the install script treating the empty string as the latest version
```
sodic, line 11: @cprecioso This was a mistake, right?

cprecioso: Maybe yeah, I don't really remember. Still, good change to be explicit.

cprecioso (commented):
LGTM

PR #3145, opened 2025-09-10
Update tutorial example apps

infomiho:


examples/tutorials/TodoAppTs/src/MainPage.tsx
```diff
@@ -1,25 +1,24 @@
...
   createTask,
   getTasks,
   updateTask,
   useQuery,
 } from "wasp/client/operations";
-import { Task } from "wasp/entities";
+import type { Task } from "wasp/entities";
 
-export const MainPage = () => {
+export const MainPage = ({ user }: { user: AuthUser }) => {
```
sodic, line 12: @infomiho Are we sure this is supposed to be here? It's unused and I believe the tutorial finishes without it.

sodic: Miho said that we don't use it but we're keeping it here because it's mentioned in the tutorial. Fair enough, though we should come up with a way to use it.

Issue #3149, opened 2025-09-11
`EmailSignupData` has `username` instead of `email`


infomiho:
We have a bug in our signup action on the client: the `EmailSignupData` has `username` and `password` fields instead of `email` and `password` fields.

Link to the code https://github.com/wasp-lang/wasp/blob/main/waspc/data/Generator/templates/sdk/wasp/auth/email/actions/signup.ts#L8

Reported by user: https://discord.com/channels/686873244791210014/1415343740230307982/1415343740230307982

PR #3150, opened 2025-09-12
Fix email singup data type error + more robust todoApp sigunup testing

FranjoMindek:
This PR fixes the type bug we had with email `signup` method.
It was defined as `username+password` object instead of `email+password` object.
As not to repeat the same bug I also introduce "manual signup" feature to `kitchen-sink` app, and add tests for it via `playwright`.

While implementing tests, I've noticed our auth helpers could be cleaned up.
I've split `helpers.ts` into 3 files: 
- `helpers.ts` - general helper funcs
- `auth.ts` - auth helpers
- `mailcrab.ts` - mailcrab helpers (SMTP client)

The `auth.ts` is insipred by `waspc/starters-e2e-tests/tests/auth.ts` changes, but a little bit modified for what we need in `todoApp`. 
`mailcrab.ts` is fully copied from `starters-e2e-tests`. This one most likely doesn't need an additional review.

I've made it so that functions names now reflect their actions.

`performLogin` and `performSignup` now :
1. Navigate
2. Fill up the form
3. Submit the form

Which means they no longer validate the end result (to expect failure or success).
This is left to the caller to verify since he knows whether to expect failure or success.

Fixes: https://github.com/wasp-lang/wasp/issues/3149


waspc/examples/todoApp/src/features/auth/pages/CustomSignupPage.tsx
```diff
@@ -82,7 +82,7 @@ export const SignupPage = () => {
             <FormError>{errors.address?.message}</FormError>
           </FormItemGroup>
           <FormItemGroup>
-            <Button type="submit">Signup</Button>
+            <Button type="submit">Sign up</Button>
```
FranjoMindek, line 85: This makes it consistent with Auth UI, otherwise we have different `locator` logic in our tests.

waspc/examples/todoApp/src/features/auth/pages/CustomSignupPage.tsx
```diff
@@ -67,7 +67,7 @@ export const SignupPage = () => {
             </Alert>
           )}
           <FormItemGroup>
-            <FormLabel>Email</FormLabel>
+            <FormLabel>E-mail</FormLabel>
```
FranjoMindek, line 70: This makes it consistent with Auth UI, otherwise we have different `locator` logic in our tests.

waspc/examples/todoApp/src/features/auth/pages/ManualSignupPage.tsx
```diff
@@ -0,0 +1,91 @@
+import { useState } from "react";
+import {
```
FranjoMindek, line 2: Copy of `CustomSignupPage.tsx` with changed `handleSubmit` and `success` messasge (same as Auth UI, implies email verificaiton is needed).

waspc/examples/todoApp/e2e-tests/tests/auth.ts
```diff
@@ -0,0 +1,113 @@
...
+import { randomUUID } from "crypto";
+import { isRunningInDevMode } from "./helpers";
+import { getMailCrabEmailVerificationLink } from "./mailcrab";
+
+interface BaseEmailCredentials {
+  email: string;
+  password: string;
+}
+
+interface UserSignupFields {
```
FranjoMindek, line 10: Would be nice if we could source this from the wasp app itself.

waspc/examples/todoApp/e2e-tests/tests/auth-hooks.spec.ts
```diff
@@ -16,9 +17,10 @@ test.describe("auth hooks", () => {
     const emailThatThrowsError = "notallowed@email.com";
     const password = "12345678";
 
-    await submitSignupForm(page, {
```
FranjoMindek, line 19: Makes no sense that submit form:
1) Navigates
2) Fills up the form
3) Submits the form
4) Verifies the form submitted successfully

I reduced it to 2) and 3)

And `performSignup` does 1) 2) and 3).  
Its up to you to perform the verification (since you know if you sent good or bad credentials).

waspc/examples/todoApp/e2e-tests/tests/auth.ts
infomiho: That's an argument for publishing out internal libs 😄

waspc/examples/todoApp/src/features/auth/pages/CustomSignupPage.tsx
sodic: I don't understand this. Can you elaborate?

waspc/examples/todoApp/e2e-tests/tests/async-jobs.spec.ts
```diff
@@ -1,11 +1,12 @@
 import { expect, test } from "@playwright/test";
-import { performLogin, setupTestUser } from "./helpers";
+import { performLogin, setupTestUser } from "./auth";
 
 test.describe("async jobs", () => {
   const credentials = setupTestUser();
 
   test("submits a job and gets result", async ({ page }) => {
     await performLogin(page, credentials);
+    await expect(page).toHaveURL("/");
```
sodic, line 9: What's this?

waspc/examples/todoApp/src/features/auth/pages/CustomSignupPage.tsx
FranjoMindek: Auth UI has button with `"Sign up"` text.
Our custom signup page had button with `"Signup"` text.

In tests we might look for button with text `"Sign up"` button so we can click it.
This will find Auth UI button, but not this one.

So you have to search for a different button while essentially you have the same form.
This makes these two forms behave more similarly. 
We can more easily abstract testing of signup form then.

sodic: Ok, got it. Thanks!

waspc/examples/todoApp/e2e-tests/tests/async-jobs.spec.ts
FranjoMindek: Since `playwright` interacts with the application runtime through browsers, as in reality all actions take time.
When we `performLogin` we submit the login form, but we still haven't received any response from the server.
When we anticipate success, we will receive an auth token and redirect to `/` (by default).
 
`await expect(page).toHaveURL("/");` ensures above happens before we continue with further test logic.
It also has an internal timeout which will throw if it waits too long. 

So we either successfully login or we throw.

waspc/examples/todoApp/e2e-tests/tests/auth.ts
FranjoMindek: https://github.com/wasp-lang/wasp/issues/3155

FranjoMindek:
Both pass the test locally and in CI

PR #3156, opened 2025-09-15
Changelog for `EmailSignupData` type fix

FranjoMindek:


FranjoMindek:
> Maybe we can add more details - but like this is also OK

What details? Can you expand? As in what was the mistake?

FranjoMindek:
Added more detail

PR #3158, opened 2025-09-15
Move React Query to the SDK's peerDependencies

cprecioso:
- Adds support for setting `peerDependencies` in the SDK package.json.

  Peer dependencies signal to npm that this is dependency that can/must be shared with the depending package (in this case, React Query is installed in the SDK package, and the depending package is the user code). This ensure that both will use the same version (and the same global singletons -- React Query's cache), and that it is accessible to both (installed on the main `node_modules` instead of `wasp/out/sdk/node_modules`).

  When the depending package doesn't explicitly ask for the dependency that is requested as a `peer`, npm will still install it by default, therefore the installation of that dep will happen whether the user defines it or not, and we don't impact anyone's usecase.

- Sets React query version to a `^` range with the minimum version to the one that fixed bugs that affected us.

Docs weren't touched because the user can still access our internal React Query in the same way.

### DIFF
--- a/waspc/e2e-tests/test-outputs/waspBuild-golden/waspBuild/.wasp/build/.waspchecksums
+++ b/waspc/e2e-tests/test-outputs/waspBuild-golden/waspBuild/.wasp/build/.waspchecksums
@@ -480,7 +480,7 @@
             "file",
             "server/package.json"
         ],
-        "04173ba393b33ca8d5c5c1cd7700db31b4a29b52a7500ba769d0b5dc011bdc97"
+        "c383ad9dc6a627391d43876d1548f2e475a54ded41e6d025b946041c2694d19d"
     ],
     [
         [
@@ -585,7 +585,7 @@
             "file",
             "web-app/package.json"
         ],
-        "be99038bc812a8af761734c82eebbcf1a17e10bd92888a9abe33e3dfbed6d591"
+        "4597c24e80e5cea6cfa147cf923c3d1504250bd2e9f92fb6c77e5fad031af184"
     ],
     [
         [
--- a/waspc/e2e-tests/test-outputs/waspBuild-golden/waspBuild/.wasp/build/server/package.json
+++ b/waspc/e2e-tests/test-outputs/waspBuild-golden/waspBuild/.wasp/build/server/package.json
@@ -28,7 +28,6 @@
   "nodemonConfig": {
     "delay": "1000"
   },
-  "peerDependencies": {},
   "prisma": {},
   "private": true,
   "scripts": {
--- a/waspc/e2e-tests/test-outputs/waspBuild-golden/waspBuild/.wasp/build/web-app/package.json
+++ b/waspc/e2e-tests/test-outputs/waspBuild-golden/waspBuild/.wasp/build/web-app/package.json
@@ -28,7 +28,6 @@
   "overrides": {
     "rollup": "4.44.0"
   },
-  "peerDependencies": {},
   "private": true,
   "scripts": {
     "build": "tsc --build && vite build",
--- a/waspc/e2e-tests/test-outputs/waspCompile-golden/waspCompile/.wasp/out/.waspchecksums
+++ b/waspc/e2e-tests/test-outputs/waspCompile-golden/waspCompile/.wasp/out/.waspchecksums
@@ -487,7 +487,7 @@
             "file",
             "server/package.json"
         ],
-        "04173ba393b33ca8d5c5c1cd7700db31b4a29b52a7500ba769d0b5dc011bdc97"
+        "c383ad9dc6a627391d43876d1548f2e475a54ded41e6d025b946041c2694d19d"
     ],
     [
         [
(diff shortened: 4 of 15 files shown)


waspc/data/Generator/templates/server/package.json
```diff
@@ -23,5 +23,6 @@
   },
   "prisma": {=& prisma =},
   {=& depsChunk =},
-  {=& devDepsChunk =}
+  {=& devDepsChunk =},
+  {=& peerDepsChunk =}
```
sodic: I don't think we need this in the server or the client. They're treated as a project, not a library, so peer dependencies probably don't make much sense there.

@cprecioso Is there a scenario in which this becomes useful that I'm missing?

waspc/ChangeLog.md
```diff
@@ -27,6 +28,7 @@
 - Better error message when choosing a template that doesn't exist. ([#2992](https://github.com/wasp-lang/wasp/pull/2992))
 - Bash completion setup is simplified. ([#3100](https://github.com/wasp-lang/wasp/pull/3100))
 - Increased the font size in Auth UI components, preventing automatic zooming on mobile devices. (by @Vickram-T-G) ([#3110](https://github.com/wasp-lang/wasp/pull/3110/files))
+- Wasp's React Query version was bumped in order to avoid a bug on their side. ([#3158](https://github.com/wasp-lang/wasp/pull/3158))
```
sodic, line 32: I don't think this description is correct. Which bug are we avoiding?

The current version doesn't have the bug, nor does the latest version in the range set by this PR. So I think this should just say "Wasp now allows more flexibility in versions of React Query" or something along those lines.

waspc/data/Generator/templates/server/package.json
cprecioso: I agree that this is not useful, but, given that the underlying data structure does allow `peerDependencies`, I thought it was more correct to output it in case they were set, than not, so data doesn't become magically lost. no strong opinion anyway, wdyt?

sodic:
Oh, and should we regenerate package-lock.json files for our examples too? I think they currently have the package in the wrong place (under `wasp/node_modules`). But maybe it doesn't matter.

waspc/ChangeLog.md
cprecioso: [According to Miho](https://discord.com/channels/686873244791210014/1417054511490666588/1417095318608281610), the bug was in 4.39.0 but fixed on 4.39.1, so that's why I say we increased it and therefore avoidiing the bug.

cprecioso:
> Oh, and should we regenerate package-lock.json files for our examples too? I think they currently have the package in the wrong place (under `wasp/node_modules`). But maybe it doesn't matter.

This is part of the release checklist, that's why I didn't do it myself.

waspc/ChangeLog.md
sodic: Ok, I get it. I guess I just found the wording weird because the text implies that the bug is present in the current version of Wasp (0.17.1), which it isn't. Not a big deal in any case, it can stay as is.

Resolve when you read this.

sodic:
> This is part of the release checklist, that's why I didn't do it myself.

Oh, right.

If you agree we should do it, then please do it here. [That part of the checklist](https://www.notion.so/wasp-lang/0-18-0-26518a74854c80eebfc7c05ad9233205#26518a74854c80eebfc7c05ad9233205) was already taken care of, so this is a new addition. Everything is already migrated, so all that's necessary is regenerating the `package-lock.json` file and verifying all examples behave well.

P.S. For future versions, we'll make all these updates a rolling thing because leaving them for release caused loads of problems - I'm writing a post mortem now.

waspc/src/Wasp/Generator/NpmDependencies.hs
```diff
@@ -200,6 +206,10 @@ getDependenciesPackageJsonEntry = dependenciesToPackageJsonEntryWithKey "depende
 getDevDependenciesPackageJsonEntry :: NpmDepsForPackage -> String
 getDevDependenciesPackageJsonEntry = dependenciesToPackageJsonEntryWithKey "devDependencies" . devDependencies
 
+-- | Construct peerDependencies entry in package.json
```
sodic: I'd say this comment is redundant.

waspc/ChangeLog.md
cprecioso: I just split it in two :D

waspc/data/Generator/templates/server/package.json
sodic: Hmm, good question. TBH I'd rather have server and web app generator ignore them completely (but I would leave that comment you left in `combineDependencies`).

We want to discourage someone adding a peer dependency to the framework code. We want them to rethink what they're doing. This way they'll probably catch themselves before making a PR. If we enable all the peer dependency plumbing, then no one catches it before the reviewer.

Plus, we're testing for stuff we don't care about. So I vote for "kick them out"

Issue #2482, opened 2025-02-03
Update to React 19


infomiho:
**EDIT** from @cprecioso:

We'll keep this issue as the work item for just _updating_ the version (changing the version number and make everything work with it).

We have another task for actually reviewing and adopting the new features in React 19: https://github.com/wasp-lang/wasp/issues/3453

---

React 19 is the latest version as of now and we should migrate to it.

It doesn't seem there are a lot of [breaking changes](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#breaking-changes) but we should investigate how does the bump affect Wasp's code and our users' code.

cprecioso:
Related #1239

infomiho:
We can upgrade the workarounds for `useEffect` being called 2 times after we upgrade to React 19 - check this comment: https://github.com/wasp-lang/wasp/issues/3407#issuecomment-3569901368

cprecioso:
@infomiho should we actually have that note in #3453?

PR #1533, opened 2023-10-24
Relax node version

Martinsos:
### Description

Fixes #1433 .

Now we check if the node version is older than some specific version that Wasp is happy with, and if so, we are ok with that.

I also updated this minimal node version to the latest LTS which is node 20.

I updated the docs, e2e tests, and tested running a new application -> both `wasp start` and `wasp build`, and it seems all works fine.

### Select what type of change this PR introduces:

1. [ ] **Just code/docs improvement** (no functional change). 
2. [ ] **Bug fix** (non-breaking change which fixes an issue).
3. [ ] **New feature** (non-breaking change which adds functionality).
4. [X] **Breaking change** (fix or feature that would cause existing functionality to not work as expected).

### Update Waspc ChangeLog and version if needed
If you did a bug fix, new feature, or breaking change, that affects waspc, make sure you satisfy the following:
1. [X] I updated [ChangeLog.md](https://github.com/wasp-lang/wasp/blob/main/waspc/ChangeLog.md) with description of the change this PR introduces.
2. [X] I bumped waspc version in [waspc.cabal](https://github.com/wasp-lang/wasp/blob/main/waspc/waspc.cabal) to reflect changes I introduced, with regards to the version of the latest wasp release, if the bump was needed.

PR #527, opened 2022-03-25
[Experimental] Enumerate wasp keywords

sodic:
This PR tries to tidy up `Wasp.Analyzer.Token`. The main idea was to reduce duplication and make different token type groups less arbitrary and interleaved.

PR #3159, opened 2025-09-15
Move to npm workspaces

### LINKED ISSUE #3130: Share dependencies through an npm workspace
### Description 

Turn the user’s `package.json` into an [npm workspace](https://docs.npmjs.com/cli/v8/using-npm/workspaces), containing the user project and the `./wasp/*` generated projects.

- Projects refer to one another through workspace dependencies instead of path aliases, copying files, or relative `../../..` paths.
- Dependencies are installed once overall, not once per `package.json`
- Versions are resolved taking into account all of the ranges, not just the current `package.json`, and de-duplicated.
- `package-lock.json` also locks the versions in the generated project, instead of upgrading them to the latest allowable version every time.

Edge cases to look for: what happens when the generated projects…

- don’t exist yet?
- need to be recreated?
- have just been recreated?
- we’ve updated the Wasp version?
- is there a situation in which we’d have out-of-sync generated projects or dependencies?

> [!WARNING]
> Apparently, `npm` doesn't allow multiple subprojects with the same name in the same workspace. We generate the same projects at `.wasp/out/` and `.wasp/build`, so we need to find a solution or workaround.

### Steps

- [ ] Check Wasp behaves correctly with `package.json#workspaces`
- [ ] Check that dependencies are correctly shared
- [ ] Edit the starter	`package.json`'s to have the `workspaces` key
- [ ] Add a check for `workspaces` in the compiler
- [ ] Adapt `wasp deps` command
- [ ] Test that dependency checking logic inside waspc still works correctly
- [ ] Add a migration guide and documentation

### Related issues
- https://github.com/wasp-lang/wasp/issues/559
- https://github.com/wasp-lang/wasp/issues/1644
- https://github.com/wasp-lang/wasp/issues/1640
- https://github.com/wasp-lang/wasp/issues/1854
- https://github.com/wasp-lang/wasp/issues/1838
  - Probably solves https://github.com/wasp-lang/wasp/issues/1845
- https://github.com/wasp-lang/wasp/issues/1859
- #2726

cprecioso:
Closes #3130

WIP

### DIFF
--- a/waspc/src/Wasp/Generator/ServerGenerator.hs
+++ b/waspc/src/Wasp/Generator/ServerGenerator.hs
@@ -129,7 +129,8 @@ genPackageJson spec waspDependencies = do
       (C.asServerFile [relfile|package.json|])
       ( Just $
           object
-            [ "depsChunk" .= N.getDependenciesPackageJsonEntry combinedDependencies,
+            [ "packageName" .= ("@wasp.sh/generated-server-" ++ (if AS.isBuild spec then "build" else "dev")),
+              "depsChunk" .= N.getDependenciesPackageJsonEntry combinedDependencies,
               "devDepsChunk" .= N.getDevDependenciesPackageJsonEntry combinedDependencies,
               "nodeVersionRange" .= (">=" <> show NodeVersion.oldestWaspSupportedNodeVersion),
               "startProductionScript"
--- a/waspc/src/Wasp/Generator/WebAppGenerator.hs
+++ b/waspc/src/Wasp/Generator/WebAppGenerator.hs
@@ -109,7 +109,7 @@ genPackageJson spec waspDependencies = do
       (C.asWebAppFile [relfile|package.json|])
       ( Just $
           object
-            [ "appName" .= (fst (getApp spec) :: String),
+            [ "packageName" .= ("@wasp.sh/generated-client-" ++ (if AS.isBuild spec then "build" else "dev")),
               "depsChunk" .= N.getDependenciesPackageJsonEntry combinedDependencies,
               "devDepsChunk" .= N.getDevDependenciesPackageJsonEntry combinedDependencies,
               "overridesChunk" .= N.getDependencyOverridesPackageJsonEntry dependencyOverrides,
--- a/waspc/data/Cli/starters/basic/package.json
+++ b/waspc/data/Cli/starters/basic/package.json
@@ -1,6 +1,7 @@
 {
   "name": "__waspAppName__",
   "type": "module",
+  "workspaces": [".wasp/out/*", ".wasp/build/*"],
   "dependencies": {
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
--- a/waspc/data/Cli/starters/minimal/package.json
+++ b/waspc/data/Cli/starters/minimal/package.json
@@ -1,6 +1,7 @@
 {
   "name": "__waspAppName__",
   "type": "module",
+  "workspaces": [".wasp/out/*", ".wasp/build/*"],
   "dependencies": {
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
--- a/waspc/data/Generator/templates/react-app/package.json
+++ b/waspc/data/Generator/templates/react-app/package.json
@@ -1,6 +1,6 @@
 {{={= =}=}}
 {
-  "name": "{= appName =}",
+  "name": "{= packageName =}",
   "version": "0.0.0",
   "private": true,
   "type": "module",
--- a/waspc/data/Generator/templates/server/package.json
+++ b/waspc/data/Generator/templates/server/package.json
@@ -1,6 +1,6 @@
 {{={= =}=}}
 {
-  "name": "server",
+  "name": "{= packageName =}",
   "version": "0.0.0",
   "private": true,
   "type": "module",


waspc/src/Wasp/Generator/Valid/Common.hs
cprecioso, line 1: This file is just stuff extracted from `TsConfig.hs` that I also wanted to use in `PackageJson.hs`.

waspc/src/Wasp/Generator/Valid/PackageJson.hs
```diff
@@ -135,8 +149,8 @@ getExpectedPackageJsonDependencyKey = \case
   RequiredDevelopment -> "devDependencies"
   Optional -> "dependencies or devDependencies"
 
-getOppositePackageJsonDepedencyKey :: PackageRequirement -> Maybe String
-getOppositePackageJsonDepedencyKey = \case
+getOppositePackageJsonDependencyKey :: PackageRequirement -> Maybe String
+getOppositePackageJsonDependencyKey = \case
```
cprecioso, line 171: Just fixing a typo

```diff
-getOppositePackageJsonDepedencyKey
+getOppositePackageJsonDependencyKey
                           ^ here
```

infomiho:
@cprecioso one thing I remember from when I tried implementing npm workspaces is that we no longer need to run the `npm install` for `web-app` and `server` apps, the `npm install` at the root is enough, is that still true? 

EDIT: left some comments, looking good, I'm glad it seems _that simple_, I hope it stays that way.

mage/Dockerfile
```diff
@@ -32,6 +32,10 @@ RUN cd .wasp/build/server && npx prisma generate --schema='../db/schema.prisma'
 # Building the server should come after Prisma generation.
 RUN cd .wasp/build/server && npm run bundle
 
+# Depending on `npm`'s dependency resolution, these folders may or may not exist.
```
infomiho: What is causing this discrepancy? Can we somehow make it deterministic in the way `npm` works?

waspc/ChangeLog.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### 🎉 New Features
+
+- Wasp now uses npm workspaces to manage our generated code. This makes our dependency system more reliable, improves install performance and reduces the size of each project on disk. ([#3159](https://github.com/wasp-lang/wasp/pull/3159))
```
infomiho, line 13: Do we want to list this maybe as a breaking change since people won't be able to upgrade without adding the required fields in their `package.json`?

waspc/src/Wasp/Generator/ServerGenerator.hs
```diff
@@ -129,7 +129,8 @@ genPackageJson spec waspDependencies = do
       (C.asServerFile [relfile|package.json|])
       ( Just $
           object
-            [ "depsChunk" .= N.getDependenciesPackageJsonEntry combinedDependencies,
+            [ "packageName" .= ("@wasp.sh/generated-server-" ++ (if AS.isBuild spec then "build" else "dev")),
```
infomiho: I'd probably extract the package name logic into a separate module e.g. `Wasp.Generator.Workspaces` or something similar. This way we can encapsulate the workspaces implementation details in one place.

waspc/src/Wasp/Generator/Valid/PackageJson.hs
```diff
@@ -58,6 +60,18 @@ validateOptionalDependencies packageJson =
   where
     validateOptional packageSpec = validatePackageJsonDependency packageJson packageSpec Optional
 
+validateWorkspaces :: P.PackageJson -> [GeneratorError]
+validateWorkspaces packageJson =
+  validateFieldValue
+    "package.json"
+    (FieldName ["workspaces"])
+    -- TODO: Add SDK as a workspace.
```
infomiho: Is the comment below related to this TODO? When do we resolve this TODO?

waspc/src/Wasp/Generator/NpmDependencies.hs
```diff
@@ -137,45 +138,38 @@ instance Eq NpmDepsForPackage where
...
---   which describes which dependencies are in conflict.
--- TODO: The comment above and function name are not exactly correct any more,
--- as user deps don't get combined with the wasp deps any more, instead user deps
--- are just checked against wasp deps to see if there are any conflicts, and then
--- wasp deps are more or less returned as they are (maybe with some changes? But certainly no user deps added).
--- This function deserves rewriting / rethinking. This should be addressed while solving
--- GH issue https://github.com/wasp-lang/wasp/issues/1644 .
-combineNpmDepsForPackage :: NpmDepsForWasp -> NpmDepsForUser -> Either NpmDepsForPackageError NpmDepsForPackage
-combineNpmDepsForPackage npmDepsForWasp npmDepsForUser =
+waspDepsToPackageDeps :: NpmDepsForWasp -> NpmDepsForPackage
```
infomiho, line 133: I got confused with the naming of these types, I know it's old code but could you please quickly explain the:
- `NpmDepsForWasp`
- `NpmDepsForPackage`
- `NpmDepsForUser`

types since it's fresh to you now.

mage/Dockerfile
cprecioso: > Can we somehow make it deterministic in the way npm works?

Nope. As soon as the user starts installing dependencies on their project, `npm` will decide where to install stuff.

In general, it prefers to install in the top-level `node_modules` (so the lower-level `node_modules` might not exist), but if it find e.g. two incompatible transitive dependencies deep in the tree, it might decide to install each in the package that needs it (so it will create the lower-level folder).

This is expected, as that's the way `npm` can output correct trees while still being compatible with Node's resolution algorithm; and will not affect our build further than just making sure that the process works with or without that folder.

waspc/src/Wasp/Generator/Valid/PackageJson.hs
cprecioso: Yes the whole comment is the TODO. Is it not clear, how can I make it clear?

> When do we resolve this TODO?

When we upgrade to React 19 (#2482)

waspc/src/Wasp/Generator/NpmDependencies.hs
cprecioso: All these data types have the same general shape, but we're using them for different things, so that's why I supposed we're creating different ones, to get type-safety on this.

- ForWasp: Dependencies we specify for our own packages. Both the `ServerGenerator` and the `WebApp` generator have theirs.
- ForUser: Depdencies in the user's `package.json`.
- ForPackage: The agnostic representation of which dependencies _any package.json_ has. The `SdkGenerator` defines one of these.
- ForFramework: Holds the `NpmDepsForPackage` of both the Server and the WebApp

I am not sure the mess of names is worth the type-safety, but I didn't want to embark on a wider refactor of this.

cprecioso:
@infomiho ready

waspc/src/Wasp/Generator/NpmDependencies.hs
cprecioso: Hmmm I took a stab at it in another PR, wdyt?
https://github.com/wasp-lang/wasp/pull/3178/files#diff-117bae4304f8622274f600ffaa47dc30da81ed63e8d7d7c75f72a3eeefa2330d

waspc/src/Wasp/Generator/Valid/PackageJson.hs
infomiho: > how can I make it clear?

Continue in the same line or say "Read below for more details" or add a colon ":" at the end etc.

waspc/src/Wasp/Generator/Workspaces.hs
```diff
@@ -0,0 +1,18 @@
+module Wasp.Generator.Workspaces
```
infomiho: I just realized my suggested named is a bit too generic, so I'd go for `Wasp.Generator.NpmWorkspaces`

infomiho: Also, is there a way we maybe keep some source of truth related to the list of workspaces here? e.g. list of generated workspaces `[".wasp/out/*", ".wasp/build/*"]` and then we import the list into the validation module. This way we centralise all the important knowledge about npm workspaces into one place.

mage/Dockerfile
infomiho: Ok, this is then business as usual which means we have to cover it in the case it exists and in the case when it does not exist. Okay, this is an okay solution 👍

waspc/src/Wasp/Generator/NpmDependencies.hs
infomiho: The PR is a clear improvement in communicating that the types are basically semantic wrappers which I like. I'll let others chime in as well 👍

waspc/src/Wasp/Generator/Workspaces.hs
cprecioso: Added some type-safety for extra marks :D

waspc/src/Wasp/Generator/NpmWorkspaces.hs
```diff
@@ -0,0 +1,43 @@
...
+import StrongPath (Dir, Path', Rel, reldir, (</>))
+import Wasp.AppSpec (AppSpec, isBuild)
+import Wasp.Generator.Common (ProjectRootDir)
+import Wasp.Project.Common (WaspProjectDir, buildDirInDotWaspDir, dotWaspDirInWaspProjectDir, generatedCodeDirInDotWaspDir)
+
+-- TODO: Add SDK as a workspace:
+-- Currently a overly-zealous resolution makes an incompatible resolution for `@types/react` that
+-- would make the workspace installation fail.
+-- Review when we upgrade React 19 (#2482).
+workspaces :: [Path' (Rel WaspProjectDir) (Dir ())]
```
infomiho: I dislike that we modified the glob pattern from `.../*` to `.../*/` because we introduced type safety. I get that both of the forms are equivalent and work but I'd maybe go the extra mile to make it so that we can use the glob you came up with in the first place.

One way we can do that - instead of returning `[Path' (Rel WaspProjectDir) (Dir ())]` we return `[String]` as the type from this function and inside of it we do something like:

`fromRelDir $ dotWaspDirInWaspProjectDir </> generatedCodeDirInDotWaspDir ++ "*"` (using a `"*"` string instead of  `Path' (Rel ProjectRootDir) (Dir ())` for the wildcard)

Let me know your thoughts on this, I'm on the fence if this is important enough to change.

cprecioso: >  I'm on the fence if this is important enough to change.

I don't think it is. I do prefer it without the `/`, but as you said it works the same. And fixing it would either require us to return un-type-safe `String`, or an un-semantic `Path' ... File`. I like that we can pass paths around with always knowing where they stem from, and I wouldn't like to lose it. I can clearly see both arguments though, so if you feel strongly about this, I'll make the needed changes.

infomiho:
Can we get rid of these warnings when running in dev?
```
[ Client!] npm warn config ignoring workspace config at /Users/ilakovac/dev/wasp/waspc/examples/todoApp/.wasp/out/web-app/.npmrc
[ Server!] npm warn config ignoring workspace config at /Users/ilakovac/dev/wasp/waspc/examples/todoApp/.wasp/out/server/.npmrc
```

infomiho:
I benchmarked it a little bit with the `basic` template running the `wasp compile` command.

**Using workspaces is around 3 seconds faster and it produces around ~50 MB smaller `node_modules`.**

<details>
<summary>Details</summary>

| Setup             | Disk Usage | Time          |
|-------------------|------------|-------------------------|
| **With Workspace**  | **359MB (86.7%)**   | **17.71s (85.6%)**        |
| **Without Workspace** | 414MB       | 20.70s                     |

```
hyperfine --warmup 1 \
  'wasp new benchmark_wasp && cd benchmark_wasp && wasp compile && cd .. && rm -rf benchmark_wasp' \
  'wasp-cli new benchmark_lw && cd benchmark_lw && wasp-cli compile && cd .. && rm -rf benchmark_lw'
Benchmark 1: wasp new benchmark_wasp && cd benchmark_wasp && wasp compile && cd .. && rm -rf benchmark_wasp
  Time (mean ± σ):     20.701 s ±  5.990 s    [User: 19.823 s, System: 10.092 s]
  Range (min … max):   18.536 s … 37.737 s    10 runs
 
  Warning: The first benchmarking run for this command was significantly slower than the rest (37.737 s). This could be caused by (filesystem) caches that were not filled until after the first run. You are already using the '--warmup' option which helps to fill these caches before the actual benchmark. You can either try to increase the warmup count further or re-run this benchmark on a quiet system in case it was a random outlier. Alternatively, consider using the '--prepare' option to clear the caches before each timing run.
 
Benchmark 2: wasp-cli new benchmark_lw && cd benchmark_lw && wasp-cli compile && cd .. && rm -rf benchmark_lw
  Time (mean ± σ):     17.714 s ±  6.238 s    [User: 16.703 s, System: 8.253 s]
  Range (min … max):   14.840 s … 35.005 s    10 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
 
Summary
  wasp-cli new benchmark_lw && cd benchmark_lw && wasp-cli compile && cd .. && rm -rf benchmark_lw ran
    1.17 ± 0.53 times faster than wasp new benchmark_wasp && cd benchmark_wasp && wasp compile && cd .. && rm -rf benchmark_wasp
  ```
</details>

File: waspc/src/Wasp/Generator/NpmWorkspaces.hs (hunk 8 above)
Comment (infomiho) [reply waspc/src/Wasp/Generator/NpmWorkspaces.hs]: On a second thought, since this is a user-facing value i.e. it lives in their `package.json` and we want it to look as natural as possible, I'd go with the initial values without the trailing slash.

**Comment by FranjoMindek** [comment]
> I benchmarked it a little bit with the `basic` template running the `wasp compile` command.

Edited your table a bit
image

File: waspc/src/Wasp/Generator/NpmWorkspaces.hs (hunk 8 above)
Comment (cprecioso) [reply waspc/src/Wasp/Generator/NpmWorkspaces.hs]: Wdyt? I keep the semantic-ness of it until the last moment, and the trailing slash just becomes a thing of rendering the path to the package.json:

[`9454f77` (#3159)](https://github.com/wasp-lang/wasp/pull/3159/commits/9454f77ff4b61fc77b9e4335c56af7e629cdf5b3)

**Comment by cprecioso** [comment]
@infomiho ready

**Comment by infomiho** [comment]
I've noticed one thing while testing it on `ask-the-document` and migrating: I have to put the values in the `workspaces` field in the right order.  I found that slightly annoying since the order doesn't isn't important and we are being stricter than we have to.

What do you say we do something like `set(what they have).includes(set(what we need))`? This is a nitpick, but again, it's user facing so I'm thinking it's worth to make it more flexible?

**Comment by cprecioso** [comment]
@infomiho ready

**Review by infomiho — APPROVED** [review]
Looking good, I'm okay with the current state. I've tested it out locally as well. I'd like someone else from the team to also do a sanity check of the impl since it's a structural change.

File: mage/Dockerfile (hunk 2 above)
Comment (sodic) [reply mage/Dockerfile]: But the top-level folder will always exist, right? If not, we might have a bigger problem because I think other parts of Wasp rely on it existing.

File: waspc/ChangeLog.md [hunk 9]
```diff
@@ -2,8 +2,15 @@
 
 ## Unreleased
 
+### ⚠️ Breaking Changes
+
+Remember to check out the [migration guide](https://wasp.sh/docs/migration-guides/migrate-from-0-18-to-0-19) for step-by-step documentation on how to upgrade.
+
+- Wasp now requires your project's `package.json` to contain `"workspaces": [".wasp/build/*", ".wasp/out/*"]`. ([#3159](https://github.com/wasp-lang/wasp/pull/3159))
```
sodic, line 9: I was hoping we could get away without leaking this to the user.

@cprecioso Do you maybe know how other frameworks do it? Do they use workspaces at all? If so, do they leak it in their package.json files?

waspc/data/Generator/templates/Dockerfile
```diff
@@ -47,6 +47,10 @@ RUN cd .wasp/build/server && npx prisma generate --schema='{= dbSchemaFileFromSe
 # Building the server should come after Prisma generation.
 RUN cd .wasp/build/server && npm run bundle
 
+# Depending on `npm`'s dependency resolution, these folders may or may not exist.
+# We ensure they are created so that the COPY instruction later does not fail.
+RUN mkdir -p /app/node_modules /app/.wasp/build/server/node_modules
```
sodic: None of the other commands use explicit `/app` since it's set by the workdir. How come this one does? It threw me off for a second, I thought something special was going on.

waspc/src/Wasp/Generator/WebAppGenerator.hs
```diff
@@ -123,13 +124,23 @@ genPackageJson spec waspDependencies = do
...
+  -- We do expect users to manually go into the generated directories when bundling the built ouput.
+  -- So we do add the `.npmrc` there to help them avoid using an incompatible Node.js version.
+  | AS.isBuild spec =
+      return
+        [ C.mkTmplFdWithDstAndData
+            (C.asTmplFile [relfile|npmrc|])
+            (C.asWebAppFile [relfile|.npmrc|])
+            Nothing
+        ]
+  | otherwise = return []
```
sodic, line 143: There's a neat pattern with list comprehensions to handle cases like this:
```haskell
genNpmrc :: AppSpec -> Generator [FileDraft]
genNpmrc spec = return [npmrcFileDraft | AS.isBuild spec]
  where
    npmrcFileDraft =
      C.mkTmplFdWithDstAndData
        (C.asTmplFile [relfile|npmrc|])
        (C.asWebAppFile [relfile|.npmrc|])
        Nothing
```
Use it if you like it. Keep your comment of course, I've removed it for brevity.

waspc/src/Wasp/Generator/ServerGenerator.hs
```diff
@@ -180,13 +182,24 @@ npmDepsForWasp spec =
...
+  -- So we do add the `.npmrc` there to help them avoid using an incompatible Node.js version.
+  | AS.isBuild spec =
+      return
+        [ C.mkTmplFdWithDstAndData
+            (C.asTmplFile [relfile|npmrc|])
+            (C.asServerFile [relfile|.npmrc|])
+            Nothing
+        ]
+  | otherwise =
+      return []
```
sodic, line 202: Hmm, any ideas of how we can reduce this duplication without getting too fancy?

waspc/src/Wasp/Generator/NpmDependencies.hs
```diff
@@ -137,45 +138,38 @@ instance Eq NpmDepsForPackage where
...
-          }
+    then Nothing
     else
-      Left $
+      Just $
         NpmDepsForPackageError
           { dependenciesConflictErrors = conflictErrors,
             devDependenciesConflictErrors = devConflictErrors
           }
   where
```
sodic: This is subjective, but I find guards much nicer than if expressions. In fact, we tell the students of the Haskell course we teach to only ever use an if expression if you need its property of being an expression (e.g., to inline it somewhere).

So this is how I'd write it:
```haskell
getNpmDepsConflicts :: NpmDepsForWasp -> NpmDepsForUser -> Maybe NpmDepsForPackageError
getNpmDepsConflicts npmDepsForWasp npmDepsForUser
  | null conflictErrors && null devConflictErrors = Nothing
  | otherwise =
    Just $
      NpmDepsForPackageError
        { dependenciesConflictErrors = conflictErrors,
          devDependenciesConflictErrors = devConflictErrors
        }
```
Not a large difference in this case and, like I said, a little subjective. So do as you please :)

More importantly, this function is a little lacking in terms of its implementation and return type. By the end of itthe function's evaluation, we _know_ that at least one of the returned lists is not empty, and then lose this information.

Since lists already come with an "empty" state (an empty list), the function should have simply return `NpmDepsForPackageErrors` or a more elaborate sum type that better communicates what we know.

waspc/src/Wasp/Generator/NpmWorkspaces.hs
```diff
@@ -0,0 +1,62 @@
...
+import Wasp.Project.Common (WaspProjectDir, buildDirInDotWaspDir, dotWaspDirInWaspProjectDir, generatedCodeDirInDotWaspDir)
+
+-- | Returns the list of workspaces that should be included in the generated package.json file. Each
+-- workspace is a glob that matches all packages in a certain directory.
+-- The path syntax here is always POSIX, because Windows syntax does not allow globs, and `npm`
+-- prefers it.
+--
+-- Order doesn't matter, but we sort the packages to ensure a deterministic order. Otherwise, we
+-- might inadvertently change the order and the compiler will complain about it in user's projects.
+workspaces :: [Path Posix (Rel WaspProjectDir) (Dir ())]
```
sodic: StrongPath has an alias for `Dir ()` which is `Dir'`. But are you sure we don't want `Dir a` or `Dir WorkspaceDir` (the last one assumes we add an extra type)?

waspc/src/Wasp/Generator/NpmWorkspaces.hs
```diff
@@ -0,0 +1,62 @@
...
+serverPackageName :: AppSpec -> String
+serverPackageName = workspacePackageName "server"
+
+webAppPackageName :: AppSpec -> String
+webAppPackageName = workspacePackageName "webapp"
+
+workspacePackageName :: String -> AppSpec -> String
+workspacePackageName baseName spec = "@wasp.sh/generated-" ++ baseName ++ "-" ++ mode
+  where
+    mode = if isBuild spec then "build" else "dev"
```
sodic: Another example where I'd personally [use guards](https://media.tenor.com/eHIRFWRKeQoAAAAe/marge-i-just-think-theyre-neat.png).

waspc/src/Wasp/Generator/NpmWorkspaces.hs
```diff
@@ -0,0 +1,62 @@
...
+  fmap (FP.Posix.dropTrailingPathSeparator . fromRelDirP)
+
+serverPackageName :: AppSpec -> String
+serverPackageName = workspacePackageName "server"
+
+webAppPackageName :: AppSpec -> String
+webAppPackageName = workspacePackageName "webapp"
+
+workspacePackageName :: String -> AppSpec -> String
+workspacePackageName baseName spec = "@wasp.sh/generated-" ++ baseName ++ "-" ++ mode
```
sodic: A nice use case for `intercalate` but might be overkill. You be the judge.

waspc/src/Wasp/Generator/NpmWorkspaces.hs
```diff
@@ -0,0 +1,62 @@
...
+        unlines
+          [ "This should never happen: our paths should always be POSIX-compatible, but they're not.",
+            displayException exception
+          ]
+
+toWorkspacesField :: [Path Posix (Rel WaspProjectDir) (Dir ())] -> [String]
+toWorkspacesField =
+  -- While the trailing slashes do not matter, we drop them because they will be user-visible in
+  -- their `package.json`, and it is more customary without them.
+  fmap (FP.Posix.dropTrailingPathSeparator . fromRelDirP)
```
sodic: The type signature already limits the function to work only on lists, so IMO the polymorphic `fmap` is unnecessary and confusing.

This is a loong running discussion in Wasp btw, so don't feel obliged to change it. If anything, Martin will be happy he finally got someone on his side :)

If you're curious:
- https://github.com/wasp-lang/wasp/pull/1601#discussion_r1431408030
- https://github.com/wasp-lang/wasp/pull/527#discussion_r837410567
- https://github.com/wasp-lang/wasp/pull/1533#discussion_r1424351552

web/docs/migration-guides/migrate-from-0-18-to-0-19.md
```diff
@@ -0,0 +1,49 @@
...
+  "workspaces": [".wasp/build/*", ".wasp/out/*"]
+}
+```
+
+And, to recalculate the dependencies with the new workspace setup, run the following commands in your terminal:
+
+```shell
+wasp clean
+rm package-lock.json
+wasp ts-setup # ONLY if you are using the Wasp TS Config
```
sodic, line 44: Did we test it with the TS config?

waspc/src/Wasp/Generator/NpmWorkspaces.hs
```diff
@@ -0,0 +1,62 @@
...
+-- prefers it.
+--
+-- Order doesn't matter, but we sort the packages to ensure a deterministic order. Otherwise, we
+-- might inadvertently change the order and the compiler will complain about it in user's projects.
+workspaces :: [Path Posix (Rel WaspProjectDir) (Dir ())]
+workspaces =
+  sort
+    [ globFor generatedCodeDirInDotWaspDir,
+      globFor buildDirInDotWaspDir
+      -- TODO: Add SDK as a workspace:
```
sodic: Should we have a dedicated issue for this thing in particular. As far as I can see, we only have the React 19 issue.

waspc/src/Wasp/Generator/NpmWorkspaces.hs
```diff
@@ -0,0 +1,62 @@
...
+      forceRelDirToPosix (dotWaspDirInWaspProjectDir </> dir)
+        </> packageWildcard
+    packageWildcard = [reldirP|*|]
+
+    forceRelDirToPosix = either nonPosixError id . relDirToPosix
+    nonPosixError exception =
+      error $
+        unlines
+          [ "This should never happen: our paths should always be POSIX-compatible, but they're not.",
+            displayException exception
```
sodic: We rarely use this to show errors. How does it look like?

waspc/src/Wasp/Generator/NpmWorkspaces.hs
```diff
@@ -0,0 +1,62 @@
...
+  sort
+    [ globFor generatedCodeDirInDotWaspDir,
+      globFor buildDirInDotWaspDir
+      -- TODO: Add SDK as a workspace:
+      -- Currently a overly-zealous resolution makes an incompatible resolution for `@types/react`
+      -- that would make the workspace installation fail.
+      -- Review when we upgrade React 19 (#2482).
+    ]
+  where
+    globFor dir =
```
sodic: I think we can make this function clearer with a type signature. After all, it can't just take _any_ dir. I believe we have a special type somewhere that captures the values for `build` and `out` dir.

sodic (commented):
I reviewed most of it but have to go now. I'll come back and review the remaining 3 or 4 files.

I pushed my comments so far in hope you'll have something to do by the time I'm back :)

Great work on this, can't wait to merge it!

waspc/ChangeLog.md
cprecioso: The `workspaces` key is our best solution for participating in dependency resolution, without changing our whole generated code architecture. It has to be at the top-level package (the user's project), there's no way around that.

Other frameworks can solve these problems by generating code into the user's project directly, and shipping their runtime dependencies into an npm-published package. The templates &rarr; libs initiatives is already taking steps in this direction, if we want to go all the way in the future

<details>
<summary>Blog-post-length explanation</summary>


### The meta-framework problem
While we allow the user to believe that their project files are the entry point, a meta-framework like ours actually wants to generate their own entry point files (in our case, the `server` and `web-app` projects) that are the ones that are in charge, setup the server or the client, and choose when to call into the user-authored code.

These entry points might also have their own runtime dependencies themselves, that we want to both hide from users (they don't need to manually install them or care about them) and make them participate in the dependency tree resolution (so we participate in dependency deduplication and version resolution).

So we end up with these requirements:
1. We want user-authored code to be the top-level project
2. We want to import user-authored code
3. We want to hide framework dependencies
4. We want to participate in dep resolution 

### Wasp until now
- By generating code in `.wasp/out/`, we satisfy (1). 
- We take advantage of the compiler to locate and generate precise relative paths (`../../../../src/my-entry.ts`), so we satisfy (2).
- By making it a package by itself (it has a `package.json`), it can independently declare its own dependencies, so it satisfies (3).
- **not (4)**, since we were running the installation for the framework dependencies separately from the user project.

The requirement 4 is the one we want to address.
 
### Other frameworks
In general what other frameworks do is:

- For (1) and (2), they generate code straight in the user's project.

They do this by either actually writing files (like TanStack Router/Start), or by generating virtual files at build time with bundler plugins (like Next)

- For (3) and (4), they ship all their possible runtime dependencies as part of an npm-published package, that exports functions for the generated code to call into.

This can be done both in the same package as their compiler (like the `next` package having internal imports), or in a support package (like `prisma` making you install `@prisma/client`).

### This solution
Our main problem is that we're not generating the code in the user's project, but as separate packages.

- We can't write these packages to `node_modules` because `npm` only resolves the dep tree when fetching the packages from the registry, and doesn't look into the written `package.json`s for changes.

- We can't use `file:.wasp/out/server` because that doesn't resolve transitive dependencies.

So then the final option is to tell `npm` that these projects are actually user-authored. This signals to `npm` that they can change their dependency, and that it should look into their `package.json`s on disk to correctly resolve the dep tree. We do that through the `workspaces` key.

The main constraint is that the `workspaces` key has to be in the top-level project. Because we want to maintain the fiction of requirement (1), we have no other way but to write this key into the user project's `package.json`.

### Wasp in the future

For us, the same architecture as other frameworks could be reached by

1. First option
	1. Completely adopting the templates &rarr; libs RFC
	2. Reducing our generated code to the minimum
	3. Writing this minimum-viable generated code to the user project
2. Second option
	1. Creating a package that declares all of the possible dependencies we might need
	2. Moving the Wasp compiler to be a bundler plugin and emit virtual files, using dependencies from that package

</details>

mage/Dockerfile
cprecioso: Yes, it will!

waspc/data/Generator/templates/Dockerfile
cprecioso: I think just a leftover, fixed!

waspc/src/Wasp/Generator/WebAppGenerator.hs
cprecioso: It's cool, but I am not sure that (to my untrained eye) it reads well, since we're not really iterating through a list. I'll keep the guard in this instance, but add that pattern to my toolbox, thanks!

waspc/src/Wasp/Generator/ServerGenerator.hs
cprecioso: What duplication?

waspc/src/Wasp/Generator/NpmDependencies.hs
cprecioso: I ended up removing `NpmDepsForPackageError` altogether in favor of `[DependencyConflictError]`. We weren't using the dependency/devDependency separation anywhere, so a single list can be appropriate, and simplifies the checking to a single `null` call instead of checking the maybe and then each separate list.

cprecioso: And I used guards everywhere in the ensuing refactor :)

waspc/src/Wasp/Generator/NpmWorkspaces.hs
cprecioso: I didn't want to name the type because other code should not be further operating on these paths, they are not real paths but globs. I moved it to `Dir'`.

waspc/src/Wasp/Generator/NpmWorkspaces.hs
cprecioso: I think it is overkill, left it out.

waspc/src/Wasp/Generator/NpmWorkspaces.hs
cprecioso: I'll turn it to `map`. Reading that context, I also realized that sneaky Martin also led me to believe that "all the true Haskellers" use `<>` instead of `++` while I prefer the latter too. 😂

web/docs/migration-guides/migrate-from-0-18-to-0-19.md
cprecioso: Yes! I tested with Waspello, works fine.

waspc/src/Wasp/Generator/NpmWorkspaces.hs
cprecioso: I thought it'd come up when doing the React 19 thing, but you're right. I added a specific issue for it https://github.com/wasp-lang/wasp/issues/3233 and put it into this comment too.

waspc/src/Wasp/Generator/NpmWorkspaces.hs
cprecioso: image

It doesn't seem to be showing anything interesting with the `displayException`, it's apparently just a `show` over an opaque error value. I'll show only our message plus the input that caused the error

cprecioso:
@sodic ready

waspc/ChangeLog.md
Martinsos: Nice write up! I certainly would like to understand this better in the future, I hope we discuss this more.

Couple of quick questions without getting too deep into it right nowm fccused on the very end of your comment, the "Wasp in the future" section:
1. "Writing this minimum-viable generated code to the user project" -> what does that mean, we would actually spam their project with our code? Would it just sit there, or would they need to import it / use it somehow?
2. "Moving the Wasp compiler to be a bundler plugin and emit virtual files, using dependencies from that package" -> what could this look like? Does it mean we need to rewrite it to TypeScript? Does it mean we don't have wasp CLI more? Are we limited for our big viison by being "just" a bundler plugin? Or maybe there is a way to have bundler plugin that calls Wasp CLI or something like that?

cprecioso: Thanks! Just a disclaimer that those are my own opinions on where I'd take Wasp in the future, haven't talked about it with anyone, just ideas (haven't even tried them out).

> 1. [...] what does that mean, we would actually spam their project with our code?

Yeah, maybe, if our generated code becomes small enough to not be a hassle. That's what TanStack does, and it provides you great visibility, and no need for LSP plugins or anything because the code is actually all there.

> 2. [...] Does it mean we need to rewrite it to TypeScript? Does it mean we don't have wasp CLI more? Are we limited for our big viison by being "just" a bundler plugin?

None of this, necessarily. Implementing part of our functionality can just be an implementation detail. The bundler plugin can call out to external programs no problem (e.g. `esbuild`), so our current compiling model doesn't need to change dramatically. (And while there's no need to rewrite to TS, I would argue that we could compile to WASM so no need for native binaries, but even that is optional). The CLI can still exist and call out e.g. `vite` when needed, as an implementation detail.

But implementing the compiler as a bundler plugins could give us a lot of functionality "for free", wide hooks into modifying or adapting the user code unobtrusively; and give users a know quantity for adding functionality as e.g. vite plugins. To drive the point home, large swathes of `next` are implemented as Webpack plugins, but you don't need to know that at all.

Martinsos: You lost me on some points here, but I think got some idea of it. Ok, future discussions!

waspc/src/Wasp/Generator/NpmWorkspaces.hs
Martinsos: They indeed do! You will see them using `<$>` and `<>`, like I also do these days. I was once in the past in the `++` and `map` camp, but I grew out of it. @infomiho and @sodic haven't yet :D.

p.s. `map` and `++` were made for learning really, there isn't much purpose for them when there are `fmap` and `<>`. But they had/have simpler types and are therefore more friendly for beginners (simpler error messages, no typeclasses and stuff in them).

waspc/src/Wasp/Generator/Valid/PackageJson.hs
```diff
@@ -58,6 +62,14 @@ validateOptionalDependencies packageJson =
   where
     validateOptional packageSpec = validatePackageJsonDependency packageJson packageSpec Optional
 
+validateWorkspaces :: P.PackageJson -> [GeneratorError]
```
Martinsos, line 65: Nice job on adding validation!

waspc/src/Wasp/Generator/Valid/Common.hs
```diff
@@ -0,0 +1,106 @@
...
+    validateArrayFieldIncludesRequired,
+    FullyQualifiedFieldName (FieldName),
+  )
+where
+
+import Data.List (intercalate)
+import qualified Data.Set as S
+import Wasp.Generator.Monad (GeneratorError (GenericGeneratorError))
+
+class JsonValue a where
```
Martinsos: I see we are doing some logic in this file for modeling JSON and validating it.

I wonder if we could reuse some existing library for this instead of implementing it on our own?

How about Aeson itself, could it also be used for this? Maybe not, I am not sure, but worth checking.

Extremely quick googling gave me https://hackage.haskell.org/package/forma (by mark karpov, he is serious), and also https://github.com/silkapp/json-schema , maybe worth checking.

cprecioso: I agree with the intention, just a note that this is not new code but extracting our old code for tsconfig validation into a common file. I'd not change this code for now to keep the PR scoped, but I'll create an issue for it (the task sounds fun actually)

cprecioso: #3242

mage/Dockerfile
sodic: Then we don't need the `mkdir` :)

waspc/ChangeLog.md
sodic: This is awesome. @cprecioso Do you mind copying your comment with Martin's points addressed into a notion document. Everyone on the team should read and understand this better (myself included).

waspc/src/Wasp/Generator/ServerGenerator.hs
sodic: This code is essentially the same as the one in WebGenerator, including the comment.

waspc/src/Wasp/Generator/NpmWorkspaces.hs
sodic: Isn't then `Dir a` the preffered option? This is polymorphic and means "Any directory". `Dir ()` and `Dir'` mean "a specific directory I don't care to name".

Also, toWorkspaceField still uses `Dir ()`.

waspc/src/Wasp/Generator/NpmWorkspaces.hs
sodic: > I grew out of it. @infomiho and @sodic haven't yet :D.

I did grow out of Haskell tho. Some day I hope you join me.

@cprecioso Welcome to the right side :)

waspc/ChangeLog.md
cprecioso: Did so at https://www.notion.so/wasp-lang/Fictions-of-a-meta-framework-and-why-we-use-npm-workspaces-28718a74854c800d96d5efbe9cc0f6a8?source=copy_link

waspc/src/Wasp/Generator/NpmWorkspaces.hs
sodic: ... and now we can also change the argument name :)

sodic: Plus, I advise naming functions using verbs:
- `makeGlobFor`
- `makeNonPosixError`

I know we're all serious Haskellers pretending that "yo function is no different than a number, why would it have special treatment" (render in sponge bob mock text). But, in reality, the human mind _does_ consider functions different than regular values, and naming them with verbs helps out.

I won't block on this though, it's a matter of style.

waspc/src/Wasp/Generator/ServerGenerator.hs
cprecioso: Oh I see... Mmmh I can extract maybe to the `NpmWorkspaces.hs` seeing as it is tangentially related? Not 100% on it tho

cprecioso:
@sodic ready

waspc/src/Wasp/Generator/NpmWorkspaces.hs
```diff
@@ -1,54 +1,47 @@
...
 
-    forceRelDirToPosix = either nonPosixError id . relDirToPosix
-    nonPosixError exception =
-      error $
-        unlines
-          [ "This should never happen: our paths should always be POSIX-compatible, but they're not.",
-            displayException exception
-          ]
+    -- We force this to be POSIX because Windows-style paths do not accept wildcard characters.
+    packageWildcard = [SP.reldirP|*|]
```
sodic: This is also something that probably doesn't need to be in StrongPath. When we're constructing wildcards, we're already in string teritorry, so there's no point in pretending we're dealing with real paths, discussing Posix etc.

waspc/src/Wasp/Generator/NpmDependencies.hs
```diff
@@ -150,27 +142,20 @@ waspDepsToPackageDeps npmDepsForWasp =
...
 getNpmDepsConflicts npmDepsForWasp npmDepsForUser =
-  if null conflictErrors && null devConflictErrors
-    then Nothing
-    else
-      Just $
-        NpmDepsForPackageError
-          { dependenciesConflictErrors = conflictErrors,
-            devDependenciesConflictErrors = devConflictErrors
-          }
+  conflictErrors ++ devConflictErrors
```
sodic, line 147: Do we lose something by not separating dev dependency conflicts from regular dependency conflicts?

waspc/src/Wasp/Generator/Valid/Common.hs
```diff
@@ -0,0 +1,106 @@
...
+
+instance Show FullyQualifiedFieldName where
+  show (FieldName fieldPath) = intercalate "." fieldPath
+
+validateRequiredField :: (Eq a, JsonValue a) => String -> FullyQualifiedFieldName -> Maybe a -> a -> [GeneratorError]
+validateRequiredField fileName fullyQualifiedFieldName fieldValue expectedValue =
+  validateFieldValue fileName fullyQualifiedFieldName (Just expectedValue) fieldValue
+
+validateFieldValue :: (Eq a, JsonValue a) => String -> FullyQualifiedFieldName -> Maybe a -> Maybe a -> [GeneratorError]
+validateFieldValue fileName fullyQualifiedFieldName expectedValue actualValue =
```
sodic: When doing refactors, it's best to do it in a PR different than the one introducing a feature.
If a refactor includes both code migration and functional changes, then those two should be separated as well.
I'll leave [another pointer](https://mtlynch.io/code-review-love/#6-separate-functional-and-non-functional-changes) to this great article :)

This PR includes:
- NPM workspaces changes
- A refactor of validation modules with both:
  - Functional changes
  - Moving code around

Since Git treats everything as new code, changes to logic are extremely easy to miss. Examples:
- I think `validateArrayFieldIncludesRequired` is new.
- The `fileName` argument in this function is also new.
- I think the rest of it is just moved.

This results in:
- Properly managing context and focusing attention on new changes is more difficult, so most reviewers will only do a shallow review of both the refactor and the new change.
- If the refactor needs changing, NPM workspaces have to wait.
- The review takes longer because there's more context to manage and more things to comment on. Specifically, half of my time went to the workspaces, the other half went to these three files.

I've taken a look at the refactor and have some reservations:
- A lot of the things in `Common` aren't common. Most of it is used either by one module or the other, the only common thing being the `JsonValue` type class.
- It seems like the refactor is half done and we could go for more reusability. Only a single function in the package.json validation does things the "TS Config" way, all others do it "The old PackageJson way" without using `JsonValue` (i.e., there's little overlap between the two call stacks). I didn't go too deep, but that seems fishy.
- If we're going down the JSONValue route, I think a better approach is deriving `instance JsonValue a => JsonValue [a]` than doing `(JsonValue [a]) => ...` in a type signature.
- etc.  

So, since I have more reservations about the refactor than I have for the worksapces part, instead of discussing the refactor here, I suggest we move it out of this PR and into a different one? Or is it instrumental to the change?
I took a look and it doesn't seem instrumental, but I could be wrong.

I approve everything else btw (the workspaces I mean). Once this has been taken care of, you can merge without me :)

cprecioso: @sodic I updated the validation system at https://github.com/wasp-lang/wasp/pull/3262, once that is approved I can use that system here.

waspc/src/Wasp/Generator/Valid/Common.hs
cprecioso: @Martinsos I updated the validation system at https://github.com/wasp-lang/wasp/pull/3262, once that is approved I can use that system here.

waspc/src/Wasp/Generator/Valid/PackageJson.hs
```diff
@@ -41,7 +44,7 @@ validateRuntimeDependencies packageJson =
 
 validateDevelopmentDependencies :: P.PackageJson -> [GeneratorError]
 validateDevelopmentDependencies packageJson =
-  concat $
+  concat
```
cprecioso, line 48: Fixing a linter warning

cprecioso:
@sodic ready for re-review

- Undid the validator refactor and made an smaller change for validation
- When defining the globs, convert from StrongPath to regular String earlier to simplify code
- Turned `PackageJson.workspaces` into a `Set` as it fits better how I'm checking it
- Updated `main` and solved conflicts

waspc/src/Wasp/Generator/Valid/PackageJson.hs
```diff
@@ -58,6 +61,34 @@ validateOptionalDependencies packageJson =
...
+
+    makeWrongWorkspacesError definedWorkspaces =
+      GenericGeneratorError $
+        unwords
+          [ "Wasp requires \"workspaces\" to include",
+            showSet NW.workspaceGlobs,
+            "but is missing",
+            showSet $ NW.workspaceGlobs `S.difference` definedWorkspaces,
+            "in package.json."
+          ]
```
sodic:
```suggestion
          [ "Wasp requires package.json \"workspaces\" to include",
            showSet NW.workspaceGlobs,
            ", but",
            showSet $ NW.workspaceGlobs `S.difference` definedWorkspaces,
            "is missing."
          ]
```
This sounds a little more natural. But the entire message might leave users confused (because of the arrays surrounding the output).

image


I recommend saying:
```
- Wasp requires package.json "workspaces" to include: ".wasp/build", ".wasp/out". You are missing: ".wasp/build".

File: waspc/src/Wasp/ExternalConfig/Npm/PackageJson.hs [hunk 29]
```diff
@@ -14,14 +14,16 @@ where
...
 import GHC.Generics (Generic)
 import Wasp.ExternalConfig.Npm.Dependency (Dependency)
 import qualified Wasp.ExternalConfig.Npm.Dependency as D
 
 data PackageJson = PackageJson
   { name :: !String,
     dependencies :: !DependenciesMap,
-    devDependencies :: !DependenciesMap
+    devDependencies :: !DependenciesMap,
+    workspaces :: !(Maybe (Set String))
```
sodic: How come we went with a set here?

This structure should mirror how the file works, not how we want it to work. So I believe this should be an array. If we need the set, we should introduce it in the processing.

waspc/src/Wasp/Generator/Valid/PackageJson.hs
```diff
@@ -58,6 +61,34 @@ validateOptionalDependencies packageJson =
...
+
+    missingWorkspacesError =
+      GenericGeneratorError $
+        unwords
+          [ "Wasp requires \"workspaces\" to be present with the value",
+            showSet NW.workspaceGlobs,
+            "in package.json."
+          ]
```
sodic:
```suggestion
          [ "Wasp requires package.json \"workspaces\" to be present and include values:",
            showSet NW.workspaceGlobs, -- we do a comma separated list without the brackets.
          ]
```
Slighly more correct and readable I think.

sodic: Also, do we want to allow them to have extra workspaces? Perhaps it's simpler keeping it airtight for now: You must have this and only this.

sodic (approved):
All good!

I left a few comments with suggested changes on the messages. Please consider them.
I don't want to dwell on this for too long since we're refactoring it soon.

You don't need to ping me again before merging :)

waspc/src/Wasp/Generator/Valid/PackageJson.hs
cprecioso: I initially went for not allowing it, but Miho asked me to allow it.

PR #3162, opened 2025-09-17
Loosen linting-focused Typescript config options

infomiho:
### Error preventing us from deploying an app

After we updated to Vite 7 and our `tsconfig.json` extends their config `@tsconfig/vite-react/tsconfig.json`, we run into some `tsc` errors when we build the `web-app`. 

We got this when building the `ask-the-documents` app:
```ts
src/router.tsx:5:1 - error TS6133: 'createAuthRequiredPage' is declared but its value is never read.

5 import createAuthRequiredPage from "./auth/pages/createAuthRequiredPage"
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error.
```

### We want to ignore those errors

Sometimes there will be some unused imports or params in the generated code (it's easier to write the Mustache templates like that sometimes) - so we want to keep the Typescript rules loose.

Discussed with @sodic on a call and agreed to go with this solution for now. In the future, most of the templated code will in the libs which will be strict, so this should be a okay fix for now.

### DIFF
--- a/waspc/data/Generator/templates/react-app/tsconfig.app.json
+++ b/waspc/data/Generator/templates/react-app/tsconfig.app.json
@@ -5,11 +5,19 @@
     // Temporary loosen the type checking until we can address all the errors.
     "jsx": "preserve",
     "allowJs": true,
-    "strict": false,
     "skipLibCheck": true,
     // Allow importing pages with the .tsx extension.
     "allowImportingTsExtensions": true,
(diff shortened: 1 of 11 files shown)


sodic (approved):
Nice, thank you

PR #3168, opened 2025-09-18
Correctly set dir to serve in wasp-app-runner

cprecioso:
Made as part of
- #3159

---

There is a strange interaction when running `npx` with a custom `cwd`, pointing to a folder inside of a npm workspace. It's a mouthful.

But basically, doing `npx serve .` with `cwd=.wasp/build/web-app/build`, actually searches for the parent package and turns the cwd into it (so `cwd=.wasp/build/web-app`), and thus `serve` doesn't serve the correct directory.

To workaround this npm bug, we explicitly tell `npx serve` the location of the folder we want it to run.

---

I send this PR ahead of time because the `wasp-app-runner` used for E2E examples is taken from `npm`, so we should do this change first and then publish so that #3159 can pass the tests.

### DIFF
--- a/wasp-app-runner/src/build/client.ts
+++ b/wasp-app-runner/src/build/client.ts
@@ -64,11 +64,13 @@ async function startClientApp({
 }): Promise<{
   exitCode: number | null;
 }> {
+  const servePath = path.resolve(pathToApp, clientAppBuildOutputDir);
+
   return spawnWithLog({
     name: "client-start-app",
     cmd: "npx",
-    args: ["serve", "--single", "-p", "3000"],
-    cwd: path.join(pathToApp, clientAppBuildOutputDir),
+    args: ["serve", "--single", "-p", "3000", servePath],
+    cwd: servePath,
   });
 }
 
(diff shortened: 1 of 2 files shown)


wasp-app-runner/src/build/client.ts
```diff
@@ -64,11 +64,13 @@ async function startClientApp({
...
 }> {
+  const servePath = path.resolve(pathToApp, clientAppBuildOutputDir);
+
   return spawnWithLog({
     name: "client-start-app",
     cmd: "npx",
-    args: ["serve", "--single", "-p", "3000"],
-    cwd: path.join(pathToApp, clientAppBuildOutputDir),
+    args: ["serve", "--single", "-p", "3000", servePath],
+    cwd: servePath,
```
infomiho: Can we replace `npx serve` with `npm run preview` since this is what you ended up doing in `wasp start build`? (even better, we just call `wasp start build` in the future)?

infomiho: It would be `npm run preview -- --port 3000 --strictPort` I guess

PR #3177, opened 2025-09-19
Add --image argument to wasp start db command

cprecioso:
Fixes #1414

Adds an optional `--image` argument to the `wasp start db` command, allowing users to specify a custom Docker image for their development database.

This is part of #3174. We're adding `ask-the-documents` to the repo, which needs a specific PostgreSQL extension. This way we can more easily test the app locally.

## Changes

- Added `--image` argument parsing
- Modified `wasp start db` to accept and use custom Docker images
- Added documentation highlighting the extensions use case

## Usage

```bash
# Use default PostgreSQL image (uses `postgres`)
wasp start db

# Use specific PostgreSQL version
wasp start db --image postgres:15

# Use PostgreSQL with PostGIS extension
wasp start db --image postgis/postgis:14-3.2

# Use PostgreSQL with pgvector extension
wasp start db --image pgvector/pgvector:pg16
```

> [!NOTE]
> 🤖 Generated with [Claude Code](https://claude.ai/code)  
> and 👨🏻 reviewed by a person

Issue #169, opened 2021-02-03
Implement `wasp deploy`


Martinsos:
Right now deployment of Wasp apps is pretty manual -> you generate code with `wasp build` and then you have to manually deploy different parts (frontend, backend), the procedure depending on which hosting provider you are deploying to.

Ideally, we want the experience to be the following:
1. In .wasp, you declaratively define your deployment, e.g.:
```
deployment production {
  client: {
    provider: netlify,
    netlify: { provider specific settings },
    ... // settings that are not specific per provider.
  },
  server: {
    provider: heroku,
    heroku: { provider specific settings },
    ... // settings that are not specific per provider.
  }
}
```
2. To deploy, you call `wasp deploy`, which then based on the declarative code in .wasp decides where and how to deploy the web app. It would actually call `wasp build` and then call commands needed to deploy the code. Ideally it would utilize Terraform for this, instead of us implementing logic that checks if something already exists, creating if it does not, and so on.

I tried coming up with Terraform script for deploying backend to Heroku, here is one version of it that works. I had to put it level above the whole project, otherwise it would complain about the checksum of the project (because it in the project itself, and probably modifying state.tf, which was modifying the checksum). But the main problem was that deploying was really slow -> the only support for deploying containers to Heroku was via their heroku.yml feature, where Heroku builds image on their own on their servers, which is super slow (7 minutes). Building image on our machine and then pushing it to Heroku container registry is not supported in Terraform Heroku provider nor will it be because it is not part of Heroku API.
Solution might be for us to implement this part, deploying image to Heroku, on our own - I guess Terraform must have some way of providing custom parts like this.

<details>
  <summary>
    Terraform script (main.tf)
  </summary>

```tf  
terraform {
  required_providers {
    heroku = {
      source = "heroku/heroku"
      version = "3.2.0"
    }
  }
}

provider "heroku" {
  # Configuration options
}

variable "test_app" {
  description = "Test app"
}

resource "heroku_app" "server" {
  name = var.test_app
  region = "us"
  stack = "container" # This means we are using Docker.
}

# Build code & release to the app
resource "heroku_build" "server" {
  app = heroku_app.server.name

  # NOTE: Assumes heroku.yml exists and it cointains build.docker instructions.

  source = {
    path = "out"
  }
  # TODO: Add `version` field?
}

resource "heroku_formation" "server" {
  app        = heroku_app.server.name
  type       = "web"
  quantity   = 1
  size       = "free"
[...]

PR #3182, opened 2025-09-22
Add `--db-image` argument to `wasp start db` command

### LINKED ISSUE #1414: Enable specifying custom Docker image for managed dev database `wasp db start`
This would allow users to use custom postres docker images!

https://discord.com/channels/686873244791210014/1080864617347162122/1145605557307310101 .

cprecioso:
- Fixes #1414
- This PR was previously merged as #3177 before it was ready, and rapidly reverted

Adds an optional `--db-image` argument to the `wasp start db` command, allowing users to specify a custom Docker image for their development database.

This is part of #3174. We're adding `ask-the-documents` to the repo, which needs a specific PostgreSQL extension. This way we can more easily test the app locally.

## Changes

- Added `--db-image` argument parsing
- Modified `wasp start db` to accept and use custom Docker images
- Added documentation highlighting the extensions use case

## Usage

```bash
# Use default PostgreSQL image (uses `postgres`)
wasp start db

# Use specific PostgreSQL version
wasp start db --db-image postgres:15

# Use PostgreSQL with PostGIS extension
wasp start db --db-image postgis/postgis:14-3.2

# Use PostgreSQL with pgvector extension
wasp start db --db-image pgvector/pgvector:pg16
```

> [!NOTE]


cprecioso:
Tested with PostGIS and pgvector images, both work fine!

waspc/ChangeLog.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🔧 Small improvements
+
+- You can now specify which PostgreSQL image to use in `wasp start db` with the `--image` argument. ([#3182](https://github.com/wasp-lang/wasp/pull/3182))
```
sodic: I'd list this as a new feature, not a small improvement.

It may not be useful to everybody but it is a feature :)

waspc/cli/exe/Main.hs
```diff
@@ -176,7 +176,9 @@ printUsage =
         cmd   "    uninstall             Removes Wasp from your system.",
         title "  IN PROJECT",
         cmd   "    start                 Runs Wasp app in development mode, watching for file changes.",
-        cmd   "    start db              Starts managed development database for you.",
+        cmd   "    start db [--image <image>]",
```
sodic: We're missing this help text when we run `wasp db`. Plus, we could include some examples for `wasp start`:

image

web/docs/data-model/databases.md
```diff
@@ -87,6 +87,30 @@ Also, make sure that:
 In case you might want to connect to the dev database through the external tool like `psql` or [pgAdmin](https://www.pgadmin.org/), the credentials are printed in the console when you run `wasp db start`, at the very beginning.
 :::
 
+##### Adding extensions to the dev database {#custom-database}
+
+You can also specify a custom Docker image to use for the database with the `--image` option. This is particularly useful when you need PostgreSQL with specific extensions (like PostGIS for geographic data, pgvector for embeddings, etc.):
+
+```bash
+# Use default PostgreSQL image
```
sodic: I think we should say what the default is, maybe even communicate it as:
> `wasp start db` is a shorthand for `wasp start db -- image postgres`

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -28,12 +32,21 @@ import Wasp.Project.Env (dotEnvServer)
...
 -- Wasp creates it and connects the Wasp app with it.
 -- Wasp is smart while doing this so it checks which database is specified
 -- in Wasp configuration and spins up a database of appropriate type.
-start :: Command ()
-start = do
+start :: Arguments -> Command ()
+start args = do
+  startDbArgs <-
+    parseArguments "wasp start db" startDbArgsParser args
+      & either (E.throwError . CommandError "Invalid arguments") return
```
sodic, line 45: No need to change anything here, just FYI.

I rarely use `&` in Haskell. I believe it's more idiomatic doing stuff from right to left (which is unfortunate) so I'm not used to seeing this.

Here's how I would probably write this:
```hs
start args = do
  startDbArgs <-
    either (E.throwError . CommandError "Invalid arguments") return $
      parseArguments "wasp start db" startDbArgsParser args
```
Or even like this:
```hs
start args = do
  startDbArgs <- case parseArguments "wasp start db" startDbArgsParser args of
    Left err -> E.throwError $ CommandError "Invalid arguments" err
    Right parsedArgs -> return parsedArgs
```
You can pick whatever you like the most, all options are perfectly valid.

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -82,24 +106,31 @@ throwIfCustomDbAlreadyInUse spec = do
...
+               then [" ℹ Using custom Docker image: " <> dockerImage, ""]
+               else []
+           )
+        ++ [ "Additional info:",
+             " ℹ Connection URL, in case you might want to connect with external tools:",
+             "     " <> connectionUrl,
+             " ℹ Database data is persisted in a docker volume with the following name"
+               <> " (useful to know if you will want to delete it at some point):",
+             "     " <> dockerVolumeName
+           ]
```
sodic: At this point (and perhaps even at an earlier one), I think it makes sense to extract these messages into their own variables inside the `where` block. There's a lot of nesting.

@cprecioso What do you think?

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -82,24 +106,31 @@ throwIfCustomDbAlreadyInUse spec = do
...
-startPostgreDevDb :: Path' Abs (Dir WaspProjectDir) -> String -> Command ()
-startPostgreDevDb waspProjectDir appName = do
+startPostgreDevDb :: Path' Abs (Dir WaspProjectDir) -> String -> Maybe String -> Command ()
+startPostgreDevDb waspProjectDir appName customImage = do
   throwIfExeIsNotAvailable
     "docker"
     "To run PostgreSQL dev database, Wasp needs `docker` installed and in PATH."
   throwIfDevDbPortIsAlreadyInUse
 
+  let dockerImage = fromMaybe "postgres" customImage
```
sodic: I believe we should do this part outside of the calling function and send the final image in as an argument:
```hs
startPostgresDevDb :: Path' Abs (Dir WaspProjectDir) -> String -> String -> Command ()
startPostgresDevDb waspProjectDir appName image = do
```

There isn't really a reason for this function to take in an optional.
Also, I see the name was changed from "postgres" to "postgre." I believe this is incorrect: the official name of the project is "PostgreSQL" and its nickname is "Postgres." ([wikipedia](https://en.wikipedia.org/wiki/PostgreSQL)).

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -82,24 +106,31 @@ throwIfCustomDbAlreadyInUse spec = do
...
-        " ℹ Connection URL, in case you might want to connect with external tools:",
-        "     " <> connectionUrl,
-        " ℹ Database data is persisted in a docker volume with the following name"
-          <> " (useful to know if you will want to delete it at some point):",
-        "     " <> dockerVolumeName
+        ""
       ]
+        ++ ( if isJust customImage
+               then [" ℹ Using custom Docker image: " <> dockerImage, ""]
+               else []
```
sodic: I don't think we need any special text here. We should always say "Using Docker image ...."

The default case should just be a default value for something, not a special procedure. This means:
- We set the default value at the top of the "system."
- Everything else is completely agnostic on whether it's working with the default value or the custom value.

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -28,12 +32,21 @@ import Wasp.Project.Env (dotEnvServer)
 import Wasp.Util (whenM)
 import qualified Wasp.Util.Network.Socket as Socket
 
+-- | Command-line arguments for `wasp start db`
+data StartDbArgs = StartDbArgs
+  { customDbImage :: Maybe String
+  }
```
sodic: Since we have all this parsing logic in the same file, I suggest moving this datatype below `start`.

Our rule is:
1. Exported stuff before private stuff
2. If `A` is used in `B`, `A` should come as soon as possible after `B` without breaking the first rule (I think Clean Code has a nice name for this rule, can't remember now)

Unfortunately, many files don't stick to this rule since we can't automate it (at least I think we can't) and sometimes forget. But when we remember, we should follow it :)

waspc/cli/exe/Main.hs
```diff
@@ -176,7 +176,9 @@ printUsage =
...
+              "                          Starts managed development database for you.",
+              "                          Optionally specify a custom Docker image with --image (default: postgres).",
```
Martinsos: So I know that `--image` should be enough because there is context of "db" from `wasp db start`, but, in other PRs you went with `--db-image`, and I am tempted to say we should do the same here. It helps with unifying the CLI experience, and makes it super clear what is this image.

This is more of a nitpick kind of, but I genuinely think it would be a DX.

raw

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -28,12 +32,21 @@ import Wasp.Project.Env (dotEnvServer)
 import Wasp.Util (whenM)
 import qualified Wasp.Util.Network.Socket as Socket
 
+-- | Command-line arguments for `wasp start db`
```
Martinsos:
```suggestion
-- | Command-line arguments for `wasp start db`.
```

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
Martinsos: Do we have super strict rule for this? If we do I have to admit I forgot it.
My rule is, make it read like a story (Clean Code also). Almost always that will match what you wrote above, but there could be some exceptions.

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -28,12 +32,21 @@ import Wasp.Project.Env (dotEnvServer)
...
+data StartDbArgs = StartDbArgs
+  { customDbImage :: Maybe String
```
Martinsos: I would erither go with `dbImage` so it reflects the argument on the CLI, or I would go full exlanatory, so `customDbDockerImage`.

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
Martinsos: Opposite opinion: I think the current code with the `&` is the best one.

While `&` is not that popular as `$`, I don't see any reason why one would not use it in appropriate places("appropriate" being the key here).

In Haskell, you anyway read both from left to right and right to left, it depends on the situation, so there is no golden rule here.

It does make sense to stick with how people normally write Haskell, so I agree one shoudln't go crazy with `&`, but here I think its usage is perfect, because it allows you to read code in the right order -> business logic is in focus, error handling comes after (and is even nicely indented). So I can read it as `do this, but if error this` and I think that is perfect.

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -42,12 +55,23 @@ start = do
...
   case ASV.getValidDbSystem appSpec of
     AS.App.Db.SQLite -> noteSQLiteDoesntNeedStart
-    AS.App.Db.PostgreSQL -> startPostgreDevDb waspProjectDir appName
+    AS.App.Db.PostgreSQL -> startPostgreDevDb waspProjectDir appName (customDbImage startDbArgs)
   where
     noteSQLiteDoesntNeedStart =
       cliSendMessageC . Msg.Info $
         "Nothing to do! You are all good, you are using SQLite which doesn't need to be started."
 
+-- | Parser for `wasp start db` arguments
```
Martinsos: Redundant comment

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -42,12 +55,23 @@ start = do
...
+startDbArgsParser :: Opt.Parser StartDbArgs
+startDbArgsParser =
+  StartDbArgs
+    <$> Opt.optional
+      ( Opt.strOption $
+          Opt.long "image"
+            <> Opt.metavar "IMAGE"
+            <> Opt.help "Docker image to use for the database (default: postgres)"
```
Martinsos: Question: could we have make this default here not be hardcoded?

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -82,24 +106,31 @@ throwIfCustomDbAlreadyInUse spec = do
     throwCustomDbAlreadyInUseError msg =
       E.throwError $ CommandError "You are using custom database already" msg
 
-startPostgreDevDb :: Path' Abs (Dir WaspProjectDir) -> String -> Command ()
-startPostgreDevDb waspProjectDir appName = do
+startPostgreDevDb :: Path' Abs (Dir WaspProjectDir) -> String -> Maybe String -> Command ()
+startPostgreDevDb waspProjectDir appName customImage = do
```
Martinsos:
```suggestion
startPostgreDevDb waspProjectDir appName customDbDockerImage = do
```

web/docs/data-model/databases.md
Martinsos:
```suggestion
# use default PostgreSQL image
```

same goes for other comments below. EIther capital letter + period, or no capital letter.

web/docs/data-model/databases.md
```diff
@@ -87,6 +87,30 @@ Also, make sure that:
 In case you might want to connect to the dev database through the external tool like `psql` or [pgAdmin](https://www.pgadmin.org/), the credentials are printed in the console when you run `wasp db start`, at the very beginning.
 :::
 
+##### Adding extensions to the dev database {#custom-database}
```
Martinsos: It's a wrong title, isn't it? This should about using a custom docker image, while extensions are just one use case (e.g. you might want to use older postgres version).

web/docs/data-model/databases.md
```diff
@@ -87,6 +87,30 @@ Also, make sure that:
...
+# Use PostgreSQL with PostGIS extension for geographic data
+wasp start db --image postgis/postgis:14-3.2
+
+# Use PostgreSQL with pgvector extension for AI embeddings
+wasp start db --image pgvector/pgvector:pg16
+```
+
+:::note
+
+The images you use must respect the `POSTGRES_DB`, `POSTGRES_USER`, and `POSTGRES_PASSWORD` environment variables, as Wasp relies on them to set up the database. An easy way to ensure this is to use images based on the official [PostgreSQL Docker image](https://hub.docker.com/_/postgres).
```
Martinsos: I have to admit I am not sure what does it mean they "respect" those vars. Is there a more explicit way to put this?

waspc/cli/exe/Main.hs
cprecioso: Include some examples for `wasp start` in `wasp db`?

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
cprecioso: In this case it's written like that for consistency, like in here:

https://github.com/wasp-lang/wasp/blob/1129a070f29d7d42827ae7d5f9e9a0ae2514e9b7/waspc/cli/src/Wasp/Cli/Command/CreateNewProject.hs#L32-L36

I think a good idea in general for this would be to start using Options.Applicative for all of our CLI argument parsing, not just occasionally.

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
cprecioso: Actually the nesting was basically unnecessary, now it's all in the top level. I don't really like extracting it somewhere else, and now it's simpler, so I think I'll leave it as-is.

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
cprecioso: Huh weird thing about the Postgres thing, I didn't realize!

cprecioso:
@Martinsos 

> I wonder if it would make sense, from the DX perspective, to actually also have a central place to define a default database image

Yes, I agree. This also makes me feel like what @infomiho once said about having a deployment config file might also make sense. As you said, I think it might be out-of-scope for this PR, but we could start looking into it. (If/when we conclude that custom DB images are a feature that people do use)

waspc/cli/exe/Main.hs
cprecioso: I do think that having the command be unequivocally about `db` gives it enough context not to be confused. But you're right that it might be nice to have it all consistent. I'll make the change.

cprecioso:
@sodic @Martinsos ready

Martinsos:
> @Martinsos
> 
> > I wonder if it would make sense, from the DX perspective, to actually also have a central place to define a default database image
> 
> Yes, I agree. This also makes me feel like what @infomiho once said about having a deployment config file might also make sense. As you said, I think it might be out-of-scope for this PR, but we could start looking into it. (If/when we conclude that custom DB images are a feature that people do use)

I do think they are needed for sure, the custom DB images. But sure more data will be helpful.

There is one older related issue: https://github.com/wasp-lang/wasp/issues/169 . And I know Miho did some work on it in the meantime where i also aprticipated but can't find it right now, I asked him to link it there.

EDIT: I also created issue for it here: https://github.com/wasp-lang/wasp/issues/3193 .

waspc/cli/exe/Main.hs
```diff
@@ -238,21 +240,25 @@ printDbUsage =
...
-              "                If there are multiple seeds, you can specify a seed to execute by providing its name,",
-              "                or if not then you will be asked to provide the name interactively.",
         cmd $ intercalate "\n" [
-              "  migrate-dev   Ensures dev database corresponds to the current state of schema(entities):",
-              "                  - Generates a new migration if there are changes in the schema.",
-              "                  - Applies any pending migrations to the database either using the",
-              "                    supplied migration name or asking for one.",
+              "  start [--db-image <image>]   Alias for `wasp start db`.",
+              "                               Starts managed development database for you.",
+              "                               Optionally specify a custom Docker image."
```
Martinsos, line 246: Not the same as for `wasp start db` (missing "with --db-image). I would add it here or remove it up there.

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -32,22 +36,42 @@ import qualified Wasp.Util.Network.Socket as Socket
...
 
+startDbArgsParser :: Opt.Parser StartDbArgs
+startDbArgsParser =
+  StartDbArgs
+    <$> Opt.strOption
+      ( Opt.long "db-image"
+          <> Opt.metavar "IMAGE"
+          <> Opt.help "Docker image to use for the database"
+          <> Opt.showDefault
+          <> Opt.value "postgres"
```
Martinsos: Is this default value? If so, you should not have it hardcoded, same as in that other place.

waspc/ChangeLog.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🎉 New Features
+
+- You can now specify which PostgreSQL image to use in `wasp start db` with the `--db-image` argument. ([#3182](https://github.com/wasp-lang/wasp/pull/3182))
```
Martinsos, line 7: Maybe also explain right here why this is useful (extensions). Although they can click. But still you coudl explain it shortly. Ok not sure, just an idea.

raw

web/docs/data-model/databases.md
```diff
@@ -87,6 +87,32 @@ Also, make sure that:
 In case you might want to connect to the dev database through the external tool like `psql` or [pgAdmin](https://www.pgadmin.org/), the credentials are printed in the console when you run `wasp db start`, at the very beginning.
 :::
 
+##### Customising the dev database {#custom-database}
```
Martinsos, line 90:
```suggestion
##### Customising the dev database docker image {#custom-database-image}
```

Suggestion, you don't have to take it.

It is just more precise, less likely to colide with some other concept of customizing the datase.

web/docs/data-model/databases.md
```diff
@@ -87,6 +87,32 @@ Also, make sure that:
...
+
+You can also specify a custom Docker image to use for the database with the `--db-image` option. This is particularly useful when you need PostgreSQL with specific extensions (like PostGIS for geographic data, pgvector for embeddings, etc.), or an older version. By default, Wasp uses the latest official PostgreSQL image.
```
Martinsos: latest official postgreSQL image -> what if this text linked to it on DockerHub? That could be helpful I think.

web/docs/data-model/databases.md
```diff
@@ -87,6 +87,32 @@ Also, make sure that:
...
+# Use PostgreSQL with PostGIS extension for geographic data:
+wasp start db --db-image postgis/postgis:14-3.2
+
+# Use PostgreSQL with pgvector extension for AI embeddings:
+wasp start db --db-image pgvector/pgvector:pg16
+```
+
+:::note
+
+The images you specify must use the `POSTGRES_DB`, `POSTGRES_USER`, and `POSTGRES_PASSWORD` environment variables to configure themselves. Wasp will use those values when connecting to the database. We recommend basing your images on the official [PostgreSQL Docker image](https://hub.docker.com/_/postgres), as it automatically uses these environment variables to set up the database name, user, and password.
```
Martinsos:
```suggestion
The custom docker image you specify must use the `POSTGRES_DB`, `POSTGRES_USER`, and `POSTGRES_PASSWORD` environment variables when configuring the database. Wasp will use those values when connecting to the database. We recommend basing your image on the official [PostgreSQL Docker image](https://hub.docker.com/_/postgres), as it automatically uses these environment variables to set up the database name, user, and password.
```

Martinsos (approved):
Couple of comments, but once you resolve them, I am good with merging!

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
cprecioso: Not hardcoded? Where should I get it from? Or do you just mean to move it to its own top-level constant?

Martinsos: Ohhh yes sorry it is ok -> my mind was left in the state where I thought there are other places where you need this default value of `postgres`, like in error messages or something, but you don't anymore becuase you define it here and then it just flows.
All good!

Although, what does feel tempting, is to make it a bit more "serious". This "postgres" here, next to other stuff, it is kind of easy to miss it as just another part of CLI command information. I would be tempted to move its definition to sometihn glike `defaultDockerImageForPostgreDatabase :: String` and put that somewhere in the `Wasp.Db.Postgres` (if we have such module hah can't remember), and then use it here.
But I am not sure, this is just an idea, maybe it is too much. I will leave it to you, do as you like, I think you have a better feel at this point.

raw

waspc/cli/exe/Main.hs
cprecioso: I don't get this comment, can you clarify? I do handle the argument in both cases.

waspc/ChangeLog.md
cprecioso: I think I will leave it as-is, we don't do much of that in the changelog.

web/docs/data-model/databases.md
cprecioso: We already mention Docker first thing in this section. I don't especially feel like mentioning Docker in the header because some of our users might not know or care that we're using Docker for the dev database and might not think this can help them.

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
cprecioso: I did extract it, also added a goodie in the shape of a `type DockerImageName = String` to be clearer.

waspc/cli/exe/Main.hs
Martinsos: Aha sorry, I should have been more verbose! So CLI docs for `wasp start db` are very similar, but a bit different, and there is no reason for them to be different, so I was suggesting to make them the same. THe difference is in "with --db-image" part).

cprecioso: ahh understood, my eyes were skipping over that haha

PR #3183, opened 2025-09-22
Add `--db-image` argument to `wasp-app-runner`

cprecioso:
- On top of #3182

Adds an optional `--db-image` argument to the `run-wasp-app` command, allowing us to specify a custom Docker image for the database.

This is part of #3174. We're adding `ask-the-documents` to the repo, which needs a specific PostgreSQL extension. This way we can more easily test the app.

## Changes

- Modified `run-wasp-app` to accept and use custom Docker images
- Added explanation to README

> [!NOTE]

### DIFF
--- a/wasp-app-runner/src/args.ts
+++ b/wasp-app-runner/src/args.ts
@@ -1,5 +1,6 @@
 import { Argument, program } from "@commander-js/extra-typings";
 import packageJson from "../package.json" with { type: "json" };
+import { defaultPostgresDbImage } from "./db/postgres.js";
 import { Branded } from "./types.js";
 
 export type Mode = "dev" | "build";
@@ -11,7 +12,7 @@ export function parseArgs(): {
   mode: Mode;
   pathToApp: PathToApp;
   waspCliCmd: WaspCliCmd;
-  dbImage: DockerImageName;
+  dbImage?: DockerImageName;
 } {
   const parsedProgram = program
     .name("run-wasp-app")
@@ -26,8 +27,7 @@ export function parseArgs(): {
     .option("--wasp-cli-cmd <command>", "Wasp CLI command to use", "wasp")
     .option(
       "--db-image <image>",
-      "Custom PostgreSQL Docker image to use",
-      "postgres:16",
+      `Custom PostgreSQL Docker image to use (default: "${defaultPostgresDbImage}")`,
     )
     .parse();
 
--- a/wasp-app-runner/src/index.ts
+++ b/wasp-app-runner/src/index.ts
@@ -7,6 +7,8 @@ import {
   type WaspCliCmd,
 } from "./args.js";
 import { startAppInBuildMode } from "./build/index.js";
+import { DbType } from "./db/index.js";
+import { defaultPostgresDbImage } from "./db/postgres.js";
 import { checkDependencies } from "./dependencies.js";
 import { startAppInDevMode } from "./dev/index.js";
 import { createLogger } from "./logging.js";
@@ -38,12 +40,12 @@ async function runWaspApp({
   mode,
   waspCliCmd,
   pathToApp,
-  dbImage,
+  dbImage: dbImageArg,
 }: {
   mode: Mode;
   waspCliCmd: WaspCliCmd;
   pathToApp: PathToApp;
-  dbImage: DockerImageName;
+  dbImage?: DockerImageName;
 }): Promise<void> {
   await checkDependencies();
 
@@ -51,6 +53,14 @@ async function runWaspApp({
     waspCliCmd,
     pathToApp,
   });
+
+  if (dbImageArg && dbType !== DbType.Postgres) {
+    logger.error(
+      `The --db-image option is only valid when using PostgreSQL as the database.`,
+    );
+    process.exit(1);
+  }
+  const dbImage = dbImageArg ?? defaultPostgresDbImage;
 
   if (await isWaspTypescriptConfigProject(pathToApp)) {
     await waspTsSetup({
--- a/wasp-app-runner/src/db/postgres.ts
+++ b/wasp-app-runner/src/db/postgres.ts
@@ -9,6 +9,8 @@ import { Branded } from "../types.js";
 import type { AppName } from "../waspCli.js";
 import type { SetupDbResult } from "./types.js";
 
+export const defaultPostgresDbImage = "postgres:16" as DockerImageName;
+
 type DatabaseConnectionUrl = Branded<string, "DatabaseConnectionUrl">;
 
 const logger = createLogger("postgres");


package.json
```diff
@@ -9,7 +9,7 @@
     "prettier-plugin-organize-imports": "4.1.0",
     "prettier-plugin-pkg": "0.19.1",
     "prettier-plugin-sh": "0.17.4",
-    "prettier-plugin-tailwindcss": "^0.6.13",
```
sodic: Why does the app runner need a tailwind plugin for prettier anyway?

wasp-app-runner/README.md
```diff
@@ -68,6 +69,15 @@ If Postgres is used, the script automatically sets the `DATABASE_URL` env variab
 DATABASE_URL=postgresql://postgres:devpass@localhost:5432/postgres
 ```
 
+#### Custom database image
+
+You can override the Docker image used for Postgres via `--db-image`. If not provided, the default `postgres` is used.
+
+Examples: `--db-image postgres:15`, `--db-image pgvector/pgvector:pg16`, `--db-image postgis/postgis:14-3.2`.
```
sodic, line 76: I'd prefer examples with the full command, if you think it makes sense.

wasp-app-runner/README.md
```diff
@@ -68,6 +69,15 @@ If Postgres is used, the script automatically sets the `DATABASE_URL` env variab
...
+
+> [!NOTE]
+> The same requirements as in `wasp start db` apply for the Postgres image used. See [Wasp docs](../web/docs/data-model/databases.md#custom-database) for more details.
```
sodic, line 79: We'll forget about this link at one point (there's no system keeping it in check). I'd just repeat what we said there.

The link might die/change, but the text isn't likely to become untrue IMO.

wasp-app-runner/src/args.ts
```diff
@@ -22,6 +23,7 @@ export function parseArgs(): {
     )
     .option("--path-to-app <path>", "Path to the Wasp application", ".")
     .option("--wasp-cli-cmd <command>", "Wasp CLI command to use", "wasp")
+    .option("--db-image <image>", "Custom PostgreSQL Docker image to use")
```
FranjoMindek: Instead of doing it the current way, I would make this option have a default value (`"postgres:16"`) instead.

Then we can simplify code (no more handling of `undefined` and picking where to switch to a default value) and we have a global configuration instead of it being kept hidden deep inside of the function call stack. 

This makes sense because `--db-image` is indeed a global configuration (we won't have multiple different values).

FranjoMindek (changes requested):
I think we can make the `db-image` configuration global rather than local. 
Very straightforward and clean RP otherwise.

wasp-app-runner/src/args.ts
```diff
@@ -10,6 +10,7 @@ export function parseArgs(): {
   mode: Mode;
   pathToApp: PathToApp;
   waspCliCmd: WaspCliCmd;
+  dbImage?: string;
```
sodic: It seems this code has the same problem I talked about in my review of https://github.com/wasp-lang/wasp/pull/3182#pullrequestreview-3258864291.

The "optionality" of `dbImage` is stretching itself too deep into the code. The default value should be handled as soon as possible. Most of the code shouldn't even know whether it's working with a default value or a custom value.

Think of it like this:
> If we want to change the default value, the change should be in a single, obvious place. And the place should be the same for all flags/default values. We shouldn't have to dig all the way down to specific functions to change specific values. 

Btw, am I missing something here? Since all other values are handled this way already and only the new one isn't, it's possible I'm not seeing what you're seeing.

package.json
FranjoMindek: ~~It just makes the version fixed, which is how you should install prettier plugins.~~
I didn't even notice it was `tailwindcss`. Most likely copy-paste from `wasp` repo from the time when it was outside of `wasp` repo.

wasp-app-runner/README.md
Martinsos: Sorry to pull in the opposite direction, but I have to admit I don't mind not having the full command, I don't see what it would bring.

wasp-app-runner/README.md
Martinsos: Again me with the opposite opinion hah, sorry about that!
If requirements change, who will remember to update them? It is the same problem as with link. But with link, I can at least likely go and manually figure out the correct place even if it is outdated. While with requirements, I will never know if they are outdated or not. Plus, if we ever will have some way to verify links, well then we will have some verification, we certainly won't have it for duplicated requirements.

I think the question is, are requirements the same because the same actual logic is used in the bacgrkound, or because requirements are duplicated in the wasp app runner, in a way that they are the same as in `wasp start db`? If they are duplicated, I don't mind duplicating them here also. If they instead "transitive", then I think linking is better.

Martinsos (commented):
Others covered main stuff well, I just poped a couple of extra comments. I will remain in the role of weak reviewer here and just comment, I won'd be doing approving.

wasp-app-runner/README.md
sodic: Have you seen what trips up our users tho?

People copy paste the most random stuff. There's no stopping them, so I'd rather give them something that will work when they paste it.

Not important though so do as you please @cprecioso.

wasp-app-runner/README.md
sodic: Yeah alright, it's transitive. Fair point. Let's keep the link.

> Again me with the opposite opinion hah, sorry about that!

I'll just quote you for this one :)
>  No need to apologize, why would I be offended by you not agreeing?

wasp-app-runner/README.md
Martinsos: But full command will more easliy get outdated, and it certainly won't work for them anyway as it would be too specific. This way they can't just copy and paste the full command, they willl have to figure out how to plug this in, so I would say this is a better option then.

wasp-app-runner/README.md
Martinsos: I am trying to accomodate all of you special snowflakes but ok ok I see you want me to go strong I can do that

package.json
cprecioso: Bad merge, sorry!

wasp-app-runner/src/args.ts
cprecioso: I did it on purpose really, because of two reasons:

- IMO semantically, it doesn't make sense to handle the default value in the CLI parsing level. Imagine we add support for a MySQL DB, then the default value of `--db-image` would actually be dynamically determined through the flow of the program, since we wouldn't want to set `postgres` image then, but a `mysql` one.

  You can see this in the current code where the `sqlite` and `postgres` drivers have the same input parameters, so if I'm marking the `dbImage` as required and giving it a value at the top-level, I'm at some point passing `dbImage: postgres` into the sqlite driver. Of course (1) that's a no-op, and (2) nothing a refactor can't fix. But it points to this issue I'm talking about.

- The fact that we are using a default value or not is sometimes relevant. For example, in the Railway PR, we need to set some variables if we're not using the default image. Of course, here's not an issue, but in my mind they're both working at the same level of abstraction.

But yeah this is me overvaluing a "correct" implementation ready to get extended, and undervaluing the straightforward-ness to handle this at the top level.

Wdyt?

Martinsos: > But yeah this is me overvaluing a "correct" implementation ready to get extended, and undervaluing the straightforward-ness to handle this at the top level.

This is not clear to me hm, why do you think you are overvaluing a correct implementation vs simpler implementation? I think @sodic was also pushing for the most correct implementation.

I am not sure if I am getting the whole thing right. I think what you are saying is, we can, in case when CLI arg is not provided, set it to the default value right at the top, or we can recognize it is not set (set it to `Nothing`), and then when we want to do something with it, we can at that moment decide how to handle that and grab that default value from somewhere.

I think @sodic suggested the first approach because it means you have that logic of "what do I do when it is empty, aha I set it to default" at only one place in the code, avoiding duplication and desync in the future.

We could alternatively have one central function which does this ("give me the value of db docker image") and then leave the value of image to be `Nothing`, and then expect other parts of code to call that function. This gives the rest of the code more knowledge, because there is extra state of `Nothing` that they can react on (if relevant), but it does leave some extra space for a mistake, since they might not call that function as intended and do something else, by accident.

So first approach is simpler, and less code, but one "state" less, therefore less "knowledge" in the system.
Second approach is more complex, more code, more space for future mistakes, but one "state" more and more "knowledge" in the system.

I would say, unless it is clear that more knowledge is useful, or that it feels semantically correct (I know this sounds vague, but sometimes it just makes sense, I can't say right now because I am not immersed enough into this code), let's go for simpler option (approach one).

Btw I didn't get the point with mysql and postgres, so my bad if there was something there that would inform this discussion further. I do like the idea of tying up the default docker image with the default db system at the type (or some hm) level, if that is what you are suggesting.
Aaaaah I get what you mean now. If user set `db.system` to `mysql`, then we wouldn't set the value of `--db-image` to postgres, but to mysql docker image. Ok, true, that makes sense to me. But we can just say in the CLI docs that we will set it to default image for their chosen db system. We could even dynamically determine which one is that, as you said :D.

cprecioso: > why do you think you are overvaluing a correct implementation vs simpler implementation? 

oof not calling the other thing "incorrect", sorry! just that IMO with the constraints i had in mind, that was the most semantically accurate.

> or that it feels semantically correct (I know this sounds vague, but sometimes it just makes sense, I can't say right now because I am not immersed enough into this code),

yeah this, it felt correct

Martinsos: Ok, I don't have anything else to say here so I will let the two of you figure it out, as I think @sodic has a better grasp of the code than me.

FranjoMindek: > IMO semantically, it doesn't make sense to handle the default value in the CLI parsing level. Imagine we add support for a MySQL DB, then the default value of --db-image would actually be dynamically determined through the flow of the program, since we wouldn't want to set postgres image then, but a mysql one.

Me personally, I would still do it at the start of the program.
If above happens we will most likely need to parse the `AppSpec` to know what to do with the application (it also includes the `prisma` schema).

So the CLI and the `AppSpec` would be the two configuration inputs into the program.
As is usually with these programs, CLI would override the `AppSpec` provided values.

If so I would declare default values as part of the `AppSpec` parsing process.

That way we:
We have a single source of truth for configuration.
The information is known at the start of the program. If we ever need it at some other place we don't have to refactor the code to move the default value declaration to shared parent.
`wasp-app-runner` core/driver has to knew less about how Wasp applications work.

My thinking is if this operation can be overriden at the top, why not just default value it at the top?
The program must work with values defined at the top anyway.
This way we don't have to care about undefined values in the core/driver itself.

> You can see this in the current code where the sqlite and postgres drivers have the same input parameters, so if I'm marking the dbImage as required and giving it a value at the top-level, I'm at some point passing dbImage: postgres into the sqlite driver. Of course (1) that's a no-op, and (2) nothing a refactor can't fix. But it points to this issue I'm talking about.

We shouldn't enforce the same interface if the needs are not the same.
I would say this was a mistake on our part.

We will have some branching logic no matter the solution, but I'm sure we could abstract some docker behavior when we support multiple databases if we do it this way. 

> The fact that we are using a default value or not is sometimes relevant. For example, in the Railway PR, we need to set some variables if we're not using the default image. Of course, here's not an issue, but in my mind they're both working at the same level of abstraction.

Yes, if we don't mind if it's default or not, let's just combine all the input data at the top?

sodic: > You can see this in the current code where the sqlite and postgres drivers have the same input parameters, so if I'm marking the dbImage as required and giving it a value at the top-level, I'm at some point passing dbImage: postgres into the sqlite driver. 

This is a great observation, but my conclusion is different :)

The core problem is that our current implementation has a "type error." It's not a real type error, but our groupings and parameters are set up in a way that allows invalid states.
Effective TS explains this in _Item 29: Prefer Types That Always Represent Valid States._

Specifically, the parameter combination of `dbType: DbType, dbImage?: string` does not properly model our constraints. We lose information and are forced to think about a state which shouldn't exist, `dbType: SQLite, dbImage: somethingOtherThanUndefined`.

That is the root issue we should address. I think we should address it as high up as possible and propagate this information through the type system.

This problem likely affects the CLI too - we shouldn't really allow users to set `--db-image` if they're using `dev` mode. We should instead tell them something like "You can only use `--db-image` in build mode."  If I'm not mistaken, the current implementation will just ignore the `--db-image` argument when used in `dev` mode.

> Imagine we add support for a MySQL DB, then the default value of --db-image would actually be dynamically determined

@FranjoMindek did a great job explaining this point.

Wasp app runner currently asssumes that "build" means "Use a postgres database" and "dev" means "Use an SQLite database."

If we do add MySQL, we'll have to break this assumption and pass an extra input to the runner - either the full app specification like Franjo suggested, or at least an argument specifying the database. In any case, the app runner will get more info and still be able to decide which default to use at the top-level.

Finally, notice how "dev" doesn't have to mean "Use an SQLite database." Apps in dev mode can also use Postgres and, if that becomes necessary, we will again want to communicate this through argument codependencies and the type system.

For example, let's assume we add MySQL to Wasp. Let's also assume we want to work with real databases in development, so we add a `--db-type` flag.

`run-wasp-app dev` should then allow the `--db-type` to be either `SQLite`, `MySql`, or `PostgreSQL`.
  - If `--db-type` is `SQLite`, then we don't allow a `--db-image` flag.
  - If `--db-type` is `MySQL` or `PostgreSQL`, then we allow a `--db-image` flag and choose an appropriate default if one isn't provided. We could even check the image compatibility with the database if we decide it's useful (but again, near the top level).

These constraints and codependencies between arguments should then again be properly modelled by the type system.

> The fact that we are using a default value or not is sometimes relevant. For example, in the Railway PR, we need to set some variables if we're not using the default image. Of course, here's not an issue, but in my mind they're both working at the same level of abstraction.

I don't think those are related. Wasp app runner needs the image to start a docker container. It's not doing anything fancy and I'd expect it to stay that way.

Finally, @FranjoMindek said:
> We shouldn't enforce the same interface if the needs are not the same.
I would say this was a mistake on our part.

Yes, I agree. We should drop the `SetupDbFn` type. It doesn't provide anything useful and just forces our drivers to have the same inputs. I think the correct way to handle this would be doing something like:
```ts
// This would be parsed somewhere up top
// I'm making up type names and don't have this fleshed out, but you get the idea
type DbConfig = { dbType: SQLite } | { dbType: PostgreSQL, appName: string, pathToApp: string, dbImage: string }

export function setupDb(dbConfig: DbConfig): DbSetupResult {
  // ...
}
```

cprecioso: Yep I see the point. So as far as I can see, due to the needs of the code right now, I'm just pushing the default value to as top as I can get it (argument parser). In the future if we add more DB systems, we can then try and model the types in the way of "Making illegal states invalid".

cprecioso:
Ready for re-review @sodic @FranjoMindek @Martinsos

wasp-app-runner/src/args.ts
sodic: Nice, that's the best thing to do right now, I agree.

sodic: > we can then try and model the types in the way of "Making illegal states invalid".

@cprecioso Do you mind creating an issue for this? :grimacing: 

Also, I still think we should warn the caller that this option can't be used in dev mode instead of silently ignoring it.

wasp-app-runner/README.md
```diff
@@ -68,6 +69,15 @@ If Postgres is used, the script automatically sets the `DATABASE_URL` env variab
 DATABASE_URL=postgresql://postgres:devpass@localhost:5432/postgres
 ```
 
+#### Custom database image
```
sodic, line 72: The docs should mention that this can't be used in dev mode too.

sodic (commented):
Good stuff, I'd just like us to inform the user about the inability to use `--db-image` in dev mode, both in the docs and when using the CLI.

wasp-app-runner/README.md
cprecioso: Where did you get the impression that it can't be used on dev mode? It can.

wasp-app-runner/src/index.ts
```diff
@@ -43,6 +54,14 @@ async function runWaspApp({
     pathToApp,
   });
 
+  if (dbImageArg && dbType !== DbType.Postgres) {
+    logger.error(
+      `The --db-image option is only valid when using PostgreSQL as the database.`,
+    );
+    process.exit(1);
+  }
+  const dbImage = dbImageArg ?? defaultPostgresDbImage;
```
sodic, line 63: It's a shame we had to push this all the way down here but I see no other way :/

cprecioso: yep 🫠

PR #3184, opened 2025-09-22
Add `--db-image` argument to `wasp deploy railway`

cprecioso:
Adds an optional `--db-image` argument to the `wasp deploy railway {launch,setup}` command, allowing the user to specify a custom Docker image for the database.

Their default Postgres service only allows adding extensions through the web interface, and only a few "blessed" ones (e.g. neither `PostGIS` or `pgvector` are available). If you want it any other way, you must deploy the image while setting the env variables yourself. For those, I copied Miho's already-running ask-the-documents app on Railway, and everything seems work correctly.

This is part of #3174. We're adding `ask-the-documents` to the repo, which needs a specific PostgreSQL extension. This way we can more easily test the app.

> [!NOTE]
> 🤖 Generated with [Claude Code](https://claude.ai/code)  
> and 👨🏻 reviewed and edited by a person

PR #3187, opened 2025-09-23
Add `--db-image` argument to `wasp deploy fly`

cprecioso:
- On top of #3186 

Adds an optional `--db-image` argument to the `wasp deploy fly {launch,setup,create-db}` command, allowing the user to specify a custom Docker image for the database.

This is part of #3174. We're adding `ask-the-documents` to the repo, which needs a specific PostgreSQL extension. This way we can more easily test the app.

> [!NOTE]


waspc/ChangeLog.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🔧 Small improvements
+
+- You can now specify which PostgreSQL image to use in `wasp deploy fly` with the `--db-image` argument. ([#3187](https://github.com/wasp-lang/wasp/pull/3187))
```
FranjoMindek, line 9: I think it is a new feature.

web/docs/deployment/deployment-methods/wasp-deploy/fly.md
```diff
@@ -193,6 +193,22 @@ Fly.io offers support for both **locally** built Docker containers and **remotel
 
 If you want to build locally, supply the `--build-locally` option to `wasp deploy fly launch` or `wasp deploy fly deploy`.
 
+## Using a Custom PostgreSQL Image
+
+By default, Wasp uses the standard PostgreSQL image provided by Fly.io when creating a new database for your app. However, if your application requires specific PostgreSQL extensions (e.g., PostGIS), you can specify a custom PostgreSQL image during any of the commands that create a database.
```
FranjoMindek: Only thing, I would try to standardize the text between here and `railway`.
https://github.com/wasp-lang/wasp/pull/3184

There will be changes, but e.g. start of the explanation could be the same:

By default standard version by {provider}.. if want different because of e.g. plugins ... specify yourself...

waspc/packages/deploy/src/providers/fly/index.ts
```diff
@@ -41,6 +41,10 @@ class FlyCommand extends Command {
         "--volume-size <volumeSize>",
         "flyctl postgres create option",
         "1",
+      )
+      .option(
+        "--db-image <dbImage>",
+        "custom postgres image with extensions (must be based on flyio/postgres-flex)",
```
Martinsos: I would make it clear it is Docker image.
Also, why "with extensions"? Does it need to have extensions, why would that be so?

web/docs/deployment/deployment-methods/wasp-deploy/fly.md
```diff
@@ -193,6 +193,22 @@ Fly.io offers support for both **locally** built Docker containers and **remotel
...
+
+```shell
+wasp deploy fly create-db <region> --db-image <custom-postgres-image>
+wasp deploy fly setup <app-name> <region> --db-image <custom-postgres-image>
+wasp deploy fly launch <app-name> <region> --db-image <custom-postgres
```
Martinsos: Sometihng missing here at the end?

web/docs/deployment/deployment-methods/wasp-deploy/fly.md
```diff
@@ -193,6 +193,22 @@ Fly.io offers support for both **locally** built Docker containers and **remotel
 
 If you want to build locally, supply the `--build-locally` option to `wasp deploy fly launch` or `wasp deploy fly deploy`.
 
+## Using a Custom PostgreSQL Image
```
Martinsos:
```suggestion
## Using a Custom PostgreSQL Docker Image
```
I feel like we are not mentioning word Docker enough, we mostly just say image. That can be fine is context of Docker image is strongly established, but I don't think it is here, so I would at leats put it here in the title, to establish that context. ANd maybe I would even sprinkle Docker here and there below.

web/docs/deployment/deployment-methods/wasp-deploy/fly.md
Martinsos: So this is me not knowing enough about how `wasp deploy fly` works, but what is the idea here, I need to keep repeating this argument for every command? How often do I need to repeat it, what does the workflow look like? How cumbersome / impractical is that -> it looks pretty impractical? What if they forget to pass it, what happens? If this is a problem, is there a better solution?

web/docs/deployment/deployment-methods/wasp-deploy/fly.md
Martinsos, line 1: What about discoverability -> I am person building a Wasp app, and I used postgres extensions. Now I want to deploy to Fly. How likely am I to see this new piece of docs about custom docker image on Fly?

I guess I will look for it explicitly because I will know I need a custom Docker image, right? In that case I think we are fine.

But another thought. Wouldn't it be cool if Wasp knew you are using a custom docker image for development already, with extensions, and then when you are doing deployment, it prompted you also to provide a custom docker image? Or maybe even reused the same one you used for development? That might be interesting hm. Although I guess it can't be the same image because Fly has some config it expects from it. ALthough we could use that image locally, why not, based on Fly's image. Hm. So maybe optimal solution would be to have a place in Wasp where one can define databaes image, and then that same image is used by default both during development and also when deploying the database. This is also good because you normally want the development environment to be as similar to production environment as possible.

This certainly shouldn't be happening in this issue, but I thought it is a good moment to think about this.

web/docs/deployment/deployment-methods/wasp-deploy/fly.md
```diff
@@ -193,6 +193,22 @@ Fly.io offers support for both **locally** built Docker containers and **remotel
...
+
+```shell
+wasp deploy fly create-db <region> --db-image <custom-postgres-image>
+wasp deploy fly setup <app-name> <region> --db-image <custom-postgres-image>
+wasp deploy fly launch <app-name> <region> --db-image <custom-postgres
+```
+
+:::warning
+
+Custom PostgreSQL images must be compatible with Fly.io, as the platform requires certain configurations to work properly. Since these requirements are not readily documented, en easy way to ensure compatibility is to base your custom image off the official Fly.io PostgreSQL image: [`flyio/postgres-flex`](https://hub.docker.com/r/flyio/postgres-flex).
```
Martinsos:
```suggestion
Custom PostgreSQL images must be compatible with Fly.io, as the platform requires certain configurations to work properly. Since these requirements are not readily documented, an easy way to ensure compatibility is to base your custom image off the official Fly.io PostgreSQL image: [`flyio/postgres-flex`](https://hub.docker.com/r/flyio/postgres-flex).
```

Martinsos (commented):
Makes sense, but I didn't dig into "is this correctly wired in all the places", I just look at the code changes and commented.
If you want somebody to check the first thing, probably best to have @infomiho take a look, or if you are confident, we can just go with it.

web/docs/deployment/deployment-methods/wasp-deploy/fly.md
cprecioso: You're right, made them more consistent between each other!

web/docs/deployment/deployment-methods/wasp-deploy/fly.md
cprecioso: I put it in the text, but I wanted to have the heading be the problem they're having (I need a custom DB), not the solution they might not know beforehand they need (Docker)

web/docs/deployment/deployment-methods/wasp-deploy/fly.md
cprecioso: I added some clarifications that this only needs to be added the first time you set up the DB

web/docs/deployment/deployment-methods/wasp-deploy/fly.md
cprecioso: I think this is related to the issue you opened about declaring extensions in the wasp file somewhere

cprecioso:
Ready @Martinsos @sodic @FranjoMindek

web/docs/deployment/deployment-methods/wasp-deploy/fly.md
```diff
@@ -193,6 +193,26 @@ Fly.io offers support for both **locally** built Docker containers and **remotel
 
 If you want to build locally, supply the `--build-locally` option to `wasp deploy fly launch` or `wasp deploy fly deploy`.
 
+#### Using a custom PostgreSQL database
+
+By default, Wasp uses the standard PostgreSQL image provided by Fly.io when creating a new database for your app. However, if your application requires specific PostgreSQL extensions (e.g., PostGIS), you can specify a Docker image with a custom PostgreSQL installation, with the `--db-image <docker-image>` flag.
```
Martinsos:
```suggestion
By default, Wasp uses the standard PostgreSQL Docker image provided by Fly.io when creating a new database for your app. However, if you have a need for a custom Docker image, e.g., your application requires specific PostgreSQL extensions (e.g., PostGIS), you can specify a Docker image with a custom PostgreSQL installation, with the `--db-image <docker-image>` flag.
```

web/docs/deployment/deployment-methods/wasp-deploy/fly.md
```diff
@@ -193,6 +193,26 @@ Fly.io offers support for both **locally** built Docker containers and **remotel
...
+
+:::tip
+You only need to specify the Docker image once, when first creating the app.
```
Martinsos: Ha ok cool! With what command is that? `launch`? I forgot a bit. Woudl it make sense tom ention that here? Or can it be more commands, I know `launch` is just an alias for running 3 commands one after another I think?

Is there really no need to mention it again later? What if I am doing sometihng directly with the database? It doesn't matter because Docker image is "fixed" up there on Fly? Or in toml files? I am sking this part for myself.

web/docs/deployment/deployment-methods/wasp-deploy/fly.md
```diff
@@ -193,6 +193,26 @@ Fly.io offers support for both **locally** built Docker containers and **remotel
...
+:::
+
+:::warning
+Custom PostgreSQL images must be compatible with Fly.io, as the platform requires certain configurations to work properly. Since these requirements are not readily documented, an easy way to ensure compatibility is to base your custom image off the official Fly.io PostgreSQL image: [`flyio/postgres-flex`](https://hub.docker.com/r/flyio/postgres-flex).
```
Martinsos:
```suggestion
Custom PostgreSQL images must be compatible with Fly.io, as Fly.io platform requires certain configurations to work properly. Since these requirements are not readily documented, an easy way to ensure compatibility is to base your custom image off the official Fly.io PostgreSQL image: [`flyio/postgres-flex`](https://hub.docker.com/r/flyio/postgres-flex).
```

web/docs/deployment/deployment-methods/wasp-deploy/fly.md
```diff
@@ -193,6 +193,26 @@ Fly.io offers support for both **locally** built Docker containers and **remotel
...
+:::
+
+```shell
+# Add the flag the first time you create your app, with any of these commands:
```
Martinsos: Ok this helps! Maybe move this above next to the tip about having to pass it only once. Maybe combine those together somehow. Because I was missing this info there.

web/docs/deployment/deployment-methods/wasp-deploy/fly.md
cprecioso: Reordered the comments a bit so it's clearer.

> Is there really no need to mention it again later

Nope, we don't regenerate the DB any other time (doing so would delete the user's data)

> Docker image is "fixed" up there on Fly?

Yep this

web/docs/deployment/deployment-methods/wasp-deploy/fly.md
```diff
@@ -193,6 +193,25 @@ Fly.io offers support for both **locally** built Docker containers and **remotel
...
+
+By default, Wasp uses the standard PostgreSQL Docker image provided by Fly.io when creating a new database for your app. However, if you have a need for a custom Docker image, e.g., your application requires specific PostgreSQL extensions (e.g., PostGIS), you can specify a Docker image with a custom PostgreSQL installation, with the `--db-image <docker-image>` flag.
+
+:::tip
+You only need to specify the Docker image once, when first creating the app with any of these commands:
+
+```shell
+wasp deploy fly create-db <region> --db-image <custom-postgres-image>
+wasp deploy fly setup <app-name> <region> --db-image <custom-postgres-image>
+wasp deploy fly launch <app-name> <region> --db-image <custom-postgres-image>
```
infomiho, line 210: These docs feel less useful than the Railway docs here: https://github.com/wasp-lang/wasp/pull/3184/files#diff-f9669fe5fac834d02bc236a50a01480fb863df8d83691bb4c3fa6b059127d6a0R13

In those docs I have a copy-pasteable command for most popular use cases pgvector and PostGIS. 

Can we do the same for Fly as well? I like the tip to base the image on `flyio/postgres-flex` but if that's the _only thing we can advise_ let's also give them some pointers on how to craft an image for pgvector or PostGIS. Maybe link to https://community.fly.io/t/adding-pgvector-to-fly-postgres/12202/21

cprecioso: Updated the guidance, uploaded a gist at https://gist.github.com/cprecioso/e19e883138241c1a446f48d6187aae75

recheck pls

Issue #3147, opened 2025-09-11
`wasp new -t saas MyCoolSaasProject` is slow


Martinsos:
For some reason it is quite slow till it check out open saas repo, takes like 10 seconds or something. I am guessing the git command that pulls in the data is probably pulling in too much? Maybe we don't pull in just the latest commit, depth 0 or what is it, but more? Look into it.
Confirmed on wasp version 0.16.x, and also the latest one.
I haven't tried any other versions, or other starter templates (which we should also check while at it).

infomiho:
Tested it now and it's really slow:
```
wasp new miho-saas -t saas  2.82s user 1.37s system 21% cpu 19.140 total
```

cprecioso:
On further review:
- The download does not do any git operations, it downloads the .tar.gz of a tag. (GitHub automatically creates that for each tag)
- These archives are heavily throttled on GitHub's side to conserve resources:
	- The archive has almost zero compression applied
	- The download speed is very limited
- The problem is that the archive (as of 0.18) has two main folders:
  - `opensaas-sh` **(81.2 MB)**: the opensaas website (with e.g. heavy assets)
  - `template` **(13.6MB)**: the actual template
- `wasp new` only needs the `template` folder but it downloads the whole thing, 6x more data than needed, and at very low speeds.

**We should find a way of downloading _only_ the `template` folder, and preferably in some other way than the auto-generated archives.**

Martinsos:
Uff ok! I guess we can just run `git` command and use it to download only content of `template` dir, from specific tag, depth of 0 or 1 or whateveris correct depth?

cprecioso:
> I guess we can just run git command and use it to download only content of template dir

Yeah we might, but we don't require `git` installed for our users, that'd add a new requirement. I think I can find a workaround.

infomiho:
User reported this as well: https://discord.com/channels/686873244791210014/1420096261423501353/1420096261423501353

cprecioso:
Based on a [quick Discord poll](https://discord.com/channels/686873244791210014/920312576133443634/1419645486935048245), I'm working on the following solution design:

> On pushing or editing the release tag for opensaas, a github action automatically runs. It archives only the correct folder ahead of time and adds it to the release. This will get then get served through GitHub's CDN, much faster.
> This is not a complex release pipeline, just a script with on: tags that contains basically git clone && tar -c && gh release upload. 

[...]

PR #3196, opened 2025-09-29
Use pre-built tarballs for `wasp new -t saas`

cprecioso:
Resolves https://github.com/wasp-lang/wasp/issues/3147

In https://github.com/wasp-lang/open-saas/pull/511 we made a GitHub action that pre-compresses the OpenSaaS template and allows it to be delivered via the GH CDN.

In this PR, we point `wasp new` to use that asset.

> I didn't remove the code for declaring a template that downloads a GitHub repo the old way, but it is now unused. I would prefer to remove it, so we don't have dead code lying around, but somebody might feel it's useful so I'm waiting on your opinion.

---

On my laptop, it's **20x faster** (from ~40s to ~2s)

```
Benchmark 1: local-wasp new -t saas project
  Time (mean ± σ):      2.029 s ±  0.556 s    [User: 1.418 s, System: 0.422 s]
  Range (min … max):    1.384 s …  3.047 s    10 runs
 
Benchmark 2: wasp new -t saas project
  Time (mean ± σ):     39.976 s ±  7.191 s    [User: 5.033 s, System: 2.069 s]
  Range (min … max):   28.133 s … 53.633 s    10 runs
 
Summary
  local-wasp new -t saas project ran
   19.71 ± 6.46 times faster than wasp new -t saas project
```


waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
```diff
@@ -159,19 +164,20 @@ openSaasStarterTemplate =
 styleText :: String -> String
 styleText = id
 
-simpleGhRepoTemplate :: (String, Path' Rel' Dir') -> (String, String) -> StartingInstructionsBuilder -> StarterTemplate
-simpleGhRepoTemplate (repoName, tmplPathInRepo) (tmplDisplayName, tmplDescription) buildStartingInstructions =
-  GhRepoStarterTemplate
```
cprecioso, line 164: I had to remove the old function, or the linter complained

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
```diff
@@ -159,19 +164,20 @@ openSaasStarterTemplate =
...
           GhRepo._repoName = repoName,
           GhRepo._repoReferenceName = waspVersionTemplateGitTag
         }
     )
+    assetName
     ( DirBasedTemplateMetadata
         { _name = tmplDisplayName,
           _description = tmplDescription,
-          _path = tmplPathInRepo,
+          _path = [reldir|.|],
```
cprecioso: Given this function is called _`simple`_`GhReleaseAssetTemplate`, I'm assuming the tarball has the files in the top level. This is easy to adapt if that changes in the future.

waspc/cli/src/Wasp/Cli/Archive.hs
```diff
@@ -29,6 +30,7 @@ fetchArchiveAndCopySubdirToDisk archiveDownloadUrl targetFolder destinationOnDis
 
         downloadFile archiveDownloadUrl archiveDownloadPath
         unpackArchive archiveDownloadPath archiveUnpackPath
+        removeFile archiveDownloadPath
```
cprecioso: Previously we always just copied a subfolder so we didn't have to worry about the original archive, which we was kept in the top level of the extracted code. Now we copy the top level, so we first need to remove the just-extracted archive to clean up.

waspc/ChangeLog.md
```diff
@@ -6,6 +6,10 @@
 
 - You can now specify which PostgreSQL image to use in `wasp start db` with the `--db-image` argument. ([#3182](https://github.com/wasp-lang/wasp/pull/3182))
 
+### 🔧 Small improvements
+
+- Creating a new OpenSaaS project is now much faster (around 20x faster in our testing!). ([#3196](https://github.com/wasp-lang/wasp/pull/3196))
```
cprecioso, line 15: I felt like boasting a bit 😁

waspc/README.md
```diff
@@ -438,7 +438,7 @@ Do the non-bold steps when necessary (decide for each step depending on the chan
...
   - Ensure that all starter templates are working with this new version of Wasp.
-    Update Wasp version in their main.wasp files, and update their code as neccessary. Finally, in their repos (for those templates that are on Github), create new git tag that is the same as the new one in `StarterTemplates.hs` (e.g. `wasp-v0.13.1-template`). Now, once new wasp release is out, it will immediately be able to pull the correct and working version of the starter templates, which is why all this needs to happen before we release new wasp version.
+    Update Wasp version in their main.wasp files, and update their code as neccessary. Finally, in their repos (for those templates that are on Github), create new git tag that is the same as the new one in `StarterTemplates.hs` (e.g. `wasp-v0.13.1-template`), and confirm that the GitHub action correctly ran and uploaded a `template.tar.gz` file. Now, once new wasp release is out, it will immediately be able to pull the correct and working version of the starter templates, which is why all this needs to happen before we release new wasp version.
```
cprecioso, line 441: On merging, I will update the release checklist on notion to add this note too.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/GhReleaseAsset.hs
```diff
@@ -0,0 +1,28 @@
...
+import Control.Monad.IO.Class (liftIO)
+import StrongPath (Abs, Dir, Dir', Path', Rel')
+import Wasp.Cli.Command (Command)
+import Wasp.Cli.Command.CreateNewProject.Common (throwProjectCreationError)
+import Wasp.Cli.Command.CreateNewProject.ProjectDescription (NewProjectAppName, NewProjectName)
+import Wasp.Cli.Command.CreateNewProject.StarterTemplates.Templating (replaceTemplatePlaceholdersInTemplateFiles)
+import Wasp.Cli.GithubRepo (GithubReleaseAssetName, GithubRepoRef, fetchFolderFromGithubReleaseAssetToDisk)
+import Wasp.Project (WaspProjectDir)
+
+createProjectOnDiskFromGhReleaseAssetTemplate ::
```
infomiho: I'd replace asset with archive everywhere e.g. `createProjectOnDiskFromGhReleaseArchive`. We are dealing only with archives here and we decompress them always, so we should be explicit in the name

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/GhReleaseArchive.hs
```diff
@@ -0,0 +1,28 @@
...
+createProjectOnDiskFromGhReleaseAssetTemplate ::
+  Path' Abs (Dir WaspProjectDir) ->
+  NewProjectName ->
+  NewProjectAppName ->
+  GithubRepoRef ->
+  GithubReleaseAssetName ->
+  Path' Rel' Dir' ->
+  Command ()
+createProjectOnDiskFromGhReleaseAssetTemplate absWaspProjectDir projectName appName ghRepoRef assetName templatePathInRepo = do
+  fetchTheTemplateFromGhToDisk >>= either throwProjectCreationError pure
```
infomiho: This threw me off for a seconds because it wasn't clear to me that it will fetch it to the `absWaspProjectDir` dir and then the next line didn't connect for me.

I'd maybe write it like a story `fetch the template to the wasp project dir` so the next line reads more naturally `replace template placeholders in wasp project dir`. This leads me to the name `fetchTemplateFromGhToWaspProjectDir`.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
infomiho: We should encode the assumption that we use `.` always somewhere on the other side e.g. the Open Saas action that packs the template. Ideally we'd have our own "wasp-lang/upload-starter-template" action that has this assumption always enforced.

This is mostly "what would be nice" comment, I think it's okay that we maybe just add a comment to the Open Saas action for now.

waspc/cli/src/Wasp/Cli/Archive.hs
infomiho: Could maybe explain this one more time? Before we didn't remove the archive because we copied just a subfolder? But how did that remove the original archive then?

waspc/cli/src/Wasp/Cli/GithubRepo.hs
```diff
@@ -26,6 +26,27 @@ type GithubRepoName = String
...
+type GithubReleaseAssetName = String
+
+fetchFolderFromGithubReleaseAssetToDisk ::
+  GithubRepoRef ->
+  GithubReleaseAssetName ->
+  Path' (Rel archiveRoot) (Dir folderInArchive) ->
+  Path' Abs (Dir destinationDir) ->
+  IO (Either String ())
+fetchFolderFromGithubReleaseAssetToDisk githubRepoRef assetName folderInArchiveRoot destinationOnDisk = do
+  let downloadUrl = getGithubReleaseAssetDownloadURL githubRepoRef
```
infomiho: I'd prefer if we also had the `assetName` as one of the parameters, so it's clear that it's used to construct the URL, I was confused for a second why didn't we use it.

I know it exists in the scope, but I think it's clear if you are explicit.

waspc/cli/src/Wasp/Cli/GithubRepo.hs
```diff
@@ -26,6 +26,27 @@ type GithubRepoName = String
...
+
+  fetchArchiveAndCopySubdirToDisk downloadUrl folderInArchiveRoot destinationOnDisk
```
infomiho, line 40: `fetchFolderFromGithubReleaseAssetToDisk` uses `fetchArchiveAndCopySubdirToDisk` so I'd say it's okay if we are strict in saying "we fetch only archives from GitHub releases" e.g. `fetchFolderFromGithubReleaseArchiveToDisk`.

I know "release asset" is the term Github uses, but I just feel like archive might be better in our codebase.

waspc/cli/src/Wasp/Cli/Archive.hs
cprecioso: **Old way:**

```
temp-dir/
	- open-saas-repo.tgz

-> untar open-saas-repo.tgz

temp-dir/
	- open-saas-repo.tgz // <- this is not removed
	- template/
		- app/
		- blog/ 
	- opensaas-sh/
	- readme.md
	- etc

-> copy template/ to $destination
// we're only copying template/ so the original .tgz is not copied
```

**New way:**

```
temp-dir/
	- template.tgz

-> untar template.tgz

temp-dir/
	- template.tgz // <- this is not removed
	- app/
	- blog/
	- readme.md

-> copy . to $destination
// if we don't remove `template.tgz`, we will end up copying to the destination
```

infomiho: Can we modify our command to untar to a folder named `template`? That would not require us to delete the `tgz` but I still think it's nice that we cleanup after ourselves.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates.hs
cprecioso: https://github.com/wasp-lang/open-saas/pull/528

waspc/cli/src/Wasp/Cli/Archive.hs
cprecioso: I modified the logic so we extract to a `extracted` folder and get whatever we need from it (in most cases, just `.`). Removed the `removeFile` part because the `withTempDir` helper deletes the whole parent directory anyway.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/GhReleaseArchive.hs
cprecioso: Okey, rewrote it a bit

cprecioso:
@infomiho ready

waspc/README.md
cprecioso: Done

Issue #3085, opened 2025-08-14
The name and location of `waspc/examples/todoApp`


FranjoMindek:
Our kitchen sink application `todoApp` is the only example application under `waspc` folder.
The reasons for this are historic from when it was difficult to run `wasp-cli` outside of the `waspc` folder.
Currently we have no reasons to hold it inside of the `waspc` folder.

We should co-locate it with other `/examples` applications.
This will make it accessible to people who want to see a kitchen sink application.

Secondly, we should rename the `todoApp` to represent what it is, an kitchen sink application.
It should be named something like `kitchen-sink`.

Sidenote: 
This is bigger than it looks like, as it will affect a lot of scripts we have.
We should also consolidate `todoApp` under other generic `example` apps scripts.

cprecioso:
I wouldn't have it under `examples` really, it's not made for people to learn from it.
It's a test app, that's what I would call it

FranjoMindek:
It's the best application to view examples from though? Since it has the complete feature set.

cprecioso:
IMO "<repo>/examples" means for outside people to check examples of how to do stuff in wasp. Those need to be clear and curated.

`todoApp` is not such an example for users to learn, we just use it as a dumping ground of stuff we need to test; and it's purpose is not itself, but a vessel for our headless e2e tests. 

That's why I don't think it should be called an `example`, but a `test` app.

Martinsos:
I suggested @FranjoMindek to open this issue because I understood from our latest efforts that we want Kitchen Sink to be public.

@cprecioso while this was so before, I understood that the intention behind the kitchen sink app changed, and that we decided that we will use it for both at the same time: keep it representable for anybody from the outside, but also make sure it covers all of the Wasp features. Maybe I got that wrong though, and we decided we are nto doing that because those two are too hard to do at the same time? Or we said we are doing it because we avoid duplication that way and we want it to be pretty and well organized anyway?

@infomiho I think you were the main force behind reorganizing our example apps, can you weight in here, what was the final idea behind examples/todoApp (kitchen sink)?

infomiho:
One of the points my [Consolidating example apps RFC](https://www.notion.so/wasp-lang/RFC-Consolidating-example-apps-1e718a74854c808a9d3de8cd8442c4ca?source=copy_link) was:
> 1. Make the `waspc/todoApp` a proper kitchen sink app https://github.com/wasp-lang/wasp/issues/2740
>	...
[...]

PR #3197, opened 2025-09-29
Rename `todoApp` to `kitchen-sink` and move it to `examples/` dir

FranjoMindek:
Fixes https://github.com/wasp-lang/wasp/issues/3085

The goal is to make `todoApp` more public and accessible to users.
For that reason we are renaming it to `kitchen-sink` and moving it to `examples/` (from `waspc/examples/`) dir.

As such `kitchen-sink` becomes one of our example apps and will be treated so.

To go along with that, we are moving `kitchen-sink` tests from `waspc-ci.yaml` workflow to `examples-e2e-tests.yaml`.
Additionally, we are adding it as one of the `deploy-examples.yml` options.

NOTE:
This will currently make it so we don't run `kitchen-sink` tests on each PR.
HOWEVER, in our release post mortem we did declare that we will want to run `examples-e2e-tests.yaml` (along with other test worklows) on each PR. So when we do remove `paths` from their PR triggers, this will go back to old behavior of running `kitchen-sink` tests on each PR.

In this PR I also:
- update the `kitchen-sink`'s `README.md`
- remove the `ensure_app_compiles_and_builds.sh` script
	- we don't lose anything, it was `wasp build` + `tsc` 
- update the `run` script
	- `test:examples` now includes `kitchen-sink`
	- there is still specific `test:kitchen-sink`
	- I've added `npx playwright install --with-deps` to `playwright` based tests to make sure we have all the deps installed
	
Sidenote: It's really hard to check if I covered all `todoApp` renames since our tutorial apps are also called `todoApp`. It's good that we are renaming because confusion.

TODO: Generate `kitchen-sink` production API keys.


Martinsos:
Nice stuff! I will let @infomiho approve since he knows this part better, but I support the change, and PR description sounds reasonable to me, so soft approve from me :D.

infomiho (approved):
I think it makes sense and the changes look OK to me. 

I'd say we are blocked by:
>  This will currently make it so we don't run kitchen-sink tests on each PR.

I'd merge this only AFTER we implement the change that will run example apps tests on each PR, so we are not in a limbo with PRs that are not running tests on the kitchen-sink app.

Let me know if you need help with generating any of the production API keys for the kitchen sink app.

Issue #3202, opened 2025-10-01
Rework the "Before" and "After" tabs in the migration guide


sodic:
"Before" and "After" tabs in our migration guide are tricky to deal with:

- You can’t see both versions at the same time and compare them. Understanding the differences requires flipping between the tabs and remembering stuff.
- It’s easy to mistakenly take the “Before” version of the code and paste that into your app. You sometimes wouldn't even notice you got the same code (e.g., `package.json` with lots of random versions). But even if you do notice, you still have to go back to the documentation to see what's happening

Ideas:
- Replace tabs with side-by-side views.
- Replace tabs with showing the old and the new version one after the other.

nehaaprasad:
Hello! @sodic 
fixed it by implementing a side-by-side BeforeAfter component so users can easily compare both versions.
let me know please everything is correct.

PR #3205, opened 2025-10-01
docs: replace tabs with BeforeAfter component in migration guides

naaa760:
fixes: #3202

### Description

- I've fixed a UX issue in our migration guide documentation where users had trouble comparing "Before" and "After" code examples. The old tabbed interface made it easy to accidentally copy the wrong code version or lose context when switching between tabs.

**What I changed:**
- Created a new `BeforeAfter` component that displays code examples side-by-side with clear visual separation
- Updated all migration guide files to use this new component instead of tabs
- Users can now see both versions simultaneously, making comparisons much easier

**Why this matters:**
- No more tab-flipping to understand differences
- Reduces risk of copying outdated code
- Better visual comparison of changes
- Cleaner, more intuitive documentation experience


infomiho:
Hello, thank you for trying but can't accept LLM generated contributions. The submitted code doesn't compile and after I fixed it, I noticed it doesn't look good either:
image

PR #3235, opened 2025-10-07
Create automation for merging `release` back to `main`

### LINKED ISSUE #3200: Ensure main is always ahead of release
We often make hotfixes or docs changes on the `release` branch but forget to merge them back into `main`.

That's problematic because:
- Branching out of `main` no longer guarantees the most up-to-date version of the code.
- The more `main` and `release` diverge, the more likely it is that we'll have to solve conflicts during the release procedure. And that's the worst time to handle them: we're in a rush and have long forgotten the context.

Let's somehow ensure that all changes we make to the `release` branch are immediately merged back into `main`.

Ideas:
- @cprecioso said: Maybe we can create an automation that automatically creates a PR back to main and tags the person who pushed

End goal: `main` is a fast-forward of `release` at all times (except that short window between the commit on `release` and the merge back to `main`).

cprecioso:
- Resolves #3200

Creates a GH Action to merge `release` back to `main`. Adds the info to the release checklist. Once this PR is accepted, I'll add the same info to the checklist on Notion.

### DIFF
--- a/waspc/README.md
+++ b/waspc/README.md
@@ -397,7 +397,7 @@ How do I know where I want to target my PR, to `release` or `main`?
   - `main` represents near future, and is for changes to the to-be-published stuff.
 
 > [!IMPORTANT]
-> If you target the `release` branch with a PR, once merged, a new PR will automatically be created to update `main` with the changes from `release`. You will be assigned it , and it's your responsibility to merge it.
+> If you merge a PR or push any changes to `release`, a new PR will automatically be created to update `main` with the changes from `release`. You will be assigned it, and it's your responsibility to merge it.
 
 ## Deployment / CI
 
@@ -460,7 +460,7 @@ Do the non-bold steps when necessary (decide for each step depending on the chan
   - This will automatically create a new draft release.
 - 👉 Find new draft release here: https://github.com/wasp-lang/wasp/releases and edit it with your release notes.
 - 👉 Publish the draft release when ready.
-- 👉 You will have been tagged in an automated PR to merge `release` back to `main`. Make sure to merge that PR.
+- 👉 You will have been tagged in an automated PR to merge `release` back to `main`. Make sure to merge that PR. This ensures that `main` is ahead of `release` and we won't have merge conflicts in future releases.
 - Deploy the example apps to Fly.io by running the [deploy-examples workflow](/.github/workflows/deploy-examples.yml) (see "Deployment / CI" section for more details).
 - If there are changes to the docs, [publish the new version](/web#deployment) from the `release` branch.
 - If you published new docs, rerun the Algolia Crawler to update the search index. If you published a new version of the docs, the search won't work until you do this.


.github/workflows/merge-release-to-main.yml
```diff
@@ -0,0 +1,37 @@
...
+    > Please remember to **Rebase and merge** this PR to avoid conflicts in the future.
+
+jobs:
+  create-pr:
+    name: Create Pull Request
+    runs-on: ubuntu-latest
+
+    steps:
+      - run: |
+          gh pr create \
```
cprecioso: This will fail if there is already a PR for release &rarr; main, or if there are no changes to be ported from one to the other. That is to be expected.

.github/workflows/merge-release-to-main.yml
```diff
@@ -0,0 +1,37 @@
...
+env:
+  SOURCE_BRANCH: release
+  TARGET_BRANCH: main
+
+  PR_TITLE: "Merge `release` into `main`"
+  PR_BODY: |
+    Automated PR to merge changes from `release` branch back into `main`.
+
+    > [!IMPORTANT]
+    > Please remember to **Rebase and merge** this PR to avoid conflicts in the future.
```
cprecioso: I am unsure about this, should we recommend to Rebase or to Regular-merge? Can we set a branch protection rule to enforce that?

Martinsos: While I am a big proponent of rebasing and squashing, the case of meging master into release and vice versa is one where I think normal merging makes more sense. Both can move fast, and trying to rebase and similar it just a recipe for trouble -> this is not a good use case for rebase.
So let's go for merging, that is also how we have been doing it so far.

.github/workflows/merge-release-to-main.yml
Martinsos: What happens when it fails? Are we going to see a failed action, or a successful action? It should not fail, the action, right?

Another question: "if there is already a PR for release -> main" -> but not any, right only if if it has this same title and all the other details? Hm I am surprised taht even then it would fail, why wouldn't it create a new one?

waspc/README.md
```diff
@@ -396,6 +396,9 @@ How do I know where I want to target my PR, to `release` or `main`?
   - `release` represents the present, and is for changes to the already published stuff.
   - `main` represents near future, and is for changes to the to-be-published stuff.
 
+> [!IMPORTANT]
+> If you target the `release` branch with a PR, once merged, a new PR will automatically be created to update `main` with the changes from `release`. You will be assigned it , and it's your responsibility to merge it.
```
Martinsos: Only if I target the release branch with PR and merge it? Or on any changes to the release branch?

Martinsos:
```suggestion
> If you target the `release` branch with a PR, once merged, a new PR will automatically be created to update `main` with the changes from `release`. You will be assigned it, and it's your responsibility to merge it.
```

waspc/README.md
```diff
@@ -457,7 +460,7 @@ Do the non-bold steps when necessary (decide for each step depending on the chan
   - This will automatically create a new draft release.
 - 👉 Find new draft release here: https://github.com/wasp-lang/wasp/releases and edit it with your release notes.
 - 👉 Publish the draft release when ready.
-- 👉 Merge `release` back into `main` (`git merge release` while on the `main` branch), if needed.
+- 👉 You will have been tagged in an automated PR to merge `release` back to `main`. Make sure to merge that PR.
```
Martinsos: Maybe even a bit of motivation wouldn't hurt? A sentence that explains this is important to not have master and release drift too much? Maybe not if we want to keep it brief here though.

raw

.github/workflows/merge-release-to-main.yml
cprecioso: > What happens when it fails? Are we going to see a failed action 

My initial design was yes because I didn't want to just ignore all errors and wasn't sure on how to check if there was already a PR. But I checked the `gh` docs a bit more and found how to do it, if in a bit more complex way.

> Another question: "if there is already a PR for release -> main" -> but not any, right only if if it has this same title and all the other details? 

No. Only one PR is allowed for each head to base branch. In GitHub, the PRs are not static, and if the underlying branches are updated, so are the PRs. So if there's already a PR, we don't need to update it, it will just pick up the new changes from the `release` branch. That's also why GitHub doesn't allow you to create a new one for it.

Martinsos: Only one PR per branch -> ok yeah of course, makes sense! Brainfart moment here.

cprecioso:
@Martinsos I addded some stuff to detect if there's already an existing PR. re-check pls?

PR #3236, opened 2025-10-08
Bump version to 0.19

cprecioso:
Bumps version

### DIFF
--- a/mage/Dockerfile
+++ b/mage/Dockerfile
@@ -35,7 +35,7 @@ RUN cd .wasp/build/server && npm run bundle
 # TODO: Use pm2?
 # TODO: Use non-root user (node).
 FROM base AS server-production
-RUN curl -sSL https://get.wasp.sh/installer.sh | sh -s -- -v 0.19.0
+RUN curl -sSL https://get.wasp.sh/installer.sh | sh -s -- -v 0.18.1
 ENV PATH "$PATH:/root/.local/bin"
 ENV NODE_ENV production
 WORKDIR /app
--- a/mage/main.wasp
+++ b/mage/main.wasp
@@ -1,6 +1,6 @@
 app waspAi {
   wasp: {
-    version: "^0.19.0"
+    version: "^0.18.1"
   },
   title: "MAGE - GPT Web App Generator ✨",
   head: [
--- a/waspc/ChangeLog.md
+++ b/waspc/ChangeLog.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 0.19.0 (Unreleased)
+## 0.18.1 (Unreleased)
 
 ### 🎉 New Features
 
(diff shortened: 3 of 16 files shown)


mage/Dockerfile
```diff
@@ -35,7 +35,7 @@ RUN cd .wasp/build/server && npm run bundle
 # TODO: Use pm2?
 # TODO: Use non-root user (node).
 FROM base AS server-production
-RUN curl -sSL https://get.wasp.sh/installer.sh | sh -s -- -v 0.18.0
+RUN curl -sSL https://get.wasp.sh/installer.sh | sh -s -- -v 0.19.0
```
cprecioso: Note that this does not exist atm

Martinsos: Hm yeah. But that is ok, right?

mage/main.wasp
```diff
@@ -1,6 +1,6 @@
 app waspAi {
   wasp: {
-    version: "^0.18.0"
+    version: "^0.19.0"
```
Martinsos: Note that this you don't really need to bump with the rest. Mage is not an example app, nor is it used for testing. We can bump it at its own tempo, when we feel like updating it.

waspc/ChangeLog.md
```diff
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## 0.19.0 (Unreleased)
```
Martinsos: Btw why are we bumping to 0.19.0? Because of the new feature, `--db-image`?

@sodic @infomiho please correct me if I am wrong, but the scheme we have for versioning Wasp right now is that we treat that X in 0.X.Y as major version, and, and Y as minor version, and there is no patch version -> we also bump minor version then. So in this case, we actually woudn't bump it to 0.19, but 0.18.1. At least that is what I remember, but I am not 100% as I haven't been doing the PRs for some time, so asking @sodic and @infomiho to confirm this / comment on this.

Martinsos (changes requested):
All makes sense but I think our versioning scheme might not be the same as what you think, invited others to comment!

mage/main.wasp
cprecioso: In that case I'd actually vote for taking it out of the repo 😁

waspc/ChangeLog.md
infomiho: I believe we are doing the ["major version zero"](https://semver.org/#spec-item-4) paradigm where in `0.X.Y` we keep the zero, and `X` is major and `Y` is minor. 

So I think we should bump to `0.18.1` since we didn't have any breaking changes between `0.18.0` and unreleased changes: https://github.com/wasp-lang/wasp/blob/main/waspc/ChangeLog.md#unreleased

cprecioso:
@Martinsos @infomiho bumped _less_

PR #3237, opened 2025-10-08
Modernize blog file types

FranjoMindek:
Docusaurus allows us to work with two file types for blog posts: MD and MDX.
MDX is an extensions of MD which allows you to use JSX inside of your markdown file.

For now I've converted all MDX files to have `.mdx` extensions instead of `.md`.
Note that we do have mure MD files with `.md` extensions.
But people can find this switching between `.md` and `.mdx` confusing.

I'm not sure if we want to:
1. Use exact extensions name (`.md` or `.mdx`)
2. Always use `.mdx` extensions

Other fixes:
- blogs 
	- added missing truncate indicators
	- prettier fixed duplicated tailwindcss classes in one blog post
	- i converted manual images to ImgWithCaption components in one blog post
	- i made imports consistent across blog posts
- converted blog components to `.tsx` from `.js`
	- converted default to named exports
- made it a requirement to have a truncate indicator (this should allows us to have blog descriptions, but it doesn't work correctly currently, we should fix this)


cprecioso:
@FranjoMindek 

>I'm not sure if we want to:
> - Use exact extensions name (.md or .mdx)
> - Always use .mdx extensions

Yeah, I'm also torn. I'd like to use `.md` wherever available as then we could tell prettier to format those. But I don't want to have to think about if we need `md` or `mdx` when authoring, or have to rename and possibly forget references to this file. ~~I am good with both options, slightly prefer the first option.~~ Realized that changing between md and mdx would also mean stuff like updating the comment syntax. I don't want to have people think about that or have the pain of changing from one to the other, so I change my vote for clearly the second option.

web/blog/2021-02-23-journey-to-ycombinator.mdx
```diff
@@ -14,7 +14,7 @@ Wasp became part of Winter 2021 YCombinator batch!
 
 Here we describe our journey and how we got in after applying for the third time.
 
-<!--truncate-->
+{/* truncate */}
```
cprecioso, line 17: Having a different truncate comment marker depends on md or mdx, right?

web/blog/2023-04-17-How-I-Built-CoverLetterGPT.md
```diff
@@ -8,17 +8,16 @@ tags: [wasp, ai, gpt, fullstack, PERN, stripe, chakra, saas]
 import Link from '@docusaurus/Link';
 import useBaseUrl from '@docusaurus/useBaseUrl';
 
-import InBlogCta from './components/InBlogCta';
-import WaspIntro from './_wasp-intro.md';
-import ImgWithCaption from './components/ImgWithCaption'
```
cprecioso, line 13: I guess this change was automated? I don't think any of these components were actually used. In that case, why only remove the WaspIntro component?

web/blog/components/DiscordLink.js
```diff
@@ -1,9 +1,7 @@
-const DiscordLink = (props) => {
+export function DiscordLink() {
   return (
     <span>
       <a href="https://discord.gg/rzdnErX"> Discord </a>
     </span>
   );
-};
-
-export default DiscordLink;
```
cprecioso, line 9: Personally I don't have a strong preference of default vs. named, so I'd have kept this as a default export so that we don't have to change the imports everywhere. Same with the other components.

web/docusaurus.config.ts
```diff
@@ -213,7 +213,7 @@ const config: Config = {
           blogSidebarTitle: "All our posts",
           postsPerPage: "ALL",
           editUrl: "https://github.com/wasp-lang/wasp/edit/release/web",
-          onUntruncatedBlogPosts: "log",
+          onUntruncatedBlogPosts: "throw",
```
cprecioso, line 216: Good! 🙌

web/blog/2023-04-17-How-I-Built-CoverLetterGPT.md
FranjoMindek: I manually removed `WaspIntro` from most of components since it's not used.
People copy-pasted without knowing. I wanted to stop the cycle.
Forgot to mention that.

Didn't check it for all components, but I should.

web/blog/2021-02-23-journey-to-ycombinator.mdx
FranjoMindek: Yes, and that's why I dislike having two different types.

FranjoMindek:
> Yeah, I'm also torn. I'd like to use .md wherever available as then we could tell prettier to format those.

I thought we disabled prettier for blog posts since formatting causes issues.
e.g. it tries to escape chars from MDX import links, because prettier doesn't recognize it as TypeScript/JavaScript

web/blog/components/DiscordLink.js
FranjoMindek: I generally prefer `named` over `default` exports, I think also Filip had a rant somewhere about this problem.
For me personally, default exports make refactoring a pain. A lot of automated renaming etc. actions don't work properly with them. Especially since the name of exported value can be different in the importer's file.

cprecioso:
> > Yeah, I'm also torn. I'd like to use .md wherever available as then we could tell prettier to format those.
> 
> I thought we disabled prettier for blog posts since formatting causes issues. e.g. it tries to escape chars from MDX import links, because prettier doesn't recognize it as TypeScript/JavaScript

Yep, we did. Prettier supports Markdown but not MDX (v3). We had everything with the `.md` extension (which was basically a lie), so Prettier tried to parse it as regular Markdown, and it ended badly.

But, if we had `.md` files with only Markdown and not MDX, we could enable Prettier for those. Still, moot point since I think we both prefer naming everything as MDX.

---

Btw for future reviewers, a good part of doing this extension rename is that the next major version of Docusaurus will require the correct extensions, and using the wrong one is already considered deprecated.

web/blog/components/DiscordLink.js
cprecioso: Yep, I get it, I just thought the effort vs. return on this would not have been worth it. But you've already done it and this is better, I agree, so perfect!

cprecioso (approved):
LGTM and I don't see any build errors or broken links on a quick click-through around the site, so ok by me

web/blog/2023-04-17-How-I-Built-CoverLetterGPT.md
Martinsos: Yeah people often seem to copy paste without removing redundant stuff, which is not great.

Martinsos:
+1 on argument for going for all .mdx. Not going into deeper review, just commenting in the passing.

web/blog/2024-09-03-OS-builders-interview-with-tim-jones-pgboss.mdx
```diff
@@ -9,6 +9,8 @@ tags: [webdev, postgresql, database, os-maintainers]
 
 We’re launching a new series of posts where we'll sit down with the folks who help us run our projects without expecting anyting in return. Yes, we're talking about open-source maintainers and builders, the people who dedicate their free time to make tech better. This is our way to say "Thank you!" to all of those who help us build and improve [Wasp](https://wasp.sh/), as well as shape the webdev ecosystem.
 
+<!-- truncate -->
```
infomiho: Why not `{/* truncate */}`?

web/blog/2024-12-11-armadajs-2024-a-conference-that-feels-like-home.mdx
```diff
@@ -9,6 +9,8 @@ tags: [conference, community, wasp, webdev]
 
 It took us a few days to wrap our thoughts after attending [ArmadaJS](https://armada-js.com/), a small but exciting conference in Novi Sad, Serbia. People usually believe that bigger conferences are better, but the fact that this one was on a smaller scale meant that it was easer for us to closely connect and talk with other attendees.
 
+<!-- truncate -->
```
infomiho: Another instance of HTML comment

web/blog/2024-09-03-OS-builders-interview-with-tim-jones-pgboss.mdx
FranjoMindek: Fixed

web/blog/2024-12-11-armadajs-2024-a-conference-that-feels-like-home.mdx
FranjoMindek: Fixed

PR #3238, opened 2025-10-08
Workflow consistency and refactor

### LINKED ISSUE #3204: Improve workflow triggers
Our workflows and jobs are a little chaotic. Each workflow has its own triggers and it's easy to forget when to apply what.

For example:
- Prettier doesn't run on `main` or `release` (this may have been fixed, double check)
- Some workflows don't run on version tags (e.g., `test-wasp-deploy`)
- We had an incident when a contributor PR caused a mistake that wasn’t caught on [the PR itself](https://github.com/wasp-lang/wasp/pull/3027) or on main. It was caught on [a different, unrelated PR.](https://github.com/wasp-lang/wasp/actions/runs/17760612682/job/50472336713?pr=3157) It probably wasn’t caught on main because prettier didn't run on `main`. I’m not sure why it wasn’t caught in the contributor PR.
@Martinsos commented that he believes this was already recognized by @cprecioso and fixed by @FranjoMindek. If so, add a comment with appropriate references.

What we should do:
- Investigate and understand everything listed above.
- Bring order into our workflows. @cprecioso suggested a possible solution [here](https://notion.so/wasp-lang/RFC-CI-Architecture-The-Fast-and-the-Furious-24818a74854c8049af39f3e0b2af82a4?pvs=25#24818a74854c806caf73f43b930c0eeb)

### LINKED ISSUE #3206: Ensure rolling updates to example apps
Most release delays come down to finding and fixing problems in the example apps:
- It’s hard to figure out what has been migrated and in what capacity.
- It’s hard to migrate stuff you haven’t worked on.
- Bugs sometimes pop up. They then either have to be fixed by the person doing the release (not well acquainted with the feature) or by the person who introduced them (which is better because they have context).

We should solve this with rolling updates to the example apps. In short:
> If you introduce a change to Wasp. It's also your job to adapt all example apps to use this change.

This should apply to all apps we maintain: examples in the repo (including ask-the-documents once we move it in #3141), Mage, Open Saas (this one's going to be tricky).

End goals:
- The CI continuously tests all example apps and starters (including Open Saas) with the current version on `main`.
- The person doing the release doesn’t need to migrate anything as long as the CI is passing.

### LINKED ISSUE #3219: Move away from `paths` triggers in Github actions
The `paths` trigger is misleading. Example:
  - Example apps (including `todoApp`) had version `0.17.1` in their `main.wasp` files
  - I updated `waspc.cabal` to `0.18.0`
  - The PR didn’t run tests on the example apps because the `paths` trigger didn’t match
  - I merged the PR to main, and [the CI failed](https://github.com/wasp-lang/wasp/actions/runs/17477938511/job/49642272815)

Unless we can truly cover anything that can break a test with a `paths` trigger, we should remove it.

@cprecioso says:
> I try to steer clear of paths, as they are now declaring dependencies between code in a parallel way to the actual code itself. Best thing IMO is to nominally run everything everytime, but caching whatever we can.

cprecioso:
- Resolves #3204 
- Resolves #3219
- Resolves #3206

In this PR we do multiple things at once:

- Renames all the workflows in the same way "{workflow kind}-{feature}-{action}.yaml" so that we can quickly find what we need (workflow kind is one of ci, automation, or release).
- For all the workflows related to CI, we take away all of their `push` and `pull_request` triggers, and `paths` filters.
- We have a main `ci.yaml` workflow that calls all the others. This also simplifies the `release.yaml` workflow since it can just call `ci` before doing its stuff.
- Updates `actions/checkout` and `actions/setup-node` to latest version (no breaking changes)
- Removes the `actions/checkout` workaround that use the PR head commit instead of the PR merge commit
	- This was a "workaround" in quotes because it wasn't working around a bug but the stated, expected GitHub actions behaviour.
	- Original motivation was to avoid the situation were `main` was broken, but the PR was based on a previous non-broken `main`, so you wouldn't readily know where the breakage was coming from.
	- Now that we have the same tests on `main` and PR, this won't happen anymore
	- For the I in CI, we need to be checking how the new code behaves when merged with the current version, so it is more correct to test on the merge commit
	- Plus, GH will give us the cache for the _current_ `main`, not the version of `main` the PR was based off. Testing on the merge commit means it is more probable that we can reuse the warmed cache.
- Adds `WASP_TELEMETRY_DISABLE` to workflows where we had forgot it

Result:
**We run the exact same CI checks both in PRs and `main`, continuously (hey that's the C in CI!)** We do so in the same consistent way, with easy-to-understand entry points.

Future improvement if we feel any pain:
- This PR doesn't add any new caching. We can do that if it becomes too long.
- We also don't mandate any ordering, so every workflow launches in parallel. We can change this if we want.
- We can build the wasp cli only once and use it everywhere we need it, for deduplication

### DIFF
--- a/.github/workflows/ci-waspc-build.yaml
+++ b/.github/workflows/ci-waspc-build.yaml
@@ -112,6 +112,10 @@ jobs:
               brew install --formula "$repo_name/$formula_name"
               brew link --force "$repo_name/$formula_name"
 
+              # Workaround for GHC 9.0.2 and Apple Silicon
+              # https://gitlab.haskell.org/ghc/ghc/-/issues/20592
+              echo C_INCLUDE_PATH="`xcrun --show-sdk-path`/usr/include/ffi" >> $GITHUB_ENV
+
     runs-on: ${{ matrix.env.runner }}
     container: ${{ matrix.env.container }}
     name: build (${{ matrix.env.name }})


FranjoMindek:
https://github.com/wasp-lang/wasp/pull/3239#pullrequestreview-3316178771

Before reviewing, could we tackle this?
Meaning we pass params to `setup-haskell` action only if we want to use non-default values.

Seems like a good match for consistency.

edit: Seems like Filip is doing it here: https://github.com/wasp-lang/wasp/pull/3241

.github/workflows/check-formatting.yaml
```diff
@@ -38,9 +33,7 @@ jobs:
           git config --global core.autocrlf input
           git config --global core.eol lf
 
-      - uses: actions/checkout@v4
-        with:
```
FranjoMindek, line 42: I thought we fixed this one already. Well good thing we do now.

waspc/src/Wasp/Node/Version.hs
```diff
@@ -18,7 +18,7 @@ import Wasp.Util (indent)
 -- | Wasp supports any node version equal or greater to this version.
 -- | We usually keep this one equal to the latest LTS.
 -- NOTE: If you change this value, make sure to also update it on some other places:
---   - /.github/workflows/ci.yaml -> actions/setup-node -> node-version
+--   - /.github/workflows/* -> actions/setup-node -> node-version
```
FranjoMindek:
```suggestion
--   - /.github/workflows/*.yaml -> actions/setup-node -> node-version
```

.github/workflows/release-create.yaml
```diff
@@ -5,36 +5,24 @@ on:
     tags:
       - v*
 
+concurrency:
+  group: "release-create"
+  cancel-in-progress: true
```
sodic, line 10: Why do we need these things again?

waspc/examples/todoApp/README.md
```diff
@@ -86,4 +88,4 @@ npm run test
 
 #### How to run in CI
 
-We set up a GitHub Action to run the tests in CI. See `.github/workflows/waspc-ci.yaml` for details.
+We set up a GitHub Action to run the tests in CI. See `.github/workflows/ci-waspc-test.yaml` for details.
```
sodic, line 91: Doesn't this talk about all tests, not just the CI?

.github/workflows/ci-deploy-test.yaml
```diff
@@ -1,17 +1,16 @@
...
-      - ".github/**"
-    branches:
-      - main
-    tags:
-      - v*
+  workflow_call:
+    secrets:
+      FLY_GITHUB_TESTING_TOKEN:
+        description: "Fly API token for deploying testing apps"
+        required: true
```
sodic, line 8: Just to double check, we need this because the workflow is called from another one, right? There's no way to get the env var without explicitly sending it?

.github/workflows/release-create.yaml
cprecioso: This means that if a release is created while another is running, it will cancel the old one, wait until it's done, and then start the new one

waspc/examples/todoApp/README.md
cprecioso: I was thinking it talks about the tests regarding todoApp, no?

.github/workflows/ci-deploy-test.yaml
cprecioso: Yes to both

.github/workflows/release-create.yaml
infomiho: Two pushes -> two CI runs. If you push faster than the CI finishes, the older CI in progress will be cancelled.

waspc/examples/todoApp/README.md
sodic: Oh, right, didn't see where we were at. IN that case, is `ci-waspc-test` the correct workflow to point it to (might be, just double checking)?

cprecioso: it is! that worflow is too big and does a lot of stuff

cprecioso: somebody braver than me might want to take a stab at it some time 😁

FranjoMindek: TodoApp renaming PR already moves TodoApp stuff into examples workflows.
Best thing is because the abstraction work well, it's just adding the `examples/kitchen-sink` name to the matrix.

PR #3239, opened 2025-10-08
Update GHC to 9.0.2

sodic:
Partially takes care of #1446

### DIFF
--- a/waspc/run
+++ b/waspc/run
@@ -76,7 +76,7 @@ RUN_CMD="cabal --project-dir=${PROJECT_ROOT} run wasp-cli -- ${ARGS[@]}"
 # Utility
 GHCID_CMD="ghcid --command=cabal repl"
 # ghc version below should be the one from `cabal.project` file, `with-compiler` field.
-GHCUP_SET="ghcup set ghc 8.10.7 && ghcup set hls 2.2.0.0"
+GHCUP_SET="ghcup set hls 2.2.0.0"
 DEV_TOOLS_BIN="$PROJECT_ROOT/.bin"
 function install_dev_tool() {
   echo "cabal --project-file=$PROJECT_ROOT/dev-tool.project install $1 --installdir=$DEV_TOOLS_BIN --install-method=copy --overwrite-policy=always"
--- a/waspc/src/Wasp/Analyzer/TypeDefinitions/TH/Decl.hs
+++ b/waspc/src/Wasp/Analyzer/TypeDefinitions/TH/Decl.hs
@@ -9,7 +9,6 @@ where
 import Control.Applicative ((<|>))
 import qualified Data.HashMap.Strict as H
 import Language.Haskell.TH
-import Language.Haskell.TH.Syntax (VarBangType)
 import qualified Wasp.Analyzer.Evaluator.Evaluation as E
 import qualified Wasp.Analyzer.Type as T
 import Wasp.Analyzer.TypeDefinitions (DeclType (..), EnumType (..), IsDeclType (..), IsEnumType (..))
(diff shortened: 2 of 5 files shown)


waspc/src/Wasp/AppSpec/Core/Decl/JSON/TH.hs
```diff
@@ -1,6 +1,5 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE BlockArguments #-}
-{-# LANGUAGE GADTs #-}
```
sodic, line 3: This has nothing to do with the update but I noticed the redundancy and removed it.

waspc/src/Wasp/Analyzer/TypeDefinitions/TH/Decl.hs
```diff
@@ -188,7 +188,7 @@ genIsDeclTypeInstanceDefinition typeName dataConstructorName bodyTypeE evaluateE
           { dtName = $(nameToLowerFirstStringLiteralExpr dataConstructorName),
             dtBodyType = $bodyTypeE,
             dtEvaluate = \typeDefs bindings declName declBodyExpr ->
-              makeDecl @ $(conT typeName) declName <$> declEvaluate typeDefs bindings declBodyExpr
+              makeDecl @($(conT typeName)) declName <$> declEvaluate typeDefs bindings declBodyExpr
```
sodic, line 191: GHC 9.0.2 changed whitespace sensitivity around some special characters, one of them being `@` ([source](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.0#whitespace-sensitive----and-)).

This is the part that concerns us:

image

Because the splice `$(conT typeName)` starts with a `$`, there had to be a space after the type application operator `@`. GHC no longer allows this space (i.e., it interprets it as an operator function call) so we had to remove it and add the parantheses to ensure the parser doesn't interpret `@$` as the same character.

Same goes for the rest of the file.

waspc/waspc.cabal
```diff
@@ -126,7 +126,7 @@ library
     , aeson                 ^>= 1.5.6
     , aeson-pretty          ^>= 0.8
     , text                  ^>= 1.2.4
-    , template-haskell      ^>= 2.16.0
+    , template-haskell      ^>= 2.17.0
```
sodic, line 129: I couldn't update GHC without updating `template-haskell`.

Our `base` is pretty permissive and accounts for all versions that don't make a breaking change. There hasn't been a breaking change between 8.10.7 and 9.0.2 (or even 9.12.x), so there's nothing we have to do there.

waspc/run
```diff
@@ -76,7 +76,7 @@ RUN_CMD="cabal --project-dir=${PROJECT_ROOT} run wasp-cli -- ${ARGS[@]}"
 # Utility
 GHCID_CMD="ghcid --command=cabal repl"
 # ghc version below should be the one from `cabal.project` file, `with-compiler` field.
-GHCUP_SET="ghcup set ghc 8.10.7 && ghcup set hls 2.2.0.0"
+GHCUP_SET="ghcup set hls 2.2.0.0"
```
sodic, line 80: GHC setting is redundant because cabal chooses the compiler version specified in `cabal.project` (which I bumped).

.github/workflows/waspc-build.yaml
```diff
@@ -60,9 +60,6 @@ jobs:
               # Cabal dependencies
               apt-get install -y zlib1g-dev
 
-          # TODO: Add a Linux ARM64 build once we update the GHC version (#1446)
-          # GHC 8.10.7 does not support ARM64 on Linux yet
-
```
sodic, line 65: #3240

FranjoMindek:
> But Github Actions are famous for not being able to de-duplicate stuff, so I don't think we will have any luck there / it is worth bothering with it.

The problem is that we never use default values from setup Haskell action, even if we can.
We always use the same GHC version across all the workflows. We should be able to remove passing GHC version in everything except the setup Haskell version itself.

So we can definitely reduce part of the duplication.

Edit: now mentioned here https://github.com/wasp-lang/wasp/pull/3238#issuecomment-3384482870

waspc/run
cprecioso: What happens when this compiler version is not installed? Does ghcup automatically install too?

sodic: It doesn't - it complains that you're missing a version it needs:

image

cprecioso: So maybe we can modify this command to `ghcup install 9.0.2`?

sodic:
@FranjoMindek said
> We should be able to remove passing GHC version in everything except the setup Haskell version itself.

I did it here: https://github.com/wasp-lang/wasp/pull/3241

@FranjoMindek @cprecioso @Martinsos Check it out and tell me what you think.

PR #3241, opened 2025-10-09
Remove GHC repetitions from Github actions

sodic:
This PR reduces the number of places where we explicitly list our GHC/Cabal version by relying on the defualt versions in the `setup-haskell` action.

### DIFF
--- a/.github/actions/setup-haskell/action.yaml
+++ b/.github/actions/setup-haskell/action.yaml
@@ -15,7 +15,7 @@ inputs:
     description: |
       The version of Cabal to install.
     required: false
-    default: "latest"
+    default: "3.10.2.0"
 
   cabal-project-dir:
     description: |
--- a/.github/workflows/examples-e2e-tests.yaml
+++ b/.github/workflows/examples-e2e-tests.yaml
@@ -32,8 +32,6 @@ jobs:
 
       - uses: ./.github/actions/setup-haskell
         with:
-          ghc-version: "9.0.2"
-          cabal-version: "3.10.2.0"
           cabal-project-dir: waspc
 
       - uses: actions/setup-node@v4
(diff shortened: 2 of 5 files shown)


.github/actions/setup-haskell/action.yaml
```diff
@@ -15,7 +15,7 @@ inputs:
     description: |
       The version of Cabal to install.
     required: false
-    default: "latest"
+    default: "3.10.2.0"
```
sodic, line 18: Better to pin down cabal too.

.github/actions/setup-haskell/action.yaml
sodic, line 1: This is now the only file that lists both versions. We get much less repetition at the cost of some flexibility (e.g., we can no longer dispatch or call `waspc-build` with a different compiler version.

Is this important to us @cprecioso?

.github/workflows/examples-e2e-tests.yaml
```diff
@@ -32,8 +32,6 @@ jobs:
 
       - uses: ./.github/actions/setup-haskell
         with:
-          ghc-version: "9.0.2"
-          cabal-version: "3.10.2.0"
           cabal-project-dir: waspc
```
FranjoMindek, line 35: If we want we could also remove `cabal-project-dir` because `waspc` is the default.

.github/actions/setup-haskell/action.yaml
cprecioso: @sodic not that i know of

.github/actions/setup-haskell/action.yaml
Martinsos: Yeah always better to pin down stuff. Although cabal specifically is mostly ok to always use the newest version. But one day breaking change will come, so why not pin down :).

PR #3245, opened 2025-10-10
Fix wrong names in "merge release to main" automation

cprecioso:
Fixes some names:

- Envs for CLI  `GITHUB_REPOSITORY` -> `GH_REPO` (same with `GH_TOKEN`), those are the correct ones for the `gh` cli
- File was named `.yml` instead of `.yaml` for consistency
- Renamed the `GITHUB_ACTOR` -> `USERNAME` env so it doesn't get confused with the `gh` CLI environment variables


cprecioso:
You're right, I wrote the description a little too fast. All of the changes are intended, updated the description.

PR #3258, opened 2025-10-14
Better Win support: correct resolving of node/npm/npx & prisma

Martinsos:
## Description

I added no tests here because to unit test changes here is too hard because of IO and global vars, and while we could do it if we really wanted to, I don't think it is worth the effort, taking into account the content of the changes. This will all be well covered on the e2e tests and is not that logic heavy.

I did create a separate issue to get wasp cli e2e tests working for Windows (#3404), that will be important.

I based this PR on previous work by @MetaMModern and @nodtem66 + quite some help from @nodtem66 , so big thanks to both, and I added you both as co-authors, as is more than deserved.

## Type of change

- [x] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

  ~- [ ] I added **unit tests** for my change. ~
  ~- [ ] _(if you fixed a bug)_ I added a **regression test** for the bug I fixed. ~
  ~- [ ] _(if you added/updated a feature)_ I added/updated **e2e tests** in `examples/kitchen-sink/e2e-tests`.~
  ~- [ ] _(if you added/updated a feature)_ I updated the **starter templates** in `waspc/data/Cli/templates`, as needed.~
  ~- [ ] _(if you added/updated a feature)_ I updated the **example apps** in `examples/`, as needed.~
    ~- [ ] _(if you updated `examples/tutorials`)_ I updated the tutorial in the docs (and vice versa).~

- 📜 Documentation:

  ~- [ ] _(if you added/updated a feature)_ I **added/updated the documentation** in `web/docs/`.~

  ~- [ ] _(if you did a breaking change)_ I added a step to the current **migration guide** in `web/docs/migration-guides/`.~
  ~- [ ] I **bumped the `version`** in `waspc/waspc.cabal` to reflect the changes I introduced.~


waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
```diff
@@ -178,4 +178,7 @@ absPrismaExecutableFp :: Path' Abs (Dir WaspProjectDir) -> FilePath
 absPrismaExecutableFp waspProjectDir = SP.fromAbsFile prismaExecutableAbs
   where
     prismaExecutableAbs :: Path' Abs File'
-    prismaExecutableAbs = waspProjectDir </> [relfile|./node_modules/.bin/prisma|]
+    prismaExecutableAbs =
+      waspProjectDir
+        </> nodeModulesDirInWaspProjectDir
+        </> SP.castRel (getPathToExecutableInNodeModules "prisma")
```
Martinsos, line 184: We are now being system-agnostic here, when figuring out abs path to prisma binary.

waspc/src/Wasp/Node/Executables.hs
```diff
@@ -0,0 +1,43 @@
...
+import Wasp.Util.System (ExecName, resolveExecNameIO)
+
+-- | Node executable name to be passed to Haskell's "System.Process" functions.
+--
+-- This function being top level form in combo with NOINLINE guarantees that IO action will get
+-- executed only once per lifetime of the Haskell program.
+{-# NOINLINE nodeExec #-}
+nodeExec :: ExecName
+nodeExec =
+  -- NOTE: We are taking whole resolved absolute path here because just using the resolved exec name
```
Martinsos: This I am not super happy with.
So basically just resolving `npm` to `npm.cmd` was not enough, we were still getting errors on Windows. Not the same ones, we moved forward, but different kind, I wish I wrote them down, something about not being able to find some .js files in node_modules.
But maybe it is for the best to have absolute paths to executables we use? Not 100% sure, I think if `npm.cmd` was working I woudl have stuck with it.

Interested in discussion here. Maybe I should try I again and write down the errors, that might help us make a decision.

waspc/src/Wasp/Node/Executables.hs
```diff
@@ -0,0 +1,43 @@
...
+  -- was still flaky on Windows in some situations.
+  fromAbsFile $ snd $ unsafePerformIO $ resolveExecNameIO "node"
```
Martinsos: Instead of doing unsaferPerformIO, we could have actually obtained these in real IO, somweher at the start of our app, and then propagated them through the code, likely via our monad stack.
While that would be the "best" way to do it, I didn't think it makes much sense to bother with such big refactoring at the moment, for no real gain. We can do it if it turns out it is needed.

waspc/src/Wasp/Node/Executables.hs
nodtem66: It is weird that just `npm.cmd` is not enough. But I vote for using absolute path if it works for `node` and `npx`. AFAIK, it works with my Windows 11 without any issues. I don't see any drawbacks of using the full path, except, for the logging, it will be longer than just the executable name.

Martinsos: Yeah I agree. I still don't like the idea that I don't know what is causing this. Plus shorter names feel less fragile in some way, but that is not quite a weak argument in this case, as it could be made the opposite way easily also.

Will put at least one more attempt into this, to gather more info before making a final decision!

Martinsos: Ok, so just to replicate the error I was mentioning, I modified this code to use `fst` instead of `snd`, therefore the command name (e.g. `npm.cmd`) that it successfully resolved, and then I get

image

So it fails when calling the command with `npm.cmd` instead of the full path.

This confuses me because of this:
```hs
-- The resolved path corresponds to the program that would be executed by
-- 'System.Process.createProcess' if exec name was provided as 'System.Process.RawCommand'. Check
-- 'System.Process.findExecutable' for more details since that is what we use internally.
```
This is what `System.Process.findExecutable` promises, and we are using it. But that doesn't sem to be correct (or we are somehow using it somewhere in some way that is not `createProcess`, although what could that be).

Martinsos: Some related issues I found online, but I am not sure if those inform us much beyond that it is indeed problem with exec path resolution:
- https://stackoverflow.com/questions/24721903/npm-npm-cli-js-not-found-when-running-npm
- https://github.com/asdf-vm/asdf-nodejs/pull/426

Martinsos: GPT tells me that what happens is that Windows resolution picks up the local `npm` shim in local `node_modules` but that then fails. Instead of picking up the global one.

But why does `findExecutable` then pick up the global `npm`? Shouldn't they be using the same mechanism and therefore hitting the same problem?

Here is from `findExecutable` docs:
image

Ok, so `findExecutable` uses this `SearchPath` and it depends on it what it will find. And it sounds like that one also will look into relative dir, based on these docs, but obviously not in our case, or somehow skips `.node_modules`? Yeah I mean `.node_modules` shoudln't even be hit by that, it is a neighbouring dir in a way. I am actually wondering why does the `npm.cmd` with `proc` even hit the local `node_modules`.

Some ideas for further testing:
1. Try calling `SearchPath` from powershell with `npm.cmd` and see what it gives us. Do this from inside the `waspc` dir. Confirm it is absolute path to the global npm installation. Or call `findExecutable` from `ghci`, should be the same.
2. Try calling `proc` with `npm.cmd` from `waspc` dir, via `ghci`. Try confirming somehow it resolves to the local `npm` installation in local `.node_modules`, and if so, figure out why.

nodtem66: I think I knew why. Node has both `npm.cmd` and `npm.ps1` to support all shells that users might use.
Both scripts set the variable `NPM_CLI_JS` to the absolute path of `npm-cli.js`.
`npm.cmd` 
```
SET "NPM_CLI_JS=%~dp0\node_modules\npm\bin\npm-cli.js"
```
`npm.ps1`
```
$NPM_CLI_JS="$PSScriptRoot/node_modules/npm/bin/npm-cli.js"
```
The behavior of `$PSScriptRoot` and `%~dp0` varies on the current working directory (`cwd`) and the document of System process suggests that
> If [cwd](https://hackage-content.haskell.org/package/process-1.6.26.1/docs/System-Process.html#v:cwd) is provided, it is implementation-dependent whether relative paths are resolved with respect to [cwd](https://hackage-content.haskell.org/package/process-1.6.26.1/docs/System-Process.html#v:cwd) or the current working directory, so absolute paths should be used to ensure portability. 

That's why we need the absolute path of `npm.cmd` because we set cwd to the `waspc` folder.

Martinsos: @nodtem66  thanks so much for this! Ok yeah that makes sense, so it is at the end cwd that is problematic -> it was my initial hypothesis in the issue, but I was missing understanding why is it problematic.
Ok, so absolute paths are then great because we don't care about what cwd is set to in haskell's calls to processes.
Or, we could make sure that cwd is kept undefined in those calls. But that sounds like a recipe for future problems, better to stick to absolute paths and we are good.

nodtem66: Without setting `cwd`, the variable `%~dp0` will be the current directory that the command `wasp` runs.
So, using absolute path is the proper way to do. We will have more flexible option to set or unset `cwd`.  Plus, we could use the absolute paths of all executables for logging, so that users can monitor the fake `npm.cmd` or the invalid path of `npm.cmd`.

Martinsos: Ok makes sense! I polished the code a bit, removed duplication of comments by moving that knowledge to the resolveExecNameIO, and also made it return only absolute path, not execName + absolutepath, as we don't want to make the mistake of using the first one.

Martinsos:
@nodtem66 in case you want to take a final look, feel free to, although I didn't change it much: however I finally made it ready for review (out of draft)! I did separate out the terminal unstyling into another PR to keep this one clean. I think this is now in good shape. Thanks again for all the help, and I added you as a co-author.

Btw apologies for my intermittent activity on the PRs: these days I am mostly taken away by tasks that are not programming but stuff like management, research and similar, that benefit the team the most, but I do make sure to dedicate at least one day per sprint on development, and that is when I usually grab the time for these PRs.

nodtem66 (approved):
Nice, all of these look great. Thank you for the co-authorship. I'm happy to help out and pick up new things along the way.

Don't worry about that. I get that someone needed to do that job. I usually use my free time for this, and I really enjoy it. The best part was definitely what I earned from the discussion. I hope it is useful for the project more or less.

Martinsos:
> Nice, all of these look great. Thank you for the co-authorship. I'm happy to help out and pick up new things along the way.
> 
> Don't worry about that. I get that someone needed to do that job. I usually use my free time for this, and I really enjoy it. The best part was definitely what I earned from the discussion. I hope it is useful for the project more or less.

It is super useful! I probably wouldn't have the energy to drive through some of these obstacles on my own with the limited time I had available, and I really appreciate the quality of your responses and research/work you put behind each one. Hope to see you around more.

waspc/src/Wasp/Node/NodeModules.hs
```diff
@@ -0,0 +1,25 @@
...
+-- This function returns a path to such an executable with a given name, taking into account
+-- details like current operating system.
+--
+-- Example: @getPathToExecutableInNodeModules "npm"@ -> @".bin/npm.cmd"@
+getPathToExecutableInNodeModules :: String -> Path' (Rel NodeModulesDir) (File f)
+getPathToExecutableInNodeModules execName =
+  [reldir|.bin|] </> fromJust (parseRelFile systemSpecificExecFilename)
+  where
+    systemSpecificExecFilename
+      | isSystemWindows = execName <> ".cmd"
```
cprecioso, line 24: Can we have a comment with a link to somewhere that says that npm adds `.cmd` on Windows?
I found this, don't know if there's a better one: https://docs.npmjs.com/cli/v11/configuring-npm/package-json#bin

waspc/src/Wasp/Node/NodeModules.hs
cprecioso, line 1: I see this logic is only used for calling `prisma`

idea: Could it be possible to call `npx prisma`, instead of duplicating npm's executable resolution algorithm ourselves? Then we can remove this file.

waspc/src/Wasp/Util/System.hs
```diff
@@ -0,0 +1,62 @@
...
+  firstJustM findExecutable execNamesToLookForByPriorityDesc >>= \case
+    Just execPath -> parseAbsFile execPath
+    Nothing ->
+      (throwIO . userError . unlines)
+        [ "Could not find '" <> execName <> "' executable on your system.",
+          "Please ensure " <> execName <> " is installed and available in your PATH."
+        ]
+  where
+    execNamesToLookForByPriorityDesc
+      | isSystemWindows = (execName <>) <$> ["", ".cmd", ".exe", ".ps1"]
```
cprecioso, line 55: We should get this list from the `PATHEXT` environment variable since it is configurable.

waspc/ChangeLog.md
```diff
@@ -26,6 +26,7 @@ Remember to check out the [migration guide](https://wasp.sh/docs/migration-guide
 
 ### 🐞 Bug fixes
 
+- Fixed node/npm/npx/prisma command resolution on Windows, which was one of the main stoppers of supporting Wasp on native Windows. ([#3258](https://github.com/wasp-lang/wasp/pull/3258)).
```
cprecioso, line 29: This should go on "unreleased". Also I'd remove or at least rephrase the last part. Now it is not clear if this change means Wasp on Windows finally works, or if it doesn't yet.

waspc/ChangeLog.md
```diff
@@ -47,6 +48,10 @@ Remember to check out the [migration guide](https://wasp.sh/docs/migration-guide
 
 - Wasp's `kitchen-sink` application has been moved to the public examples (`examples/`) directory in our repo. ([#3085](https://github.com/wasp-lang/wasp/pull/3197))
 
+### 🌟 Contributors
+
+- Big thanks to @MetaMModern and @nodtem66 for crucial work in releasing the fix for node commands resolution on Windows ([#3258](https://github.com/wasp-lang/wasp/pull/3258)) and other related issues.
```
cprecioso, line 53: In the past we've just noted the contributors by tagging them in the changelog item. Leaving this to your choice since I understand this was a lot of work.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
cprecioso: See my comment on the `NodeModules.hs` file

cprecioso (commented):
Shallow review, but wanted to comment a couple of things on the design:

waspc/src/Wasp/Node/Executables.hs
```diff
@@ -0,0 +1,22 @@
...
+nodeExec :: ExecName
+nodeExec = fromAbsFile $ unsafePerformIO $ resolveExecNameIO "node"
+
+{-# NOINLINE npmExec #-} -- To ensure single execution, for unsafePerformIO.
+npmExec :: ExecName
+npmExec = fromAbsFile $ unsafePerformIO $ resolveExecNameIO "npm"
+
+{-# NOINLINE npxExec #-} -- To ensure single execution, for unsafePerformIO.
+npxExec :: ExecName
+npxExec = fromAbsFile $ unsafePerformIO $ resolveExecNameIO "npx"
```
Martinsos, line 22:
```suggestion

-- We are using NOINLINE in the code below to ensure that
-- unsafePerformIO executes only once.

{-# NOINLINE nodeExec #-}
nodeExec :: ExecName
nodeExec = fromAbsFile $ unsafePerformIO $ resolveExecNameIO "node"

{-# NOINLINE npmExec #-}
npmExec :: ExecName
npmExec = fromAbsFile $ unsafePerformIO $ resolveExecNameIO "npm"

{-# NOINLINE npxExec #-}
npxExec :: ExecName
npxExec = fromAbsFile $ unsafePerformIO $ resolveExecNameIO "npx"
```

waspc/src/Wasp/Node/NodeModules.hs
Martinsos: Do we need to? If it didn't why would we have this logic here? What I am saying is, this code says that if on Windows, we add .cmd to the exec name. Stating next to it that we do so because npm on windows add .cmd is just repeating what the code said.

waspc/src/Wasp/Node/NodeModules.hs
Martinsos: Hm so much time passed so I can't remember why we didn't just use `npx`, was there a good reason for it or just ommision. But I agree that if we can use `npx`, we should, instead of replicating the resolution ourselves. Will look into this!

waspc/src/Wasp/Util/System.hs
```diff
@@ -0,0 +1,62 @@
...
+import Control.Exception (throwIO)
+import Control.Monad.Extra (firstJustM)
+import StrongPath (Abs, File', Path', parseAbsFile)
+import System.Directory (findExecutable)
+import qualified System.Info
+
+-- | Executable name as expected by Haskell's "System.Process" and its 'System.Process.RawCommand',
+-- therefore suited for passing to their functions for creating/executing processes.
+-- It can be just "node", or "node.exe", or relative or full path, ... .
+type ExecName = FilePath
```
Martinsos, line 18: This one is a bit weird. I remember I had reasons for it but still.

Name sounds general enough for person not to know it is crafted for System.Process and System.Process.RawCommand. And also to think it is actual exec and not maybe path or something.

The main offender is really System.Process and System.Process.RawCommand I guess, but still it is our problem now.

Maybe one solution is to just say it is path? ExecPath? And yeah, maybe it is justu file name, maybe it is relative, maybe it is absolute, but well all those are paths. But I guess System.Process uses "executable name" nomenclature? Check if that is so and how important is it to stick to it.

Then, another question: do we want to make this consumable only by System.Process and System.Process.RawCommand, if we indeed assume whatever is inside is specifically tailored for them? Do we want to express that on the type level?
I don't think I would go as far as creating a proxy API for System.Process that consumes only our data type, but I do thikn we could use `data` instead of `type` that can be unboxed only with `forSystemProcess` function (that is trivial and is here just for name). So desctructor would be hidden and you would have to explicitly unpack it for specific usage, which sounds right.

waspc/src/Wasp/Util/System.hs
```diff
@@ -0,0 +1,62 @@
...
+--
+-- Motivation for this function was mainly driven by how exec names are resolved when executing a
+-- process on Windows.
+-- On Linux/MacOS situation is simple, the system will normally do the name resolution for us, so
+-- e.g. if we pass "npm" to 'System.Process.proc', that will work out of the box.
+-- But on Windows, that will normally fail, since there is no "npm" really but instead "npm.cmd" or
+-- "npm.exe". In that case, we want to figure out what exactly is the right exec name to use.
+-- Note that we don't have to bother with this when using 'System.Process.shell' instead of
+-- 'System.Process.proc', becuase then the shell will do system resolution for us,
+-- but at the price of abandoning any argument escaping.
```
Martinsos, line 39:
```suggestion
-- but at the price of abandoning any argument escaping, which is often not what we want.
```

waspc/src/Wasp/Util/System.hs
Martinsos: Interesting name potentially: SystemExecName. Or SystemProcName. Or SystemProcPath.

waspc/src/Wasp/Node/Executables.hs
```diff
@@ -0,0 +1,22 @@
...
+    npxExec,
+  )
+where
+
+import GHC.IO (unsafePerformIO)
+import StrongPath (fromAbsFile)
+import Wasp.Util.System (ExecName, resolveExecNameIO)
+
+{-# NOINLINE nodeExec #-} -- To ensure single execution, for unsafePerformIO.
+nodeExec :: ExecName
```
Martinsos, line 13: These names, nodeExec, npmExec, ... -> I actually don't like reading them in places of usage because they look as if we they stand for some kind "npm exec" command or "execute npm". So I would consider better name here. Can be longer if needed. `npmExe`? `npmExecutable`? `npmProcName`? This is connected to how I decide to potentially rename ExecName so first let's do that, then do this.

waspc/packages/wasp-config/__tests__/cli.unit.test.ts
```diff
@@ -0,0 +1,86 @@
...
+  test("should throw an error if the entity names JSON is not an array", () => {
+    expectParseProcessArgsToError([
+      "main.wasp.js",
+      "output.json",
+      JSON.stringify({ entity1: "entity1" }),
+    ]);
+  });
+
+  function expectParseProcessArgsToSucceed(args: string[], nodeExec: string = "node") {
+    const result = parseProcessArgsOrThrow([nodeExec, "run.js", ...args]);
```
Martinsos, line 71: Hm I should remember what is this, why are we modifying / writing test in cli.unit.test.ts file in wasp-config.

Martinsos (commented):
Quick round of comments to myself, coming back to the PR after long time and doing initial review!

PR #3261, opened 2025-10-15
Fix dev-tools.project versions

FranjoMindek:
Updating `ormolu` version made it so that our project is no longer formatted correctly:
https://github.com/wasp-lang/wasp/actions/runs/18526820452/job/52799513371?pr=3261

So I reverted the `ormolu` version update for now.
I suppose we will do that once we finish updating GHC fully.

### DIFF
--- a/waspc/dev-tool.project
+++ b/waspc/dev-tool.project
@@ -6,7 +6,7 @@ packages: .
 jobs: $ncpus
 
 constraints:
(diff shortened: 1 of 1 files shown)


Martinsos:
Hm but you haven't touch ormolu version in this PR?

FranjoMindek:
I did, but I reverted it.

image

Issue #2953, opened 2025-07-14
Automatically test Wasp AI


sodic:
Wasp AI is usually the last thing we test. We do it right before the releasae when everything is ready. Most of the time, it goes smoothly. But when it doesn't, it can get pretty painful.

One painful example was the release of Wasp 0.17.0. We discovered [this issue](https://github.com/wasp-lang/wasp/issues/2951) during RC testing and it turned out to be a huge can of worms. The problem was caused by [something we implemented](https://github.com/wasp-lang/wasp/pull/2625) more than two months prior. 

The conclusion was that we should test `wasp ai` in our CI, and that we should run it as often as possible.

FranjoMindek:
As noted in https://github.com/wasp-lang/wasp/pull/3268#issue-3526412846,
the `mage/Dockerfile` file features a Wasp install scripts which is tied to a specific version.
The problem is the version specified is the current `main` branch version, while the installer script relies only on released versions, e.g.:
```sh
# main has version 0.18.2
# release is version 0.18.1

RUN curl -sSL https://get.wasp.sh/installer.sh | sh -s -- -v 0.18.2
```

In short, `mage` Dockerfile works on `release`, but it is broken on `main`.

We will need to solve this to make `mage` work as part of the CI tests.

PR #3268, opened 2025-10-17
Bump project version on `main` post hotfix on `release`

FranjoMindek:
Project version bump to `0.18.2` since we released a hotfix on `release`.
This is to ensure `waspc` is once again ahead of `release` on `main`.

I have 1 possible problem I want to talk about:

Locally, `mage/Dockerfile` will fail with the current setup because we also update the version for the installer script:
```dockerfile
RUN curl -sSL https://get.wasp.sh/installer.sh | sh -s -- -v 0.18.2
```
This is because `0.18.2` version isn't released yet.

`waspc/README.md` does say that we need to bump the `mage` version. So this is an intended part of our process. 

While this does not conflict with our current CI setup since we don't test `mage` on it, it may in future.


Martinsos:
Can you simplify this for me:
- why are we doing this change/PR?
- I didn't get the thing with Mage. What is the problem, and what are the options? Doesn't sound to me like we want Mage to fail hah. But you sound as if that is still the right way to go? Is there no better way? Give me some context / options / explanation.

FranjoMindek:
I've updated the PR description.
I suppose we don't want to change anything `mage` related in this PR (especially since it's the process we describe in `waspc/README.md`), but the explained problem is something we will need to solve when we start testing mage in CI.

Related?: https://github.com/wasp-lang/wasp/issues/2953

Martinsos (approved):
Ok, all makes sense to me! I didn't check the code super thoroughly but it makes sense. As for mage -> please create another issue for this, or add it to existing issue, mabye that one you linked.

PR #3283, opened 2025-10-29
Update GHC to recommended version (9.6.7)

sodic:


sodic:
Ah if it only was this simple, this was supposed to be a draft. New PR: #3302 (builds on the previous too)

PR #3286, opened 2025-11-01
Update GHC to 9.2.8

sodic:


.github/workflows/ci-waspc-build.yaml
```diff
@@ -112,10 +112,6 @@ jobs:
               brew install --formula "$repo_name/$formula_name"
               brew link --force "$repo_name/$formula_name"
 
-              # Workaround for GHC 9.0.2 and Apple Silicon
-              # https://gitlab.haskell.org/ghc/ghc/-/issues/20592
-              echo C_INCLUDE_PATH="`xcrun --show-sdk-path`/usr/include/ffi" >> $GITHUB_ENV
```
sodic, line 114: No longer necessary on GHC 9.2.8.

waspc/cli/tests/DbMigrateTest.hs
```diff
@@ -1,7 +1,7 @@
 module DbMigrateTest where
 
 import Data.Either (isLeft)
-import Test.Tasty.Hspec
```
sodic, line 4: Newer versions of Tasty no longer reexport Hspec (since 1.1.7, [changelog](https://hackage.haskell.org/package/tasty-hspec-1.2.0.4/changelog)).

waspc/e2e-tests/snapshots/kitchen-sink-golden/wasp-app/.wasp/out/installedNpmDepsLog.json
sodic, line 1: Aeson changed how they order object keys when printing JSON.

waspc/src/Wasp/Analyzer/Parser/Lexer/Internal.hs
```diff
@@ -57,9 +57,9 @@ getInput = lstateInput <$> get
 updateInput :: Int -> Lexer ()
 updateInput !consumed = do
   (LexInput _ _ remaining) <- getInput
-  let newInput =
-        let (prevChar : remaining') = drop (consumed - 1) remaining
```
sodic, line 61: Newer versions of GHC throw errors on pattern matches like this one ([already encoutered](https://github.com/wasp-lang/strong-path/pull/54#discussion_r2472330271) in Strong Path).

I used the same pattern/message we use in the rest of the project.

waspc/src/Wasp/AppSpec/Job.hs
```diff
@@ -2,6 +2,8 @@
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DuplicateRecordFields #-}
+-- TODO: Remove once we start using record dot syntax
+{-# OPTIONS_GHC -Wno-ambiguous-fields #-}
```
sodic, line 6: image

I'm ignoring the error for now. When it becomes necessary, we can start using record dot syntax.

I left a todo, should I make an issue for switching over to record dot syntax? How do we want to move, at once or gradually? For example, I could start using it in this module immediately, and that would allow me not to silence the warning.

waspc/src/Wasp/Generator/Crud.hs
```diff
@@ -68,4 +68,4 @@ crudDeclarationToOperationsList crud =
 makeCrudOperationKeyAndJsonPair :: AS.Crud.CrudOperation -> Aeson.Value -> Aeson.Types.Pair
 makeCrudOperationKeyAndJsonPair operation json = key .= json
   where
-    key = T.pack . show $ operation
+    key = Key.fromString . show $ operation
```
sodic, line 71: This was a breaking change in Aeson.

waspc/src/Wasp/Util.hs
```diff
@@ -48,11 +48,12 @@ import Control.Applicative (liftA2)
 import Control.Monad (unless, when)
 import qualified Crypto.Hash.SHA256 as SHA256
 import qualified Data.Aeson as Aeson
+import qualified Data.Aeson.Key as Key
+import qualified Data.Aeson.KeyMap as KM
```
sodic, line 52: Another breaking change in Aeson 2.0.0.0. Check the [changelog](https://hackage.haskell.org/package/aeson-2.2.3.0/changelog) for more details.

waspc/src/Wasp/Util.hs
```diff
@@ -107,8 +107,8 @@ headSafe xs = Just (head xs)
 second3 :: (b -> d) -> (a, b, c) -> (a, d, c)
 second3 f (x, y, z) = (x, f y, z)
 
-jsonSet :: Text.Text -> Aeson.Value -> Aeson.Value -> Aeson.Value
-jsonSet key value (Aeson.Object o) = Aeson.Object $ M.insert key value o
+jsonSet :: Key.Key -> Aeson.Value -> Aeson.Value -> Aeson.Value
+jsonSet key value (Aeson.Object o) = Aeson.Object $ KM.insert key value o
```
sodic, line 111: Since Aeson keys can no longer be `Text`, it doesn't make sense for this function to accept `Text` either.

waspc/src/Wasp/Util.hs
```diff
@@ -138,10 +138,11 @@ indent numSpaces = intercalate "\n" . map (toEmptyStringIfAllWhiteSpace . (inden
...
 concatShortPrefixAndText :: String -> String -> String
-concatShortPrefixAndText prefix "" = prefix
-concatShortPrefixAndText prefix text =
-  let (l : ls) = lines text
-   in prefix ++ l ++ if null ls then "" else "\n" ++ indent (length prefix) (intercalate "\n" ls)
+concatShortPrefixAndText prefix text = case lines text of
+  [] -> prefix
+  (l : ls)
+    | null ls -> prefix ++ l
+    | otherwise -> prefix ++ l ++ "\n" ++ indent (length prefix) (intercalate "\n" ls)
```
sodic: This code alo had an "incomplete patterns" warning.

waspc/src/Wasp/Version.hs
```diff
@@ -5,6 +5,6 @@ import qualified Paths_waspc
 import qualified Wasp.SemanticVersion as SV
 
 waspVersion :: SV.Version
-waspVersion = SV.Version (toEnum major) (toEnum minor) (toEnum patch)
-  where
-    DV.Version [major, minor, patch] _ = Paths_waspc.version
+waspVersion = case Paths_waspc.version of
+  DV.Version [major, minor, patch] _ -> SV.Version (toEnum major) (toEnum minor) (toEnum patch)
+  _ -> error "This should never happen. Wasp binary version have exactly three digits, but it doesn't."
```
sodic: Another non-exhaustive pattern warning.

Btw, this got me wondering: What happens if we mistakenly forget to specify the minor or patch version number in `waspc.cabal`. When would it fail?

And the answer is: right here.

image

Without this new message, the error is a generic "non-exaustive pattern in..."

waspc/tests/Analyzer/EvaluatorTest.hs
```diff
@@ -226,7 +227,9 @@ spec_Evaluator = do
                 "  booleanValue: {=json false json=},",
                 "}"
               ]
-        let Right [("Test", allJson)] = takeDecls <$> eval typeDefs source
+        let allJson = case takeDecls <$> eval typeDefs source of
+              Right [(_, aj)] -> aj
+              left -> error $ "Expected Right but got Left:" ++ show left
```
sodic: I got a couple of these "non-exhaustive pattern" errors in tests too. I wasn't sure about the best way to handle them, so I took 3 different approaches in three different places:
- Here
- In WaspGeneratorTest.hs
- In TypeCheckerTest.hs

Tell me which one you prefer @Martinsos

waspc/waspc.cabal
sodic, line 1: All version bumps are driven either by:
- A `base` version mismatch.
- An integration error with sister libraries. For example, Tasty was bumped to a higher version than `base` would require because lower versions didn't when combined with available versions of `tasty-discover`.

I also had to add `hspec` because `tasty-hspec` no longer re-exports their symbols.

Here's a claude generated summary of the changes:

1. **aeson**: `^>= 1.5.6` → `^>= 2.0`
   - Reason: aeson 1.5.6 requires ghc-prim < 0.8, incompatible with GHC 9.2.8 (ghc-prim 0.8.0)
   - Breaking change: Object type changed from HashMap to KeyMap

2. **mustache**: `^>= 2.3.2` → `^>= 2.4` (2 occurrences)
   - Reason: mustache 2.3.2 requires aeson < 2
   - Locations: main library and waspls library

3. **cryptonite**: `^>= 0.29` → `^>= 0.30`
   - Reason: cryptonite 0.29 requires base < 4.16, GHC 9.2.8 has base 4.16.4.0

4. **bytestring**: `^>= 0.10.12` → `^>= 0.11` (2 occurrences)
   - Reason: bytestring 0.10.x incompatible with base 4.16
   - GHC 9.2.8 ships with bytestring 0.11.4.0
   - Locations: main library and cli-lib

5. **tasty-hspec**: `>= 1.1 && < 1.1.7` → `>= 1.2` (4 occurrences)
   - Reason: tasty-hspec 1.1.6 doesn't support base 4.16
   - Locations: waspc-tests, waspls-tests, wasp-cli-tests, waspc-e2e-tests
   - Breaking change: Spec, it, shouldBe etc. no longer exported

6. **hashable**: `^>= 1.3.5.0` → `^>= 1.4`
   - Reason: stm-containers 1.2.1+ requires hashable >= 1.4

7. **stm-containers**: `^>= 1.2` → `^>= 1.2.1`
   - Reason: stm-containers 1.2 pulls in stm-hamt 1.2.0.10 which has compilation errors
   - stm-hamt 1.2.1.1 (pulled by stm-containers 1.2.1) fixes the issue

8. **hspec**: Added `>= 2.11` to all test suites
   - Reason: tasty-hspec 1.2+ requires hspec as separate dependency
   - Added to: waspc-tests, waspls-tests, wasp-cli-tests, waspc-e2e-tests

sodic (commented):
@Martinsos Please check this out and tell me what you think.

waspc/src/Wasp/AppSpec/Job.hs
sodic: Ok, so it turns out they deprecated it in the very next major, so I started using record dots: https://github.com/wasp-lang/wasp/pull/3287#discussion_r2489620174

waspc/cabal.project
```diff
@@ -16,8 +16,10 @@ package waspc
 -- building multiple packages at once, for example external dependencies.
 jobs: $ncpus
 
+tests: True
```
Martinsos, line 24: What is this?

waspc/src/Wasp/Util.hs
Martinsos: Makes sense! Rewrite suggestion (extracted `prefix`, replaced guards with cases):
```suggestion
concatShortPrefixAndText prefix text =
  prefix <> case lines text of
    [] -> ""
    [l] -> l
    (l:ls) -> l <> "\n" <> indent (length prefix) (intercalate "\n" ls)
```

Martinsos: This communicates what this is about: we add something to prefix, in all cases, and I find it easier to read because conditions are at single level, so one tree with three leaves, not one tree with two branches where one of them is leaf and another is tree wiht two leaves. We are taking about decision tree here.

But resolve this as you like, I don't have anything else to add: RAW.

waspc/tests/Analyzer/EvaluatorTest.hs
```diff
@@ -178,7 +178,8 @@ spec_Evaluator = do
         let typeDefs =
               TD.addDeclType @Business $
                 TD.addEnumType @BusinessType $
-                  TD.addDeclType @Person $ TD.empty
+                  TD.addDeclType @Person $
```
Martinsos, line 181: but why
new ormolu?

waspc/tests/Analyzer/EvaluatorTest.hs
Martinsos: I would maybe handle it with `either`, I think there is such function, like `maybe`. BUt this is also ok at the end, doesn't matter too much. `aj` is a bit poor name but ok it is very localized.

waspc/tests/Analyzer/TypeCheckerTest.hs
```diff
@@ -68,7 +68,7 @@ spec_TypeChecker = do
       it "Properly hoists declarations" $ do
         let mAst = P.parseStatements "llnode Head { value: 2, next: Tail } llnode Tail { value: 3 }"
         mAst `shouldSatisfy` isRight
-        let (Right ast) = mAst
+        let ast = fromRight (error "Impossible") mAst
```
Martinsos: This is ok! Certainly nicer than casing, right? One thing I love doing is writing it like this when there is error handling:
```hs
let ast = mAst `fromRight` (error "Impossible")
```
I know this won't work because of order or args, but the idea is that I prefer putting error handling "later", because indeed I do n't want to read it first. But here this is ok I think.

waspc/tests/Generator/WebAppGeneratorTest.hs
```diff
@@ -98,7 +98,10 @@ spec_WebAppGenerator = do
     --   that they will successfully be written, it checks only that their
     --   destinations are correct.
     it "Given a simple AppSpec, creates file drafts at expected destinations" $ do
-      let (_, Right fileDrafts) = runGenerator $ genWebApp testAppSpec
+      let (_, generatorResult) = runGenerator $ genWebApp testAppSpec
+      fileDrafts <- case generatorResult of
+        Right fds -> pure fds
+        Left err -> expectationFailure ("genWebApp failed: " ++ show err) >> error "unreachable"
```
Martinsos:
```suggestion
      fileDrafts <- case (snd $ runGenerator $ genWebApp testAppSpec) of
        Right fds -> pure fds
        Left err -> expectationFailure ("genWebApp failed: " ++ show err) >> error "unreachable"
```

but yeah would maybe be even nicer if you used `fromRight` or something (not sure if doable here). This is also not terrible though.

waspc/tests/Analyzer/EvaluatorTest.hs
sodic: Yes, we had to update it and they changed the formatting rules.

waspc/cabal.project
sodic: I was sure I explained this thing. Anyway...

So, for some reason, LSP no longer worked in tests until I added this. This was caused by HLS trying to run `cabal v2-repl waspc:test:waspc-tests` and getting this message:
```
Cannot open a repl for the test suite 'waspc-tests' because the solver did not find a plan that included the test suites for waspc-0.18.1. It is probably worth trying again with test suites explicitly enabled in the configuration in the cabal.project{.local} file. This will ask the solver to find a plan with the test suites available. It will either fail with an explanation or find a different plan that uses different versions of some other packages. Use the '--dry-run' flag to see package versions and check that you are happy with the choices.
```

So enabling tests specifically with `tests: True` is the way to do it. The real question is why it hasn't been a problem so far. The answer seems to be "If the dependency tree is simple enough/similar enough, cabal will include tests in the default build plan." Until now, that seemed to be the case. But it isn't anymore.

I didn't fully understand what was happening, but I don't think it's worth digging deeper. I can dig in if you think it's necessary.

waspc/tests/Analyzer/TypeCheckerTest.hs
sodic: > Certainly nicer than casing, right?

It's shorter but kind of hacky - instead of relying on the type system (as we do with the case), we rely on previous lines of code. It's fine, I know, but still.

waspc/src/Wasp/Version.hs
Martinsos: I don't know, but I wouldn't worry too much, doesn't sound like a potential issue, right? If we do, it will fail in development anyway.

waspc/cabal.project
Martinsos: No need, I just wanted to understand the change.

Here is screenshot from cabal's docs:
image

From this, it sounds like potentially we are doing something wrong, and then we really just patched it this way -> it says in most normal cases this shouldn't be needed. And I doubt our case is so special. So question is, what are we doing that is so weird that this is required for us?

I see that in the error messages it told you to enable them though.

So I guess what I am suspicious of is that the real problem is somewhere in who depends on whom in our cabal file, and that we just "patched" over the real problem. But I am not sure.

Maybe ask on cabal github? Or haskell reddit? Or we should dig a bit deeper, understand exactly what does that error message mean?

EDIT: OK I GPTed a bit more, it seems that this might be related to speciifcally how HLS is calling cabal in this case, and that in such case, `tests: True` actually is needed. I am not sure if that is true or not for real, but it sounds like it might be.

Ok, I would say let's not waste more time then, but maybe let's add one comment explainnig why we have that (that it is needed for HLS), because I don't see how else can someobdy know that (it certinaly isn't obvious from cabal.project docs, quite the opposite, it sounds like tests: True is redundant, at least to me).

waspc/tests/Analyzer/TypeCheckerTest.hs
Martinsos: Ah because of `shouldSatisfy`? OK I missed that one because it wasn't a change. Ok, in any case, I trust your judgement, and I don't mind us going with case if this feels too hacky, sometimes it is better to not be too "smart".

RAW

waspc/cabal.project
sodic: > EDIT: OK I GPTed a bit more, it seems that this might be related to speciifcally how HLS is calling cabal in this case, and that in such case, tests: True actually is needed. I am not sure if that is true or not for real, but it sounds like it might be.

Lol I went down the exact same path. It told me the same thing in one iteration. In another one, it told me something else.

I tried to pin down why it didn't include the test suite in the build plan (i.e., where the dependency conflict is coming from), but I decided it only makes sense to do that after I update fully (to GHC 9.6.7). Perhaps not even then because, if the behavior is as implicit and difficult to reverse engineer, I prefer the flag.

Here's one of my chats if you're curious: https://chatgpt.com/c/6909fe86-584c-8328-bd85-6b2d94d12f7d

sodic: I'll add a comment in any case.

waspc/src/Wasp/Version.hs
sodic: You say "I don't know" but I didn't ask a question :)

I wouldn't know what else to do anyway.

waspc/tests/Generator/WebAppGeneratorTest.hs
sodic: Used `fromRight`.

PR #3287, opened 2025-11-01
Update GHC to 9.4.8

sodic:
Only look at changes in these files:
```
.github/actions/setup-haskell/action.yaml
waspc/cabal.project
waspc/src/Wasp/AppSpec/App/Auth.hs
waspc/src/Wasp/AppSpec/Job.hs
waspc/cli/src/Wasp/Cli/Command/Build.hs
waspc/waspc.cabal
waspc/dev-tool/project
```
Everything else is Ormolu's autoformat.

### DIFF
--- a/waspc/waspc.cabal
+++ b/waspc/waspc.cabal
@@ -6,7 +6,7 @@ cabal-version: 2.4
 --    Consider using hpack, or maybe even hpack-dhall.
 
 name:           waspc
-version:        0.18.1
+version:        0.18.2
 description:    Please see the README on GitHub at <https://github.com/wasp-lang/wasp/waspc#readme>
 homepage:       https://github.com/wasp-lang/wasp/waspc#readme
 bug-reports:    https://github.com/wasp-lang/wasp/issues
--- a/.github/workflows/ci-waspc-build.yaml
+++ b/.github/workflows/ci-waspc-build.yaml
@@ -76,16 +76,13 @@ jobs:
               apk add zlib-dev zlib-static
 
           - name: macos-x86_64
-            runner: macos-15-intel # Latest image still based on Intel architecture that can be used for free
-
+            runner: macos-15-intel
             # macOS's syscalls are private and change between versions, so we
-            # can't statically link the binary. However, on macOS programs link
-            # to `libSystem`, which is quite stable between releases, so we're
-            # fine depending on it.
+            # can't statically link the binary.
             static: false
 
           - name: macos-aarch64
-            runner: macos-latest # Latest macOS images are already Apple Silicon-based
+            runner: macos-15 # By default an Apple Silicon-based runner.
             static: false # Check the comment above for why we can't statically link on macOS
             install-deps: |
               # We need to install llvm@13 for building on Apple Silicon (prebuilt libraries
@@ -148,7 +145,7 @@ jobs:
 
   build-universal:
     needs: build
-    runs-on: macos-latest
+    runs-on: macos-15
     steps:
       - name: Download macOS binaries
         uses: actions/download-artifact@v5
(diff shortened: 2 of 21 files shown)


waspc/waspc.cabal
sodic, line 1: All version bumps are driven by either:
- Core libraries mismatch (template haskell, base, ghc-prim)
- Dependency conflicts driven by those changes.

Here's a Claude-assisted summary.

### 1. template-haskell: ^>= 2.18.0 → ^>= 2.19.0
- GHC 9.4.8 ships with template-haskell-2.19.0.0
- The previous constraint `^>= 2.18.0` means `>= 2.18.0 && < 2.19.0`, which excludes 2.19.0.0
- GHC to template-haskell mapping:
  - GHC 9.2.8 → template-haskell 2.18.0.0
  - GHC 9.4.8 → template-haskell 2.19.0.0

### 2. lens: ^>= 5.1 → ^>= 5.2
- lens-5.2+ has explicit support for template-haskell 2.19.0.0 (GHC 9.4)
- lens-5.1.x requires template-haskell < 2.19, which conflicts with GHC 9.4.8

### 3. aeson: ^>= 2.0 → ^>= 2.2
- GHC 9.4.8 ships with base-4.17.2.1 which includes ghc-prim-0.9.1
- aeson-2.0.x requires ghc-prim < 0.9, excluding version 0.9.1
- aeson-2.2.x supports ghc-prim-0.9.1 and is compatible with GHC 9.4.8

### 4. lens-aeson: ^>= 1.1.3 → ^>= 1.2
- lens-aeson-1.1.3 requires aeson < 2.1
- With aeson updated to 2.2, we need lens-aeson-1.2+ which supports aeson 2.2

waspc/src/Wasp/AppSpec/App/Auth.hs
sodic, line 1: Looks like they deprecated ambiguous record fields pretty quickly so I had to start using the `OverloadedRecordDot`. We plan to use it anyway, so this seemed like the best solution

Previous context: https://github.com/wasp-lang/wasp/pull/3286#discussion_r2483802679

waspc/src/Wasp/AppSpec/Job.hs
sodic, line 1: Same story as with Auth.hs: https://github.com/wasp-lang/wasp/pull/3287#discussion_r2489620174

waspc/dev-tool.project
```diff
@@ -1,14 +1,14 @@
...
 
 -- Speeds up builds by parallelizing them over the number of available CPUs.
 jobs: $ncpus
 
 constraints:
-  ormolu ==0.4.0.0,
+  ormolu ==0.5.3.0,
   stan ==0.2.1.0,
-  hlint ==3.5.0,
+  hlint ==3.6.1,
```
sodic, line 11: We had to update these two because their bases were too old.

@Martinsos Do we ever use anything except Ormolu? If not, should I remove those?

Martinsos: Yeah I would actually love to drop these.

I learned that HLS actually brings hlint, stan and ormolu with it, so we are getting those already as code actions and formatting. Check it out here, under plugins: https://haskell-language-server.readthedocs.io/en/latest/features.html .

So ideally, we wouldn't have this dev-tool.project at all, instead we would piggy back on what HLS is exposing.

That said, I don't know if that is possible -> can we externally use call these plugins as they come with HLS?

If so, then all we care is that we all use the same version of HLS, and we are good. We can have editors use the formatter as provided from HLS, that should work. THe question I guess is can we, in CI, also somehow trigger the same formatter as is used in HLS, ideally that exact one.

If not, then we probably do need this file so we can at least run formatter, and maybe also stan and hlint, if those are anyway ran by HLS.

Maybe I complicated this, but in summary, HLS brings and runs all of these three when using it. But we also need it in CI to enforce stuff. So do we have to specify and install them separately, or can we somehow get them from HLS?

We might want to create separate issue for this.

sodic: https://github.com/wasp-lang/wasp/issues/3306

Issue #3220, opened 2025-10-02
Automate or document Playwright requirements for app e2e tests


sodic:
It’s not obvious we should install playwright when running the tests. I had to look at what the CI does to figure it out.

Let's make it obvious or, even better, unnecessary (by taking care of it automatically).

FranjoMindek:
In this PR: https://github.com/wasp-lang/wasp/pull/3197, i make it so `run` script takes care of installing `playwright` browser requirements (`npx playwright install --with-deps`).

PR #3292, opened 2025-11-04
Add script to install `playwright` deps, use it automatically before each `test`

FranjoMindek:
Adds new `test:install-deps` command to all projects using `playwright`.
Call it before running the `test` command.
In case you already have deps installed the `test:install-deps` command won't do anything.

Removes temporary `playwright` deps installs we had in the `run` script (now part of `npm run test`).
Also removes explicit `playwright` deps installs from CI `.yaml` files (now part of `npm run test`).

Fixes: #3220

### DIFF
--- a/examples/ask-the-documents/package.json
+++ b/examples/ask-the-documents/package.json
@@ -9,7 +9,7 @@
     "env:pull": "npx dotenv-vault@latest pull development .env.server",
     "env:push": "npx dotenv-vault@latest push development .env.server",
     "test": "npm run test:install-deps && DEBUG=pw:webserver playwright test --config e2e-tests/",
-    "test:install-deps": "npx playwright install --with-deps"
+    "test:install-deps": "playwright install --with-deps"
   },
   "dependencies": {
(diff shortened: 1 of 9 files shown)


examples/ask-the-documents/package.json
```diff
@@ -8,7 +8,8 @@
   "scripts": {
     "env:pull": "npx dotenv-vault@latest pull development .env.server",
     "env:push": "npx dotenv-vault@latest push development .env.server",
-    "test": "DEBUG=pw:webserver playwright test --config e2e-tests/"
+    "test": "npm run test:install-deps && DEBUG=pw:webserver playwright test --config e2e-tests/",
+    "test:install-deps": "npx playwright install --with-deps"
```
infomiho:
```suggestion
    "test:install-deps": "playwright install --with-deps"
```

I think we can drop the `npx` since Playwright is installed as a dev dep?

FranjoMindek: Yes.

PR #3295, opened 2025-11-04
Remove ghcup workaround

cprecioso:
We had a workaround to avoid an issue with an old GitHub runner image version: we did a `chown` of the ghcup folder.
- https://github.com/actions/runner-images/issues/7061

This issue was fixed upstream and the workaroud is no longer needed.
CI runs correctly: https://github.com/wasp-lang/wasp/actions/runs/19065146636/job/54453848914?pr=3295


sodic (approved):
Ship it!

PR #3293, opened 2025-11-04
CI: Update runner to macOS 15

cprecioso:
### Description

Replaces the `macos-13` image with `macos-15-intel`. macOS 13 is deprecated in GitHub Runners and will be intentionally disabled on Tuesdays until Dec 8th when it will be disabled for good. This will make our CI fail.

The only other supported macOS Intel version is macOS 15, so we migrate to that.
No foreseeable complications, and build runs correctly. The binary now declares its minimum version to be macOS 15, so this raises our minimum requirement.

PR #3296, opened 2025-11-04
CI: Pin macOS version for Apple Silicon

cprecioso:
In my previous research I declared that libc was compatible between versions in macOS so that it was not needed to pin macOS to an older version.

While researching for #3293, I researched a bit more and learned that macOS binaries do declare a minimum compatible version, and they might be using new features aside from libc.

Therefore, we do need to avoid automatically updating our macOS version if possible. `macos-latest` currently points to `macos-15`, so I updated the reference to that. It also has the benefit of being the same version as for Intel-based macOS (since #3293), so it doesn't differ based on architecture.

### DIFF
--- a/.github/workflows/ci-waspc-build.yaml
+++ b/.github/workflows/ci-waspc-build.yaml
@@ -76,16 +76,13 @@ jobs:
               apk add zlib-dev zlib-static
 
           - name: macos-x86_64
-            runner: macos-15-intel # Latest image still based on Intel architecture that can be used for free
-
+            runner: macos-15-intel
             # macOS's syscalls are private and change between versions, so we
-            # can't statically link the binary. However, on macOS programs link
-            # to `libSystem`, which is quite stable between releases, so we're
-            # fine depending on it.
+            # can't statically link the binary.
             static: false
 
           - name: macos-aarch64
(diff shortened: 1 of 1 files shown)


cprecioso:
* **#3297** <a href="https://app.graphite.com/github/pr/wasp-lang/wasp/3297?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3296** <a href="https://app.graphite.com/github/pr/wasp-lang/wasp/3296?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/wasp-lang/wasp/3296?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3293** <a href="https://app.graphite.com/github/pr/wasp-lang/wasp/3293?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

cprecioso:
Yes

Issue #3165, opened 2025-09-18
Update outdated code examples (e.g. README)


FranjoMindek:
Currently our `README.md` example is outdated.
It shows code for Wasp `"^0.15.0"`.

We should update all our non-versioned code examples to latest Wasp version.

TODO: see if more code examples are outdated

FranjoMindek:
Solved with #3298 and #3315

PR #3298, opened 2025-11-04
Refactor repo `README.md`

FranjoMindek:
I've changed most of the README but im small manners.

One bigger change is that I reordered the titles, now its:
`"What is Wasp" -> "Get started"`
instead of:
`"What is Wasp" -> "Get Swag"`
Which seems more natural to me (we are wasting their time less).

Also I removed the `"This repository"` title, which seems useless? At least to me.

The `"AI"` title still feels out of place to me. I would rather remove it, but I feel that is a controversial one. 
@Martinsos what do you think about that?

Rest of not so obvious decisions are explained in separate comments below.

You can check it out rendered here: https://github.com/wasp-lang/wasp/tree/franjo/update-main-readme

Part of: [#3165](https://github.com/wasp-lang/wasp/issues/3165)

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,6 @@
     <a href="https://wasp.sh">Website</a>
     |
     <a href="https://wasp.sh/docs">Docs</a>
-    |
-    <a href="https://e44cy1h4s0q.typeform.com/to/EPJCwsMi">Deployed? Get swag! 👕</a>
   </div>
   <div align=center>
     <a href="https://discord.gg/rzdnErX">Discord</a>
@@ -26,6 +24,9 @@
     <a href="https://x.com/WaspLang">Twitter</a>
     |
     <a href="https://www.youtube.com/@wasplang">Youtube</a>
+  </div>
+  <div align=center>
+    <a href="https://e44cy1h4s0q.typeform.com/to/EPJCwsMi">Deployed? Get swag! 👕</a>
   </div>
 </div>
 


README.md
```diff
@@ -1,23 +1,44 @@
...
+</div>
 
 ---
 
-<p align="center">
-  <a href="https://wasp.sh">Web page</a> | <a href="https://wasp.sh/docs">Docs</a> | <a href="https://e44cy1h4s0q.typeform.com/to/EPJCwsMi">Deployed? Get swag! 👕</a>
-</p>
-
-<br>
+## What is Wasp?
```
FranjoMindek, line 35: New title for the first text.

README.md
```diff
@@ -26,11 +47,15 @@ Build your app in a day and deploy it with a single CLI command!
...
 - 🚀 **Quick start**: Due to its expressiveness, you can create and deploy a production-ready web app from scratch with very few lines of concise, consistent, declarative code.
 - 😌 **No boilerplate**: By abstracting away complex full-stack features, there is less boilerplate code. That means less code to maintain and understand! It also means easier upgrades.
-- 🔓 **No lock-in**: You can deploy the Wasp app anywhere you like. There is no lock-in into specific providers; you have full control over the code (and can actually check it out in .wasp/ dir if you are interested ).
+- 🔓 **No lock-in**: You can deploy the Wasp app anywhere you like. There is no lock-in into specific providers; you have complete control over the code (and can actually check it out in `.wasp/` directory if you are interested ).
 
-### Features
+#### Notable features
 
-🔒 Full-stack Auth, 🖇️ RPC (Client <-> Server), 🚀 Simple Deployment, ⚙ ️Jobs, ✉️ Email Sending, 🛟 Full-stack Type Safety, ...
+- 🔒 Full-stack Auth & Type Safety
```
FranjoMindek: Connected auth and type safety into one "security" point.

README.md
```diff
@@ -57,7 +83,11 @@ query getTasks {
   fn: import { getTasks } from "@server/tasks.js", // Your Node.js code.
   entities: [Task] // Automatic cache invalidation.
 }
+```
+
+And a Prisma schema for the database:
```
FranjoMindek, line 84: I've split `.wasp` and Prisma files into separate code blocks.

README.md
```diff
@@ -67,73 +97,69 @@ model Task { // Your Prisma data model.
...
 The core of Wasp is built in Haskell, but there are also a lot of non-Haskell parts of Wasp, so you will certainly be able to find something for you!
 
-Even if you don't plan to submit any code, just joining the discussion on discord [![Discord](https://img.shields.io/discord/686873244791210014?label=chat%20on%20discord)](https://discord.gg/rzdnErX) and giving your feedback is already great and helps a lot (motivates us and helps us figure out how to shape Wasp)!
+Even if you don't plan to submit any code, just joining the discussion on Discord [![Discord](https://img.shields.io/discord/686873244791210014?label=chat%20on%20discord)](https://discord.gg/rzdnErX) and giving your feedback is already great and helps a lot (motivates us and helps us figure out how to shape Wasp)!
 
 You can also:
 
-- :star: Star this repo to show your interest/support.
-- :mailbox: Stay updated by subscribing to our [email list](https://wasp.sh#signup).
-- :speech_balloon: Join the discussion at https://github.com/wasp-lang/wasp/discussions .
```
FranjoMindek, line 129: I tried opening the discussion and the link was dead. So I removed it.

README.md
```diff
@@ -1,23 +1,44 @@
...
+    </a>
+    <a href="https://discord.gg/rzdnErX">
+      <img alt="discord" src="https://img.shields.io/discord/686873244791210014?label=chat%20@%20discord"/>
+    </a>
+  </div>
+  <br />
+  <!-- LINKS -->
+  <div align=center>
+    <a href="https://wasp.sh">Website</a>
+    <span>|</span>
```
Martinsos: What is with all the spans, why do we need those, why not just `|`?

README.md
```diff
@@ -1,23 +1,44 @@
...
+    <a href="https://wasp.sh/docs">Docs</a>
+    <span>|</span>
+    <a href="https://discord.gg/rzdnErX">Discord</a>
+    <span>|</span>
+    <a href="https://x.com/WaspLang">Twitter</a>
+    <span>|</span>
+    <a href="https://www.youtube.com/@wasplang">Youtube</a>
```
Martinsos, line 26: Hm I don't know if we need Youtube :D.
Could we have two rows of stuff? Maybe first one is webpage and docs, and the second is then discord, twitter, youtube, deployed?
We also have discord repeated both as a badge and as a link, but hm I don't know maybe it is ok. I wonder if we should make it all badges, probably not.

Btw there are these blue "link" lines in between badges, that haven't been there before:
image

README.md
```diff
@@ -26,11 +47,15 @@ Build your app in a day and deploy it with a single CLI command!
 
 - 🚀 **Quick start**: Due to its expressiveness, you can create and deploy a production-ready web app from scratch with very few lines of concise, consistent, declarative code.
 - 😌 **No boilerplate**: By abstracting away complex full-stack features, there is less boilerplate code. That means less code to maintain and understand! It also means easier upgrades.
-- 🔓 **No lock-in**: You can deploy the Wasp app anywhere you like. There is no lock-in into specific providers; you have full control over the code (and can actually check it out in .wasp/ dir if you are interested ).
+- 🔓 **No lock-in**: You can deploy the Wasp app anywhere you like. There is no lock-in into specific providers; you have complete control over the code (and can actually check it out in `.wasp/` directory if you are interested ).
 
-### Features
+#### Notable features
```
Martinsos: I would turn this back into just "Features" and also make it level 3 heading (I am not sure why you went with level 4?), while also adding bullet point with `...`. That makes stronger impression that there are more features IMO.

README.md
Martinsos: Hm I don't think we should, they are not really connected things. This confused me quite a bit as I was reading it. Type safety is not about security at all in a way, it is about code quality. Anyway, I would keep them separate as they were, "Full-stack Type Safety" was nice, with that nice emoji.

README.md
```diff
@@ -67,73 +97,69 @@ model Task { // Your Prisma data model.
...
-This is the main repo of the Wasp universe, containing core code (mostly `waspc` - Wasp compiler) and the supporting materials.
+Wasp comes with experimental AI code generator to help you kickstart your next Wasp project. You can use it via `wasp new` in the CLI (select the `ai-generated` option) if you provide your OpenAI keys. Alternatively, you can use our [Mage web app](https://usemage.ai), in which case our OpenAI keys are used in the background.
 
-# Project status
+## Project status
 
-Currently, Wasp is in beta, with most features fleshed out and working well.
-However, there are still a lot of improvements and additions that we have in mind for the future, and we are working on them constantly, so you can expect a lot of changes and improvements in the future.
+Currently, Wasp is in beta, with most features fully developed and functioning well.
+However, we still have many improvements and additions in mind for the future, and we are continually working on them. As a result, you can expect numerous changes and improvements in the future.
```
Martinsos, line 133: I see you did some changes to the text, minor ones, and while I don't mind them, for some it is not clera if it is an improvement -> what was your reasoning, how did you do these? Was it grammarly suggesting them, your own inspiration?

README.md
Martinsos: Ah yes we removed those! Nice job on adding the roadmap. Actually we should probasbly link the roadmap also under the part where we explain we are in beta and will be changing stuff.

README.md
FranjoMindek: Firstly I went by myself, mostly did technical stuff like missing code blocks, or ending too long sentences.
Then I put it inside of grammarly, and picked the results which I liked from it (not everything).

README.md
FranjoMindek: Before it was just a normal paragraph of text, it wasn't a list, I just didn't want to make it too long.
I that is also why I connected auth and typesafety. I do see the appeal of `...`.

For level4, it seems like it could work as a subtitle, rather tan separately.

README.md
FranjoMindek: I think it's good to interlink community stuff so it can be more easily discovered.
I still think it's okay to have a single row.

I'll see what I'll do regarding badges.


Regarding the blue line, I also noticed it at 2nd sponsor block:
image

I've experimented a bit and found out that this creates the blue lines:
```md
<a href="https://github.com/MarianoMiguel">
	<img src="https://github.com/MarianoMiguel.png" width="50px" alt="MarianoMiguel" />
</a>
```
while this does not:
```md
<a href="https://github.com/MarianoMiguel"><img src="https://github.com/MarianoMiguel.png" width="50px" alt="MarianoMiguel" /></a>
```

I wanted to make it more readable, but I guess this doesn't really work fully like `.html` file.

README.md
FranjoMindek: We can do any one of those two.
I just thought this looked out of place:
image

I did revert it to `|`.

README.md
FranjoMindek: Okay from what I can see, badges are usually used only for project status (not socials) and are usually only at the beginning of the README, i.e. not further below.

IMO we should drop the discord badge.
People usually put license, version, tests (ci, coverage...), PRs status

README.md
FranjoMindek: One good idea would be for each bullet point to lead to a different docs page.
Like, "we have proof".

README.md
Martinsos: Ok do as you like, but this is easier to read for me.
I would maybe even do
```html
<a>...</a>
| <a>...</a>
| <a>...</a>
```

although hah I am not sure if that is controversial regarding formatting.

ANyway, raw, and I like this as it is.

README.md
Martinsos: I wouldn't drop the discord badge because it shows number of people active in the chat, which signals healthiness / is inviting. I have seen in multiple places badges used for stuff like this (for chats/communiities + number of people), so that is all right, let's keep the discord badge.
We can have it duplicated though (also discord mentioned below).

The reason why I was saying we go two rows is because web/docs are more important, socials come second, and now with all in one row, it is hard to distiguish what is less and what is more important. Maybe there is other way to make socials less prominent, be it usign their icons instead of text or something, I am not sure. But I would seriously consider making somehow web/docs stand out (my only idea is still them being in the first row).

README.md
Martinsos: Yup, that also might be nice (links)! Although I guess they can just go to docs to check it out. Btw I am ok Features being lvl3 heading, that sounds right to me.

RAW

README.md
FranjoMindek: We've talked about this in the DMs.

I've restored discord badge at the top, as it is our main source of communication.
The open below in text is replaced by normal link.

We now have 3 rows of links in the initial section:
- important links
- socials
- get swag

Issue #3073, opened 2025-08-12
`open-saas` development version is out of sync with `wasp` development version


FranjoMindek:
Currently, we make it so that `next` version of `open-saas` (`main` branch) can be developed with `wasp` release version.
If we do breaking changes on `next` version of `wasp` (`wasp-cli` from now on), we create PRs on `open-saas` and leave them open. We only merge the just before releasing the new `open-saas` version, after we release a new `wasp` version.

While this makes it so that people can contribute to `open-saas` without having to build `wasp` themselves, this is counter-intuitive because `open-saas` really is a `wasp` starter template, and like every other template, should be developed with `wasp-cli`.

This also make it hard for `open-saas` to work with other system we built around the `wasp-cli`.
Like our testing evironment.

e.g. as soon as we make a breaking change on `wasp-cli`, `starter-e2e-tests.yaml` workflow will fail, because it tries to run the `open-saas` with `wasp-cli`, but for `open-saas` we don't merge these PRs
https://github.com/wasp-lang/wasp/actions/runs/16910912920/job/47912029896?pr=3070

Instead I propose a new approach:
1) We immediately merge `wasp-cli` changes to `open-saas` `main` branch instead of leaving them open
2) We should provide make it so that wasp installer script (`curl -sSL https://get.wasp.sh/installer.sh | sh`) can install `nightly` builds
   - that means we should expand it with `nightly` version: `curl -sSL https://get.wasp.sh/installer.sh | sh -s -- -v nightly`
   - the way to do this would be to create a release artifact each time we merge a commit to `wasp-lang/wasp`'s `main` branch
   - the `nightly` version would then install that version of `wasp`
   - this would allow people to contribute to `open-saas` easily, without having to build the `wasp` themsleves

infomiho:
Since Wasp is distributed via Github releases, I guess each new PR to `main` would release a new "pre-release" release. The Wasp install script will consider the last release as `nightly`?

cprecioso:
I'm in favor of tying `open-saas#main` with `wasp#main`. And I'm in favor of publishing nightly releases. 

But I don't think we should be polluting our releases with every commit, people can get confused, and there could be two pages of "pre-releases" before an actual release.

Nightly releases should be published to some other feed (e.g. npm, gh packages, gh containers, aws bucket, cloudflare workers)

Martinsos:
[...]

PR #3304, opened 2025-11-05
Create fetch nightly action

cprecioso:
- Part of https://github.com/wasp-lang/wasp/issues/3073
- Preparation for https://github.com/wasp-lang/get-wasp-sh/pull/6

Creates an action for _other repos_ to fetch a "nightly" version created from the last correct CI run for a given branch (by default `main`).

This makes sense to keep in this repo because:

1. It depends on specific implementation details of stuff in this repo (e.g. the name of the workflow that builds the tarballs)

3. The actions system allows us to share this action out without any trouble, just with:

  ```yaml
  uses: wasp-lang/wasp/.github/actions/fetch-nightly-cli@main
  ```

### DIFF
--- a/.github/actions/fetch-nightly-cli/action.yaml
+++ b/.github/actions/fetch-nightly-cli/action.yaml
@@ -54,6 +54,6 @@ runs:
         RUN_ID: ${{ steps.get_run_id.outputs.run_id }}
         OUTPUT_DIR: ${{ inputs.output-dir }}
         ARTIFACT_NAME:
-          # This is the name of the artifact (not the file), as defined in:
-          # .github/workflows/ci-waspc-build.yaml
+          # This is the name of the artifact (not the file), which must be in sync with:
+          # /.github/workflows/ci-waspc-build.yaml
           "wasp-cli-${{ runner.os == 'Linux' && 'linux' || 'macos' }}-${{ runner.arch == 'ARM64' && 'aarch64' || 'x86_64' }}"
--- a/.github/workflows/ci-waspc-build.yaml
+++ b/.github/workflows/ci-waspc-build.yaml
@@ -147,7 +147,10 @@ jobs:
       - uses: actions/upload-artifact@v4
         with:
           path: ./waspc/artifacts/*
-          name: wasp-cli-${{ matrix.env.name }}
+          name:
+            "wasp-cli-${{ matrix.env.name }}"
+            # This name must be in sync with the artifact name used in:
+            # /.github/actions/fetch-nightly-cli/action.yaml
           if-no-files-found: error
 
   build-universal:


.github/actions/fetch-nightly-cli/action.yaml
```diff
@@ -0,0 +1,58 @@
...
+            --limit 1 \
+            --json databaseId \
+            --jq '.[0].databaseId'
+        )
+        echo "run_id=$run_id" >> $GITHUB_OUTPUT
+      env:
+        GH_REPO: wasp-lang/wasp
+        GH_TOKEN: ${{ inputs.token }}
+        branch: ${{ inputs.branch }}
+        workflow_name:
```
infomiho: Why not go with all uppercase as with the other env vars? It feels like a GH action convention (I might be wrong).

.github/actions/fetch-nightly-cli/action.yaml
```diff
@@ -0,0 +1,58 @@
...
+          # This name must be in sync with the name used in:
+          # /.github/workflows/ci.yaml
+          "CI"
+
+    - name: Download artifact
+      if: matrix.args.needs-download == true
```
infomiho: This looks like a leftover from extraction?

.github/actions/fetch-nightly-cli/action.yaml
```diff
@@ -0,0 +1,58 @@
...
+      shell: bash
+      run: |
+        gh run download "$run_id" --name "$artifact_name" --dir "$output_dir"
+      env:
+        GH_REPO: wasp-lang/wasp
+        GH_TOKEN: ${{ inputs.token }}
+        run_id: ${{ steps.get_run_id.outputs.run_id }}
+        output_dir: ${{ inputs.output-dir }}
+        artifact_name:
```
infomiho: I'd also go with all uppercase names:

```suggestion
        RUN_ID: ${{ steps.get_run_id.outputs.run_id }}
        OUTPUT_DIR: ${{ inputs.output-dir }}
        ARTIFACT_NAME:
```

.github/actions/fetch-nightly-cli/action.yaml
```diff
@@ -0,0 +1,58 @@
...
+          # This is the name of the artifact (not the file), as defined in:
+          # .github/workflows/ci-waspc-build.yaml
+          "wasp-cli-${{ runner.os == 'Linux' && 'linux' || 'macos' }}-${{ runner.arch == 'ARM64' && 'aarch64' || 'x86_64' }}"
```
infomiho, line 59: This feels fragile since it's very low level and probably repeated in multiple places - is there a way to compute this in one place and use it all of our CI workflows?

.github/actions/fetch-nightly-cli/action.yaml
cprecioso: Resolving this thread in favor of the other comment https://github.com/wasp-lang/wasp/pull/3304#discussion_r2493776245

.github/actions/fetch-nightly-cli/action.yaml
cprecioso: I tried to match the energy of the shell, where uppercase names are globals or env vars for other programs, and lowercase names are the script's own variables.

.github/actions/fetch-nightly-cli/action.yaml
cprecioso: I get you, I'm also not in love with it. I _could_ extract it to its own action that just computes this name, but I don't think it's worth it for now. I will do it if you think it's better though.

.github/actions/fetch-nightly-cli/action.yaml
infomiho: Having different casings implies something that's not true - that there is a difference in availability of these env variables - but all of them are globally available in the step? That sounds like a reason enough not to do different casings IMHO

.github/actions/fetch-nightly-cli/action.yaml
infomiho: Can we analyze the instances where we compute the artifact name and maybe come up with a helper script? Let's give it go and see what comes of it - if it's more bad than good, I'm okay with keeping it inline like this.

cprecioso: I tried it out locally already. This only shows up in the `ci-waspc-build.yaml` workflow and here. I don't want to extract it to a regular script, because then we'd have to checkout the full Wasp repo whenever we run this action in another repo, just to run that script. That's why I think it's not worth it for now.

infomiho: Ok fair enough, if we can't do anything meaningful in code, let's add some comments in both places to try to keep these two places in sync.

cprecioso: Ready

PR #3312, opened 2025-11-06
Add fetch-nightly-cli GitHub Action documentation

cprecioso:
## Summary
Add documentation for the fetch-nightly-cli GitHub Action, which enables testing against the latest nightly CLI build from Wasp's CI runs.

## Details
This README explains how to use the action to fetch the latest build from `main` or a specific branch/PR, with examples for basic usage, custom branches, and output directories.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/actions/fetch-nightly-cli/README.md
+++ b/.github/actions/fetch-nightly-cli/README.md
@@ -13,9 +13,9 @@ A GitHub Action that fetches the latest nightly CLI build from Wasp's CI runs. T
 
 This action provides an internal "nightly" system for Wasp, allowing other repositories and workflows to download and install the binary from the latest successful CI run. The action:
 
-- Fetches the most recent successful build from the specified branch (defaults to `main`)
-- Downloads the appropriate binary for the current OS and architecture
-- Can be used to test against PRs by specifying a different branch
+- Fetches the most recent successful build from the specified branch (defaults to `main`).
+- Downloads the appropriate binary for the current OS and architecture.
+- Can be used to test against PRs by specifying a different branch.
 
 ## Usage
 


.github/actions/fetch-nightly-cli/README.md
```diff
@@ -0,0 +1,53 @@
...
+> - run: curl -sSL https://get.wasp.sh/installer.sh | sh -- -v [version]
+> ```
+
+## Background
+
+This action provides an internal "nightly" system for Wasp, allowing other repositories and workflows to download and install the binary from the latest successful CI run. The action:
+
+- Fetches the most recent successful build from the specified branch (defaults to `main`)
+- Downloads the appropriate binary for the current OS and architecture
+- Can be used to test against PRs by specifying a different branch
```
Martinsos: Again me, but periods at the end.

.github/actions/fetch-nightly-cli/README.md
```diff
@@ -0,0 +1,53 @@
...
+### Basic example
+
+Fetch the latest nightly build from `main` and install it:
+
+```yaml
+- uses: wasp-lang/wasp/.github/actions/fetch-nightly-cli@main
+- run: curl -sSL https://get.wasp.sh/installer.sh | sh -- -f wasp-*.tar.gz
+```
+
+The wildcard `wasp-*.tar.gz` is used because the artifact has a different name depending on the OS and architecture. The `-f` flag tells the installer to use the locally downloaded file.
```
Martinsos, line 31: Is there any chance it might get the wrong .tar.gz? Maybe tehre is some other wasp-*.tar.gz on the disk that we are not expecting, from some previous step? Are nightlies maybe marked better, with `main` or `nightly` or something?

cprecioso: There is, but if there's more than one argument specified, the installer will fail.
I am waiting for GitHub to reply on the issue I opened in their repo https://github.com/cli/cli/issues/12097, and will see if I can make it more robust once they do.

Issue #3247, opened 2025-10-10
We should run unit test for Tutorial Actions Executor in the CI


infomiho:
We implemented the unit tests in this PR https://github.com/wasp-lang/wasp/pull/2732 but in order to keep the PR as lean as possible, we didn't want to add the CI improvements in this PR.

We should run the unit tests when there are changes done to the Tutorial Actions Executor.

Martinsos:
I put this into product backlog, since we want to do it relatively soon, not wait.

infomiho:
I've moved this as a child issue of the open sourcing effort - since it'll go into a separate repo, makes more sense to add the CI in the separate repo.

PR #3316, opened 2025-11-06
Add e2e snapshot tests to the tutorial app executor

infomiho:
## Description

Follow-up on #2732 

It adds e2e snapshot tests using Vitest. There is a mock tutorial that the executor parses and executes. We take snapshots of the file list, certain files that are patched and the `git log`. 

To be able to run TACTE for different apps like in our tests, we had to enable configuring the tutorial app via the command options.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests:

- 📜 Documentation:

### DIFF
--- a/web/tutorial-actions-executor/README.md
+++ b/web/tutorial-actions-executor/README.md
@@ -54,15 +54,14 @@ npm run edit-patch-action -- --wasp-cli-command wasp
 
 This command:
 
-- Generates the app from scratch using `generate-app` command.
+- Generates the app from scratch using generate-app command.
   - Generating the apps results in having each action as a separate commit in Git history.
-- "Rewinds" the app to the commit _before_ the specified action.
-- Applies changes from specified action to the working directory but does not commit them.
-- Allows you to edit the code as needed.
-- Asks you to confirm when you are done, and then creates a new patch for the action.
-- Reapplies all subsequent actions on top of the modified action.
-  - There might be conflicts between your new changes and subsequent actions.
-  - If so, you will need to resolve them manually.
+- "Rewinds" the app to the commit of the specified action and allows you to make changes
+  to its code as needed.
+- Asks you to confirm when you are done, updates the commit of the action with the changes
+  you did, and reapplies all subsequent commits on top of it.
+  - There might be conflicts between your new changes and subsequent commits.
+    If so, you will need to resolve them manually.
 
 ### 3. List Actions (`npm run list-actions`)
 


web/tutorial-actions-executor/e2e-tests/__snapshots__/generate-app.test.ts.snap
```diff
@@ -0,0 +1,79 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
```
infomiho, line 1: Vitest generated snapshot file. This is a "binary" file for our purposes, nothing to see here.

web/tutorial-actions-executor/e2e-tests/fixtures/tutorial/patches/02-patch__add-test-file.patch
```diff
@@ -0,0 +1,9 @@
+diff --git a/src/testUtils.ts b/src/testUtils.ts
```
infomiho, line 1: Example patch file for the mock tutorial.

web/tutorial-actions-executor/e2e-tests/fixtures/tutorial/01-init.md
```diff
@@ -0,0 +1,9 @@
+---
```
infomiho, line 1: Mock tutorial Markdown file with `<TutorialAction>` components.

web/tutorial-actions-executor/e2e-tests/generate-app.test.ts
```diff
@@ -0,0 +1,70 @@
...
+import { $, fs, glob } from "zx";
+
+const TEST_APP_NAME = "e2e-test-app";
+const TEST_OUTPUT_DIR = "./e2e-tests/.result";
+const TEST_TUTORIAL_DIR = "./e2e-tests/fixtures/tutorial";
+const WASP_CLI_COMMAND = process.env.WASP_CLI_COMMAND ?? "wasp-cli";
+
+describe("generate-app e2e", () => {
+  it(
+    "should generate app from tutorial actions and match snapshots",
```
infomiho, line 19: Test which runs the generator and takes note of:
- file list
- file contents of files we expect to be modified by patches
- git log output which shows all the commits the generator creates while working

web/tutorial-actions-executor/src/tutorialApp.ts
```diff
@@ -8,24 +8,35 @@ export type AppParentDirPath = Branded<string, "AppParentDirPath">;
...
+}
 
-export const tutorialApp = {
-  name: tutorialAppName,
-  parentDirPath: tutorialAppParentDirPath,
-  dirPath: tutorialAppDirPath,
-  docsTutorialDirPath,
-  docsTutorialPatchesPath,
-} as const;
+export function createTutorialApp(options: {
```
infomiho, line 19: Instead of hard-coding the locations of the tutorial files, patches and the app name - we now require the options to be specified when using the CLI.

web/tutorial-actions-executor/package.json
```diff
@@ -4,9 +4,10 @@
   "type": "module",
   "license": "MIT",
   "scripts": {
-    "edit-patch-action": "tsx ./src/index.ts edit-patch-action",
-    "generate-app": "tsx ./src/index.ts generate-app",
-    "list-actions": "tsx ./src/index.ts list-actions",
+    "tacte": "tsx ./src/index.ts",
+    "edit-patch-action": "npm run tacte -- edit-patch-action --app-name TodoApp --output-dir ./.result --tutorial-dir ../docs/tutorial",
+    "generate-app": "npm run tacte -- generate-app --app-name TodoApp --output-dir ./.result --tutorial-dir ../docs/tutorial",
+    "list-actions": "npm run tacte -- list-actions --app-name TodoApp --output-dir ./.result --tutorial-dir ../docs/tutorial",
```
infomiho: I find the repetitiveness of this annoying, but I couldn't come up with the right solution. 

Ideas considered:
1. Define one more script called `tacte-default` with all the default options which other scripts call (It throws and error since it requires `list-actions` to have one extra option i.e. `app-name` which is doesn't need)
2. Add support for a config file (Sounds maybe too complex)
3. Add defaults to each option (I don't want to hard-code the paths again)

cprecioso: My idea:
- You don't need to run through `tsx`, we're on Node 22 so we support ts natively
- This package exposes a program, so let's put it in the `package.json#bin`
- I actually like the `tacte-default` script thing, you can just add extra flags to it as needed.

So it would end up being something like:

```json
{
  "bin": { "tacte": "./src/index.ts" },
  "scripts": {
    "tacte-default": "tacte --common-option-1 --common-option-2",
    "edit-patch-action": "npm run tacte-default -- edit-patch-action --specific-option-1",
    "generate-app": "npm run tacte-default -- generate-app",
    "list-actions": "npm run tacte-default -- list-actions --specific-option-1 --specific-option-2"
  }
}
```

And I would entertain the idea of a config file for future work (it could be as simple as [c12](https://github.com/unjs/c12) + a zod schema)

cprecioso: Actually, thinking of encapsulating everything, probably these scripts should not be part of this package, but of the `web` package.

web/tutorial-actions-executor/README.md
```diff
@@ -104,3 +118,32 @@ The tool extracts these components and uses:
...
+
+### E2E Snapshot Tests
+
+E2E tests verify the entire tutorial action execution flow using snapshot testing.
+
+Snapshot testing captures the output of the tutorial action executor (generated files, git history, etc.)
+and stores it as a "snapshot". On subsequent test runs, the output is compared against the stored snapshot
+to ensure nothing has changed unexpectedly.
+
+**E2E Test Structure:**
```
cprecioso:
```suggestion
#### E2E Test Structure
```

web/tutorial-actions-executor/README.md
```diff
@@ -104,3 +118,32 @@ The tool extracts these components and uses:
...
+
+- Test fixtures: `e2e-tests/fixtures/tutorial/` contains minimal tutorial files used for testing.
+- Generated output: `e2e-tests/.result/`.
+- Snapshots: `e2e-tests/__snapshots__/`.
+
+**Updating Snapshots:**
+
+When you intentionally change the tutorial action executor behavior, delete the `__snapshots__` folder
+and re-run the tests to generate new snapshots.
```
cprecioso: No need to delete the snapshots

~~~suggestion
#### Updating Snapshots

When you intentionally change the tutorial action executor behavior, run the tests in update mode:

```bash
$ npm run test --update
```
~~~

web/tutorial-actions-executor/src/waspCli.ts
```diff
@@ -15,8 +15,11 @@ export async function waspDbMigrate({
   migrationName: string;
 }): Promise<void> {
   await $({
-    // Needs to inhert stdio for `wasp db migrate-dev` to work
-    stdio: "inherit",
+    // We ignore stdin to avoid hangs in non-interactive environments e.g. e2e tests when
+    // `wasp db migrate-dev` runs. When `stdin` is not ignored, the command waits forever
+    // for user input. We don't need any input since we are providing the migration name
+    // via the --name flag.
+    stdio: ["ignore", "pipe", "pipe"],
```
cprecioso, line 22: I think this might be the default behaviour

web/tutorial-actions-executor/src/tutorialApp.ts
cprecioso, line 19: A general suggestion, no need to do it now (or at all):

```ts
import * as z from "zod";
import path from "node:path";

const AbsPath =
  z.string()
  .overwrite((s) => path.resolve(s))
  .brand<"AbsPath">();

const AppName = z.string().brand<"AppName">();
const ParentDir = AbsPath.brand<"ParentDir">();
// etc

const TutorialApp = z.object({
  appName: AppName,
  parentDir: ParentDir,
  // etc
});

const createTutorialApp = (data: z.input<typeof TutorialApp>) =>
  TutorialApp.parse(data);
```

web/tutorial-actions-executor/src/commands/edit-patch-action/index.ts
```diff
@@ -57,7 +82,7 @@ export const editPatchActionCommand = new Command("edit-patch-action")
 
     await editPatchActionPatch({ appDir: tutorialApp.dirPath, action });
 
-    await extractCommitsIntoPatches(actions);
+    await extractCommitsIntoPatches(actions, tutorialApp);
```
cprecioso, line 78: In these times I long for a React Context for regular JS code (it's there in [`AsyncLocalStorage`](https://nodejs.org/docs/latest-v22.x/api/async_context.html) but it's not as easy to work with as React Context)

web/tutorial-actions-executor/e2e-tests/generate-app.test.ts
```diff
@@ -0,0 +1,70 @@
...
+        "--wasp-cli-command",
+        WASP_CLI_COMMAND,
+      ];
+      await $`npm run tacte -- generate-app ${generateAppOptions}`.verbose();
+
+      const projectDirPath = path.join(TEST_OUTPUT_DIR, TEST_APP_NAME);
+
+      function getProjectFileContent(filePath: string): string {
+        return fs.readFileSync(path.join(projectDirPath, filePath), "utf-8");
+      }
```
cprecioso: minor, style: i'd bump this up to the top level scope

web/tutorial-actions-executor/e2e-tests/generate-app.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    60 * 1000, // Test timeout
+  );
+});
+
+async function getProjectFileList(projectDirPath: string): Promise<string[]> {
+  const files = await glob("**/*", {
+    cwd: projectDirPath,
+    ignore: ["node_modules/**", ".wasp/**", ".git/**"],
+    onlyFiles: true,
+  });
```
cprecioso, line 62: nbd but maybe we can avoid using `zx` things when the built-in modules have what we need: [`fs.glob`](https://nodejs.org/docs/latest-v22.x/api/fs.html#fspromisesglobpattern-options). 

Same for your use of `fs.readFile` above

web/tutorial-actions-executor/e2e-tests/generate-app.test.ts
```diff
@@ -0,0 +1,70 @@
...
+      ];
+      await $`npm run tacte -- generate-app ${generateAppOptions}`.verbose();
+
+      const projectDirPath = path.join(TEST_OUTPUT_DIR, TEST_APP_NAME);
+
+      function getProjectFileContent(filePath: string): string {
+        return fs.readFileSync(path.join(projectDirPath, filePath), "utf-8");
+      }
+
+      const fileList = await getProjectFileList(projectDirPath);
```
cprecioso: strong: I agree with this `await` but then you should make `fs.readFile` async 😁

web/tutorial-actions-executor/e2e-tests/generate-app.test.ts
```diff
@@ -0,0 +1,70 @@
...
+
+describe("generate-app e2e", () => {
+  it(
+    "should generate app from tutorial actions and match snapshots",
+    async () => {
+      const generateAppOptions = [
+        "--app-name",
+        TEST_APP_NAME,
+        "--output-dir",
+        TEST_OUTPUT_DIR,
```
cprecioso, line 25: Should we clean up the `output-dir`? We don't want previous leftovers to linger and make it into the snapshots.

web/tutorial-actions-executor/e2e-tests/generate-app.test.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  return normalizeTimestamps(gitLogResult.text());
+}
+
+/**
+ * Normalizes timestamps in the given content by replacing them with a placeholder.
+ * Prisma migration files contain timestamps which change between test runs.
+ */
+function normalizeTimestamps(content: string): string {
+  return content.replace(/\d{14}/g, "TIMESTAMP");
```
cprecioso: nice!

web/tutorial-actions-executor/src/waspCli.ts
infomiho: image

Just checked, and it's  `pipe` for `stdin`

cprecioso: got it, the ignore this! 😅

web/tutorial-actions-executor/e2e-tests/__snapshots__/generate-app.test.ts.snap
Martinsos: Check how we marked snapshot test files as "generated" here, you will want to do the samething: https://github.com/wasp-lang/wasp/blob/main/waspc/.gitattributes#L11 .

web/tutorial-actions-executor/e2e-tests/fixtures/tutorial/patches/02-patch__add-test-file.patch
Martinsos: I don't understand this comment, but I do understand this is a patch file for mock tutorial.

web/tutorial-actions-executor/e2e-tests/generate-app.test.ts
Martinsos:
```suggestion
      function getProjectFileContent(pathInProject: string): string {
        return fs.readFileSync(path.join(projectDirPath, pathInProject), "utf-8");
      }
```

Martinsos: +1, since another one is also outside, and I don't nee dto know this implementation here yet.

web/tutorial-actions-executor/e2e-tests/generate-app.test.ts
```diff
@@ -0,0 +1,70 @@
...
+      expect(fileList).toMatchSnapshot("file-structure");
+
+      const testUtilsContent = getProjectFileContent("src/testUtils.ts");
+      expect(testUtilsContent).toMatchSnapshot("test-utils");
+
+      const schemaContent = getProjectFileContent("schema.prisma");
+      expect(schemaContent).toMatchSnapshot("prisma-schema");
+
+      const gitLog = await getGitLog(projectDirPath);
+      expect(gitLog).toMatchSnapshot("git-commits");
```
Martinsos: Small thing, but for readibility, I would reduce this to:
```suggestion
      expect(await getProjectFileList(projectDirPath)).toMatchSnapshot("file-structure");
      expect(getProjectFileContent("src/testUtils.ts")).toMatchSnapshot("test-utils");
      expect(getProjectFileContent("schema.prisma")).toMatchSnapshot("prisma-schema");
      expect(await getGitLog(projectDirPath)).toMatchSnapshot("git-commits");
```

Less repetition, less boilerplate for me to read, faster to scan.

web/tutorial-actions-executor/e2e-tests/generate-app.test.ts
Martinsos: But what if output we are testing by some weird accident actually starts producing TIMESTAMP instead of some of the timestamps, we won't pick that up :D.
Kidding, that is very unlikely hah. Although I wonder if there is a bit more "correct" way, but I can't thikn of any, excpet for maybe enriching the name to be "TIMESTAMP_STUB_E2E". It also helps if somebody is cheaking he file, to know what that TIMESTAMP there is -> aha, it is something we did in e2e tests.

web/tutorial-actions-executor/src/commands/edit-patch-action/index.ts
```diff
@@ -34,7 +43,23 @@ export const editPatchActionCommand = new Command("edit-patch-action")
     ),
   )
   .addOption(waspCliCommandOption)
-  .action(async ({ actionId, skipGeneratingApp, waspCliCommand }) => {
+  .addOption(appNameOption)
+  .addOption(outputDirOption)
+  .addOption(tutorialDirOption)
```
Martinsos: Sucks that you have to repeat for all the commands, no way to centralize this logic?

web/tutorial-actions-executor/src/commands/commonOptions.ts
```diff
@@ -8,3 +8,18 @@ export const waspCliCommandOption = new Option(
...
+
+export const appNameOption = new Option(
+  "--app-name <name>",
+  "Name of the app to generate",
+).makeOptionMandatory();
+
+export const outputDirOption = new Option(
+  "--output-dir <path>",
+  "Directory where the app will be generated",
+).makeOptionMandatory();
```
Martinsos: This should maybe have some default value? `out/`?

web/tutorial-actions-executor/src/tutorialApp.ts
```diff
@@ -8,24 +8,35 @@ export type AppParentDirPath = Branded<string, "AppParentDirPath">;
...
-  tutorialAppName,
-) as AppDirPath;
-const docsTutorialDirPath = path.resolve("../docs/tutorial") as TutorialDirPath;
-const docsTutorialPatchesPath = path.resolve(
-  docsTutorialDirPath,
-  "patches",
-) as PatchesDirPath;
+export interface TutorialApp {
+  name: AppName;
+  parentDirPath: AppParentDirPath;
```
Martinsos: Cant' we just get parent dir from dirPath? This is duplication that can potentially be incorrect, while we can easily extract that info, no?

web/tutorial-actions-executor/src/tutorialApp.ts
Martinsos: For my learning, how does this work, what is happening here?

web/tutorial-actions-executor/README.md
```diff
@@ -74,9 +74,23 @@ npm run list-actions
 
 Shows actions grouped by tutorial file, including each action's `id` and `kind`.
 
+### Required Command Options
+
+All commands require the following options to set up the tutorial app configuration:
+
+- `--app-name <name>`: Name of the app to generate.
+- `--output-dir <path>`: Directory where the app will be generated.
+- `--tutorial-dir <path>`: Directory containing the tutorial tutorial files.
```
Martinsos:
```suggestion
- `--tutorial-dir <path>`: Directory containing the tutorial files.
```

web/tutorial-actions-executor/README.md
```diff
@@ -104,3 +118,32 @@ The tool extracts these components and uses:
...
+
+```bash
+npm run test
+```
+
+### E2E Snapshot Tests
+
+E2E tests verify the entire tutorial action execution flow using snapshot testing.
+
+Snapshot testing captures the output of the tutorial action executor (generated files, git history, etc.)
```
Martinsos, line 135: Do we want to add running of these tests to the CI? They could run only if tacte files changes. I guess we will move tacte to its own repo soon though, so maybe we can skip it.

web/tutorial-actions-executor/e2e-tests/generate-app.test.ts
```diff
@@ -0,0 +1,70 @@
+import path from "path";
+import { describe, expect, it } from "vitest";
+import { $, fs, glob } from "zx";
+
+const TEST_APP_NAME = "e2e-test-app";
+const TEST_OUTPUT_DIR = "./e2e-tests/.result";
+const TEST_TUTORIAL_DIR = "./e2e-tests/fixtures/tutorial";
```
Martinsos: Would using `__dirname` here make us robust to change in the directory structure here? If so, let's do it.

web/tutorial-actions-executor/e2e-tests/fixtures/tutorial/patches/02-patch__add-test-file.patch
infomiho: Now that I read it myself, I wrote it weirdly, I wanted to say "this is a example patch file used in the mock tutorial"

web/tutorial-actions-executor/src/commands/edit-patch-action/index.ts
infomiho: Actually, there is - we can create a sub-class of `Command` with the options included. I disliked that way of doing it, but I see the upside. Let me investigate a bit more.

web/tutorial-actions-executor/src/commands/commonOptions.ts
infomiho: This would make sense, output dir can have a default value that would be okay for different apps.

web/tutorial-actions-executor/src/tutorialApp.ts
infomiho: We have the parent dir where we run `wasp new` which created a new folder which then becomes our app dir. So I encoded those two paths explicitly. 

I'd like to keep it like that, we compute one using the other anyways:
```ts
const parentDirPath = path.resolve(options.outputDir) as AppParentDirPath;
const dirPath = path.resolve(parentDirPath, appName) as AppDirPath;
```

I guess we could use some better names here e.g. `outputDir` and `waspAppDir`. But as we make it less Wasp-specific, we won't have the `waspAppDir` baked in anymore.

web/tutorial-actions-executor/README.md
infomiho: We have an issue for this https://github.com/wasp-lang/wasp/issues/3247

web/tutorial-actions-executor/src/tutorialApp.ts
cprecioso: - Miho is using branded types everywhere for this library

- Branded types are a soft version of `newtype`s for TS, trying to scurry out of structural typing and into nominal typing. e.g.:

  ```ts
  type FilePath = string & { __brand: "FilePath" }
  //                         ⬆️ this property doesn't really exist
  // we're lying to the typechecker so we can force nominal-ish typing
  // the `__brand` property name is a convention but we could choose anything

  const ensurePath = (str: string)=> path.isPath(str) ? str as FilePath : null

  const needsPath = (path: FilePath) => { /* ... */ }

  needsPath("~/myfile.txt")             // -> type error, missing `__brand` property
  needsPath(ensurePath("~/myfile.txt")) // -> type correct
  ```

- Zod can handle branding types for you, especially useful if you pair it with specific validation or transformation

  ```ts
  const FilePathSchema = z.string().check(p => path.isPath(p)).brand<"FilePath">()
  const AbsPathSchema = FilePathSchema.transform(p => path.resolve(p)).brand<"AbsPath">()

  AbsPathSchema.parse("./my-path.txt")
  // input string
  // output AbsPath (guaranteed to be absolute)
  ```

- By composing the Zod schemata, we can avoid parsing each string individually, and doing it all at once in the full object.

  ```ts
  // instead of:
  const foo = AbsPathSchema.parse("./foo")
  const bar = AbsPathSchema.parse("./bar")
  const config = { foo, bar }

  // we can do:
  const ConfigSchema = z.object({ foo: AbsPathSchema, bar: AbsPathSchema })
  const config = ConfigSchema.parse({ foo: "./foo", bar: "./bar" })
  ```

- And the last trick up our sleeve, `schema.parse()` takes an `unknown` for convenience, so we ask Zod for the input type so we can have safer hardcoded types

  ```ts
  type ConfigInput = z.input<typeof ConfigSchema> // -> { foo: string, bar: string }
  // Zod knows it needs at least an input of that shape so that it doesn't immediately fail

  // and as a convenience for hardcoded values, we just wrap the parse with this type
  const parseConfig = (input: ConfigInput): Config => ConfigSchema.parse(input)

  ConfigSchema.parse(1234) // <- no type error, but it would never work
  ConfigSchema.parse({ foo: "./foo", bar: "./bar" }) // <- correct, but no autocomplete

  parseConfig(1234) // <- type error
  parseConfig({ foo: "./foo" }) // <- type error, missing property
  parseConfig({ foo: "./foo", bar: "./bar" }) // <- autocomplete available too!
  ```

web/tutorial-actions-executor/e2e-tests/generate-app.test.ts
cprecioso: > Would using `__dirname` here make us robust to change in the directory structure here?

No, why would it?

web/tutorial-actions-executor/e2e-tests/generate-app.test.ts
infomiho: Using the built-in glob is so much more verbose, check it out:
```ts
async function getProjectFileList(projectDirPath: string): Promise<string[]> {
  const files: string[] = [];

  const filesIterator = fs.glob("**/*", {
    cwd: projectDirPath,
    exclude: (name) => {
      const parts = name.split(path.sep);
      return (
        parts.includes("node_modules") ||
        parts.includes(".wasp") ||
        parts.includes(".git")
      );
    },
  });

  for await (const file of filesIterator) {
    const fullPath = path.join(projectDirPath, file);
    const stats = await fs.stat(fullPath);
    if (!stats.isFile()) {
      continue;
    }
    files.push(file);
  }

  return files.map(normalizeTimestamps).sort();
}
```

`zx` exports the `[globby`](https://github.com/sindresorhus/globby) package and since `zx` is a hard-dep for us - I don't see the harm in using.

I've changed the code to use `fs` from `fs/promises` 👍

web/tutorial-actions-executor/e2e-tests/generate-app.test.ts
infomiho: > But what if output we are testing by some weird accident actually starts producing TIMESTAMP instead of some of the timestamps, we won't pick that up :D.

That's fine - there will be TIMSTAMPS in different locations in the output then.

I get the argument of someone looking at snapshot files, so I'll update the replacement to `NORMALIZED_TIMESTAMP_E2E`.

web/tutorial-actions-executor/e2e-tests/generate-app.test.ts
cprecioso: Oh ok, I was looking at `glob` from the latest node, which does have `exclude` as globs, and `withFileTypes` option.
Go on then

web/tutorial-actions-executor/src/tutorialApp.ts
Martinsos: Nice, thank you! I think I am still missing a bit of TS type-level intuition + zod specifics to understand this 100%, but I do feel know I understand enough to know where to look further / understand the situation, which helps.

web/tutorial-actions-executor/e2e-tests/generate-app.test.ts
Martinsos: Ah no but you don't want TIMESTAMP to be in there, that is weird, but you will never know :D. Nevermind, it is silly argument, not someting we need to crea bout in practice, resolving.

web/tutorial-actions-executor/src/commands/edit-patch-action/index.ts
Martinsos: Ok, up to you, whatever seems the best, it is not a big thing.

web/tutorial-actions-executor/src/tutorialApp.ts
Martinsos: Yeah I guess what I was proposing is that you never have these two, instead, when you need parent, you calculate it at that very moment. This way you ensure that when you get this in some function, you don't have to wonder if someobdy up there messed it up. It reduces space for mistakes. And calculating it on the fly is cheap. This is general principle of denormalizing data I guess, or whatever is the correct term. Why are you actually storing it, because you want to avoid repeating calculation? But is it really so expensive that that is needed? Because what you are doing makes sense to me only if you are doing optimization.

web/tutorial-actions-executor/e2e-tests/generate-app.test.ts
Martinsos: Because right now, it has hardcoded `e2e-tests` here. Which is an assumption that goes "I assume this file is positioned in e2e-tests".
But if that gets renamed to e2e-test, or there is an added level of depth to dirs, this will become incorrect. If it used `__dirname`, it would still be correct, since there is no assumption.

web/tutorial-actions-executor/src/commands/edit-patch-action/index.ts
infomiho: I've managed to do this in the end using a factory fn - I think you'll like it:
```ts
export const editPatchActionCommand = createTacteCommand("edit-patch-action")
```

web/tutorial-actions-executor/e2e-tests/generate-app.test.ts
infomiho: Sure, I've added the current dir path computation 👍

web/tutorial-actions-executor/src/tutorialApp.ts
infomiho: Hm, yep, now that you said it and I've renamed them to `outputDir` and `appDirPath` I like having them both since they serve different purposes:
- there is an output dir which is used for any actions we might perform
- there is an app dir path which we use any app actions e.g. `wasp db migrate` (which is Wasp specific)

When we move away from Wasp specific command e.g. just use patches and commands - I'd then also probably kill the `appDirPath` since each command will need to `cd` into the app dir if needed because the tool won't be Wasp specific anymore.

web/tutorial-actions-executor/e2e-tests/generate-app.test.ts
cprecioso: @Martinsos oh I get it! Sorry I didn't realize this was in the `e2e-tests` folder itself. You are right!

web/tutorial-actions-executor/e2e-tests/generate-app.test.ts
```diff
@@ -1,14 +1,20 @@
 import * as fs from "node:fs/promises";
 import * as path from "node:path";
-import { describe, expect, it } from "vitest";
+import { beforeAll, describe, expect, it } from "vitest";
 import { $, glob } from "zx";
 
+const currentDirPath = path.dirname(new URL(import.meta.url).pathname);
```
cprecioso: You're looking for `import.meta.dirname` :)

web/tutorial-actions-executor/package.json
infomiho: - Dropping `tsx` would require rewriting all the imports to match what Node.js can resolve e.g. `index.js` instead of dir imports.
- Doing `tacte-default` breaks for me now...
	```
	npm run generate-app
	
	> tutorial-app-generator@0.0.1 generate-app
	> npm run tacte-default -- generate-app
	
	
	> tutorial-app-generator@0.0.1 tacte-default
	> tsx ./src/index.ts --app-name TodoApp --tutorial-dir ../docs/tutorial generate-app
	
	error: unknown option '--app-name'
	```
	with this kind of setup:
	```json
	    "generate-app": "npm run tacte-default -- generate-app",
	    "tacte-default": "tsx ./src/index.ts --app-name TodoApp --tutorial-dir ../docs/tutorial",
	```

I think I'll optimize this bit (Wasp defaults) and having the `bin` option in a separate PR 👍

PR #3317, opened 2025-11-06
Remove LLVM workaround

cprecioso:
We had to setup an old LLVM in Apple Silicon with a workaround, since it was an old version. This is no longer needed, we can remove it.


Martinsos (approved):
Nice!

Issue #3148, opened 2025-09-11
Get core `wasp` CLI functionality working on Windows


Martinsos:
## 🎯 Goal

- [ ] Be able to run `wasp` CLI on Windows (native, not WSL) with all the core functionality working. I mean `wasp` CLI version built during wasp development, not installed one (binary).
- [ ] This also includes tests running in CI to ensure no regressions happen. This might be challenging due to having to run test infrastructure on Win? Depends on the tests I guess, let's see.

Core functionality is `wasp start`, `wasp build`, `wasp db migrate-dev`, `wasp waspls`, `wasp clean`, ... .

Non-core functionality are commands that one can do without: `wasp deploy` (because you can deploy manually), `wasp start db` (because you can start the db manually), ... . These we will handle separately, in another issue.

## 🗒️ Currently known core features that are not working:

- [ ] Calling `npm` or `npx` from the `wasp` CLI: can't find `npm`.
      Latest ongoing attempt at fixing that: #3258 (inspired by attempts #658 and #3020).
- [ ] Encoding issues (potentially not just Win issue though).
      Solved in #658 ? If so, we should extract that part and create PR.
- [ ] Some wrong file paths.
      Solved in #658 ? If so, we should extract that part and create PR.
- [ ] Prisma migrate not working on Win (due to `script`).
      Solved in #658 but also got solved on its own in the meantime? If so, confirm and mark as already done.

## What works
After #3258 gets merged, we will be able to run local dev version of `wasp` using `cabal run wasp-cli` and execute some basic commands.
Notice however that there is more to be "built" in order to run fully working dev version of `wasp`, like setting up the TS packages/libs that come with it. So just running `cabal run wasp-cli` will crash again at points where those are needed, if they are not set up.
Since we don't have the equivalent of `./run` script for Windows, right now that has to be done manually.
- #3318 adds script for taking care of the TS packages, which is a crucial component for making this easier.

Martinsos:
There is an old PR where a big part of the issues with Windows is fixed: https://github.com/wasp-lang/wasp/pull/658 . Specifically:
1. Calling of `npm` and `npx` and similar works (logic was added to invoke them differently when on Win, taking into account they are .ps1 scripts and .cmd and so on).
2. Wasp CLI was made more robust regarding encoding (this is a general thing, actually).
3. Some wrong paths were fixed.
4. We made prisma migrate work on Win (`script` was problem). I think this got solved on `main` also in the meantime because `script` is not needed anymore.
[...]

PR #3318, opened 2025-11-06
install_packages_to_data_dir.ps1 for Window 

### LINKED ISSUE #3043: Support npm run build for Wasp Studio on Windows
## This is an issue for the PR  https://github.com/wasp-lang/wasp/pull/3020

**Describe the bug**

Cannot build Wasp studio on Windows due to the Unix commands in package.json

**To Reproduce**

- Install [NodeJS](https://nodejs.org/en/download)
- Clone commit bba407d
- Change directory to ./waspc/packages/studio/
- Install dependencies
```bash
npm ci
```
- Run build
```bash
npm run build
```
- The console will show the error:
```bash
> wasp-studio-server@1.0.0 build
> npm run build:client && rm -rf dist && tsc && cp -r ./public ./dist/public

> wasp-studio-server@1.0.0 build:client
> npm --prefix ./client install && npm --prefix ./client run copy

> wasp-studio-client@0.0.0 copy
> npm run build && cp -r ./dist/* ../public

> wasp-studio-client@0.0.0 build
> tsc && vite build

vite v4.5.3 building for production...
Browserslist: caniuse-lite is outdated. Please run:
  npx update-browserslist-db@latest
  Why you should do it regularly: https://github.com/browserslist/update-db#readme
✓ 1057 modules transformed.
dist/index.html                     0.40 kB │ gzip:   0.27 kB
dist/assets/index-17ce6ed4.css    157.34 kB │ gzip:  18.98 kB
dist/assets/index-62a9d21a.js     496.83 kB │ gzip: 162.59 kB
dist/assets/Flow-b5112d3d.js    1,373.26 kB │ gzip: 421.00 kB

(!) Some chunks are larger than 500 kBs after minification. Consider:
- Using dynamic import() to code-split the application
- Use build.rollupOptions.output.manualChunks to improve chunking: https://rollupjs.org/configuration-options/#output-manualchunks
- Adjust chunk size limit for this warning via build.chunkSizeWarningLimit.
✓ built in 13.36s

'cp' is not recognized as an internal or external command,
operable program or batch file.
```

**Problem**

`cp -r` and `rm -rf` are not valid commands on Windows

**Environment:**

- OS: Windows 11 (Build 10.0.26100)
- Node v22.14.0
- NPM 11.4.2
- Powershell 5.1.26100.4768
- CMD Version 10.0.26100.4770

PS. I've already fixed this issue

nodtem66:
# PR for #3148 Wasp CLI on Windows

## Description

We used the script `waspc/tools/install_packages_to_data_dir.sh` for preparing dependent packages. This PR extends this script for Window users. The logic derived from the original bash script. 

## Related issues and fixes

The current `packages/studio` uses `cp` and `rmdir` command to build the project. But these are not the valid commands in Window. To make this cross-platform, I replaced these commands by NodeJS's [`fs.cpSync`](https://nodejs.org/api/fs.html#fscpsyncsrc-dest-options) and [`fs.rmSync`](https://nodejs.org/api/fs.html#fsrmsyncpath-options). This PR will fix #3043 as well.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests:

- 📜 Documentation:

### DIFF
--- a/waspc/tools/install_packages_to_data_dir.ps1
+++ b/waspc/tools/install_packages_to_data_dir.ps1
@@ -29,7 +29,9 @@ foreach ($package in $packageDirs) {
 
 Push-Location "$dir/.."
 try {
-    Remove-Item -Path "./data/packages" -Recurse -Force
+    if (Test-Path "./data/packages") {
+        Remove-Item -Path "./data/packages" -Recurse -Force
+    }
     Copy-Item -Path "./packages" -Destination "./data/packages" -Recurse
 }
 finally {
--- a/waspc/packages/studio/package.json
+++ b/waspc/packages/studio/package.json
@@ -11,8 +11,8 @@
     "build": "npm run build:client && npm run remove:dist && tsc && npm run copy:public",
     "build:client": "npm --prefix ./client install && npm --prefix ./client run copy",
     "copy:public": "node -e \"require('fs').cpSync('./public', './dist/public', { recursive: true });\"",
-    "remove:dist": "node -e \"require('fs').rmSync('./dist', { recursive: true, force: true });\"",
     "dev": "tsx --watch ./src/index.ts",
+    "remove:dist": "node -e \"require('fs').rmSync('./dist', { recursive: true, force: true });\"",
     "start": "node ./dist/index.js"
   },
   "dependencies": {


waspc/tools/install_packages_to_data_dir.bat
```diff
@@ -0,0 +1,49 @@
+@echo off
+setlocal enableDelayedExpansion 
+
+REM Window BAT script to compile the waspc/packages/* packages locally and in CI.
+REM It will then move it into the Cabal data dir (and thus, the installer archive in CI releases).
+REM This script derived from tools/install_packages_to_data_dir.sh using Copilot (6 Nov. 2025).
+REM Reviewed, revised, and commented by @nodtem66 (6 Nov. 2025)
```
Martinsos: This is good to know! But we can remove it, it will become stale with time.

waspc/tools/install_packages_to_data_dir.bat
```diff
@@ -0,0 +1,49 @@
...
+
+REM Gets the directory of where this script lives.
+set dir=%~dp0
+REM Remove trailing backslash
```
Martinsos:
```suggestion
REM Remove trailing backslash.
```

waspc/tools/install_packages_to_data_dir.bat
```diff
@@ -0,0 +1,49 @@
...
+if "%dir:~-1%"=="\" set "dir=%dir:~0,-1%"
+
+REM set alias cmd for cp
```
Martinsos:
```suggestion
REM Set alias for cp.
```

waspc/tools/install_packages_to_data_dir.bat
```diff
@@ -0,0 +1,49 @@
...
+DOSKEY cp=COPY $* 
+
+REM Loop through each package directory
+REM For /d Loop to iterate over directories
```
Martinsos:
```suggestion
```
I think these are redundant, we can remove them.

waspc/tools/install_packages_to_data_dir.bat
```diff
@@ -0,0 +1,49 @@
...
+    cd /d "!package_dir!"
+    
+    REM We're only installing the dependencies here to verify that the build
+    REM works, that's why the node_modules folder is removed immediately after.
+    REM The real dependency installation happens in Haskell.
+    call npm install
+    call npm run build
+    
+    REM Remove node_modules directory
+    REM rmdir /s delete a directory tree, /q for quiet mode
```
Martinsos:
```suggestion
```
I would even delete these, to keep it as similar to the original script, the first comment is certainly redundant, the second is somewhat informative but also we can go without it and check online if needed.

waspc/tools/install_packages_to_data_dir.bat
```diff
@@ -0,0 +1,49 @@
...
+for /d %%G in ("%dir%\..\packages\*") do (
+    set package_dir=%%G
+    set package=%%~nxG
+
+    echo Installing !package! ^(!package_dir!^)
+    cd /d "!package_dir!"
+    
+    REM We're only installing the dependencies here to verify that the build
+    REM works, that's why the node_modules folder is removed immediately after.
+    REM The real dependency installation happens in Haskell.
```
Martinsos: Maybe let's try to replicate the original script as close as we can and put this above the `echo Installing ...`, same like in the original script.

waspc/tools/install_packages_to_data_dir.bat
```diff
@@ -0,0 +1,49 @@
...
+    call npm install
+    call npm run build
+    
+    REM Remove node_modules directory
+    REM rmdir /s delete a directory tree, /q for quiet mode
+    if exist "node_modules" rmdir /s /q "node_modules"
+)
+
+REM Change back to parent directory
```
Martinsos:
```suggestion
```
Redundant comment.

waspc/tools/install_packages_to_data_dir.bat
```diff
@@ -0,0 +1,49 @@
...
+    set package=%%~nxG
+
+    echo Installing !package! ^(!package_dir!^)
+    cd /d "!package_dir!"
+    
+    REM We're only installing the dependencies here to verify that the build
+    REM works, that's why the node_modules folder is removed immediately after.
+    REM The real dependency installation happens in Haskell.
+    call npm install
+    call npm run build
```
Martinsos: Maybe let's try to replicate the original script as close as we can and put this above the `echo Installing ...`, same like in the original script.

```suggestion
    REM We're only installing the dependencies here to verify that the build
    REM works, that's why the node_modules folder is removed immediately after.
    REM The real dependency installation happens in Haskell.
    echo Installing !package! ^(!package_dir!^)
    cd /d "!package_dir!"
    call npm install
    call npm run build
```

waspc/tools/install_packages_to_data_dir.bat
```diff
@@ -0,0 +1,49 @@
...
+    REM Remove node_modules directory
+    REM rmdir /s delete a directory tree, /q for quiet mode
+    if exist "node_modules" rmdir /s /q "node_modules"
+)
+
+REM Change back to parent directory
+cd /d "%dir%\.."
+
+REM Remove existing data/packages directory and copy packages to data
+if exist "data\packages" rmdir /s /q "data\packages"
```
Martinsos: This check is in theory redundant, as data/packages should exist. I would maybe remove it, because if it doesn't exist, something is wrong, so I would rather have it fail.
```suggestion
rmdir /s /q "data\packages"
```

waspc/tools/install_packages_to_data_dir.bat
```diff
@@ -0,0 +1,49 @@
...
+    
+    REM Remove node_modules directory
+    REM rmdir /s delete a directory tree, /q for quiet mode
+    if exist "node_modules" rmdir /s /q "node_modules"
+)
+
+REM Change back to parent directory
+cd /d "%dir%\.."
+
+REM Remove existing data/packages directory and copy packages to data
```
Martinsos: Redundant comment.
```suggestion
```

waspc/tools/install_packages_to_data_dir.bat
```diff
@@ -0,0 +1,49 @@
...
+if exist "data\packages" rmdir /s /q "data\packages"
+REM xcopy /e copy directories and subdirectories, including empty ones
+REM /i if in doubt always assume the destination is a folder
+REM /y suppresses prompting to confirm you want to overwrite an existing destination file
+REM /q suppresses the display of file names while copying
+REM /c continues copying even if errors occur
```
Martinsos:
```suggestion
REM Explanation of xcopy arguments:
REM  - /e -> Copy directories and subdirectories, including empty ones.
REM  - /i -> If in doubt always assume the destination is a folder.
REM  - /y -> Suppresses prompting to confirm you want to overwrite an existing destination file.
REM  - /q -> Suppresses the display of file names while copying.
REM  - /c -> Continues copying even if errors occur.
```

Martinsos: Btw I am a bit suspicious about the `/c` option. If there are errors, maybe we should rather have the whole thing fail, instead of it continuing? Nothing worse than thinking stuff worked, but it didn't.

waspc/tools/install_packages_to_data_dir.bat
```diff
@@ -0,0 +1,49 @@
...
+xcopy "packages" "data\packages" /e /i /y /q /c
+
+echo Done^!
```
Martinsos:
```suggestion
echo Done installing packages to data dir^!
```
if this is valid syntax, not sure.

Martinsos:
Oh one extra thing: check out `.github/workflows/ci-waspc-test.yaml`. It also runs on Windows (latest). For windows it skips some steps, like running e2e tests, but it does compile the project and run unit tests.

Anyway, there is part where `install_packages_to_data_dir.sh` is called. But it runs only for ubuntu/mac. Now however we can also add the same thing for Windows! Let's add it? It won't make a big difference until we also enable some other parts of that workflow for Windows, but it is a good step in the right direction.

waspc/tools/install_packages_to_data_dir.bat
nodtem66: That makes sense. I'll remove /c from it.

nodtem66:
@Martinsos Thank you for the review. I've revised the batch script and added it to the `.github/workflows/ci-waspc-test.yaml`.

Regarding the `./run.bat`, I'll raise a new PR for it. Indeed, It is a good idea to keep it simple. 
I'll implement just:
- run build
- run build:all
- run wasp-cli

PS. In term of maintainability, I prefers powershell over batch. So, I'll create `run.ps1` instead.

Martinsos:
> @Martinsos Thank you for the review. I've revised the batch script and added it to the `.github/workflows/ci-waspc-test.yaml`.
> 
> Regarding the `./run.bat`, I'll raise a new PR for it. Indeed, It is a good idea to keep it simple. I'll implement just:
> 
> * run build
> * run build:all
> * run wasp-cli
> 
> PS. In term of maintainability, I prefers powershell over batch. So, I'll create `run.ps1` instead.

Sounds good! Also, let me know when this PR is ready for another round. Btw, is there a reason why you then implement `install_packages_to_data_dir` in batch, and not in powershell? I guess because it is more similar to the .sh file it is emulating?

nodtem66:
Okay. It's ready for reviewing.

It was my blunder to not think that powershell is much better and a default shell for Windows nowadays. You can see in the new commit that the powershell script looks more clear than the batch script.

.github/workflows/ci-waspc-test.yaml
```diff
@@ -84,6 +84,10 @@ jobs:
         if: matrix.os == 'ubuntu-22.04' || matrix.os == 'macos-13'
         run: ./tools/install_packages_to_data_dir.sh
 
+      - name: Compile TS packages and move it into the Cabal data dir on Windows
+        if: matrix.os == 'windows-latest'
+        run: ./tools/install_packages_to_data_dir.bat
```
Martinsos:
```suggestion
        run: ./tools/install_packages_to_data_dir.ps1
```

waspc/tools/install_packages_to_data_dir.ps1
```diff
@@ -0,0 +1,42 @@
+# Helper to compile the waspc/packages/* packages locally and in CI.
+# It will then move it into the Cabal data dir (and thus, the installer archive in CI releases).
+# PSVersion 5.1 or higher is required.
+
+# Gets the directory of where this script lives.
+$dir = Split-Path -Parent $MyInvocation.MyCommand.Path
+
+# Get all package directories.
```
Martinsos:
```suggestion
```

Redundant comment. We are sticking to Clean Code to some degree (not fanatically but there is some good stuff), and one thing we do is avoid comments that can be expressed via code or types, and also avoid comments that are stating the same thing as code they describe (as it is duplication and risks desyncing as time goes).

In this case, I think this code line is so clear, this comment is redundant.

waspc/tools/install_packages_to_data_dir.ps1
```diff
@@ -0,0 +1,42 @@
...
+    }
+    finally {
+        Pop-Location
+    }
+}
+
+# Change to parent directory and copy packages to data
+Push-Location "$dir/.."
+try {
+    # Remove existing data/packages directory
```
Martinsos:
```suggestion
```
Redundant.

waspc/tools/install_packages_to_data_dir.ps1
```diff
@@ -0,0 +1,42 @@
...
+    Remove-Item -Path "./data/packages" -Recurse -Force
+    # Copy packages to data directory
```
Martinsos:
```suggestion
```
Redundant.

waspc/tools/install_packages_to_data_dir.ps1
```diff
@@ -0,0 +1,42 @@
...
+        npm run build
+        # Remove node_modules directory
+        Remove-Item -Path "./node_modules" -Recurse -Force
+    }
+    finally {
+        Pop-Location
+    }
+}
+
+# Change to parent directory and copy packages to data
```
Martinsos:
```suggestion
```
Redundant.

Martinsos (commented):
Ok I think this is it!

Can you please fix the comments, and I think we can merge. Oh one more thing: why don't you make it point toward `main` branch? You don't really need any of the work I did in `wasp-lang:martin-win-cmd-resolution`, and this merged in `main` won't do any problems, if I am correct.

nodtem66:
Thanks for the code review. The new commit is ready to review. 
I totally forgot to change the script name in `ci-waspc-test.yaml`. 

My intention is using your branch as a baseline to check errors and fix them as much as possible before merging into `main`.
But I can change to `main` if you would like.

Martinsos:
> Thanks for the code review. The new commit is ready to review. I totally forgot to change the script name in `ci-waspc-test.yaml`.
> 
> My intention is using your branch as a baseline to check errors and fix them as much as possible before merging into `main`. But I can change to `main` if you would like.

I assumed so, and it is a good idea, but I would rather merge these chunks we know work sooner, and not accumulate them on my branch, since my branch still needs some refining. And since these chunks are quite standalone, we can indeed merge them sooner. So let's make this go agains `main`, and we can merge!

nodtem66:
It's ready. Please check it again in case I missed something.

waspc/tools/install_packages_to_data_dir.ps1
```diff
@@ -0,0 +1,38 @@
...
+    # We're only installing the dependencies here to verify that the build
+    # works, that's why the node_modules folder is removed immediately after.
+    # The real dependency installation happens in Haskell.
+    Write-Host "Installing $packageName ($packageDir)"
+    
+    Push-Location $packageDir
+    try {
+        npm install
+        npm run build
+        # Remove node_modules directory
```
Martinsos:
```suggestion
```
Redunant.

waspc/tools/install_packages_to_data_dir.ps1
```diff
@@ -0,0 +1,38 @@
+# Helper to compile the waspc/packages/* packages locally and in CI.
+# It will then move it into the Cabal data dir (and thus, the installer archive in CI releases).
```
Martinsos:
```suggestion
# It will then move them into the Cabal data dir (and thus, the installer archive in CI releases).
```

.github/workflows/ci-waspc-test.yaml
```diff
@@ -77,6 +77,10 @@ jobs:
         if: matrix.os == 'ubuntu-22.04' || matrix.os == 'macos-15-intel'
         run: ./tools/install_packages_to_data_dir.sh
 
+      - name: Compile TS packages and move it into the Cabal data dir on Windows
```
Martinsos:
```suggestion
      - name: Compile TS packages and move them into the Cabal data dir on Windows
```

Martinsos:
Hm CI fails on multiple places.

Most obvious to me was this, so let's maybe start with that:

image


frmom here https://github.com/wasp-lang/wasp/actions/runs/19207250164/job/54904233053 .

It is running on windows-latest which is Win Server 2025, might that be the issue?

nodtem66:
The script was run on the bash shell.
image
I'll fix this by changing the shell to `pwsh`

Martinsos:
Great @nodtem66 nice catch, it makes sense now but I wasn't aware that shell needs to be specified if not `bash`! But yeah how would it know otherwise. Merged!

PR #3323, opened 2025-11-07
Fix `wasp start db` to use Postgres 18 explicitly

### LINKED ISSUE #3321: `wasp db start` fails with latest Postgres 18 version
### Reproduction 
Wasp version: **0.18.1** (latest at the time of writing)

Reproduction steps:
```bash
# Force Docker to fetch PG 18
docker image rm postgres:latest

# Create a new app
cd /tmp
wasp new saas-app -t saas
cd saas-app/app
wasp db start
```

The error we get is:
```
wasp db start

Additional info:
     postgresql://postgresWaspDevUser:postgresWaspDevPass@localhost:5432/OpenSaaS-299c73562d
     wasp-dev-db-OpenSaaS-299c73562d

...
Unable to find image 'postgres:latest' locally
latest: Pulling from library/postgres
Digest: sha256:41fc5342eefba6cc2ccda736aaf034bbbb7c3df0fdb81516eba1ba33f360162c
Status: Downloaded newer image for postgres:latest
Error: in 18+, these Docker images are configured to store database data in a
       format which is compatible with "pg_ctlcluster" (specifically, using
       major-version-specific directory names).  This better reflects how
       PostgreSQL itself works, and how upgrades are to be performed.

       See also https://github.com/docker-library/postgres/pull/1259

       Counter to that, there appears to be PostgreSQL data in:
         /var/lib/postgresql/data (unused mount/volume)

       This is usually the result of upgrading the Docker image without
       upgrading the underlying database using "pg_upgrade" (which requires both
       versions).

       The suggested container configuration for 18+ is to place a single mount
       at /var/lib/postgresql which will then place PostgreSQL data in a
       subdirectory, allowing usage of "pg_upgrade --link" without mount point
       boundary issues.

       See https://github.com/docker-library/postgres/issues/37 for a (long)
       discussion around this process, and suggestions for how to do so.
wasp-bin: callCommand: docker run --name wasp-dev-db-OpenSaaS-299c73562d --rm --publish 5432:5432 -v wasp-dev-db-OpenSaaS-299c73562d:/var/lib/postgresql/data --env POSTGRES_PASSWORD=postgresWaspDevPass --env POSTGRES_USER=postgresWaspDevUser --env POSTGRES_DB=OpenSaaS-299c73562d postgres (exit 1): failed
```

### What's going on

From Postgres 18, Docker images now require us to use ` /var/lib/postgresql` as the mounting dir for the volume and not ` /var/lib/postgresql/data` as we did before. 

Erasing the existing volumes used by the development database didn't help. 

### Solution

We need to change the container mounting location to ` /var/lib/postgresql`. 

Running `docker run --name wasp-dev-db-OpenSaaS-299c73562d --rm --publish 5432:5432 -v wasp-dev-db-OpenSaaS-299c73562d:/var/lib/postgresql --env POSTGRES_PASSWORD=postgresWaspDevPass --env POSTGRES_USER=postgresWaspDevUser --env POSTGRES_DB=OpenSaaS-299c73562d postgres` worked for me locally.

## Workarounds for users

This affects all Wasp versions since all Wasp version use `postgres:latest` for the development DB. The fix will be implement in the future Wasp 0.18.2 but for older versions of Wasp this issue will persist. So, we are offering a couple of workarounds:

### Tag `postgres:latest` as `postgres:17`

- Run `docker tag postgres:17 postgres:latest` which will make Docker use PG 17 instead of PG 18
- If you get a error about an existing database, you probably just need to delete the dev DB volume with `docker volume rm wasp-dev-db-<specific-to-your-app>` (you'll find the volume name in the error message)

### Use Docker directly for your dev needs 

You can run the Docker command yourself with a specific Postgres version:
```
docker run --name wasp-dev-db-custom --rm --publish 5432:5432 -v wasp-dev-db-custom:/var/lib/postgresql/data --env POSTGRES_PASSWORD=postgresWaspDevPass --env POSTGRES_USER=postgresWaspDevUser --env POSTGRES_DB=custom postgres:17
```
And you'll need to set the `DATABASE_URL` in the `.env.server`:
```
postgresql://postgresWaspDevUser:postgresWaspDevPass@localhost:5432/custom
```

I've used `custom` here as the app name - you can change this.

infomiho:
## Description

Fixes #3321

It does two things:
- pins the Postgres version used by `wasp start db` to version 18
- uses the correct path for the Docker volume mount for Postgres 18

## Type of change

- [x] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests:

  - [ ] ~~I added **unit tests** for my change.~~ This is a hotfix and there were no existing unit tests to update.
  - [ ] ~~_(if you fixed a bug)_ I added a **regression test** for the bug I fixed.~~ This was a configuration change, not a regression.
  - [ ] ~~_(if you added/updated a feature)_ I added/updated **e2e tests** at `examples/kitchen-sink/e2e-tests`.~~

- 📜 Documentation:

  - [ ] ~~_(if you added/updated a feature)_ I **added/updated the documentation** in `web/docs/`.~~

  - [ ] ~~_(if you did a breaking change)_ I added a step to the current **migration guide** at `web/docs/migration-guides/`.~~

### DIFF
--- a/examples/tutorials/TodoApp/package-lock.json
+++ b/examples/tutorials/TodoApp/package-lock.json
@@ -53,255 +53,6 @@
         "@tanstack/react-query": "~4.41.0"
       }
     },
-    ".wasp/out/sdk/wasp/node_modules/@node-rs/argon2": {
-      "version": "1.8.3",
-      "resolved": "https://registry.npmjs.org/@node-rs/argon2/-/argon2-1.8.3.tgz",
-      "integrity": "sha512-sf/QAEI59hsMEEE2J8vO4hKrXrv4Oplte3KI2N4MhMDYpytH0drkVfErmHBfWFZxxIEK03fX1WsBNswS2nIZKg==",
-      "license": "MIT",
-      "engines": {
-        "node": ">= 10"
-      },
-      "optionalDependencies": {
-        "@node-rs/argon2-android-arm-eabi": "1.8.3",
-        "@node-rs/argon2-android-arm64": "1.8.3",
-        "@node-rs/argon2-darwin-arm64": "1.8.3",
-        "@node-rs/argon2-darwin-x64": "1.8.3",
-        "@node-rs/argon2-freebsd-x64": "1.8.3",
-        "@node-rs/argon2-linux-arm-gnueabihf": "1.8.3",
-        "@node-rs/argon2-linux-arm64-gnu": "1.8.3",
-        "@node-rs/argon2-linux-arm64-musl": "1.8.3",
-        "@node-rs/argon2-linux-x64-gnu": "1.8.3",
-        "@node-rs/argon2-linux-x64-musl": "1.8.3",
-        "@node-rs/argon2-wasm32-wasi": "1.8.3",
-        "@node-rs/argon2-win32-arm64-msvc": "1.8.3",
-        "@node-rs/argon2-win32-ia32-msvc": "1.8.3",
-        "@node-rs/argon2-win32-x64-msvc": "1.8.3"
-      }
-    },
-    ".wasp/out/sdk/wasp/node_modules/@node-rs/argon2-android-arm-eabi": {
-      "version": "1.8.3",
-      "resolved": "https://registry.npmjs.org/@node-rs/argon2-android-arm-eabi/-/argon2-android-arm-eabi-1.8.3.tgz",
-      "integrity": "sha512-JFZPlNM0A8Og+Tncb8UZsQrhEMlbHBXPsT3hRoKImzVmTmq28Os0ucFWow0AACp2coLHBSydXH3Dh0lZup3rWw==",
-      "cpu": [
-        "arm"
-      ],
-      "license": "MIT",
-      "optional": true,
-      "os": [
-        "android"
-      ],
-      "engines": {
-        "node": ">= 10"
-      }
-    },
-    ".wasp/out/sdk/wasp/node_modules/@node-rs/argon2-android-arm64": {
-      "version": "1.8.3",
-      "resolved": "https://registry.npmjs.org/@node-rs/argon2-android-arm64/-/argon2-android-arm64-1.8.3.tgz",
-      "integrity": "sha512-zaf8P3T92caeW2xnMA7P1QvRA4pIt/04oilYP44XlTCtMye//vwXDMeK53sl7dvYiJKnzAWDRx41k8vZvpZazg==",
-      "cpu": [
-        "arm64"
-      ],
-      "license": "MIT",
-      "optional": true,
-      "os": [
-        "android"
-      ],
-      "engines": {
-        "node": ">= 10"
-      }
-    },
-    ".wasp/out/sdk/wasp/node_modules/@node-rs/argon2-darwin-arm64": {
-      "version": "1.8.3",
-      "resolved": "https://registry.npmjs.org/@node-rs/argon2-darwin-arm64/-/argon2-darwin-arm64-1.8.3.tgz",
-      "integrity": "sha512-DV/IbmLGdNXBtXb5o2UI5ba6kvqXqPAJgmMOTUCuHeBSp992GlLHdfU4rzGu0dNrxudBnunNZv+crd0YdEQSUA==",
-      "cpu": [
-        "arm64"
-      ],
-      "license": "MIT",
-      "optional": true,
-      "os": [
-        "darwin"
-      ],
-      "engines": {
-        "node": ">= 10"
-      }
-    },
-    ".wasp/out/sdk/wasp/node_modules/@node-rs/argon2-darwin-x64": {
-      "version": "1.8.3",
-      "resolved": "https://registry.npmjs.org/@node-rs/argon2-darwin-x64/-/argon2-darwin-x64-1.8.3.tgz",
-      "integrity": "sha512-YMjmBGFZhLfYjfQ2gll9A+BZu/zAMV7lWZIbKxb7ZgEofILQwuGmExjDtY3Jplido/6leCEdpmlk2oIsME00LA==",
-      "cpu": [
-        "x64"
-      ],
-      "license": "MIT",
-      "optional": true,
-      "os": [
-        "darwin"
-      ],
-      "engines": {
-        "node": ">= 10"
-      }
-    },
-    ".wasp/out/sdk/wasp/node_modules/@node-rs/argon2-freebsd-x64": {
-      "version": "1.8.3",
-      "resolved": "https://registry.npmjs.org/@node-rs/argon2-freebsd-x64/-/argon2-freebsd-x64-1.8.3.tgz",
-      "integrity": "sha512-Hq3Rj5Yb2RolTG/luRPnv+XiGCbi5nAK25Pc8ou/tVapwX+iktEm/NXbxc5zsMxraYVkCvfdwBjweC5O+KqCGw==",
-      "cpu": [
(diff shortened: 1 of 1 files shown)


waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -107,31 +107,24 @@ startPostgreDevDb waspProjectDir appName = do
...
-                "--env POSTGRES_USER=%s",
-                "--env POSTGRES_DB=%s",
-                "postgres"
+                printf "--publish %d:5432" Dev.Postgres.defaultDevPort,
+                printf "-v %s:%s" dockerVolumeName postgresDockerVolumePath,
+                printf "--env POSTGRES_PASSWORD=%s" Dev.Postgres.defaultDevPass,
+                printf "--env POSTGRES_USER=%s" Dev.Postgres.defaultDevUser,
+                printf "--env POSTGRES_DB=%s" dbName,
+                postgresDockerImage
               ]
```
infomiho: Instead of doing:
```hs
printf (unwords [...]) var1 var2
```
I went with:
```hs
 unwords [
  printf "" var1,
  printf "" var1 var2
]
```

The reason is that the list of `printf` vars became long and hard to follow.

I needed to add `postgresDockerVolumePath` in the middle of the list and I had to count the placeholder occurrences to know where to put it. This felt like a bug waiting to happen e.g. somebody puts something at the wrong position without realizing.

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -171,3 +164,18 @@ makeWaspDevDbDockerContainerName waspProjectDir appName =
...
+type PostgresDockerImage = String
+
+type PostgresDockerVolumeMountPath = String
+
+-- | We pin the Postgres Docker image to avoid issues when a new major version of Postgres
+-- is released. We aim to occasionally update this version in Wasp releases.
+-- If you bump the Postgres version, make sure to check if `postgresDockerVolumeMountPath`
+-- is still correct.
+waspDevDbPostgresDockerImageSpec :: (PostgresDockerImage, PostgresDockerVolumeMountPath)
+waspDevDbPostgresDockerImageSpec = ("postgres:18", postgresDockerVolumeMountPath)
```
infomiho, line 177: Since the Docker volume mount path is affected by major Postgres version changes - I coupled them in the "docker image spec" (maybe there is a better name, let me know).

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -171,3 +164,18 @@ makeWaspDevDbDockerContainerName waspProjectDir appName =
...
 maxDockerContainerNameLength = 63
+
+type PostgresDockerImage = String
+
+type PostgresDockerVolumeMountPath = String
+
+-- | We pin the Postgres Docker image to avoid issues when a new major version of Postgres
+-- is released. We aim to occasionally update this version in Wasp releases.
+-- If you bump the Postgres version, make sure to check if `postgresDockerVolumeMountPath`
+-- is still correct.
```
Martinsos: If you bump it where exactly? Aha, here, right? Ok!

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -171,3 +164,18 @@ makeWaspDevDbDockerContainerName waspProjectDir appName =
...
+waspDevDbPostgresDockerImageSpec :: (PostgresDockerImage, PostgresDockerVolumeMountPath)
+waspDevDbPostgresDockerImageSpec = ("postgres:18", postgresDockerVolumeMountPath)
+  where
+    -- Path inside the Postgres Docker container where the database files are stored.
```
Martinsos, line 179:
```suggestion
    -- | Path inside the Postgres Docker container where the database files are stored.
```
I thikn this is correct per Haddock? If not ignore.

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -171,3 +164,18 @@ makeWaspDevDbDockerContainerName waspProjectDir appName =
 
 maxDockerContainerNameLength :: Int
 maxDockerContainerNameLength = 63
+
+type PostgresDockerImage = String
+
+type PostgresDockerVolumeMountPath = String
```
Martinsos, line 170: You have "Mount" in it here, but above, you unpack it into just `postgresDockerVolumePath` wihtout Mount in it, that is probably by accident and we want to have those two match, for consistency?

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -107,31 +107,24 @@ startPostgreDevDb waspProjectDir appName = do
...
+          [ "docker run",
+            printf "--name %s" dockerContainerName,
+            "--rm",
+            printf "--publish %d:5432" Dev.Postgres.defaultDevPort,
+            printf "-v %s:%s" dockerVolumeName postgresDockerVolumePath,
+            printf "--env POSTGRES_PASSWORD=%s" Dev.Postgres.defaultDevPass,
+            printf "--env POSTGRES_USER=%s" Dev.Postgres.defaultDevUser,
+            printf "--env POSTGRES_DB=%s" dbName,
+            postgresDockerImage
+          ]
```
Martinsos, line 120: Before it was following the idea of "first goes the string tempalte with placeholders, then come the variables", so printf style.
Now, you went it mixing them all together, which someone might find better or not, I see pros and cons, but what you also did is used `printf` isnide for each one, I am not sure you got much with it? Why not just drop prints then and do normal concatenation? Ok there is one line wher eyou have two of htem. But I thikn `printf` here might be more noise than benefit.

But ok I see this is nuanced, so I am not sure -> let's put it this way, I am challenging this a bit, but I am ok with you leaving it like this if you still after that think this is the best.

RAW

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
Martinsos: Yup, that is great, both coupling and name I think.

Martinsos (approved):
@infomiho Looking good! I approved, but also left minor comments, please take care of those and then proceed.

One question: is there any other part of Wasp that assumes version of postgress? Maybe jobs? Shoudl we be thikning about having a central place where we specify the versino of postgress we use? PRobalby not in this PR, and maybe not at all, but I thought it might be worth considering it quickly.

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
FranjoMindek: Not for `where` block functions.

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
infomiho: Check my comment for some arguments against the previous approach: https://github.com/wasp-lang/wasp/pull/3323#discussion_r2502958262


I tried rewritten to concatenation, I still like the current state better:
SCR-20251107-mgjl

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
Martinsos: Ok yeah good argument!

Issue #3151, opened 2025-09-12
Rewrite our ./run development script so it also works on Windows


Martinsos:
We could rewrite it to:
- NodeJS script (requires node)
- https://github.com/casey/just script (requires `sh`)
- ...

FranjoMindek:
Google's [`zx`](https://github.com/google/zx) is really nice and we already use it in places.
e.g. `starters-e2e-tests`
It's written in javascript/typescript.

cprecioso:
I'd prefer to do Node since we already need it for development, and not introduce a new dependency such as `just`.

infomiho:
+1 for `zx` since we used it for `wasp-app-runner` and the `deploy` script already 👍

nodtem66:
It's just my curiosity to see how `zx` handles the shell process. This is a simple test to check if it can recognize `npm.ps1` and etc.
```js
import { usePowerShell, os } from 'zx'

if (os.type().includes("Windows")) {
   usePowerShell();
}
console.log(await $`npm --version`)
console.log(await $`npx --version`)
console.log(await $`node --version`)
```

It works! 
```
ProcessOutput {
  stdout: '11.4.2\n',
  stderr: '',
  signal: null,
  exitCode: 0,
  duration: 2532
}
ProcessOutput {
  stdout: '11.4.2\n',
  stderr: '',
  signal: null,
  exitCode: 0,
  duration: 2488
}
ProcessOutput {
  stdout: 'v22.14.0\r\n',
  stderr: '',
  signal: null,
  exitCode: 0,
  duration: 2150
}
```

infomiho:
I'd add to this issue - let's migrate all of our scripts 😄

cprecioso:
We decided to just make Bash a dependency on Windows #4347

PR #3324, opened 2025-11-07
[Issue/3148] ./run.ps1 for building waspc on Windows

nodtem66:
## Description

As discussed in #3318, we need a small equivalent to the `run` script that works, at least, for building Wasp on Windows. It is a temporary script before having a cross-platform script to handle this task (see #3151). This PR adds a `run.ps1` script that is compatible to PSVersion 5.1, which is the default version on Window 10/11. 

Example:
```
./run
USAGE
  run <command>

COMMANDS
  build             Builds the Haskell project.
  build:all         Builds the Haskell project + all sub-projects (i.e. TS packages).
  wasp-cli <args>   Runs the dev version of wasp executable while forwarding arguments.
                    Builds the project (hs) first if needed. Doesn't require you to be in the waspc project to run it.
```

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests:

- 📜 Documentation:


Martinsos:
This also contains commits from the other branch, with .bat for moving packages to data dir, which makes this PR a bit tricky to work on.

Suggestion: make this PR point to `main` branch instead, and remove the logic from "install packages script" PR. The only problem will be that the "install packages" script is missing on main, so in theory this PR is then incorrect as `run.ps1` will fail if it tries to run it, but I would solve that by just waiting a bit till that other PR gets merged into `main`, and in the meantime you can even create a "sub" "install packges" script in this one that says "Hello I am a stub" just to so you can test for example.

waspc/run.ps1
```diff
@@ -0,0 +1,38 @@
+# This script defines common commands used during building / developing `waspc` project
+# and makes it easy to run them.
```
Martinsos: I would that it duplicates (part of) logic from `run` shell script, and that in the future we plan to merge these two into one, platform-agnostic script, to avoid the duplication.

waspc/run.ps1
```diff
@@ -0,0 +1,38 @@
...
+
+param(
+    [string]$Command = ""
+)
+
+$PROJECT_ROOT = Split-Path -Parent $MyInvocation.MyCommand.Path
+$REPOSITORY_ROOT = Split-Path -Parent $PROJECT_ROOT
+    
+# Building
+$WASP_PACKAGES_COMPILE = Join-Path $PROJECT_ROOT "tools\install_packages_to_data_dir.bat"
```
Martinsos:
```suggestion
$WASP_PACKAGES_COMPILE = Join-Path $PROJECT_ROOT "tools\install_packages_to_data_dir.ps1"
```
to follow the latest changes on that side.

waspc/run.ps1
```diff
@@ -0,0 +1,38 @@
...
+        Invoke-Expression $RUN_CMD
+    }
+    Default {
+        Write-Host "USAGE"
+        Write-Host "  run <command>"
+        Write-Host ""
+        Write-Host "COMMANDS"
+        Write-Host "  build             Builds the Haskell project."
+        Write-Host "  build:all         Builds the Haskell project + all sub-projects (i.e. TS packages)."
+        Write-Host "  wasp-cli <args>   Runs the dev version of wasp executable while forwarding arguments."
```
Martinsos, line 35: Will indeed argument forwarding happen? That is something that happens by default in Powershell scripts?

Martinsos (commented):
This is looking great! Left some comments, let's take care of those and I think we can merge.

waspc/run.ps1
nodtem66: Yes, the script defines the first positional argument as `$Command` and `$Args` refers to the rest of arguments by default ([source](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_automatic_variables?view=powershell-5.1#args)).

waspc/run.ps1
nodtem66: I've changed the comments in the new commit.

nodtem66:
Thanks for the comments. I removed the unnecessary commits and added a stub file for `./install_packages_to_data_dir.ps1`.

waspc/run.ps1
Martinsos: Ah I missed the `$Args`! Ok, great!

Martinsos:
> Thanks for the comments. I removed the unnecessary commits and added a stub file for `./install_packages_to_data_dir.ps1`.

Ok all looking good, now just make it point to `main` branch and I thikn we are good! If you can't do that, I can also do it for you I think, let me know.

nodtem66:
Everything is done. It takes more time than I expected to change the base to `main`. I need to cherry-pick, rebase, and sadly force push to keep the same name of the target branch.

Martinsos (approved):
Great! Approving, but will wait for that other PR to get merged first, and then we will merge this one.

Martinsos:
> Everything is done. It takes more time than I expected to change the base to `main`. I need to cherry-pick, rebase, and sadly force push to keep the same name of the target branch.

Ok yeah that was expected! But now we have really clean situation here, easy to review and easy to merge.

Martinsos:
2 checks are failing but I thikn that is normal, it is a permission thing happening for contributor PRs.

waspc/run.ps1
```diff
@@ -0,0 +1,38 @@
...
+        Write-Host "                    Builds the project (hs) first if needed. Doesn't require you to be in the waspc project to run it."
+    }
+}
```
Martinsos:
```suggestion
}

```

waspc/tools/install_packages_to_data_dir.ps1
```diff
@@ -36,4 +36,4 @@ try {
 }
 finally {
     Pop-Location
-}
+}
```
Martinsos:
```suggestion
}

```

PR #3329, opened 2025-11-07
Release -> main

infomiho:
## Description

Merges `release` -> `main` with some logic changes as a result of a merge conflict. Talked with @Martinsos about it IRL and we agreed that it makes sense to do it like this:
1. The `--db-image` option still works
2. The volume mount path is not configurable - but we should make it configurable before we release the `--db-image` feature

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests:

- 📜 Documentation:


infomiho:
@cprecioso I get your point of view - at which point do we manage the experience vs. let go?

I feel like adding these kind of escape hatches frees us up in the future from needing to hotfix the mount path. So, I think it still makes sense to add it, just in case 😄

PR #3335, opened 2025-11-10
Add error to action

cprecioso:
## Description

The action was not erroring out in the case of it not finding the a successful workflow run. The empty string was then passed to `gh run download` which then just picks up the first run it finds instead of erroring out (this seems to be the intended behaviour).

Now we explicitly check and error out.

image

Example run: https://github.com/wasp-lang/open-saas/actions/runs/19230001646/job/54966078262?pr=540#step:6:36

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

### Not applicable

- 🧪 Tests:

- 📜 Documentation:

### DIFF
--- a/.github/actions/fetch-nightly-cli/action.yaml
+++ b/.github/actions/fetch-nightly-cli/action.yaml
@@ -34,6 +34,8 @@ runs:
         )
 
         if [ -z "$run_id" ]; then
+          # Using the ::error:: prefix lets GitHub handle it as a special string, and shows up as a red error in the logs.
+          # https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-an-error-message
           echo "::error::No successful workflow runs found for branch '$BRANCH'."
           exit 1
         fi


.github/actions/fetch-nightly-cli/action.yaml
```diff
@@ -32,6 +32,12 @@ runs:
             --json databaseId \
             --jq '.[0].databaseId'
         )
+
+        if [ -z "$run_id" ]; then
+          echo "::error::No successful workflow runs found for branch '$BRANCH'."
```
cprecioso, line 39: Using the `::error::` prefix lets GitHub handle it as a special string, and shows up as a red error in the logs.

Docs: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-an-error-message

Martinsos: Sounds good! Since you had the urge to ocmment this here, I would say let's also add this comment to the code here, as it is quite niche fact.

PR #3322, opened 2025-11-07
Reenable serving the web app from a subdirectory

sodic:
Fixes #3253 

Read the issue description. It tells you everything you need to know.

Issue #3021, opened 2025-08-01
Make it easy to write e2e tests that run on test-specific one-off Wasp project (for testing invariants)


sodic:
What often happens:
1. A user reports a bug and gives us a project to reproduce it.
2. We reproduce the bug and fix it.
4. We keep working on Wasp and break the same behavior again.

It would be nice if we had an easy way to write and maintain tests to ensure this doesn't happen, a way to automatically perform regression testing for all our bugs.

Some open source repos have a directoriy called `bugs` with many subdirectories named after github issues, each with a test that verifies the code doesn't have the bug described in that issue.

@Martinsos's formulation:
> A way to easily write/maintain tests that set up a specific wasp app for themselves and check its behaviour [...] 
> an easy way to define a test where I pick some kind of starter template for Wasp, define some changes to that code unique to this test (likely as a diff or patch or what is it called), and then also define some expected behaviour (be it wasp CLI behaviour, or something we do via playwright), and I have a test.
>
> So kind of an ability to define test-specific Wasp apps with minimal code + tests to run on them?

@cprecioso's idea:
> My 2c is that I'd take one main action: invest in having a easy-to-author test harness -- it is a bother to create a full wasp app to test only one aspect of our code. The boilerplate should be abstracted away. (Incidentally I think @miho's work on #2814 can be applied here)
Once we have that, we have to use it and create a lot of tests, and no new feature or bug without a regression test

sodic:
An example of what we should cover with this:
- https://github.com/wasp-lang/wasp/issues/3253#issuecomment-3406301125
- https://github.com/wasp-lang/wasp/pull/3339#issuecomment-3515915040

Implementing tests for that behavior didn't make sense at the time:
- Fringe feature.
- Complicated to test it.

After we standardize this kind of stuff, we should look into it again.

Issue #1591, opened 2023-12-06
Better support for deploying Wasp frontend to Github Pages


Martinsos:
Currently directly deploying Wasp frontend on GH pages doesn't work because of how client side routing is done, GH pages explain the issue here: https://create-react-app.dev/docs/deployment/#notes-on-client-side-routing .
They also explain different solutions: switching to different kind of routing, or doing this https://github.com/rafgraph/spa-github-pages#usage-instructions .

What we could do:
1. Switch from using HTML5 history API to routing with hashes. But I am not sure if we can also guarantee users of Waps will stick with it, so it sounds problematic.
2. Do the suggested trick with to teach GitHub Pages to handle 404s by redirecting to your index.html page with a custom redirect parameter. If they are deploying to GH pages, we could make sure to generate client files so that they are already ready for this.
3. Add documentation about this, so people are aware of the issue and can handle it on their own.

Relevant GH convo: https://discord.com/channels/686873244791210014/1170037967574802513/1170037967574802513 .

Bojun-Feng:
I ran into a similar issue when trying to deploy a Wasp project front-end (modified from OpenSaaS, version 0.17.0) on GitHub Pages via GitHub Actions. I tried several solutions and spend probably way more time than I should fixing the issue. I would like to document my attempts here. Since I am documenting failed attempts as well, this will be a long comment. 

If you are unfortunately suffering from similar issues and just needs a solution, I would recommend looking at the **Epilogue** section at the very end before reading the rest of this comment.

Do note I followed the migration guide to update my project to the newest version from version 0.15.0, which might be why some of the seemingly promising attempts did not work.

### Issue Description

The issue itself is pretty simple:
- I was visiting `https://user.github.io/project-name/` or the new homepage.
- The router knows nothing of the route `/project-name`.
- Router is mad and throws error. As a result:

The page shows:
```
There was an error rendering this page. Check the browser console for more information.
```
The console shows:
```
ErrorResponseImpl {status: 404, statusText: 'Not Found', internal: true, data: 'Error: No route matches URL "/project-name/"', error: Error: No route matches URL "/project-name/"
    at getInternalRouterError (http://loc…}
```

### Attempt 1: Update `package.json` (Failed)

I added the following field in `package.json`:
```diff
{
  "name": "opensaas",
  "type": "module",
+ "homepage": "https://user.github.io/project-name/",
  ...
```

[...]

PR #3339, opened 2025-11-10
Fix CORS when client is deployed to subdir

### LINKED ISSUE #3254: Make sure CORS works when client is deployed to a subdir
**Describe the bug**
We support `client.baseDir` option as a way for users to deploy the client app in a sub-directory. Users need to set the `WASP_WEB_CLIENT_URL` env variable on the server to an URL that includes the sub-directory i.e. `https://<something>/subdir`. Wasp then uses this full URL in the CORS settings. This breaks Wasp's CORS setup since subdirs are not considered different origins and this setup is wrong. Wasp should in this case set the allowed origin to just `https://<something>` (without the subdir part).

**To Reproduce**

1. Use the `client.baseDir` option
2. Deploy the client to a sub-directory e.g. Github Pages
3. Try to access the server from the client

**Expected behavior**
CORS should work.

sodic:
## Description

Fixes #3254 

## Type of change

- [x] **🐞 Bug fix** 
- [x] **💥 Breaking change** 

## Checklist

- 🧪 Tests: No tests, explained below.

- 📜 Documentation:

  - [x] _(if you did a breaking change)_ I added a step to the current **migration guide** at `web/docs/migration-guides/`.

### DIFF
--- a/web/docs/migration-guides/migrate-from-0-18-to-0-19.md
+++ b/web/docs/migration-guides/migrate-from-0-18-to-0-19.md
@@ -49,8 +49,6 @@ wasp ts-setup # ONLY if you are using the Wasp TS Config
 ```
 
 ### 3. Fix type errors caused by `config.allowedCORSOrigins`
-Search your codebase for the string `allowedCORSOrigins` and fix any potential type errors around its usage.
-### 3. Fix type errors caused by `config.allowedCORSOrigins`
 Search your codebase for the string `allowedCORSOrigins` and fix any potential type errors around its usage. You can follow [our middleware guide](../advanced/middleware-config.md) for the recommended way to extend the CORS configuration.
 
 If the search returns no results, it means you aren't using this feature and there's nothing to fix.


web/docs/advanced/middleware-config.md
```diff
@@ -98,7 +98,7 @@ If you would like to modify the middleware for _all_ operations and APIs, you ca
 
     export const serverMiddlewareFn = (middlewareConfig) => {
       // Example of adding extra domains to CORS.
-      middlewareConfig.set('cors', cors({ origin: [config.frontendUrl, 'https://example1.com', 'https://example2.com'] }))
+      middlewareConfig.set('cors', cors({ origin: [...config.allowedCorsOrigins, 'https://example1.com', 'https://example2.com'] }))
```
sodic: This seems more correct to me. Tell me if I'm missing something.

waspc/data/Generator/templates/sdk/wasp/universal/url.ts
```diff
@@ -3,3 +3,7 @@ export function stripTrailingSlash(url: undefined): undefined
 export function stripTrailingSlash(url?: string): string | undefined {
     return url?.replace(/\/$/, "");
 }
+
+export function getOrigin(url: string) {
+    return new URL(url).origin;
+}
```
sodic, line 9: A couple of comments related to @infomiho's [original implementation](https://github.com/wasp-lang/wasp/issues/1591#issuecomment-3397308595).
1. The original suggestion puts  this logic in `globalMiddleware`. I thought it made more sense to put it in `config.ts`, at the source.
2. The original does a `try..catch` to cover the case when the URLs aren't valid. I don't think this is necessary since we check URLs during the env validation phase. @infomiho tell me if I missed something.

Other possible improvements:
- It would be nice if we could transfer the info that "this URL is confirmed to be valid" through the type system. If we had that, it would have been obvious that the function here doesn't have to recheck it. But I'm guessing Zod doesn't offer this and we'd have to do it manually?
- It would be even better if we could parse the url string into an `URL` right away and have `config.frontendUrl` be of type `URL`. We'd have an easier time than calling `new URL` in a bunch of places. Is there a reason we wouldn't want to do this (e.g., some rule saying config values should be primitive types)?

If I got everything right, I can create issues for those refactors, something we can use to onboard new team members :)

sodic:
I did not write the tests for this (similar story to https://github.com/wasp-lang/wasp/pull/3322#issuecomment-3511678349):
- The cost outweights the benefits (fringe feature, very complex test).
- There's an effort to better support tests like these. I've mentioned this issue [there](https://github.com/wasp-lang/wasp/issues/3021).

examples/kitchen-sink/src/serverSetup.ts
```diff
@@ -49,7 +49,7 @@ export const serverMiddlewareFn: MiddlewareConfigFn = (middlewareConfig) => {
   // Example of adding an extra domain to CORS.
   middlewareConfig.set(
     "cors",
-    cors({ origin: [config.frontendUrl, "http://127.0.0.1:3000"] }),
+    cors({ origin: [...config.allowedCORSOrigins, "http://127.0.0.1:3000"] }),
```
infomiho, line 52: Nice catch!

waspc/ChangeLog.md
```diff
@@ -18,6 +18,8 @@ Remember to check out the [migration guide](https://wasp.sh/docs/migration-guide
 ### 🐞 Bug fixes
 
 - Fixed a type error with the default `NODE_ENV` value in the server env validation schema. ([#3189](https://github.com/wasp-lang/wasp/pull/3189))
+- Fixed frontend routing when serving the client from a non-root subpath/basedir ([#3322](https://github.com/wasp-lang/wasp/pull/3322))
```
infomiho: Is this added now because you missed it in the linked PR?

web/docs/advanced/middleware-config.md
infomiho: Have you tested this? `allowedCorsOrigins` is `*` in development, does that break the `cors` middleware?

I'd suggest looking into the implementation to double check: https://github.com/expressjs/cors/blob/master/lib/index.js#L41 

Based on what I saw, it looks like the effect of adding `*` in the array would be that it doesn't match the origin and that's it. But it will fail to match the localhost if not explicitly listed in the array.

waspc/data/Generator/templates/sdk/wasp/universal/url.ts
infomiho: Just to note: the "original solution" was me trying to debug the issue, so it's not authoritative in any way 😄 

1. I think it makes sense to put it in the config, at the source as you said.
2. The `try..catch` bit was there in the original because I was doing it directly on the origins which can be `*` - you are doing it here on an URL, so no need for that.
3. I agree with encoding the parsed/validated type in the type system. I believe we can use the [`brand()`](https://v3.zod.dev/?id=brand) option. 
4. Encoding it as URL makes sense since it's a URL. It would be a breaking change for users using the `config.frontendUrl` (but nothing a `.toString()` can't fix)

+1 for creating issues for these refactors

infomiho (commented):
Left some comments

waspc/ChangeLog.md
sodic: You know it.

web/docs/advanced/middleware-config.md
sodic: Great catch, this didn't even cross my mind. I only tested in production.
I'll double check.

waspc/data/Generator/templates/sdk/wasp/universal/url.ts
sodic: The brand thing is a great idea, I didn't know about it. I'll create the issues, thanks

examples/kitchen-sink/src/serverSetup.ts
sodic: Ah, if only I caught it before going crazy over my Wasp app gaslighting me.

waspc/data/Generator/templates/sdk/wasp/server/config.ts
```diff
@@ -11,20 +11,20 @@ type Config = {
   databaseUrl: string;
   frontendUrl: string;
   serverUrl: string;
-  allowedCORSOrigins: string | string[];
+  allowedCORSOrigins: '*' | string[];
```
sodic: I did not extract this into its own type on purpose - types are nicely displayed if I do it like this.

web/docs/advanced/middleware-config.md
sodic: @infomiho You are absolutely right!

We indeed had a problem with this. I slightly changed my approach, let me know what you think. I'm not 100% happy but don't know what to improve. Perhaps I could put comment in the docs and the kitchen sink app explaining the logic? Then again, I don't want to explain in so too much detail how `cors` works.

examples/kitchen-sink/src/serverSetup.ts
```diff
@@ -47,9 +47,11 @@ function addCustomRoute(app: Application) {
...
-  middlewareConfig.set(
-    "cors",
-    cors({ origin: [config.frontendUrl, "http://127.0.0.1:3000"] }),
-  );
+  if (config.allowedCORSOrigins !== "*") {
+    middlewareConfig.set(
+      "cors",
+      cors({ origin: [...config.allowedCORSOrigins, "http://127.0.0.1:3000"] }),
+    );
+  }
```
sodic: Users see that `config.allowedCORSOrigins` is of type `"*" | string[]` on hover.

Unfortunately, TS will still allow spreading eve if they don't eliminate the `"*"` option (since strings are iterable).

infomiho: Yep, I get why you dislike this solution, it's really indirect way of checking if it's dev mode vs. prod mode.

What if we maybe maybe the `allowedCORSOrigins` for example `string[] | undefined` where it's undefined in dev and we set it to `*` or something like that. I'm looking for a way to create a stronger signal. Maybe we could investigate how other frameworks handle this scenario and create an issue to improve the `config`?

I'll think about this and report back.

waspc/data/Generator/templates/sdk/wasp/universal/url.ts
sodic: Done: #3359, #3361

waspc/data/Generator/templates/sdk/wasp/server/config.ts
```diff
@@ -11,20 +11,20 @@ type Config = {
...
   {=# isAuthEnabled =}
   auth: {
     jwtSecret: string;
   }
   {=/ isAuthEnabled =}
 }
 
-const frontendUrl = stripTrailingSlash(env["{= clientUrlEnvVarName =}"])
-const serverUrl = stripTrailingSlash(env["{= serverUrlEnvVarName =}"])
+const frontendUrl = stripTrailingSlash(env['{= clientUrlEnvVarName =}'])
```
FranjoMindek, line 22: We are not following our `prettier` formatting standards here.
I know we can't really format this, but our manual code should still try to follow the rules:

```json
  "semi": true,
  "singleQuote": false,
  "jsxSingleQuote": false,
  "trailingComma": "all"
```

FranjoMindek (approved):
I didn't really notice anything other what has already been commented (e.g. using `URL` type for URLs).

examples/kitchen-sink/src/serverSetup.ts
cprecioso: What if we always have it always as an array? The origin option allows a heterogeneous array of string and regex, so we could keep our behaviour in dev and users wouldn't need to worry about if it's an array or not.

Our definitions would be
```ts
const allowedCORSOriginByMode = {
  development: [/.*/],
  production: [getOrigin(frontendUrl)]
}
```

and our users can do something like this, without the `if`:
```ts
middlewareConfig.set(
  "cors",
  cors({ origin: [...config.origin, "http://my-random-other-domain.com"] }),
);
```

cprecioso: see my suggestion at https://github.com/wasp-lang/wasp/pull/3339/files#r2526939903

sodic: Unfortunately this wouldn't work because of the problem Miho explained in a different thread: https://github.com/wasp-lang/wasp/pull/3339#discussion_r2513787641. In short, this would fail in development because the array passed to CORS does not include `localhost` and `*` inside the array is no longer treated as a magic value.

Plus, we don't want breaking changes and improvements with this PR, only to fix the bug.

cprecioso: > Plus, we don't want breaking changes

We're doing a breaking change anyway, so IMO let's do the breaking change if it needs to be one

> and * inside the array is no longer treated as a magic value

I didn't suggest passing `*`, I suggested passing an always-true regex `/.*/`

sodic: > I didn't suggest passing *, I suggested passing an always-true regex /.*/

Oh, I see. I didn't notice this. That's a very clever solution!

> We're doing a breaking change anyway

Hm, where'd you say the breaking change is? The way I see it, this is a bugfix:
- Our types now more precisely describe what was already happening.
- Our runtime now behaves as advertised for subdir stuff. Non-subdir logic is not affected.
- We've fixed an issue in the docs.

But, even if there was a breaking change, the emphasis is on the "...and improvements" part. I don't want to implement and test a new CORS system in this PR, refine our API and its types, change docs, etc. All I want is to do the bare minimum that fixes CORS issues in subdirs.

This is a great suggestion so I'm very tempted. But I'd still say it's of scope for this PR. I'll create a new issue and we can have it as a Kaizen thing or whatever in the next weeks.

Does that sound reasonable?

cprecioso: > Hm, where'd you say the breaking change is

No, I was just responding to your initial comment of "we don't want breaking changes with this PR". I mean to say that if we had to do a breaking change that'd be fine because we're releasing a major this week.

> All I want is to do the bare minimum that fixes CORS issues in subdirs.

IMO the total cost of putting this in the PR is way lower than the cost of leaving this off. Compared to recovering the shared context for author and reviewer in a new PR and have the merged fix wait for the next release. Be it in a week or, even worse, in a handful of months because we didn't get to Kaizen this.

You may have a different opinion, so I leave it in your hands! 💪

infomiho: > I'll think about this and report back.

Nothing new to comment here - I'll approve the PR as-is - let's encode any improvements not implemented now in a new issue that should ideally end up in the Product Backlog.

waspc/data/Generator/templates/sdk/wasp/server/config.ts
sodic: I decided against doing this and used formatting more consistent with the current file. The choice was either that, having mixed formatting rules, or fixing the entire file manually.

I didn't want to mix formatting fixes and the bugfix in the same PR.

examples/kitchen-sink/src/serverSetup.ts
sodic: I implemented Carlos's suggestion in the end. He was right, it wasn't that much work. I guess I'm just paranoid.

Anyway, it's pretty neat!

waspc/ChangeLog.md
```diff
@@ -15,12 +16,15 @@ Remember to check out the [migration guide](https://wasp.sh/docs/migration-guide
 - You can now specify a custom volume mount path for the development database with the `--db-volume-mount-path` option in `wasp start db`. ([#3362](https://github.com/wasp-lang/wasp/pull/3362))
 - You can now specify which PostgreSQL image to use in `wasp deploy railway` with the `--db-image` argument. ([#3184](https://github.com/wasp-lang/wasp/pull/3184))
 - You can now specify which PostgreSQL image to use in `wasp deploy fly` with the `--db-image` argument. ([#3187](https://github.com/wasp-lang/wasp/pull/3187))
+- Improved DX around `allowedCORSOrigins` ([#3339](https://github.com/wasp-lang/wasp/pull/3339))
```
cprecioso:
```suggestion
- It's now simpler to extend our default CORS rules. ([#3339](https://github.com/wasp-lang/wasp/pull/3339))
```

waspc/ChangeLog.md
```diff
@@ -15,12 +16,15 @@ Remember to check out the [migration guide](https://wasp.sh/docs/migration-guide
...
 
 ### 🐞 Bug fixes
 
 - Fixed a type error in the generated server app when `process.env.NODE_ENV` is also declared by another dependency in the project. ([#3189](https://github.com/wasp-lang/wasp/pull/3189))
 - Fixed an incompatibility between `wasp deploy fly` and Fly CLI v0.3.214. ([#3372](https://github.com/wasp-lang/wasp/pull/3372))
 - Raised the internal requirement of Tailwind CSS from `^3.2.7` to `^3.4.17` to ensure compatibility with ESM config files. ([#3380](https://github.com/wasp-lang/wasp/issues/3380))
+- Fixed frontend routing when serving the client from a non-root subpath/basedir ([#3322](https://github.com/wasp-lang/wasp/pull/3322))
```
cprecioso:
```suggestion
- Fixed an error with routing when serving the client from a subpath in a domain. ([#3322](https://github.com/wasp-lang/wasp/pull/3322))
```

waspc/ChangeLog.md
```diff
@@ -15,12 +16,15 @@ Remember to check out the [migration guide](https://wasp.sh/docs/migration-guide
...
+- Fixed CORS issues when serving the client on a non-root subpath/basedir ([#3339](https://github.com/wasp-lang/wasp/pull/3339))
```
cprecioso:
```suggestion
- Fixed an error with CORS when serving the client from a subpath in a domain. ([#3339](https://github.com/wasp-lang/wasp/pull/3339))
```

web/docs/migration-guides/migrate-from-0-18-to-0-19.md
```diff
@@ -8,6 +8,10 @@ title: Migration from 0.18.X to 0.19.X
 
 Wasp now enables npm workspaces for managing the generated app. This change makes our dependency system more reliable, and better prepares us for future features. It also makes installs faster overall and reduces the size of each project on your disk. This is largely transparent, and you shouldn't notice any difference in how you develop your app.
 
+### Breaking API change for `config.allowedCORSOrigins`
+The type of `config.allowedCORSOrigins` (imported from `wasp/server`) was changed from `string | string[]` to `(String | RegExp)[]`.
+This was done to make Wasp's CORS system more compsable with the `cors` package.
```
cprecioso:
```suggestion
### The type of `config.allowedCORSOrigins` has changed

The type of `config.allowedCORSOrigins` (imported from `wasp/server`) was changed from `string | string[]` to `(string | RegExp)[]`, which is always an array. Now, it's simpler to extend our default CORS rules if you just want to add extra domains to the list.
```

web/docs/advanced/middleware-config.md
cprecioso:
```suggestion
      middlewareConfig.set('cors', cors({ origin: [...config.allowedCORSOrigins, 'https://example1.com', 'https://example2.com'] }))
```

web/docs/advanced/middleware-config.md
```diff
@@ -121,7 +121,7 @@ If you would like to modify the middleware for _all_ operations and APIs, you ca
 
     export const serverMiddlewareFn: MiddlewareConfigFn = (middlewareConfig) => {
       // Example of adding an extra domains to CORS.
```
cprecioso:
```suggestion
      // Example of adding extra domains to CORS.
```

web/docs/migration-guides/migrate-from-0-18-to-0-19.md
```diff
@@ -44,6 +48,10 @@ rm package-lock.json
 wasp ts-setup # ONLY if you are using the Wasp TS Config
 ```
 
-### 3. Enjoy your updated Wasp app
+### 3. Fix type errors caused by `config.allowedCORSOrigins`
+Search your codebase for the string `allowedCORSOrigins` and fix any potential type errors around its usage.
+If the search returns no results, it means you aren't using this feature and there's nothing to fix.
```
cprecioso, line 54:
~~~suggestion
### 3. Fix type errors caused by `config.allowedCORSOrigins`
Search your codebase for the string `allowedCORSOrigins` and fix any potential type errors around its usage. You can follow [our middleware guide](../advanced/middleware-config.md) for the recommended way to extend the CORS configuration.

If the search returns no results, it means you aren't using this feature and there's nothing to fix.
~~~

PR #3344, opened 2025-11-10
Extract file path argument parser

cprecioso:
## Description

As part of https://github.com/wasp-lang/wasp/pull/3338, I need to add a dir path argument parser (https://github.com/wasp-lang/wasp/pull/3343).
When doing that, it makes sense to extract the file path argument parser from `EnvVarArgument` into its own file.

## Connections

- Working towards https://github.com/wasp-lang/wasp/issues/3073
- Part of https://github.com/wasp-lang/wasp/pull/3338 -- I split the refactors into their own PRs for ease of review:
	- https://github.com/wasp-lang/wasp/pull/3340
	- https://github.com/wasp-lang/wasp/pull/3341
	- https://github.com/wasp-lang/wasp/pull/3342
	- https://github.com/wasp-lang/wasp/pull/3343
	- https://github.com/wasp-lang/wasp/pull/3344
	- https://github.com/wasp-lang/wasp/pull/3345
	- https://github.com/wasp-lang/wasp/pull/3346

## Type of change

- [x] **🔧 Just code/docs improvement** 
- [ ] **🐞 Bug fix** 
- [ ] **🚀 New/improved feature** 
- [ ] **💥 Breaking change** 

## Checklist

- [x] I tested my change in a Wasp app to verify that it works as intended.

- 🧪 Tests:
  - [ ] ~~I added **unit tests** for my change.~~ 
  - [ ] ~~_(if you fixed a bug)_ I added a **regression test** for the bug I fixed. ~~
  - [ ] ~~_(if you added/updated a feature)_ I added/updated **e2e tests** at `examples/kitchen-sink/e2e-tests`.~~
  - [ ] ~~_(if you added/updated a feature)_ I updated the **starter templates** at `waspc/data/Cli/templates`, as needed.~~
  - Just moving stuff

- 📜 Documentation:
  - [ ] ~~_(if you added/updated a feature)_ I **added/updated the documentation** in `web/docs/`.~~

- 🆕 Changelog: _(if change is more than just code/docs improvement)_
  - [ ] ~~I updated `waspc/ChangeLog.md` with a **user-friendly** description of the change.~~
  - [ ] ~~_(if you did a breaking change)_ I added a step to the current **migration guide** at `web/docs/migration-guides/`.~~
  - [ ] ~~I **bumped the `version`** in `waspc/waspc.cabal` to reflect the changes I introduced.~~

PR #3343, opened 2025-11-10
Add dir path argument parser

cprecioso:
## Description

As part of https://github.com/wasp-lang/wasp/pull/3338, I need to add a dir path argument parser.

## Connections

- On top of https://github.com/wasp-lang/wasp/pull/3344
- Working towards https://github.com/wasp-lang/wasp/issues/3073
- Part of https://github.com/wasp-lang/wasp/pull/3338 -- I split the refactors into their own PRs for ease of review:
	- https://github.com/wasp-lang/wasp/pull/3340
	- https://github.com/wasp-lang/wasp/pull/3341
	- https://github.com/wasp-lang/wasp/pull/3342
	- https://github.com/wasp-lang/wasp/pull/3344
	- https://github.com/wasp-lang/wasp/pull/3345
	- https://github.com/wasp-lang/wasp/pull/3346

## Type of change

- [x] **🔧 Just code/docs improvement** 
- [ ] **🐞 Bug fix** 
- [ ] **🚀 New/improved feature** 
- [ ] **💥 Breaking change** 

> [!NOTE]
> If we're strict, this is adding a new feature, but as it is not yet exposed, I am opting to bump the changelog in the PR that actually exposes this feature. (https://github.com/wasp-lang/wasp/pull/3338)

## Checklist

- [x] I tested my change in a Wasp app to verify that it works as intended.

- 🧪 Tests:
  - [ ] ~~I added **unit tests** for my change.~~
    - I skipped adding unit tests for this as it is a quite straight-forward wrapping and unwrapping of FilePath and StrongPath types
  - [ ] ~~_(if you fixed a bug)_ I added a **regression test** for the bug I fixed. ~~
  - [ ] ~~_(if you added/updated a feature)_ I added/updated **e2e tests** at `examples/kitchen-sink/e2e-tests`.~~
  - [ ] ~~_(if you added/updated a feature)_ I updated the **starter templates** at `waspc/data/Cli/templates`, as needed.~~

- 📜 Documentation:
  - [ ] ~~_(if you added/updated a feature)_ I **added/updated the documentation** in `web/docs/`.~~

- 🆕 Changelog: _(if change is more than just code/docs improvement)_
  - [ ] ~~I updated `waspc/ChangeLog.md` with a **user-friendly** description of the change.~~
  - [ ] ~~_(if you did a breaking change)_ I added a step to the current **migration guide** at `web/docs/migration-guides/`.~~
[...]

PR #3342, opened 2025-11-10
Refactor ProjectDescription

cprecioso:
## Description

The logic in ProjectDescription is not easy to follow. While adding the functionality for #3338, I refactored it to make it simpler. I extracted the refactor to this PR.

In the previous version, we were calling down to slightly different implementations of similar functions that each dealt with different variables being missing, and the functions were calling down to each other, which made the logic more difficult to follow. 
It also made extension difficult (like the one I need to do in #3338), because the functions needed very specific combinations of missing/present information.

In the new version, the missing information is handled at the top level, and all the extra functions consolidated to a single one that only deals with present information. That way we remove the multiple layers of indirection, and it's very easy to see where we get or ask for each piece of information. This structure also makes it easy for me to extend and share logic in #3338.

## Connections

- Working towards https://github.com/wasp-lang/wasp/issues/3073
- Part of https://github.com/wasp-lang/wasp/pull/3338 -- I split the refactors into their own PRs for ease of review:
	- https://github.com/wasp-lang/wasp/pull/3340
	- https://github.com/wasp-lang/wasp/pull/3341
	- https://github.com/wasp-lang/wasp/pull/3342
	- https://github.com/wasp-lang/wasp/pull/3343
	- https://github.com/wasp-lang/wasp/pull/3344
	- https://github.com/wasp-lang/wasp/pull/3345
	- https://github.com/wasp-lang/wasp/pull/3346

## Type of change

- [x] **🔧 Just code/docs improvement** 
- [ ] **🐞 Bug fix** 
- [ ] **🚀 New/improved feature** 
- [ ] **💥 Breaking change** 

## Checklist

- [x] I tested my change in a Wasp app to verify that it works as intended.

- 🧪 Tests:
  - [ ] ~~I added **unit tests** for my change. ~~
  - [ ] ~_(if you fixed a bug)_ I added a **regression test** for the bug I fixed. ~
  - [ ] ~~_(if you added/updated a feature)_ I added/updated **e2e tests** at `examples/kitchen-sink/e2e-tests`.~~
  - [ ] ~~_(if you added/updated a feature)_ I updated the **starter templates** at `waspc/data/Cli/templates`, as needed.~~
  - [x] Already covered by starters tests

- 📜 Documentation:
  - [ ] ~~_(if you added/updated a feature)_ I **added/updated the documentation** in `web/docs/`.~~

- 🆕 Changelog: _(if change is more than just code/docs improvement)_
[...]

PR #3340, opened 2025-11-10
Refactor StarterTemplate data structure

cprecioso:
## Description

In the `StarterTemplate` record, we were overloading `DirBasedTemplateMetadata` with multiple different meanings depending on the which kind of `StarterTemplate` we had.

I converted each constructor into record syntax, and extracted there the relevant fields from `DirBasedTemplateMetadata` (now called just `TemplateMetadata`).

## Connections

- On top of https://github.com/wasp-lang/wasp/pull/3346
- On top of https://github.com/wasp-lang/wasp/pull/3345
- Working towards https://github.com/wasp-lang/wasp/issues/3073
- Part of https://github.com/wasp-lang/wasp/pull/3338 -- I split the refactors into their own PRs for ease of review:
	- https://github.com/wasp-lang/wasp/pull/3340
	- https://github.com/wasp-lang/wasp/pull/3341
	- https://github.com/wasp-lang/wasp/pull/3342
	- https://github.com/wasp-lang/wasp/pull/3343
	- https://github.com/wasp-lang/wasp/pull/3344
	- https://github.com/wasp-lang/wasp/pull/3345
	- https://github.com/wasp-lang/wasp/pull/3346

## Type of change

- [x] **🔧 Just code/docs improvement** 
- [ ] **🐞 Bug fix** 
- [ ] **🚀 New/improved feature** 
- [ ] **💥 Breaking change** 

## Checklist

- [x] I tested my change in a Wasp app to verify that it works as intended.

- 🧪 Tests:
  - [ ] ~~I added **unit tests** for my change. ~~
  - [ ] ~~_(if you fixed a bug)_ I added a **regression test** for the bug I fixed. ~~
  - [ ] ~~_(if you added/updated a feature)_ I added/updated **e2e tests** at `examples/kitchen-sink/e2e-tests`.~~
  - [ ] ~~_(if you added/updated a feature)_ I updated the **starter templates** at `waspc/data/Cli/templates`, as needed.~~
  - Only moving stuff at the type level

- 📜 Documentation:
  - [ ] ~~_(if you added/updated a feature)_ I **added/updated the documentation** in `web/docs/`.~~

- 🆕 Changelog: _(if change is more than just code/docs improvement)_
  - [ ] ~~I updated `waspc/ChangeLog.md` with a **user-friendly** description of the change.~~
  - [ ] ~~_(if you did a breaking change)_ I added a step to the current **migration guide** at `web/docs/migration-guides/`.~~
[...]

PR #3345, opened 2025-11-10
Split StarterTemplates to AvailableTemplates

cprecioso:
## Description

As part of adding one more kind of StarterTemplate (https://github.com/wasp-lang/wasp/pull/3338), the `StarterTemplates.hs` file was getting crowded, so I left in this file all the types and definitions, and moved the specifications of available templates to their own file.

## Connections

- Working towards https://github.com/wasp-lang/wasp/issues/3073
- Part of https://github.com/wasp-lang/wasp/pull/3338 -- I split the refactors into their own PRs for ease of review:
	- https://github.com/wasp-lang/wasp/pull/3340
	- https://github.com/wasp-lang/wasp/pull/3341
	- https://github.com/wasp-lang/wasp/pull/3342
	- https://github.com/wasp-lang/wasp/pull/3343
	- https://github.com/wasp-lang/wasp/pull/3344
	- https://github.com/wasp-lang/wasp/pull/3345
	- https://github.com/wasp-lang/wasp/pull/3346

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests:
  - [ ] ~~I added **unit tests** for my change. ~~
  - [ ] ~~_(if you fixed a bug)_ I added a **regression test** for the bug I fixed. ~~
  - [ ] ~~_(if you added/updated a feature)_ I added/updated **e2e tests** at `examples/kitchen-sink/e2e-tests`.~~
  - Just moving stuff

- 📜 Documentation:
  - [ ] ~~_(if you added/updated a feature)_ I **added/updated the documentation** in `web/docs/`.~~

  - [ ] ~~I updated `waspc/ChangeLog.md` with a **user-friendly** description of the change.~~
  - [ ] ~~_(if you did a breaking change)_ I added a step to the current **migration guide** at `web/docs/migration-guides/`.~~
  - [ ] ~~I **bumped the `version`** in `waspc/waspc.cabal` to reflect the changes I introduced.~~

### DIFF
--- a/waspc/cli/src/Wasp/Cli/Command/CreateNewProject/AvailableTemplates.hs
+++ b/waspc/cli/src/Wasp/Cli/Command/CreateNewProject/AvailableTemplates.hs
@@ -106,7 +106,7 @@ openSaasStarterTemplate =
     )
 
 -- NOTE: As version of Wasp CLI changes, so we should update this tag name here,
---   and also create it on gh repos of templates.
+--   and also create it on the GitHub repos of templates.
 --   By tagging templates for each version of Wasp CLI, we ensure that each release of
 --   Wasp CLI uses correct version of templates, that work with it.
 waspVersionTemplateGitTag :: String


waspc/cli/src/Wasp/Cli/Command/CreateNewProject/AvailableTemplates.hs
```diff
@@ -0,0 +1,113 @@
...
+              styleCode $ "    wasp start",
+              styleText $ "",
+              styleText $ "Check the " <> styleCode "README.md" <> " for additional guidance!"
+            ]
+      }
+
+{- HLINT ignore openSaasStarterTemplate "Redundant $" -}
+
+openSaasStarterTemplate :: StarterTemplate
+openSaasStarterTemplate =
```
cprecioso, line 66: This previously used a `simpleGhReleaseArchiveTemplate`, which was just one more layer that provided little abstraction nor extra functionality, and was only used here. I removed it and now the function creates the structure directly.

waspc/src/Wasp/Util/Terminal.hs
```diff
@@ -14,6 +15,12 @@ import Data.List (foldl')
 styleCode :: String -> String
 styleCode = applyStyles [Bold]
 
+-- | Applies the Wasp CLI standardized text styling to a string. Currently, this
+-- is a no-op, but it exists for future-proofing. And to make consecutive calls
+-- to styleCode and styleText look similar.
+styleText :: String -> String
+styleText = id
```
cprecioso, line 22: Extracted this here since it is used in both `AvailableTemplates` and `StarterTemplates`, and in this way is colocated with `styleCode`.

waspc/README.md
```diff
@@ -432,9 +432,9 @@ Do the non-bold steps when necessary (decide for each step depending on the chan
 - Update the starter templates if necessary (i.e., if there are breaking changes or new features they should make use of):
   - Context: they are used by used by `wasp new`, you can find reference to them in `Wasp.Cli. ... .StarterTemplates`.
   - Check and merge all PRs with the label `merge-before-release`.
-  - In `StarterTemplates.hs` file, update git tag to new version of Wasp we are about to release (e.g. `wasp-v0.13.1-template`).
+  - In `AvailableTemplates.hs` file, update git tag to new version of Wasp we are about to release (e.g. `wasp-v0.13.1-template`).
```
cprecioso: I will update the Notion checklist once we merge this

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/AvailableTemplates.hs
```diff
@@ -0,0 +1,113 @@
...
+                styleText $ "  6. Last step: run the app!",
+                styleCode $ "    wasp start",
+                styleText $ "",
+                styleText $ "Check the README for additional guidance and the link to docs!"
+              ]
+        }
+    )
+
+-- NOTE: As version of Wasp CLI changes, so we should update this tag name here,
+--   and also create it on gh repos of templates.
```
infomiho:
```suggestion
--   and also create it on Github repos of templates.
```

I know it's old code, but I thought why not.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/AvailableTemplates.hs
```diff
@@ -0,0 +1,113 @@
...
+        }
+    )
+    "template.tar.gz"
+    ( DirBasedTemplateMetadata
+        { _name = "saas",
+          _description =
+            "Everything a SaaS needs! Comes with Auth, ChatGPT API, Tailwind, Stripe payments and more."
+              <> " Check out https://opensaas.sh/ for more details.",
+          -- We assume that the archive contains files at its root.
+          _path = [reldir|.|],
```
infomiho, line 81: I know it's old code, but... can we encode this assumption in a variable (ideally we'd have a helper function like `createGhRepoReleaseArchiveTemplate` that would set this value always). 

Imagine adding the next remote template like `OpenTool` and we'd have to repeat this path + the comment. Of course we'd extract it into a separate variable at least or even introduce the helper function.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/AvailableTemplates.hs
```diff
@@ -0,0 +1,113 @@
...
+            ]
+      }
+
+{- HLINT ignore openSaasStarterTemplate "Redundant $" -}
+
+openSaasStarterTemplate :: StarterTemplate
+openSaasStarterTemplate =
+  GhRepoReleaseArchiveTemplate
+    ( GhRepo.GithubRepoRef
+        { GhRepo._repoOwner = "wasp-lang",
```
infomiho, line 69: We also dropped the `waspGhOrgName` which was a nice way to encode this is the "official" repo name vs. just one of the Open SaaS specific values.

infomiho (commented):
The refactor is good, but I think there are some assumptions we can better encode.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/AvailableTemplates.hs
cprecioso: I disagree, this is "just one of the OpenSaaS specific values". We're just saying where OS is, the fact that it is in the wasp gh org here holds no meaning by itself.
For example, if we moved OS to its own org, this proposed variable would just have to change too, because it doesn't encode any actual reusable data.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/AvailableTemplates.hs
cprecioso: I disagree. All of the data in this record _are_ assumptions. We're assuming OS will reside in that repo, with that tag name, and that path. The path is no different, and doesn't warrant an assumption IMO.

If we added a hypothetic OpenTool, yep, we'll extract what we need in a reusable helper. But with only OS here, the helper is just one more indirection which we could skip.

cprecioso: I also think that, together with the refactor at https://github.com/wasp-lang/wasp/pull/3340, this code reads much nicer and is really clear on where we are hardcoding assumptions.

infomiho: Ok, so you would treat all the assumptions here as Open SaaS specific until we introduce another remote template. That's a fair thing to assume. We can refactor this with more data points if/when we add a new remote template.

waspc/cli/src/Wasp/Cli/Command/CreateNewProject/AvailableTemplates.hs
infomiho: I think we can assume that all Wasp provided templates will come from the Wasp Github org. I'm okay with not doing anything now since there is only one remote GH template.

waspc/README.md
cprecioso: This is not in the Notion checklist anymore 💁‍♀️

PR #3347, opened 2025-11-11
Fix naming in CI workflows.

sodic:


.github/workflows/ci-waspc-build.yaml
```diff
@@ -22,6 +22,7 @@ env:
 
 jobs:
   build:
+    name: Build Wasp
```
cprecioso: Can we have this be something like "Build Wasp (${{ matrix.env.name }}))"?

cprecioso (approved):
LGTM, Left one suggestion

.github/workflows/ci-waspc-build.yaml
cprecioso: This is wrong, we are already defining a name on L90 https://github.com/wasp-lang/wasp/blob/main/.github/workflows/ci-waspc-build.yaml#L90

cprecioso: Disregard this, look at the following review

PR #3349, opened 2025-11-12
First experiment with using Brick for TUI.

Martinsos:
This is just a quick experiment with Brick to show that we can easily get a nice TUI in haskell for our CLI. Code is very nice I think, and library is very powerful and handles a lot of stuff out of the box (it is also cross-platform!).

image

As mentioned in the comments in the code, main thing that will be missing for us to use it for `wasp start` is logic for replacing ansi control characters in the output received from the child processes with `Vty` styling, but that is basically one function/module that we can then reuse accross the whole TUI and LLM should be able to help us a lot to write it. We can even open source it named `brick-ansi`, I saw other people on `brick` github asking for it.

I will leave this draft here as a potential starting point. Whoever is implementing this, first read the Brick guide.

### DIFF
--- a/waspc/cli/src/Wasp/Cli/Command/Start.hs
+++ b/waspc/cli/src/Wasp/Cli/Command/Start.hs
@@ -84,7 +84,7 @@ ui =
                   B.padLeftRight 1 $ B.str "3:Db",
                   B.padLeftRight 1 $ B.str "4:All"
                 ],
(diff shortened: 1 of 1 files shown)


Martinsos:
Closing but let's use it as a reference if we go down this road, I was happy with the work done here, I think it is a good foundation to continue.

PR #3362, opened 2025-11-13
Add --db-volume-mount-path option for dev db

### LINKED ISSUE #3331: Support for DB mount path in `wasp db start`
We recently had a [hotfix](https://github.com/wasp-lang/wasp/pull/3323) that required us to modify the database mount path for the dev database (`wasp db start`). 

For our dev database, we now have [two default values](https://github.com/wasp-lang/wasp/blob/31c619e61131c7448a9374c355963c8a477cfac1/waspc/src/Wasp/Db/Postgres.hs#L26):
- database image (`postgres:18`)
- and the database mount path (`/var/lib/postgresql`)

These two values influence each-other and in the future when we change the default image, we'll see if we need to change the default mount path.

**The trouble is that we have the option for `wasp db start` called [`--db-image`](https://github.com/wasp-lang/wasp/blob/31c619e61131c7448a9374c355963c8a477cfac1/waspc/cli/src/Wasp/Cli/Command/Start/Db.hs#L66) which allows you to change the image used by the command. However, we don't allow you to change the default mount path.** 

You probably wouldn't need to change it - but as the hotfix showed - maybe it becomes important later on. So, in the name of future-proofing our setup, let's also add `--db-image-mount-path` or something like that.

infomiho:
## Description

Closes #3331

Adds `--db-volume-mount-path` option to `wasp start db` to allow overriding the default Docker volume mount path.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests:

  - [ ] ~~I added **unit tests** for my change.~~ Adding a new option to `optparse-applicative`
  - [ ] _(if you added/updated a feature)_ ~~I added/updated **e2e tests** at `examples/kitchen-sink/e2e-tests`.~~ Not using a custom DB image.

- 📜 Documentation:

### DIFF
--- a/web/docs/data-model/databases.md
+++ b/web/docs/data-model/databases.md
@@ -89,9 +89,7 @@ In case you might want to connect to the dev database through the external tool 
 
 ##### Customising the dev database {#custom-database}
 
-Wasp development database by default uses:
-- [`postgres:18`](https://hub.docker.com/_/postgres/tags?name=18) Docker image,
-- and the `/var/lib/postgresql` volume mount path.
+The Wasp development database uses the [PostgreSQL 18 Docker image](https://hub.docker.com/_/postgres/tags?name=18) by default, and will set up its data volumes according to their guidance.
 
 If you need to customise the development database, you can use the following options:
 


waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -63,18 +63,23 @@ startDbArgsParser :: Opt.Parser StartDbArgs
...
-      -- the db image AND the mount path in Wasp.Db.Postgres.defaultPostgresDockerImageSpec
-      -- but we should allow users to override both of these and not just the db image.
       ( Opt.long "db-image"
           <> Opt.metavar "IMAGE"
           <> Opt.help "Docker image to use for the database"
           <> Opt.showDefault
           <> Opt.value (fst defaultPostgresDockerImageSpec)
       )
+    <*> Opt.strOption
+      ( Opt.long "db-volume-mount-path"
```
infomiho, line 78: I'm on the fence if we should name it `db-data-dir` but this name is more precise while `db-data-dir` is more close to the user.

.github/workflows/automation-pr-label-external.yaml
cprecioso, line 1: i guess this will be deduplicated on merge

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
cprecioso: I also think the current name is a bit too verbose, but I think that any shorter name could get confused with "this flag controls where the db is stored in _their_ filesystem", so I vote to keep this one.

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -53,7 +53,7 @@ start args = do
 
   case ASV.getValidDbSystem appSpec of
     AS.App.Db.SQLite -> noteSQLiteDoesntNeedStart
-    AS.App.Db.PostgreSQL -> startPostgresDevDb waspProjectDir appName (dbImage startDbArgs)
+    AS.App.Db.PostgreSQL -> startPostgresDevDb waspProjectDir appName (dbImage startDbArgs) (dbVolumeMountPath startDbArgs)
```
cprecioso: weak, idea: should we just pass `StartDbArgs` to the `startPostgresDevDb` fn? I realize this could be "crossing domains" so feel free to disregard.

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
```diff
@@ -124,9 +129,10 @@ startPostgresDevDb waspProjectDir appName dbDockerImage = do
         "",
         "Additional info:",
         " ℹ Using Docker image: " <> dbDockerImage,
+        " ℹ Docker volume mount path inside container: " <> dbDockerVolumeMountPath,
```
cprecioso: IMO feels like using too many "ℹ" characters kind of overloads this message. I'm not 100% on this suggestion so feel free to disregard or go with what you think is best.

```suggestion
        "   with data volume mounted at: " <> dbDockerVolumeMountPath,
```

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
cprecioso: or, being that you're defininig the default as a `(imageName, volumePath)` pair, you can keep using that in here too

web/docs/data-model/databases.md
```diff
@@ -89,22 +89,32 @@ In case you might want to connect to the dev database through the external tool
...
-You can also specify a custom Docker image to use for the database with the `--db-image` option. This is particularly useful when you need PostgreSQL with specific extensions (like PostGIS for geographic data, pgvector for embeddings, etc.), or an older version. By default, Wasp uses the latest official [PostgreSQL Docker image](https://hub.docker.com/_/postgres).
+  You can customize the development database using two options:
+
+  - `--db-image`: Specify a custom Docker image
+      
+    Useful for PostgreSQL extensions or specific versions (for example, PostGIS, pgvector, etc.). Wasp uses the [`postgres:18`](https://hub.docker.com/_/postgres/tags?name=18) Docker image by default.
+
+  - `--db-volume-mount-path`: Specify the volume mount path inside the container
+  
+    You only need to worry about this if your custom `--db-image` is based on **Postgres 17 or older**. If the volume mount path is incorrect, the data won't be persisted in your development database.
```
cprecioso: This message is a bit vague... I can see users thinking Why would I be on PG17? What do I actually need to do with this flag then?


```suggestion
    If you have an existing database from an app using Wasp 0.18 or older, you can use this flag in this way: blah blah
```

cprecioso: I see there's an example below, maybe we can call it out explictly so we avoid people feeling lost?

waspc/cli/src/Wasp/Cli/Command/Start/Db.hs
infomiho: I think I like the idea of a tuple better - the default is defined as a tuple, so why not pass it here as a tuple. I dislike the idea of passing the args directly for reason you mentioned :)

infomiho: Now that I'm seeing it - I don't think we gain much with the tuple in terms of readability or type information - so I've stayed the initial impl. 

I've reformatted the code so the args are on separate lines - it looks a bit better.

web/docs/data-model/databases.md
infomiho: Updated the text to maybe better explain there is a default value vs. customization options.

web/docs/data-model/databases.md
```diff
@@ -89,22 +89,38 @@ In case you might want to connect to the dev database through the external tool
 
 ##### Customising the dev database {#custom-database}
 
-You can also specify a custom Docker image to use for the database with the `--db-image` option. This is particularly useful when you need PostgreSQL with specific extensions (like PostGIS for geographic data, pgvector for embeddings, etc.), or an older version. By default, Wasp uses the latest official [PostgreSQL Docker image](https://hub.docker.com/_/postgres).
+Wasp development database by default uses:
+- [`postgres:18`](https://hub.docker.com/_/postgres/tags?name=18) Docker image,
+- and the `/var/lib/postgresql` volume mount path.
```
cprecioso:
```suggestion
The Wasp development database uses the [PostgreSQL 18 Docker image](https://hub.docker.com/_/postgres/tags?name=18) by default, and will set up its data volumes according to their guidance.
```

web/docs/data-model/databases.md
```diff
@@ -89,22 +89,38 @@ In case you might want to connect to the dev database through the external tool
...
+
+If you need to customise the development database, you can use the following options:
+
+- `--db-image`: Specify a custom Docker image
+    
+  Useful for PostgreSQL extensions or specific versions (for example, PostGIS, pgvector, etc.). 
+
+- `--db-volume-mount-path`: Specify the volume mount path inside the container
+
+  You only need to set this option if your custom `--db-image` is based on **Postgres 17 or older** (check `postgres:15` example below).
```
cprecioso:
```suggestion
  You only need to set this option if your custom `--db-image` is based on **PostgreSQL 17 or older** (check the `postgres:15` example below). 
```

PR #3368, opened 2025-11-13
Mention examples and tutorial in the PR checklist

sodic:
## Description

Replace this message here, and write a high-level overview with any additional
context (motivation, trade-offs, approaches considered, concerns, ...).

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests:

- 📜 Documentation:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -29,14 +29,14 @@ context (motivation, trade-offs, approaches considered, concerns, ...).
 
 - [ ] I tested my change in a Wasp app to verify that it works as intended.
 
-- 🧪 Tests:
+- 🧪 Tests and apps:
 
   - [ ] I added **unit tests** for my change. <!-- If not, explain why. -->
   - [ ] _(if you fixed a bug)_ I added a **regression test** for the bug I fixed. <!-- If not, explain why. -->
   - [ ] _(if you added/updated a feature)_ I added/updated **e2e tests** in `examples/kitchen-sink/e2e-tests`.
   - [ ] _(if you added/updated a feature)_ I updated the **starter templates** in `waspc/data/Cli/templates`, as needed.
   - [ ] _(if you added/updated a feature)_ I updated the **example apps** in `examples/`, as needed.
-    - [ ] _(if you updated `examples/tutorials`)_ I updated the appropriate section of the docs (and vice versa).
+    - [ ] _(if you updated `examples/tutorials`)_ I updated the tutorial in the docs (and vice versa).
 
 - 📜 Documentation:
 


.github/pull_request_template.md
```diff
@@ -33,16 +33,18 @@ context (motivation, trade-offs, approaches considered, concerns, ...).
 
   - [ ] I added **unit tests** for my change. <!-- If not, explain why. -->
   - [ ] _(if you fixed a bug)_ I added a **regression test** for the bug I fixed. <!-- If not, explain why. -->
-  - [ ] _(if you added/updated a feature)_ I added/updated **e2e tests** at `examples/kitchen-sink/e2e-tests`.
-  - [ ] _(if you added/updated a feature)_ I updated the **starter templates** at `waspc/data/Cli/templates`, as needed.
+  - [ ] _(if you added/updated a feature)_ I added/updated **e2e tests** in `examples/kitchen-sink/e2e-tests`.
+  - [ ] _(if you added/updated a feature)_ I updated the **starter templates** in `waspc/data/Cli/templates`, as needed.
+  - [ ] _(if you added/updated a feature)_ I updated the **example apps** in `examples/`, as needed.
+    - [ ] _(if you updated the tutorial)_ Update the appropriate section of the docs.
```
Martinsos: I don't get this one, why is it indented, and what does it mean to update the appropriate section of the docs? If I updated the tutorial, and tutorial is in docs, which other part of the docs would I need to update? Aha ok so I think I get it only no -> you meant to say, if I updated "tutorial app", not "tutorial", right? So if I updated "tutorial app", then I should also update tutorial docs, right?

Ok, then let's do this:
```suggestion
  - [ ] _(if you updated an app in examples/tutorials/)_ I updated tutorial documentation in `web/docs`.
```

But this not enough actually -> what if they did the opposite, they updated tutorial in the docs, but forgot about the actual tutorial app in examples/? Ok I will let you figure it out. But do we really need to even tell them this? This is like telling them that they need to update docs if htey updated the code. Ok I guess we do hah yeah :D. But ttuorial should be more obvious? Hm.

sodic: > you meant to say, if I updated "tutorial app", not "tutorial", right? So if I updated "tutorial app", then I should also update tutorial docs, right?

I thought the indentation (under examples) communicates that:
- First we send the to update the example apps.
- Then, if one of those was the tutorial, we tell them to update the docs too.

But I can make it more explicit.

EDIT: OK, I did it. I think it's enough. We can revisit if it becomes a problem. I'd say the most important thing is that we remember it when reviewing contributor PRs even if they forget.

Martinsos: I want push too much if you both think this is ok, for me this is still quite confusing.
I now get the thing with indentation, but I don't think I would figure it out on my own.

Btw this is about updating documentation, and we actually have section for that just below. So maybe best is to move this there, under Documentation, and it will be a bit clearer, and no indenting.

I would love to see you put some extra effort into this, so that it is done right, but if you really don't think it is useful, ok I accept.

sodic: > I would love to see you put some extra effort into this, so that it is done right, but if you really don't think it is useful, ok I accept.

Ok then, this one's on me then. If it bites us, I'll buy you a coffee :)

Issue #3354, opened 2025-11-12
Release 0.19


cprecioso:

cprecioso:
Released https://github.com/wasp-lang/wasp/releases/tag/v0.19.0

PR #3370, opened 2025-11-13
Update package.json template breaking changes in Wasp AI

cprecioso:
- Part of https://github.com/wasp-lang/wasp/issues/3354

Breaking change in package.json was not done in the Wasp AI templates


infomiho:
Yep, we really could use #2953 not to forget

PR #3114, opened 2025-09-04
Modernize `e2e-tests`

FranjoMindek:
The goal of this PR is to refactor `waspc/e2e-tests` to make the whole project more understandable, easier to use and up to date with current Wasp project. This PR wont try to expand `e2e-tests` with additional tests.

How this is achieved:
- Refactored the all mentions of "Golden tests" to "Snapshot tests", which is a much more known term for the type of testing we are doing here. It is also much more understandable for people without any prior terminology knowledge.
- Changed `waspComplex` snapshot test with `kitchen-sink` snapshot tests. Instead of creating our own complex Wasp application for testing purposes, we use the existing `kitchen-sink` (old `todoApp`) Wasp app. (simplification)
- Added `e2e-tests/README.md` 
- Improved "Snapshot Test"/"Golden Test" definition DX

Along the way we also expanded upon old logic to make it easier to reason when can specific shell commands be executed. Or more specifically in which context.

This is done by splitting the old `ShellCommandContext` into:
- `WaspAppContext`
- `SnapshotTestContext`

Where each context assumes in which environment their commands are run.
This change was also followed by project structure change, where context specific commands exist in their separate folders: `ShellCommands.hs`, `SnapshotTest/ShellCommands.hs`, `WaspApp/ShellCommands.hs`.

Part of: https://github.com/wasp-lang/wasp/issues/1963

PR #3390, opened 2025-11-19
Readd filtering of .tgz files

infomiho:
## Description

We improved the e2e tests in https://github.com/wasp-lang/wasp/issues/3114 and the merge didn't save my changes from https://github.com/wasp-lang/wasp/pull/3078

I redid the changes but with maybe a more DRYed up implementation.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

  - [ ] ~~I added **unit tests** for my change.~~  We don't have unit tests for e2e-tests.

- 📜 Documentation:


waspc/e2e-tests/SnapshotTest.hs
```diff
@@ -198,3 +198,14 @@ defineSnapshotTestCases currentSnapshotDir goldenSnapshotDir currentSnapshotFile
...
+isTgzFile :: FilePath -> Bool
+isTgzFile = (".tgz" `isExtensionOf`)
+
+type FileName = String
+
+createFilenameFilter :: [FileName -> Bool] -> FilePath -> Bool
+createFilenameFilter predicates =
+  \path ->
+    let filename = takeFileName path
+     in all ($ filename) predicates
```
infomiho: I like this more than `createFilenameFilter path` even though it's the same thing, I know.

FranjoMindek: I really feel like is easier to read?
```hs
createFileNameFilter :: [FileName -> Bool] -> FilePath -> Bool
createFileNameFilter predicates filePath = all ($ takeFileName filePath) predicates
```

Anyhow I would use `fileName` instead of `filename`.
So also `createFileNameFilter` instead of `createFilenameFilter`.

infomiho: I like your rewrite, one downside is that `takeFileName` is called for each predicate (I believe). But since these are e2e-tests I don't think we care 😄

PR #3403, opened 2025-11-20
Terminal output is now unstyled (ansi codes) on Windows.

Martinsos:
## Description

Terminal output from `wasp` CLI on windows was messed up due to windows shells not showing ansi codes correctly. This PR makes default behaviour on Windows to not do any styling. It also brings a new env var that users can use to explicitly control this behaviour (styling or not).

## Type of change

- [X] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

  ~- [ ] I added **unit tests** for my change. ~
  ~- [ ] _(if you fixed a bug)_ I added a **regression test** for the bug I fixed. ~
  ~- [ ] _(if you added/updated a feature)_ I added/updated **e2e tests** in `examples/kitchen-sink/e2e-tests`.~
  ~- [ ] _(if you added/updated a feature)_ I updated the **starter templates** in `waspc/data/Cli/templates`, as needed.~
  ~- [ ] _(if you added/updated a feature)_ I updated the **example apps** in `examples/`, as needed.~
    ~- [ ] _(if you updated `examples/tutorials`)_ I updated the tutorial in the docs (and vice versa).~

- 📜 Documentation:

  ~- [ ] _(if you added/updated a feature)_ I **added/updated the documentation** in `web/docs/`.~

  ~- [ ] _(if you did a breaking change)_ I added a step to the current **migration guide** in `web/docs/migration-guides/`.~
  ~- [ ] I **bumped the `version`** in `waspc/waspc.cabal` to reflect the changes I introduced.~


waspc/src/Wasp/Util/Terminal.hs
```diff
@@ -5,10 +5,23 @@ module Wasp.Util.Terminal
...
+import GHC.IO (unsafePerformIO)
+import System.Environment (lookupEnv)
+import Wasp.Util.System (isEnvVarValueTruthy, isSystemWindows)
+
+-- Combo of this function being top level form and NOINLINE ensures that the IO action
+-- executes only once per lifetime of the program.
+{-# NOINLINE isStylingDisabled #-}
+isStylingDisabled :: Bool
+isStylingDisabled =
+  case unsafePerformIO (lookupEnv "WASP_TERMINAL_STYLING") of
```
Martinsos, line 22: Reviewing myself: should we document this somewhere, that we have this functionality now (this env var)? Hm probably should.

waspc/ChangeLog.md
```diff
@@ -26,6 +26,9 @@ Remember to check out the [migration guide](https://wasp.sh/docs/migration-guide
 
 ### 🐞 Bug fixes
 
+- Fixed terminal output on Windows: now it is unstyled on Windows, since ansi codes are often not shown correctly on Win shells.
+  This default behaviour can be explicitly controlled via `WASP_TERMINAL_STYLING=0|1` env var. ([#3403](https://github.com/wasp-lang/wasp/pull/3403))
+- Fixed a type error with the default `NODE_ENV` value in the server env validation schema. ([#3189](https://github.com/wasp-lang/wasp/pull/3189))
```
Martinsos:
```suggestion
```

waspc/src/Wasp/Util/System.hs
```diff
@@ -0,0 +1,17 @@
...
+  )
+where
+
+import qualified System.Info
+
+isSystemWindows :: Bool
+isSystemWindows = System.Info.os == "mingw32"
+
+isSystemMacOS :: Bool
+isSystemMacOS = System.Info.os == "darwin"
```
cprecioso, line 14: idea, weak: if we'd be expecting further development that needs specific platform support, we could use a `data System = Windows | MacOS | Linux` thing that abstracts us from this.

waspc/src/Wasp/Util/Terminal.hs
cprecioso: Yes

cprecioso, line 22: note: I dislike doing an `unsafe` thing for this. It'd make sense that this is something controlled at printing time, when we already have access to IO so that we don't need to do this. I understand this could mean a larger refactor so I'll let you evaluate if that's something we should do.

waspc/src/Wasp/Util/Terminal.hs
```diff
@@ -5,10 +5,23 @@ module Wasp.Util.Terminal
...
+    Nothing -> isSystemWindows
+    Just v -> isEnvVarValueTruthy v
```
cprecioso, line 24: This reads strangely, I'd prefer to have something explicit like

```suggestion
isStylingDisabled =
  case unsafePerformIO (lookupEnv "WASP_TERMINAL_STYLING") of
    Nothing -> defaultValue
    Just v -> isEnvVarValueTruthy v
  
  where
  	defaultValue
  	  | isSystemWindows = False
  	  | otherwise = True
```

I understand it is not as concise (and there's some repetition in the cases) but it is really zippy to read through and understand.

waspc/ChangeLog.md
```diff
@@ -26,6 +26,8 @@ Remember to check out the [migration guide](https://wasp.sh/docs/migration-guide
 
 ### 🐞 Bug fixes
 
+- Fixed terminal output on Windows: now it is unstyled on Windows, since ansi codes are often not shown correctly on Win shells.
+  This default behaviour can be explicitly controlled via `WASP_TERMINAL_STYLING=0|1` env var. ([#3403](https://github.com/wasp-lang/wasp/pull/3403))
```
cprecioso, line 30: We don't usually have multiline items in the changelog. Maybe actually noting this behaviour in the documentation is better that in the changelog.

cprecioso: Also, this is now added in the wrong release, make an "Unreleased" heading

cprecioso: Also, I think this could be better worded (or at least avoid the abbreviations)


```suggestion
- On Windows, we disable terminal styling by default, in order to improve support. You can manually enable it back through [these instructions](link to docs). ([#3403](https://github.com/wasp-lang/wasp/pull/3403))
```

waspc/src/Wasp/Util/System.hs
```diff
@@ -0,0 +1,17 @@
...
+
+isEnvVarValueTruthy :: String -> Bool
+isEnvVarValueTruthy envVarValue = envVarValue `notElem` ["0", "false", "no", "off"]
```
cprecioso, line 17: I'm not sure this is the best place for it but I'm not sure there's a better one so I'll let it go.

cprecioso (changes requested):
Good job tracking this down!

Requesting changes:
- Strongly on adding docs and improving changelog
- Medium on removing the `unsafePerformIO` (though willing to let it go)
- Weak in everything else

Btw, I can see on MS documentation that there should be ANSI support on both ConHost and WinTer: https://learn.microsoft.com/en-us/powershell/module/Microsoft.PowerShell.Core/about/about_ansi_terminals?view=powershell-5.1; I assume I might be missing something.
I'd have liked to see an issue with a reproduction, or at least more information on this same PR with an overview of affected systems and links to documentation supporting this solution.

Martinsos:
Ok, so question is: do we want to turn it off on Windows for real?
What we know is that on Win server 22, in powershell, ansii codes work. In cmd, they don't.
I haven't tried it yet on Win 11.

So it might be we want to leave it on by default, or maybe we do some smart checking for support. But we still could offer the option to not color the output via env flag even if by default we go with it. Btw is there a standardized name for such env flag? Also look into that.

Potentially useful info: https://learn.microsoft.com/en-us/powershell/module/Microsoft.PowerShell.Core/about/about_ansi_terminals?view=powershell-5.1 .

Also, might we worth checking what others re doing.
Here is a quick output from asking Claude Code to analyze `bun` codebase (gotta check how/if it is correct, this is direct from AI):
```
1. Bun Explicitly Enables ANSI on Windows

  Bun proactively enables Virtual Terminal Processing on Windows at startup by calling
  SetConsoleMode() with the ENABLE_VIRTUAL_TERMINAL_PROCESSING flag:

  // src/output.zig:172-218
  _ = c.SetConsoleMode(stdout,
      w.ENABLE_PROCESSED_OUTPUT |
      std.os.windows.ENABLE_VIRTUAL_TERMINAL_PROCESSING |
      w.ENABLE_WRAP_AT_EOL_OUTPUT |
      mode
  );

  This enables ANSI escape sequences on Windows 10+ (including Windows Server 2016+).

  2. No Differentiation Between cmd.exe and PowerShell

  Critical finding: Bun does NOT differentiate between cmd.exe and PowerShell. Both are treated
  identically:

  - Uses uv_guess_handle(fd) on Windows which returns UV_TTY for any console
  - Applies the same console mode flags to both shells
  - No shell-specific detection code exists

  3. Windows Build-Based Color Detection

  Bun detects color support based on Windows build numbers, not shell type:

  // src/js/internal/tty.ts:56-172
  if (process.platform === "win32") {
    const { release } = require("node:os");
    OSRelease = release().split(".");

    if (+OSRelease[0] >= 10) {
      const build = +OSRelease[2];
      if (build >= 14931) return COLORS_16m;  // TrueColor (16M colors)
      if (build >= 10586) return COLORS_256;  // 256 colors
    }
    return COLORS_16;  // 16 colors for older Windows
  }

  Windows Server 2022 has build 20348+, which means it should support full TrueColor (24-bit) in
  both cmd.exe and PowerShell.

  4. ANSI Support Detection Method

  To detect ANSI support, Bun checks:

  // src/js/internal/util/colors.ts:16-22
  shouldColorize(stream) {
    if (process.env.FORCE_COLOR !== undefined) {
      return getColorDepth(process.env) > 2;
    }

    return stream?.isTTY &&
           (stream.getColorDepth?.() > 2 : true);
  }

  5. Environment Variable Support

  Bun respects these environment variables (in priority order):
  - FORCE_COLOR - Force colors (empty/"1"/"true" = 16, "2" = 256, "3" = TrueColor)
  - NODE_DISABLE_COLORS - Disable colors
  - NO_COLOR - Disable colors
  - TERM=dumb - Disable colors
```

Martinsos:
@FranjoMindek and @cprecioso linked https://clig.dev/#output , and this is the most relevant part:

> Disable color if your program is not in a terminal or the user requested it. These things should disable colors:
> 
> - stdout or stderr is not an interactive terminal (a TTY). It’s best to individually check—if you’re piping stdout to another program, it’s still useful to get colors on stderr.
> - The NO_COLOR environment variable is set and it is not empty (regardless of its value).
> - The TERM environment variable has the value dumb.
> - The user passes the option --no-color.
> - You may also want to add a MYAPP_NO_COLOR environment variable in case users want to disable color specifically for your program.
> 
> Further reading: [no-color.org](https://no-color.org/)

PR #3412, opened 2025-11-24
Add error message when user doesn't specify env vars in `wasp build start`

### LINKED ISSUE #3307: `wasp build start`: Show warning message when no environment variables were passed
We got a user struggling with `wasp build start` as they were trying to pass environment variables throught the shell instead of CLI arguments: https://discord.com/channels/686873244791210014/1433736962195329124/1433736962195329124

There's no reasonable situation in which there wouldn't be any environment variables, as at a minimum we'd require the `DATABASE_URL`.

We could detect the case when no environment variables were passed and print a warning message telling users.

sodic:
## Description

Fixes #3307

image

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

      - The current function is not tested and setting up the infrastructure to test it would be too time-consuming (at least for me).

- 📜 Documentation:

### DIFF
--- a/examples/kitchen-sink/package-lock.json
+++ b/examples/kitchen-sink/package-lock.json
@@ -188,7 +188,6 @@
       "integrity": "sha512-e7jT4DxYvIDLk1ZHmU/m/mB19rex9sv0c2ftBtjSBv+kVM/902eh0fINUzD7UwLLNR+jU585GxUJ8/EBfAM5fw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
         "@babel/generator": "^7.28.5",
@@ -1729,7 +1728,6 @@
       "integrity": "sha512-x30GFguInsgt+4z5I4WbkZP2CGpotJMUXy+Gl/aaUjHn2o1DnLYNTA+q9XdYmAQZM8fIIkvUiA2NpgosM3fneg==",
       "hasInstallScript": true,
       "license": "Apache-2.0",
-      "peer": true,
       "engines": {
         "node": ">=16.13"
       },
@@ -2190,7 +2188,6 @@
       "resolved": "https://registry.npmjs.org/@tanstack/react-query/-/react-query-4.42.0.tgz",
       "integrity": "sha512-j0tiofkzE3CSrYKmVRaKuwGgvCE+P2OOEDlhmfjeZf5ufcuFHwYwwgw3j08n4WYPVZ+OpsHblcFYezhKA3jDwg==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@tanstack/query-core": "4.41.0",
         "use-sync-external-store": "^1.2.0"
@@ -2322,7 +2319,8 @@
       "version": "5.0.4",
       "resolved": "https://registry.npmjs.org/@types/aria-query/-/aria-query-5.0.4.tgz",
       "integrity": "sha512-rfT93uj5s0PRL7EzccGMs3brplhcrghnDoV26NqKhCAS1hVo+WdNsPvE/yb6ilfr5hi2MEk6d5EWJTKdxg8jVw==",
-      "license": "MIT"
+      "license": "MIT",
+      "peer": true
     },
     "node_modules/@types/babel__core": {
       "version": "7.20.5",
@@ -2768,7 +2766,6 @@
       "resolved": "https://registry.npmjs.org/@vitest/ui/-/ui-1.6.1.tgz",
       "integrity": "sha512-xa57bCPGuzEFqGjPs3vVLyqareG8DX0uMkr5U/v5vLv5/ZUrBrPL7gzxzTJedEyZxFMfsozwTIbbYfEQVo3kgg==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@vitest/utils": "1.6.1",
         "fast-glob": "^3.3.2",
@@ -3597,7 +3594,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "baseline-browser-mapping": "^2.9.0",
         "caniuse-lite": "^1.0.30001759",
@@ -3925,7 +3921,6 @@
       "integrity": "sha512-/rFeCpNJQbhSZjGVwO9RFV3xPqbnERS8MmIQzCtD/zl6gpJuV/bMLuN92oG3F7d8oDEHHRrujSXNUr8fpjntKw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=18"
       }
@@ -4240,7 +4235,8 @@
       "version": "0.5.16",
       "resolved": "https://registry.npmjs.org/dom-accessibility-api/-/dom-accessibility-api-0.5.16.tgz",
       "integrity": "sha512-X7BJ2yElsnOJ30pZF4uIIDfBEVgF4XEBxL9Bxhy6dnrm5hkzqmsWHGTiHqRiITNhMyFLyAiWndIJP7Z1NTteDg==",
-      "license": "MIT"
+      "license": "MIT",
+      "peer": true
     },
     "node_modules/domexception": {
       "version": "4.0.0",
@@ -4504,7 +4500,6 @@
       "dev": true,
       "hasInstallScript": true,
       "license": "MIT",
-      "peer": true,
       "bin": {
         "esbuild": "bin/esbuild"
       },
@@ -5622,7 +5617,6 @@
       "resolved": "https://registry.npmjs.org/jiti/-/jiti-1.21.7.tgz",
       "integrity": "sha512-/imKNG4EbWNrVjoNC/1H5/9GFy+tqjGBHCaSsN+P2RnPqjsLmv6UD3Ej+Kj8nBWaRAwyk7kK5ZUc+OEatnTR3A==",
       "license": "MIT",
-      "peer": true,
       "bin": {
         "jiti": "bin/jiti.js"
       }
@@ -5647,7 +5641,6 @@
       "resolved": "https://registry.npmjs.org/jsdom/-/jsdom-21.1.2.tgz",
       "integrity": "sha512-sCpFmK2jv+1sjff4u7fzft+pUh2KSUbUrEHYHyfSIbGTIcmnjyp83qg6qLwdJ/I3LpTXx33ACxeRL7Lsyc6lGQ==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "abab": "^2.0.6",
         "acorn": "^8.8.2",
@@ -5841,7 +5834,6 @@
       "integrity": "sha512-P1FlFBGCMPMXu+EGdVD9W4Mjm0DqsusmKgO7Xc33mI5X1bklmsQb0hfzPhXomQr9waWIBDsiOjvr1e6BTaUqpA==",
       "deprecated": "This package has been deprecated. Please see https://lucia-auth.com/lucia-v3/migrate.",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@oslojs/crypto": "^1.0.1",
         "@oslojs/encoding": "^1.1.0"
@@ -5861,6 +5853,7 @@
       "resolved": "https://registry.npmjs.org/lz-string/-/lz-string-1.5.0.tgz",
       "integrity": "sha512-h5bgJWpxJNswbU7qCrV0tIKQCaS3blPDrqKWx+QxzuzL1zGUzij9XCWLrSLsJPu5t+eWA/ycetzYAO5IOMcWAQ==",
       "license": "MIT",
+      "peer": true,
       "bin": {
         "lz-string": "bin/bin.js"
       }
@@ -6919,7 +6912,6 @@
       "resolved": "https://registry.npmjs.org/pg/-/pg-8.16.3.tgz",
       "integrity": "sha512-enxc1h0jA/aq5oSDMvqyW3q89ra6XIIDZgCX9vkMrnz5DFTw/Ny3Li2lFQ+pt3L6MCgm/5o2o8HW9hiJji+xvw==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "pg-connection-string": "^2.9.1",
         "pg-pool": "^3.10.1",
@@ -7144,7 +7136,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "nanoid": "^3.3.11",
         "picocolors": "^1.1.1",
@@ -7340,6 +7331,7 @@
       "resolved": "https://registry.npmjs.org/pretty-format/-/pretty-format-27.5.1.tgz",
       "integrity": "sha512-Qb1gy5OrP5+zDf2Bvnzdl3jsTf1qXVMazbvCoKhtKqVs4/YK4ozX4gKQJJVyNe+cajNPn0KoC0MC3FUmaHWEmQ==",
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "ansi-regex": "^5.0.1",
         "ansi-styles": "^5.0.0",
@@ -7354,6 +7346,7 @@
       "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-5.2.0.tgz",
       "integrity": "sha512-Cxwpt2SfTzTtXcfOlzGEee8O+c+MmUgGrNiBcXnuWxuFJHe6a5Hz7qwhwe5OgaSYI0IJvkLqWX1ASG+cJOkEiA==",
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=10"
       },
@@ -7367,7 +7360,6 @@
       "integrity": "sha512-c5K9MiDaa+VAAyh1OiYk76PXOme9s3E992D7kvvIOhCrNsBQfy2mP2QAQtX0WNj140IgG++12kwZpYB9iIydNQ==",
       "hasInstallScript": true,
       "license": "Apache-2.0",
-      "peer": true,
       "dependencies": {
         "@prisma/engines": "5.19.1"
       },
@@ -7498,7 +7490,6 @@
       "resolved": "https://registry.npmjs.org/react/-/react-19.2.3.tgz",
       "integrity": "sha512-Ku/hhYbVjOQnXDZFv2+RibmLFGwFdeeKHFcOTlrt7xplBnya5OGn/hIRDsqDiSUcfORsDC7MPxwork8jBwsIWA==",
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=0.10.0"
       }
@@ -7508,7 +7499,6 @@
       "resolved": "https://registry.npmjs.org/react-dom/-/react-dom-19.2.3.tgz",
       "integrity": "sha512-yELu4WmLPw5Mr/lmeEpox5rw3RETacE++JgHqQzd2dg+YbJuat3jH4ingc+WPZhxaoFzdv9y33G+F7Nl5O0GBg==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "scheduler": "^0.27.0"
       },
@@ -7536,7 +7526,8 @@
       "version": "17.0.2",
       "resolved": "https://registry.npmjs.org/react-is/-/react-is-17.0.2.tgz",
       "integrity": "sha512-w2GsyukL62IJnlaff/nRegPQR94C/XXamvMWmSHRJ4y7Ts/4ocGRmTHvOs8PSE6pB3dWOrD/nueuU5sduBsQ4w==",
-      "license": "MIT"
+      "license": "MIT",
+      "peer": true
     },
     "node_modules/react-refresh": {
       "version": "0.17.0",
@@ -7713,7 +7704,6 @@
       "resolved": "https://registry.npmjs.org/rollup/-/rollup-4.53.3.tgz",
       "integrity": "sha512-w8GmOxZfBmKknvdXU1sdM9NHcoQejwF/4mNgj2JuEEdRaHwwF12K7e9eXn1nLZ07ad+du76mkVsyeb2rKGllsA==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@types/estree": "1.0.8"
       },
@@ -8545,7 +8535,6 @@
       "resolved": "https://registry.npmjs.org/tailwindcss/-/tailwindcss-3.4.19.tgz",
       "integrity": "sha512-3ofp+LL8E+pK/JuPLPggVAIaEuhvIz4qNcf3nA1Xn2o/7fb7s/TYpHhwGDv1ZU3PkBluUVaF8PyCHcm48cKLWQ==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@alloc/quick-lru": "^5.2.0",
         "arg": "^5.0.2",
@@ -8949,7 +8938,6 @@
       "integrity": "sha512-dZwN5L1VlUBewiP6H9s2+B3e3Jg96D0vzN+Ry73sOefebhYr9f94wwkMNN/9ouoU8pV1BqA1d1zGk8928cx0rg==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "esbuild": "^0.27.0",
         "fdir": "^6.5.0",


waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
```diff
@@ -44,22 +46,23 @@ data BuildStartConfig = BuildStartConfig
...
-          (Server.serverUrlEnvVarName, serverUrl')
-        ]
   userServerEnvVars <-
     liftIO $
       combineEnvVarsWithEnvFiles (Args.serverEnvironmentVariables args) (Args.serverEnvironmentFiles args)
-  serverEnvVars' <- overrideEnvVarsCommand waspServerEnvVars userServerEnvVars
+  userClientEnvVars <-
+    liftIO $
+      combineEnvVarsWithEnvFiles (Args.clientEnvironmentVariables args) (Args.clientEnvironmentFiles args)
+  when (null userClientEnvVars && null userServerEnvVars) $ throwError noEnvVarsSpecifiedErrorMsg
```
sodic: Since:
- Our output is pretty noisy and warnings easily get lost
- And @cprecioso [said](https://github.com/wasp-lang/wasp/issues/3307#issue-3590393647): 
  > There's no reasonable situation in which there wouldn't be any environment variables, as at a minimum we'd require the DATABASE_URL.

I decided to throw an error instead of a warning. This is how it looks like:

image

Potential caveats:
- What if the user wants to test how their system behaves when it lacks all env vars? They'd have to pass in a dummy env var just to satisfy our check.

Reviewers: Let me know what you think. Do you think I should log a warning instead?

waspc/src/Wasp/Util/Terminal.hs
```diff
@@ -4,15 +4,16 @@ module Wasp.Util.Terminal
...
   )
 where
 
 import Data.List (foldl')
 
 -- | Applies the Wasp CLI standardized code styling to a string.
 styleCode :: String -> String
-styleCode = applyStyles [Bold]
+styleCode "" = ""
+styleCode str = ansiEscapeCode ++ getAnsiCodeFor Bold ++ str ++ ansiEscapeCode ++ ansiResetBoldDimCode
```
sodic, line 17: I had to change this because the `applyStyles` command is too agressive - it resets all the styles, not just the ones it applied.

Here's an example, this is what happens when I use the old version of `styleCode` inside an error message:

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
sodic, line 1: The `makeBuildStartConfig` already does env var resolution (i.e., collects env vars from all user-defined sources) and executes within the `Command` monad. These characteristic made it an appropriate place to check for missing environment variables.

But there is a caveat.

The current approach checks _"Does any of the user-provided env var sources include an env variable,"_ which is different from _"Did the user specify any env var soruces."_ For example, if the user does `wasp build start --server-env-file .env.server` but the file `.env.server` is empty, we show them the same error message as if they haven't passed the `--server-env-file` argument.

If that isn't acceptable:
- I could also check whether they specified an env var file and enrich the error message, saying adding something like _"You specified a file called `.env.server` but it is empty"_ 
- I could move the entire check to the calling function. We'd then lose the ability to reason about env vars after their resolution, meaning that `wasp build start --server-env-file .env.server` called with an empty `.env.server` passes our checks and fails during the server's runtime.

The bottom line is, do we care about catching:
1. Users not using the proper flags for specifying env vars.
2. Users ending up with an empty list of environment variables, either by not using proper flags (option **1**) or by specifying empty files.

The current solution checks for **2**, which also captures **1**. But it doesn't differentiate between the two cases.
If we only care about **1**, we should move the check to the `buildStart` function.
If we want to differentiate, I should add an extra check. The downside is that we're complicating the code for a potentially fringe feature.

waspc/src/Wasp/Util/Terminal.hs
```diff
@@ -41,7 +42,7 @@ data Style
 applyStyles :: [Style] -> String -> String
 applyStyles [] str = str
 applyStyles _ "" = ""
-applyStyles styles str = foldl' applyStyle str styles ++ ansiEscapeCode ++ ansiResetCode
+applyStyles styles str = foldl' applyStyle str styles ++ ansiEscapeCode ++ ansiResetAllCode
```
sodic, line 52: As mentioned in the other comment, this command is too agressive. `ansiResetAllCode` removes all the styles including the ones this command didn't apply, which makes it unsuable in all pre-styled text (e.g., error messages).

There are more specific reset codes depending on what you want to reset ([source](https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797#colors--graphics-mode)). Unfortunately, only styles (bold, underline) have specific reset codes. Colors only have "reset all."

Anyway, we should:
- Ideally find a library that does this for us.
- If that's impossible, refactor our internal library:
  - Differentiate between style and color
  - Implement something like `minimalResetFor :: Style -> String` or do something even better to get more control.
  - Encapuslate `ansiEscapeCode`. Callers shouldn't have to import both `getAnsiCodeFor` and `ansiEscapeCode`.  They are never used independently.
 
 I'll create an issue for this.

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
```diff
@@ -87,6 +90,15 @@ makeBuildStartConfig appSpec args projectDir = do
...
 
     serverUrl' = defaultDevServerUrl
+    noEnvVarsSpecifiedErrorMsg =
+      CommandError
+        "No env vars specified"
+        $ "You called "
+          ++ styleCode "wasp build start"
+          ++ " without specifying any environment variables for the started apps (client and server)."
+          ++ "\nThis is a mistake. Run "
+          ++ styleCode "wasp build start --help"
```
sodic: It would be nice if I could draw the help command (i.e., `wasp build start --help`) without hardcoding it, but I don't think I can?

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
```diff
@@ -44,22 +46,23 @@ data BuildStartConfig = BuildStartConfig
...
         ]
-  userClientEnvVars <-
-    liftIO $
-      combineEnvVarsWithEnvFiles (Args.clientEnvironmentVariables args) (Args.clientEnvironmentFiles args)
+      waspServerEnvVars =
+        [ (Server.clientUrlEnvVarName, clientUrl'),
+          (Server.serverUrlEnvVarName, serverUrl')
+        ]
   clientEnvVars' <- overrideEnvVarsCommand waspClientEnvVars userClientEnvVars
+  serverEnvVars' <- overrideEnvVarsCommand waspServerEnvVars userServerEnvVars
```
sodic, line 67: I propose refactoring this part of the function into something like `resolveEnvVars`. This function would then just call:
```hs
(serverEnvVars', clientEnvVars') <- resolveEnvVars (Args.serverEnvironmentVariables args) ...
return $
  BuildStartConfig
    { appUniqueId = appUniqueId',
      buildDir = buildDir',
      serverUrl = serverUrl',
      clientPort = clientPort',
      clientUrl = clientUrl',
      serverEnvVars = serverEnvVars'
      clientEnvVars = clientEnvVars'
    }
```
Should I do that? If yes, should I do it here after we settle on the approach?

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
cprecioso: Seems reasonable to have error and not warning.

> What if the user wants to test how their system behaves when it lacks all env vars?

I don't think we need to support that usecase

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
cprecioso: I don't think we need to complicate ourselves doing (1) explicitly.

In general, IMO `wasp build start` is a convenience feature, so we can do enough to catch the most common cases. The peculiarities of each edge case is something I'd deem out of scope. We can always addd more specific errors if this comes back to bite us.

waspc/src/Wasp/Util/Terminal.hs
cprecioso: Together with #3403, I was thinking that we for sure need to either use a library, or make ourselves some API like

```hs
errorMessage = CliMsg $ Bold "Wasp Server: " <> Yellow appName <> " is not defined"
```

and then we compute the ansi codes at print time (or strip them if needed)

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
cprecioso: Maybe using [`Options.Applicative.execParserPure`](https://hackage-content.haskell.org/package/optparse-applicative-0.19.0.0/docs/src/Options.Applicative.Extra.html#execParserPure) can make the parser output the help text.
IMO this is nice-to-have but not key, so I'd try it and see if it can be easily done, skip it otherwise.

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
cprecioso: No strong opinion, do it if you think is best. The only comment I'd say:

```hs
(serverEnv, clientEnv) <- resolveEnvVars etc
-- I think this too specific needlessly, we might want
-- to reuse the logic somewhere else that doesn't have server/client

-- I' prefer this, feels more reusable
serverEnv <- resolveEnvVars serverInput serverEtc
clientEnv <- resolveEnvVars clientInput clientEtc
```

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
```diff
@@ -44,22 +46,23 @@ data BuildStartConfig = BuildStartConfig
...
-          (Server.serverUrlEnvVarName, serverUrl')
-        ]
   userServerEnvVars <-
     liftIO $
       combineEnvVarsWithEnvFiles (Args.serverEnvironmentVariables args) (Args.serverEnvironmentFiles args)
-  serverEnvVars' <- overrideEnvVarsCommand waspServerEnvVars userServerEnvVars
+  userClientEnvVars <-
+    liftIO $
+      combineEnvVarsWithEnvFiles (Args.clientEnvironmentVariables args) (Args.clientEnvironmentFiles args)
+  when (null userClientEnvVars && null userServerEnvVars) $ throwError noEnvVarsSpecifiedMsg
```
cprecioso, line 57: something to think about:
if the user passes client envs but not server envs, that's incorrect. the opposite is probably fine. maybe we serve our users better by only checking server envs.
I'm open to being convinced otherwise.

waspc/src/Wasp/Util/Terminal.hs
cprecioso: It seems [`prettyprinter`](https://hackage.haskell.org/package/prettyprinter-1.7.1) and [`prettyprinter-ansi-terminal`](https://hackage.haskell.org/package/prettyprinter-ansi-terminal) is what `optparse-applicative` uses, we might want to reuse too.

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
sodic: Hm, I didn't want to bake too many assumptions about how other parts of Wasp, or their app and its env vars work. Perhaps they have a client env var that's essential as well?

So my idea was to verify _"Did the user manage to get at least one env var to one of the apps?"_
  - If the answer is "no," then something is wrong and we warn them.
  - If the answer is "yes," then they know what to do, they know about the flags, and knowing about one flag makes the other obvious. If they forgot some (or all) necessary variables, the runtime will warn them.

We muddied the waters here by verifying env vars after resolution (discussion at https://github.com/wasp-lang/wasp/pull/3412#discussion_r2556544262), meaning we do perform _a part_ of the runtime's job, but I don't think we should do more than that.

waspc/src/Wasp/Util/Terminal.hs
Martinsos: True! I think when I was implementing it, I was just using the general reset code. Which now that you say this, doesn't sounds great hah, because you can't envelop one style with another style.

I see you instead used a specific reset (I dind't even know those exist, awesome).
Why not, while at it, fix the `applyStyles` function so it resets specific ansii codes, and fix it for good, instead of doing this exception here? Sounds like a simple fix?

waspc/src/Wasp/Util/Terminal.hs
Martinsos: We could define default color that is used to reset colors, and then specific resets for styles, that would probably work well enough for now.

I remember I was looking at other libraries, but I think I found them complicated or didn't like the API, so I did it manually, that it is simple enough. But I am ok for using library of course.

@cprecioso you are suggesting we basically decorate the text with metadata but then later decide how to interpret it (style it, not style it, ...).
I like that, it sounds very nice. I am only wondering if it might complicate things too much, since these are not strings then, and we can't use normal functions to operate on them or print them. But maybe that is ok: because once we add ansi codes in, they are also not normal strings hah, so better not to be able to just simply operate on them. Yeah, I like this, it is semantically correct.

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
Martinsos: What does it mean you could "draw" it? Aha, you mean refer to it somehow, so if its exact invocation changes, this is still correct?

Hm, yeah that would be really nice. We could certainly refactor our system to support this, but probably not worth it now.

waspc/ChangeLog.md
```diff
@@ -6,6 +6,10 @@
 
 - Updated GitHub social login docs to use GitHub apps over OAuth apps per GitHub recommendation. ([#3277](https://github.com/wasp-lang/wasp/issues/3277))
 
+### 🔧 Small improvements
+
+- `wasp build start` now errors when users forget to specify environment variables ([#3412](https://github.com/wasp-lang/wasp/pull/3412))
```
Martinsos: > errors

I thought we are just warning them? Are we sure we want to actually error, is that a bit too strong?

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
```diff
@@ -87,6 +90,15 @@ makeBuildStartConfig appSpec args projectDir = do
     clientUrl' = getDefaultDevClientUrl appSpec
 
     serverUrl' = defaultDevServerUrl
+    noEnvVarsSpecifiedMsg =
+      CommandError
+        "No env vars specified"
+        $ "You called "
+          ++ styleCode "wasp build start"
+          ++ " without specifying any environment variables for the started apps (client and server)."
+          ++ "\nThis is a mistake. Run "
```
Martinsos: > This is a mistake.

Hah this made me laugh. Sounds very short/harsh/robotic somehow.

Couldn't we offer some more information? I think we should explain why this is a mistake -> obviously they don't know, and while just telling them this is a mistake is a first step, but a bit extra info would be nice (even if full explanation is in `wasp build start --help`).

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
```diff
@@ -87,6 +90,15 @@ makeBuildStartConfig appSpec args projectDir = do
     clientUrl' = getDefaultDevClientUrl appSpec
 
     serverUrl' = defaultDevServerUrl
+    noEnvVarsSpecifiedMsg =
+      CommandError
```
Martinsos, line 96: So we are going for error, not for warning? I saw you say in another comment it will be warning + it makes sense to me it is warning. I can see it being an error though if we are 100% sure they might not have done this on purpose (are we?).

Martinsos (commented):
Quick review as per invite, I will let @cprecioso do the main reviewing/approving, I just commented on couple of things I might be relevant for.

waspc/src/Wasp/Util/Terminal.hs
sodic: I tried it, it was too fiddly to fix and test (some do have their own reset codes, others don't, different treatment for colors and styles, etc.). Details explained here: https://github.com/wasp-lang/wasp/pull/3412#discussion_r2556580162. And I'll create an issue.
 
Also, I'd probably rather use a library for it then implement it here, and that's a seprate discussion which also kind of started in that other thread.

waspc/ChangeLog.md
sodic: Discussed [here](https://github.com/wasp-lang/wasp/pull/3412#discussion_r2556426281). Resolve this thread please and reply there if necessary :)

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
sodic: Discussed [here](https://github.com/wasp-lang/wasp/pull/3412#discussion_r2556426281). Resolve this thread please and reply there if necessary :)

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
sodic: Hm, what do you suggest we say? I personally wouldn't add anything else but am open to suggestions and will very likely accept them.

> obviously they don't know

I'd argue that they _do know_ that not sending env vars is a mistake. They just maybe don't know how to send them. Parts of this are already discussed here: https://github.com/wasp-lang/wasp/pull/3412#discussion_r2556544262

sodic: Btw, I managed to refer to the help command's text and now this is my output:
image

So you can suggest improvements on that. I can't change anything that comes after "Available options." That comes from the parser.

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
sodic: I couldn't do the second thing because I'd then have no place to throw the error. I left it as is for the time being.

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
```diff
@@ -21,8 +22,9 @@ import qualified StrongPath as SP
 import Wasp.AppSpec (AppSpec)
 import qualified Wasp.AppSpec.Valid as ASV
 import Wasp.Cli.Command (Command, CommandError (CommandError))
-import Wasp.Cli.Command.BuildStart.ArgumentsParser (BuildStartArgs)
+import Wasp.Cli.Command.BuildStart.ArgumentsParser (BuildStartArgs, buildStartArgsParser)
```
sodic, line 25: This creates some coupling but I've decided it's alright. Let me know if you disagree.

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
```diff
@@ -88,6 +92,13 @@ makeBuildStartConfig appSpec args projectDir = do
     clientUrl' = getDefaultDevClientUrl appSpec
 
     serverUrl' = defaultDevServerUrl
+    noEnvVarsSpecifiedMsg =
+      CommandError
+        "No env vars specified"
+        $ "You called "
+          ++ styleCode "wasp build start"
+          ++ " without specifying any environment variables for the started apps (client and server).\n"
+          ++ getParserHelpMessage buildStartArgsParser
```
sodic, line 106: This prints out the entire help text (see [here](https://github.com/wasp-lang/wasp/pull/3412#discussion_r2589380739)). So if we add anything other than env var options, it might become noisy.

I decided this was also OK and that we'd cross that bridge when we get there. Let me know if you disagree.

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
cprecioso: I see. I also don't like the coupling but I don't think there's a practical solution otherwise. LGTM

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
```diff
@@ -90,15 +92,13 @@ makeBuildStartConfig appSpec args projectDir = do
...
+    noEnvVarsSpecifiedMsg =
       CommandError
         "No env vars specified"
         $ "You called "
           ++ styleCode "wasp build start"
-          ++ " without specifying any environment variables for the started apps (client and server)."
-          ++ "\nThis is a mistake. Run "
-          ++ styleCode "wasp build start --help"
-          ++ " to see how to to specify them."
+          ++ " without specifying any environment variables for the started apps (client and server).\n"
```
cprecioso: I see there was a back and forth here between you and Martin. I liked your initial message better.

If we want to keep this variant, I think we're missing something like

```
You called wasp build start without specifying any environment variables for the started apps (client and server). Please set at least one environment variable. 

Usage:
[... the help message ...]
```

If we wanted to be extra helpful (I leave it to your decision if this is acceptable coupling):

```
You called wasp build start without specifying any environment variables for the started apps (client and server). Please set at least one environment variable (usually DATABASE_URL, and JWT_SECRET).

Find more information about required environment variables at https://wasp.sh/docs/project/env-vars.

Usage:
[... the help message ...]
```

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
```diff
@@ -120,8 +120,11 @@ overrideEnvVarsCommand forced existing =
...
     Right combined -> return combined
 
-combineEnvVarsWithEnvFiles :: [EnvVar] -> [EnvVarFileArgument] -> IO [EnvVar]
-combineEnvVarsWithEnvFiles pairs files = do
-  pairsFromFiles <- mapM readEnvVarFile files
-  let allEnvVars = pairs <> concat pairsFromFiles
+combineEnvVarsWithEnvFiles :: [EnvVar] -> [FilePathArgument] -> IO [EnvVar]
+combineEnvVarsWithEnvFiles inlineEnvVars files = do
+  envVarsFromFiles <- mapM readEnvVarsFromFile files
+  let allEnvVars = inlineEnvVars <> concat envVarsFromFiles
```
cprecioso, line 131: Nice

waspc/cli/src/Wasp/Cli/Util/Parser.hs
```diff
@@ -1,22 +1,48 @@
...
+withArguments cmdName parser onSuccess args =
+  case parseArguments cmdName parser args of
+    (ArgsParsed result) -> onSuccess result
+    (ParseFailure helpMessage) -> throwError $ CommandError "Parsing arguments failed" helpMessage
+    (ShowHelp helpMessage) -> cliSendMessageC $ Msg.Info helpMessage
+
+getParserHelpMessage :: Opt.Parser a -> String
+getParserHelpMessage =
+  Opt.Help.renderHelp (Opt.prefColumns parserPreferences) . parserHelp Opt.defaultPrefs
+
```
cprecioso, line 28: I'm not looking deeper into this, I trust your investigation that this is the best way to do it ✌️

waspc/ChangeLog.md
```diff
@@ -2,9 +2,29 @@
...
+
+## 0.19.1
+
+### 🐞 Bug fixes
+
+- Fixed logic for getting Railway service domains in `wasp deploy railway` with Railway CLI 4.11.2 and greater. ([#3454](https://github.com/wasp-lang/wasp/pull/3454))
+
+### 🔧 Small improvements
+
+- `wasp build start` now errors when users forget to specify environment variables ([#3412](https://github.com/wasp-lang/wasp/pull/3412))
```
cprecioso: I'd have it a bit less robotic, but feel free to disregard

```suggestion
- `wasp build start` now shows an error if you forget to specify any environment variables ([#3412](https://github.com/wasp-lang/wasp/pull/3412))
```

cprecioso (approved):
LGTM, just left a couple of comments on phrasing. Leaving it approved in case you want to go ahead when you're done.

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
sodic: This is great, I'll link the docs. Never mind the coupling. I might change "please" to "you must" because they truly must :)

waspc/cli/src/Wasp/Cli/Util/Parser.hs
sodic: Unfortunately it seems so.

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
sodic: This is what I did:
image

I can't do "usage" instead of "available options" - that's part of the parser's help text.

waspc/cli/src/Wasp/Cli/Command/BuildStart/Config.hs
sodic: I found a simpler way and did it (it's still ugly though, but less).

cprecioso (approved):
Approving the new message ✅

PR #3414, opened 2025-11-24
Fix error swallowing in db reset and db seed

### LINKED ISSUE #3348: Wasp CLI Command `wasp db seed` does not properly propagate the exit codes
When failing inside of the Typescript portion of the `wasp db seed` command execution, we return a exit code of 0 instead of 1.

### How I initially found out about it:

By calling `wasp db seed` with a valid seed name e.g. `devSeedSimple`
```ts
// main.wasp  
db: {
    seeds: [
      import { devSeedSimple } from "@src/features/db/seeds",
      import { prodSeed } from "@src/features/db/seeds"
    ],
    prismaSetupFn: import { setUpPrisma } from "@src/features/db/prisma",
  },
```
But not actually exporting the declared function, we error out on `wasp db seed` but return exit code 0:

Image

### Easy reproduction

Same can be done by throwing any `Error` inside of the seed function:
```ts
export const devSeedSimple: DbSeedFn = async (prismaClient) => {
  throw Error("AAAA");
}
```

Image

### Extra context I checked for now

Calling the seed directly from `package.json` script (`npm run db-seed`) inside of `.wasp/out/server` returns 1 properly:

Image

sodic:
## Description

Fixes #3348.

## Type of change

- [x] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

    - The modified function isn't unit-tested. Adding tests is tricky (thanks to IO) and would blow through the estimate.
     - Same reason as above.

- 📜 Documentation:

    - I updated it in a different PR that's getting merged soon, no reason to cause conflicts.

### DIFF
--- a/waspc/ChangeLog.md
+++ b/waspc/ChangeLog.md
@@ -6,6 +6,10 @@
 
 - Updated GitHub social login docs to use GitHub apps over OAuth apps per GitHub recommendation. ([#3277](https://github.com/wasp-lang/wasp/issues/3277))
 
+### 🔧 Small improvements
+
+- Wasp returns a proper exit code when `wasp db seed` and `wasp db reset` fail. ([#3414](https://github.com/wasp-lang/wasp/pull/3414))
+
 ## 0.19.0
 
 ### ⚠️ Breaking Changes


waspc/cli/src/Wasp/Cli/Command/Db/Seed.hs
```diff
@@ -36,7 +36,7 @@ seed maybeUserProvidedSeedName = do
   cliSendMessageC $ Msg.Start $ "Running database seed " <> nameOfSeedToRun <> "..."
 
   liftIO (dbSeed genProjectDir nameOfSeedToRun) >>= \case
-    Left errorMsg -> cliSendMessageC $ Msg.Failure "Database seeding failed" errorMsg
```
sodic, line 39: This is where the old code was swallowing errors. We'd just output a failure message and continue as if nothing happened.

waspc/cli/src/Wasp/Cli/Command/Db/Reset.hs
```diff
@@ -21,5 +22,5 @@ reset = do
   cliSendMessageC $ Msg.Start "Resetting the database..."
 
   liftIO (dbReset genProjectDir) >>= \case
-    Left errorMsg -> cliSendMessageC $ Msg.Failure "Database reset failed" errorMsg
```
sodic, line 24: This command had the same problem. I tested to confirm happens and fixed it.

Issue #3404, opened 2025-11-20
Make CLI e2e tests work for Windows


Martinsos:
In our [CI (waspc tests)](https://github.com/wasp-lang/wasp/blob/main/.github/workflows/ci-waspc-test.yaml), we do have windows in the testing matrix, and we do build wasp on windows, and we run unit tests on windows.

But, we don't run Wasp CLI e2e tests on windows. CI actually runs them for windows builder also, but inside the Wasp CLI e2e tests code (at the top level, in Haskell code), we have a check that doesn't run anything if we are on Windows, since tests currently crash on Windows due to heavy reliance on posix commands.

What we need to do is refactor these tests so they can run on Windows.

Likely the best solution is to change ShellCommandBuilder to not be ReaderT, but ReaderT + IO, and instead of concatenating strings that are shell commands, we would be combining IO actions that are platform agnostic (e.g. we would replace `mkdir` with appropriate Haskell's function for creating a directory, that is platform agnostic). This will both make the code better typed/nicer and make e2e tests work on Windows.

I did a first attempt at this at branch `waspc-e2e-tests-win`, but it is purely vibe-coded output from my chat with Claude Code, I haven't even looked at what it output yet. If I take this on, I will probably start from there, as it this task is mostly about translating shell commands into Haskell's functions + some refactoring, and AI can give a good start with that.

cprecioso:
Let's check, since we made `bash` a required dev tool for developing on Wasp, on Windows too, if the current tests work already.

PR #3415, opened 2025-11-24
Smoke test `wasp ai:new` in CI

### LINKED ISSUE #2953: Automatically test Wasp AI

sodic:
## Description

Fixes #2953.

Some decisions I made, feel free to challenge:
 - We're only testing whether `wasp new:ai` exits successfully and generates the project. We're not looking at the project:
   - This was decided in the issue. Testing much more than this isn't worth it. We just want to catch big bugs like [this one](https://github.com/wasp-lang/wasp/issues/2951) before the release.
- Wasp AI test is its own thing:
  - It's not part of Snapshot tests because we don't care about the generated context. Adapting snapshot tests to work with this would be both costly and incorrect, as they wouldn't be testing snapshots.
  - It's not a part of starters tests for a similar reason. Starter tests are headless test verifying the apps work correctly. In this case, we only care about the exit code of `wasp new:ai`.
  - We don't yet have a platform for running these kinds of tests (other than CI). I believe @FranjoMindek is working on one but the PR is not yet ready.
- Testing steps are only listed in CI. There is no local `./run something` way to run this test. Reasons:
  - The local test would be cumbersome to set up (where to put the generated project, how to clean it up).
  - The "test" is simple enough for us to run manually.
  - We don't plan to often run the test locally.
  - Even if we're running the test locally, we'll probably want to run all the steps manually anyway.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:


.github/workflows/ci-waspc-test.yaml
```diff
@@ -17,6 +18,7 @@ jobs:
     name: Test waspc
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
```
sodic, line 20: Doesn't cancel all jobs if one of them fails.

.github/workflows/ci-waspc-test.yaml
```diff
@@ -67,6 +69,12 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
+      - name: Debug npm PATH on Windows
+        if: runner.os == 'Windows'
+        run: |
+          which npm || echo "npm not found in PATH"
+          echo "PATH: $PATH"
```
sodic: I will remove this debug step before merging.

I put it here because `wasp new:ai` fails to find NPM on Windows. even though we installed it. Here's an example of a failing action: https://github.com/wasp-lang/wasp/actions/runs/19654022582/job/56286837757

This debug steps shows that `bash` successfully finds NPM on Windows, meaning it only fails when Wasp calls it. @Martinsos This sounds like something you dealt with recently (executing commands and resolving `PATH` on Windows). Can you help?

.github/workflows/ci-waspc-test.yaml
```diff
@@ -89,13 +97,18 @@ jobs:
       - name: Build wasp code
         run: cabal build all
 
+      - name: Run `wasp new:ai` smoke test
+        if: runner.os != 'Windows' # Skip windows for now
```
sodic: I should remove this before merge too. Details about the Windows problem are in the other comment.

.github/actions/setup-haskell/action.yaml
```diff
@@ -15,7 +15,7 @@ inputs:
     description: |
       The version of Cabal to install.
     required: false
-    default: "3.10.2.0"
+    default: "3.12.1.0"
```
sodic, line 18: I updated cabal to the recommended version:
- It matches our local setup.
- It allows me to use `--project-dir`.

.github/workflows/ci-waspc-test.yaml
```diff
@@ -6,17 +6,22 @@ on:
...
 env:
   WASP_TELEMETRY_DISABLE: 1
+  OPENAI_API_KEY: ${{ secrets.OPENAI_API_TESTING_KEY }}
+  # NOTE: Duplicated in defaults->run->working-directory
+  PROJECT_DIR: ${{ github.workspace }}/waspc
 
 defaults:
   run:
     shell: bash
+    # NOTE: Duplicated in env.PROJECT_DIR
```
sodic: I unfortunately can't use this value in env vars or vice versa. We must duplicate it so I left a note.

.github/workflows/ci-waspc-test.yaml
```diff
@@ -89,13 +100,18 @@ jobs:
       - name: Build wasp code
         run: cabal build all
 
+      - name: Run `wasp new:ai` smoke test
+        if: runner.os != 'Windows' # Skip windows for now
+        working-directory: ${{ runner.temp }}
```
sodic: I wanted to execute this test somewhere outside the project where there's no context meddling with the test.

.github/workflows/ci-waspc-test.yaml
```diff
@@ -89,13 +100,18 @@ jobs:
...
+          APP_TITLE: todoApp
+          APP_DESCRIPTION: "A simple todo app with one main page that lists all the tasks. User can create new tasks by providing their description, toggle existing ones, or edit their description. User owns tasks. User can only see and edit their own tasks. Tasks are saved in the database."
+          CONFIG_JSON: "{}"
+
       - name: Run unit tests
         run: cabal test wasp-cli-tests waspc-tests waspls-tests
 
-      - name: Install `wasp-cli`
-        if: matrix.os == 'ubuntu-22.04'
-        run: cabal install
```
sodic, line 97: This was a leftover from when we were doing headless tests.

.github/workflows/ci-waspc-test.yaml
Martinsos: This is a well known Wasp problem, and the main obstacle where we stopped with Win support for Wasp: Wasp doesn't correctly resolve npm/node/npx/... on Windows, and errors when it tries to call them. So if you hit that, nothing new. And this is exactly the thing I am solving in the PR I am working on, and good news is, it is about to get merged, and then this will work. So it isn't anything specific to your PR here, in that case, and you are all good. Here is PR if you are interested: https://github.com/wasp-lang/wasp/pull/3258 .

Ok but you hit this because you wanted to also run the test on Win and it didn't work.
So what shall we do now? I suggest you merge comment out the test and leave a mention here https://github.com/wasp-lang/wasp/issues/3404 to uncomment it.

That said, I have a new question: isn't this just another e2e test for Wasp CLI? If so, we have a mechanism for doing those, in `e2e-tests`, and you should probably not have this test as a standalone test, but instead as just another test inside of `e2e-tests`. @FranjoMindek knows the most about the whole thing, so you can ask him for expert opinion, but I also know enough to answer most of the stuff. But yeah I think you will want to put this test there, and then you won't even have do add anything to the CI here, since those tests are already ran by CI. And also they already skip Windows. And we already have an issue to uncomment that once Windows work. So if you do this, you don't have do to anything in regards to windows.

sodic: This was covered in the description and in discord :)

Conclusion: it stays as is and I'll create an issue blocked by Franjo's issue.

cprecioso (changes requested):
General approach seems fine, but won't review in-depth yet, as I have one big blocker:

`ci-waspc-test` is already quite big and monolithic. We'd be better served by extracting steps from it to their own files, rather than adding steps here.

So I'd suggest to create a new workflow file altogether, and call it from the `ci.yaml`. You can look at the testing deploy or testing starters worflows to copy any needed boilerplate.

**EDIT:** I mean to create a new workflow file _for this functionality only_, no need to undo the whole big file here.

.github/workflows/ci-wasp-ai-test.yaml
```diff
@@ -0,0 +1,36 @@
...
+        with:
+          node-version: lts/*
+
+      - name: Install `wasp-cli`
+        working-directory: waspc
+        run: ./run install
+
+      - name: Run `wasp new:ai` smoke test
+        working-directory: ${{ runner.temp }}
+        run: cabal --project-dir=${{ env.PROJECT_DIR }} run wasp-cli -- new:ai "$APP_TITLE" "$APP_DESCRIPTION" "$CONFIG_JSON"
```
cprecioso, line 35: Why not merge both of these here into one? No need to install then.

```yaml
run: ./run wasp-cli new:ai blah blah
```

.github/workflows/ci-wasp-ai-test.yaml
```diff
@@ -0,0 +1,36 @@
...
+        env:
+          APP_TITLE: todoApp
+          APP_DESCRIPTION: "A simple todo app with one main page that lists all the tasks. User can create new tasks by providing their description, toggle existing ones, or edit their description. User owns tasks. User can only see and edit their own tasks. Tasks are saved in the database."
+          CONFIG_JSON: "{}"
```
cprecioso, line 43: I like this being extracted for easier modification!

.github/workflows/ci-wasp-ai-test.yaml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Install `wasp-cli`
+        working-directory: waspc
+        run: ./run install
+
+      - name: Run `wasp new:ai` smoke test
+        working-directory: ${{ runner.temp }}
+        run: cabal --project-dir=${{ env.PROJECT_DIR }} run wasp-cli -- new:ai "$APP_TITLE" "$APP_DESCRIPTION" "$CONFIG_JSON"
+        env:
+          APP_TITLE: todoApp
+          APP_DESCRIPTION: "A simple todo app with one main page that lists all the tasks. User can create new tasks by providing their description, toggle existing ones, or edit their description. User owns tasks. User can only see and edit their own tasks. Tasks are saved in the database."
```
cprecioso: nitpick, further idea:


```suggestion
          APP_DESCRIPTION: >
            A simple todo app with one main page that lists all the tasks.
            User can create new tasks by providing their description, toggle
            existing ones, or edit their description. User owns tasks. User
            can only see and edit their own tasks. Tasks are saved in the
            database.
```

I eyeballed the line length, make sure it's the correct one if you take this suggestion 😁

.github/workflows/ci-wasp-ai-test.yaml
cprecioso, line 1: Only two small things
- check that the external actions we use are updated (they're now on v6 IIRC)
- Add a note saying why we don't try to compiler or run the created app

cprecioso (approved):
LGTM! Thanks for splitting it 🤗

I added some suggestions, left the PR approved already since they're simple things.

.github/workflows/ci-waspc-test.yaml
Martinsos: Great! RAW

Martinsos:
I see @cprecioso has a grasp on it, so leaving the main reviewing to him, I shall retire myself.

.github/workflows/ci-wasp-ai-test.yaml
sodic: I wanted to run `wasp new:ai` in a folder outside the project to avoid accidentally relying on some hidden context. Also, it seemed neater to generate temporary stuff in a temp folder.

I added a note that clarifies this.

cprecioso: ahh got it 🫡

.github/workflows/ci-wasp-ai-test.yaml
sodic: I updated them here: https://github.com/wasp-lang/wasp/commit/8d49cbb518ce923f0c8ecf02a13f545ca7459a19

PR #3440, opened 2025-11-26
Do not fail on `--help` argument

### LINKED ISSUE #3438: `wasp build start --help` says "Parsing failed" and returns non-zero exit code
When you run `wasp build start --help`, the command says that "Parsing arguments failed" and returns an exit code of `1`, but does mention `--help` as a valid argument.

We should not say "parsing failed" and give a `0` error code.

```
$ wasp build start --help


❌ --- [Error] Parsing arguments failed: ------------------------------------------

Usage: wasp build start [-c|--client-env NAME=VALUE] 
                        [--client-env-file FILE_PATH] 
                        [-s|--server-env NAME=VALUE] 
                        [--server-env-file FILE_PATH]

Available options:
  -c,--client-env NAME=VALUE
                           Set an environment variable for the client (can be
                           used multiple times)
  --client-env-file FILE_PATH
                           Load environment variables for the client from a file
                           (can be used multiple times)
  -s,--server-env NAME=VALUE
                           Set an environment variable for the server (can be
                           used multiple times)
  --server-env-file FILE_PATH
                           Load environment variables for the server from a file
                           (can be used multiple times)
  -h,--help                Show this help text
```

cprecioso:
- Fixes #3438

## Description

We are exiting the CLI with a failure when using the `--help` flags in Options.Applicative parsers.

This is because, in that case, the library return a `Failure` but with an `ExitCode.ExitSuccess` inside of it. We were not inspecting it, and just throwing an error even when it didn't need to be one.

Because we need to short-circuit the full execution, but without throwing an error; and `parseArguments` was being used in the exact same way in every caller; I transformed the `parseArguments` function into a `withArguments` one that wraps the `Command` and can decide whether to run it or not.

## Type of change

- [x] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

  > [!NOTE]
  > I didn't add any tests as we don't have testing at such high-level in the CLI itself, we'd end up testing `Options.Applicative` itself.

- 📜 Documentation:

### DIFF
--- a/waspc/cli/src/Wasp/Cli/Util/Parser.hs
+++ b/waspc/cli/src/Wasp/Cli/Util/Parser.hs
@@ -15,13 +15,9 @@ import qualified Wasp.Message as Msg
 withArguments :: String -> Opt.Parser a -> (a -> Command ()) -> Arguments -> Command ()
 withArguments cmdName parser onSuccess args =
   case parseArguments cmdName parser args of
-    (ArgsParsed result) -> do
-      onSuccess result
-    (ParseFailure helpMessage) -> do
-      throwError $ CommandError "Parsing arguments failed" helpMessage
-    (ShowHelp helpMessage) -> do
-      cliSendMessageC $ Msg.Info helpMessage
-      return ()
+    (ArgsParsed result) -> onSuccess result
+    (ParseFailure helpMessage) -> throwError $ CommandError "Parsing arguments failed" helpMessage
+    (ShowHelp helpMessage) -> cliSendMessageC $ Msg.Info helpMessage
 
 data ArgsParseResult args
   = ArgsParsed args
@@ -36,9 +32,7 @@ parseArguments cmdName parser args =
       error $ "Completion invoked when parsing '" <> cmdName <> "', but this should never happen"
     (Opt.Failure failure) ->
       case Opt.execFailure failure cmdName of
-        (help, EC.ExitSuccess, _) ->
-          ShowHelp $ show help
-        (help, EC.ExitFailure _, _) ->
-          ParseFailure $ show help
+        (help, EC.ExitSuccess, _) -> ShowHelp $ show help
+        (help, EC.ExitFailure _, _) -> ParseFailure $ show help
   where
     parserInfo = Opt.info (parser <**> Opt.helper) Opt.fullDesc


waspc/src/Wasp/Psl/Parser/Common.hs
```diff
@@ -18,7 +18,6 @@ module Wasp.Psl.Parser.Common
   )
 where
 
-import Control.Applicative (liftA2)
```
cprecioso: Compiler was giving me a warning

sodic: Yeah, this is now in global scope.

waspc/cli/src/Wasp/Cli/Util/Parser.hs
```diff
@@ -1,22 +1,47 @@
...
+withArguments cmdName parser onSuccess args =
+  dispatch $ parseArguments cmdName parser args
+  where
+    dispatch (Success result) = do
+      onSuccess result
+    dispatch (Failure _ helpMessage) = do
+      throwError $ CommandError "Parsing arguments failed" helpMessage
+    dispatch (Stop helpMessage) = do
+      cliSendMessageC $ Msg.Info helpMessage
+      return ()
```
sodic: Seems like we can simplify this:
```hs
withArguments :: String -> Opt.Parser a -> (a -> Command ()) -> Arguments -> Command ()
withArguments cmdName parser onSuccess args = case parseArguments cmdName parser args of
  (Success result) -> onSuccess result
  (Failure _ helpMessage) -> throwError $ CommandError "Parsing arguments failed" helpMessage
  (Stop helpMessage) -> cliSendMessageC $ Msg.Info helpMessage
```

waspc/cli/src/Wasp/Cli/Util/Parser.hs
```diff
@@ -1,22 +1,47 @@
...
+
+data ArgsResult a
+  = Success a
+  | Failure Int String
+  | Stop String
```
sodic:
```hs
data ParsingResult a  -- or even 'args' instead of just a
  = ParsedArguments a
  | ParsingFailed Int String
  | HelpCalled String  -- or ShowHelp or HelpRequested, something along those lines
```
I think this is a little more informative. No need to copy it verbatim, just consider the general idea.

waspc/cli/src/Wasp/Cli/Util/Parser.hs
```diff
@@ -1,22 +1,47 @@
...
+parserResultToArgsResult _ (Opt.Success success) = Success success
+parserResultToArgsResult cmdName (Opt.CompletionInvoked _) =
   error $ "Completion invoked when parsing '" <> cmdName <> "', but this should never happen"
-parserResultToEither cmdName (Opt.Failure failure) = Left $ show help
+parserResultToArgsResult cmdName (Opt.Failure failure) =
+  failureToArgsResult $ Opt.execFailure failure cmdName
   where
-    (help, _, _) = Opt.execFailure failure cmdName
+    failureToArgsResult (help, EC.ExitSuccess, _) = Stop $ show help
+    failureToArgsResult (help, EC.ExitFailure exitCodeInt, _) = Failure exitCodeInt $ show help
```
sodic: We can simplify this as well:
```hs
parserResultToArgsResult cmdName (Opt.Failure failure) =
  case Opt.execFailure failure cmdName of
    (help, EC.ExitSuccess, _) -> Stop $ show help
    (help, EC.ExitFailure exitCodeInt, _) -> Failure exitCodeInt $ show help
```

cprecioso:
Ready

waspc/cli/src/Wasp/Cli/Util/Parser.hs
```diff
@@ -1,22 +1,44 @@
...
+withArguments :: String -> Opt.Parser a -> (a -> Command ()) -> Arguments -> Command ()
+withArguments cmdName parser onSuccess args =
+  case parseArguments cmdName parser args of
+    (ArgsParsed result) -> do
+      onSuccess result
+    (ParseFailure helpMessage) -> do
+      throwError $ CommandError "Parsing arguments failed" helpMessage
+    (ShowHelp helpMessage) -> do
+      cliSendMessageC $ Msg.Info helpMessage
+      return ()
```
sodic: Why the `return ()`?

waspc/cli/src/Wasp/Cli/Util/Parser.hs
```diff
@@ -1,22 +1,44 @@
...
-parseArguments :: String -> Opt.Parser a -> Arguments -> Either String a
-parseArguments cmdName parser args =
-  parserResultToEither cmdName $
-    Opt.execParserPure Opt.defaultPrefs parserInfo args
-  where
-    parserInfo = Opt.info (parser Opt.<**> Opt.helper) Opt.fullDesc
+withArguments :: String -> Opt.Parser a -> (a -> Command ()) -> Arguments -> Command ()
+withArguments cmdName parser onSuccess args =
+  case parseArguments cmdName parser args of
+    (ArgsParsed result) -> do
```
sodic: Why the `do`?

waspc/cli/src/Wasp/Cli/Util/Parser.hs
```diff
@@ -1,22 +1,44 @@
...
+  case Opt.execParserPure Opt.defaultPrefs parserInfo args of
+    (Opt.Success success) -> ArgsParsed success
+    (Opt.CompletionInvoked _) ->
+      error $ "Completion invoked when parsing '" <> cmdName <> "', but this should never happen"
+    (Opt.Failure failure) ->
+      case Opt.execFailure failure cmdName of
+        (help, EC.ExitSuccess, _) ->
+          ShowHelp $ show help
+        (help, EC.ExitFailure _, _) ->
+          ParseFailure $ show help
```
sodic: I find cases more readable when on a single line. It makes the mappings look cleaner and more like mappings IMO. This applies to both the `withArguments` function and to the two cases here.

But do what you prefer. RAW.

waspc/cli/src/Wasp/Cli/Util/Parser.hs
cprecioso: Just to make clear that we're not running anything else, but i can remove it

waspc/cli/src/Wasp/Cli/Util/Parser.hs
cprecioso: for consistency, eveyrthing is `do`, but i can change it

waspc/cli/src/Wasp/Cli/Util/Parser.hs
cprecioso: 👍

Issue #3432, opened 2025-11-25
Remove specialized `cd*` functions and use `cd` directly


infomiho:
We have some specialized `cd*` functions in the `deploy` package:
- `cdToServerBuildDir`
- `cdToClientBuildDir`

They just called `zx`'s `cd` function with the specific folders. Since `cd` is special function which has the side-effect of changing the internal state of `zx` we should make it explicit when it's called. 

### Refactoring proposal

Remove the specialized `cd*` functions and call `cd` directly.

ksuraev:
Hey, I'd like to work on this :) The plan was to remove specialised functions from the deploy package, find all calls to those functions, change to using cd directly, and update imports.

infomiho:
@ksuraev Hey, sure, take it, I'm looking forward to the PR. I'd recommend reading the `waspc` README to start: https://github.com/wasp-lang/wasp/tree/main/waspc#waspc

If you get stuck with something, we'll help you in the PR comments :)

infomiho:
Closed by #3456

PR #3456, opened 2025-11-27
Refactor: Remove specialized cd* functions in deploy package

ksuraev:
## Description

Removed specialised cd functions from waspProject.ts, updated *.ts files using specialised cd functions to use cd directly and updated imports.  Any advice/help on getting this complete would be appreciated :) 

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

Doesn't seem to be any existing tests in the deploy package so far, unless I'm missing something. Out of scope for this PR to set up testing for the pkg. 

- 📜 Documentation:

Closes #[3432](https://github.com/wasp-lang/wasp/issues/3432)

### DIFF
--- a/waspc/packages/deploy/src/providers/fly/commands/deploy/deploy.ts
+++ b/waspc/packages/deploy/src/providers/fly/commands/deploy/deploy.ts
@@ -1,4 +1,4 @@
-import { $, fs, cd } from "zx";
+import { $, cd, fs } from "zx";
 
 import { buildClient } from "../../../../common/clientApp.js";
 import { getFullCommandName } from "../../../../common/commander.js";
@@ -7,6 +7,10 @@ import {
   waspSays,
 } from "../../../../common/terminal.js";
 import { ensureWaspProjectIsBuilt } from "../../../../common/waspBuild.js";
+import {
+  getClientBuildDir,
+  getServerBuildDir,
+} from "../../../../common/waspProject.js";
 import {
   createDeploymentInstructions,
   DeploymentInstructions,
@@ -26,7 +30,6 @@ import {
   serverTomlExistsInProject,
 } from "../../tomlFile.js";
 import { DeployCmdOptions } from "./DeployCmdOptions.js";
-import { getClientBuildDir, getServerBuildDir } from "../../../../common/waspProject.js";
 
 export async function deploy(cmdOptions: DeployCmdOptions): Promise<void> {
   waspSays("Deploying your Wasp app to Fly.io!");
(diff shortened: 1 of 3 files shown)


waspc/packages/deploy/src/providers/fly/commands/deploy/deploy.ts
```diff
@@ -87,7 +84,8 @@ async function deployServer(
 ) {
   waspSays("Deploying your server now...");
 
-  cdToServerBuildDir(deploymentInstructions.cmdOptions.waspProjectDir);
+  const serverBuildDir = getServerBuildDir(deploymentInstructions.cmdOptions.waspProjectDir);
+  cd(serverBuildDir);
```
infomiho:
```suggestion
  cd(getServerBuildDir(deploymentInstructions.cmdOptions.waspProjectDir));
```

I'd suggest inlining the getting of the dir since we don't use the `serverBuildDir` variable anywhere below. In a way, we don't need this variable and we have to think about its name now. If we don't introduce it, there is one variable name less to think about :) 

I'd do this for all of the places you made the change.

infomiho (commented):
Thank you for the contribution, it's looking good already. I left one comment.

The best thing to finish this is to _test the change locally_. You'd do this by going through the instructions in the [`waspc/README.md`](https://github.com/wasp-lang/wasp/tree/main/waspc)

Most important bits for you would be:
- `./run build:all` which would build the `deploy` package as well as the compiler
- `./run install` it will install the compiler locally as `wasp-cli`
- `wasp-cli deploy railway launch <some-name>` for one of the example apps in the `examples` folder

This might be a bit of challenge because it does involve setting up the Haskell compiler, registering for one of the cloud providers ([Railway](https://wasp.sh/docs/deployment/deployment-methods/wasp-deploy/railway) or [Fly](https://wasp.sh/docs/deployment/deployment-methods/wasp-deploy/fly)) and deploying one of the example apps.

One more thing you'd ideally, would be to run the `./run prettier:format` so the formatting CI check passes.

If you do this - that's great, you did the whole process of testing. If it proves too difficult, let me know and I'll test it myself and report back :)

waspc/packages/deploy/src/providers/fly/commands/deploy/deploy.ts
ksuraev: Yeah that makes sense! I'll update that everywhere in one commit :) And should be fine to test the change locally, as I have most things set up already, but will let you know if I run into any issues! Thanks for the feedback :)

ksuraev:
Ran kitchen-sink with railway successfully, ran prettier and inlined the build directories :)

ksuraev:
> That's it, thank you for the contribution

Thanks for your help :)

Issue #3453, opened 2025-11-26
Adopt React 19 features


cprecioso:
- Blocked by https://github.com/wasp-lang/wasp/issues/2482

This issue is the work item tracking reviewing the contents of the new React 19 update, and adopting the ones that make sense into Wasp.

Check the initial task #2482 for a bit more discussion.

jourlez:
## React 19 Features — Codebase Investigation

I've audited the Wasp SDK templates (`waspc/data/Generator/templates/sdk/`) to identify opportunities for adopting React 19 features. Here are the findings:

---

### 1. `forwardRef` Removal (React 19 makes `ref` a regular prop) — **Low effort, recommended**

React 19 allows passing `ref` as a regular prop to function components, eliminating the need for `forwardRef`.

**Current usage** (8+ instances in auth form components):
- `sdk/wasp/auth/forms/internal/Form.tsx` — `Form`, `FormItemGroup`, `FormLabel`, `FormInput`, `FormTextarea`, `FormError`, `SubmitButton` all use `forwardRef`
- `sdk/wasp/auth/forms/internal/Message.tsx` — `Message` uses `forwardRef`

**Migration**: Each component can be simplified from:
```tsx
export const FormInput = forwardRef<ComponentRef<"input">, ComponentPropsWithoutRef<"input">>(
  ({ className, ...props }, ref) => (
    <input className={clsx(styles.formInput, className)} {...props} ref={ref} />
  )
);
```
to:
```tsx
export const FormInput = ({ className, ref, ...props }: ComponentPropsWithRef<"input">) => (
  <input className={clsx(styles.formInput, className)} {...props} ref={ref} />
);
```

**Effort**: Low — straightforward mechanical change, no behavioral impact.

---

### 2. Suspense Support (`useSuspenseQuery`) — **Medium effort, recommended as follow-up**

**Current state**: Zero `<Suspense>` usage in templates. Wasp wraps TanStack React Query's `useQuery` in `sdk/wasp/client/operations/hooks.ts` but does **not** expose `useSuspenseQuery`.

TanStack React Query already ships `useSuspenseQuery` (since v5, currently at ~4.42.0 in Wasp). This is tracked separately in #1239.

**What's needed**:
1. TanStack Query upgrade to v5+ (tracked in #2244 / #3482, assigned to @sodic)
2. Add a `useSuspenseQuery` wrapper alongside the existing `useQuery` wrapper in `hooks.ts`
[...]

Issue #3256, opened 2025-10-14
Complete e2e tests for `ask-the-documents`


cprecioso:
_Originally posted by @infomiho in https://github.com/wasp-lang/wasp/pull/3250#discussion_r2425859000_

We should test out the full flow of the `ask-the-documents` app. Probably by mocking the Google and OpenAI communication somehow (MSW, MitM?)

Martinsos:
Ok, it is important to have good test.
Let's brainstorm more on concrete solutions, try to visalize them a bit, then we will decide how to proceed.

PR #3458, opened 2025-11-27
Fix email verification hook being called multiple times

### LINKED ISSUE #3407: `onAfterEmailVerified` hook called twice in development
The `onAfterEmailVerified` hook is being triggered twice in development mode when using Wasp's `<VerifyEmailForm />` component, causing duplicate operations (e.g., welcome emails being sent twice).

**Technically this isn't a bug but expected behaviour:** React 18's Strict Mode renders components twice in development to check if components are doing their cleanup correctly. Since `<VerifyEmailForm />` triggers the verification backend endpoint as soon as it's mounted, this results in the request being sent twice in development.

**But it's unexpected for users** and they might think it will happen in production.

### Proposed solution

Implement a similar workaround as was done for the [OAuth component](https://github.com/wasp-lang/wasp/blob/main/waspc/data/Generator/templates/react-app/src/auth/pages/OAuthCallback.tsx#L70-L76) to prevent the backend from being triggered twice in development mode.

Based on https://discord.com/channels/686873244791210014/1438285961225044039/1441432541402763446

infomiho:
## Description

Closes #3407

It makes sure the endpoint is not called twice and updates the Playwright tests to test for that. This is an _okay_ workaround to help users not be confused while developing as to why the `onAfterEmailVerified` is called twice.

## Type of change

- [x] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/e2e-tests/snapshots/kitchen-sink-golden/snapshot-file-list.manifest
+++ b/waspc/e2e-tests/snapshots/kitchen-sink-golden/snapshot-file-list.manifest
@@ -84,6 +84,7 @@ wasp-app/.wasp/out/sdk/wasp/client/crud/operationsHelpers.ts
 wasp-app/.wasp/out/sdk/wasp/client/crud/tasks.ts
 wasp-app/.wasp/out/sdk/wasp/client/env.ts
 wasp-app/.wasp/out/sdk/wasp/client/env/schema.ts
+wasp-app/.wasp/out/sdk/wasp/client/hooks.ts
 wasp-app/.wasp/out/sdk/wasp/client/index.ts
 wasp-app/.wasp/out/sdk/wasp/client/operations/actions/core.ts
 wasp-app/.wasp/out/sdk/wasp/client/operations/actions/index.ts
@@ -337,6 +338,10 @@ wasp-app/.wasp/out/sdk/wasp/dist/client/env/schema.d.ts
 wasp-app/.wasp/out/sdk/wasp/dist/client/env/schema.d.ts.map
 wasp-app/.wasp/out/sdk/wasp/dist/client/env/schema.js
 wasp-app/.wasp/out/sdk/wasp/dist/client/env/schema.js.map
+wasp-app/.wasp/out/sdk/wasp/dist/client/hooks.d.ts
+wasp-app/.wasp/out/sdk/wasp/dist/client/hooks.d.ts.map
+wasp-app/.wasp/out/sdk/wasp/dist/client/hooks.js
+wasp-app/.wasp/out/sdk/wasp/dist/client/hooks.js.map
 wasp-app/.wasp/out/sdk/wasp/dist/client/index.d.ts
 wasp-app/.wasp/out/sdk/wasp/dist/client/index.d.ts.map
 wasp-app/.wasp/out/sdk/wasp/dist/client/index.js
--- a/waspc/e2e-tests/snapshots/kitchen-sink-golden/wasp-app/.wasp/out/.waspchecksums
+++ b/waspc/e2e-tests/snapshots/kitchen-sink-golden/wasp-app/.wasp/out/.waspchecksums
@@ -165,7 +165,7 @@
             "file",
             "../out/sdk/wasp/auth/forms/internal/email/VerifyEmailForm.tsx"
         ],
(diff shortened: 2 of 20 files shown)


waspc/data/Generator/templates/sdk/wasp/auth/forms/internal/email/VerifyEmailForm.tsx
```diff
@@ -1,4 +1,4 @@
-import { useContext, useEffect } from 'react'
```
infomiho, line 1: The fix is just this file - rest is updating headless and snapshot tests.

waspc/data/Generator/templates/sdk/wasp/auth/forms/internal/email/VerifyEmailForm.tsx
```diff
@@ -34,9 +34,13 @@ export const VerifyEmailForm = () => {
...
 
+  const isFirstRender = useRef(true)
   useEffect(() => {
-    submitForm()
-  }, [location])
+    if (isFirstRender.current) {
+      isFirstRender.current = false
+      submitForm()
+    }
+  }, [])
```
cprecioso: I'll allow it 😂 since we can easily take care of this once we do #3453

waspc/ChangeLog.md
```diff
@@ -2,6 +2,10 @@
 
 ## Next
 
+### 🐞 Bug fixes
+
+- Fixed email verification hook being called multiple times. ([#3458](https://github.com/wasp-lang/wasp/pull/3458))
```
cprecioso: Mmm, same thing as in the other PR, I think we can morph this into something informative for our users who might not have all the context

```suggestion
- Fixed `onAfterEmailVerified` hook being called twice while in development mode. ([#3458](https://github.com/wasp-lang/wasp/pull/3458))
```

examples/kitchen-sink/schema.prisma
```diff
@@ -12,9 +12,10 @@ generator client {
 model User {
   id Int @id @default(autoincrement())
 
-  isOnAfterSignupHookCalled        Boolean @default(false)
-  isOnAfterLoginHookCalled         Boolean @default(false)
-  isOnAfterEmailVerifiedHookCalled Boolean @default(false)
+  isOnAfterSignupHookCalled          Boolean @default(false)
+  isOnAfterLoginHookCalled           Boolean @default(false)
+  isOnAfterEmailVerifiedHookCalled   Boolean @default(false)
+  numTimesOnAfterEmailVerifiedCalled Int     @default(0)
```
cprecioso, line 18: Good thinking!

waspc/data/Generator/templates/sdk/wasp/auth/forms/internal/email/VerifyEmailForm.tsx
sodic: Can we have a comment (both here and in the auth component) to explain what's happening and why. Better yet, we could have a custom reusable hook with a nice name. Any comments could then go to its definition.

Also, do we have a long term solution for this? If not, I think we should have an issue for creating one and point to that from the code.

examples/kitchen-sink/schema.prisma
sodic: Do we have the same test for the oauth component. If it's fast, I suggest adding it in. If not, you can skip it.

infomiho: Don't have any OAuth headless tests because we can't mock Google login yet (same as in #3256) so I can't add this kind of test easily.

waspc/data/Generator/templates/sdk/wasp/auth/forms/internal/email/VerifyEmailForm.tsx
infomiho: Long term fix will be upgrading to React 19 and using one of the methods @cprecioso proposed in the issue:
> Suspense or useAction

Extracting this as a reusable hook is a bit tricky since we need it both in the SDK and the web-app. So I'd need to export a new symbol from the SDK for the web-app. I'll see what I can do and report back.

infomiho: Extracted logic into `useEffectOnce` and added the comment there. Updated the public API table.

Issue #3135, opened 2025-09-05
Vite unleashed


cprecioso:
> [!NOTE]
> This is more of an open-ended task, as it includes the work of investigating which features we want to expose. I've called this kind of task "[x] unleashed". [Here's what I mean by it.](https://www.notion.so/wasp-lang/INFO-What-does-it-mean-to-unleash-26318a74854c818bb7b1e0adb46b5e9c)

- Do we need to give access to our `index.html`? Why would a user want it for that is not currently afforded?
- Especially, pay attention to whether problems might be solved by importing our Vite config from the user’s config, or vice-versa
- Do we support more specialized plugins, like the ones that add middleware?

---
Related issues
- https://github.com/wasp-lang/wasp/issues/1874
- https://github.com/wasp-lang/wasp/issues/3074
- #2437
- #866

PR #3459, opened 2025-11-28
Using user's `vite.config.ts` as entry point POC

infomiho:
## Description

This is POC of _one of the approaches_ to give more control over Vite to the user. Not a real PR that needs to be reviewed and merged. This PR will be deleted afterwards. 

### What's in the POC

- [x] Modified the `web-app` to use the user's config _directly_ (this means user's config is **the real config** but `web-app` is still the entry point, we are still a framework)
  - [x] [`vite` CLI is invoked with `../../../vite.config.ts` flag](https://github.com/wasp-lang/wasp/pull/3459/files#diff-3774962ea201188060cd4d0a576b8385975e53429035de66cbc6308bb0907bb3R11)
  - [x] `web-app`'s `vite.config.ts` no longer imports user's config and merges with Wasp's config
  - [x] User's config now includes the `wasp()` plugin which contains all the important options
- [x] Moved all Vite config options (e.g. client app port, `react()` plugin etc.) to the [`wasp()`](https://github.com/wasp-lang/wasp/pull/3459/files#diff-98309ea31e5116b5f1fe0b27612d96a5a13a768260e5a4e26a8fe3b23f7554e8R13) plugin
  - [x] The plugin applies all the same Wasp-specific options and it's still templated
  - [x] The plugin accepts options to configure the `react()` plugin
  - [x] The plugin extends the `test` options for Vitest to ignore `.wasp/**/*` if Vitest is used from user-land
  - [x] The plugin is located in the SDK and users import it from `wasp/client/vite` and use in the `vite.config.ts`:
      ```ts
      import { defineConfig } from "vite";
      import { wasp } from "wasp/client/vite";
	
      export default defineConfig({
        plugins: [wasp()],
      });
      ```
- [x] Moved Vitest to user-land (this is probably not 100% necessary to do, but I was cleaning up and did this at the same time, in the real PR this would come as a separate effort)
  - [x] Removed dependencies related to Vitest
  - [x] Deleted `wasp test client` command
  - [x] Deleted Wasp-specific test helpers (recreated them in `kitchen-sink`, not sure what would be the best way to ship these helpers, as part of the docs?)
  - [x] Setup Vitest in user-land with latest Vitest 4 (and not Vitest 1 which we had in Wasp)
 
Part of #3135 #3472


infomiho:

PR #3462, opened 2025-11-28
Making user's project the root for Vite POC

infomiho:
## Description

This PR is not for reviewing. It might serve for discussion later on.

What was done:
- [x] Run `npx vite` from the user project dir and not `.wasp/out/web-app`
- [x] Move `index.html` to user-land which loads the `index.tsx`
- [x] Move `index.tsx` that mounts the React app to user-land
- [x] Move `public/manifest.json` to user-land
- [x] Moves the `web-app` code to the SDK
  - The router can't import user code directly since the SDK lives in the `node_modules`. Instead, it needs to import the user code from the `src` dir that was copied into the SDK. 
 
Part of #3135 #3472


infomiho:

PR #3463, opened 2025-12-01
CI: Extract and don't hardcode artifact naming

cprecioso:
## Description

- As part of https://github.com/wasp-lang/wasp/pull/3464

---

- I extracted the computation of the artifact and tarball name out of the `waspc-build` workflow, so that we can reuse it in multiple places (mainly, the `fetch-nightly-packages`; and in the future the npm packaging script) without repeating ourselves.
- Makes the waspc-build job output the computed names of the artifacts it has created. This way we don't hardcode these names in the following workflows.
  - This requires a bit of a workaround since, in matrix jobs, each job will override the output. Instead, we store each job's output in a different variable.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/.github/actions/compute-artifact-names/action.yaml
+++ b/.github/actions/compute-artifact-names/action.yaml
@@ -24,8 +24,6 @@ runs:
         )
       shell: sh
       run: |
-        # Using the ::error:: prefix lets GitHub handle it as a special string, and shows up as a red error in the logs.
-        # https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-an-error-message
         echo '::error::Invalid build name provided: ${{ inputs.build-name }}.'
         exit 1
 
--- a/.github/workflows/ci-waspc-build.yaml
+++ b/.github/workflows/ci-waspc-build.yaml
@@ -175,26 +175,21 @@ jobs:
 
       - name: Unpack, create universal binary and pack
         run: |
-          set -ex # Fail on error and print each command
+          set -eux # Fail on error and print each command
 
-          input_files=(
-            "$X86_64_TARBALL_NAME"
-            "$AARCH64_TARBALL_NAME"
-          )
+          mkdir x86_64 aarch64 universal
 
-          # Extract each architecture
-          i=0;
-          for file in "${input_files[@]}"; do
-            mkdir "arch-$i"
-            tar -xzf "$file" -C "arch-$i"
-            i=$((i + 1))
-          done
+          # Unpack both architectures
+          tar -xzf "$X86_64_TARBALL_NAME" -C x86_64
+          tar -xzf "$AARCH64_TARBALL_NAME" -C aarch64
 
-          mkdir universal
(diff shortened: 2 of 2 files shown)


.github/actions/compute-artifact-names/action.yaml
```diff
@@ -0,0 +1,45 @@
...
+            "macos-x86_64",
+            "macos-aarch64",
+            "macos-universal"
+          ]'),
+          inputs.build-name
+        )
+      shell: sh
+      run: |
+        # Using the ::error:: prefix lets GitHub handle it as a special string, and shows up as a red error in the logs.
+        # https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-an-error-message
```
infomiho: On the fence if we need this comment, for example, you are not explaining the `fromJson` API for Github Action etc.

But I'm okay if we keep it 👍 

RAW

.github/workflows/ci-waspc-build.yaml
```diff
@@ -112,52 +124,82 @@ jobs:
...
           # Create the universal binary
           lipo -create arch-*/wasp-bin -output universal/wasp-bin
-          # Copy the data folder too
-          cp -R "arch-${input_arch[0]}/data" universal/
+          # Copy the data folder too (we take the first one, should be identical in both)
+          cp -R "arch-0/data" universal/
 
           # Pack back up
-          tar -czf wasp-macos-universal.tar.gz -C universal .
+          tar -czf ${{ steps.compute-artifact-names.outputs.tarball-name }} -C universal .
```
infomiho: Something feels off about using `$i` here for me. I asked some of my LLM buddies to help me with some suggestions:
1. You could use `for i in "${!input_files[@]}"; do` to get the array indicies which lets you do `tar -xzf "${input_files[$i]}" -C "arch-$i"`.
2. Another suggestion was to go fully explicit with something like:
    ```bash
    set -ex

    # Create semantic directories
    mkdir x86_64 aarch64 universal

    # Extract specifically to those directories
    tar -xzf "$X86_64_TARBALL_NAME" -C x86_64
    tar -xzf "$AARCH64_TARBALL_NAME" -C aarch64

    # Create the universal binary
    # Now it is obvious which inputs are being used
    lipo -create x86_64/wasp-bin aarch64/wasp-bin -output universal/wasp-bin

    # Copy data from one of the sources (explicitly choosing x86_64)
    cp -R x86_64/data universal/

    # Pack back up
    tar -czf ${{ steps.compute-artifact-names.outputs.tarball-name }} -C universal .
    ```

infomiho: I found the second option the most readable since this is an exact thing we are doing: creating 1 universal binary out of Intel and ARM binaries. We don't need the extensibility here, it'll always be that.

PR #3464, opened 2025-12-01
CI: Generate npm packages

cprecioso:
Adds a script to the repo (and CI) that takes in the built compiler artifacts and packages them as a family of npm packages. 

These packages are not uploaded as part of this PR, only built. But once uploaded, a user will be able to run the wasp cli with `npm i -g @wasp.sh/wasp-cli` or `npx @wasp.sh/wasp-cli`. 

**Architecture:**

The npm package "family" solution is similar to the one employed by other packages that include native code such as esbuild or rollup. 

These consist of a main package (`@wasp.sh/wasp-cli`) and per-target-environment sub packages (`@wasp.sh/wasp-cli-darwin-x64`, `@wasp.sh/wasp-cli-linux-x64-musl`, etc). 

Each sub package includes their own full copy of the wasp cli and the data dir, plus a JS file that exports the paths those files. Their `package.json` marks their target environment compatibility (through the `os`, `cpu`, and `libc` keys). This means that straight up installing those packages in an environment different than their declared ones would have `npm install` fail. 

The main package marks all of the subpackages as `optionalDependencies`, which means that it will install the ones that it can, and ignore the ones that it can't. From that main package then we expose a binary that tries to import all of the subpackages and calls the first binary that it finds, otherwise will fail. 

Testing is setup at #3469

### DIFF
--- a/scripts/make-npm-packages/src/generator/main-package.ts
+++ b/scripts/make-npm-packages/src/generator/main-package.ts
@@ -3,7 +3,10 @@ import * as path from "node:path";
 import type { PackageJson } from "type-fest";
 import { TEMPLATES_DIR, copyStaticFiles } from "../common.ts";
 import type { TarballData } from "../schema/input-data.ts";
-import { noLibcName, type MainPackageData } from "../schema/output-data.ts";
+import {
+  UNDEFINED_LIBC_NAME,
+  type MainPackageData,
+} from "../schema/output-data.ts";
 
 const TEMPLATE_DIR = path.join(TEMPLATES_DIR, "main-package");
 const TEMPLATE_STATIC_FILES = ["bin.js", "CLIError.js"];
@@ -76,7 +79,7 @@ function writeDataFile(
 
   const pkgPaths = { subPackages: {} } as MainPackageData;
   for (const subPackage of subPackages) {
-    const { os, cpu, libc = noLibcName } = subPackage.data.target;
+    const { os, cpu, libc = UNDEFINED_LIBC_NAME } = subPackage.data.target;
     pkgPaths.subPackages[os] ??= {};
     pkgPaths.subPackages[os][cpu] ??= {};
     pkgPaths.subPackages[os][cpu][libc] = {
--- a/scripts/make-npm-packages/src/generator/sub-package.ts
+++ b/scripts/make-npm-packages/src/generator/sub-package.ts
@@ -4,7 +4,7 @@ import * as path from "node:path";
 import type { PackageJson } from "type-fest";
 import { TEMPLATES_DIR, copyStaticFiles } from "../common.ts";
 import type { NpmTarget } from "../schema/input-data.ts";
-import { noLibcName } from "../schema/output-data.ts";
+import { UNDEFINED_LIBC_NAME } from "../schema/output-data.ts";
 import type { Path } from "../schema/util.ts";
 
 const TEMPLATE_DIR = path.join(TEMPLATES_DIR, "sub-package");
@@ -71,7 +71,7 @@ function generatePackageJson(
   pkg.bin = {
     // The name doesn't matter (and in fact we don't want users to call it
     // directly), so we'll make it clear.
-    [`__internal_wasp-${target.os}-${target.cpu}-${target.libc ?? noLibcName}`]:
+    [`__internal_wasp-${target.os}-${target.cpu}-${target.libc ?? UNDEFINED_LIBC_NAME}`]:
       "wasp-bin",
   };
 
--- a/scripts/make-npm-packages/src/schema/args.ts
+++ b/scripts/make-npm-packages/src/schema/args.ts
@@ -1,7 +1,7 @@
 import * as z from "zod";
 import type { CheckFn } from "zod/v4/core";
 import type { NpmTarget } from "./input-data.ts";
-import { noLibcName } from "./output-data.ts";
+import { UNDEFINED_LIBC_NAME } from "./output-data.ts";
 import { isDirPath, pathExists, PathSchema } from "./util.ts";
 
 const PLACEHOLDERS = {
@@ -32,7 +32,7 @@ export const ArgsSchema = z.object({
     .check(checkIncludesPlaceholder(PLACEHOLDERS.libc))
     .transform(
       (s) =>
-        ({ os, cpu, libc = noLibcName }: NpmTarget) =>
+        ({ os, cpu, libc = UNDEFINED_LIBC_NAME }: NpmTarget) =>
           s
             .replace(PLACEHOLDERS.os, os)
             .replace(PLACEHOLDERS.cpu, cpu)
--- a/scripts/make-npm-packages/src/schema/output-data.ts
+++ b/scripts/make-npm-packages/src/schema/output-data.ts
@@ -5,13 +5,13 @@ export interface SubPackageInfo {
 // This can be any string representing a lack of a defined libc for the build
 // (e.g. macOS or Windows don't use libc in the same way Linux does).
 // The usual convention is to call it "unknown".
-export const noLibcName = "unknown" as const;
+export const UNDEFINED_LIBC_NAME = "unknown" as const;
 
 export interface MainPackageData {
   subPackages: {
     [os: string]: {
       [arch: string]: {
-        [libc: string | typeof noLibcName]: SubPackageInfo;
+        [libc: string | typeof UNDEFINED_LIBC_NAME]: SubPackageInfo;
       };
     };
   };
--- a/scripts/make-npm-packages/templates/main-package/bin.js
+++ b/scripts/make-npm-packages/templates/main-package/bin.js
@@ -53,7 +53,7 @@ async function getSubPackage() {
       ? isGlibc()
         ? "glibc"
         : "musl"
-      : /** @satisfies {typeof import("../../src/schema/output-data.ts").noLibcName} */ (
+      : /** @satisfies {typeof import("../../src/schema/output-data.ts").UNDEFINED_LIBC_NAME} */ (
           "unknown"
         );
 


.github/actions/compute-artifact-names/action.yaml
```diff
@@ -14,6 +14,7 @@ runs:
       if: >
         ! contains(
           fromJson('[
+            "*",
```
infomiho: What does this special value mean? Can we swap out `*` for something more descriptive or add a comment that explains it?

infomiho: I noticed it's used in the `artifact-name` as well, what does `wasp-cli-*` artifact contain?

.github/actions/compute-artifact-names/action.yaml
```diff
@@ -36,10 +37,27 @@ runs:
         echo "artifact-name=wasp-cli-${{ inputs.build-name }}" >> $GITHUB_OUTPUT
         echo "tarball-name=wasp-${{ inputs.build-name }}.tar.gz" >> $GITHUB_OUTPUT
 
+    - name: Set npm triple
```
infomiho: Can we use a more descriptive name here instead of a triple here? It's sometimes a pair and sometimes are a triple - so the name is probably wrong 😄 

We are encoding the OS and the arch I guess and which C lib to use. So maybe a better name would be `npm-package-metadata` or something like that.

.github/workflows/ci-waspc-build.yaml
```diff
@@ -203,3 +205,56 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Compute artifact names for download
+        uses: ./.github/actions/compute-artifact-names
+        id: compute-download-artifact-names
+        with:
+          build-name: "*"
```
infomiho: Now I'm guessing that `*` means "all" of the artifacts?

.github/workflows/ci-waspc-build.yaml
```diff
@@ -203,3 +205,56 @@ jobs:
...
+
+      - uses: actions/download-artifact@v6
+        with:
+          path: "./download/"
+          pattern: ${{ steps.compute-download-artifact-names.outputs.artifact-name }}
+
+      - name: Copy artifacts to input folder
+        run: |
+          mkdir -p ./input/
+          cp ./download/*/* ./input/
```
infomiho: I guess this the folder `*` and all the files inside of it? We definitely need to change the `*` name to something more meaningful.

.github/workflows/ci-waspc-build.yaml
```diff
@@ -203,3 +205,56 @@ jobs:
...
+              tarballs:
+                with_entries(select(.key | startswith("output_")))
+                | map(
+                    fromjson
+                    | {
+                        fileName: .["tarball-name"],
+                        target: (.["npm-triple"] | fromjson)
+                      }
+                  )
+            }
```
infomiho, line 247: This feels complex enough not to be an inline `jq` call. What about a simple JS script?

I have to admit I got lost at first since I'm not that proficient in `jq`.

.github/workflows/ci-waspc-build.yaml
```diff
@@ -203,3 +205,56 @@ jobs:
...
+        with:
+          path: "./download/"
+          pattern: ${{ steps.compute-download-artifact-names.outputs.artifact-name }}
+
+      - name: Copy artifacts to input folder
+        run: |
+          mkdir -p ./input/
+          cp ./download/*/* ./input/
+
+      - name: Create data file
```
infomiho: This name could be better, what is a data file? Is it npm packages tarballs data?

.github/workflows/ci-waspc-build.yaml
infomiho: Maybe we feed the JSON as input to the `./scripts/make-npm-packages/run.bash` file?

scripts/make-npm-packages/templates/main-package/bin.js
```diff
@@ -0,0 +1,81 @@
...
+    }
+  };
+
+  static log = (error) => {
+    if (error instanceof CLIError) {
+      console.error("Error:", error.message);
+    } else {
+      console.error(
+        "Unknown Error launching Wasp:",
+        error.message || String(error),
```
infomiho:
```suggestion
        error.message ?? String(error),
```

Can we use this? Do we have any assurance about what the Node.js runtime will be?

scripts/make-npm-packages/templates/main-package/bin.js
```diff
@@ -0,0 +1,81 @@
...
+    }
+  };
+}
+
+try {
+  process.exitCode = await main();
+} catch (error) {
+  CLIError.log(error);
+  process.exitCode = -1;
+}
```
infomiho: What's going on here? Is this some sort of industry standard when shipping native binaries 😄

scripts/make-npm-packages/main-package/make.bash
```diff
@@ -0,0 +1,40 @@
+set -euo pipefail
```
infomiho: All of our scripts use `.sh` suffix, why not keep using it 😄 

I know this is green field and it's a `bash` only script, but it still feels weird to have some scripts as `sh` and some as `bash`.

scripts/make-npm-packages/main-package/make.bash
```diff
@@ -0,0 +1,40 @@
...
+    .
+    | .name = $packageName
+    | .version = $inputData.version
+    | .optionalDependencies = (
+        $subPackageNames
+        | map({ (.): $inputData.version })
+        | add
+      )
+    | .os = ($inputData.tarballs | map(.target[0]) | unique)
+    | .cpu = ($inputData.tarballs | map(.target[1]) | unique)
```
infomiho: Why did you have to use `unique` here?

scripts/make-npm-packages/main-package/template/bin.js
```diff
@@ -0,0 +1,81 @@
...
+}
+
+async function findWaspExecutable() {
+  const { default: data } = await import("./data.json", {
+    with: { type: "json" },
+  });
+
+  const { waspBinPath, dataDirPath } = await Promise.any(
+    data.subPackageNames.map((pkg) => import(pkg)),
+  ).catch(
```
infomiho: How much of a performance penalty is this? Trying to load all the sub packages until one succeeds. Can we maybe know exactly which OS and which platform is used? e.g. `os.arch()`

scripts/make-npm-packages/main-package/make.bash
```diff
@@ -0,0 +1,40 @@
...
+  ' \
+  > "$outputDirPath/package.json"
+
+jq -n \
+  --argjson subPackageNames "$subPackageNamesJson" \
+  '{ subPackageNames: $subPackageNames }' \
+  > "$outputDirPath/data.json"
```
infomiho: I'd probably add a comment to explain the usage place of this `data.json` file ("it's used in the main package to find the available binary" or similar)

scripts/make-npm-packages/sub-package/make.bash
```diff
@@ -0,0 +1,37 @@
...
+tarballPath=$3
+tarballDataJson=$4
+outputDirPath=$5
+
+templateDir="$(dirname "$0")/template"
+
+source "$(dirname "$0")/../util.bash"
+
+rm -rf "${outputDirPath:?}/*" || true
+mkdir -p "$outputDirPath"
```
infomiho: Since it's repeated for both main and sub-package - maybe this can go to the utils file?

.github/workflows/ci-waspc-build.yaml
cprecioso: In this case this was just supposed to be the wildcard glob. But I'll replace it for the output of the compute names step.

.github/actions/compute-artifact-names/action.yaml
cprecioso: Renamed it "all"

.github/actions/compute-artifact-names/action.yaml
cprecioso: Renamed it "npm-target"

scripts/make-npm-packages/templates/main-package/bin.js
cprecioso: It's Node 22, so `??` is supported. I wanted to avoid printing an empty string, which is why I use `||`.

scripts/make-npm-packages/templates/main-package/bin.js
cprecioso: Nope, just trying to make everything nice. Here we're not dealing with the errors that the Wasp CLI might exit with (those are considered just an exit code and returned); the `catch` here is only reached if there was a problem in the actual process of finding and calling the CLI itself. Given it's an unknown error, we log it and exit with `-1`.

cprecioso: I made the `CLIError` thing to ensure we always log a human-readable error from this wrapper, and not some JS traceback that confuses people.

cprecioso: I added some comments in the code

scripts/make-npm-packages/main-package/make.bash
cprecioso: For example, we have two packages for (linux, x64) (one for glibc and another for musl), but we don't want to have the value twice.

scripts/make-npm-packages/main-package/template/bin.js
cprecioso: Yep, I made a new version using this data

.github/actions/compute-artifact-names/action.yaml
```diff
@@ -3,7 +3,10 @@ description: Central place for computing the artifact names in the CI builds.
 
 inputs:
   build-name:
-    description: The build name (e.g., linux-x86_64, macos-aarch64, etc.)
+    description: >
+      The build name (e.g., linux-x86_64, macos-aarch64, etc.). The special
+      value "all" can be used to receive a glob that matches any or all of
+      the names.
```
infomiho:
```suggestion
      The build name (e.g., linux-x86_64, macos-aarch64, etc.). The special
      value "all" can be used to receive a glob that matches all of the build
      names.
```

I got confused by `any or all`

.github/actions/compute-artifact-names/action.yaml
```diff
@@ -23,18 +27,33 @@ runs:
...
         # https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-an-error-message
-        echo '::error::Invalid build name provided: ${{ inputs.build-name }}.'
-        exit 1
+        "echo '::error::Invalid build name provided: ${{ inputs.build-name }}.' && exit 1"
 
     - name: Set artifact names
       id: set-artifact-names
       shell: sh
+      env:
+        BUILD_NAME: ${{ inputs.build-name == 'all' && '*' || inputs.build-name }}
```
infomiho, line 38: Why are we swapping `all` for `*` here? I thought we decided to use `all` as the "special name".

scripts/make-npm-packages/src/schema/util.ts
```diff
@@ -0,0 +1,52 @@
...
+      }
+    }
+  };
+
+export const [isDir, isFile] = (
+  [
+    { kind: "directory", checkFn: (stat) => stat.isDirectory() },
+    { kind: "file", checkFn: (stat) => stat.isFile() },
+  ] as Parameters<typeof createStatChecker>[0][]
+).map(createStatChecker);
```
infomiho:
```suggestion
const isDir = createStatChecker({ kind: "directory", checkFn: (stat) => stat.isDirectory() });
const isFile = createStatChecker({ kind: "file", checkFn: (stat) => stat.isFile() };
```

I think the `map` bring more complexity than it's justified.

scripts/make-npm-packages/src/schema/util.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const exists = (ctx: ParsePayload<Path>) => {
```
infomiho:
```suggestion
export const pathExists = (ctx: ParsePayload<Path>) => {
```

waspc/waspc.cabal
```diff
@@ -329,7 +329,6 @@ library
     Wasp.Generator.SdkGenerator.Client.RouterGenerator
     Wasp.Generator.SdkGenerator.Common
     Wasp.Generator.SdkGenerator.CrudG
-    Wasp.Generator.SdkGenerator.DepVersions
```
infomiho: Was this part of a rebase on `main`? I feel like we already did this change here: https://github.com/wasp-lang/wasp/commit/57c6ed3e32fcf52b49dd954833724ac38ad047aa

scripts/make-npm-packages/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "include": ["./src/**/*"],
```
infomiho: Nitpick: we usually kept `include` after the `compilerOptions`.

scripts/make-npm-packages/src/generator/main-package.ts
```diff
@@ -0,0 +1,104 @@
+import fs from "fs-extra";
+import * as path from "node:path";
+import type { PackageJson } from "type-fest";
+import { COMMON_TEMPLATE_DIR } from "../common.ts";
+import type { TarballData } from "../schema/input-data.ts";
+import type { MainPackageData } from "../schema/output-data.ts";
+
+const TEMPLATE_DIR = path.join(COMMON_TEMPLATE_DIR, "main-package");
+const TEMPLATE_COPY_FILES = ["bin.js", "CLIError.js"];
```
infomiho:
```suggestion
const TEMPLATE_STATIC_FILES = ["bin.js", "CLIError.js"];
```

You use this in `copyStaticFiles` which signals to me that these files are "static files". The fact that we copy them is a detail of the `makeMainPackage` fn.

scripts/make-npm-packages/src/generator/main-package.ts
```diff
@@ -0,0 +1,104 @@
...
+  outputDirPath,
+}: {
+  packageName: string;
+  packageVersion: string;
+  subPackages: { packageName: string; data: TarballData }[];
+  outputDirPath: string;
+}) {
+  console.group(`Creating main package ${packageName}...`);
+
+  console.log("Preparing main package directory:", outputDirPath);
```
infomiho:
```suggestion
  console.log(`Preparing main package directory: ${outputDirPath}`);
```

Nitpick

scripts/make-npm-packages/src/generator/main-package.ts
```diff
@@ -0,0 +1,104 @@
...
+
+function generatePackageJson(
+  packageName: string,
+  packageVersion: string,
+  subPackages: { packageName: string; data: TarballData }[],
+) {
+  console.group("Generating package.json...");
+
+  const pkgTemplateFilePath = path.join(TEMPLATE_DIR, "package.json");
+  const pkg = fs.readJsonSync(pkgTemplateFilePath) as PackageJson;
```
infomiho, line 54: We used Zod schemas for all other input, why not use it for `PackageJson`? Also, the `PackageJson` type probably contains more fields than we have in our template `package.json`.

scripts/make-npm-packages/src/generator/main-package.ts
```diff
@@ -0,0 +1,104 @@
...
+  console.log("Read package.json template:", pkgTemplateFilePath);
+
+  pkg.name = packageName;
+  pkg.version = packageVersion;
+
+  pkg.optionalDependencies = {};
+  for (const subPkgName of subPackages)
+    pkg.optionalDependencies[subPkgName.packageName] = packageVersion;
```
infomiho:
```suggestion
  for (const subPkgName of subPackages) {
    pkg.optionalDependencies[subPkgName.packageName] = packageVersion;
  }
```

scripts/make-npm-packages/src/generator/main-package.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  pkg.os = unique(subPackages.map((subPackage) => subPackage.data.target[0]));
+  pkg.cpu = unique(subPackages.map((subPackage) => subPackage.data.target[1]));
```
infomiho: Can we somehow avoid using indexes and use properties? e.g. `target.os` and `target.cpu`

scripts/make-npm-packages/src/generator/sub-package.ts
```diff
@@ -0,0 +1,92 @@
...
+  for (const fileName of fileNames) {
+    fs.copyFileSync(
+      path.join(TEMPLATE_DIR, fileName),
+      path.join(outputDirPath, fileName),
+    );
+    console.log(`Copied ${fileName}`);
+  }
+
+  console.groupEnd();
+}
```
infomiho: We have the same fn in `main-package.ts` - let's maybe extract it into a util fn.

scripts/make-npm-packages/src/generator/sub-package.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  const pkgTemplateFilePath = path.join(TEMPLATE_DIR, "package.json");
+  const pkg = fs.readJsonSync(pkgTemplateFilePath) as PackageJson;
+  console.log("Read package.json template:", pkgTemplateFilePath);
+
+  pkg.name = packageName;
+  pkg.version = packageVersion;
+  pkg.os = [target.os];
+  pkg.cpu = [target.cpu];
+  if (target.libc) pkg.libc = [target.libc];
```
infomiho:
```suggestion
  if (target.libc) {
    pkg.libc = [target.libc];
  }
```

scripts/make-npm-packages/src/generator/sub-package.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  // npm strips the executable bit from files in packages, so we use the "bin"
+  // field to let npm know it needs to mark it as executable upon install.
+  pkg.bin = {
+    // The name doesn't matter (and in fact we don't want users to call it
+    // directly), so we'll make it clear.
+    [`__internal_wasp-${target.os}-${target.cpu}-${target.libc ?? "unknown"}`]:
```
infomiho: I've seen this `unknown` value repeated, is this a special value we need to use? Let's encode it in a constant then.

scripts/make-npm-packages/src/schema/args.ts
```diff
@@ -0,0 +1,34 @@
...
+  "input-dir": { type: "string" },
+  "output-dir": { type: "string" },
+  "main-package-name": { type: "string", default: "@wasp.sh/wasp-cli" },
+  "sub-package-name": {
+    type: "string",
+    default: "@wasp.sh/wasp-cli-$os-$cpu-$libc",
+  },
+} as const;
+
+const includesPlaceholder = (placeholder: string) =>
```
infomiho:
```suggestion
const assertIncludesPlaceholder = (placeholder: string) =>
```

It's more clear that this validates rather than just produce a boolean.

scripts/make-npm-packages/src/schema/util.ts
```diff
@@ -0,0 +1,52 @@
...
+import type { ParsePayload } from "zod/v4/core";
+
+export const PathSchema = (from = process.cwd()) =>
+  z
+    .string()
+    .overwrite((p) => path.resolve(from, p))
+    .brand<"Path">();
+export type Path = z.infer<ReturnType<typeof PathSchema>>;
+
+const createStatChecker =
```
infomiho: This reads to me more like a `createPathCheckerFn` since the error messages talk about paths.

scripts/make-npm-packages/src/common.ts
```diff
@@ -0,0 +1,6 @@
+import * as path from "node:path";
+
+export const COMMON_TEMPLATE_DIR = path.resolve(
```
infomiho: Drop the "common", it's cleaner.

Image

I think you can go with just `TEMPLATE_DIR` since there is only one template dir. The "common" part led me to believe there might be multiple template dirs.

scripts/make-npm-packages/src/index.ts
```diff
@@ -0,0 +1,82 @@
...
+import { ARGS_OPTIONS, ArgsSchema } from "./schema/args.ts";
+import { BuildDataSchema } from "./schema/input-data.ts";
+
+(function main() {
+  const args = parseArgs({ strict: true, options: ARGS_OPTIONS });
+
+  if (args.values.help) {
+    console.log("Available options:", ARGS_OPTIONS);
+    return;
+  }
```
infomiho: Nitpick: we might not need the `--help` option since we'll always run this script in the CI.

scripts/make-npm-packages/src/index.ts
```diff
@@ -0,0 +1,82 @@
...
+    const {
+      fileName: inputTarballPath,
+      target: [os, cpu, libc],
+    } = tarballData;
+
+    console.group("Creating subpackage");
+    console.log("Input tarball path:", inputTarballPath);
+    console.log("Target:", { os, cpu, libc });
+
+    const subPackageName = makeSubPackageName(os, cpu, libc || "unknown");
```
infomiho: We need to encode the `"unknown"` for sure now that I see it for the third time.

scripts/make-npm-packages/src/index.ts
```diff
@@ -0,0 +1,82 @@
+import slugify from "@sindresorhus/slugify";
+import fs from "fs-extra";
+import * as path from "node:path";
+import { parseArgs } from "node:util";
+import { makeMainPackage } from "./generator/main-package.ts";
+import { makeSubPackage } from "./generator/sub-package.ts";
+import { ARGS_OPTIONS, ArgsSchema } from "./schema/args.ts";
+import { BuildDataSchema } from "./schema/input-data.ts";
+
+(function main() {
```
infomiho: Why do we need the immediately invoked fn? We don't do anything async, but even then I think we can just call `await` on module level. Since this is a script, I think top level module code is fine.

scripts/make-npm-packages/templates/main-package/bin.js
```diff
@@ -0,0 +1,123 @@
...
+  debug("Loading main package data...");
+
+  /** @type {{ default: import("../../src/schema/output-data.ts").MainPackageData }} */
+  const { default: data } = await import(
+    // @ts-expect-error This file does not exist at dev time, only in the built package.
+    "./data.json",
+    { with: { type: "json" } }
+  );
+
+  debug("Loaded main package data:", JSON.stringify(data, null, 2));
```
infomiho: Even though this debug log is never printed - do we still execute the `JSON.stringify` bit each time?

scripts/make-npm-packages/templates/main-package/CLIError.js
```diff
@@ -0,0 +1,38 @@
+// @ts-check
+
+/*
+This Error wrapper exists to ensure a consisten human-friendly error reporting
```
infomiho:
```suggestion
This Error wrapper exists to ensure a consistent human-friendly error reporting
```

scripts/make-npm-packages/readme.md
```diff
@@ -0,0 +1,91 @@
+# make-npm-packages
```
infomiho, line 1: It would be cool if we share the workflow with others: a blog post with code examples. It would be an evergreen post and probably interesting for HN audience 🤷

scripts/make-npm-packages/readme.md
```diff
@@ -0,0 +1,91 @@
...
+2. **Sub-packages** - Platform-specific packages containing the actual executables
+
+The main package uses npm's `optionalDependencies` to install the correct platform-specific sub-package based on the user's system. The CLI wrapper then detects and runs the appropriate executable.
+
+## Architecture
+
+### Main Package
+
+- Contains a CLI wrapper (`bin.js`) that detects the installed platform-specific sub-package
+- Uses `Promise.any()` to load the first available sub-package. Only fails if none are found.
```
infomiho: We need to regenerate the README 😄

.github/actions/compute-artifact-names/action.yaml
cprecioso: Now `all` is the special parameter value we give when we want to get back a wildcard that works for all of the artifacts. We do need to insert that wildcard (`*`) into the name at some point.

scripts/make-npm-packages/src/schema/util.ts
cprecioso: you're right, i don't know what i was thinking 😂

scripts/make-npm-packages/src/generator/main-package.ts
cprecioso: > why not use [Zod] for PackageJson? 

We don't need to read or act on any of the existing data in the templates, so I don't feel the need to parse whatever it might have with Zod. We're just adding stuff.

> the PackageJson type probably contains more fields than we have in our template

Yeah, but they're all marked as optional, no? We want to create a valid PackageJson file so I'm just using this type to get autocomplete and linting. I don't feel a concern in this case.

scripts/make-npm-packages/src/common.ts
cprecioso: I mean, this is not really a template, but the folder that contains the actual templates for each package. I'll rename it to TEMPLATES_DIR

scripts/make-npm-packages/src/index.ts
cprecioso: I'll keep it just in case, it's really not maintenance intensive to have.

scripts/make-npm-packages/src/index.ts
cprecioso: We do a `return` a couple of lines below. I wanted to do an early return instead of a `process.exit` since the latter doesn't really allow Node.js to exit cleanly.

Anyway, I removed the top level function and moved everything into a big if.

scripts/make-npm-packages/templates/main-package/bin.js
cprecioso: Yeah, you're correct. I was just debugging this myself so didn't care much about this initially. But now we're keeping it, so I migrated to using the printf-style arguments in debuglog so it shouldn't be stringify-ing each time.

waspc/waspc.cabal
cprecioso: Yep, something weird. I'll see if i can fix it.

scripts/make-npm-packages/src/index.ts
```diff
@@ -0,0 +1,82 @@
...
+import { makeMainPackage } from "./generator/main-package.ts";
+import { makeSubPackage } from "./generator/sub-package.ts";
+import { ARGS_OPTIONS, ArgsSchema } from "./schema/args.ts";
+import { BuildDataSchema } from "./schema/input-data.ts";
+
+const args = parseArgs({ strict: true, options: ARGS_OPTIONS });
+
+if (args.values.help) {
+  console.log(`Available options: ${ARGS_OPTIONS}`);
+} else {
```
infomiho, line 14: Nitpick: early `process.exit(0)` maybe?

scripts/make-npm-packages/src/schema/output-data.ts
```diff
@@ -0,0 +1,23 @@
+export interface SubPackageInfo {
+  packageName: string;
+}
+
+// This can be any string representing a lack of a defined libc for the build
+// (e.g. macOS or Windows don't use libc in the same way Linux does).
+// The usual convention is to call it "unknown".
+export const noLibcName = "unknown" as const;
```
infomiho: Nitpick: I think we should go with `NO_LIBC_NAME` since it's a constant. And it will be more clear in places of usage. 

Also, `UNDEFINED_LIBC_NAME` or `UNUSED_LIBC_TARGET_NAME` or `FALLBACK_LIBC_NAME` maybe explain better what is the purpose. 

RAW

scripts/make-npm-packages/src/index.ts
cprecioso: I mentioned here https://github.com/wasp-lang/wasp/pull/3464#discussion_r2619936703 I usually avoid doing `process.exit()` since it doesn't allow Node to do cleanup.
I know this is not a real problem here, but I'd prefer to keep it away if possible.

PR #3465, opened 2025-12-01
Add CLI tests to `waspc/e2e-tests`

FranjoMindek:
### What are we adding?

Before this PR, `waspc/e2e-tests` only had snapshot tests. They function in a way where we want to keep their artifacts (snapshots) as part of the git-tracked files. This PR adds **regular CLI tests** that validate command execution without persisting any artifacts.

Before going to deep into the solution be aware that we are aiming to refactor the whole `waspc/e2e-tests` by moving them away from the current unix shell commands approach. Instead we will write all of our logic in pure Haskell. This is to make them cross-platform compatible. That is why the tests which would require complex unix shell command implementations were only partially implemented. As the unix shell commands logic will be dropped, they are not worth the time investment. 
More on cross platform compatible e2e tests here: https://github.com/wasp-lang/wasp/issues/3404

In short, we want a net-positive gain from these tests, not complete coverage. Standard good code quality rules still apply.

The CLI commands table, which the implementation was based on can be found here: https://www.notion.so/wasp-lang/CLI-Testing-24518a74854c8051b0fef88533592504?source=copy_link
In there we mention which commands we are testing, which we are not, and why.

Note, as we now have additional variant of tests besides the snapshot tests, I decided to refactor the structure so that the snapshot tests logic is moved under the `waspc/e2e-tests/SnapshotTest` folder.
Changes related to moving snapshot tests folder are in a separate PR: https://github.com/wasp-lang/wasp/pull/3468

**New Directory structure:**
```
e2e-tests/
├── SnapshotTest/          # Snapshot-based tests (compare golden vs current output)
│   ├── snapshots/         # Git-tracked golden files
│   └── *.hs               # Snapshot test modules
├── Test/                  # Regular tests (validate exit codes, no persisted artifacts)
│   ├── test-outputs/      # Temporary test outputs (gitignored)
│   └── *.hs               # Test modules
├── SnapshotTest.hs        # Snapshot test framework
├── Test.hs                # Regular test framework
└── Main.hs                # Entry point
```

### How new tests work

Tests consist of a name (which creates a same-named directory under `Test/test-outputs/`) and test cases. Each test case is a unix shell command executed **sequentially**. This is different from snapshot tests which only ever have a single unix shell command. This approach makes it much easier to build tests (since we can split them into steps), and also allows validating individual test cases separately. Validity is determined by checking exit status.

image

Sequentiality is enforced by executing the unix shell commands while creating `Hspec`'s `Spec`. The captured exit status is translated into SUCCESS/FAIL after running all of the tests. This is because I couldn't force Tasty to run them in sequence even while using the [`sequential` function](https://hackage-content.haskell.org/package/hspec-2.11.14/docs/Test-Hspec.html#v:sequential).

~~Blocked by https://github.com/wasp-lang/wasp/pull/3480 which enables us to use the `wasp db reset` command non-interactively. This PR assumes the above PR is merged.~~

Fixes: https://github.com/wasp-lang/wasp/issues/1963

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change**

### DIFF
--- a/waspc/e2e-tests/SnapshotTest.hs
+++ b/waspc/e2e-tests/SnapshotTest.hs
@@ -98,7 +98,8 @@ executeSnapshotTestCommand snapshotTest snapshotDir = do
         }
 
 generateSnapshotFileListManifest :: Path' Abs (Dir SnapshotDir) -> Path' Abs (File SnapshotFileListManifestFile) -> IO ()
-generateSnapshotFileListManifest snapshotDir snapshotFileListManifestFile = getSnapshotFilesForExistenceCheck >>= writeSnapshotFileListManifest
+generateSnapshotFileListManifest snapshotDir snapshotFileListManifestFile =
+  getSnapshotFilesForExistenceCheck >>= writeSnapshotFileListManifest
   where
     getSnapshotFilesForExistenceCheck :: IO [Path' Abs (File SnapshotFile)]
     getSnapshotFilesForExistenceCheck =
@@ -115,7 +116,8 @@ generateSnapshotFileListManifest snapshotDir snapshotFileListManifestFile = getS
     -- Creates a deterministic manifest of files that should exist in the snapshot.
     -- File paths are normalized to relative paths and sorted.
     writeSnapshotFileListManifest :: [Path' Abs (File SnapshotFile)] -> IO ()
-    writeSnapshotFileListManifest = writeFile (SP.fromAbsFile snapshotFileListManifestFile) . unlines . sort . map normalizeFile
+    writeSnapshotFileListManifest =
+      writeFile (SP.fromAbsFile snapshotFileListManifestFile) . unlines . sort . map normalizeFile
       where
         normalizeFile = makeRelative (SP.fromAbsDir snapshotDir) . SP.fromAbsFile
 
@@ -150,7 +152,9 @@ getNormalizedSnapshotFilesForContentCheck snapshotDir = do
     formatPackageJsonFiles = mapM_ formatPackageJsonFile . filter isPackageJsonFile
       where
         formatPackageJsonFile :: Path' Abs (File file) -> IO ()
-        formatPackageJsonFile packageJsonFile = BS.readFile (SP.fromAbsFile packageJsonFile) >>= BSL.writeFile (SP.fromAbsFile packageJsonFile) . formatJson . unsafeDecodeJson
+        formatPackageJsonFile packageJsonFile =
+          BS.readFile (SP.fromAbsFile packageJsonFile)
+            >>= BSL.writeFile (SP.fromAbsFile packageJsonFile) . formatJson . unsafeDecodeJson
 
         isPackageJsonFile :: Path' Abs (File file) -> Bool
         isPackageJsonFile = equalFilePath "package.json" . takeFileName . SP.fromAbsFile
@@ -173,7 +177,8 @@ defineSnapshotTestCases ::
   Path' Abs (Dir SnapshotDir) ->
   [Path' Abs (File SnapshotFile)] ->
   [TestTree]
-defineSnapshotTestCases currentSnapshotDir goldenSnapshotDir = map defineSnapshotTestCase
+defineSnapshotTestCases currentSnapshotDir goldenSnapshotDir currentSnapshotFiles =
+  map defineSnapshotTestCase currentSnapshotFiles
   where
     defineSnapshotTestCase :: Path' Abs (File SnapshotFile) -> TestTree
     defineSnapshotTestCase currentSnapshotFile =


Martinsos:
Cool! I won't get into proper reviewing if others have time because I don't have enough time do to it properly, but I would potentially recommend @cprecioso as a reviewer, he is new to this piece of code and will both benefit from learning about it + offering a fresh perspective.

My only quick feedback is, do we have to call them "ephemeral"? Can't they just be named "tests", and "snapshot" tests are then kind of a specialization? Ephemeral sounds like there is sometihng unusually ephemeral happening, but then one realized "oh ok they are just normal tests" hah, so would be nice to avoid that "misdirection". I guess you maybe did so because you hvae Tests and then you have SnapshotTests and you needed to also somehow name these guys, but maybe that is then design issue (or maybe not).

Sequential part -> I hope when you said they have to execute sequentially, you meant a single test calls its content sequentially, not that all of them have to wait for each other? If actually all tests have to execute sequentially, that kind of sucks, doesn't it? Would be good to understand better in that case if we can not hvae that limitation.

Ok this is my quick take, I hope someobdy else can take this one over.

cprecioso:
Hey, didn't take a look yet, just a heads-up that I won't be able to fit this review before I go on vacation. I can do it when I come back.

FranjoMindek:
> My only quick feedback is, do we have to call them "ephemeral"? Can't they just be named "tests", and "snapshot" tests are then kind of a specialization? Ephemeral sounds like there is sometihng unusually ephemeral happening, but then one realized "oh ok they are just normal tests" hah, so would be nice to avoid that "misdirection". I guess you maybe did so because you hvae Tests and then you have SnapshotTests and you needed to also somehow name these guys, but maybe that is then design issue (or maybe not).

Yes. It's simply giving them a name. I could name them just `Tests` instead. Though that makes me feel like it's something which should also affect snapshot tests, but it does not. It's a separate variant.

> Sequential part -> I hope when you said they have to execute sequentially, you meant a single test calls its content sequentially, not that all of them have to wait for each other? If actually all tests have to execute sequentially, that kind of sucks, doesn't it? Would be good to understand better in that case if we can not hvae that limitation.

Sadly they are fully sequential because of `Sequentiality is enforced by executing the unix shell commands while creating the Hspec's Spec`. Because spec creation is not done in parallel, their execution ends up sequential. I only ever managed to do everything in parallel, or do everything sequentially, didn't manage to run the ephemeral test in parallel where their cases are run in sequence. Even when I use `Hspec`'s `sequential` execution, they start the next shell command before the previous one was finished.

Do note that currently even snapshot tests are executed sequentially (the project generation part, which is also the most expensive part time wise, as it's done sequentially before running the `TestTree`). 

So we do have a lot to gain by optimizing both of them.

> Hey, didn't take a look yet, just a heads-up that I won't be able to fit this review before I go on vacation. I can do it when I come back.

That is fine, I still have to fix the issue that interactive tests (e.g. `wasp db reset` or `wasp new`) don't work on github runners.

cprecioso:
Maybe "Regression tests" is what we want?

Martinsos:
I don't think "regression tests" term is used in this way. I would just call them Tests then and that is ok, they are just normal, general tests.

waspc/e2e-tests/snapshots/kitchen-sink-golden/wasp-app/migrations/20251127110727_add_on_after_email_verified_hook_call_count/migration.sql
cprecioso, line 1: Is this file no longer needed?

waspc/e2e-tests/Main.hs
```diff
@@ -15,13 +29,55 @@ main = do
...
         waspBuildSnapshotTest,
         waspMigrateSnapshotTest,
         kitchenSinkSnapshotTest
       ]
+  ephemeralTests <-
+    mapM
+      runEphemeralTest
+      [ -- general Wasp commads
+        waspNewMinimalEphemeralTest,
+        waspNewMinimalInteractiveEphemeralTest,
```
cprecioso: If not doing tasty-discover: Instead of manually enumerating the tests here, can we make this file export an array of its tests?

waspc/e2e-tests/Main.hs
```diff
@@ -1,3 +1,17 @@
...
+import EphemeralTest.WaspCleanEphemeralTest (waspCleanEphemeralTest)
+import EphemeralTest.WaspCompileEphemeralTest (waspCompileEphemeralTest)
+import EphemeralTest.WaspCompletionEphemeralTest (waspCompletionEphemeralTest)
+import EphemeralTest.WaspDbMigrateDevEphemeralTest (waspDbMigrateDevEphemeralTest)
+import EphemeralTest.WaspDbResetEphemeralTest (waspDbResetEphemeralTest)
+import EphemeralTest.WaspDbSeedEphemeralTest (waspDbSeedEphemeralTest)
+import EphemeralTest.WaspDepsEphemeralTest (waspDepsEphemeralTest)
+import EphemeralTest.WaspDockerfileEphemeralTest (waspDockerfileEphemeralTest)
+import EphemeralTest.WaspInfoEphemeralTest (waspInfoEphemeralTest)
+import EphemeralTest.WaspNewEphemeralTest (waspNewBasicEphemeralTest, waspNewBasicInteractiveEphemeralTest, waspNewMinimalEphemeralTest, waspNewMinimalInteractiveEphemeralTest, waspNewSaasEphemeralTest, waspNewSaasInteractiveEphemeralTest)
```
cprecioso: If not doing tasty-discover: I'd split this import in multiple lines

waspc/e2e-tests/EphemeralTest/WaspCompletionEphemeralTest.hs
```diff
@@ -0,0 +1,34 @@
+module EphemeralTest.WaspCompletionEphemeralTest (waspCompletionEphemeralTest) where
+
+import EphemeralTest (EphemeralTest, makeEphemeralTest, makeEphemeralTestCase)
+import ShellCommands (ShellCommand, ShellCommandBuilder, (~&&))
+
+waspCompletionEphemeralTest :: EphemeralTest
+waspCompletionEphemeralTest =
```
cprecioso: Oh, nice job adding tests for this!

waspc/e2e-tests/Main.hs
```diff
@@ -15,13 +29,55 @@ main = do
 
 -- TODO: Investigate automatically discovering the tests.
```
cprecioso, line 30: I think it might be worth it to investigate `tasty-discover` in this PR, as we're adding a way longer list of tests.

waspc/e2e-tests/EphemeralTest.hs
cprecioso, line 1: Okay, so I thought a bit about this, and I think I'd scrap the "EphemeralTest" name, it's really cumbersome and not a convention.

What I see here is that we're doing E2E tests, and some of those happen to use snapshots. That's it, I wouldn't explicitly differentiate between the two. In any case, I'd add a "Snapshot" suffix to the other ones, and leave these ones without any specific names, if you want to mark those as a special case of the others.

FranjoMindek:
Removed the notion of "ephemeral tests".

waspc/e2e-tests/Test/FileSystem.hs
cprecioso, line 1: re: this file and `ShellCommand`. By being in the `Test` folder, they just look like two more test suits. I'd move them to a Common folder or something similar. I'd have the final dir tree something like:

- e2e-tests
	- Tests/
		- (all the regular tests)
		- Snapshot/
			- (all the snapshot tests)
			- _if it's not feasible to have it nested, we can keep it as it is now (e2e-tests/SnapshotTests)_
	- Src/
		- FileSystem.hs
		- ShellCommand.hs
		- SnapshotTest/
			- FileSystem.hs
			- ShellCommands.hs 
		- WaspProject/
			- FileSystem.hs
			- ShellCommands.hs 

You might also look into consolidating the `FileSystem` and `ShelCommand` from the different places (only if it makes sense, you'll know best)

waspc/e2e-tests/Test/FileSystem.hs
```diff
@@ -0,0 +1,29 @@
...
+
+-- | The directory where all 'TestDir' are created.
+data TestOutputsDir
+
+getTestDir :: String -> IO (Path' Abs (Dir TestDir))
+getTestDir testName = (</> (testsOutputsDirInTestsDir </> testDirInTestOutputsDir testName)) <$> getE2eTestsDir
+
+getTestOutputsDir :: IO (Path' Abs (Dir TestOutputsDir))
+getTestOutputsDir = (</> testsOutputsDirInTestsDir) <$> getE2eTestsDir
+
```
cprecioso: nitpick: i'd use `<&>` instead of `<$>` here, as otherwise the reading the code mixes RTL and LTR styles

waspc/e2e-tests/Test/WaspBuildStartTest.hs
```diff
@@ -0,0 +1,34 @@
...
+        waspCliBuildStartFails,
+      makeTestCase
+        "Setup: Create Wasp project from minimal starter"
+        (createTestWaspProject Minimal),
+      makeTestCase
+        "Setup: Modify Wasp project to use Postgresql"
+        (withInTestWaspProjectDir [setWaspDbToPSQL]),
+      makeTestCase
+        "Should fail inside of a unbuilt Wasp project"
+        (withInTestWaspProjectDir [waspCliBuildStart "-s DATABASE_URL=none"]),
```
cprecioso: i don't think this test is failing

waspc/e2e-tests/Test/WaspCleanTest.hs
```diff
@@ -0,0 +1,39 @@
...
+      makeTestCase
+        "Assert `.wasp` directory does not exist"
+        (withInTestWaspProjectDir [assertDirectoryDoesNotExist ".wasp"]),
+      makeTestCase
+        "Assert `node_modules` directory does not exist"
+        (withInTestWaspProjectDir [assertDirectoryDoesNotExist "node_modules"])
+    ]
+  where
+    waspCliCleanFails :: ShellCommandBuilder context ShellCommand
+    waspCliCleanFails = return "! wasp-cli clean"
```
cprecioso: we're using this pattern a lot, let's extract it to a `fails` helper or similar

waspc/e2e-tests/Test/ShellCommands.hs
```diff
@@ -0,0 +1,34 @@
...
+import StrongPath (Abs, Dir, Path', fromAbsDir)
+import Test.FileSystem (TestDir)
+import WaspProject.ShellCommands (WaspProjectContext (..))
+
+data TestContext = TestContext
+  { testDir :: Path' Abs (Dir TestDir),
+    waspProjectContext :: WaspProjectContext
+  }
+
+withInTestWaspProjectDir ::
```
cprecioso: nitpick: within is a single word


```suggestion
withinTestWaspProjectDir ::
```

cprecioso: you can even reduce it


```suggestion
inTestWaspProjectDir ::
```

waspc/e2e-tests/Main.hs
```diff
@@ -5,23 +5,87 @@ import SnapshotTest.WaspCompileSnapshotTest (waspCompileSnapshotTest)
...
+      runTest
+      [ -- general Wasp commads
+        waspNewMinimalTest,
+        waspNewMinimalInteractiveTest,
+        waspNewBasicTest,
+        waspNewBasicInteractiveTest,
+        -- Note(Franjo): These will fail because there "dev" opensaas template tag doesn't exist.
+        -- We can't really test the `saas` tempalte in  `waspc` e2e tests.
+        -- FIXME: waspNewSaasTest,
+        -- FIXME: waspNewSaasInteractiveTest,
```
cprecioso: I'd leave the `FIXME`s, remove your notes, and just put a comment on top that says " -- TODO: Move to Haskell tests so that we can test everything (#issue_number)"

waspc/e2e-tests/README.md
```diff
@@ -1,21 +1,48 @@
...
 Every command is treated as a black box.
 
-The main **outputs** of the `waspc` is a Wasp application.
-We want to validate that CLI commands correctly generate or modify applications.
+The main **outputs** of `waspc` are Wasp applications.
+We want to validate that CLI commands correctly generate or modify Wasp applications.
 In addition to applications, we also cover secondary outputs, such as the installation and uninstallation of the CLI itself, `bash` completions, and more.
 
-## Snapshot tests
+## Test varaints
```
cprecioso:
```suggestion
## Test variants
```

waspc/e2e-tests/README.md
```diff
@@ -1,21 +1,48 @@
...
+`waspc e2e-tests` consist of two different tests variants:
+
+1. **Tests**: tests whose output we don't need to save
+2. **Snapshot Tests**: tests whose output we want to save
+
+### Tests
+
+We test Wasp CLI commands whose outputs we can discard.
+
+Tests **sequentially execute their test cases** in the `TestDir`.
```
cprecioso: If we make each case self-contained, this will no longer be an important note, so we can remove.

We'd note somewhere else that we're running tests sequentially and that running them in parallel is a FIXME.

waspc/e2e-tests/Test.hs
```diff
@@ -0,0 +1,114 @@
...
+  { name :: String,
+    testCases :: [TestCase]
+  }
+
+makeTest :: String -> [TestCase] -> Test
+makeTest testName testTestCases =
+  Test
+    { name = testName,
+      testCases = testTestCases
+    }
```
cprecioso: You're making a synonym of `Test` 😁, i don't think we need it

waspc/e2e-tests/Tests/WaspBuildStartTest.hs
cprecioso, line 1: Could we make it so the CLI tests are independent from each other, even if they're still sequential? that'd make sequential or parallel just a detail, and the new tests would work in both ways (there'd be some repetition but i think it's fine and can be extracted if too much)

so instead of 
```hs
waspBuildStartTest =
  makeTest
    "wasp-build-start"
    [ makeTestCase
        "Should fail outside of a Wasp project"
        waspCliBuildStartFails,
      makeTestCase
        "Setup: Create Wasp project from minimal starter"
        (createTestWaspProject Minimal),
      makeTestCase
        "Setup: Modify Wasp project to use Postgresql"
        (withInTestWaspProjectDir [setWaspDbToPSQL]),
      makeTestCase
        "Should fail inside of a unbuilt Wasp project"
        (withInTestWaspProjectDir [waspCliBuildStart "-s DATABASE_URL=none"]),
      makeTestCase
        "Setup: Build the Wasp project"
        (withInTestWaspProjectDir [waspCliBuild]),
      makeTestCase
        "Should succeed inside of a built Wasp project"
        (withInTestWaspProjectDir [waspCliBuildStart "-s DATABASE_URL=none"])
    ]
```

we'd have something like
```hs
waspBuildStartTest =
  makeTest
    "wasp-build-start"
    [ makeTestCase
        "Should fail outside of a Wasp project"
        (fails $ return "wasp-cli build start"),
      makeTestCase
        "Should fail inside of a unbuilt Wasp project"
        (sequence [
          createTestWaspProject Minimal,
          withInTestWaspProjectDir [
            fails $ waspCliBuildStart "-s DATABASE_URL=none"
          ]
        ]),
      makeTestCase
        "Should succeed inside of a built Wasp project"
        (sequence [
          createTestWaspProject Minimal,
          withInTestWaspProjectDir [
            setWaspDbToPSQL,
            waspCliBuild,
            waspCliBuildStart "-s DATABASE_URL=none"
          ]
        ])
    ]
```

waspc/e2e-tests/Test/FileSystem.hs
FranjoMindek: TIL about `<&>`

waspc/e2e-tests/Test/ShellCommands.hs
FranjoMindek: We do the same for snapshots, I will also fix that.

waspc/e2e-tests/Test.hs
FranjoMindek: Yep, they weren't before, but got reduced to them :D

waspc/e2e-tests/Test/WaspBuildStartTest.hs
FranjoMindek: Yes, this one shuld've been:
```hs
TestCase
  "fail-unbuilt-project"
  ( sequence
      [ createTestWaspProject Minimal,
        inTestWaspProjectDir
          [ setWaspDbToPSQL,
            return waspCliBuildStartFails
          ]
      ]
  ),
```

waspc/e2e-tests/Test/FileSystem.hs
FranjoMindek: I did move the tests, but I'll leave file system and shell commands stuff separated for now.

waspc/e2e-tests/Test/WaspCleanTest.hs
FranjoMindek: Sadly that is a bit hard to do due to types constraints.
Essentially we would have to `ShellCommandBuilder WaspProjectContext ShellCommand -> ShellCommandBuilder WaspTestContext ShellCommand` which we can't really do.

waspc/e2e-tests/Main.hs
FranjoMindek: Cleaned it up a bit.

waspc/e2e-tests/Test/FileSystem.hs
cprecioso: I'd still insist on grouping the testing functions a bit, right now if I go into the dir I have

- `e2e-tests/SnapshotTest`
- `e2e-tests/Test`
- `e2e-tests/Tests`
- `e2e-tests/SnapshotTest.hs`
- `e2e-tests/Test.hs`

It feels a bit messy, they can't be told apart at first glance. I'd still like a main separation of `tests` for test files, and `src` or `lib` or `utils` for the testing functions.

If you don't want to pollute the haskell module names, you can use [`hs-source-dirs`](https://cabal.readthedocs.io/en/stable/cabal-package-description-file.html#pkg-field-hs-source-dirs) in `waspc.cabal` to mark both the `tests` and `src` folders as module roots.

waspc/e2e-tests/Tests/WaspBuildStartTest.hs
```diff
@@ -0,0 +1,40 @@
...
+import ShellCommands (ShellCommand, WaspNewTemplate (..))
+import Test (Test (..), TestCase (..))
+import Test.ShellCommands (createTestWaspProject, inTestWaspProjectDir)
+import WaspProject.ShellCommands (setWaspDbToPSQL, waspCliBuild, waspCliBuildStart)
+
+-- FIXME: @waspCliBuildStart@ - figure out long lasting processes
+waspBuildStartTest :: Test
+waspBuildStartTest =
+  Test
+    "wasp-build-start"
```
cprecioso, line 10: extra, optional: i'd love if we could have human names and we "slugified" them when writing the folder

cprecioso (approved):
Looks good! I wanted to insist in my previous comment, but apart from that, everything looks good.
Leaving approved

PR #3468, opened 2025-12-02
Move snapshot tests to their subdirectory

FranjoMindek:
## Description

Part of https://github.com/wasp-lang/wasp/pull/3465.
I am moving `snapshots` folder to the `SnapshotTest` directory, so that the each testing domain (snapshot tests, ephemeral tests) is self-contained with all of the assets they use/generate.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change**


waspc/e2e-tests/SnapshotTest/FileSystem.hs
```diff
@@ -19,8 +19,7 @@ where
 import Data.Maybe (fromJust)
 import FileSystem (E2eTestsDir, GitRootDir, e2eTestsDirInWaspcDir, getE2eTestsDir, waspcDirInGitRootDir)
 import SnapshotTest.Snapshot (SnapshotType (..))
-import StrongPath (Dir, File, Path, Path', Rel, castDir, parseRelDir, reldir, relfile, (</>))
-import StrongPath.Types (Abs)
+import StrongPath
```
cprecioso: I don't think we usually do bare imports, but RAW

PR #3469, opened 2025-12-02
CI: Test npm packages

cprecioso:
On top of #3464, we add a CI step to test that the npm packaging worked correctly.

We download the built packages, setup a local npm registry with [Verdaccio](https://verdaccio.org/), upload the packages, and run `npx @wasp.sh/wasp-cli version`.

### DIFF
--- a/.github/workflows/ci-npm-package-test.yaml
+++ b/.github/workflows/ci-npm-package-test.yaml
@@ -32,6 +32,7 @@ jobs:
       - uses: actions/download-artifact@v6
         with:
           name: wasp-npm-packages
+          path: packages/
 
       - name: Install dependencies
         if: matrix.runner.install-deps
@@ -73,7 +74,7 @@ jobs:
           npm config set registry "$TEST_REGISTRY_URL"
 
           # Publish all packages to the test registry
-          for pkg_dir in ./wasp*/; do
+          for pkg_dir in ./packages/*; do
             npm publish "$pkg_dir"
           done
 


.github/workflows/ci-npm-package-test.yaml
```diff
@@ -0,0 +1,75 @@
...
+            max_body_size: 1000mb # Needed because Wasp packages are large
+            packages: { '**': { access: $all, publish: $all } }
+        run: |
+          set -eux
+
+          # Install global packages
+          npm install -g npm-cli-login@1 verdaccio@6
+
+          # Create registry config
+          mkdir -p "$HOME/.config/verdaccio/"
```
infomiho:
```suggestion
          mkdir -p "$HOME/.config/verdaccio"
```
Nitpick

.github/workflows/ci-npm-package-test.yaml
```diff
@@ -0,0 +1,75 @@
...
+          until curl -sSf "$TEST_REGISTRY_URL/-/ping" >/dev/null; do
+            sleep 1
+          done
+
+          # Set registry and login
+          npm-cli-login -u fake-user -p fake-password -e fake@email.com -r "$TEST_REGISTRY_URL"
+          npm config set registry "$TEST_REGISTRY_URL"
+
+          # Publish all packages to the test registry
+          for pkg_dir in ./wasp*/; do
```
infomiho: Where does this `wasp` dir come from? I guess the artifacts are downloaded to that dir? Can we somehow encode it somewhere e.g. env var e.g. `DOWNLOADED_ARTIFACTS_DIR` or something like that.

infomiho:
@cprecioso One this I noticed in the test, the package is uploaded with the Wasp Studio client `node_modules`?

```
npm notice 1.2kB data/packages/studio/client/node_modules/@aashutoshrathi/word-wrap/index.d.ts
```

cprecioso:
> @cprecioso One this I noticed in the test, the package is uploaded with the Wasp Studio client `node_modules`?

Yep! I saw but it didn't really click that we should not have it 🤦‍♂️, since it comes from the workflow we already had that builds the artifacts. I am sending a fix for it: https://github.com/wasp-lang/wasp/pull/3483

cprecioso:
@infomiho I fixed the node_modules situaiton (https://github.com/wasp-lang/wasp/pull/3483) and rebased this on top of it. So, this should be fixed.

Issue #3127, opened 2025-09-05
Publish CLI on npm


cprecioso:
### Description

- Wasp distributed as an npm package, no more global installers
  - Goal: `npm install -g @wasp.sh/cli`
  - Maybe: per-project CLI as a dependency, version declaring and locking for free
  - We can look at how other packages that require native binaries do it, like `rollup`, `rolldown`, `esbuild`, etc
  - Non-goal: Wasp being used programmatically, or changing our build to generate something else than a CLI binary.

While we can have multiple approaches (global `curl | sh` installer, npm package), we should choose one and recommend that one first in our documentation:

- Pros of CLI as a npm package:
	- No need for our own installer
	- Installed at the same time as other dependencies
	- No need for global installations (although it remains a possibility)
	- Ability to lock and upgrade dependencies with the same workflows as npm
	- Feels more native to the ecosystem
- Cons of CLI as a npm package:
	- If not installed globally, you’ll need to prefix the commands like `npx wasp`
	  - FWIW there are ways of having a good experience while keeping the locally pinned approach: have the global binary search if there’s a local binary in the current project and if so defer to it. (pioneered by gulp ca. 2013)
	- node/npm is a prerequisite to install (right now is a requisite to run, but not to install)

### Steps

- [ ] Investigate prior art
- [ ] Craft a package/collection of packages and check that they work
- [ ] Modify our release code to build the packages
- [ ] (Optional) Automatically publish the packages on release (check [publishing to npm with OIDC](https://github.blog/changelog/2025-07-31-npm-trusted-publishing-with-oidc-is-generally-available/))
- [ ] (Optional) Add telemetry to the package(s)
- [ ] (Optional) Allow a global package installation to act as a proxy to a local one if detected
- [ ] As a team, agree on the preferred method of installation
- [ ] Update documentation
- [ ] As a team, discuss: what do we do with `main.wasp#version` if we lock the version through `npm`?

Martinsos:
Data point: Claude Code allows both native install and npm install and some others:

<img width="964" height="946" alt="Image" src="https://github.com/user-attachments/assets/f6095f5e-1f7c-4f9a-b002-1d14742c4ce1" />

cprecioso:
Migration completed with Wasp 0.21
https://github.com/wasp-lang/wasp/releases/tag/v0.21.0

PR #3470, opened 2025-12-03
Add `./run get-version`

cprecioso:
## Description

For https://github.com/wasp-lang/wasp/issues/3127, I'll need to query the current waspc.cabal version from the CI, so I added a `./run` command to do so.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/run
+++ b/waspc/run
@@ -75,7 +75,7 @@ TEST_CMD="cabal test && \
 RUN_CMD="cabal --project-dir=${PROJECT_ROOT} run wasp-cli -- ${ARGS[@]}"
 
 # Utility
-GET_VERSION_CMD="cabal repl waspc -v0 <<<'putStrLn $ Data.Version.showVersion Paths_waspc.version'"
+GET_WASPC_VERSION_CMD="cabal repl waspc -v0 <<< 'putStrLn $ Data.Version.showVersion Paths_waspc.version'"
 GHCID_CMD="ghcid --command=cabal repl"
 # HLS version below should be one that works with the GHC version from the
 # `with-compiler` field in `cabal.project`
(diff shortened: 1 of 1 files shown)


waspc/run
```diff
@@ -96,7 +97,7 @@ ORMOLU_CHECK_CMD="$ORMOLU_BASE_CMD --mode check "'$'"(git ls-files '*.hs' '*.hs-
 ORMOLU_FORMAT_CMD="$ORMOLU_BASE_CMD --mode inplace "'$'"(git ls-files '*.hs' '*.hs-boot')"
 
 echo_and_eval() {
-  echo -e $"${LIGHT_CYAN}Running:${DEFAULT_COLOR}" "$1" "\n"
+  echo -e $"${LIGHT_CYAN}Running:${DEFAULT_COLOR}" "$1" "\n" 1>&2
```
cprecioso, line 100: Conventionally, human-readable info messages get written to stderr and not stdout.

Otherwise, when we do `my_var=$(./run get-version)`, the contents of `$my_var` would be `Running: blah\n0.19.0` insted of `0.19.0`

waspc/run
```diff
@@ -75,6 +75,7 @@ TEST_CMD="cabal test && \
 RUN_CMD="cabal --project-dir=${PROJECT_ROOT} run wasp-cli -- ${ARGS[@]}"
 
 # Utility
+GET_VERSION_CMD="cabal repl waspc -v0 <<<'putStrLn $ Data.Version.showVersion Paths_waspc.version'"
```
cprecioso: We use a simple inline script, passing it to Cabal REPL. I found no better way to do this, but I think it's somewhat elegant.

waspc/run
```diff
@@ -176,6 +177,8 @@ print_usage() {
   echo ""
   print_usage_cmd "ghcup-set" \
     "Sets the correct version of GHC and Cabal for this project (via GHCup)."
+  print_usage_cmd "get-version" \
```
Martinsos: Maaaaybe I would name it `waspc-version`, I think it is clearer what it does then. RAW though.

waspc/run
Martinsos: Nice, sounds like the best solution to me! Good idea. I love `cabal repl` in general, so useful.

PR #3471, opened 2025-12-03
Generating routes for user-land Vite in user's dir

infomiho:
This PR is not for reviewing. It might serve for discussion later on.

What was done:
- [x] Generates the routes in user-land in `router.generated.tsx`
- [x]  Uses the generated routes in the `getWaspApp` helper that returns the `WaspApp` component that needs to be mounted in the `index.tsx`
- [x] Alongside the routes changes, we also import the `app.client.rootComponent` and the `app.client.setupFn` directly in `index.tsx`

Part of #3135 #3472


infomiho:

Issue #3233, opened 2025-10-07
Add Wasp SDK to npm Workspaces


cprecioso:
Based on https://github.com/wasp-lang/wasp/pull/3159

The Wasp SDK is not currently added as a workspace to the user project.

If we did it, we could remove the specific `npm install` step from the SDK, and have even stronger dependency sharing.

The blocker is #2482, because when npm resolves the dependencies there's one deep in the tree that asks for `@types/react@19` and we only have `@types/react@18`, so it errors out.

infomiho:
I believe that we can remove one specific `@prisma/client` hack after we implement this: https://github.com/wasp-lang/wasp/pull/3479

PR #3479, opened 2025-12-05
Move server auth deps from SDK to server package.json

infomiho:
## Description

Quick refactor since we no longer need to keep the server auth deps in the SDK `package.json` ever since we introduced npm workspaces.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:


infomiho:
The comment says:
```hs
-- Server auth deps must be installed in the SDK because "@lucia-auth/adapter-prisma"
-- lists prisma/client as a dependency.
-- Installing it inside .wasp/out/server/node_modules would also
-- install prisma/client in the same folder, which would cause our
-- runtime to load the wrong (uninitialized prisma/client).
```

After I moved the deps from SDK to server, the same problem happened but in a different place.

The old problem was that we had the `server` package and the SDK package which would both install `@prisma/client`. We got rid of the `server` install by moving its auth deps to the SDK.

The new problem is similar, the `server` is now part of the npm workspace, so its deps get installed in the root project. But we still have the SDK which is outside of the npm workspace and we still have two instances of `@prisma/client`.

I believe these deps should be moved to the server deps after the SDK becomes a part of the workspaces. This issue won't happen then. Check out #3233

PR #3480, opened 2025-12-07
Add the `--force` flag to enable non-interactive db reset

FranjoMindek:
## Description

Adds the `--force` flag option to `wasp db reset` command, which just relays the `--force` flag to the [prisma CLI](https://www.prisma.io/docs/orm/reference/prisma-cli-reference#migrate-reset). This makes it reset the database without any interaction (no y/n needed).

I picked the option which stays in line with the prisma workflow the most.

Possible variations:
1. Pick a different flag name than prisma (not `--force`)
2. Making it non-interactive by default (would not go this route, breaking change also)

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
+++ b/waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
@@ -92,11 +92,7 @@ reset :: Path' Abs (Dir ProjectRootDir) -> ResetArgs -> J.Job
 reset projectRootDir resetArgs =
   runPrismaCommandAsJobFromWaspServerDir
     projectRootDir
-    prismaArgs
-  where
-    schema = projectRootDir </> dbSchemaFileInProjectRootDir
-    prismaArgs =
-      [ "migrate",
+    ( [ "migrate",
         "reset",
         "--schema",
         SP.fromAbsFile schema,
@@ -106,6 +102,9 @@ reset projectRootDir resetArgs =
         "--skip-seed"
       ]
         ++ if force resetArgs then ["--force"] else []
+    )
+  where
+    schema = projectRootDir </> dbSchemaFileInProjectRootDir
 
 -- | Runs `prisma db seed`, which executes the seeding script specified in package.json in
 --   prisma.seed field.


waspc/cli/exe/Main.hs
```diff
@@ -224,7 +224,7 @@ dbCli args = case args of
   -- These commands don't require an existing and running database.
   "start" : optionalStartArgs -> runCommand $ Command.Start.Db.start optionalStartArgs
   -- These commands require an existing and running database.
-  ["reset"] -> runCommandThatRequiresDbRunning Command.Db.Reset.reset
+  "reset" : optionalResetArgs -> runCommandThatRequiresDbRunning $ Command.Db.Reset.reset optionalResetArgs
```
FranjoMindek: I followed the pattern that `migrate-dev` command used.

waspc/cli/src/Wasp/Cli/Command/Db/Reset.hs
```diff
@@ -3,24 +3,45 @@ module Wasp.Cli.Command.Db.Reset
...
+  liftIO tryReset >>= \case
     Left errorMsg -> throwError $ CommandError "Database reset failed" errorMsg
-    Right () -> cliSendMessageC $ Msg.Success "Database reset successfully!"
+    Right () -> cliSendMessageC $ Msg.Success "Database reset successful."
+  where
+    tryReset = runExceptT $ do
+      resetArgs <- liftEither $ parseResetArgs optionalResetArgs
+      ExceptT $ dbReset genProjectDir resetArgs
+
+parseResetArgs :: [String] -> Either String ResetArgs
```
FranjoMindek: This is pretty much the same as `migrate-dev` one, the approach is extensible, i.e. assumes we might add more flags in the future.
If we don't think we will extend the `db reset` command, we can simplify this (we don't need recursion).

waspc/cli/src/Wasp/Cli/Command/Db/Reset.hs
```diff
@@ -3,24 +3,45 @@ module Wasp.Cli.Command.Db.Reset
...
+  resetDatabase optionalResetArgs genProjectDir
 
-  liftIO (dbReset genProjectDir) >>= \case
+resetDatabase :: [String] -> Path' Abs (Dir ProjectRootDir) -> Command ()
+resetDatabase optionalResetArgs genProjectDir = do
+  cliSendMessageC $ Msg.Start "Resetting the database..."
+  liftIO tryReset >>= \case
     Left errorMsg -> throwError $ CommandError "Database reset failed" errorMsg
-    Right () -> cliSendMessageC $ Msg.Success "Database reset successfully!"
+    Right () -> cliSendMessageC $ Msg.Success "Database reset successful."
```
infomiho:
```suggestion
    Right () -> cliSendMessageC $ Msg.Success "Database reset successfully."
```

`reset` was use here as a verb, not a noun so it's correct to use `successfully` - if you want to go with a noun, I'd then say `Database reset was successful`

waspc/cli/src/Wasp/Cli/Command/Db/Reset.hs
```diff
@@ -3,24 +3,45 @@ module Wasp.Cli.Command.Db.Reset
...
+      ExceptT $ dbReset genProjectDir resetArgs
+
+parseResetArgs :: [String] -> Either String ResetArgs
+parseResetArgs resetArgs = do
+  go resetArgs defaultResetArgs
+  where
+    go :: [String] -> ResetArgs -> Either String ResetArgs
+    go [] currentResetArgs = Right currentResetArgs
+    go ("--force" : rest) currentResetArgs = go rest $ currentResetArgs {_force = True}
+    go unknown _ = Left $ "Unknown reset arg(s): " ++ unwords unknown
```
infomiho: I'd suggest to use the `optparse-applicative` even though it's quite simple, it's a way to future-proof our arg parsing (we want to move it completely to `optparse-applicative` at some point).

Take a look at https://github.com/wasp-lang/wasp/blob/main/waspc/cli/src/Wasp/Cli/Command/BuildStart.hs#L26

Carlos built some nice parsing helpers which you can use 😄

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
```diff
@@ -88,15 +88,25 @@ migrateStatus projectRootDir =
...
+      [ "migrate",
+        "reset",
+        "--schema",
+        SP.fromAbsFile schema,
+        "--skip-generate",
+        -- NOTE(martin): We do "--skip-seed" here because I just think seeding
+        --   happening automatically on reset is too aggressive / confusing.
+        "--skip-seed"
+      ]
+        ++ forceArg
```
infomiho: I think it's okay to keep it inline, we didn't gain much by moving `prismaArgs` to the `where` block.

waspc/cli/src/Wasp/Cli/Command/Db/Reset.hs
FranjoMindek: Seems very nice, just confused as to why we send the name to `withArguments`, seems only for logging. 
That seems a bit out of place to me, but I won't touch it.

waspc/src/Wasp/Generator/DbGenerator/Jobs.hs
```diff
@@ -88,15 +88,24 @@ migrateStatus projectRootDir =
...
+      [ "migrate",
+        "reset",
+        "--schema",
+        SP.fromAbsFile schema,
+        "--skip-generate",
+        -- NOTE(martin): We do "--skip-seed" here because I just think seeding
+        --   happening automatically on reset is too aggressive / confusing.
+        "--skip-seed"
+      ]
+        ++ if force resetArgs then ["--force"] else []
```
infomiho: When I commented on this line, I thought you can move the whole `prismaArgs` where it was previously, but I won't block on this.

RAW

PR #3483, opened 2025-12-10
Update binary packager

cprecioso:
## Description

We discovered at https://github.com/wasp-lang/wasp/pull/3469#issuecomment-3626594696 that we are including `node_modules/**` files from our data packages in the Wasp CLI tarballs. This is because `make_binary_package.sh` was copying everything in the data folder, and not just the globs specified on the Cabal file.

Now we use `cabal sdist --list-only` to list which files are considered part of the code (so it excludes things like `node_modules`), and only copy those.

I also added `-x` to the script so we can trace its invocations from the CI logs.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/tools/make_binary_package.sh
+++ b/waspc/tools/make_binary_package.sh
@@ -18,20 +18,21 @@ CABAL_PROJECT_ROOT_PATH="$(cabal list-bin wasp-cli | sed s/\\/dist-newstyle.*//)
 WASP_BINARY_PATH="$(cabal list-bin wasp-cli)"
 cp "$WASP_BINARY_PATH" "$TMP_DIR/wasp-bin"
 
-# This pipeline of three commands copies the data files declared in the Cabal
-# file into the tmp dir to be packaged.
-#
-# 1. List all the files considered by Cabal to be part of the source (including
-# the declared data files).
-# 2. Filter to get only the files in the "data" dir.
-# 3. Copy the files in that list from the project root to the tmp dir to be
-# packaged.
-#
-# We use rsync to recreate the directory structure properly, plus it has a nice
-# `--files-from=-` option to read from stdin the list of files to be copied.
-cabal sdist --list-only \
-  | grep "^\./data/" \
-  | rsync -av --files-from=- "$CABAL_PROJECT_ROOT_PATH" "$TMP_DIR"
+CABAL_DATA_FILE_PATHS=$(
+  # List all the files considered by Cabal to be part of the source (including
+  # the declared data files).
+  cabal sdist --list-only \
+    | grep "^\./data/" # And now filter to only get the files in the "data" dir.
+)
+for data_file_path in $CABAL_DATA_FILE_PATHS; do
+  source_path="$CABAL_PROJECT_ROOT_PATH/$data_file_path"
+  dest_path="$TMP_DIR/$data_file_path"
+
+  # Make sure the parent directories in the output exist.
+  mkdir -p "$(dirname "$dest_path")"
+
+  cp "$source_path" "$dest_path"
+done
 
 tar -czf "$DST" -C "$TMP_DIR" .
 


waspc/tools/make_binary_package.sh
```diff
@@ -1,4 +1,6 @@
-#!/bin/sh -e
+#!/usr/bin/env bash
+
+set -euxo pipefail
```
sodic: What's this, what do the options mean? Do we need it?

waspc/tools/make_binary_package.sh
```diff
@@ -9,15 +11,19 @@
...
 cp "$WASP_BINARY_PATH" "$TMP_DIR/wasp-bin"
 
-CABAL_PROJECT_ROOT_PATH="$(cabal list-bin wasp-cli | sed s/\\/dist-newstyle.*//)"
-cp -R "$CABAL_PROJECT_ROOT_PATH/data" "$TMP_DIR/data"
+cabal sdist --list-only \
+  | grep "^./data" \
+  | while read -r data_file_path; do
+    mkdir -p "$TMP_DIR/$(dirname "$data_file_path")"
+    cp "$CABAL_PROJECT_ROOT_PATH/$data_file_path" "$TMP_DIR/$data_file_path"
+  done
```
sodic: Can we make it more obvious what this piece of code does with some nicely named functions and/or variables (or, if all else fails, some comments).

Btw, I think this would possibly be clearer and less fragile with `rsync`, not sure.

waspc/tools/make_binary_package.sh
cprecioso: Huh I thought by now this is common knowledge http://redsymbol.net/articles/unofficial-bash-strict-mode/
I'll add a comment

infomiho:
I'll let @sodic handle this one 👍

waspc/tools/make_binary_package.sh
```diff
@@ -9,15 +13,25 @@
...
 
 WASP_BINARY_PATH="$(cabal list-bin wasp-cli)"
 cp "$WASP_BINARY_PATH" "$TMP_DIR/wasp-bin"
 
-CABAL_PROJECT_ROOT_PATH="$(cabal list-bin wasp-cli | sed s/\\/dist-newstyle.*//)"
-cp -R "$CABAL_PROJECT_ROOT_PATH/data" "$TMP_DIR/data"
+CABAL_DATA_FILE_PATHS=$(
+  # First we list all the files declared in the .cabal file
+  # Then we only take the ones in the "data" directory
+  cabal sdist --list-only | grep "^./data/"
```
sodic: Not a big deal, but the dot is interpreted as "any character" not a literal dot. We should escape it.

waspc/tools/make_binary_package.sh
```diff
@@ -1,4 +1,8 @@
-#!/bin/sh -e
+#!/usr/bin/env bash
+
+# Bash "Strict Mode" http://redsymbol.net/articles/unofficial-bash-strict-mode/
```
sodic, line 3: Pretty cool, this is news to me :)

waspc/tools/make_binary_package.sh
```diff
@@ -9,15 +13,27 @@
...
+# the declared data files).
+# 2. Filter to get only the files in the "data" dir.
+# 3. Copy the files in that list from the project root to the tmp dir to be
+# packaged.
+#
+# We use rsync to recreate the directory structure properly, plus it has a nice
+# `--files-from=-` option to read from stdin the list of files to be copied.
+cabal sdist --list-only \
+  | grep "^\./data/" \
+  | rsync -av --files-from=- "$CABAL_PROJECT_ROOT_PATH" "$TMP_DIR"
```
sodic: The comment is nice, but all of it can be expressed through code if you want:
```bash
list_all_files_needed_in_binary() {
  cabal sdist --list-only
}

filter_data_files() {
  grep -E '^\./data/'
}

list_all_files_needed_in_binary \
  | filter_data_files \ 
  | rsync -a --files-from=- "$CABAL_PROJECT_ROOT_PATH" "$TMP_DIR"
```

The `filter_data_files` is arguably clearer with `grep`.

Anyway, choose whichever combination of comment/function/inline you like best and think is the clearest (it might be the one you already have).

waspc/tools/make_binary_package.sh
cprecioso: Did it with rsync, check it out

waspc/tools/make_binary_package.sh
cprecioso: I don't like the overhead of declaring functions for a single statement, so I'll err on the side of keeping it as-is

cprecioso:
Undid the rsync solution since it is not installed by default on Linux runners, so we keep it portable with mkdir + cp

PR #3491, opened 2025-12-11
Add `manifest.json` for libs

infomiho:
## Description

This is a libs cache busting experiment:
- When we ship libs to users, the libs can have the same version as the compiler (e.g. 0.19.2). Compiler version 0.19.2 will always use lib version 0.19.2. Users don't need cache busting since once the lib is shipped with the compiler, that's it. Once they update the compiler, the lib has the new version and `npm` will install the new version normally.
- When we (Wasp developers) work on Wasp, we want the cache busting to work on each change to the lib. We don't want to make a change in the lib and not see those changes because `npm` cached it because the version stayed e.g. 0.20.0. We can manually change the version in lib `pacakge.json` on each change, but that's annoying. One thing we can do is to generate a random version on each lib build while in dev so `npm` always uses the latest code.

The issue with random versions is that this means random tarball names and Wasp compiler won't know the name of the file in its `data` dir that it needs to copy to the user's dir.

To enable using random versions, I introduced a `manifest.json` file that contains the lib -> tarball name mapping:
```json
{
  "@wasp.sh/lib-auth": "wasp.sh-lib-auth-0.0.0-dev-638e73b3c7071af4.tgz"
}
```

Our compiler loads the `manifest.json` to know what is the tarball name for each lib. This gives the ability to have:
- a random tarball version while developing Wasp (`wasp.sh-lib-auth-0.0.0-dev-638e73b3c7071af4.tgz`)
- and a fixed tarball name with the Wasp compiler version (`wasp.sh-lib-auth-0.19.2.tgz`).

shapes at 25-12-11 22 40 52

I made it with Template Haskell so the `manifest.json` is loaded in compile time.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/tools/install_libs_to_data_dir.ps1
+++ b/waspc/tools/install_libs_to_data_dir.ps1
@@ -7,14 +7,6 @@ $dir = Split-Path -Parent $MyInvocation.MyCommand.Path
 $wascpDir = Resolve-Path "$dir/.."
 $dataLibsDir = Join-Path $wascpDir "data/Generator/libs"
 
-if ($env:USE_RANDOM_LIB_VERSION) {
-    $randomHex = [guid]::NewGuid().ToString("N").Substring(0, 16)
-    $libVersion = "0.0.0-dev-$randomHex"
-    Write-Host "USE_RANDOM_LIB_VERSION is set - using random version for cache busting: $libVersion"
-} else {
-    $libVersion = & "$wascpDir/run.bat" get-waspc-version | Select-Object -Last 1
-    Write-Host "Setting lib versions to: $libVersion"
-}
 
 # Clean up old libs.
 if (Test-Path $dataLibsDir) {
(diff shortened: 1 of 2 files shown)


waspc/src/Wasp/Generator/WaspLibs/Manifest/TH.hs
```diff
@@ -0,0 +1,28 @@
+{-# LANGUAGE TemplateHaskell #-}
```
infomiho, line 1: Template Haskell to get the JSON from the `manifest.json` as a Haskell `Map`.

waspc/src/Wasp/Generator/WaspLibs/WaspLib.hs
```diff
@@ -30,22 +33,19 @@ data WaspLib = WaspLib
 
 makeWaspLib :: String -> IO WaspLib
```
infomiho, line 34: This no longer needs to be `IO` and we can get rid of some of complexity that we required because this function needed to be `IO`. I didn't clean that up in this PR to keep the proposal clear.

waspc/tools/install_libs_to_data_dir.ps1
```diff
@@ -7,12 +7,24 @@ $dir = Split-Path -Parent $MyInvocation.MyCommand.Path
 $wascpDir = Resolve-Path "$dir/.."
 $dataLibsDir = Join-Path $wascpDir "data/Generator/libs"
 
+if ($env:USE_RANDOM_LIB_VERSION) {
```
infomiho: I imagined this as our workflow:
1. while developing you use `USE_RANDOM_LIB_VERSION=1` to get random versions and cache bustin
2. when the libs are built in the CI, they use the compiler version

waspc/tools/install_libs_to_data_dir.sh
```diff
@@ -7,10 +7,24 @@ script_dir=$(CDPATH="" cd -- "$(dirname -- "$0")" && pwd)
 waspc_dir=$script_dir/..
 data_libs_dir=$waspc_dir/data/Generator/libs
 
+# Get the waspc version to use for the libs.
```
infomiho: Given that I had to update both scripts (Unix and Windows) to have the same logic, I'd maybe refactor this in separate PR to a cross-platform JS script.

sodic (commented):
This is very cool, but I think too much for something that's not product code (code users get benefits from) but internal code (easier workflow for us). And even if it was a product feature, I'd probably vote for compromising on it to avoid the extra complexity.

I think we should either:
  1. Stick with the hashes.
  2. Be user-first, meaning we use wasp versions instead of hashes. Plus, either:
    1. Suck up the problems that creates for us (cache busting).
    2. Find an alternative/much simpler way to make our development cycle easier (e.g., by finding a way to bust the cache manually, automate the steps in a script, or document them in a README).
    
I vote for the simplest versoin of 2.2:
  - Always use Wasp versions.
  - Document how to bust caches during development.

If that ever becomes a problem, then we can look into automating it.

infomiho:
> Document how to bust caches during development.

It's quite annoying - you have to delete the `package-lock.json` each time you want to test changes. I suspect we'll quickly get tired of this.

Check what we'd have to do here: https://github.com/wasp-lang/wasp/pull/3501/changes#diff-0ee39dc1b0edc3e3323e2f964524eba4fb605a9068c818f070400f3e145901aaR37-R48

We'd keep modifying the `package-lock.json` if we nuke it or run a complex command to refresh the cache.

infomiho:
Closing this in favor of: https://github.com/wasp-lang/wasp/pull/3501 since I found an _okay_ way to bust the cache and added a new `./run libs:bust-cache` command.

PR #3497, opened 2025-12-15
Run tests with oldest supported Wasp version

sodic:
## Description

During #2482, @cprecioso noticed that our example apps don't work on older versions of Node we claim to support. This PR explores what to do about it. Discussion welcome.

The main question is: Which tests do we run on which versions of Node?
After we decide that, we can deal with how and where we specify this:
- I first choose a repostiroy variable for now
- @cprecioso then suggested using `.nvmrc` which was a much better idea (comment below)

### Final solution
We now run all tests and builds with Wasp's lowest supported vresion of Node. Unfortunatley, not all places could read the version from `.nvmrc`, so there is some duplication. More info on this in the comments.

For consistency, I used `waspc/.nvmrc` as a source-of-truth for the Node version wherever I could, not just where it matters to the user. 

If I couldn't use that, I've added a comment that explains why.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/.github/workflows/release-examples-deploy.yaml
+++ b/.github/workflows/release-examples-deploy.yaml
@@ -51,7 +51,7 @@ jobs:
     steps:
       - name: Install Wasp
         run: |
-          curl -sSfL https://get.wasp.sh/installer.sh | sh -s -- -v "${{  inputs.version }}"
+          npm i -g @wasp.sh/wasp-cli@"${{  inputs.version }}"
 
       - name: Install Fly CLI
         uses: superfly/flyctl-actions/setup-flyctl@v1
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@
 
 ## What is Wasp?
 
-Wasp (**W**eb **A**pplication **Sp**ecification) is a Rails-like framework for React, Node.js, and Prisma.  
+Wasp (**W**eb **A**pplication **Sp**ecification) is a Rails-like framework for React, Node.js, and Prisma.
 Build your app in a day and deploy it with a single CLI command!
 
 ### Why is Wasp awesome
@@ -112,7 +112,7 @@ For more information about Wasp, check [**docs**](https://wasp.sh/docs).
 Run to install Wasp on OSX/Linux/WSL(Win):
 
 ```sh
-curl -sSL https://get.wasp.sh/installer.sh | sh
+npm i -g @wasp.sh/wasp-cli@latest
 ```
 
 From there, follow the instructions to run your first app in less than a minute!
--- a/examples/ask-the-documents/README.md
+++ b/examples/ask-the-documents/README.md
@@ -19,7 +19,7 @@ This is an example Wasp app that supports:
 1. Make sure you have the latest version of Wasp installed.
 
    ```bash
-   curl -sSL https://get.wasp.sh/installer.sh | sh
+   npm i -g @wasp.sh/wasp-cli@latest
    ```
 
 2. Start the database with:
--- a/examples/tutorials/TodoApp/README.md
+++ b/examples/tutorials/TodoApp/README.md
@@ -11,7 +11,7 @@ After you've cloned this repository, you can run the app locally by following th
 Install Wasp
 
 ```bash
-curl -sSL https://get.wasp.sh/installer.sh | sh
+npm i -g @wasp.sh/wasp-cli@latest
 ```
 
 Position yourself in the project directory, then migrate the DB:
--- a/examples/websockets-realtime-voting/README.md
+++ b/examples/websockets-realtime-voting/README.md
@@ -23,7 +23,7 @@ git clone https://github.com/wasp-lang/wasp.git
 Make sure you've downloaded and installed Wasp
 
 ```bash
-curl -sSL https://get.wasp.sh/installer.sh | sh
+npm i -g @wasp.sh/wasp-cli@latest
 ```
 
 Then navigate to the project directory
--- a/mage/Dockerfile
+++ b/mage/Dockerfile
@@ -39,8 +39,7 @@ RUN mkdir -p /app/.wasp/build/server/node_modules
 # TODO: Use pm2?
 # TODO: Use non-root user (node).
 FROM base AS server-production
-RUN curl -sSL https://get.wasp.sh/installer.sh | sh -s -- -v 0.20.0
-ENV PATH "$PATH:/root/.local/bin"
+RUN npm i -g @wasp.sh/wasp-cli@0.20.0
 ENV NODE_ENV production
 WORKDIR /app
 # Copying the top level 'node_modules' because it contains the Prisma packages
--- a/mage/src/client/pages/ResultPage.jsx
+++ b/mage/src/client/pages/ResultPage.jsx
@@ -625,7 +625,7 @@ export default function RunTheAppModal({ disabled, onDownloadZip }) {
               :
             </h2>
             <pre className="mt-2 rounded-lg bg-slate-800 p-4 text-sm text-slate-200">
-              curl -sSL https://get.wasp.sh/installer.sh | sh
+              npm i -g @wasp.sh/wasp-cli@latest
             </pre>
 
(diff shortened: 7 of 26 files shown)


cprecioso:
- For ourselves, I'd make all the team use the lowest supported Node.js version in our development enviroments (Right now we say 22.12 in our documentation).
- I wouldn't like to have one more place where we have to remember to change the version numbers, _especially_ if it's not `grep`-able in our codebase. I see you removed a comment about `.nvmrc`, maybe we can go back to some way of declaratively saying which node.js version we use and have everywhere use that?

sodic:
> For ourselves, I'd make all the team use the lowest supported Node.js version in our development enviroments

I agree

> I wouldn't like to have one more place where we have to remember to change the version numbers, especially if it's not grep-able in our codebase... maybe we can go back to some way of declaratively saying which node.js version we use and have everywhere use that?

Fair enough, but where else would I put it to make it work with GH actions?

cprecioso:
> Fair enough, but where else would I put it to make it work with GH actions?

Something like this?

```yaml
steps:
  - uses: actions/checkout@v6

  - id: get-node-version
    run: echo "node-version=$(cat .nvmrc)" >> $GITHUB_OUTPUT

  - uses: actions/setup-node@v6
    with:
      node-version: ${{ steps.get-node-version.outputs.node-version }}
```

cprecioso:
In fact, reading from the docs, this is allowed

```yaml
steps:
  - uses: actions/checkout@v6

  - uses: actions/setup-node@v6
    with:
      node-version-file: .nvmrc
```

.github/workflows/ci-waspc-build.yaml
```diff
@@ -3,17 +3,14 @@ name: "Build - Wasp CLI binaries"
...
   workflow_call:
     inputs:
       node-version:
-        description: "Node.js version to use"
-        default: "22"
+        # NOTE: This should be equal to Wasp's lowest supported Node version.
+        # Don't change this version without updating it in all required
+        # places (full list in /waspc/.nvmrc).
+        description: "Node.js version to use for building"
+        default: "22.12.0"
```
sodic, line 13: The only way to avoid duplication here is doing something too crazy IMO:
- Having a separate setup job that calls `setup-node`
- Relying on `setup-node` to have an output called `node-version` and ignoring the fact that it already installs something and does some setup
- Reading that output in the next step and using conditionals when calling `setup-node` again.

If you can think of something better, I'm up for it.

.github/workflows/ci-waspc-test.yaml
```diff
@@ -39,10 +39,13 @@ jobs:
...
         # To reduce the number of jobs, we only test the Node.js versions on
-        # Ubuntu 20.04.
+        # Ubuntu 22.04.
         include:
           - os: ubuntu-22.04
-            node-version: 22
+            # NOTE: This should be equal to Wasp's lowest supported Node
+            # version. Don't change this version without updating it in all
+            # required places (full list in /waspc/.nvmrc).
+            node-version: "22.12.0"
```
sodic, line 48: Duplication again: matrix is a compile time thing, so it's impossible to get the version from `.nvmrc`. We'd have to do weird stuff again.

The version doesn't change that often so I opted for repetition.

.github/workflows/ci-waspc-build.yaml
cprecioso: > calling setup-node again

I have done that in the past (to deal with corepack) 😂 but I agree that it's unnecessary. no prob

.github/workflows/ci-waspc-test.yaml
cprecioso: ok

.github/workflows/release-create.yaml
```diff
@@ -51,9 +51,15 @@ jobs:
...
-        uses: actions/setup-node@v6
+      - uses: actions/setup-node@v6
+        with:
+          node-version-file: waspc/.nvmrc
+
+        # Tokenless npm publishing requires npm v11.5.1+ so we install it
+        # explicitly.
+        # Read more: https://docs.npmjs.com/trusted-publishers/
+      - name: Install latest npm
+        run: npm i -g npm@^11.5.1
```
cprecioso: Thanks for the assist bro 🙏💪

.github/workflows/release-create.yaml
```diff
@@ -51,9 +51,15 @@ jobs:
...
-      - # We don't set a specific node version here because we want to use the
-        # latest, to get token-less npm publishing.
-        uses: actions/setup-node@v6
+      - uses: actions/setup-node@v6
+        with:
+          node-version-file: waspc/.nvmrc
+
+        # Tokenless npm publishing requires npm v11.5.1+ so we install it
+        # explicitly.
+        # Read more: https://docs.npmjs.com/trusted-publishers/
```
cprecioso: Extreme nitpick, I don't think this comment is indented properly, I think right now it's part of the previous list item.

.github/workflows/ci-npm-package-publish.yaml
```diff
@@ -10,6 +10,14 @@ jobs:
...
     steps:
       - uses: actions/setup-node@v6
+        with:
+          node-version-file: waspc/.nvmrc
+
+        # Tokenless npm publishing requires npm v11.5.1+ so we install it
+        # explicitly.
+        # Read more: https://docs.npmjs.com/trusted-publishers/
+      - name: Install latest npm
+        run: npm i -g npm@^11.5.1
```
cprecioso: We don't do npm tokenless publishing here, this is not needed. We're publishing with `pkg.pr.new` and that does a whole different thing with its own authorization.

.github/workflows/ci-waspc-test.yaml
cprecioso: Optional: how about doing it such that `node-version` matrix prop is optional, and if empty we read from .nvmrc? RAW

.github/workflows/ci-npm-package-build.yaml
cprecioso: Maybe we can keep our own nvmrc version but do:

```suggestion
- name: Run packager script
run: node --experimental-type-stripping ./scripts/make-npm-packages/src/index.ts --input-dir ./input/ --output-dir ./output/
```

This is the flag that was switched on by default in v22.18 and v24 to support TS.

cprecioso:
Mmmm one more thing I realized, what about moving `waspc/.nvmrc` to just `.nvmrc`? Since it seems like it's the runtime for everything in our repo, not only waspc.

.github/workflows/ci-npm-package-publish.yaml
sodic: Ok, makes sense. What should I write, or can we then use the version from `.nvmrc`?

sodic:
> Mmmm one more thing I realized, what about moving `waspc/.nvmrc` to just `.nvmrc`? Since it seems like it's the runtime for everything in our repo, not only waspc.

I thought about it, but then we wouldn't get autodetection in `waspc`. But that's not a big deal, I'll do it.

.github/workflows/ci-npm-package-publish.yaml
sodic: I concluded I should remove it and did so.

.github/workflows/ci-waspc-test.yaml
sodic: I thought about it but decided I prefer the epxlicitness of the matrix. This was too much IMO.

.github/workflows/ci-npm-package-build.yaml
cprecioso: The suggestion seems to be malformed in GitHub UI. the change I'm suggesting is to use the .nvmrc version in the `setup-node` action, and just add the `--experimental-type-stripping` flag to this command.

cprecioso:
> we wouldn't get autodetection in waspc

what autodetection are you referring to?

PR #3534, opened 2025-12-22
[Wasp News] Core types and display module

sodic:


waspc/tests/UtilTest.hs
```diff
@@ -141,6 +143,38 @@ spec_insertAt = do
...
+    prop "does not produce lines longer than max length unless unavoidable" $
+      \maxLen str ->
+        let lns = lines (wrapString maxLen str)
+         in all (\l -> length l <= maxLen || length (words l) == 1) lns
+
+    prop "does not produce empty lines" $
+      \maxLen str -> null str || notElem "" (lines (wrapString maxLen str))
+
+    prop "treats negative numbers as zero" $
+      \(Negative n) str -> wrapString n str == wrapString 0 str
```
sodic: This behavior is debatable but I thought it made sense considering how we want to treat news that are too long.

waspc/src/Wasp/Util.hs
```diff
@@ -187,6 +188,23 @@ insertAt theInsert idx host =
...
+    groupWordsIntoLines :: [String] -> [[String]]
+    groupWordsIntoLines [] = []
+    groupWordsIntoLines ws =
+      let (line, rest) = takeLine maxLength ws
+       in line : groupWordsIntoLines rest
+
+    takeLine :: Int -> [String] -> ([String], [String])
+    takeLine _ [] = ([], [])
+    takeLine remainingLength (w : ws)
+      | length w > remainingLength && remainingLength /= maxLength = ([], w : ws)
```
sodic, line 203: The second part of the condition might be a little unclear. It means "Don't include a word if it's longer than the remaining length, except when it's too long and exceeds maxLength."

I couldn't find a nice way to extract it (i.e., without nested where or something) so I left it in. Let me know if it's too confusing.

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
```diff
@@ -0,0 +1,33 @@
+{-# LANGUAGE OverloadedRecordDot #-}
+
+module Wasp.Cli.Command.News.Display (displayNewsEntry) where
+
+import Data.Time (defaultTimeLocale, formatTime)
+import Wasp.Cli.Command.News.Common (NewsEntry (..), NewsLevel (..))
+import Wasp.Util (indent, wrapString)
+import qualified Wasp.Util.Terminal as Term
+
+displayNewsEntry :: NewsEntry -> IO ()
```
sodic: To see how the output looks like:
```sh
cabal repl wasp-cli --build-depends=aeson
ghci> import Wasp.Cli.Command.News.Display; import Data.Aeson; import Data.Maybe
ghci> displayNewsEntry (fromJust (decode "{\"level\":\"high\",\"id\":\"cve-2025-8472-auth-bypass\",\"title\":\"Authentication Bypass in JWT Verification\",\"body\":\"CVE-2025-8472: Improperly validated JWT tokens in authentication middleware allow attackers to forge tokens and bypass authentication. Affects all apps using usernameAndPassword, Google, or GitHub OAuth. Upgrade to 0.16.8 immediately.\",\"publishedAt\":\"2025-10-22T14:30:00.000Z\",\"waspVersionsAffected\":\">=0.13.0 <0.16.8\"}"))
```
You can play around with input lengths and see what happens.

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
```diff
@@ -0,0 +1,33 @@
...
+displayNewsEntry entry = do
+  putStrLn ""
+  putStrLn $
+    Term.applyStyles [Term.Bold] entry.title
```
sodic: We don't cut off the title if it's too long. Should we?

waspc/cli/src/Wasp/Cli/Command/News/Core.hs
```diff
@@ -0,0 +1,33 @@
+{-# LANGUAGE DeriveGeneric #-}
```
sodic, line 1: Needed to derive `FromJSON`.

waspc/src/Wasp/Util.hs
```diff
@@ -187,6 +188,23 @@ insertAt theInsert idx host =
 trim :: String -> String
 trim = reverse . dropWhile isSpace . reverse . dropWhile isSpace
 
+wrapString :: Int -> String -> String
```
sodic, line 191: I didn't use `Natural` because it would complicate both calling and implementation (e.g., `QuickCheck` doesn't have an `arbitrary` instance for `Natural`, the callers and the implementation would have to call `fromIntegral`...).

waspc/tests/UtilTest.hs
```diff
@@ -141,6 +143,35 @@ spec_insertAt = do
       insertAt [0] 3 [1, 2, 3] `shouldBe` ([1, 2, 3, 0] :: [Int])
       insertAt [0] 4 [1, 2, 3] `shouldBe` ([1, 2, 3, 0] :: [Int])
 
+spec_wrapString :: Spec
+spec_wrapString = do
+  describe "wrapString" $ do
+    it "returns an empty string for empty input" $ do
```
sodic, line 147: All the `do`s are redundant, but so are all other `do`s in the file. I decided it's best to leave it as is :)

waspc/cli/src/Wasp/Cli/Command/News/Core.hs
Martinsos, line 1: Tiny idea: would in this case name `Core.hs` work better? If not, never mind.

Martinsos, line 1: General thought: does it make more sense to keep all the News logic here in CLI, or would it make more sense to keep the core of it in waspc lib itself, and CLI only uses it + adds display?

But maybe better not to complicate and just have it all here, yeah ok. I will leave the comment as possibly interesting though, but I wouldn't move it, I would leave it here.

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
```diff
@@ -0,0 +1,33 @@
...
+    dateText = formatTime defaultTimeLocale "%Y-%m-%d" newsEntry.publishedAt
+    dotCount = max minDotsCount (maxColumns - length newsEntry.title - length dateText - 2)
+    maxColumns = 80
+    minDotsCount = 5
+
+showLevelInColor :: NewsLevel -> String
+showLevelInColor newsLevel = case newsLevel of
+  High -> Term.applyStyles [Term.Red] "High"
+  Moderate -> Term.applyStyles [Term.Yellow] "Moderate"
+  Low -> Term.applyStyles [Term.Blue] "Low"
```
Martinsos: This is already nice, but it would be great if it was easier to read. If I could take one look and have an idea of what I can expect as an output. Here are some changes I would consider:
1. Turn `displayNewsEntry` into `showNewsEntry`. Because why not have it pure. You could even write a test then. And then `showLevelInColor` to just `showNewsLevel`.
2. Extract more from `showNewsEntry` (I will call it like that from now on) to `where`. Then, what you can achieve at the top level, is something easy to understand, like
```hs
(title <> " " <> separatorDots <> " " <> publishedAtDate)
  <> "\n" <> level
  <> "\n" <> body
```
You cuold make it even nicer by using `trimming`, like this, we use it in some parts of Wasp:
```hs
  [trimming|
    ${title} ${separatorDots} ${publishedAtDate}
    ${level}
    ${body}
  |]
  where
    title = T.pack $ ... -- NOTE: trimming takes Text, not String bits.
```

And you would more stuff in `where` obviously, and also maybe use nested `where` to make it clear what is used for what, or otherwise extract some stuff as separate functions potentially. But you know, let's make it clean code.

Yeah we should do this, it is easy to do, and it greatly increases the readibility of this piece of code. You could then also potentially write a test for this output since it is pure, having a test sounds nice, we can feel confident it works then. Even if just a single test case.

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
```diff
@@ -0,0 +1,33 @@
...
+      <> Term.applyStyles [Term.Bold] (replicate dotCount '.')
+      <> " "
+      <> Term.applyStyles [Term.Yellow, Term.Bold] dateText
+  putStrLn $
+    showLevelInColor newsEntry.level
+      <> "\n"
+      <> Term.applyStyles [Term.Grey] (indent 2 $ wrapString (maxColumns - 2) newsEntry.body)
+  where
+    dateText = formatTime defaultTimeLocale "%Y-%m-%d" newsEntry.publishedAt
+    dotCount = max minDotsCount (maxColumns - length newsEntry.title - length dateText - 2)
```
Martinsos: This looks likely to become wrong hm. If we change this first line where it is used, in any way, even add a space. This `2` here, that is for 2 spaces, right?

I would say, either let's add comment at least warning about this, or let's fix it for real, for example you could have a function called `separateWith :: String -> String -> Char -> Int -> String; separateWith start end sep width = ...` and then the function will take care of it and we will use it to connect these two.

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
```diff
@@ -0,0 +1,33 @@
...
+  putStrLn $
+    Term.applyStyles [Term.Bold] newsEntry.title
+      <> " "
+      <> Term.applyStyles [Term.Bold] (replicate dotCount '.')
+      <> " "
+      <> Term.applyStyles [Term.Yellow, Term.Bold] dateText
+  putStrLn $
+    showLevelInColor newsEntry.level
+      <> "\n"
+      <> Term.applyStyles [Term.Grey] (indent 2 $ wrapString (maxColumns - 2) newsEntry.body)
```
Martinsos: Would be nice if it was clear this `2` here is the one and same number, in the sense that if first one changes, the second one changes.
If you make this a separate binding, as I suggest below, it will be easy to put this `2` in where, bound to `numIndentSpaces` or `indentSize` or wahtever you want to call it.

waspc/src/Wasp/Util.hs
Martinsos: Yeah, I was there couple of times myself, sounds great but complicates things too much. Makes sense.

waspc/cli/src/Wasp/Cli/Command/News/Core.hs
```diff
@@ -0,0 +1,33 @@
...
+import Data.Char (toLower)
+import Data.Time (UTCTime)
+import GHC.Generics
+
+data NewsEntry = NewsEntry
+  { id :: !String,
+    title :: !String,
+    body :: !String,
+    level :: !NewsLevel,
+    publishedAt :: !UTCTime
```
cprecioso, line 20: question: what's the rule of thumb for setting properties as strict or lazy?

waspc/cli/src/Wasp/Cli/Command/News/Common.hs
```diff
@@ -0,0 +1,33 @@
...
+  ( NewsEntry (..),
+    NewsLevel (..),
+  )
+where
+
+import Data.Aeson (FromJSON (..), genericParseJSON)
+import qualified Data.Aeson as Aeson
+import Data.Char (toLower)
+import Data.Time (UTCTime)
+import GHC.Generics
```
cprecioso: nitpick
```suggestion
import GHC.Generics (Generic)
```

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
cprecioso: Nice, here I leave the output so it's easier to see :)

```

Authentication Bypass in JWT Verification ........................... 2025-10-22
High
  CVE-2025-8472: Improperly validated JWT tokens in authentication middleware
  allow attackers to forge tokens and bypass authentication. Affects all apps
  using usernameAndPassword, Google, or GitHub OAuth. Upgrade to 0.16.8
  immediately.
```

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
```diff
@@ -0,0 +1,33 @@
...
+module Wasp.Cli.Command.News.Display (displayNewsEntry) where
+
+import Data.Time (defaultTimeLocale, formatTime)
+import Wasp.Cli.Command.News.Common (NewsEntry (..), NewsLevel (..))
+import Wasp.Util (indent, wrapString)
+import qualified Wasp.Util.Terminal as Term
+
+displayNewsEntry :: NewsEntry -> IO ()
+displayNewsEntry newsEntry = do
+  putStrLn ""
```
cprecioso: nitpick: i think this empty line is probably the responsibility of the caller, not of this fn

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
```diff
@@ -0,0 +1,33 @@
...
+      <> " "
+      <> Term.applyStyles [Term.Yellow, Term.Bold] dateText
+  putStrLn $
+    showLevelInColor newsEntry.level
+      <> "\n"
+      <> Term.applyStyles [Term.Grey] (indent 2 $ wrapString (maxColumns - 2) newsEntry.body)
+  where
+    dateText = formatTime defaultTimeLocale "%Y-%m-%d" newsEntry.publishedAt
+    dotCount = max minDotsCount (maxColumns - length newsEntry.title - length dateText - 2)
+    maxColumns = 80
```
cprecioso: Let's note in the server part somewhere what the max length for the title is and link to this line

waspc/src/Wasp/Util.hs
```diff
@@ -187,6 +188,23 @@ insertAt theInsert idx host =
...
+wrapString maxLength = intercalate "\n" . map unwords . groupWordsIntoLines . words
```
cprecioso, line 192: Good job! I can't help but wonder if it'd be more useful for us to offload this code to a library. e.g. `prettyprinter` ([1](https://hackage.haskell.org/package/prettyprinter), [2](https://hackage.haskell.org/package/prettyprinter-ansi-terminal)) seems to be what `optparse-applicative` uses, and a quick read of the docs look like it might offer lots of things we need (wrapping text, filling lines with repeated characters, etc).

waspc/src/Wasp/Util.hs
cprecioso: Maybe you could store the different parts of the conditional in variables and name them. I think you gave a good overview in this comment, so I'd try to make the code say that.

At any rate, I do expect line wrapping algorithms to become messy, so I am not too bothered by the current state.

waspc/src/Wasp/Util.hs
cprecioso: (I understand we're on a schedule so I don't expect this suggestion to be done now)

waspc/cli/src/Wasp/Cli/Command/News/Core.hs
sodic: Hm, good question. My rule is:
  - I use strict for all new code except when I need lazy (which is never).
  - I don't change old code for this.

I don't think it matters much though.

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
sodic: Here's a picture too:

image

waspc/cli/src/Wasp/Cli/Command/News/Core.hs
```diff
@@ -0,0 +1,33 @@
...
+instance FromJSON NewsEntry
+
+data NewsLevel
+  = Low
+  | Moderate
+  | High
+  deriving (Eq, Ord, Show, Generic)
+
+instance FromJSON NewsLevel where
+  parseJSON = genericParseJSON $ Aeson.defaultOptions {Aeson.constructorTagModifier = map toLower}
```
infomiho, line 33: Does `constructorTagModifier` make `high` `moderate` `low` values work?

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
```diff
@@ -0,0 +1,31 @@
...
+import Wasp.Util (indent, wrapString)
+import qualified Wasp.Util.Terminal as Term
+
+displayNewsEntry :: NewsEntry -> String
+displayNewsEntry newsEntry =
+  Term.applyStyles [Term.Bold] newsEntry.title
+    <> " "
+    <> Term.applyStyles [Term.Bold] (replicate dotCount '.')
+    <> " "
+    <> Term.applyStyles [Term.Yellow, Term.Bold] dateText
```
infomiho: I got confused for a bit because I thought we use a newline between each line, but not really since for the first three items we are actually using a space. The first three items are all just one line.

I'd maybe put this into a new variable named `header` so you can simply use `unlines [header, level, body]`.

RAW

waspc/src/Wasp/Util.hs
Martinsos: Yup this one is a bit confusing!

One thing I find confusing is the name, `takeLine`. It makes sense in some way, but `takeWords` would be a better name I think. `words` in the name indicates we are treating them as separate words to be concatenated with spaces, so that is also ok.
I am ok with you sticking to `takeLine` if you liek that better though.

As for the arg name -> yeah, `remainingLength` is confusing. But why not just name it `maxLineLength`? From the perspective of the `takeLine` function, it doesn't have to know anything about the concept of "remaining" -> each time it is called, it is building a line of certain length, that is it.

Auff wait, ok I see now what is happening, I started reading the implementation. Yeah I didn't expect this -> you are chekcing inside if remainingLength /= maxLength, because you are handling special case where you are just starting and you have to take at least one word, even if it is longer than the remaining length. So special case where the first word to make a line is longer than the maxLength. But I don't think you should do it like this, this is quite unexpected / confusing.

Btw what is with my implementation that I did during the hackaton, I thikn I had a nice solution going there?

But ok let's focus on this.

Special case is when you are building a line and the very first word is longer than the maxLineLength.

You can't know when you are at the start, from inside the `takeLine`, because it doesn't have that info. You went around that by using `maxLength`, but that is in a way "global" variable and this solution makes the whole thing ugly and unexpected.

One solution is to make `takeLine` simpler, it would not handle this special case, instead it would return no words in such case, and outside you would handle the special case of "first words", be it directly in `groupWordsIntoLines`, or in a wrapper function called `takeLine1` (or some better name).

Example of doing this directly in `groupWordsIntoLines`:
```hs
  groupWordsIntoLines (firstWord:wordsRest) =
    let (lineRest, leftoverWords) = takeLine (maxLength - length firstWord - 1) wordsRest
        line = firstWord : lineRest
     in line : groupWordsIntoLines leftoverWords
```

waspc/tests/UtilTest.hs
Martinsos: Yeah it's ok, they make it easier to add more code and make it look more uniform.

waspc/tests/UtilTest.hs
```diff
@@ -141,6 +143,35 @@ spec_insertAt = do
...
+      wrapString 20 "hello world" `shouldBe` "hello world"
+
+    it "wraps a string that exceeds max length" $ do
+      wrapString 10 "hello world" `shouldBe` "hello\nworld"
+
+    it "includes a long word at the start of the line even if it exceeds max length" $ do
+      wrapString 5 "extraordinary day is extraordinary" `shouldBe` "extraordinary\nday\nis\nextraordinary"
+
+    prop "preserves all words in order" $
+      \maxLen str -> words str == words (wrapString maxLen str)
```
Martinsos: How likely is string generated by quickcheck to have something that looks like proper input? I am assuming it will have random characters, and while I guess it will throw in space sometimes, I somehow doubt that most of this input will actually be representative or interesting. Not only should it throw in at least a couple of spaces, but it should also choose such maxLen that the lengths of words are in some way interesting.
I tihnk what you end up here with, let's say it produces 1000 inputs, is mostly inputs that are nothing alike real life data, therefore meaning that you are not really testing anything useful. Maybe I am wrong, you can test it by printing the inputs.

If you wanted to make it represntative, I think you would want to configure how it produces inputs. It should draw maxLength from 1 to 200 with some reasonable distribution (normal?), and also you would have it generate a list of strings, not one big string, and you would also make sure its length is from some reasonable distribution, and lengths of each word in it. Then you concatenate those and you get something proper I think. You should also throw in soem numbers, special chars, maybe extra spaces, ... .

Maybe that is too much work, and in that case I would even drop these tests, if it is ndeed true they are not testing anything useful.

waspc/tests/UtilTest.hs
```diff
@@ -141,6 +143,35 @@ spec_insertAt = do
...
+spec_wrapString :: Spec
+spec_wrapString = do
+  describe "wrapString" $ do
+    it "returns an empty string for empty input" $ do
+      wrapString 10 "" `shouldBe` ""
+
+    it "returns a single line when text fits" $ do
+      wrapString 20 "hello world" `shouldBe` "hello world"
+
+    it "wraps a string that exceeds max length" $ do
```
Martinsos, line 153: Some cases I would also consider adding:
1. Multiple spaces between the words: "Oh yeah  this is   great".
2. Separators, special chars, numbers: "It's a great day, number 9 said to 6: and indeed it was.".
3. Newlines: "This is my list\n - buy eggs\n - wash hands\n - pet a dog.".

waspc/tests/UtilTest.hs
Martinsos: Hm I am not sure what is that behaviour, what happens when length is 0? I don't see a test case for that one, I thikn you should add it.

Also I am not sure what do you mean by `it made sense ... news that are too long` -> how is that connected to this?

waspc/cli/src/Wasp/Cli/Command/News/Core.hs
```diff
@@ -0,0 +1,33 @@
...
+  }
+  deriving (Generic, Show, Eq)
+
+instance FromJSON NewsEntry
+
+data NewsLevel
+  = Low
+  | Moderate
+  | High
+  deriving (Eq, Ord, Show, Generic)
```
Martinsos, line 30: Maybe I completely forgot, but I thought we are doing "Critical", "Important" and "Info", how did we end up with High, Moderate and Low?

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
sodic: Hm, I thought I made this pure before you took a look at it :)

Anyway:
- It's pure already. I still didn't write a test because the colors were annoying. I could if you think it's important.
- I'll do the trimming changes
- `displayNewsEntry` -> `showNewsEntry` - I thought about this but skipped it on purpose because I thought `show` has some implied semantics this doesn't have. I know the `Show` typeclass does, but is there a convention for functions named `showX` too? If not, I'll renam it, `showNewsEntry` is a nicer name.

waspc/cli/src/Wasp/Cli/Command/News/Core.hs
sodic: Yes!

waspc/cli/src/Wasp/Cli/Command/News/Core.hs
sodic: Yes, I agree

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
sodic: I didn't use trimming in the end because I found the text conversions annoying, and I think this does the job.

Martinsos: From what I know, `show` usually means it is converted to String, that is it. `display` indicates it is being printed (IO). I can't remember though if I made that up or saw somewhere.

Ah ok, yeah trimming needs all to be Text, but result is nice. All right.

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
Martinsos: I don't konw, I woudl say it is fine.

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
```diff
@@ -0,0 +1,37 @@
...
+module Wasp.Cli.Command.News.Display (showNewsEntry) where
+
+import Data.List (intercalate)
+import Data.Time (defaultTimeLocale, formatTime)
+import Wasp.Cli.Command.News.Core (NewsEntry (..), NewsLevel (..))
+import Wasp.Util (indent, wrapString)
+import qualified Wasp.Util.Terminal as Term
+
+showNewsEntry :: NewsEntry -> String
+showNewsEntry newsEntry =
```
Martinsos, line 12: I still like the better what I suggested: you could have had it obvious that news is title and separator dots and then date, and then next line is level,a dn then body.
THis is better for sure, we have headerLine, but still, styling is mixed with the structure, so I have to parse it and read stuff below.

While
```hs
[trimming|
  {title} {dotsSeparator} {publishedAt}
  {level}
  {body}
|]
```
is just what it needs to be -> at a glance, I know what the news are going to look like. Awesome.

And even if not using trimming, although I don't really see why doing `T.pack` is a big deal, you cuold accomplish this with just `<>`, a bit less nice but still.

I guess the main thing I am advocating is that you "exapnd" the header title, instead of having it below and its struture also mixed with styling, which makes it even harder to read.

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
```diff
@@ -0,0 +1,37 @@
...
+        <> " "
+        <> Term.applyStyles [Term.Bold] (replicate dotCount '.')
+        <> " "
+        <> Term.applyStyles [Term.Yellow, Term.Bold] date
+
+    levelLine = showNewsLevel newsEntry.level
+
+    bodyBlock = Term.applyStyles [Term.Grey] $ indent indentSize $ wrapString (maxColumns - indentSize) newsEntry.body
+
+    dotCount = max minDotsCount (maxColumns - length newsEntry.title - length date - 2)
```
Martinsos, line 26: Why minus 2? As I said above in the comment, you should either explain this is caused by the two spaces around the dot separator, which is fragile but at least there is a comment, or better, you should write a function that adds repeated separator between the two string, and repeats it till it fills the length.
Check the earlier comment, I think you maybe missed it because you never commented on it.

waspc/tests/UtilTest.hs
```diff
@@ -141,6 +141,47 @@ spec_insertAt = do
...
+    it "never breaks long words even if they exceed max length" $ do
+      wrapString 5 "extraordinary day is extraordinary" `shouldBe` "extraordinary\nday\nis\nextraordinary"
+
+    it "wraps text with many words into multiple lines" $ do
+      wrapString 20 "one two three four five six seven eight nine ten"
+        `shouldBe` "one two three four\nfive six seven eight\nnine ten"
+
+    it "treats newlines as spaces" $ do
+      wrapString 15 "hello\nworld this\nis a test"
+        `shouldBe` "hello world\nthis is a test"
```
Martinsos, line 165: That mabye sucks a bit, so we can't use a newline in the news? I think we will want to, to have nicer formatting. This means we can have no newlines, no paragraphs, ... . Yeah, I think we should support newlines.

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
sodic: Not addressed in https://github.com/wasp-lang/wasp/pull/3554. Added to the list for fixing later.

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
sodic: Not addressed in https://github.com/wasp-lang/wasp/pull/3554. Added to the list for fixing later.

waspc/tests/UtilTest.hs
sodic: Not addressed in https://github.com/wasp-lang/wasp/pull/3554. I'll create an issue for this.

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
sodic: Not addressed in https://github.com/wasp-lang/wasp/pull/3554. Added to the list for fixing later.

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
sodic: Not addressed in https://github.com/wasp-lang/wasp/pull/3554. Added to the list for fixing later.

waspc/src/Wasp/Util.hs
sodic: Not addressed in https://github.com/wasp-lang/wasp/pull/3554. Added to the list for fixing later.

PR #3535, opened 2025-12-22
[Wasp News] Fetching module

sodic:


waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
```diff
@@ -0,0 +1,36 @@
...
+fetchNews :: IO (Either String [NewsEntry])
+fetchNews = do
+  waspNewsUrl <- fromMaybe "https://news.wasp.sh" <$> lookupEnv "WASP_NEWS_SERVER_URL"
+
+  requestResult <- try $ httpJSONThatThrowsIfNot2xx =<< parseRequest waspNewsUrl
+
+  return $ case requestResult of
+    -- TODO: Can I get the status code easily
+    Left (_ :: HttpException) -> Left "Failed to fetch news from server."
+    Right (Left jsonError) -> Left $ "Failed to decode news JSON: " ++ jsonError
```
sodic, line 31: Do you think this is enough error info? The response status is not easily obtainable and it might not be that useful.

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
sodic, line 1: Do you think this module should be more general and extracted? It seems like it's applicable to more than news fetching.

If I did that, I'd have to make the error messages more general and create a separete news fetching module that will take care of news-specific stuff (e.g., reading the news URL env var ).

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
```diff
@@ -0,0 +1,32 @@
...
+import Data.Functor ((<&>))
+import Data.Maybe (fromMaybe)
+import Network.HTTP.Simple (HttpException, parseRequest)
+import System.Environment (lookupEnv)
+import Wasp.Cli.Command.News.Common (NewsEntry)
+import Wasp.Util.Network.HTTP (httpJSONThatThrowsIfNot2xx)
+
+fetchNewsWithTimeout :: Int -> IO (Either String [NewsEntry])
+fetchNewsWithTimeout timeoutSeconds =
+  timeout timeoutMicroseconds fetchNews <&> fromMaybe (Left "News fetching timed out")
```
sodic, line 18: We could technically handle the timeout by adding the option to `httpLBS` inside `httpJSONThatThrowsIfNot2xx` but it would make the function more complex and probably isn't worth it.

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
```diff
@@ -0,0 +1,32 @@
...
+import System.Timeout (timeout)
+import Control.Exception (try)
+import Data.Functor ((<&>))
+import Data.Maybe (fromMaybe)
+import Network.HTTP.Simple (HttpException, parseRequest)
+import System.Environment (lookupEnv)
+import Wasp.Cli.Command.News.Common (NewsEntry)
+import Wasp.Util.Network.HTTP (httpJSONThatThrowsIfNot2xx)
+
+fetchNewsWithTimeout :: Int -> IO (Either String [NewsEntry])
```
sodic, line 16: Should we use `Double` instead of `Int`?

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
```diff
@@ -0,0 +1,32 @@
+module Wasp.Cli.Command.News.Fetching
+  ( fetchNewsWithTimeout,
+    fetchNews,
```
sodic, line 3: Here's how you can test the module:
```
cabal repl wasp-cli
ghci> import Wasp.Cli.Command.News.Fetching
ghci> fetchNews
ghci> fetchNewsWithTimeout 1
```
You'll likely have to throttle your connection to get the timeout to trigger, but you can trust me it works :)

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
```diff
@@ -0,0 +1,32 @@
...
+
+fetchNewsWithTimeout :: Int -> IO (Either String [NewsEntry])
+fetchNewsWithTimeout timeoutSeconds =
+  timeout timeoutMicroseconds fetchNews <&> fromMaybe (Left "News fetching timed out")
+  where
+    timeoutMicroseconds = timeoutSeconds * 1000000
+
+fetchNews :: IO (Either String [NewsEntry])
+fetchNews = do
+  waspNewsUrl <- fromMaybe "https://news.wasp.sh" <$> lookupEnv "WASP_NEWS_SERVER_URL"
```
sodic, line 24: The env var is useful for local development and automated testing.

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
cprecioso: Nah, it's fine if there isn't an evident need right now. It's easy to change.

You could always do

```suggestion
fetchNewsWithTimeout :: Num a => a -> IO (Either String [NewsEntry])
```

but i think that'd be overkill

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
cprecioso: I think it's fine. You can print the URL of the server so the debugging-inclined ones can `curl` it.
I'd remove the casting if you're not doing anything with it though.

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
```diff
@@ -0,0 +1,32 @@
...
+
+  requestResult <- try $ httpJSONThatThrowsIfNot2xx =<< parseRequest waspNewsUrl
+
+  return $ case requestResult of
+    -- TODO: Can I get the status code easily
+    Left (_ :: HttpException) -> Left "Failed to fetch news from server."
+    Right (Left jsonError) -> Left $ "Failed to decode news JSON: " ++ jsonError
+    Right (Right news) -> Right news
```
cprecioso, line 32: I dislike this double Either checking. Maybe a `join` would make this code nicer? RAW

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
```diff
@@ -0,0 +1,32 @@
...
+    timeoutMicroseconds = timeoutSeconds * 1000000
+
+fetchNews :: IO (Either String [NewsEntry])
+fetchNews = do
+  waspNewsUrl <- fromMaybe "https://news.wasp.sh" <$> lookupEnv "WASP_NEWS_SERVER_URL"
+
+  requestResult <- try $ httpJSONThatThrowsIfNot2xx =<< parseRequest waspNewsUrl
+
+  return $ case requestResult of
+    -- TODO: Can I get the status code easily
```
infomiho, line 29: There is the `getHttpExceptionStatusCode` from the `Wasp.Util.Network.HTTP` module.

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
infomiho: > Do you think this module should be more general and extracted? 

I'd create an issue to maybe move some of the timeout logic to ` Wasp.Util.Network.HTTP`. 

Other than that, I wouldn't extract anything else.

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
infomiho: I think you can set the timeout on the result of `parseRequest` and not need to touch the `httpJSONThatThrowsIfNot2xx` by using: https://hackage.haskell.org/package/http-conduit-2.3.9/docs/Network-HTTP-Simple.html#v:setRequestResponseTimeout

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
Martinsos: it's fine to use Int, I don't see why you would want to use `Double`?

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
Martinsos: You can see our code doing this in Wasp.AI.OpenAI.ChatGPT.hs -> queryChatGPT function, look for `setRequestResponseTimeout`. I would maybe do both, both this and the `timeout` above, but if using just one, would stick to just `timeout`. You can check this module in general for some ways how to do these types of fetchings, I know I spent some time going in circles there till I figured out how to correctly use these http functions.

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
Martinsos: Yeah wanted to comment the same thing. I am not sure what is the solution, and this is explicit which is good, but it is a bit weird. But maybe better not to waste time on it.

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
Martinsos: What @infomiho said, I wouldn't waste time on over-engineering this.
Maybe you don't even have to bother with extracting it to utils, you could, but it is quite simple really.

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
sodic: The type is there to help type inference, can't remove it unfortunately.

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
sodic: I'll create an issue to have a single source of truth for all our network request but am leaving it as is for now.

Done: https://github.com/wasp-lang/wasp/issues/3661

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
sodic: Covered here: https://github.com/wasp-lang/wasp/pull/3535#discussion_r2642681984.

I'll create an issue for a proper network module consolidating all of this.

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
sodic: Just in case I wanted something like 0.5 seconds. But int it is.

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
sodic: I know, but it's cumersome to use, should have been more specific with "easily" :)

I'd have to nest a pattern match over maybe to get the code out, and it's probably not useful to users. Do you think it's worth it.

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
sodic: I'd have to first convert the errors to the same type to use `join` which complicates the code, so I'll stick with the current version.

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
infomiho: It depends on what you need the status code for - if it's just to generate a more specific error message - I'd say okay, no need to go into that. If you think it'll help to be more precise in error handling, I'd go for it. You can extract the cumbersome bit into a helper function. 

RAW

sodic: Yes, I don't need it for the message. Ok then, not doing it :)

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
Martinsos: raw

PR #3536, opened 2025-12-22
[Wasp News] Local persistence

sodic:


waspc/src/Wasp/Util.hs
```diff
@@ -292,3 +295,11 @@ secondsToMicroSeconds = (* 1000000)
 
 findDuplicateElems :: (Ord a) => [a] -> [a]
 findDuplicateElems = map head . filter ((> 1) . length) . group . sort
+
+isOlderThanNHours :: Natural -> T.UTCTime -> IO Bool
+isOlderThanNHours nHours time = do
```
sodic, line 300: I didn't write tests for this because:
  - It's an old function.
  - It works in `IO`. I could have created a new pure function that compares timestamps and then tested that, but I decided to skip it (no great reason, just to save time).

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
```diff
@@ -0,0 +1,95 @@
...
+import GHC.Generics
+import StrongPath (Abs, File', Path', fromAbsDir, parent, relfile, (</>))
+import qualified System.Directory as SD
+import Wasp.Cli.Command.News.Common (NewsEntry (..))
+import Wasp.Cli.FileSystem (getUserCacheDir, getWaspCacheDir)
+import Wasp.Util (ifM, isOlderThanNHours)
+import qualified Wasp.Util.IO as IOUtil
+
+-- | News state stored on disk.
+data LocalNewsState = LocalNewsState
```
sodic, line 30: FYI: I didn't call it `NewsCache` on purpose as it's not really cache.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
```diff
@@ -0,0 +1,80 @@
...
+setLastReportTimestamp time state =
+  state
+    { lastReportAt = Just time
+    }
+
+markNewsAsSeen :: [NewsEntry] -> LocalNewsState -> LocalNewsState
+markNewsAsSeen newsEntries state =
+  state
+    { seenNewsIds = seenNewsIds state `Set.union` Set.fromList ((.id) <$> newsEntries)
+    }
```
sodic: I skipped tests for this. Let me know if you think they're necessary.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
```diff
@@ -0,0 +1,80 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
```
sodic, line 2: Needed to derive `FromJSON` and `ToJSON`

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
```diff
@@ -0,0 +1,69 @@
...
+
+markNewsAsSeen :: [NewsEntry] -> LocalNewsState -> LocalNewsState
+markNewsAsSeen newsEntries state = state {seenNewsIds = unionOfOldAndNewIds}
+  where
+    unionOfOldAndNewIds = seenNewsIds state <> Set.fromList (map (.id) newsEntries)
+
+getNewsStateFilePath :: IO (Path' Abs File')
+getNewsStateFilePath = do
+  waspCacheDir <- getWaspCacheDir <$> getUserCacheDir
+  return $ waspCacheDir </> [relfile|news.json|]
```
sodic, line 69: This can be a oneliner but I find the `do` block more readable.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
```diff
@@ -0,0 +1,69 @@
...
+  SD.createDirectoryIfMissing True $ fromAbsDir $ parent newsStateFile
+  writeJsonFile newsStateFile localNewsState
+
+obtainLocalNewsState :: IO LocalNewsState
+obtainLocalNewsState = do
+  stateFile <- getNewsStateFilePath
+  ifM
+    (IOUtil.doesFileExist stateFile)
+    (tryReadingLocalNewsStateOrUseEmpty stateFile)
+    (return emptyLocalNewsState)
```
sodic, line 48: We treat malformed files as non-existing files. There's nothing special to do here as they'll be overwritten on write.

Tell me if you think that's a problem.

waspc/src/Wasp/Util/Json.hs
```diff
@@ -28,7 +30,10 @@ parseJsonWithComments jsonStr = do
...
-  liftIO $ writeJsonValue jsonFilePath $ updateFn jsonContent
+  liftIO $ writeJsonFile jsonFilePath $ updateFn jsonContent
 
-writeJsonValue :: Path' Abs (File f) -> Value -> IO ()
-writeJsonValue file = IOUtil.writeFileBytes file . encode
+readJsonFile :: (FromJSON a) => Path' Abs (File f) -> IO (Maybe a)
+readJsonFile = fmap decode . IOUtil.readFileBytes
+
+writeJsonFile :: (ToJSON a) => Path' Abs (File f) -> a -> IO ()
+writeJsonFile file = IOUtil.writeFileBytes file . encode
```
cprecioso, line 39: Aeson has `decodeFile` and `encodeFile`

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
```diff
@@ -0,0 +1,69 @@
...
+    markNewsAsSeen,
+  )
+where
+
+import Data.Aeson (FromJSON, ToJSON)
+import Data.Maybe (fromMaybe)
+import Data.Set (Set)
+import qualified Data.Set as Set
+import qualified Data.Time as T
+import GHC.Generics
```
cprecioso: nitpick

```suggestion
import GHC.Generics (Generic)
```

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
```diff
@@ -0,0 +1,69 @@
...
+
+markNewsAsSeen :: [NewsEntry] -> LocalNewsState -> LocalNewsState
+markNewsAsSeen newsEntries state = state {seenNewsIds = unionOfOldAndNewIds}
+  where
+    unionOfOldAndNewIds = state.seenNewsIds <> Set.fromList (map (.id) newsEntries)
+
+getNewsStateFilePath :: IO (Path' Abs File')
+getNewsStateFilePath = do
+  waspCacheDir <- getWaspCacheDir <$> getUserCacheDir
+  return $ waspCacheDir </> [relfile|news.json|]
```
cprecioso, line 69: suggestion: I'd bump the relative path to the top-level scope, so it has some prominence.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
```diff
@@ -0,0 +1,69 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE OverloadedRecordDot #-}
+
+module Wasp.Cli.Command.News.Persistence
+  ( LocalNewsState (..),
```
cprecioso, line 6: I am not sure I'd expose the fields, it seems like we want people to use the functions we define in this file.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
```diff
@@ -0,0 +1,69 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE OverloadedRecordDot #-}
```
cprecioso, line 3: This feels like cheating in Haskell 😂

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
```diff
@@ -0,0 +1,69 @@
...
+import qualified System.Directory as SD
+import Wasp.Cli.Command.News.Common (NewsEntry (..))
+import Wasp.Cli.FileSystem (getUserCacheDir, getWaspCacheDir)
+import Wasp.Util (ifM)
+import qualified Wasp.Util.IO as IOUtil
+import Wasp.Util.Json (readJsonFile, writeJsonFile)
+
+-- | News state stored on disk.
+data LocalNewsState = LocalNewsState
+  { lastReportAt :: Maybe T.UTCTime,
```
cprecioso, line 31: I am not sure it makes sense for this to be a Maybe, it doesn't seem like there'd be a valid state that has `seenNewsIds` but not a report time

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
cprecioso, line 48: Given we're ignoring errors altogether, and returning a fallback value, would it make more sense to skip the checking of the file and just do stuff, falling back if it errored out?

Something like (pseudo-code-ish):
```hs
fromRight emptyLocalNewsState <$> try (readLocalNewsState stateFile)
```

waspc/src/Wasp/Util/IO.hs
```diff
@@ -109,11 +111,14 @@ doesDirectoryExist = SD.doesDirectoryExist . SP.fromAbsDir
...
+readFileStrict :: Path' Abs (File f) -> IO Text
+readFileStrict = T.IO.readFile . SP.toFilePath
+
 readFileBytes :: Path' Abs (File f) -> IO B.ByteString
 readFileBytes = B.readFile . SP.fromAbsFile
 
-readFileStrict :: Path' Abs (File f) -> IO Text
-readFileStrict = T.IO.readFile . SP.toFilePath
+readFileBytesStrict :: Path' Abs (File f) -> IO BS.ByteString
+readFileBytesStrict = BS.readFile . SP.fromAbsFile
```
sodic, line 121: I had to add this one due to lock issues.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
```diff
@@ -0,0 +1,69 @@
...
+    seenNewsIds :: Set String
+  }
+  deriving (Generic, Show, Eq, FromJSON, ToJSON)
+
+saveLocalNewsState :: LocalNewsState -> IO ()
+saveLocalNewsState localNewsState = do
+  newsStateFile <- getNewsStateFilePath
+  SD.createDirectoryIfMissing True $ fromAbsDir $ parent newsStateFile
```
infomiho, line 39: I'd maybe pack this into a `ensureNewStateFilePath` so you can't get the path without the dir existing. 

EDIT: Maybe I spoke too soon, this might be overengineering it.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
infomiho: Yep, I'd give it a `File LocalNewsState` treatment as well.

waspc/src/Wasp/Util.hs
```diff
@@ -292,3 +295,11 @@ secondsToMicroSeconds = (* 1000000)
...
+  now <- T.getCurrentTime
+  let secondsSinceLastCheckIn = T.nominalDiffTimeToSeconds (now `T.diffUTCTime` time)
```
infomiho, line 302:
```suggestion
  let diffSeconds = T.nominalDiffTimeToSeconds (now `T.diffUTCTime` time)
```

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
sodic: You're right, it's a leftover from the previous system. I'll take care of it.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
sodic: Hmm, I think we did this because it was tricky to handle the error in non-pseudocode :)

But I'll double check, I like the suggestion.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
sodic: I tried but gave up it couldn't just be a path - we'd need a function that takes the `userCacheDir`.

But I can change it and have an extra top-level function, no preference on my side.

waspc/src/Wasp/Util/Json.hs
sodic: Nice! Didn't know about those.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
sodic: You're right. I needed to expose `lastReportAt` to detect frist time users, but I don't anymore (leftover from the same system as `Maybe T.UTCTime` :))

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
sodic: Yeah, I can finally work like a human :)

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
```diff
@@ -0,0 +1,69 @@
...
+import GHC.Generics (Generic)
+import StrongPath (Abs, File', Path', fromAbsDir, parent, relfile, (</>))
+import qualified System.Directory as SD
+import Wasp.Cli.Command.News.Common (NewsEntry (..))
+import Wasp.Cli.FileSystem (getUserCacheDir, getWaspCacheDir)
+import Wasp.Util (ifM)
+import qualified Wasp.Util.IO as IOUtil
+import Wasp.Util.Json (readJsonFile, writeJsonFile)
+
+-- | News state stored on disk.
```
Martinsos, line 29: This comment indicates that maybe name is not the best. If you had the need to say here it is stored on the disk. Or is it an impelmentation detail? If so, maybe comment should not be here.

LocalNews is also confusing, sounds like "local news" for the city or something :D. At least we don't have "local news" and "state news" :D.

But it's not easy to find a better name. `NewsStateOnDisk`? Too much "implementation" detail. And that comment does feel useful. Ok, I am fine with it I think.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
```diff
@@ -0,0 +1,69 @@
...
+data LocalNewsState = LocalNewsState
```
Martinsos, line 30: Oh this is all hm! I somehow expected it will also contain the actual news downloaded, their data. Where is that stored then? Or is re-fetched each time? Does that make sense, to refetch it each time? Maybe simplifies things for now?

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
```diff
@@ -0,0 +1,69 @@
...
+  { lastReportAt :: Maybe T.UTCTime,
```
Martinsos, line 31: This `report` keyword surprised me a bit -> where did it come from? We have terms "news", we have the concept of "seen". What is "report"?

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
```diff
@@ -0,0 +1,69 @@
...
+    seenNewsIds :: Set String
```
Martinsos, line 32: Nitpick, but why not "read" instead of "seen", that sounds more natural to me for news.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
Martinsos: Yeah maybe that would be a bit much, although nice idea in general.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
```diff
@@ -0,0 +1,69 @@
...
+  }
+  deriving (Generic, Show, Eq, FromJSON, ToJSON)
+
+saveLocalNewsState :: LocalNewsState -> IO ()
+saveLocalNewsState localNewsState = do
+  newsStateFile <- getNewsStateFilePath
+  SD.createDirectoryIfMissing True $ fromAbsDir $ parent newsStateFile
+  writeJsonFile newsStateFile localNewsState
+
+obtainLocalNewsState :: IO LocalNewsState
```
Martinsos, line 42: Nitpick: You had `save`, so I would go with `load` here which is the opposite.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
```diff
@@ -0,0 +1,69 @@
...
+wasNewsEntrySeen :: LocalNewsState -> NewsEntry -> Bool
+wasNewsEntrySeen state entry = entry.id `Set.member` state.seenNewsIds
+
+setLastReportTimestamp :: T.UTCTime -> LocalNewsState -> LocalNewsState
+setLastReportTimestamp time state = state {lastReportAt = Just time}
+
+markNewsAsSeen :: [NewsEntry] -> LocalNewsState -> LocalNewsState
+markNewsAsSeen newsEntries state = state {seenNewsIds = unionOfOldAndNewIds}
+  where
+    unionOfOldAndNewIds = state.seenNewsIds <> Set.fromList (map (.id) newsEntries)
```
Martinsos, line 64: Maybe nicer, RAW:
```suggestion
markNewsAsSeen :: [NewsEntry] -> LocalNewsState -> LocalNewsState
markNewsAsSeen newsEntries state = state {seenNewsIds = state.seenNewsIds <> newIds}
  where newIds = Set.fromList ((.id) <$> newsEntries)
```

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
Martinsos: You could have `localNewsStateInUserCacheDir :: Path' (Dir UserCacheDir) (File LocalNewsStateFile)`.

But I on the other hand am not sure if we need this at the top level scope at the moment, so RAW.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
Martinsos, line 1: Maybe rename this module to LocalNewsState? Persistence is a cool name :D, but one has to guess what it is about, and it is really about LocalNewsState.

waspc/src/Wasp/Util.hs
Martinsos: What if you wrote a test that runs in IO? Is that possible in HSpec? If it is, maybe you could do that -> have it run this functoin in IO, and test it by doing getCurrentTime, adding 5 hours go it, calling this one with 6 hours, ... . It is fast enough really, and it is not any kind of side effect taht we would mind.

waspc/src/Wasp/Util.hs
```diff
@@ -292,3 +295,11 @@ secondsToMicroSeconds = (* 1000000)
...
+  return $
+    let numSecondsInHour = 3600
```
Martinsos, line 304: Nitpick but I would put this in where, I don't see any reason to have to read this line here.

waspc/src/Wasp/Util.hs
Martinsos: Bump on this one

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
Martinsos: Hm, a small test would be ok I tihnk.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
Martinsos: Ok if you don't like fun code

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
Martinsos: nice catch

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
Martinsos: bump on this one

waspc/src/Wasp/Util.hs
sodic: This is an old function extracted to a common module. When I do that, I keep everything identical on purpose.

But, since the change is already merged now, I fixed it in https://github.com/wasp-lang/wasp/pull/3554

waspc/src/Wasp/Util.hs
sodic: Fixed in https://github.com/wasp-lang/wasp/pull/3554

waspc/src/Wasp/Util/Json.hs
sodic: Fixed in https://github.com/wasp-lang/wasp/pull/3554

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
sodic: Fields made private in  https://github.com/wasp-lang/wasp/pull/3554

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
sodic: Comment removed in https://github.com/wasp-lang/wasp/pull/3554

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
sodic: Yes, I believe we agreed it does.

Since we need to check whether there are new updates, we never show news without fetching anyway  And we don't really want to do that, there's no use case - we want to check daily and we want to show daily.

The cache therefore wouldn't spare us from making a network request. It could only reduce the payload size (e.g., if we first checked whether there's anything new to fetch and only then fetched the payload). But:

- That's an unnecessary optimization. The news are tiny and the round trip time is the same regardless of their size.
- It would require a smarter server.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
sodic: I wanted to do read but the trouble is that the past tense (read) is the same as the present tense (again read), and I wanted to explicitly communicate the past tense.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
sodic: Done in https://github.com/wasp-lang/wasp/pull/3554

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
sodic: I trust it became clear in the next PR. We might rename it but nothing to do here anyway.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
sodic: Fixed in https://github.com/wasp-lang/wasp/pull/3554

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
sodic: Fixed in https://github.com/wasp-lang/wasp/pull/3554

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
sodic: I didn't do it on purpose because we lose information.

I wanted to communicate: "We're setting the seen IDs to be the union of previous seen IDs and the new seen IDs" without requiring the look at the set operations (especially if they need to know the semantics of how `<>` works on `Set`s to understand what's happening).

So perhaps it's best to just stick with `Set.union` and then inline it. Yes, I think that's what I'll do.

sodic: I did something in between in the end. Check https://github.com/wasp-lang/wasp/pull/3554

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
sodic: Done in https://github.com/wasp-lang/wasp/pull/3554

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
sodic: Changed in https://github.com/wasp-lang/wasp/pull/3554

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
sodic: Added in https://github.com/wasp-lang/wasp/pull/3554

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
sodic: Didn't do this in the end. We had a chat over Discord. It's not so easy because the maybe bubbles out to everything then.

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
sodic: Yes, it involved type applications or annotations with an extra fromMaybe:

```haskell
loadLocalNewsState :: IO LocalNewsState
loadLocalNewsState = do
  stateFile <- getNewsStateFilePath
  eitherState <- try (fromMaybe emptyLocalNewsState <$> readJsonFile stateFile)
  return $ fromRight emptyLocalNewsState (eitherState :: Either SomeException LocalNewsState)

```

So I decided the old version was clearer and more explicit about what might fail.

waspc/src/Wasp/Util.hs
sodic: I didn't do it in the end. It's an old function and the test was annoying to write. I do have it on my list for issue creation.

waspc/src/Wasp/Util.hs
Martinsos: Ok, but I wouldn't restrict yourself to keeping all the same. I get that you do it so it is easier to see in PR that you only copied stuff, but then you are missing the opoprtuinty to improve things on the fly, which I tihnk is a shame. So I would rather loose the "look I just copied it" part.
RAW

Issue #3494, opened 2025-12-12
Implement `wasp news`


sodic:
We want a channel to tell users about important updates in wasp (e.g., new versions, security updates). We want to do this with a new CLI command: `wasp news`.

Here's how our prototype currently works:
- Calling `wasp news` explicitly prints out all news of all levels.
- `wasp start` commands prints all news once per day until the user runs explicitly runs `wasp news` to acknowledge them.

What's left to decide:
- How many priorities/levels do we want and what's their semantics (i.e., when to show and repeat what).
- Decide when to show news (just `wasp start`, all commands, etc.).
- Decide how we want to mark news as "read:"
  - When the user runs `wasp news` as we currently do.
  - Interactively - we block and wait until they acknowledge them. This might be annoying in the CI.
- Decide whether and how we want to mark unseen news.

What's left to implement:
- Implement the new/changed features we decide to have.
- Polish the code and remove all the hacks.
- Replace the news server with a static file.

sodic:
Decisions @Martinsos and I made:
- Wasp news have 3 levels: **Critical** (e.g., security vulnerabilities), **Important** (e.g., Wasp versions), and **Info** (not sure yet but maybe launch weeks).
- Users can see all news with the `wasp news` command.
- Running `wasp news` marks all shown news as "seen."
- We draw the news from a static json file on our website.
- We show some news during `wasp start` and we check for new news once per day:
  - **Critical** news (e.g., security vulenrabilities) block `wasp start` and ask users for explicit confirmation before the command continues. When the user interactively confirms they've read the news, they are considered "seen" and no longer show up on `wasp start`. Since we're only showing news on `wasp start`, this won't cause problems in the CI (plus, we'll add an env var for disabling `wasp news`).
  - **Important** news (e.g., new Wasp versions) are shown on `wasp start` but don't block execution. We fetch and show them once per day until the user explicitly runs `wasp news` to mark them as seen.
- **Info** news are only shown when users run `wasp news`, never on `wasp start`.

infomiho:
> Important news (e.g., new Wasp versions) are shown on wasp start but don't block execution. We fetch and show them once per day until the user explicitly runs wasp news to mark them as seen.

I think this might be too aggressive, I'd maybe show them up to 3 times or something like that. Or just once. 

EDIT: Okay, it's once per day. But still, what if there are 4, 5 important news that the user never cleared. Do news have an expiry date?

PR #3537, opened 2025-12-22
[Wasp News] Command and CLI wiring

sodic:


waspc/cli/src/Wasp/Cli/Command/Start.hs
```diff
@@ -23,6 +23,7 @@ import Wasp.Project.Common (dotWaspDirInWaspProjectDir, generatedCodeDirInDotWas
 -- It also listens for any file changes and recompiles and restarts generated project accordingly.
 start :: Command ()
 start = do
+  liftIO fetchAndReportMandatoryNews
```
sodic, line 26: Should this perhaps go out of the `start` command? We could also handle it in `Main.hs`.

NOTE: The current solution will block Wasp start once per day for a maximum of 2 seconds. I didn't want to do anything fancy like:
- Fech the news in the background and show them on the next run of `wasp start`.
- Fetch them in background and show them when the output quites down.
- etc

I didn't like the cost/benefit ratio of those ideas but am willing to reconsider.

waspc/cli/src/Wasp/Cli/Command/News.hs
```diff
@@ -0,0 +1,56 @@
...
+  liftIO fetchNews >>= \case
+    Left err -> throwError $ CommandError "Wasp news failed" err
+    Right newsEntries -> liftIO $ do
+      localNewsState <- obtainLocalNewsState
+      printNewsReportAndUpdateLocalState localNewsState $
+        makeVoluntaryNewsReport localNewsState newsEntries
+
+fetchAndReportMandatoryNews :: IO ()
+fetchAndReportMandatoryNews = do
+  isWaspNewsDisabled <- isJust <$> lookupEnv "WASP_NEWS_DISABLE"
```
sodic: This means that setting `WASP_NEWS_DISABLE=false` will also disable `wasp news`. Is that OK?

waspc/cli/src/Wasp/Cli/Command/News.hs
```diff
@@ -0,0 +1,56 @@
...
+  unless isWaspNewsDisabled $ do
+    localNewsState <- obtainLocalNewsState
+    whenM (isTimeForMandatoryReport localNewsState) $ do
+      fetchNewsWithTimeout 2 >>= \case
+        -- TODO: missing prefix for nicer output. Should we even output anything?
+        Left _err -> putStrLn "Couldn't fetch Wasp news, skipping."
```
sodic: I'd like this to be prefixed with `[ Wasp ]` but it's a pain to set up (I'd have to call a job, etc.).

Can it stay like this? Should I even print anything?

P.S. The current timeout is 2 seconds, what do we think about that?

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
sodic, line 1: The idea behind this module is to capture as much of [this logic](https://github.com/wasp-lang/wasp/issues/3494#issuecomment-3670417497) as we can with pure functions. I might add a comment on top of the file explaining what I'm about to explain now if you think it's useful.

The type `NewsReport` captures all variables we care about when displaying news to the user:
  - Which news to show.
  - Which news to consider seen after displaying the report.
  - Whether the user needs to interactively confirm they've read the report.

For now, we only support two kinds of reports:
- Voluntary reports - Reports that users ask for (currently done only when they call `wasp news`).
- Mandatory reports - The reports we force on users (currently only during `wasp start`).

All `NewsReports` are constructed from:
  - The local state of the news (`LocalNewsState`)
  - A list of `NewsEntry`s to consider for the report.
  
That should be enough to understand what's going on :)

Since this module provides a pure mapping between inputs (local state, news entries) and outputs (what to show, what to mark as seen, whether to require confirmation or not), it allows our unit and property tests to clearly specify exactly how `wasp news` should behave. Doing so is especially useful for this module because manually testing news is cumbersome and error-prone.

So I might even suggest looking at the tests before looking at the code.

waspc/cli/src/Wasp/Cli/Command/News.hs
```diff
@@ -0,0 +1,56 @@
...
+news =
+  liftIO fetchNews >>= \case
+    Left err -> throwError $ CommandError "Wasp news failed" err
+    Right newsEntries -> liftIO $ do
+      localNewsState <- obtainLocalNewsState
+      printNewsReportAndUpdateLocalState localNewsState $
+        makeVoluntaryNewsReport localNewsState newsEntries
+
+fetchAndReportMandatoryNews :: IO ()
+fetchAndReportMandatoryNews = do
```
sodic, line 33: Heads up for @Martinsos

You originally wanted a different name for this function: `ifNewsStaleUpdateAndShowUnseen`. I felt that name was too heavy on implementation details. As a caller, I just wan't to know someone will take care of reporting mandatory news for me.

The one part that is missing from the name is optionality. To fix that, we could call it something like `fetchAndReportMandatoryNewsIfDue` if you want. I probably wouldn't but don't mind it.

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
```diff
@@ -0,0 +1,97 @@
...
+    isRelevant = (>= Moderate) . level
+    isUnseen = not . wasNewsEntrySeen currentState
+
+isTimeForMandatoryNewsReport :: LocalNewsState -> IO Bool
+isTimeForMandatoryNewsReport state = case state.lastReportAt of
+  Nothing -> return True
+  Just lastReportAt' -> isOlderThanNHours 24 lastReportAt'
+
+printNewsReportAndUpdateLocalState :: LocalNewsState -> NewsReport -> IO ()
+printNewsReportAndUpdateLocalState localNewsStateBeforeReport newsReport = do
```
sodic, line 82: The only reason this function takes an instance of `LocalNewsState` is to avoid reading it twice if the caller already has it (e.g., like in `fetchAndReportMandatoryNews`).

On the flipside, this means that if the caller doesn't have it (e.g., like in the `news` function) they have to fetch it and pass it down.

Let me know if there's a better way to do it.

waspc/cli/src/Wasp/Cli/Command/News.hs
```diff
@@ -0,0 +1,56 @@
...
+  - Test what happens when we add new news on the server.
+
+  Future:
+  - Figure out what to do with the versions affected field.
+  - Properly type and validate stuff on the server.
+  - Decide how to deliver the news on the server.
+  - Maybe include the project in the monorepo (might make deployment more difficult).
+  - Figure out how to end tests.
+  - In `wasp news` output, mark the unread/new news
+-}
```
sodic: I'll remove this list before merging.

waspc/waspc.cabal
```diff
@@ -778,16 +780,20 @@ test-suite wasp-cli-tests
     , waspc
     , cli-lib
     , QuickCheck
+    , generic-arbitrary
```
sodic, line 783: More info on this in ReportTest.hs.

waspc/cli/tests/Wasp/Cli/Command/News/ReportTest.hs
```diff
@@ -0,0 +1,112 @@
...
+  arbitrary = genericArbitrary
+
+instance Arbitrary LocalNewsState where
+  arbitrary = genericArbitrary
+
+instance Arbitrary T.UTCTime where
+  arbitrary = T.UTCTime <$> arbitraryDay <*> arbitraryDiffTime
+    where
+      arbitraryDay = T.fromGregorian <$> choose (2010, 2020) <*> choose (1, 12) <*> choose (1, 28)
+      arbitraryDiffTime = T.secondsToDiffTime <$> choose (0, 86399)
```
sodic, line 112: The property tests require our types (`NewsLevel`, `NewsEntry`, `LocalNewsState`) to be instances of `Arbitrary`.

To easily define those instances using `genericArbitrary`, we need to define orphan instance for `T.UTCTime`. If we didn't use `genericArbitrary`, we would have to define our instances like this:
```hs
instance Arbitrary NewsLevel where
  arbitrary = elements [Low, Moderate, High]

instance Arbitrary NewsEntry where
  arbitrary =
    NewsEntry
      <$> arbitrary
      <*> arbitrary
      <*> arbitrary
      <*> arbitrary
      <*> arbitraryUTCTime

instance Arbitrary LocalNewsState where
  arbitrary =
    LocalNewsState
      <$> oneof [pure Nothing, Just <$> arbitraryUTCTime]
      <*> (Set.fromList <$> arbitrary)
``` 
The downsides of the `genericArbitrary` approach are:
- We're creating an orphan instance for a type from `Data.Time`, which isn't ideal.
- We need the `generic-arbitrary` dependency

Should I stick to the manual approach or keep the generic one and accept its downsides?

waspc/cli/tests/Wasp/Cli/Command/News/ReportTest.hs
```diff
@@ -0,0 +1,112 @@
...
+            { newsToShow = [],
+              requireConfirmation = False,
+              newsToConsiderSeen = newsEntries
+            }
+
+    prop "delegates to makeMandatoryNewsReportForExistingUser for existing users" $
+      \localNewsState newsEntries ->
+        not (isFirstTimeUser localNewsState)
+          ==> makeMandatoryNewsReport localNewsState newsEntries
+          == makeMandatoryNewsReportForExistingUser localNewsState newsEntries
```
sodic: This test is questionable, but I'd keep it. I want to capture:
  - For first time users, do this specific thing.
  - For everyone else, just call this other function and it's got it covered.

waspc/cli/tests/Wasp/Cli/Command/News/ReportTest.hs
sodic, line 1: These tests (especially property tests) should capture almost the exact specification of our news system.

I won't elaborate here because the tests should speak for themselves. If they don't, tell me how to improve them :)

waspc/cli/src/Wasp/Cli/Command/News/Persistence.hs
```diff
@@ -49,8 +50,10 @@ obtainLocalNewsState = do
   where
     tryReadingLocalNewsStateOrUseEmpty stateFile =
       fromMaybe emptyLocalNewsState <$> readJsonFile stateFile
-    emptyLocalNewsState =
-      LocalNewsState {lastReportAt = Nothing, seenNewsIds = Set.empty}
+
+emptyLocalNewsState :: LocalNewsState
+emptyLocalNewsState =
+  LocalNewsState {lastReportAt = Nothing, seenNewsIds = Set.empty}
```
sodic, line 56: I extracted this to use in other modules.

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
```diff
@@ -0,0 +1,99 @@
...
+      newsToConsiderSeen = newsEntries,
+      requireConfirmation = False
+    }
+
+makeMandatoryNewsReport :: LocalNewsState -> [NewsEntry] -> NewsReport
+makeMandatoryNewsReport currentState newsEntries
+  | isFirstTimeUser = showNothingAndMarkAllAsSeen
+  | otherwise = makeMandatoryNewsReportForExistingUser currentState newsEntries
+  where
+    isFirstTimeUser = currentState == emptyLocalNewsState
```
sodic, line 52: Users who delete or end up with malformed news state files are treated as first users, which means they won't be shown any news on the first `wasp start` they call after this event.

I decided that's OK. Let me know if you think otherwise.

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
```diff
@@ -0,0 +1,100 @@
...
+  Just lastReportAt' -> isOlderThanNHours 24 lastReportAt'
+
+printNewsReportAndUpdateLocalState :: LocalNewsState -> NewsReport -> IO ()
+printNewsReportAndUpdateLocalState localNewsStateBeforeReport newsReport = do
+  reportNews
+  when newsReport.requireConfirmation askForConfirmation
+  updateLocalNewsState
+  where
+    reportNews =
+      putStrLn $ intercalate "\n\n" $ map displayNewsEntry newsReport.newsToShow
```
sodic: This is how it looks like:

image

To see it for yourself, just run `wasp news`.

To test the `wasp start` behavior, you'll have to edit `~/.cache/wasp/news.json`. Here's some help:
```bash
alias first-time-user=`rm /home/filip/.cache/wasp/news.json`

# You need to have the news.old.json file for this to work
alias stale-news='yes | cp /home/filip/.cache/wasp/news.old.json /home/filip/.cache/wasp/news.json'
```
To test whether specific news are shown or not, I manually edit the file.

waspc/cli/exe/Main.hs
```diff
@@ -191,6 +194,7 @@ printUsage =
         cmd   "    info                  Prints basic information about the current Wasp project.",
         cmd   "    test                  Executes tests in your project.",
         cmd   "    studio                (experimental) GUI for inspecting your Wasp app.",
+        cmd   "    news                  Prints out the latest Wasp-related news.",
```
infomiho:
```suggestion
        cmd   "    news                  Shows the latest Wasp-related news and marks them as read.
```

waspc/cli/tests/Wasp/Cli/Command/News/ReportTest.hs
infomiho: I don't understand what is an orphan instance?

I have no issue with the manual approach 👍  it's clear

waspc/cli/src/Wasp/Cli/Command/Start.hs
infomiho: Doing it in `Main` would ensure the news are shown no matter the command. It would also keep the `start` command cleaner IMHO.

waspc/cli/src/Wasp/Cli/Command/News.hs
```diff
@@ -0,0 +1,56 @@
...
+
+news :: Command ()
+news =
+  liftIO fetchNews >>= \case
+    Left err -> throwError $ CommandError "Wasp news failed" err
```
infomiho:
```suggestion
    Left err -> throwError $ CommandError "Getting Wasp news failed" err
```

waspc/cli/tests/Wasp/Cli/Command/News/ReportTest.hs
sodic: > I don't understand what is an orphan instance?

From Haskell wiki:
> An orphan instance is a type class instance for class C and type T which is neither defined in the module where C is defined nor in the module where T is defined.

It's a problem when you define an orphan instance for something that doesn't belong to you (e.g., `Time`) and then the authors of that package later define their own instance.

waspc/cli/exe/Main.hs
Martinsos: The fact they are marked as read is maybe an implemetation detail. Although it could be useful to be explicit hm.
But I like `Shows` better also then "Prints out", that was a bit weird.
Btw it could be just "Read the latest Wasp-related news.", that kind of covers it all, although it is a bit ambigious, but I think it would work.

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
Martinsos: I like the idea of the module, having core logic done like this is the right way to go. I am not yet sure about the exact abstractions and stuff (Voluntary vs Mandatory sounds weird), but the basic principles sound correct.

And I am in general in favor of comments at the top of the module explaining the intention -> they dont' get stale so easily because they usually don't go into the implementation detail, but they do provide the idea of "intention", which at the module level is not so obvious from the code often.

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
```diff
@@ -0,0 +1,100 @@
...
+makeVoluntaryNewsReport _currentState newsEntries =
+  NewsReport
+    { newsToShow = newsEntries,
+      newsToConsiderSeen = newsEntries,
+      requireConfirmation = False
+    }
+
+makeMandatoryNewsReport :: LocalNewsState -> [NewsEntry] -> NewsReport
+makeMandatoryNewsReport currentState newsEntries
+  | isFirstTimeUser = showNothingAndMarkAllAsSeen
```
Martinsos, line 49: It's a bit weird reading this.

Mandatory -> I don't konw really at this point what mandatory means as a term, in the news context.

Also, I am sure there is a reason why we are showing nothing and marking all as seen, but it is hard to understand here, out of context completely, why is that the right move/logic.

I think the problem is that this function here is so isolated from the surrounding context that gives it meaning, that it loses the meaning itself and becomes kind of arbitraty. It is trying to be standalone, but it it is not doing a great job. I am not saying we should not try to extract code, but it doesn't seem at the moment this was the right way to do it.

One example: `newsEntries` input arg, I have no idea which news entries are those, whose is giving them to us here. Is it all news ever? Is it some specific news? What is it? You could say it doesn't matter, but what does matter? How can I look at this functoin and say, looking at what it should do, inputs, and outputs, that it makes sense? I don't undrestand inputs, and I also don't know what it is supposed to do.

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
```diff
@@ -0,0 +1,100 @@
...
+  | otherwise = makeMandatoryNewsReportForExistingUser currentState newsEntries
```
Martinsos, line 50: Ok and you make me read this other branch of logic in new function, although the one above you covered here. Seems assymetrical, I would rather this one is also heppnign here, or both are in its own function.

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
Martinsos: I don't think I would call it first time user. First time user of what? Of Wasp? What if they just had news shut off before? What if they just pulled in, for the first time, version of Wasp that can handle the news? Those are not first time users. The term `user` here came from nowhere also a bit, we haven't considered this term so far in News "universe".
You could say `noLocalNewsState` and that should be enough? Why do you need to make these assumptions?
I guess you wanted to make the reasoning clear -> the user never received any news so far, so we will mark the first batch of news as all read, to not overwhelm them. Maybe then call it `userNeverReceivedNews`? Or find another way to make this clear?

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
```diff
@@ -0,0 +1,100 @@
...
+  NewsReport
+    { newsToShow = allRelevantUnseenNews,
+      requireConfirmation,
+      newsToConsiderSeen =
+        if requireConfirmation
+          then allRelevantUnseenNews
+          else []
+    }
+  where
+    requireConfirmation = any ((== High) . level) allRelevantUnseenNews
```
Martinsos: I don't think ou got anything by putting this one into `where`, only made it more confusing, as you are mixing levels of detail.

Obviously you are using `allRelevantUnseenNews` as a thing above to define the fields of the record. So use it for this one also. It makes easier for me to read that piece of code, I can be like "aha these are all defined based on this allRelevantUnseenNews, ok I will see what that is later, but I can see how they relate". I can't if you pull this one out though.

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
```diff
@@ -0,0 +1,100 @@
...
+    allRelevantUnseenNews = filter isRelevant . filter isUnseen $ newsEntries
+    isRelevant = (>= Moderate) . level
```
Martinsos: How do we decide these are relevant? Because they are relevant ones in the context of mandatory news? I think I know this because I remmeber the discussoin from before, but I don't thikn I would be able to conclude this on my own if I didn't know that.

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
```diff
@@ -0,0 +1,100 @@
...
+    isUnseen = not . wasNewsEntrySeen currentState
+
+isTimeForMandatoryNewsReport :: LocalNewsState -> IO Bool
```
Martinsos, line 76: This name feels like it could be better. "Is it time" -> maybe time was long time ago hm.
How would you describe in words, to me, what this function isa bout? Let's try to create a name based on that.
I also wonder if this decision if it is time should come from outside. But it is connected to the whole concept of volontary and mandatory and so on which are still a bit unusual to me at this point of review.

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
Martinsos: I would be more worried about the desync -> what if the caller has one state of news, but state on the disk is different? In that case, I think it is better to go with what the caller has, so maybe this is ok as you did it. Although hm yeah it depends on how desync happened, it is not clear cut. I am not sure, maybe better not to overthink it.

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
```diff
@@ -0,0 +1,100 @@
...
+isTimeForMandatoryNewsReport state = case state.lastReportAt of
+  Nothing -> return True
+  Just lastReportAt' -> isOlderThanNHours 24 lastReportAt'
+
+printNewsReportAndUpdateLocalState :: LocalNewsState -> NewsReport -> IO ()
+printNewsReportAndUpdateLocalState localNewsStateBeforeReport newsReport = do
```
Martinsos, line 82: `printNewsReportAnd...` is misleading due to `print`.
You are not always just printing, you are possibly also asking for confirmation.
Maybe instead do `reportNewsAndUpdateLocalState`.

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
```diff
@@ -0,0 +1,100 @@
...
+  reportNews
+  when newsReport.requireConfirmation askForConfirmation
+  updateLocalNewsState
+  where
+    reportNews =
+      putStrLn $ intercalate "\n\n" $ map displayNewsEntry newsReport.newsToShow
+
+    askForConfirmation = do
+      let requiredAnswer = "ok"
```
Martinsos: Tiny thing, but why not go with "y" which is more standard?

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
```diff
@@ -0,0 +1,100 @@
...
+      answer <-
+        askForInput $
+          "\nPlease type '" ++ requiredAnswer ++ "' to confirm you've read the announcements: "
```
Martinsos:
```suggestion
          "\nDo you want to mark these news as read and stop seeing them? Type "y" if so:"
```

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
Martinsos, line 1: Ok I read this module.

I still think voluntary and mandatory are not great names here.

I thought instead of how they do it with LLMs now: "user invoked" vs "agent invoked". Maybe that is better direction for naming. Because if I asked you to explain these terms, how would you explain them?

I would go something like:
Voluntary means that user asked to see the news, while Mandatory means that we decided (the Wasp CLI) that they should see these news now, which we might want to do when there are important news and they haven't read them yet.

So yeah, reading this, I would say it depends on who is triggering the news? So it could be "userInvokedNewsReport" and "waspInvokedNewsReport". Not sure I even like the name report, I would be tempted to go for "Listing" instead hah, or "Showing". Ok maybe I am going too far now.

Or maybe we shold go with `AllNewsListing` and `FreshNewsListing` (or `NewsNeedingAttentionListing` or `NewsInAttentionListing`).

And you could also impelment the `NewsListing` in a different way. Meaning you would have
```hs
data NewsListing = AllNewsListing [NewsEntry] | FreshNewsListing [NewsEntry]

newsToShow :: LocalNewsState -> NewsListing -> [NewsEntry]
newsToShow state (AllNewsListing news) = ...
newsToShow state (FreshNewsListing news) = ...

requiresConfirmation :: LocalNewsState -> NewsListing -> Bool
```

This feels better because you capture possible states more tightly and I can observe them in the types, instead of having a more loose type where actually some of the fields are just derived from the rest of the state. Your approach is more flexible, but I don't think we are very likely to have more than these two possible reports/listings, maybe 3 at some point.

I am not 100% sure about all this, but it feels more "natural", while what you have right now feels somewhat "engineered" but not in a great way. It feels to me like if we first wrote code to be "normal" and then extracted this from it, it would not be like you have it. Although you probably did exactly that hm. But maybe you went too much into "abstract" thinking in this process.

I would have to explore this further to exaclty figure out what is the right direction, but I think this might be enough for you to make the next step.

waspc/cli/src/Wasp/Cli/Command/News.hs
Martinsos: I would say "Fetching" hehe

waspc/cli/src/Wasp/Cli/Command/News.hs
```diff
@@ -0,0 +1,56 @@
...
+    Right newsEntries -> liftIO $ do
+      localNewsState <- obtainLocalNewsState
+      printNewsReportAndUpdateLocalState localNewsState $
+        makeVoluntaryNewsReport localNewsState newsEntries
```
Martinsos, line 30: I thikn you should be able to use shortcircuiting of Command in similar fashion like this (this is kind of pseudocode, I don't thikn it compiles, but I think you will know how to fix it:

```suggestion
news =
  newsEntries <- (liftIO fetchNews) <&> fromLeft (throwError . CommandError "Wasp news failed")
  liftIO $ do
    localNewsState <- obtainLocalNewsState
    printNewsReportAndUpdateLocalState localNewsState $
      makeVoluntaryNewsReport localNewsState newsEntries
```

waspc/cli/src/Wasp/Cli/Command/News.hs
Martinsos: Makes sense. I am not a fan of "mandatory" though, I commented in another place, so let's see how that will affect this. But I am ok with the idea.
I would probably add the "optionality" to it because I don't like that being a surprise, if I am reading this at the call site, I wouldn't think it has that optionality built in, even if one might claim it to be an impelentation dteails, but it is a ehaviour taht affects user. But RAW.

waspc/cli/src/Wasp/Cli/Command/News.hs
Martinsos: Ha uff good question! Well hm that sounds weird, why would it also disable `wasp news` if that is an explicit message from the user that they want to see the news? I think that is why you asked also. So I don't think it should disable `wasp news`. I don't think it is a disaster if we merge it as it is and fix it afterwards, but I do think we should change this.

waspc/cli/src/Wasp/Cli/Command/News.hs
```diff
@@ -0,0 +1,56 @@
...
+
+fetchAndReportMandatoryNews :: IO ()
+fetchAndReportMandatoryNews = do
+  isWaspNewsDisabled <- isJust <$> lookupEnv "WASP_NEWS_DISABLE"
```
Martinsos: How will this affect CI -> we can expect some people will get tripped by this?If they are using `wasp start` in the CI? And while generally they shouldn't, they could use it for e2e tests. Ok but that is why we have timeout. All right I guess! I wanted to suggest we use that unfction for detecting CI and if it gives true, we skip this with the news. Yeah why don't we do that? If it saves people from waiting for 2 or 5 seonds in some situations, why not?

waspc/cli/src/Wasp/Cli/Command/News.hs
Martinsos: Sounds potentially short if someobdy has a slower connection? But there i snot many news and fethicng them should be fast, so it should be enough even on the slow connections. Hm I don't know.

Would be cool to do what you suggested kind of, although you suggested it a bit differnetly, but we could wait here for 2 seconds, if they don't arrive, we continue with wasp start, but we still keep fetching them in the background so they are fetched for the next time.

Btw this is quite easy to implement in a way where you, if they were not fetched here, just concurrently run fetching (but not showing) and continueation of wasp start, so it happens in the background. Aha but you don't have storage of news do you? Yeah ok it was easier this way.

Ok just thinking here, not sure about anything. RAW.

Martinsos: Oh btw I think no prefix is ok, we don't need it, also, you don't need to mabye print we skipped them. Maybe we shoud print we are fetching them so they know why they are waiting? But 2 seconds is not long. Let's maybe not print anything for now? Not sure myself.

waspc/cli/src/Wasp/Cli/Command/Start.hs
Martinsos: Yeah but doing it in Main means it will trigger in all possible situations, like CI and what not, and also I think it is annotying if news attack you when you are doing wasp db migrate dev or something. Better to keep it contained / expected, so we can control which workflwos it possibly interrupts (be it CI, LLMs using it, ...). And `wasp start` is safe in a way that we know they will run it if they are doing anything serious.

Btw @sodic  this is not obvious at a glance, and this deciison might be forgotten, so I would write this down somewhere maybe in some comment, here probably, for future develoeprs looking at this and thinking they might want to move it into Main.hs .

waspc/cli/tests/Wasp/Cli/Command/News/ReportTest.hs
```diff
@@ -0,0 +1,112 @@
...
+import Wasp.Cli.Command.News.Persistence (LocalNewsState (..), emptyLocalNewsState)
+import Wasp.Cli.Command.News.Report
+  ( NewsReport (..),
+    makeMandatoryNewsReport,
+    makeMandatoryNewsReportForExistingUser,
+    makeVoluntaryNewsReport,
+  )
+
+spec_makeVoluntaryNewsReport :: Spec
+spec_makeVoluntaryNewsReport = do
```
Martinsos, line 22: On one hand this is very cool, on the othe rhand, I wonder it these are expensive and if we will be doing them for "trivial" tihngs like this, will it slow our tests down. But on the other hand, it is ok to have test for stuff like this, even if somewhat trivial. I guess it is ok. ALthough it would probably also be fine if it was just a single, non-prop test case. But all good. RAW

waspc/cli/tests/Wasp/Cli/Command/News/ReportTest.hs
Martinsos: Yeah orphan instances are mostly rpoblem if you are writing a library, but if you are not, it is not such a big deal.

And this is probably a good example when it is ok to use orphan instances. At least for News... stuff, for UTCTime it is a bit trickier because you in theory just defined here how you like it for these tests, but what if some other tests also need it? I would suggest to move Arbitrary instance of UTCTime to a more general module in tests. Or to go manual route, as you said.

Btw something rings a bell to me that it is possible to do all these prop checking without having arbitrary instances like this. I can't remember what were the methods. I can imagine having newtypes here for each of these, that you use like `\(TestNewsLevel newsLevel) ...`, and they are instances of Arbitrary and local to this module, but yeah those would have to be defined manually, but you avoid having orphaned instances + you can in theory have multiple definitions of how NewsLevel is arbitrary for different tests. But then one what about nested stuff -> yeah well the same way, manually. Hm.
But I guess this might be too much for now and we can just go as you did and can adapt it in the future if needed. Just move the UTCTime to some more general file.

waspc/cli/src/Wasp/Cli/Command/News.hs
sodic: Keep in mind that this is how Telemetry works too. Setting it to anything (even false) disables telemetry. Also, Bash doesn't have booleans, so we'd have to decide which strings to interpret as truthy or falsly (e.g., "ok", "yes", "false", "on"...).

So if we're changing it, I'd change it for all env vars we read (news, telemetry, perhaps something else).

Martinsos: Yeah but I am not talking about that, I am ok with us handling value of the env var in "stupid" fashion, I am instead saying that `wasp news` command should never get disabld, there is no need, we only want to disable automatic showing of news.

Btw here is my not yet merged code for checking the truthiness: https://github.com/wasp-lang/wasp/pull/3403/changes#diff-4773d6c26558eedb85689e81c2323eedb7725ed7f0f53bb9ce7e2de6a39cfcecR16

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
infomiho: I'd love it if we encoded this at the module level e.g. `requiresConfirmation` or `isRelevant`, kinda like a part of the type system.

infomiho:
SCR-20251223-jogl

I found it kinda hard to read - maybe we should go with white instead of gray for the text color?

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
infomiho: I don't we should go with `Do you want to mark` since it makes it look like they have a choice. If they say no, the next it will pop up again, and again etc.

I'm missing some sort of reasoning here why am I interrupted and my input is required. I want an explanation like "Since these are news of high priority, please confirm that you saw them" or similar. "Some announcements have a high or critical priority, so we need you to confirm that you saw them. We'll show them again every 24 hours until confirmed." 

I'm asking for this extra explanation since it's quite the interrupt and there better be a good explanation why or I'm annoyed.

waspc/cli/src/Wasp/Cli/Command/Start.hs
infomiho: Image

Extra question, do we want to introduce some delay between showing the news and starting with the compilation? Just to give it more chance of being seen.

waspc/cli/src/Wasp/Cli/Command/News.hs
sodic: There was a misunderstanding here. Nothing disables `wasp news`, the command. So we're all good.

waspc/cli/src/Wasp/Cli/Command/News.hs
sodic: This wouldn't work because of `throwError`. I'd have to do something like:

```haskell
news = do
  newsEntries <- either (throwError . CommandError "Getting Wasp news failed") pure =<< liftIO fetchNews
  liftIO $ do
    localNewsState <- loadLocalNewsState
    executeNewsAction localNewsState $ makeUserInvokedNewsAction newsEntries
```

It's shorter, but I might even prefer the original version because of the consistent left to right flow: fetch the news -> do this when it fails -> or do this when it succeeds.  
  
I'll think about it.

sodic: I did it in the end: https://github.com/wasp-lang/wasp/pull/3554

waspc/cli/src/Wasp/Cli/Command/Start.hs
sodic: We had a misunderstanding. I didn't mean "do we want to do it on other commands," I meant "should I move the code that calls this function into the branch that handles the start command in main.hs.  
  
@infomiho Timeout - I tried it in https://github.com/wasp-lang/wasp/pull/3554 but removed it in the end. Not sure what to do since critical news already have a timeout.  
  
@martinsos Comment - Added in https://github.com/wasp-lang/wasp/pull/3554

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
sodic: Done back before I merged this PR. Latest version in https://github.com/wasp-lang/wasp/pull/3554

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
sodic: I changed this code in the end. Didn't bring it up to the top-level because it's very specific to showing the news under one particular scenario. But I did choose a better name: https://github.com/wasp-lang/wasp/pull/3554

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
sodic: We talked about a new module and I implemented it in https://github.com/wasp-lang/wasp/pull/3554. Also added a comment.

waspc/cli/src/Wasp/Cli/Command/News.hs
sodic: We talked about this in person. In https://github.com/wasp-lang/wasp/pull/3554, I changed the name to `fetchAndListMustSeeNewsIfDue` which should cover all concerns.

waspc/cli/tests/Wasp/Cli/Command/News/ReportTest.hs
sodic: Took a new approach in https://github.com/wasp-lang/wasp/pull/3554 and got rid of orphan instances.  
  
I realized these property tests aren't ideal because, since news are just random strings, the `localNewsState` will never contain an ID that might appear in the report.

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
sodic: I called it `userHasNoNewsHistory` in https://github.com/wasp-lang/wasp/pull/3554. Resolving because we talked about it in person.

waspc/cli/src/Wasp/Cli/Command/News.hs
sodic: Implemented Miho's change before merging this PR. I ddin't use fetching because the internal error already talks about fetching.

waspc/cli/src/Wasp/Cli/Command/News.hs
sodic: I addressed this before merging the PR.

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
sodic: I implemented this approach in https://github.com/wasp-lang/wasp/pull/3554.

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
sodic: There's new code and a new name in https://github.com/wasp-lang/wasp/pull/3554

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
sodic: Name changed in https://github.com/wasp-lang/wasp/pull/3554.  
  
New name is `shouldWaspListMustSeeNews`

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
sodic: Already done in this PR.

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
sodic: No longer applicable, changed in https://github.com/wasp-lang/wasp/pull/3554

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
sodic: No longer applicable, changed in https://github.com/wasp-lang/wasp/pull/3554

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
sodic: We changed this in https://github.com/wasp-lang/wasp/pull/3554

waspc/cli/exe/Main.hs
sodic: Done before I merged this PR.

waspc/cli/src/Wasp/Cli/Command/Start.hs
Martinsos: Aha ok! So does it get moved to caller, just before calling `start`. Ok, doesn't matter much to me, I am ok with anything. Maybe even better in `start`.

PR #3539, opened 2025-12-22
Create WaspInfo

cprecioso:
## Description

- Part of #3540
- On top of #3538

Adds `BuildType` information to our existing `.wasp/{out,build}/.waspinfo` file and makes it machine-readable.

Checks it in `wasp compile` and removes the full directory output if it was built with a different wasp version or build type.

### Motivation

I'm preparing to "unsplit" `.wasp/out` and `.wasp/build` in #3540. Given that there's variance in not only the contents of the files, but in which files are created (like whether an `.env` file is copied or not), it is safer to do a clean build of the directory when changing from `wasp start` to `wasp build` (this was already happening), but also from `wasp build` back to `wasp start`.

## All the PRs

- #3540
- #3539
- #3538

<details>
<summary>Checklist not applicable, click here to show anyway</summary>

## Type of change

- [ ] **🔧 Just code/docs improvement** 
- [ ] **🐞 Bug fix** 
- [ ] **🚀 New/improved feature** 
- [ ] **💥 Breaking change** 

## Checklist

- [ ] I tested my change in a Wasp app to verify that it works as intended.

- 🧪 Tests and apps:

  - [ ] I added **unit tests** for my change. 
  - [ ] _(if you fixed a bug)_ I added a **regression test** for the bug I fixed. 
  - [ ] _(if you added/updated a feature)_ I added/updated **e2e tests** in `examples/kitchen-sink/e2e-tests`.
  - [ ] _(if you added/updated a feature)_ I updated the **starter templates** in `waspc/data/Cli/templates`, as needed.
  - [ ] _(if you added/updated a feature)_ I updated the **example apps** in `examples/`, as needed.
    - [ ] _(if you updated `examples/tutorials`)_ I updated the tutorial in the docs (and vice versa).

- 📜 Documentation:

  - [ ] _(if you added/updated a feature)_ I **added/updated the documentation** in `web/docs/`.

- 🆕 Changelog: _(if change is more than just code/docs improvement)_
  - [ ] I updated `waspc/ChangeLog.md` with a **user-friendly** description of the change.
  - [ ] _(if you did a breaking change)_ I added a step to the current **migration guide** in `web/docs/migration-guides/`.
  - [ ] I **bumped the `version`** in `waspc/waspc.cabal` to reflect the changes I introduced.

 

</details>

Issue #3504, opened 2025-12-16
`package-lock.json` changes depending on `wasp build`, `wasp clean` and `wasp compile`


sodic:
Version: 0.19.2

Running `wasp clean`, `wasp build` or `wasp compile` (directly or through `wasp start`) changes the user's `package.json`. This was probably caused by #3130.

It's a problem because:
- Our project looks unstable.
- User's can't use their `package.json` files normally.
- This might be causing our CI randomly to fail on example apps without much info (check waspello in [this run](https://github.com/wasp-lang/wasp/actions/runs/20238073858/job/58098086634)), commiting the `package.json` after running `wasp compile` usually fixes this (check waspello in [this run](https://github.com/wasp-lang/wasp/actions/runs/20240475543/job/58106933265)). I'm not sure this is the cause though, the CI logs suggest it's something else.

#3163 might solve a part of the problem but I'm not sure it's going to solve the entire problem.

FranjoMindek:
I've encountered the same thing.
Ephemeral tests fail two random example app tests: https://github.com/wasp-lang/wasp/pull/3465

cprecioso:
Probably fixed by #3540

cprecioso:
Merged #3163, which deduplicates the workspaces. Now `package-lock.json` will not change if running `wasp compile` or `wasp build` anymore. It will still change if you run `wasp clean` and `npm i` without running a compile in between, but I think that's well to be expected.
Closing

PR #3540, opened 2025-12-22
Merge `.wasp/out` and `.wasp/build`

### LINKED ISSUE #3163: Unsplit `.wasp/build` and `.wasp/out`
While working on #3130, I ran into a problem wrt us outputting the server and web-app packages twice, with minimal differences. We should review whether it makes sense for us to still split the `.wasp/build` and `.wasp/out` folders, since they mostly have the same things.

This task is synergic with #2150 

### Current state review

As of [`b10ebbf` (2025-09-17)](https://github.com/wasp-lang/wasp/tree/b10ebbf24c8f97258b50ffe0e8675468b6ba437f), the changes in behaviour for `out` and `build` are that `wasp build`...:
- In Haskell code:
  - Checks email sender is not `Dummy`
  - Checks database is not SQLite
  - Doesn't check migrations
  - Doesn't emit `.env` files in server and webapp dirs
  - Removes all the previous output and outputs stuff from scratch
  - Copies some files from the user project to the build dir so Docker picks them up [(code)](https://github.com/wasp-lang/wasp/blob/b10ebbf24c8f97258b50ffe0e8675468b6ba437f/waspc/cli/src/Wasp/Cli/Command/Build.hs#L88)
- In JS code:
  - Ignores the env to skip email verification

> [!NOTE]
> These were found by doing a grep for `isBuild` and `isDevelopment`, and by comparing the `wasp build` and `wasp start` steps.

How to solve differences
- Checks can still happen differently based on the command
- Difference in outputted files (.env and docker-needed files) can be brought into the `FileDraft` system, marking them specifically for deletion if they are not needed.
- Different behaviour on JS code can happen on `process.env.NODE_ENV` instead of statically on compilation.

cprecioso:
- Related to #1769
- On top of #3539
- Resolves #3163

## Description

- Removes the `.wasp/out` and `.wasp/build` split.
- Updates references that pointed to `.wasp/build`.
- `wasp build` explicitly built `.wasp/out/sdk`. Removed it as it's now unneeded.
- Updates the `Command` `Requirable`s, since they were checking for `.wasp/build` existence as a proxy for knowing if the `wasp start` or `wasp build` had been done. Now they read the [.waspinfo file](https://github.com/wasp-lang/wasp/pull/3539) to be certain of which build has been performed.
- Removes the requirement to have the `.wasp/build/*` workspace.
	- It is now forbidden to have it, since we don't want npm to be picking up leftovers and fudging the resolution.
- Makes build and dev generated packages have the same name, avoiding double resolutions.

> [!NOTE]
> This PR does not attempt to solve the differences between our `wasp build` and `wasp start` processes (e.g. the extra copied files like the user `tsconfig` or `.env` files). However, we ensure one build process does not contaminate the other by [doing a cleanbuild if needed](https://github.com/wasp-lang/wasp/pull/3539).

### Motivation

Context: #1769

Our generated code is working on an assumption that does not hold true:
- **Assumption:** production output is independent of development output (we build to different folders)
- **Breaks:** `.wasp/build` depends on `.wasp/out/sdk`

We are doing work to maintain this separation when it was not needed, trying to make as if our process is not modal. Some complications we have are: [copying user code to our generated output](https://github.com/wasp-lang/wasp/issues/1769#issuecomment-1952938100); or [having to add each npm workspace twice to the user's package-lock](https://github.com/wasp-lang/wasp/issues/3504).

We're better off accepting that these two builds aren't independent and make them mutually exclusive, so we can build them in the same place, avoiding workarounds.

## Type of change

- [ ] **🐞 Bug fix** 
- [x] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/cli/src/Wasp/Cli/Command/Compile.hs
+++ b/waspc/cli/src/Wasp/Cli/Command/Compile.hs
@@ -16,6 +16,7 @@ import Control.Monad.Except (throwError)
 import Control.Monad.IO.Class (liftIO)
 import Data.List (intercalate)
 import StrongPath (Abs, Dir, Path', (</>))
+import qualified StrongPath as SP
 import qualified Wasp.AppSpec as AS
 import Wasp.Cli.Command (Command, CommandError (..))
 import Wasp.Cli.Command.Message (cliSendMessageC)
@@ -49,21 +50,23 @@ compile = do
 compileWithOptions :: CompileOptions -> Command [CompileWarning]
 compileWithOptions options = do
   InWaspProject waspProjectDir <- require
-  let outDir =
-        waspProjectDir
-          </> dotWaspDirInWaspProjectDir
-          </> generatedCodeDirInDotWaspDir
 
-  isCleanBuildNeeded <-
-    liftIO $
-      buildType options `WaspInfo.isCompatibleWithExistingBuildAt` outDir
+  let relOutDir = dotWaspDirInWaspProjectDir </> generatedCodeDirInDotWaspDir
+      outDir = waspProjectDir </> relOutDir
 
-  doesOutDirExist <- liftIO $ doesDirectoryExist outDir
+  generatedCodeIsCompatible <-
+    liftIO $ buildType options `WaspInfo.isCompatibleWithExistingBuildAt` outDir
 
-  when (isCleanBuildNeeded && doesOutDirExist) $ do
-    cliSendMessageC $ Msg.Start "Clearing the content of the .wasp/out directory..."
+  outDirExists <- liftIO $ doesDirectoryExist outDir
+
+  when (outDirExists && not generatedCodeIsCompatible) $ do
+    cliSendMessageC $
+      Msg.Start $
+        "Clearing the content of the " ++ SP.fromRelDir relOutDir ++ " directory..."
     liftIO $ removeDirectory outDir
-    cliSendMessageC $ Msg.Success "Successfully cleared the contents of the .wasp/out directory."
+    cliSendMessageC $
+      Msg.Success $
+        "Successfully cleared the contents of the " ++ SP.fromRelDir relOutDir ++ " directory."
 
   cliSendMessageC $ Msg.Start "Compiling wasp project..."
   (warnings, errors) <- liftIO $ compileIOWithOptions options waspProjectDir outDir


wasp-app-runner/src/waspCli.ts
```diff
@@ -66,6 +67,40 @@ export function waspBuild({
...
+  });
+  const stdoutDataWithoutAnsiChars = stripVTControlCharacters(stdoutData);
+
+  if (exitCode !== 0) {
+    logger.error(`Failed to get wasp version: ${stdoutDataWithoutAnsiChars}`);
+    process.exit(1);
+  }
+
+  const [firstLine] = stdoutData.split("\n");
+  const waspVersion = firstLine?.trim();
```
infomiho: Does `semver` package offer some sort of "parse version from input" we can maybe use here?

waspc/src/Wasp/Generator/Valid/PackageJson/Workspaces.hs
```diff
@@ -15,19 +15,30 @@ type WorkspaceName = String
...
+forbiddenWorkspaces :: [WorkspaceName]
+forbiddenWorkspaces = [".wasp/build/*"]
+
 workspacesValidator :: V.Validator P.PackageJson
 workspacesValidator =
   V.inField ("workspaces", P.workspaces) $
-    V.all [requiredWorkspaceValidator]
+    V.all
+      [ requiredWorkspaceValidator,
+        forbiddenWorkspaceValidator
```
infomiho: Would it make sense to simplify this to a simple equals check? "The `workspaces` field must be this list."

If were writing the compiler from scratch we'd do it like this. I understand that many people will have the workspaces field set to the <=0.20.0 value. I'm on the fence if we should cater to that. It's a breaking change and we are in beta right now.

waspc/cli/src/Wasp/Cli/Command/Build.hs
```diff
@@ -55,22 +54,13 @@ build = do
   let buildDir =
         waspProjectDir
           </> dotWaspDirInWaspProjectDir
-          </> buildDirInDotWaspDir
+          </> generatedCodeDirInDotWaspDir
 
   doesBuildDirExist <- liftIO $ doesDirectoryExist buildDir
   when doesBuildDirExist $ do
-    cliSendMessageC $ Msg.Start "Clearing the content of the .wasp/build directory..."
+    cliSendMessageC $ Msg.Start "Clearing the content of the .wasp/out directory..."
```
infomiho: I'm wondering if a something like `show waspProjectDir` would be a useful refactor here?

web/docs/migration-guides/migrate-from-0-20-to-0-21.md
```diff
@@ -0,0 +1,84 @@
...
+
+```sh
+curl -sSL https://get.wasp.sh/installer.sh | sh -s -- -v 0.21.0
+```
+
+## What's new in 0.21.X?
+
+### Merged the `.wasp/build` and `.wasp/out` directories
+
+In previous versions of Wasp, there were two separate directories for generated code: `.wasp/build` (used in development mode) and `.wasp/out` (used in production mode). Starting from Wasp 0.21.X, only the `.wasp/out` directory is used for generated code in both development and production modes. This change simplifies the project structure and reduces confusion.
```
infomiho:
```suggestion
In previous versions of Wasp, there were two separate directories for generated code: `.wasp/out` (used in development mode) and `.wasp/build` (used in production mode). Starting from Wasp 0.21.X, only the `.wasp/out` directory is used for generated code in both development and production modes. This change simplifies the project structure and reduces confusion.
```

waspc/src/Wasp/Generator/Valid/PackageJson/Workspaces.hs
```diff
@@ -15,16 +15,26 @@ type WorkspaceName = String
...
 
+forbiddenWorkspaces :: [WorkspaceName]
+forbiddenWorkspaces = [".wasp/build/*"]
+
 workspacesValidator :: V.Validator P.PackageJson
 workspacesValidator =
   V.inField ("workspaces", P.workspaces) $
     maybe noWorkspacesDefinedError $
-      V.all [requiredWorkspaceValidator]
+      V.all
```
infomiho, line 25: Do we allow any other workspace value e.g. user includes some workspace of their own? If we want to disallow that, let's maybe collapse the required and forbidden into "the array must have this value".

I think it's easier to reason about what the expected value is.

I know with "included" to keep the sort unimportant - we can maybe sort the input before validating it. 

RAW

waspc/src/Wasp/Generator/SdkGenerator/Common.hs
```diff
@@ -48,14 +47,6 @@ mkTmplFdWithData relSrcPath tmplData = mkTmplFdWithDstAndData relSrcPath relDstP
 mkTmplFd :: Path' (Rel SdkTemplatesDir) File' -> FileDraft
 mkTmplFd path = mkTmplFdWithDst path (SP.castRel path)
 
--- To understand what's going on here, read this issue:
--- https://github.com/wasp-lang/wasp/issues/1769
```
infomiho, line 52: Noice

waspc/cli/src/Wasp/Cli/Command/Require.hs
```diff
@@ -114,19 +108,43 @@ instance Requirable (DbConnectionEstablished FromOutDir) where
...
+  checkRequirement =
+    checkGeneratedCode BuildType.Production >>= \case
+      True -> return GeneratedCodeIsProduction
+      False ->
+        throwError $
+          CommandError
+            "Built app does not exist"
+            "You can build the app with the `wasp build` command."
+
+checkGeneratedCode :: BuildType.BuildType -> Command Bool
```
infomiho:
```suggestion
isExpectedCodeGenerated:: BuildType.BuildType -> Command Bool
```

That's a mouthful, but I think the existing name is ambiguous. Let's find a better name for this function.

Something also feels off with `Bool` here. The result is a negation of "is clean build needed". Somehow it would be nice if can get a result out e.g. `BuildValid | UnexpectedBuildType` which we can use here and match against it in the case expression.

waspc/src/Wasp/Generator/Valid/PackageJson/Workspaces.hs
cprecioso: Given that we already have this, I don't see a strong reason to go out of our way to remove it.

waspc/src/Wasp/Generator/Valid/PackageJson/Workspaces.hs
cprecioso: Same answer as previously

waspc/cli/src/Wasp/Cli/Command/Require.hs
cprecioso: > Somehow it would be nice if can get a result out e.g. BuildValid | UnexpectedBuildType which we can use here and match against it in the case expression.

I don't really agree, it doesn't seem like we'd do anything new with the new values out of this enum, just use it as a `True/False`.

Renamed it to `compareBuildTypeToGeneratedCode`

cprecioso:
@infomiho i replied

wasp-app-runner/src/versions.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export const getSettingsForVersion = (
+  version: WaspVersion,
+): VersionSettings => {
+  if (semver.lt(version, "0.21.0")) {
+    return {
+      clientAppDir: ".wasp/build/web-app",
+      serverAppDir: ".wasp/build",
+    };
```
infomiho, line 16: Looking at this for the second time, will we ever need this? `wasp-app-runner` is versioned and if we _really_ needed to test something with an older Wasp version, we use the older version. We pinned the version in the Open SaaS template as well: https://github.com/wasp-lang/open-saas/blob/main/template/e2e-tests/package.json#L22C30-L22C32?

waspc/cli/src/Wasp/Cli/Command/Build.hs
```diff
@@ -86,14 +78,14 @@ build = do
     Right () -> return ()
 
   cliSendMessageC $
-    Msg.Success "Your wasp project has been successfully built! Check it out in the .wasp/build directory."
+    Msg.Success "Your wasp project has been successfully built! Check it out in the .wasp/out directory."
```
infomiho:
```suggestion
    Msg.Success "Your wasp project has been successfully built! Check it out in the " ++ fromRelDir relBuildDir ++ " directory."
```

waspc/data/Cli/starters/basic/package.json
```diff
@@ -1,9 +1,7 @@
 {
   "name": "__waspAppName__",
   "type": "module",
-  "workspaces": [
-    ".wasp/build/*",
-    ".wasp/out/*"
+  "workspaces": [".wasp/out/*"
   ],
```
infomiho:
```suggestion
  "workspaces": [".wasp/out/*"],
```

web/docs/migration-guides/migrate-from-0-20-to-0-21.md
```diff
@@ -42,7 +46,46 @@ app MyApp {
...
+
+Now, we will clean up the old `.wasp/build` directory (to avoid any potential conflicts), and let `npm` pick up the new workspace configuration. You can do this by running the following command in your terminal:
+
+```sh
+wasp clean
+wasp ts-setup # if you're using the TS Spec
+wasp compile
+```
+
+### 23 Upgrade Tailwind CSS to v4
```
infomiho:
```suggestion
### 3. Upgrade Tailwind CSS to v4
```

waspc/src/Wasp/AI/GenerateNewProject/InitialFiles.hs
```diff
@@ -151,7 +150,6 @@ generatePackageJson newProjectDetails =
         "name": "${appName}",
         "type": "module",
         "workspaces": [
```
infomiho, line 152: Nitpick: adjust formatting to match the skeleton files.

waspc/src/Wasp/Generator/Valid/PackageJson/Workspaces.hs
infomiho: > Do we allow any other workspace value e.g. user includes some workspace of their own? 

What about this bit?

waspc/src/Wasp/Generator/Valid/PackageJson/Workspaces.hs
infomiho: I don't think that's a great argument for not simplifying our code - if we see an opportunity to clean up, we should take it. It's a sunken cost since we invented new ways to validate "required" and "forbidden" items, but I still think it's cleaner if we just required the `workspaces` array to be of certain value. (If we disallow any other workspace values, I had that question in the other thread)

I'd refactor this to be simpler if we don't allow any other workspace value. I'd make the validation use a `Set` so the order doesn't matter once we add the SDK workspace.

waspc/cli/src/Wasp/Cli/Command/Require.hs
infomiho: On a second thought, I think I don't have a problem with this bit necessarily, but the bit before it i.e. `checkIfCleanBuildNeeded`. I think that name is causing me to question the naming/approach. 

If we named `checkIfCleanBuildNeeded` as `isExistingBuildValid` that would read much nicer. It would be:
```hs
return $ isExistingBuildValid
```
instead of 
```hs
return $ not cleanBuildNeeded
```

This way we are not saying something is "not valid" but we are saying something is "valid" so it's much easier to think about it.

waspc/src/Wasp/Generator/Valid/PackageJson/Workspaces.hs
cprecioso: It is currently allowed. We decided to allow it in the PR that initially implemented this functionality. It shouldn't affect us that there are more workspaces (but I haven't tested it).

wasp-app-runner/src/versions.ts
cprecioso: I like having `wasp-app-runner` as a single tool for us to deal with any app we might need -- this also includes reproductions for users on older versions. I would prefer not to juggle multiple versions of WAR if I need to test older projects -- especially since the WAR versions don't directly correlate with Wasp versions.

waspc/src/Wasp/Generator/Valid/PackageJson/Workspaces.hs
cprecioso: > I don't think that's a great argument for not simplifying our code

Sorry, I wasn't referring to the code itself, but for the fact that we would be arbitrarily constraining our users to not allow extra workspaces.

waspc/cli/src/Wasp/Cli/Command/Require.hs
cprecioso: Okay, check it out now

waspc/src/Wasp/AI/GenerateNewProject/InitialFiles.hs
cprecioso: Everything is autoformatting to `[\n".wasp/out/*"\n]`, so I kept that.

cprecioso:
@infomiho i replied

waspc/src/Wasp/Generator/Valid/PackageJson/Workspaces.hs
infomiho: Can we test it and know it if it's something we want to disallow? I'd like to know if there is a user-error waiting to happen which we can prevent :)

wasp-app-runner/src/versions.ts
infomiho: Ok, but at which point do we remove this code? At Wasp 1.0, Wasp 2.0? Do we add all the cases for all the different breaking changes and keep maintaining the old API compatibility until we decide to clean up? Why not clean up now and just use and older version for stuff we might need. We use this tool mostly with the latest version of Wasp like 99% of the time anyways.

waspc/src/Wasp/Generator/Valid/PackageJson/Workspaces.hs
infomiho: Let's continue in the other thread, I'll resolve this one.

waspc/cli/src/Wasp/Cli/Command/Compile.hs
```diff
@@ -55,7 +55,8 @@ compileWithOptions options = do
           </> generatedCodeDirInDotWaspDir
 
   isCleanBuildNeeded <-
-    liftIO $ WaspInfo.checkIfCleanBuildNeeded outDir $ buildType options
+    liftIO $
+      buildType options `WaspInfo.isCompatibleWithExistingBuildAt` outDir
```
infomiho: Do we maybe have an error here? 

Shouldn't it be: `isCleanBuildNeeded` == **not** `isCompatibleWithExistingBuildAt`?

waspc/cli/src/Wasp/Cli/Command/Require.hs
```diff
@@ -114,19 +108,38 @@ instance Requirable (DbConnectionEstablished FromOutDir) where
...
+  checkRequirement =
+    isBuildTypeCompatibleWithGeneratedCode BuildType.Production >>= \case
+      True -> return GeneratedCodeIsProduction
+      False ->
+        throwError $
+          CommandError
+            "Built app does not exist"
+            "You can build the app with the `wasp build` command."
+
+isBuildTypeCompatibleWithGeneratedCode :: BuildType.BuildType -> Command Bool
```
infomiho, line 135: I like this much better, easier to reason about 👍

waspc/cli/src/Wasp/Cli/Command/Compile.hs
cprecioso: You're right! Fixed. I also renamed some variables here to make it easier to follow.

wasp-app-runner/src/versions.ts
cprecioso: > Ok, but at which point do we remove this code? 

I don't have that answer 😔 but IMO it's probably in the following months, if it becomes a hassle, but not yet.

I think it's going to be more comfortable to have WAR support both things at least until our daily work has firmly transitioned. Especially for such a relative low "price".

> Do we add all the cases for all the different breaking changes and keep maintaining the old API compatibility until we decide to clean up? 

I don't think supporting both versions in the transition period needs to translate to having to suffer through maintenance of this forever.

This is a tool to help us. I think it helps us more than not to support both styles for now. As soon as it stops being helpful, we can kill that feature.

Moreover, if we move it to use `wasp build start` in the near future, this will become moot because each wasp version will know what to do.

waspc/src/Wasp/Generator/Valid/PackageJson/Workspaces.hs
cprecioso: So, I tested it. Works fine, even when alternative versions of dependencies are defined in other workspaces.

Maaaybe something like a transitive dependency gets hoisted somewhere and that messes up the resolution structure in some unknown way? idk, but that's an edge case of an edge case

So anyway I leave it up to you since it you're evidently quite invested in it 😁

cprecioso:
@infomiho fixed and replied

wasp-app-runner/src/versions.ts
infomiho: Using `wasp build start` sounds like a way to go down the line. Okay, you convinced me, let's keep it like until the foreseeable future :)

waspc/src/Wasp/Generator/Valid/PackageJson/Workspaces.hs
infomiho: > So anyway I leave it up to you

What do you mean? I wanted to see if allowing extra workspace would break Wasp and we want to defend against that. If we allow extra workspaces, let's keep the current validation, if not, let's change the validation.

Based on what you told me, I think it's okay if we allow it.

waspc/cli/src/Wasp/Cli/Command/Require.hs
```diff
@@ -82,29 +81,24 @@ instance Requirable InWaspProject where
...
 
-data DbConnectionEstablished fromDir = DbConnectionEstablished fromDir deriving (Typeable)
-
-instance Requirable (DbConnectionEstablished FromOutDir) where
+instance Requirable DbConnectionEstablished where
   checkRequirement = do
     -- NOTE: 'InWaspProject' does not depend on this requirement, so this
     -- call to 'require' will not result in an infinite loop.
     InWaspProject waspProjectDir <- require
+    GeneratedCodeIsDevelopment <- require
```
sodic, line 91: Something seems strange here. Why would database connection rely on the generated code being in development?

waspc/data/packages/deploy/src/common/waspProject.ts
```diff
@@ -50,11 +50,11 @@ export function getClientBuildArtefactsDir(
 }
 
 function getWaspBuildDir(waspProjectDir: WaspProjectDir): string {
```
sodic, line 52: Should we change the names in functions and variables too now that we no longer call it `build`?

waspc/cli/src/Wasp/Cli/Command/Build.hs
```diff
@@ -52,25 +51,18 @@ build :: Command ()
 build = do
   InWaspProject waspProjectDir <- require
 
-  let buildDir =
-        waspProjectDir
-          </> dotWaspDirInWaspProjectDir
-          </> buildDirInDotWaspDir
+  let relBuildDir = dotWaspDirInWaspProjectDir </> generatedCodeDirInDotWaspDir
+      buildDir = waspProjectDir </> relBuildDir
```
sodic, line 55: We have a lot of duplication with `buildDir` (constructing it in multiple commands), but I don't know what we could do about it.

The path could become a part of the `InWaspProject` requirement, but it doesn't seem like that's its place - being inside the wasp project necessarily means you get `WaspProjectDir` but it doesn't necessarily mean you want to output something to the `buildDir`.

web/docs/deployment/deployment-methods/paas.md
```diff
@@ -65,9 +65,9 @@ While these are the general instructions on deploying the server anywhere, we al
 
 <BuildingTheWebClient />
 
-The command above will build the web client and put it in the `build/` directory in the `.wasp/build/web-app/`.
+The command above will build the web client and put it in the `build/` directory in the `.wasp/out/web-app/`.
```
sodic, line 68: Just grammar: We're missing "directory" at the end.

waspc/cli/src/Wasp/Cli/Command/Require.hs
cprecioso: I don't remember exactly but it stems from when I first implemented this Requirement. Some `.env` files are copied in `dev` mode that are not copied in `build` mode, and those include the DB connection string.

cprecioso: There was this comment with info about it just aboce but I removed it 🤦‍♂️, will adapt it and add it back.

Image

waspc/data/packages/deploy/src/common/waspProject.ts
cprecioso: This is still the folder where the wasp app gets built to, no?

waspc/cli/src/Wasp/Cli/Command/Build.hs
cprecioso: I don't think I have a solution for it :/
I could make it so that it's returned in `GeneratedCodeIsDevelopment/Production`, but that won't apply to this specific instance.

cprecioso:
@sodic addressed your feedback here: https://github.com/wasp-lang/wasp/pull/3640

waspc/data/packages/deploy/src/common/waspProject.ts
infomiho: I've changed this to deployment dir in my Vite PRs.

PR #3542, opened 2025-12-22
Refactor Validator

cprecioso:
## Description

- Part of #3543

The `PackageJson` validator was becoming quite large, so we split it.
With the better organization, I could move the lists of properties we're checking, to better showcase them at the top of the file.

I also removed the specific case check for when `workspaces` was not set, or a dependency is not in the list at all. Now they just fall through to the regular checks, which give the same information, removing our duplication.

### Motivation

In forthcoming PRs I will be adding more kinds of checks for [workspaces](https://github.com/wasp-lang/wasp/pull/3543) and [dependencies](https://github.com/wasp-lang/wasp/pull/3544), so I needed to reorganize this file a bit or it would become too large.

## All the PRs

- #3544
- #3543
- #3542

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/src/Wasp/Generator/Valid/PackageJson/Workspaces.hs
+++ b/waspc/src/Wasp/Generator/Valid/PackageJson/Workspaces.hs
@@ -27,9 +27,9 @@ workspacesValidator =
 
     noWorkspacesDefinedError =
       V.failure $
-        "Wasp requires the field to be an array, and include the following values: "
+        "Wasp requires \"workspaces\" to have the value: ["
           ++ intercalate ", " (show <$> requiredWorkspaces)
-          ++ "."
+          ++ "]."
 
 makeWorkspaceIncludedValidator :: WorkspaceName -> V.Validator [WorkspaceName]
 makeWorkspaceIncludedValidator expectedWorkspace =


waspc/src/Wasp/Generator/Valid/PackageJson/Workspaces.hs
```diff
@@ -0,0 +1,40 @@
...
+      V.all $ makeWorkspaceIncludedValidator <$> requiredWorkspaces
+
+makeWorkspaceIncludedValidator :: WorkspaceName -> V.Validator [WorkspaceName]
+makeWorkspaceIncludedValidator expectedWorkspace =
+  bool missingWorkspaceError V.success
+    . elem expectedWorkspace
+  where
+    missingWorkspaceError =
+      V.failure $
+        "Wasp requires "
```
infomiho:
```suggestion
        "Wasp requires workspace "
```

Nitpick

waspc/src/Wasp/Generator/Valid/PackageJson/Dependencies.hs
```diff
@@ -0,0 +1,154 @@
...
+
+    correctVersionValidator :: V.Validator (Maybe P.PackageVersion)
+    correctVersionValidator (Just actualVersion)
+      | actualVersion == expectedPkgVersion = V.success
+    correctVersionValidator _ = incorrectPackageVersionError
+
+    inCorrectDepList :: V.Validator (Maybe P.PackageVersion) -> V.Validator P.PackageJson
+    inCorrectDepList = inDependency depType pkgName
+    inOppositeDepList :: V.Validator (Maybe P.PackageVersion) -> V.Validator P.PackageJson
+    inOppositeDepList = inDependency (oppositeForDepType depType) pkgName
```
infomiho, line 101:
```suggestion
    inCorrectDepList :: V.Validator (Maybe P.PackageVersion) -> V.Validator P.PackageJson
    inCorrectDepList = inDependency depType 
    
    inOppositeDepList :: V.Validator (Maybe P.PackageVersion) -> V.Validator P.PackageJson
    inOppositeDepList = inDependency (oppositeForDepType depType) pkgName
```

RAW

waspc/src/Wasp/Generator/Valid/PackageJson/Dependencies.hs
```diff
@@ -0,0 +1,154 @@
...
+          ++ " if present."
+
+-- | Runs the validator on a specific dependency of the given PackageJson
+-- record, setting the appropriate path for errors.
+inDependency ::
+  DependencyType ->
+  P.PackageName ->
+  V.Validator (Maybe P.PackageVersion) ->
+  V.Validator P.PackageJson
+inDependency depType pkgName versionStringValidator =
```
infomiho:
```suggestion
inDependency depType pkgName versionValidator =
```

We didn't use this `string` prefix elsewhere, I'd either drop it here or use it everywhere.

waspc/src/Wasp/Generator/Valid/PackageJson.hs
```diff
@@ -45,164 +21,3 @@ validatePackageJson pkgJson =
...
             dependenciesValidator
           ]
-
-type WorkspaceName = String
-
-workspacesValidator :: V.Validator P.PackageJson
-workspacesValidator =
-  V.inField ("workspaces", P.workspaces) $ \case
-    Just actualWorkspaces -> requiredWorkspacesIncludedValidator actualWorkspaces
-    Nothing -> workspacesNotDefinedError
```
infomiho, line 55: I think we should still have this error as it's useful for people who don't really understand workspaces.

With this new version you get something like this;
```

- In "package.json" → "workspaces":
    Wasp requires ".wasp/build/*" to be included.

- In "package.json" → "workspaces":
    Wasp requires ".wasp/out/*" to be included.
```

If I never even dealt with `workspaces` field - it might be confusing for me what is my course of action. 

I think the old check was useful so we can say "Wasp required that you have the `worksapces` field in your `package.json` with the following value."

cprecioso:
@infomiho ready for re-review

waspc/src/Wasp/Generator/Valid/PackageJson/Workspaces.hs
```diff
@@ -0,0 +1,43 @@
...
+workspacesValidator =
+  V.inField ("workspaces", P.workspaces) $
+    maybe noWorkspacesDefinedError $
+      V.all [requiredWorkspaceValidator]
+  where
+    requiredWorkspaceValidator :: V.Validator [WorkspaceName]
+    requiredWorkspaceValidator =
+      V.all $ makeWorkspaceIncludedValidator <$> requiredWorkspaces
+
+    noWorkspacesDefinedError =
```
infomiho, line 28: This still feels a bit cold and neutral, but we have enough the context to be specific.

For the current error:
```
- In "package.json" → "workspaces":
    Wasp requires the field to be an array, and include the following values: ".wasp/build/*", ".wasp/out/*".
```
It tells me it needs to be an array and I need to add these two values. I feels like:
```js
const a = []
a.push(".wasp/build/*")
a.push(".wasp/out/*")
```

I'd maybe go for a more "custom" explanation:
```
- In "package.json" → "workspaces":
    Wasp requires the "workspaces" field to have the value: [".wasp/build/*", ".wasp/out/*"]
```
which feels more like:
```js
const a = [".wasp/build/*", ".wasp/out/*"]
```

I like it because I can copy paste the value.

cprecioso: Okay, tweaked it to:

```
- In "package.json" → "workspaces":
    Wasp requires "workspaces" to have the value: [".wasp/build/*", ".wasp/out/*"].
```

cprecioso:
@infomiho ready for re-review

infomiho (approved):

PR #3544, opened 2025-12-22
SDK as workspace

### LINKED ISSUE #3233: Add Wasp SDK to npm Workspaces

cprecioso:
- Resolves #3233
- On top of #3540

## Description

Not yet added (waiting for go-ahead): snapshots (already made), changelog, migration guide

- Adds the SDK as a workspace to the users' package.json.
- Prints a warning if it is in their dependencies.

### Motivation

Context: #3233

By adding the SDK as a workspace, it participates as a top-level package in the dependency resolution avoiding hypothetical issues where:
1. npm could assign the SDK its own, different sub-dependencies in order to have a correct resolution
2. the behaviour of `file:` dependencies changes and dependencies are no longer implicitly installed

We also now forbid `wasp` in the `dependencies`, since the workspaces will just be linked together so they can be called through that mechanism instead of duplicating the SDK in deps.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:


examples/ask-the-documents/package.json
```diff
@@ -23,8 +24,7 @@
     "react-hook-form": "^7.45.4",
     "react-markdown": "^10.1.0",
     "react-router-dom": "^6.26.2",
-    "tailwindcss": "^3.2.7",
-    "wasp": "file:.wasp/out/sdk/wasp"
```
infomiho, line 27: I'm surprised that this works like this? I know in `pnpm` you have to do `workspace:*` and stuff like that to get to use the package. I'd just like some of your npm foo here :)

waspc/src/Wasp/Generator/NpmWorkspaces.hs
```diff
@@ -24,16 +24,19 @@ import Wasp.Project.Common
...
-    [ makeGlobFromProjectRoot $ dotWaspDirInWaspProjectDir </> generatedCodeDirInDotWaspDir
-    -- TODO: Add SDK as a workspace (#3233)
+    [ allProjectsInDir $ dotWaspDirInWaspProjectDir </> generatedCodeDirInDotWaspDir,
+      relDirToGlob $ dotWaspDirInWaspProjectDir </> generatedCodeDirInDotWaspDir </> sdkRootDirInGeneratedCodeDir
     ]
   where
-    makeGlobFromProjectRoot :: Path' (Rel WaspProjectDir) (Dir ProjectRootDir) -> String
-    makeGlobFromProjectRoot projectRootDir =
-      relDirToPosixString projectRootDir FP.</> "*"
+    allProjectsInDir :: Path' (Rel WaspProjectDir) (Dir a) -> String
```
infomiho: I feel like the old name was more clear. `makeGlobFromProjectRoot` vs. `allProjectsInDir`

I think you are [creating package list](https://docs.npmjs.com/cli/v8/using-npm/workspaces) for the workspaces property. 

So I'd maybe go with, `makePackage` and `makePackagesGlobForDir` or maybe even `makeWorkspacePacakge` and `makeWorkspacePackagesGlobForDir`.

Or if we don't want to use package name here, `makeWorkspace` and `makeWorkspacesGlobForDir`.

Since we are now combining paths and globs, we probably also need to rename `requiredWorkspaceGlobs` to `requiredWorkspaces`.

infomiho (commented):
Nice job! I think we can improve the naming a bit, left a comment

examples/ask-the-documents/package.json
cprecioso: I know, in yarn you have to do that too. Apparently npm does not need/want it, i checked.

waspc/src/Wasp/Generator/NpmWorkspaces.hs
cprecioso: Renaming makes sense, I did `makeGlobForAllSubdirs` and `makeGlobForDir`.

> Since we are now combining paths and globs, we probably also need to rename requiredWorkspaceGlobs to requiredWorkspaces.

I don't agree with this. The string is still a glob, it just so happens that it doesn't have a wildcard character; but it still gets interpreted as a glob.

PR #3551, opened 2025-12-23
[Wasp News] Fix some edge cases

sodic:


waspc/cli/src/Wasp/Cli/Command/News/Report.hs
```diff
@@ -80,10 +76,9 @@ isTimeForMandatoryNewsReport state = case state.lastReportAt of
 
 printNewsReportAndUpdateLocalState :: LocalNewsState -> NewsReport -> IO ()
 printNewsReportAndUpdateLocalState localNewsStateBeforeReport newsReport = do
-  reportNews
-  if newsReport.requireConfirmation
-    then askForConfirmation
-    else putStrLn $ "\nRun " ++ styleCode "wasp news" ++ " if you don't want to see these news again."
+  unless (null newsReport.newsToShow) $ do
```
sodic, line 79: Avoids an empty line when there are no news.

waspc/cli/src/Wasp/Cli/Command/News/Report.hs
```diff
@@ -62,10 +61,7 @@ makeMandatoryNewsReportForExistingUser currentState newsEntries =
   NewsReport
     { newsToShow = allRelevantUnseenNews,
       requireConfirmation,
-      newsToConsiderSeen =
-        if requireConfirmation
-          then allRelevantUnseenNews
-          else []
+      newsToConsiderSeen = allRelevantUnseenNews
```
sodic, line 64: I had to give up on this condition for the time being because the old code would show "Run `wasp news` to stop seeing these news" even if:
  - There are no news in the report.
  - The user runs `wasp news` explicitly.

Fixing that properly would take to long so I've decided to termporarily treat all important news as seen after showing them (i.e., they won't repeat after every 24 hours).

infomiho: Let's create an issue to fix this properly 👍

sodic: Already in the notes. I'll continue working on news after the release.

PR #3554, opened 2025-12-29
Refactor and fix Wasp news

sodic:
## Description

Addressing the fixes from:
- #3534
- #3535 
- #3536 
- #3537

Check those PRs for suggestions and their status. Most of the checklist was done in the main feature PR that includes those 4, here: #3533 

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/cli/src/Wasp/Cli/Interactive.hs
+++ b/waspc/cli/src/Wasp/Cli/Interactive.hs
@@ -15,6 +15,7 @@ where
 import Control.Applicative ((<|>))
 import Data.Foldable (find)
 import Data.Function ((&))
+import Data.Functor ((<&>))
 import Data.List (intercalate)
 import Data.List.NonEmpty (NonEmpty ((:|)))
 import qualified Data.List.NonEmpty as NE
@@ -136,13 +137,13 @@ tryGettingConfirmationWithTimeout message requiredAnswer timeoutSeconds = do
   isInteractive <- hIsTerminalDevice stdin
   if not isInteractive
     then return $ Left NonInteractiveShell
-    else do
-      maybeAnswer <- timeout timeoutMicroseconds (askForInput message)
-      return $ case maybeAnswer of
-        Nothing -> Left Timeout
-        Just actualAnswer
-          | actualAnswer == requiredAnswer -> Right ()
-          | otherwise -> Left $ WrongAnswer actualAnswer
+    else
+      timeout timeoutMicroseconds (askForInput message)
+        <&> \case
+          Nothing -> Left Timeout
+          Just actualAnswer
+            | actualAnswer == requiredAnswer -> Right ()
+            | otherwise -> Left $ WrongAnswer actualAnswer
   where
     timeoutMicroseconds = timeoutSeconds * 10 ^ (6 :: Int)
 


waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -128,6 +131,13 @@ askToChoose question options = do
 askForInput :: String -> IO String
 askForInput question = putStr (Term.applyStyles [Term.Bold] question) >> prompt
 
+askForConfirmationWithTimeout :: String -> String -> Int -> IO Bool
+askForConfirmationWithTimeout message requiredAnswer timeoutSeconds = do
+  maybeAnswer <- timeout timeoutMicroseconds $ askForInput message
+  return $ fromMaybe "" maybeAnswer == requiredAnswer
```
infomiho:
```suggestion
  return $ maybeAnswer == Just requiredAnswer
```

waspc/cli/tests/Wasp/Cli/Command/News/ActionTest.hs
```diff
@@ -0,0 +1,100 @@
...
+        forExistingUser $ \localNewsState newsEntries ->
+          let news = getNewsFromAction $ makeWaspInvokedNewsAction localNewsState newsEntries
+           in all ((>= Important) . level) news
+
+      prop "does not include news that were previously seen" $
+        forExistingUser $ \localNewsState newsEntries ->
+          let news = getNewsFromAction $ makeWaspInvokedNewsAction localNewsState newsEntries
+           in not $ any (wasNewsEntrySeen localNewsState) news
+
+      prop "creates ShowWithConfirmation iff at least one critical news entry exists" $
```
infomiho:
```suggestion
      prop "creates ShowWithConfirmation if at least one critical news entry exists" $
```

waspc/cli/tests/Wasp/Cli/Command/News/ActionTest.hs
```diff
@@ -0,0 +1,100 @@
...
+    someTime = T.UTCTime (T.fromGregorian 2024 1 1) 0
+    forExistingUser prop' localNewsState newsEntries =
+      not (localNewsState == emptyLocalNewsState) ==> prop' localNewsState newsEntries
+
+getNewsFromAction :: NewsAction -> [NewsEntry]
+getNewsFromAction = \case
+  ShowAllAndMarkSeen news -> news
+  MarkSeenWithoutShowing news -> news
+  ShowWithConfirmation news -> news
+  ShowWithoutMarkingSeen news -> news
```
infomiho:
```suggestion
  ShowAllAndMarkSeen news -> news
  ShowNoneAndMarkSeen news -> news
  ShowImportantAndWait news -> news
  ShowWithoutMarkingSeen news -> news
```

Played around with the names, I'm not saying they are better, they might be worse 😂

waspc/cli/src/Wasp/Cli/Command/News/Action.hs
```diff
@@ -0,0 +1,86 @@
...
+    hasCriticalNews = any ((== Critical) . level) relevantUnseenNews
+    relevantUnseenNews = filter isRelevant . filter isUnseen $ allNewsEntries
+    isRelevant = (>= Important) . level
+    isUnseen = not . wasNewsEntrySeen currentState
+
+executeNewsAction :: LocalNewsState -> NewsAction -> IO ()
+executeNewsAction localState = \case
+  ShowAllAndMarkSeen news -> do
+    printNews news
+    updateStateWithNews news
```
infomiho: Nipick: `ShowAllAndMarkSeen` gives me print + update state. Maybe the `updateStateWithNews` should be named `markNewsAsSeen` to be super clear?

waspc/cli/src/Wasp/Cli/Command/News/Action.hs
```diff
@@ -0,0 +1,86 @@
...
+  MarkSeenWithoutShowing news ->
+    updateStateWithNews news
+  ShowWithConfirmation news -> do
+    printNews news
+    ifM askUserForConfirmation (updateStateWithNews news) updateStateOnly
```
infomiho:
```suggestion
    ifM askUserForConfirmation (markNewsAsSeen news) updateLastShowTimestamp
```

infomiho:
I know it's a draft, but I left some thoughts, feel free to ignore them if they are missing the mark.

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
```diff
@@ -26,7 +26,6 @@ fetchNews = do
   requestResult <- try $ httpJSONThatThrowsIfNot2xx =<< parseRequest waspNewsUrl
 
   return $ case requestResult of
-    -- TODO: Can I get the status code easily
     Left (_ :: HttpException) -> Left "Failed to fetch news from server."
-    Right (Left jsonError) -> Left $ "Failed to decode news JSON: " ++ jsonError
+    Right (Left _) -> Left "Failed to decode news JSON. This is a temporary problem with the Wasp News server"
```
sodic: The original message was too low level - not useful for the user and looked broken.

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
```diff
@@ -0,0 +1,123 @@
...
+getNewsToMarkAsSeen :: LocalNewsState -> NewsListing -> [NewsEntry]
+getNewsToMarkAsSeen localState listing = case listing of
+  UserRequestedAllNews allNews -> allNews
+  WaspRequestedMustSeeNews allNews
+    | not $ doesUserHaveNewsHistory localState -> allNews
+    | isConfirmationRequired localState listing -> getNewsToShow localState listing
+    | otherwise -> []
+
+processNewsListing :: LocalNewsState -> NewsListing -> IO ()
+processNewsListing localState listing = do
```
sodic: The new behavior is:
  - When Wasp shows critical news during `wasp start`, wait for confirmation but allow an answer other than "y" and time out after 10 seconds. If the confirmation times out or isn't "y," the news are shown again in a day.
  - When Wasp shows important news during wasp start, tell the users they have to run `wasp news` to mark them as read. @infomiho I didn't do the tiemout because it felt to similar to how we behave when there are critical news.
  - When the user runs `wasp news` themselves, they get all news and no other confirmation/messages

waspc/cli/tests/Wasp/Cli/Command/News/ListingTest.hs
```diff
@@ -0,0 +1,102 @@
...
+      it "always returns False" $ do
+        let newsState = markNewsAsSeen allNews emptyLocalNewsState
+        isConfirmationRequired newsState (UserRequestedAllNews allNews) `shouldBe` False
+
+    -- TODO: bad test, too similar to implementation
+    describe "for WaspRequestedMustSeeNews" $ do
+      prop "returns True if and only if there is previous news history and newsToShow includes critical news" $
+        forAll scenario $ \(newsState, listing) ->
+          isConfirmationRequired newsState listing
+            === (newsState /= emptyLocalNewsState && any ((== Critical) . level) (getNewsToShow newsState listing))
```
sodic: @Martinsos What do you think about this test?

waspc/cli/tests/Wasp/Cli/Command/News/ActionTest.hs
sodic: We changed this up in the end.

waspc/cli/src/Wasp/Cli/Command/News/Action.hs
sodic: Changed this up but did this.

waspc/cli/src/Wasp/Cli/Command/News/Fetching.hs
Martinsos:
```suggestion
    Right (Left _) -> Left "Failed to decode news JSON. This is a temporary problem with the Wasp News server."
```

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
```diff
@@ -0,0 +1,123 @@
+{-# LANGUAGE LambdaCase #-}
+
+module Wasp.Cli.Command.News.Listing
+  ( -- * News Listing
+
+    -- This modules captures all decisions Wasp has to make when showing Wasp
+    -- news under different contexts (i.e., who requested the news listing):
+    --   - Which news to show under which context.
+    --   - Which news to mark as seen.
```
Martinsos, line 10:
```suggestion
    --   - Which news to mark as seen after they have been shown.
```

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
```diff
@@ -0,0 +1,123 @@
...
+    --   - Whether the user should interactively confirm they've read the news.
```
Martinsos: let's keep the nomenclature consistent:
```suggestion
    --   - Whether the user should interactively confirm they have seen the news.
```

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
```diff
@@ -0,0 +1,123 @@
...
+    --   - etc.
+    --
+    -- We tried to capture most of the logic in pure function to simplify
```
Martinsos: Hah I don't mind this comment, but we are in pure functional language, obviously we will try to do this always when we can, for multiple reasons, only one of them being easier testing :). So this comment is redundant.

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
```diff
@@ -0,0 +1,123 @@
...
+    saveLocalNewsState,
+    setLastListingTimestamp,
+    wasLastLisingMoreThanNHoursAgo,
+    wasNewsEntrySeen,
+  )
+import Wasp.Cli.Interactive (askForConfirmationWithTimeout)
+import Wasp.Util.Terminal (styleCode)
+
+shouldWaspListMustSeeNews :: LocalNewsState -> IO Bool
+shouldWaspListMustSeeNews = wasLastLisingMoreThanNHoursAgo 24
```
Martinsos:
```suggestion
shouldWaspListMustSeeNews = wasLastListingMoreThanNHoursAgo 24
```

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
```diff
@@ -0,0 +1,123 @@
...
+
+data NewsListing
+  = UserRequestedAllNews [NewsEntry]
+  | WaspRequestedMustSeeNews [NewsEntry]
```
Martinsos: Don't do this if it is too big of a change now, but I had a bit of an epiphany and think this would be even nicer name, look at it:

```suggestion
  = UserListingAllNews [NewsEntry]
  | WaspListingMustSeeNews [NewsEntry]
```

Martinsos: Also, one thing that I feel is missing here, is the information that these `[NewsEntry]` are all the news ever. That is not written anywhere.

One way to go about it is to just add comment here, well, two commens, for each data constructor.

Another might be to make these named fields: `allNews: [NewsEntry]`. But then that is a bit of an antipattern. Maybe not if we have two records that are exactly the same, maybe that is fine, yeah. I would probably do this, yeah:
```hs
data NewsListing
  = UserRequestedAllNews { allNews :: [NewsEntry] }
  | WaspRequestedMustSeeNews { allNews :: [NewsEntry] }
```
if this compiles hm, I think it should, even with new records stuff.

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
```diff
@@ -0,0 +1,123 @@
...
+  deriving (Show, Eq)
+
+getNewsToShow :: LocalNewsState -> NewsListing -> [NewsEntry]
+getNewsToShow localState = \case
+  UserRequestedAllNews allNews -> allNews
+  WaspRequestedMustSeeNews allNews
+    | doesUserHaveNewsHistory localState -> unseenNewsThatUserMustSee allNews
+    | otherwise -> []
```
Martinsos, line 55: Mayb ea comment here explaining the decision for the `otherwise` branch would help. It kidn of is obvious one you think about it, but it does require some thinking, and it would help make it more obvious so they don't have to guess and well do the thinking from scratch.

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
```diff
@@ -0,0 +1,123 @@
...
+    isUnseen = not . wasNewsEntrySeen localState
+
+isConfirmationRequired :: LocalNewsState -> NewsListing -> Bool
+isConfirmationRequired localState listing = case listing of
+  UserRequestedAllNews _ -> False
+  WaspRequestedMustSeeNews _ -> any ((== Critical) . level) $ getNewsToShow localState listing
+
+getNewsToMarkAsSeen :: LocalNewsState -> NewsListing -> [NewsEntry]
+getNewsToMarkAsSeen localState listing = case listing of
+  UserRequestedAllNews allNews -> allNews
```
Martinsos: Ok you are going a bit nuclear here, you are marking them all as to be marked as seen, while that is not really true, you really want to mark only the new news.

This works for now. But what if the caller expects this to be more precise? I guess it depends what is expected from this function. I would say this is somehat unexpected behaviour. So maybe, if not too much effort, I would make this more precise. Then for example it might be easier to, when all news are listed to user, mark the "new" ones -> they will be the ones returned by this function. While right now you don't have that info, if I am correct.

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
```diff
@@ -0,0 +1,123 @@
...
+  WaspRequestedMustSeeNews allNews
+    | not $ doesUserHaveNewsHistory localState -> allNews
+    | isConfirmationRequired localState listing -> getNewsToShow localState listing
+    | otherwise -> []
```
Martinsos, line 74: Hm yeah this part is a bit out of context here, I think this is what you were talking about.

Just reading this, it is a bit hard to understand why this logic is like this.
So if confirmation is required, then all news to show will be marked as seen. I guess it make sense.
However otherwise, and I guess that is the case when confirmation is NOT required (can it be anything else?), no news are to be marked as seen. Why is that so?

I would say, either we provide some coments here, or, what you could do, you could unite two functions in one, so that logic is shared. For example, `getNewsToShow` could merge with `getNewsToMarkAsSeen` and return type would become richer for it. Of course, risking that one fucntion is doing too much, but on the other hand, providing unified context in a way.

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
```diff
@@ -0,0 +1,123 @@
...
+
+processNewsListing :: LocalNewsState -> NewsListing -> IO ()
```
Martinsos: Nitpick: `process` didn't sound like it will do IO, more like it will do some kind of transformation.

Maybe better names: `executeNewsListing`, `doNewsListing`, `listNews`. Personally I vote for `listNews`.

raw

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
Martinsos: This second bullet point is a bit unusual (if only important, then don't do the timeout and they have to run `wasp news`).
Why not be consistent? I would do the same regardless of critical or important: timeout + y. Otherwise, for important news, you are making it harder to see them + you are making me go and run `wasp news`, I don't see the need for that. Plus you are complicating it both for us and for them. One behaviour ftw.

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
```diff
@@ -0,0 +1,123 @@
...
+        ++ " to confirm you've seen them."
+
+  shouldMarkNewsAsSeen <-
+    if isConfirmationRequired localState listing
+      then getConfirmationFromUser
+      else return True
+
+  if shouldMarkNewsAsSeen
+    then updateTimestampAndMarkAsSeen (getNewsToMarkAsSeen localState listing)
+    else updateTimestampAndMarkAsSeen []
```
Martinsos, line 93:
```suggestion
  updateTimestampAndMarkAsSeen (if shouldMarkNewsAsSeen 
                                then getNewsToMarkAsSeen localState listing
                                else [])
```
or split it into two functions, `updateTimestamp` and `markAsSeen` (unless that doesn't make sense because you don't want to write to file twice I guess).

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
```diff
@@ -0,0 +1,123 @@
...
+  where
+    printNews =
+      unless (null newsToShow) $
+        putStrLn $
+          intercalate "\n\n" $
+            map showNewsEntry newsToShow
```
Martinsos: Look at this beauty
```suggestion
    printNews =
      unless (null newsToShow) $ putStrLn $ intercalate "\n\n" $ showNewsEntry <$> newsToShow
```
hehe ok kidding a bit, feel free to reject.

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
```diff
@@ -0,0 +1,123 @@
...
+        "y"
+        10
+
+    updateTimestampAndMarkAsSeen newsToMarkAsSeen = do
+      currentTime <- T.getCurrentTime
+      saveLocalNewsState $
+        setLastListingTimestamp currentTime $
+          markNewsAsSeen newsToMarkAsSeen localState
+
+getNewsFromListing :: NewsListing -> [NewsEntry]
```
Martinsos:
```suggestion
getAllNewsFromListing :: NewsListing -> [NewsEntry]
```

waspc/cli/src/Wasp/Cli/Command/News/LocalNewsState.hs
```diff
@@ -53,20 +63,31 @@ obtainLocalNewsState = do
...
-getNewsStateFilePath :: IO (Path' Abs File')
+getNewsStateFilePath :: IO (Path' Abs (File LocalNewsStateFile))
 getNewsStateFilePath = do
   waspCacheDir <- getWaspCacheDir <$> getUserCacheDir
-  return $ waspCacheDir </> [relfile|news.json|]
+  return $ waspCacheDir </> newsStateFileInUserCacheDir
+
+data LocalNewsStateFile
+
+newsStateFileInUserCacheDir :: Path' (Rel WaspCacheDir) (File LocalNewsStateFile)
```
Martinsos, line 92: Nice, StrongPath ftw.

waspc/cli/src/Wasp/Cli/Command/News.hs
```diff
@@ -11,33 +11,29 @@ import Data.Maybe (isJust)
...
+  newsEntries <- either (throwError . CommandError "Getting Wasp news failed") pure =<< liftIO fetchNews
+  liftIO $ do
+    localNewsState <- loadLocalNewsState
+    processNewsListing localNewsState $ UserRequestedAllNews newsEntries
 
-fetchAndReportMandatoryNews :: IO ()
-fetchAndReportMandatoryNews = do
+fetchAndListMustSeeNewsIfDue :: IO ()
+fetchAndListMustSeeNewsIfDue = do
   isWaspNewsDisabled <- isJust <$> lookupEnv "WASP_AUTO_NEWS_DISABLE"
```
Martinsos, line 31: Maybe let's sync it with how we call it? But if you don't want to (there are reasons), I do get it, and am also fine with that.

```suggestion
  isWaspNewsDisabled <- isJust <$> lookupEnv "WASP_MUST_SEE_NEWS_DISABLE"
```

waspc/cli/src/Wasp/Cli/Command/News.hs
```diff
@@ -11,33 +11,29 @@ import Data.Maybe (isJust)
...
   isOnCi <- checkIfOnCi
   unless (isWaspNewsDisabled || isOnCi) $ do
-    localNewsState <- obtainLocalNewsState
-    whenM (isTimeForMandatoryNewsReport localNewsState) $ do
+    localNewsState <- loadLocalNewsState
+    whenM (shouldWaspListMustSeeNews localNewsState) $ do
       fetchNewsWithTimeout 2 >>= \case
```
Martinsos, line 36: We give it only 2 seconds fo rthem to be fetched? Yeahy ok I guess it makes sense.

waspc/cli/src/Wasp/Cli/Command/News.hs
```diff
@@ -11,33 +11,29 @@ import Data.Maybe (isJust)
...
         Left _err -> return () -- Wasp stays silent on purpose
```
Martinsos: If you went this far, maybe say why? Obviously it is on purpose :D.

waspc/cli/src/Wasp/Cli/Command/Start.hs
```diff
@@ -23,7 +23,15 @@ import Wasp.Project.Common (dotWaspDirInWaspProjectDir, generatedCodeDirInDotWas
...
 start = do
-  liftIO fetchAndReportMandatoryNews
+  -- We only perform the periodic news check in `wasp start` to avoid being too
+  -- agressive. For example:
+  --   - We don't want to accidentally trigger news in CI.
+  --   - It would be annoying if news came out at you while you were doing
+  --   something like `wasp db migrate-dev`
+  -- Therefore, it's best to keep the periodic news check contained and
+  -- expected. This way we know exactly which workflows it could possibly
+  -- interrupt (LLMs, CIs, people...).
```
Martinsos: For soem reason I can't do a suggestion here, so I will just write it from scratch:

```hs
  -- We check for the news only in `wasp start`, and only periodically,
  -- to avoid being too aggressive. Specifically:
  --   - We don't run it in other `wasp` commands because we don't want to accidentally trigger news in CI (and `wasp start` is rarely used in normal CI, except for maybe e2e testing).
  --   - It would be annoying if news came out at you while you were doing
  --   something like `wasp db migrate-dev`.
  -- Therefore, it's best to keep the periodic news check contained and
  -- expected. This way we know exactly which workflows it could possibly
  -- interrupt (LLMs, CIs, people...).
```

waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -128,6 +133,13 @@ askToChoose question options = do
 askForInput :: String -> IO String
 askForInput question = putStr (Term.applyStyles [Term.Bold] question) >> prompt
 
+askForConfirmationWithTimeout :: String -> String -> Int -> IO Bool
+askForConfirmationWithTimeout message requiredAnswer timeoutSeconds = do
+  maybeAnswer <- timeout timeoutMicroseconds $ askForInput message
+  return $ maybeAnswer == Just requiredAnswer
```
Martinsos:
```suggestion
  (== Just requiredAnswer) <$> timeout timeoutMicroseconds $ askForInput message
```

waspc/cli/src/Wasp/Cli/Interactive.hs
```diff
@@ -128,6 +133,13 @@ askToChoose question options = do
...
+  where
+    timeoutMicroseconds = timeoutSeconds * 1000000
```
Martinsos: So I don't have to count zeroes each time I am looking at this piece of code:
```suggestion
    timeoutMicroseconds = timeoutSeconds * 10^6
```

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
sodic: Not necessarily, we sometimes don't show anything and mark everything as seen (when there's no previous news history).

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
sodic: I added it because of this: https://github.com/wasp-lang/wasp/pull/3537#discussion_r2641343364

But am open to suggestions. 

Oh, you mean just the last part of the comment. Ok, I'll remove it.

waspc/cli/tests/Wasp/Cli/Command/News/ListingTest.hs
```diff
@@ -0,0 +1,101 @@
...
+    emptyLocalNewsState,
+    markNewsAsSeen,
+  )
+
+spec_getNewsToShow :: Spec
+spec_getNewsToShow = do
+  describe "getNewsToShow" $ do
+    describe "for UserRequestedAllNews" $ do
+      it "returns all news in the listing" $ do
+        let newsState = markNewsAsSeen allNews emptyLocalNewsState
```
Martinsos: I am guessing you are using `markNewsAsSeen` here because that is the only way you currently have implemented to add news to the local news state? But it is weird / confusing in this test, feels like sometihng special is happening. Actually I am not sure myself.

Why do you have to do this? Can't you do something like:

```hs
getNewsToShow emptyLocalNewsState (UserRequestedAllNews mockAllNews) `shouldBe` allNews
getNewsToShow mockLocalNewsState (UserRequestedAllNews mockAllNews) `shouldBe` allNews
```

Yeah I guess I don't actually mind you using `markNewsAsSeen`, it is more that your only test here is running on exapmle where all the news are marked as seen -> isn't that a bit of an edge case?

Test for `WaspRequestedMustSeeNews` below looks better -> yeah ok, you didn't cover all the edge cases with empty news and what not, sure, but still, test captures the basic "happy path" and that is a big part of it.

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
sodic: Claude wouldn't have made this mistake. It's over.

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
sodic: > Don't do this if it is too big of a change now,

It's not, just a rename. It would take long before Claude, but now we're good.

> Also, one thing that I feel is missing here, is the information that these [NewsEntry] are all the news ever. 

Yes, I agree it's nicer.

> But then that is a bit of an antipattern.

I didn't do it because of this. But maybe I could do this:
```hs
data ListingContext = UserListingAllNews | WaspListingMustSeeNews
  
data NewsListing = NewsListing
  { listingContext :: NewsListingContext
  , allNews :: [NewsEntry]
  }
```
What do you think?

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
sodic: What would you put in the comment other than "when the user has no news history?"

I can put in a comment, but I'd rather put the condition in the variable and properly name the variable (if it's simple enough). We had `isFirstTimeUser` for this purpose before but it was too broad.

waspc/cli/tests/Wasp/Cli/Command/News/ListingTest.hs
```diff
@@ -0,0 +1,101 @@
...
+        getNewsToShow newsState (WaspRequestedMustSeeNews [info1, important1, important2, critical1, critical2])
+          `shouldMatchList` [important2, critical2]
+
+spec_isConfirmationRequired :: Spec
+spec_isConfirmationRequired = do
+  describe "isConfirmationRequired" $ do
+    describe "for UserRequestedAllNews" $ do
+      it "returns False" $ do
+        let newsState = markNewsAsSeen allNews emptyLocalNewsState
+        isConfirmationRequired newsState (UserRequestedAllNews allNews) `shouldBe` False
```
Martinsos: Again you are testing on an edge case where all the news are seen. I know you know that for the implementation it doesn't matter, but that is an implemenation detail.

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
sodic: It's an easy change. But you probably meant "only the shown news", not "only new news."

I have that function available. I'll do it.

waspc/cli/tests/Wasp/Cli/Command/News/ListingTest.hs
Martinsos: Ha I don't know. I commented on the implementation itself, it feels a bit too out of nowhere, maybe this is just consequence of that.

On the other hand, I would rather have test that too much duplicates the implementation, than not test it at all. At least it will fail if impelmentation changes, even if not super nice or easy to maintain.

How could it be made better though? Maybe ditch the property test here and just give me normal unit tests, one or two. And make description a bit more normal. So basically giving up on integration test and doing unit test(s).

But on one hand, what you have is ok. This is what we dsicussed, obviously this fucntion is somewhere in the middle, and well, it's logic is complex in this way. Unless we can fix the logic itself, I think it make sense the tests will be somewhat complex also.

I would think about it a bit, but not too much.

waspc/cli/tests/Wasp/Cli/Command/News/ListingTest.hs
```diff
@@ -0,0 +1,101 @@
...
+spec_getNewsToMarkAsSeen = do
+  describe "getNewsToMarkAsSeen" $ do
+    describe "for UserRequestedAllNews" $ do
+      it "returns all news entries" $ do
+        let newsState = markNewsAsSeen allNews emptyLocalNewsState
+        getNewsToMarkAsSeen newsState (UserRequestedAllNews allNews) `shouldMatchList` allNews
+
+    describe "for WaspRequestedMustSeeNews" $ do
+      prop "returns all news when there is no previous news history" $
+        forAll waspInvokedScenario $ \(_, listing) ->
```
Martinsos: waspInvokedScenario never got renamed to new naming, I think.

waspc/cli/tests/Wasp/Cli/Command/News/ListingTest.hs
```diff
@@ -0,0 +1,101 @@
...
+          isConfirmationRequired newsState listing ==>
+            getNewsToMarkAsSeen newsState listing === getNewsToShow newsState listing
+
+waspInvokedScenario :: Gen (LocalNewsState, NewsListing)
+waspInvokedScenario = do
+  seenNews <- subsetOf allNews
+  newsInListing <- subsetOf allNews
+  let newsState = markNewsAsSeen seenNews emptyLocalNewsState
+      listing = WaspRequestedMustSeeNews newsInListing
+  return (newsState, listing)
```
Martinsos: Nice

waspc/cli/tests/Wasp/Cli/Command/News/ListingTest.hs
```diff
@@ -0,0 +1,101 @@
...
+
+subsetOf :: [a] -> Gen [a]
+subsetOf = elements . subsequences
+
+allNews :: [NewsEntry]
+allNews = [info1, info2, important1, important2, critical1, critical2]
```
Martinsos: I usually prefix mocks with `mock` because then when reading stuff I know it is mocks, I was a bit confused up there with figuring out what is `allNews`, but if I saw `mockAllNews`, there is no doubt.

But will you do it then for all the critical1 and otehrs? I don't know. I guess you should. Sounds nosiy though. Hm.

raw

waspc/cli/tests/Wasp/Cli/Command/News/LocalNewsStateTest.hs
```diff
@@ -0,0 +1,31 @@
...
+      let state = markNewsAsSeen [newsEntry1, newsEntry2] emptyLocalNewsState
+      wasNewsEntrySeen state newsEntry1 `shouldBe` True
+      wasNewsEntrySeen state newsEntry2 `shouldBe` True
+      wasNewsEntrySeen state newsEntry3 `shouldBe` False
+
+    it "correctly joins seen entries" $ do
+      let state1 = markNewsAsSeen [newsEntry1] emptyLocalNewsState
+          state2 = markNewsAsSeen [newsEntry2] state1
+      wasNewsEntrySeen state2 newsEntry1 `shouldBe` True
+      wasNewsEntrySeen state2 newsEntry2 `shouldBe` True
```
Martinsos, line 25: Maybe give me one False check for newsEntry3 here, since yo ualready have i tin where. Or remove it from where, But I would check, for consistency.

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
sodic: We decided to keep the current behavior for now.

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
sodic: Martin said he didn't like it and wanted to use the antipattern. He's the boss, so I implemented the antipattern :)

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
sodic: I did something. Check it.

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
sodic: I did it this way on purpose. I prefered the repetition than another if branch, made it difficult to follow IMO.

And the two functions - yes, I did not want to write to the file twice.

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
sodic: No longer necessary.

waspc/cli/tests/Wasp/Cli/Command/News/ListingTest.hs
sodic: On purpose, but renamed now

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
Martinsos: Yeah that it is, good!

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
```diff
@@ -0,0 +1,117 @@
...
+    isUnseen = not . wasNewsEntrySeen localState
+
+isConfirmationRequired :: LocalNewsState -> NewsListing -> Bool
+isConfirmationRequired localState listing = case listing of
+  UserListingAllNews {} -> False
+  WaspListingMustSeeNews {} -> any ((== Critical) . level) $ getNewsToShow localState listing
+
+getNewsToMarkAsSeen :: LocalNewsState -> NewsListing -> [NewsEntry]
+getNewsToMarkAsSeen localState listing = case listing of
+  UserListingAllNews {} -> getNewsToShow localState listing
```
Martinsos, line 67: Hm. So I remember I had old comment here saying that we should not just return all the news, but only those that haven't been seen before, so the new ones.
And you did change this part, but to reurn all news to show, which are again all news, so the same behaviour but semantically different, and that might be ok if those sematnics made sense, but I don't see how they do, why would we mark as seen all the news that you will show? I guess because they are shown. Ok that is certianly an improvement over just marking all news, yeah I get it. But still hm. It still might again mark already marked news. If we are ok with that, then ok I guess.

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
Martinsos: Hoho :D. Let' see how your further experiments will go (after this PR).

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
Martinsos: Ok I get why not two functions.

However, for branching going first -> I found it confusing, becuase made me think the following logic is depended on this. But only part is dependent, while part is repeated. And since we are in language where `if` can be an expression, you are given perfect mechanism to capture this correclty, the way I described. So for me, the suggestion I made is a better way to represent this logic: update timestamp for sure, and mark as seen conditionally. Because that is the real logic here.

That said, this is not important enough thing to waste too much time on, so if this didn't convince you, I am ok with you sticking to what you have.

raw

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
```diff
@@ -0,0 +1,117 @@
...
+  WaspListingMustSeeNews {allNews}
+    -- If the user has no news history, we want to consider all past news seen.
+    | not $ doesUserHaveNewsHistory localState -> allNews
+    | isConfirmationRequired localState listing -> getNewsToShow localState listing
+    -- We don't want to assume the user saw the news if they haven't confirmed so
```
Martinsos, line 72: Ok, this comment helps for sure.

waspc/cli/tests/Wasp/Cli/Command/News/ActionTest.hs
sodic: No longer applicable

waspc/cli/src/Wasp/Cli/Command/News/Listing.hs
sodic: We decided we are OK.

Martinsos (approved):

PR #3500, opened 2025-12-15
Move embedded npm packages to the waspc data dir

cprecioso:
## Description

- Prompted by https://github.com/wasp-lang/wasp/pull/3483#pullrequestreview-3572325998

Moves npm packages from `waspc/packages/*` to `waspc/data/packages/*`, removes the scripts that copied them from the former to the latter, and updates scripts and documentation that pointed to the old location.

### Motivation

We embed npm packages as part of the CLI functionality (wasp deploy, tsconfig, studio, etc). These lived in `waspc/packages`, and we had a script to copy them to `waspc/data/packages`, where they were actually picked up by the build.

The reason to have this two-stage system was to enable us to remove node_modules when we copied to the data dir, so that we wouldn't ship that folder to users (the CLI does an `npm install` on these packages on first run).

This solution proved to be ineffective, as we were bundling them anyway. I realized that and updated the packaging script to read and follow the Cabal data file globs, which already filtered out the node_modules. (#3483)

This means that the initial copy step from `waspc/packages` to `waspc/data/packages` is now redundant, so we can eliminate it and simplify the process.

### Why were we bundling `node_modules` into the binary even though we filtered them at the copy step?

In the build process, we are nonetheless calling `npm install` on the packages in their new location, recreating the `node_modules` folder. The packaging script would then just blindly copy them into the final tarball.

## Type of change

- [ ] **🔧 Just code/docs improvement** 
- [ ] **🐞 Bug fix** 
- [ ] **🚀 New/improved feature** 
- [ ] **💥 Breaking change** 

## Checklist

**Checklist not applicable** I did test the final built artifacts in a Wasp project and run every command that depends on an embedded package, successfully.

<details>
<summary>Show it anyway</summary>

- [x] I tested my change in a Wasp app to verify that it works as intended.

- 🧪 Tests and apps:

  - [ ] I added **unit tests** for my change. 
  - [ ] _(if you fixed a bug)_ I added a **regression test** for the bug I fixed. 
  - [ ] _(if you added/updated a feature)_ I added/updated **e2e tests** in `examples/kitchen-sink/e2e-tests`.
  - [ ] _(if you added/updated a feature)_ I updated the **starter templates** in `waspc/data/Cli/templates`, as needed.
  - [ ] _(if you added/updated a feature)_ I updated the **example apps** in `examples/`, as needed.
    - [ ] _(if you updated `examples/tutorials`)_ I updated the tutorial in the docs (and vice versa).

- 📜 Documentation:

[...]

PR #3562, opened 2025-12-29
Refactor build commands for clarity and consistency

cprecioso:
## Description

Requested by @sodic at https://github.com/wasp-lang/wasp/pull/3500#discussion_r2650423185

Rename build commands to be more intuitive and consistent. The new naming convention makes it clearer what each command builds:

- `build` now builds Haskell + TS packages (previously `build:all`)
- `build:hs` builds Haskell only (previously `build`)
- `build:static` builds Haskell + packages statically (previously `build:all:static`)
- `build:static:hs` builds Haskell statically only (new)

### DIFF
--- a/.github/workflows/ci-waspc-build.yaml
+++ b/.github/workflows/ci-waspc-build.yaml
@@ -122,8 +122,9 @@ jobs:
         working-directory: waspc
         env:
           LC_ALL: C.UTF-8 # In some Docker containers the LOCALE is not UTF-8 by default
+          BUILD_STATIC: ${{ matrix.env.static && '1' || '' }}
         run: |
-          ./run ${{ matrix.env.static && 'build:static' || 'build' }}
+          ./run build
           mkdir -p artifacts
           ./tools/make_binary_package.sh "artifacts/${{ steps.compute-artifact-names.outputs.tarball-name }}"
 
--- a/waspc/run
+++ b/waspc/run
@@ -22,11 +22,9 @@ DEFAULT_COLOR="\033[39m"
 WASP_PACKAGES_COMPILE="for pkg in ${SCRIPT_DIR}/data/packages/*/package.json; do \
   (cd \$(dirname \$pkg) && npm install && npm run build); \
 done"
-BUILD_HS_CMD="cabal build all"
+BUILD_HS_CMD="cabal build all ${BUILD_STATIC:+--enable-executable-static}"
 BUILD_HS_FULL_CMD="cabal build all --enable-tests --enable-benchmarks"
 BUILD_ALL_CMD="$WASP_PACKAGES_COMPILE && $BUILD_HS_CMD"
-BUILD_HS_STATIC_CMD="$BUILD_HS_CMD --enable-executable-static"
-BUILD_ALL_STATIC_CMD="$WASP_PACKAGES_COMPILE && $BUILD_HS_STATIC_CMD"
 
 INSTALL_CMD="$WASP_PACKAGES_COMPILE && cabal install --overwrite-policy=always"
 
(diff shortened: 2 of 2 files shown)


.github/workflows/ci-waspc-build.yaml
```diff
@@ -123,7 +123,7 @@ jobs:
         env:
           LC_ALL: C.UTF-8 # In some Docker containers the LOCALE is not UTF-8 by default
         run: |
-          ./run build:all${{ matrix.env.static && ':static' || '' }}
+          ./run ${{ matrix.env.static && 'build:static' || 'build' }}
```
cprecioso: Moved the full name into the conditional so it appears when we do a grep

waspc/README.md
```diff
@@ -158,7 +158,7 @@ When done, new tab in your browser should open and you will see the Kitchen Sink
 3. Do a change in the codebase (most often in `src/` or `cli/src/` or `data/`), and update tests if that makes sense (see [Test](#tests)).
    Fix any errors shown by HLS/`ghcid`.
    Rinse and repeat. If you're an internal team member, postpone updating waspc e2e tests tests until approval (see [here](#note-for-team-members)).
-4. Use `./run build` to build the Haskell/cabal project, and `./run wasp-cli` to both build and run it. If you changed code in `data/packages/`, you will also need to run `./run build:packages` (check [TypeScript Packages section](#typescript-packages) for more details). Alternatively, you can also run slower `./run build:all` to at the same time build Haskell, TS packages, and any other piece of the project in one command.
+4. Use `./run build:hs` to build the Haskell/cabal project, and `./run wasp-cli` to both build and run it. If you changed code in `data/packages/`, you will also need to run `./run build:packages` (check [TypeScript Packages section](#typescript-packages) for more details). Alternatively, you can also run slower `./run build` to at the same time build Haskell, TS packages, and any other piece of the project in one command.
```
sodic, line 161: Hm, I'm thinking `./run wasp-cli` should perhaps also call `./run build` (ex `./run build:all`), no?

waspc/run
```diff
@@ -118,13 +119,15 @@ print_usage() {
...
-  print_usage_cmd "build:all:static" \
-    "Builds the Haskell project statically + all sub-projects (i.e. TS packages). Only useful for release builds. Needs to be run on a musl-based Linux distribution (e.g. Alpine)."
+  print_usage_cmd "build:hs" \
+    "Builds the Haskell project only."
   print_usage_cmd "build:packages" \
     "Builds the TypeScript projects under data/packages/."
+  print_usage_cmd "build:static" \
+    "Builds the Haskell project statically + all sub-projects (i.e. TS packages). Only useful for release builds. Needs to be run on a musl-based Linux distribution (e.g. Alpine)."
+  print_usage_cmd "build:static:hs" \
+    "Builds the Haskell project statically only. Only useful for release builds. Needs to be run on a musl-based Linux distribution (e.g. Alpine)."
```
sodic: Since static only applies to `hs`, perhaps we should flip the order into `build:hs:static`. Yes, I see this is slowly getting out of hand :)

I think the problem comes because we're mixing together two concerns. My idea:
```bash
./run build:hs
./run build:packages

./run static-build:hs
./run static-build:packages
```
And then `static-build:packages` would just delegate to `build:packages`.

Do what you think is best, maybe this will be a good inspiration.

sodic (approved):
Looks good. I didn't check for all occurrences of old names. I trust you did that :)

waspc/README.md
cprecioso: So, IMO, yeah it should, I'd like for a single command to rebuild everything so I don't have to juggle many commands when I do a change (this is also something I lobbied for in gen&rarr;libs PRs).

But in my M4 Pro laptop, `./run build:packages` takes 18s, each time. This is not something I want to put up with, every time I need to call the Wasp command. So I would say not to include it for now.

I think the solution for this would be a smarter task orchestrator that could intelligently cache and run only what is needed (see #3099 -- but maybe even a Makefile could be enough). But in the meantime I would like to keep it as is.

waspc/run
cprecioso: Mmmm I don't like having it as `static-build` as the static thing is just a variant over regular build. Maybe it would be better expressed as `BUILD_STATIC=1 ./run build`? Wdyt?

sodic: `static-build` is still my favorite option: It's less hidden and doesn't mix concerns ("static or dynamic" with "what to build". But I also don't love this option.

The env var is fine if you like it. I'd just like to avoid throwing static into the colon-separated stuff. Those should only answer the question "what to build" IMO.

But again, not very important. Do as you please :)

PR #3572, opened 2025-12-31
Update `basic` to React 19

cprecioso:
## Description

Updates the `basic` starter template to the latest best practices from React and the React Typescript Handbook.

- Types function components directly, without helper types from `@types/react` [(link)](https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/)
- Types `children` as `ReactNode` [(link)](https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/basic_type_example#useful-react-prop-type-examples)
- Uses `ComponentProps<"tag-name">` to get html component props, including their refs [(link)](https://react-typescript-cheatsheet.netlify.app/docs/react-types/componentprops/)
- Uses `import { bla } from "react"` instead of `import * as React` or `import React` [(link)](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#:~:text=The%20default%20imports%20will%20keep%20working%20in%20React%2017%2C%20but%20in%20the%20longer%20term%20we%20encourage%20moving%20away%20from%20them.)
- Stores element refs in `useState` instead of `useRef` where possible, which is preferred and improves syncing between the React and DOM worlds [(link)](https://react.dev/learn/referencing-values-with-refs#differences-between-refs-and-state)
- Moves stuff out of effects and into event callbacks where possible [(link)](https://react.dev/learn/you-might-not-need-an-effect)
- Removes `forwardRef` [(link)](https://react.dev/reference/react/forwardRef)
- And more context-specific changes that I will explain in the comments

I didn't use more modern features like [Actions](https://react.dev/reference/react/useTransition#examples) or [Suspense](https://react.dev/reference/react/Suspense) as these are generally worse understood for beginners and there's not as much mindshare for them.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/data/Cli/starters/basic/src/shared/components/Dialog.tsx
+++ b/waspc/data/Cli/starters/basic/src/shared/components/Dialog.tsx
@@ -15,7 +15,7 @@ export function Dialog({
   children,
   closeOnClickOutside = true,
 }: DialogProps) {
-  const dialogRef = useRef<HTMLDialogElement >(null);
+  const dialogRef = useRef<HTMLDialogElement>(null);
 
   useEffect(
     function handleShowOrCloseDialog() {
@@ -28,7 +28,7 @@ export function Dialog({
         dialog.close();
       }
     },
-    [open, dialogRef],
+    [open],
   );
 
   const handleClick = useCallback(
@@ -47,7 +47,7 @@ export function Dialog({
         onClose();
       }
     },
-    [closeOnClickOutside, onClose, dialogRef],
+    [closeOnClickOutside, onClose],
   );
 
   return createPortal(


waspc/data/Cli/starters/basic/src/auth/email/EmailVerificationPage.tsx
```diff
@@ -8,7 +8,7 @@ export function EmailVerificationPage() {
       <VerifyEmailForm />
       <br />
       <span className="text-sm font-medium text-neutral-900">
-        {"If everything is okay, "}
```
cprecioso, line 11: We don't need to put this in between `{}` and it makes it look weird.

waspc/data/Cli/starters/basic/src/auth/email/LoginPage.tsx
```diff
@@ -8,15 +8,15 @@ export function LoginPage() {
       <LoginForm />
       <br />
       <span className="text-sm font-medium text-neutral-900">
-        {"Don't have an account yet? "}
+        Don&apos;t have an account yet?{" "}
```
cprecioso, line 11: Linting warning if we added `'` directly, as it can be confused with a programming symbol.

waspc/data/Cli/starters/basic/src/shared/components/Portal.tsx
cprecioso, line 1: The `<Portal>` component was not adding anything to our usecase[1], and was a complex bit of logic that could scare away less savvy users. The only place where it was used was well-served with a regular `createPortal()` call.

---

[1]: The `<Portal>` component seems to be wrapping just one main bit of logic: not to mount stuff on SSR and mount it only in client mode. However, Wasp apps are always client-only, so we don't need that check.

waspc/data/Cli/starters/basic/src/shared/components/Dialog.tsx
```diff
@@ -13,73 +15,51 @@ export function Dialog({
...
+      if (clickedOutside) {
+        onClose();
+      }
     },
-    [open],
+    [closeOnClickOutside, onClose, dialogRef],
   );
 
-  return (
+  return createPortal(
```
cprecioso, line 53: Use `createPortal` directly as I removed the `<Portal>` component.

waspc/data/Cli/starters/basic/src/tags/components/CreateTagDialog.tsx
```diff
@@ -1,11 +1,11 @@
-import React from "react";
+import { useId, useState } from "react";
 import { Button } from "../../shared/components/Button";
 import { Dialog } from "../../shared/components/Dialog";
-import { Portal } from "../../shared/components/Portal";
-import { CREATE_TAG_FORM_ID, CreateTagForm } from "./CreateTagForm";
```
cprecioso, line 5: We don't need to import a global constant if we just do `useId` in the parent component and pass it down. This way we discourage hardcoded IDs which might lead to collisions.

waspc/data/Cli/starters/basic/src/shared/components/Dialog.tsx
```diff
@@ -13,73 +15,51 @@ export function Dialog({
...
-      return () => {
-        document.body.style.overflow = originalOverflow;
-      };
+      if (clickedOutside) {
+        onClose();
+      }
     },
-    [open],
+    [closeOnClickOutside, onClose, dialogRef],
   );
```
cprecioso, line 71: This effect felt iffy, as it could create conflicts between multiple instances very easily.

I migrated this to a global style (`body:has(dialog:open) { overflow: hidden }` but written in Tailwind) which should avoid those conflicts.

waspc/data/Cli/starters/basic/src/App.css
```diff
@@ -2,6 +2,10 @@
 @tailwind components;
 @tailwind utilities;
 
+body {
+  @apply has-[dialog:open]:overflow-hidden;
+}
```
cprecioso, line 7: This might be better written as [TW `group` thingy](https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state) but I lack enough TW knowledge to get the correct incantation. I'm open to suggestions.

waspc/data/Cli/starters/basic/src/shared/components/Dialog.tsx
FranjoMindek: Implication here is that every `Dialog` is a `Portal` component.
That might not be the case. 
Non-modal `Dialog` components don't have to be portalled.

I'm split between:
1. Not doing this because:
	1.1. The `Dialog` does more that one job (it's not just a `Dialog`, it also creates portals).
	1.2. The `Dialog` can never be used in a non-portal way.
2. Keeping this, as it simplifies the code. And all our `Dialog`s are modals.

RAW.
Just documenting the change.

waspc/data/Cli/starters/basic/src/tags/components/CreateTagDialog.tsx
```diff
@@ -19,31 +19,32 @@ export function CreateTagDialog() {
...
+                id={formId}
+                onTagCreated={() => setTagDialogOpen(false)}
+              />
+            </div>
+            <footer className="flex justify-end gap-2 px-4 py-6 lg:px-6 lg:py-8">
+              <Button form={formId} type="submit">
+                Create
+              </Button>
+              <Button
+                form={formId}
```
FranjoMindek: This `form={formId}` is not really necessary (logic wise) here since we change the state with `onClick` event here. But it makes sense in sense of "semantic ownership".

It was also in my old code. RAW.

waspc/data/Cli/starters/basic/src/App.css
FranjoMindek: This also assumes that all `Dialog` components are modals, and that any `Dialog` should block the screen from scrolling. It is in line with the `Dialog` component changes.

If we do keep `Dialog` the like this, I think we don't need to use a `group` here.
We specifically want to style the parent (`body`), not children based on some parent state. This is fine then.

My only concern with the `Dialog` = modal approach is that if someone continues on from the `basic` starter, they might encounter problems from this "global css" at some point.

waspc/data/Cli/starters/basic/src/auth/email/EmailVerificationPage.tsx
FranjoMindek: I did it so I don't have to deal with special chars (`'`) and it was ✨*prettier*✨ than only having the whitespace in the brackets. I'm fine with both honestly.

waspc/data/Cli/starters/basic/src/shared/components/Dialog.tsx
```diff
@@ -13,73 +15,51 @@ export function Dialog({
...
-          e.clientX > rect.right ||
-          e.clientY < rect.top ||
-          e.clientY > rect.bottom;
-
-        if (clickedOutside) {
-          onClose();
-        }
-      };
-
-      dialog.addEventListener("click", handleClick);
```
FranjoMindek, line 52: Good job on removing these.
I don't remember why I added them.
Did I think that the component gets removed so I needed to apply the event handler programmatically?

I most likely copied these from radix.
Doesn't look like removing these breaks anything.

waspc/data/Cli/starters/basic/src/shared/components/Dialog.tsx
```diff
@@ -13,73 +15,51 @@ export function Dialog({
   children,
   closeOnClickOutside = true,
 }: DialogProps) {
-  const dialogRef = React.useRef<HTMLDialogElement>(null);
```
FranjoMindek, line 16: This is one of these "rare cases" when using `useRef` is preferred:

https://react.dev/learn/referencing-values-with-refs#when-to-use-refs
https://react.dev/learn/manipulating-the-dom-with-refs

`useRef` works great with DOM element refs. It is `useRef`'s most often use case.

> Refs are a generic concept, but most often you’ll use them to hold DOM elements.
> Usually, you will use refs for non-destructive actions like focusing, scrolling, or measuring DOM elements


I think the big question for `useState` is, should it trigger re-renders?
Should setting a `ref` trigger re-renders? It might, but not in this case.

With `useState` we always have to make 2 renders (4 in dev mode) to render a `Dialog`. A single render is never possible. This is because setting `dialogRef` always triggers a re-render, and we need to set it before the component can work properly.

`useRef` solution:
https://github.com/user-attachments/assets/8ba85914-c89d-48ea-97e9-9b9389523de5

`useState` solution:
https://github.com/user-attachments/assets/3a9eac35-b6bc-49d3-9ceb-4db949dd7f67

waspc/data/Cli/starters/basic/src/tags/components/CreateTagDialog.tsx
FranjoMindek: This is a nice way to avoid hardcoding.

I usually do `CREATE_TAG_FORM_ID` because I reuse the `id` for testing purposes. I should probably use `data-testid`.

waspc/data/Cli/starters/basic/src/shared/components/Dialog.tsx
FranjoMindek: Yes, though we should definitely avoid modal dialogs inside of a modal dialog. That seems like an anti-pattern.

waspc/data/Cli/starters/basic/src/shared/components/Dialog.tsx
cprecioso: IMO the `src/client/components` is a makeshift design system, we're already offering components a punch above the basic HTML/JS primitives (and we're calling functions like `ref.showModal()` etc), so it's fine for us to offer a curated component.

waspc/data/Cli/starters/basic/src/tags/components/CreateTagDialog.tsx
cprecioso: Yeah, I get that. The recommended way is to actually [avoid testids](https://testing-library.com/docs/queries/bytestid#:~:text=it%20is%20recommended%20to%20use%20this%20only%20after%20the%20other%20queries%20don%27t%20work%20for%20your%20use%20case) and just look for text labels, the same way a user would.

waspc/data/Cli/starters/basic/src/tags/components/CreateTagDialog.tsx
cprecioso: oh you're right, removing!

waspc/data/Cli/starters/basic/src/shared/components/Dialog.tsx
cprecioso: I don't disagree with your points. I think for the current usecase it is about the same outcome, so I undid my changes.

But I think  using a state is more in line with React's semantics, even if it triggers more re-renders -- that is the point, IMO. It boils down to a (hypothetical) question: "if this object changed, would I want to re-run effects based on that?" => "if suddenly the ref started pointing to a different dialog, would I want to synchronize my `open` prop with it?" => my answer is yes.

Of course, it's hypothetical and Hyrum's Law would ensure that ReactDOM's ref lifetimes remain unchanged; but because of my answer, I'd use a state rather than a ref.

---

Anyway, small thing, and people might get confused about storing DOM in states, so I undid the change.

cprecioso:
@FranjoMindek ready for re-review

waspc/data/Cli/starters/basic/src/shared/components/Dialog.tsx
FranjoMindek: Yes, I think your point boils to the same thing I said:

> I think the big question for `useState` is, should it trigger re-renders?
Should setting a `ref` trigger re-renders? It might, but not in this case.

Current component does not change refs so it doesn't have a need for them to trigger re-renders. But if it did then it definitely would be an `useState` situation.

waspc/data/Cli/starters/basic/src/shared/components/Dialog.tsx
```diff
@@ -13,9 +15,9 @@ export function Dialog({
...
+  const dialogRef = useRef<HTMLDialogElement >(null);
```
FranjoMindek:
```suggestion
  const dialogRef = useRef<HTMLDialogElement>(null);
```

waspc/data/Cli/starters/basic/src/shared/components/Dialog.tsx
```diff
@@ -26,60 +28,40 @@ export function Dialog({
         dialog.close();
       }
     },
-    [open],
+    [open, dialogRef],
```
FranjoMindek:
```suggestion
    [open],
```

waspc/data/Cli/starters/basic/src/shared/components/Dialog.tsx
```diff
@@ -26,60 +28,40 @@ export function Dialog({
...
-      document.body.style.overflow = "hidden";
-      return () => {
-        document.body.style.overflow = originalOverflow;
-      };
+      if (clickedOutside) {
+        onClose();
+      }
     },
-    [open],
+    [closeOnClickOutside, onClose, dialogRef],
```
FranjoMindek:
```suggestion
    [closeOnClickOutside, onClose],
```

cprecioso:
Yep, thanks for catching it! Waiting for CI and merging 🚀

PR #3574, opened 2026-01-02
Add versioned docs map to llms.txt and refactor generation script

vincanger:
## Summary

- Adds a "Documentation Maps by Version" section to all `llms.txt` files, allowing LLMs to easily access docs for specific Wasp versions

- Replaced `processDocumentationFiles()` (which did 3 things) with:
- `gatherDocumentation()` - collects and structures doc data
- `buildDocumentationMap()` - pure function for llms.txt TOC
- `buildFullDocumentation()` - pure function for llms-full.txt content

### DIFF
--- a/web/scripts/generate-llm-files.ts
+++ b/web/scripts/generate-llm-files.ts
@@ -135,7 +135,7 @@ async function generateFiles() {
 
 function buildDocsMapsByVersionSection(versions: string[]): string {
   const latestVersion = versions[0];
-  let section = `## Documentation Maps by Version\n*IMPORTANT: First run \`wasp version\` to get the installed Wasp version before following the relevant link.*`;
+  let section = `## Documentation Maps by Version\n*IMPORTANT:* You should run \`wasp version\` to get the installed Wasp CLI version before choosing the correct link.\n`;
   for (const version of versions) {
     const label = version === latestVersion ? `${version} (latest)` : version;
     section += `- [${label}](${WASP_BASE_URL}llms-${version}.txt)\n`;


Martinsos:
I think I commented before that I find this whole logic to be overly complex, so I didn't go into reviewing it here, I believe we will handle that separately, but I am curious to understand what is the functional result of this change!

So what are we going to have as a result of this PR -> llms.txt for the latest version, llms-{version}.txt for all the other versions, and also each llms(-{version})*.txt is going to have a list of all the versioned llms-{version}.txt in it? Is that it? And we have one llms-full.txt for the latest version only? If not, can you please help me understand what will we have?

If it is as I described above, then that is a step in the right direction I believe, but it sounds a bit chaotic to me still. I don't like the exceptions: exception of llms.txt being the latest version, and exception of having llms-full.txt only for the latest version.
How I imagine we solve the first exception:
- we have llms.txt which is nothing but a list of llms-{version}.txt links.
- we have llms-{version}.txt for each version, including the latest one.

As for llms-full.txt -> hm not sure, I think it would make sense to generate one per version also then! Why would we have just one for the latest docs, when we can have one for each version? And if we don't need it, then let's not have it at all.

This approach I suggested above: does that make sense? Is there some practical reason why this would not be good? I guess there is one extra step, it has to fetch llms.txt first just to figure out which llms-{version}.txt it should fetch. But maybe that is ok? What do you think? We can cache all of these anyway. If this is too much and impractical, that is ok, but let's think more about it then.

Btw what is the workflow with what this PR of yours introduces -> it would fetch llms.txt first? And then it would again fetch the correct llms-{version}.txt? So I guess the same thing as with approach I described, except that for the latest wasp version, it wouldn't have to do that one extra step, but for other older wasp versions i twould have to?

Let's discuss this!

vincanger:
> So what are we going to have as a result of this PR -> llms.txt for the latest version, llms-{version}.txt for all the other versions, and also each llms(-{version})*.txt is going to have a list of all the versioned llms-{version}.txt in it? Is that it? And we have one llms-full.txt for the latest version only? If not, can you please help me understand what will we have?

Yep. This is what it does/outputs. When you frame it like that, it does seem ridiculous, but it's an attempt at adhering to the [standardized llms.txt approach](https://llmstxt.org/) while also adding our new versioned llms.txt files for Claude Code (and LLMs in general). 

To follow this proposed standard, we definitely want to keep a default llms.txt and llms-full.txt (of the latest Wasp version). So I think a hybrid approach similar to what you suggested would be best. Here's what I suggest:
- `llms.txt` becomes more of a map to the most important LLM-friendly Wasp files
- within the `llms.txt`, perhaps as the first section, we include a **Documentation Maps by Version** Section
- the Docs Map section includes an index all of our versioned `llms.txt` files renamed to e.g. `wasp-docs-map-v${x}.md`
- each `wasp-docs-map.md` file contains raw github md docs links (as llms.txt does at the moment) along with the hierarchical structure (i.e. sub-sections) of each section/guide to help the LLM retrieve understand what each guide contains (see the [claude_code_docs_map.md](https://code.claude.com/docs/en/claude_code_docs_map.md) as a good example)
- use a StartSession hook to cache this `wasp-docs-map-v${x}.md` file based on their wasp version
  - note: Claude Code doesn't cache their docs map markdown file -- so CC does a WebFetch for the file, then another for the relevant guide -- but we have multiple versioned files to choose from which makes me think caching on start is a good approach. Nevertheless, I'm wondering if we should skip this caching step/hook altogether and just let CC do multiple Fetch hops? In the end, the text files are super small and quick to access.
- create a rule in CLAUDE.md to always ground wasp knowledge by WebFetching the relevant docs via `wasp-docs-map-v${x}..md` 

As for `llms-full.txt`, We could go ahead and create versioned files as well. The `llms-full.txt` URL would default to the latest version, so the question is, do we create a duplicate at `llms-full-v${x}.txt as well? And how do we make LLMs aware of them? Perhaps a section in the `llms.txt` file, below our new Docs Maps by Version section?

> Btw what is the workflow with what this PR of yours introduces -> it would fetch llms.txt first? And then it would again fetch the correct llms-{version}.txt? So I guess the same thing as with approach I described, except that for the latest wasp version, it wouldn't have to do that one extra step, but for other older wasp versions i twould have to?

The current workflow for our plugin is similar to what I described above: StartSession hook caches versioned llms.txt with a rule that states to use that file for finding and fetching docs (fallback to fetching https:wasp.sh/llms.txt) if none is found

Martinsos:
it does seem ridiculous, but it's an attempt at adhering to the [standardized llms.txt approach](https://llmstxt.org/) while also adding our new versioned llms.txt files for Claude Code (and LLMs in general).
> 
> To follow this proposed standard, we definitely want to keep a default llms.txt and llms-full.txt (of the latest Wasp version).

So I just read the whole https://llmstxt.org/, and I haven't been able to make these same conclusions: I don't know if I am missing something, but I don't see any part of specification being about llms.txt having to be about the latest version of the software.
What I understood is that llms.txt specification is very open, and that is it just a curated web page entry point for an LLM. And there is no specification as to what should be inside, except for H1 with title of the webpage, quoted section with brief introduction, and then sections that provide further instructions, links, whatever. Which is super open.

So following that, I don't see why our llms.txt wouldn't be (if I was doing this for real I would copy their example and make sure it is simlar, I am just spitballing it here):

```md

* Wasp

Develop full-stack web apps faster ...

> Wasp is a ...
> ....
> ....

** Maybe some general knowledge, not version-specific
...

** Maybe some more general knowldge, not version-specific
...

** Wasp docs maps per wasp version
Explanation of what this is if needed.
- v0.19.0 (latest): https://wasp.sh/llms-0.19.0.txt
- v0.18.0: <link>
- ...

** Other useful links:
- Wasp Discord
- Wasp Github
- Wasp Blog

```

So llms.txt is an entry point, and we can pick how we want it to direct the LLM.

llms-full.txt is maybe a bit tricky, I admit, because it expects to have all in it. But well, it could! We could put in it all the docs, for all the versions of wasp. However, it will be quite huge! Which doesn't feel right. So another solution is to have it just have links to all the versions llms-full-{version}.txt files. Why wouldn't that work at the end? LLM are smart enough to follow that. Unless there is deterministic piece of code hm that downloads it. In that case we could go with it having the current, latest docs, full, + a section (at the start) warning this is the latest version + links to other versions.

Btw we could do the same thing for llms.txt, yeah, it could be stuff for latest version + section (I hope at start) linking to llms-{version}.txt files. But that feels like a lot of redundant stuff being given to LLM that might even confuse it, I don't like that really. Keep in mind that often, most of the users won't be using the latest version of Wasp. Ok true, likely 95% of the latest docs content will be correct for a couple of previous versions, but we don't know which aprt is not correct + there could be a bigger change.

I mean it really comes down to also testing this stuff. Does LLM mind having to do one extra fetch, when getting llms.txt? If it doesn't, then the best solution is having thin llms.txt just offering gemeral info + pointing to versions docs maps.
If it doesn't like doing additioanl fetch, then maybe we should inline into it the doc maps for a couple of latest versions, like 3 or something, and link tot he rest. But I doubt this is the case.

As for caching -> I would leave that as a seaparte issue. once we get this working first, we can play with caching, why not.

Oh btw, interesting library potentially, I don't know if it is flexible enough for what we need due to versioning, but if it is, might be cool to use instead of our own code: https://github.com/rachfop/docusaurus-plugin-llms . I wouldn't push it though if we loose to much flexibility. I see it has some support for versioning, but it looks quite crude/manual on that front.

vincanger:
Now we have:
- llms.txt as an entry point file 
- llms-{version}.txt as versioned docs maps
- llms-full.txt as the latest version full concatenated docs + an index header with links to...
- llms-full-{version}.txt files for each version

image

image

Issue #2319, opened 2024-10-03
Update `vitest` and related testing deps


infomiho:
We are using Vitest v1 and we should upgrade to Vitest v2: https://github.com/vitest-dev/vitest/releases/tag/v2.0.0

The related deps are:
- `@vitest/ui`
- `jsdom`
- `@testing-library/react`
- `@testing-library/jest-dom`
- `msw`

infomiho:
One thing we maybe want to do: make Vitest usage a purely user land feature for which we make sure to have proper Wasp support.

EranGrin:
<img width="368" height="100" alt="Image" src="https://github.com/user-attachments/assets/92ded01c-3679-4811-8004-d6dd1b2b9625" />

I get this massage from my IDE

Martinsos:
[...]

PR #3571, opened 2025-12-31
Move Tailwind to userland + update to Tailwind 4

cprecioso:
- Resolves #2483

## Description

This PR does two main things:

- Removes all special handling of Tailwind in the compiler
- Updates our support to TW v4+ (docs, starters, examples)

Both changes are as expected, no weird things are needed, and everything is largely automatic as part of following the upgrade guide:

- We outright remove every Tailwind-specific handling from AppSpec and Generator
- Tailwind is upgraded to v4
  - Changed from PostCSS installation to Vite installation, as recommended by their guide 
    > "If you're using Vite, we recommend migrating from the PostCSS plugin to our new dedicated Vite plugin for improved performance and the best developer experience"
- CSS component libraries that build upon Tailwind are also upgraded (in example apps, Hero UI and Flowbite).

The only caveat that differentiates a Wasp TW app from a regular TW app is that we need to add a `source` parameter to the Tailwind import:

```diff
 // src/Main.css
-@import "tailwindcss";
+@import "tailwindcss" source(".");
```

This is done because Vite is called from `web-app` so we need to tell TW explicitly to find the source files in the user's project.

---

This is a big PR, and we can't do the changes in it separately as they are mutually dependent on each other. However, the types of changes are quite scoped depending on the filetype so it is easy to dismiss large chunks of this PR:

- Changes to classes in JS/TS files are all the result of running the TW upgrade CLI and the formatter with the updated TW plugin. (Exceptions are marked with an explanatory comment)
- Changes to `package.json` are version upgrades
- In every example/starter app:
  - Removed the postcss and tw config
  - Put the tw config into the main CSS file
  - Added the tw plugin in the vite config
- In the compiler, I remove any specific Tailwind support, and change the `starter` code so that it doesn't write a TW config file, but writes a Vite config with the TW plugin.

I checked that every app and starter works well and hasn't changed styles.

## Type of change

- [ ] **🔧 Just code/docs improvement** 
- [ ] **🐞 Bug fix** 
- [x] **🚀 New/improved feature** 
- [x] **💥 Breaking change** 

## Checklist

- [x] I tested my change in a Wasp app to verify that it works as intended.

- 🧪 Tests and apps:

  - [ ] I added **unit tests** for my change. 
  - [ ] _(if you fixed a bug)_ I added a **regression test** for the bug I fixed. 
  - [ ] _(if you added/updated a feature)_ I added/updated **e2e tests** in `examples/kitchen-sink/e2e-tests`.
[...]

PR #3580, opened 2026-01-07
Upgrade Vitest and related testing deps

cprecioso:
## Description

- Blocking #3571 
- Resolves https://github.com/wasp-lang/wasp/issues/2319

We've been using Vitest 1 in Wasp Testing, which only supports up to Vite 4; but we're using Vite 7. That was incorrect.

One of the symptoms is that `defineConfig` from `vitest` is not up-to-date with the one from `vite`, so the Tailwind upgrade at #3571 was blocked due to incompatible plugin types.

## Type of change

- [x] **🐞 Bug fix** 
- [x] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/web/docs/migration-guides/migrate-from-0-20-to-0-21.md
+++ b/web/docs/migration-guides/migrate-from-0-20-to-0-21.md
@@ -40,15 +40,11 @@ app MyApp {
 
 **If you don't have test files in your project, you can skip this step.**
 
-We upgraded our testing support from Vitest v1 to Vitest v4. We expect most tests to work without any changes, but if you have custom test setups or configurations, please upgrade your tests by following the migration guides from Vitest, in order:
+We upgraded our testing support from Vitest v1 to Vitest v4. Most of the breaking changes are related to internal configuration, edge cases, or very advanced usage; so we recommend **first to try running your tests after bumping the Wasp version**, and only read through the migration guides if you encounter issues:
 
-:::note
-Most of the breaking changes are related to internal configuration, edge cases, and very advanced usage, so we recommend first to try running your tests after bumping the Wasp version, and only read through the migration guides if you encounter issues.
-:::
-
-1. Vitest v1 to v2: https://v3.vitest.dev/guide/migration.html#vitest-2
-2. Vitest v2 to v3: https://v3.vitest.dev/guide/migration.html#vitest-3
-3. Vitest v3 to v4: https://vitest.dev/guide/migration.html#vitest-4
+1. [Migration guide from Vitest v1 to v2](https://v3.vitest.dev/guide/migration.html#vitest-2)
+2. [Migration guide from Vitest v2 to v3](https://v3.vitest.dev/guide/migration.html#vitest-3)
+3. [Migration guide from Vitest v3 to v4](https://vitest.dev/guide/migration.html#vitest-4)
 
 ### 3. Enjoy your updated Wasp app
 


examples/kitchen-sink/src/features/operations/components/Todo.test.tsx
```diff
@@ -69,6 +70,7 @@ test("handles multiple mock data sources", async () => {
   mockQuery(getMe, mockUser);
   mockQuery(getDate, new Date());
   mockQuery(getTasks, mockTasks);
+  mockQuery(getNumTasks, mockTasks.length);
```
cprecioso, line 73: This mock was missing

waspc/data/Cli/starters/skeleton/src/vite-env.d.ts
```diff
@@ -1,7 +1,3 @@
 /// <reference types="vite/client" />
-
-// This is needed to properly support Vitest testing with jest-dom matchers.
-// Types for jest-dom are not recognized automatically and Typescript complains
-// about missing types e.g. when using `toBeInTheDocument` and other matchers.
-// Reference: https://github.com/testing-library/jest-dom/issues/546#issuecomment-1889884843
-import "@testing-library/jest-dom";
```
cprecioso, line 7: We don't really need this note, this is just the normal way to install this library as per its documentation

web/docs/migration-guides/migrate-from-0-20-to-0-21.md
```diff
@@ -0,0 +1,55 @@
...
+}
+```
+
+### 2. Upgrade Vitest tests to v4
+
+**If you don't have test files in your project, you can skip this step.**
+
+We upgraded our testing support from Vitest v1 to Vitest v4. We expect most tests to work without any changes, but if you have custom test setups or configurations, please upgrade your tests by following the migration guides from Vitest, in order:
+
+:::note
```
infomiho: Great note!

waspc/data/Cli/starters/skeleton/src/vite-env.d.ts
sodic: What about the new thing `vitest/globals`. Is that needed for Vitest 4?

Also, will we be able to kick all of this out once we move Vitest to userland?

web/docs/migration-guides/migrate-from-0-20-to-0-21.md
```diff
@@ -0,0 +1,55 @@
...
+Most of the breaking changes are related to internal configuration, edge cases, and very advanced usage, so we recommend first to try running your tests after bumping the Wasp version, and only read through the migration guides if you encounter issues.
+:::
```
sodic: I would move this note out of the callout and into the main text:
- The message is very similar to what we said in the paragraph above, so we should be able to easily merge them.
- It is a part of the main flow, not an extra note.
- Callouts kinda break the flow (especially when inserted right before a numbered list).

waspc/data/Cli/starters/skeleton/src/vite-env.d.ts
cprecioso: > What about the new thing vitest/globals. Is that needed for Vitest 4?

Not v4-specific, but we should have added it before, since it is recommended when you enable the `globals: true` option (required for `@testing-library/react`): https://vitest.dev/config/globals.html#globals

> Also, will we be able to kick all of this out once we move Vitest to userland?

Mmmm tbh I haven't thought about it, I think it will depend on our specific implementation. But we're not better off or worse off with this change so I'm not especially worried.

web/docs/migration-guides/migrate-from-0-20-to-0-21.md
cprecioso: Yep, I evaluated that already, but I think that people will see a bullet list and skip directly to it, that's why I wanted to add an attention-grabbing callout. But let me re-construct this, I think I can do away with it.

PR #3592, opened 2026-01-09
Adds cookie consent banner

infomiho:
## Description

Adds a cookie consent banner using `react-cookie-consent`. I tried using the `vanilla-cookieconsent` package that Open SaaS uses but it wouldn't properly render on the `/docs` page. 

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/web/docusaurus.config.ts
+++ b/web/docusaurus.config.ts
@@ -288,7 +288,7 @@ function mapToDocusaurusScripts(
     // so we use a custom type on the script _not to load it_ automatically.
     // Later, if cookies are accepted, the script will be loaded by the
     // `src/components/CookieConsentBanner.tsx` component.
-    type: requiresConsent ? SCRIPT_WITH_CONSENT_TYPE : undefined,
+    ...(requiresConsent ? { type: SCRIPT_WITH_CONSENT_TYPE } : {}),
   }));
 }
 
--- a/web/src/components/CookieConsentBanner.tsx
+++ b/web/src/components/CookieConsentBanner.tsx
@@ -38,15 +38,11 @@ export default function CookieConsentBanner() {
 function loadScripts() {
   document
     .querySelectorAll(`script[type="${SCRIPT_WITH_CONSENT_TYPE}"]`)
-    .forEach((oldScript) => {
-      const newScript = document.createElement("script");
-      Array.from(oldScript.attributes).forEach((attr) => {
-        if (attr.name === "type") {
-          newScript.setAttribute("type", "text/javascript");
-        } else {
-          newScript.setAttribute(attr.name, attr.value);
-        }
-      });
-      oldScript.parentNode?.replaceChild(newScript, oldScript);
+    .forEach((el) => {
+      // We need to reattach the element so that the browser picks up the new `type`.
+      const parent = el.parentNode;
+      parent.removeChild(el);
+      el.setAttribute("type", "text/javascript");
+      parent.appendChild(el);
     });
 }


web/docusaurus.config.ts
cprecioso: Oh I didn't make the connection between Plausible and this URL at first. Can you add a explicit note that this is Plausible being proxied?

web/src/components/CookieConsentBanner.module.css
cprecioso, line 1: personal preference: i might have had this css as `src/components/CookieConsentBanner/styles.css` since it is in that component's scope, and should only be loaded when that component is loaded (even if that component is loaded everywhere).
It's a semantic thing so consider it, but RAW

web/src/components/CookieConsentBanner.tsx
```diff
@@ -0,0 +1,62 @@
+import "@site/src/css/cookie-consent.css";
+import { useEffect } from "react";
+import CookieConsent, { Cookies } from "react-cookie-consent";
+
+const consentCookieName = "wasp-cookies-consent";
```
cprecioso: I think we use `CAMEL_CASE` names for hardcoded constants

web/src/components/CookieConsentBanner.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export default function CookieConsentBanner() {
+  const handleAccept = () => {
+    loadScripts();
+  };
```
cprecioso: nitpick, I'd do away with this indirection

web/src/components/CookieConsentBanner.module.css
cprecioso: moreover, being that we give the class names explicitly to the component, we can use CSS modules there https://docusaurus.io/docs/styling-layout#css-modules so our naming can be less verbose

web/src/components/CookieConsentBanner.tsx
cprecioso, line 1: I see we lost the `common`/`devOnly`/`prodOnly` thing here. Was that on purpose?

web/docusaurus.config.ts
cprecioso, line 247: Should we add a note here about the Cookie Consent component?

infomiho:
@cprecioso great suggestion on making it declarative and the `type` trick worked great! Ready for another look.

web/src/components/CookieConsentBanner.module.css
infomiho: Moved to a CSS module 👍

web/src/components/CookieConsentBanner.tsx
infomiho: Not at all, thank you for raising that issue 👍

web/docusaurus.config.ts
```diff
@@ -238,29 +239,57 @@ const config: Config = {
...
-    // Using Cloudflare Workers to proxy the analytics script
+  const prodOnlyScripts: ScriptWithConsent[] = [
     {
+      // We are using Cloudflare Workers to proxy the Plausible script.
       src: "/waspara/wasp/script.js",
       defer: true,
       "data-domain": "wasp.sh",
       "data-api": "/waspara/wasp/event",
+      // Plausible doesn't use cookies, so it can be loaded right away.
+      requiresConsent: false,
```
cprecioso, line 265: nice comments, thanks for adding them!

web/docusaurus.config.ts
```diff
@@ -238,29 +239,57 @@ const config: Config = {
...
+function mapToDocusaurusScripts(
+  scripts: ScriptWithConsent[],
+): DocusaurusScript[] {
+  return scripts.map(({ requiresConsent, ...rest }) => ({
+    ...rest,
+    // Scripts requiring consent shouldn't be loaded until the cookies are accepted,
+    // so we use a custom type on the script _not to load it_ automatically.
+    // Later, if cookies are accepted, the script will be loaded by the
+    // `src/components/CookieConsentBanner.tsx` component.
+    type: requiresConsent ? SCRIPT_WITH_CONSENT_TYPE : undefined,
```
cprecioso: This `undefined` is ending up in the final HTML, so these scripts are not loading:

```html
<script src="/js/fix-multiple-trailing-slashes.js" type="undefined"></script>
<script src="/waspara/wasp/script.js" defer="defer" data-domain="wasp.sh" data-api="/waspara/wasp/event" type="undefined"></script>
```

I guess you should explicitly say `text/javascript`.

web/src/components/CookieConsentBanner.tsx
```diff
@@ -0,0 +1,52 @@
...
+      const newScript = document.createElement("script");
+      Array.from(oldScript.attributes).forEach((attr) => {
+        if (attr.name === "type") {
+          newScript.setAttribute("type", "text/javascript");
+        } else {
+          newScript.setAttribute(attr.name, attr.value);
+        }
+      });
+      oldScript.parentNode?.replaceChild(newScript, oldScript);
+    });
```
cprecioso: This should be enough (tested in the JS Console), and is simpler:


```suggestion
    .forEach((el) => {
      // We need to reattach the element so that the browser picks up the new `type`
      const parent = el.parentNode;
      parent.removeChild(el);
      el.setAttribute("type", "text/javascript");
      parent.appendChild(el);
    });
```

infomiho:

PR #3593, opened 2026-01-09
Add REO.dev script 

infomiho:
## Description

- Add REO.dev script to the Wasp web. 
- Refactors the loading code a bit.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/web/src/components/CookieConsentBanner.tsx
+++ b/web/src/components/CookieConsentBanner.tsx
@@ -4,6 +4,21 @@ import CookieConsent, { Cookies } from "react-cookie-consent";
 
 const consentCookieName = "wasp-cookies-consent";
 
+const scriptsToLoadAfterConsent: Script[] = [
+  {
+    name: "PostHog",
+    load: () => loadScript("/scripts/posthog.js"),
+  },
+  {
+    name: "REO.dev",
+    async load() {
+      const REO_CLIENT_ID = "96f303a127451b4";
+      await loadScript(`https://static.reo.dev/${REO_CLIENT_ID}/reo.js`);
+      window.Reo.init({ clientID: REO_CLIENT_ID });
+    },
+  },
+];
+
 export default function CookieConsentBanner() {
   const handleAccept = () => {
     loadScripts();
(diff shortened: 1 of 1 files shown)


web/src/components/CookieConsentBanner.tsx
cprecioso, line 79: I'd make a explicit note that we're not `await`-ing the loads here on purpose so that 1. they all load concurrently and 2. an adblocker blocking one of them doesn't completely halt execution.

web/src/components/CookieConsentBanner.tsx
```diff
@@ -40,20 +55,24 @@ export default function CookieConsentBanner() {
 
 type Script = {
   name: string;
-  load: () => void;
+  load: () => Promise<void>;
 };
```
cprecioso, line 59: nitpick: should this type be moved to the top,  with its usage?

web/src/components/CookieConsentBanner.tsx
```diff
@@ -4,6 +4,21 @@ import CookieConsent, { Cookies } from "react-cookie-consent";
...
+  {
+    name: "PostHog",
+    load: () => loadScript("/scripts/posthog.js"),
+  },
+  {
+    name: "REO.dev",
+    async load() {
+      const REO_CLIENT_ID = "96f303a127451b4";
+      await loadScript(`https://static.reo.dev/${REO_CLIENT_ID}/reo.js`);
+      window.Reo.init({ clientID: REO_CLIENT_ID });
```
cprecioso, line 17: nitpick: It's strange for me that we need the client id both in the script url and in the initialization. I could not get clarification for it in the Reo docs since to get the script you need a login. I assume this is their instructions, but I'd double-check if possible.

web/src/components/CookieConsentBanner.tsx
cprecioso: Maybe a `Promise.all` would be more self-explanatory?

web/src/components/CookieConsentBanner.tsx
```diff
@@ -40,20 +55,24 @@ export default function CookieConsentBanner() {
...
+    Reo?: any;
+  }
+}
+
+async function loadScript(src: string): Promise<void> {
+  return new Promise((resolve) => {
+    const script = document.createElement("script");
+    script.src = src;
+    script.defer = true;
+    script.onload = () => resolve();
```
cprecioso, line 72: heads up: this might never resolve if the script trips the adblocker. We could add an `onerror` too (although it wouldn't do anything atm so RAW).

infomiho:
I'll redo the REO integration since the new system doesn't work well with the current REO script.

Issue #2631, opened 2025-04-07
Create a "Guides" section in the docs


infomiho:
Right now, we have a collection of different useful (and specific) gists that we keep track of in the learning materials repo: https://github.com/wasp-lang/learning-materials?tab=readme-ov-file

We'd like to include links to those gists in our docs somehow. Since they are quite specific, concrete examples, it makes sense to maybe include them in a "Guides" section that contains recipe styles information.

Martinsos:
What I liked so far the best is the idea of having "guides" in our docs, and we make it clear that these guides are an addition to docs for specific worfklows and might be dated.
And for the start, it could just be a link (in the docusuraus sidebar) to a gh repo with guides. that is already great.

Or we can make them part of docs immeidately.
But I would make sure they always have obvious dates on when they were updated last time, and it should be clear in them for what version of what were they written with.

I guess the question is, do we at some point want to have "guides" that we do want to keep updating it regularly? And why do we even want to start with these "not updated regularly" guides, why not immeidatelly go for the "official" ones? I don't know hm, I think I wanted to enable us to create them without having to worry about the maintainenace burden. Discussion here might be intersting, what is our policy on this, do we have two types of guides, ones that we keep up to date same as docs vs ones we don't have to be so strict with, is it different naming then, ... .

sodic:
I'll bump this because I think we should do it sooner rather than later. People have trouble finding those gists all the time. When they finally do find them, they also find them (no pun intended) exceptionally helpful.

Sources: community captaining, 100% of my friends who tried Wasp went through this.

Martinsos:
> I'll bump this because I think we should do it sooner rather than later. People have trouble finding those gists all the time. When they finally do find them, they also find them (no pun intended) exceptionally helpful.
> 
> Sources: community captaining, 100% of my friends who tried Wasp went through this.

User interviews also confirm this. Self-hosting gist is a big one. Very hard to find, immensely useful.

cprecioso:
POC at #3589 (preview: https://cprecioso-add-docs-guides.wasp-docs-on-main.pages.dev/docs/guides)
RFC at https://www.notion.so/wasp-lang/RFC-Adding-3rd-party-how-tos-to-the-docs-2e318a74854c805caef2fd9dcfe10ffd?v=02875e9c63c944d4b856d67e4a9db2da&source=copy_link

cprecioso:
Guides section is now created (#3597), we'll migrate them in a new task #3644

PR #3596, opened 2026-01-13
Docs: Add Version notice

cprecioso:
- Part of #2631

## Description

Adds a notice to docs that declare a `checked_version` metadata

https://github.com/wasp-lang/wasp/blob/f175ae04cc0d0b3203d838b294a292b9b7cd08f5/web/docs/guides/libraries/shadcn.md#L3-L5

image

- This will only show up on documents in `docs/` that set up the `checked_version` metadata.
- In the PoC this showed a date and turned into a warning if the date was older than 3 months. We decided this shouldn't happen, so this functionality is removed.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/web/src/components/VersionNotice.tsx
+++ b/web/src/components/VersionNotice.tsx
@@ -13,28 +13,28 @@ import { useMemo } from "react";
  */
 export type Version = string | number | Date;
 
-export interface CheckedVersions {
+export interface LastCheckedWithVersions {
   [name: string]: Version;
 }
 
-export default function VersionNotice({
-  checkedVersions,
+export default function LastCheckedWithVersionsNotice({
+  lastCheckedWithVersions,
 }: {
-  checkedVersions: CheckedVersions;
+  lastCheckedWithVersions: LastCheckedWithVersions;
 }) {
-  const checkedWithString = useMemo(
+  const lastCheckedWithString = useMemo(
     () =>
       listFormatter.format(
-        Object.entries(checkedVersions).map(([key, value]) =>
+        Object.entries(lastCheckedWithVersions).map(([key, value]) =>
           formatVersion(key, value),
         ),
       ),
-    [checkedVersions],
+    [lastCheckedWithVersions],
   );
 
   return (
     <Admonition type="note">
-      <p className="font-bold">Last checked with {checkedWithString}.</p>
+      <p className="font-bold">Last checked with {lastCheckedWithString}.</p>
       This guide depends on external libraries or services, so it may become
       outdated over time. We do our best to keep it up to date, but make sure to
       check their documentation for any changes.
--- a/web/src/theme/DocItem/Content/index.tsx
+++ b/web/src/theme/DocItem/Content/index.tsx
@@ -1,7 +1,7 @@
 import { useDoc } from "@docusaurus/plugin-content-docs/client";
 import type { WrapperProps } from "@docusaurus/types";
-import VersionNotice, {
-  CheckedVersions,
+import LastCheckedWithVersionsNotice, {
+  LastCheckedWithVersions,
 } from "@site/src/components/VersionNotice";
 import Content from "@theme-original/DocItem/Content";
 import type ContentType from "@theme/DocItem/Content";
@@ -10,7 +10,7 @@ import type { ReactNode } from "react";
 type Props = WrapperProps<typeof ContentType>;
 
 interface FrontMatter {
-  checked_versions?: CheckedVersions;
+  last_checked_with_versions?: LastCheckedWithVersions;
 }
 
 export default function ContentWrapper(props: Props): ReactNode {
@@ -18,9 +18,11 @@ export default function ContentWrapper(props: Props): ReactNode {
 
   return (
     <>
-      {frontMatter.checked_versions && (
+      {frontMatter.last_checked_with_versions && (
         <div className="mt-4">
-          <VersionNotice checkedVersions={frontMatter.checked_versions} />
+          <LastCheckedWithVersionsNotice
+            lastCheckedWithVersions={frontMatter.last_checked_with_versions}
+          />
         </div>
       )}
 


web/src/theme/DocItem/Content/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+import Content from "@theme-original/DocItem/Content";
+import type ContentType from "@theme/DocItem/Content";
+import type { ReactNode } from "react";
+import styles from "./styles.module.css";
+
+type Props = WrapperProps<typeof ContentType>;
+
+interface FrontMatter {
+  checked_versions?: CheckedVersions;
+}
```
cprecioso, line 14: We don't do any specific checking of the correct types, but we control the docs so I don't think we need to.

web/src/lib/formatters.ts
```diff
@@ -0,0 +1,13 @@
+// All of our docs are written in US English, so our formatters should also use it.
+export const LOCALE = "en-US";
+
+export const listFormatter = new Intl.ListFormat(LOCALE, {
```
cprecioso, line 17: To get in front of the complaints: `Intl.ListFormat` is Baseline Widely Available since 2021. 😁

web/src/components/VersionNotice/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export default function VersionNotice({
+  checkedVersions,
+}: {
+  checkedVersions: CheckedVersions;
+}) {
+  const checkedWithString = useMemo(
+    () =>
+      listFormatter.format(
```
cprecioso: `Intl.ListFormatter.format`:
- `["foo"]` -> `"foo"`
- `["foo", "bar"]` -> `"foo and bar"`
- `["foo", "bar", "baz"]` -> `"foo, bar, and baz"`
- `["foo", "bar", "baz", "qux"]` -> `"foo, bar, baz, and qux"`
- etc

web/src/components/VersionNotice/styles.module.css
```diff
@@ -0,0 +1,3 @@
+.lastCheckedLine {
```
FranjoMindek: How come we don't use `tailwindcss`? I think we have it on `web`.
I might've forgot the reasons why, if we have them.

web/src/components/VersionNotice/index.tsx
```diff
@@ -0,0 +1,47 @@
+import { dateFormatter, listFormatter } from "@site/src/lib/formatters";
+import Admonition from "@theme/Admonition";
+import { useMemo } from "react";
+import styles from "./styles.module.css";
+
+// If a string looks like a number or date, YAML will parse it.
```
FranjoMindek: Not sure if I fully understand the comment.
So if the `string` in YAML is a number, YAML parser will return a `number` type?

Could we have an example for each of the possible `Version` types?
Currently, I'm not fully sure for all examples.
`number` as in, when something only has a single version number?

If we turn this comment into jsdocs it will also popup on intellisense.

web/src/components/VersionNotice/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+// Dates are useful for unversioned dependencies, like Shadcn.
+type Version = string | number | Date;
+
+export interface CheckedVersions {
```
FranjoMindek: Okay, took me a bit exploring to fully understand the name.
Maybe, `LastCheckedWithVersions`?

Same for the fields in `.md` files:
`checked_versions` -> `last_checked_with_versions`?

web/src/components/VersionNotice/index.tsx
FranjoMindek: In a way, I wouldn't know it does this without the comment.
If `listFormatter` has a valid jsdocs which explains it I think this is fine.
If not, I would also explain it in code.

web/src/lib/formatters.ts
FranjoMindek: TIL

web/src/components/VersionNotice/index.tsx
FranjoMindek: Oh okay, I see we create it down there with `conjunction` type.
It won't be immediately transparent, but with little search you'll know what it does.

Your decision if you want to add jsdocs to `listFormatter`.

FranjoMindek (approved):
Nothing blocking (only concerned about `tailwindcss`), really nice addition.

Some suggestions to make code a bit more understandable on first off basis.

web/src/components/VersionNotice/styles.module.css
cprecioso: no specific reason, i'm just more comfortable with real css 😅
i'll ask my friend claude to write me the TW for it

web/src/components/VersionNotice/index.tsx
cprecioso:
```yaml
Wasp: 0.20.1 # parses as a string
Chrome: 143 # parses as a number
Shadcn: 2026-01-14 # parses as a Date
```

I added comments for it

cprecioso:
@FranjoMindek ready for re-review

web/src/components/VersionNotice.tsx
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * The YAML parser in Docusaurus automatically tries to parse any number-like or
+ * date-like strings into numbers or Date objects, so we need to account for that.
+ *
+ * Dates are especially useful for unversioned dependencies, like Shadcn.
+ */
+export type Version = string | number | Date;
+
+export interface LastCheckedWithVersions {
+  [name: string]: Version;
```
FranjoMindek, line 17: This `name` can also be `dependency` if you wish.
RAW.

PR #3595, opened 2026-01-13
Docs: Add comment system

cprecioso:
- Part of #2631 

## Description

Adds a comment system with [Giscus](https://giscus.app/).

Giscus acts as "just" a nicely embeddable frontend for GitHub Discussions. When a user first comments on a page, Giscus creates a discussion for that page, and through GitHub's API posts the comment to the discussion, from the user. Later comments are posted to that same discussion, as well as replies to top-level items. This forum style is nicely suited for people creating threads; and allows us to just use GitHub's moderation tools for comments.

Details:
- Pages have to explicitly opt into the comment system through setting `comments: true` in their frontmatter.
- It uses the `Doc` context from the Docusaurus Docs plugin, so it only works in the `web/docs` subolder (not the landing page nor the blog).

Example (not in this PR but in a future one):
- Enabling the comments in the document:
	https://github.com/wasp-lang/wasp/blob/bac814871916a02402e078b01f5abc83e883602c/web/docs/guides/configuration/cors-multiple-domains.md?plain=1#L3
- Live comments on the page (scroll to bottom):
	https://push-mqztwroqlrkz.wasp-docs-on-main.pages.dev/docs/guides/configuration/cors-multiple-domains
- The underlying GitHub Discussion:
	https://github.com/wasp-lang/wasp.sh-comments/discussions/3

## Type of change

- [x] **🔧 Just code/docs improvement** 
- [ ] **🐞 Bug fix** 
- [ ] **🚀 New/improved feature** 
- [ ] **💥 Breaking change** 

## Checklist

- [ ] I tested my change in a Wasp app to verify that it works as intended.

- 🧪 Tests and apps:

  - [ ] I added **unit tests** for my change. 
  - [ ] _(if you fixed a bug)_ I added a **regression test** for the bug I fixed. 
  - [ ] _(if you added/updated a feature)_ I added/updated **e2e tests** in `examples/kitchen-sink/e2e-tests`.
  - [ ] _(if you added/updated a feature)_ I updated the **starter templates** in `waspc/data/Cli/templates`, as needed.
  - [ ] _(if you added/updated a feature)_ I updated the **example apps** in `examples/`, as needed.
    - [ ] _(if you updated `examples/tutorials`)_ I updated the tutorial in the docs (and vice versa).

- 📜 Documentation:

  - [ ] _(if you added/updated a feature)_ I **added/updated the documentation** in `web/docs/`.

- 🆕 Changelog: _(if change is more than just code/docs improvement)_
[...]

PR #3597, opened 2026-01-13
Docs: Add guides section

cprecioso:
## Description

- Part of https://github.com/wasp-lang/wasp/issues/2631
- On top of https://github.com/wasp-lang/wasp/pull/3595
- On top of https://github.com/wasp-lang/wasp/pull/3596

This PR adds the "Guides" section to our website. 

Screen Recording 2026-01-09 at 11 18 47

Some notes [from the RFC](https://www.notion.so/wasp-lang/RFC-Adding-3rd-party-how-tos-to-the-docs-2e318a74854c805caef2fd9dcfe10ffd?source=copy_link):

> - Create a “Guides” section in our website, conceptually separate from “Docs”
>     - Maintains the mental separation, lets us set different maintenance expectation levels for this section vs. the docs section.
> - Letting it participate in the versioning of docs
>     - This means that when we do a release, the current state of guides gets archived in `versioned_docs`. The version selector changes between these archives.
>     - e.g. if we modify the deployment strategy from Wasp 0.20 to Wasp 0.21, the guides should also change whether you’re in one or the other.
>     - This feels like additional maintenance burden but I think with the “last checked” dates and everything, we are allowing ourselves to play more loosely with it so it doesn’t create this burden.
>     - Implementation note: Docusaurus doesn’t let us show the “Guides” menu item only on some versions, so older versions that didn’t have a Guides section will just show a “Guides not available” document.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/web/docs/guides/index.md
+++ b/web/docs/guides/index.md
@@ -7,6 +7,6 @@ title: About these guides
 
 The guides in this section will walk you through common scenarios you'll encounter when building real-world apps with Wasp. They're more hands-on and to-the-point than the main docs. Pick the one that matches what you're trying to do and follow along!
 
-Since many guides involve third-party libraries or services, some details may drift over time. Where appropriate, each guide shows when it was last updated and which versions it was tested with. If something doesn't work as expected, check the official docs for the tool in question, and feel free to let us know so we can update it.
+Since many guides involve third-party libraries or services, some details may drift over time. Where appropriate, each guide shows when it was last updated and which versions it was tested with. If something doesn't work as expected, check the official docs for the tool in question, and feel free to let us know in the comments below each page, so we can update it.
 
 Looking for a guide we don't have yet? [Open an issue](https://github.com/wasp-lang/wasp/issues) or drop by [our Discord](https://discord.gg/rzdnErX). We're always looking for new ideas.


web/docusaurus.config.ts
```diff
@@ -96,7 +103,7 @@ const config: Config = {
       ],
     },
     prism: {
-      additionalLanguages: ["shell-session", "haskell", "markdown"],
+      additionalLanguages: ["shell-session", "haskell", "markdown", "diff"],
```
cprecioso, line 107: A lot of our guides use `diff` snippets, so I add syntax highlighting for them here.

web/sidebars.ts
```diff
@@ -220,6 +220,12 @@ const sidebars: SidebarsConfig = {
       items: ["contributing", "telemetry", "vision", "contact"],
     },
   ],
+  guides: [
+    {
+      type: "autogenerated",
+      dirName: "guides",
+    },
+  ],
```
cprecioso, line 228: Q: Why is this "autogenerated"?
A: This category of docs is not trying to tell a story but just a drawer full of guides that you refer to when you need. Thus, there's no need for a specific order for them. If we use "autogenerated", we don't need to add the guides here everytime we create a new one, lowering friction.

web/docusaurus.config.ts
```diff
@@ -63,11 +63,18 @@ const config: Config = {
       },
       items: [
         {
-          type: "docsVersion",
+          type: "docSidebar",
```
cprecioso, line 67: This is relating to the links  in the  top navigation bar, and how they are colored differently when you are in their section of the website ("active").

Previously: `docsVersion` would include everything under the `docs/` folder, regardless of which sidebar it's located in. Both the "Docs" and "Guides" links would show as active when reading a guide, even if we only want the "Guides" link to be active.

Now: `docSidebar` only refers strictly to documents configured inside the docs sidebar. So, this category doesn't include guides, which are on their own alternative sidebar. Active links work as expected. No functionality change otherwise.

web/versioned_sidebars/version-0.11.8-sidebars.json
```diff
@@ -125,5 +125,6 @@
       "collapsible": false,
       "items": ["contributing", "telemetry", "vision", "contact"]
     }
-  ]
+  ],
+  "guides": [{ "type": "autogenerated", "dirName": "guides" }]
```
cprecioso, line 129: Docusaurus doesn't allow us to create a "guides" navbar section only for the current version, all versions need to have this guides sidebar.

For that reason, I add it to each version, only pointing to a document telling you that they are not available.

web/docs/guides/index.md
```diff
@@ -0,0 +1,12 @@
+---
+sidebar_position: 1
```
cprecioso, line 2: We set the position here so, even though the sidebar is autogenerated, this doc will always show up first.

web/docs/guides/index.md
```diff
@@ -0,0 +1,12 @@
...
+title: About these guides
+---
+
+# Guides
+
+The guides in this section will walk you through common scenarios you'll encounter when building real-world apps with Wasp. They're more hands-on and to-the-point than the main docs. Pick the one that matches what you're trying to do and follow along!
+
+Since many guides involve third-party libraries or services, some details may drift over time. Where appropriate, each guide shows when it was last updated and which versions it was tested with. If something doesn't work as expected, check the official docs for the tool in question, and feel free to let us know so we can update it.
```
infomiho: I'd maybe mention the comment functionality here "comment on it" or smth like that.

web/versioned_sidebars/version-0.11.8-sidebars.json
infomiho: > Docusaurus doesn't allow us to create a "guides" navbar section only for the current version

That's annoying that whenever a new section is added you have to back port it 🫠 

I know we can swizzle the Navbar probably, but I think your solution is enough. It would be maybe nice we added at the end of the message "No guides for this version... check the guides for the latest version (link)"

infomiho: Image

I guess no need for a link, it's already on the top 👍

cprecioso:
> I'd probably shout out the "guides" somewhere in the docs e.g. docs CORS section mentions the CORS guide for multiple domains.

Yep I agree, but I was planning on doing that as we migrated gists to it

PR #3600, opened 2026-01-13
Add integrations guides

cprecioso:
@infomiho says:

All of the guides have been:
- checked against the latest Wasp version
- tested in a Wasp app
- read by a human and had their text and front matter updated


web/docs/guides/integrations/custom-oauth.md
```diff
@@ -0,0 +1,244 @@
...
+  entities: []
+}
+// highlight-end
+
+// highlight-start
+api authWithSpotifyCallback {
+  httpRoute: (GET, "/auth/spotify/callback"),
+  fn: import { authWithSpotifyCallback } from "@src/auth",
+  entities: []
+}
```
sodic, line 58: How does the user know they need this? I'm guessing the provider tells them. If so, I'd expect our docs to say it.

web/docs/guides/integrations/custom-oauth.md
```diff
@@ -0,0 +1,244 @@
...
+model User {
+  id             String @id @default(cuid())
+  name           String
+  profilePicture String
+}
+```
+
+### 4. Implement the OAuth handlers
+
+Create the authentication logic:
```
sodic: Same question, does the user get this code from the provider? Or do we provide it here?

That's a lot of code that seems arbitrary, I'd expect more context.

web/docs/guides/integrations/custom-oauth.md
```diff
@@ -0,0 +1,244 @@
...
+  });
+  return spotifyUserSchema.parse(await response.json());
+}
+```
+
+:::note
+The `tokenStore` and `getRedirectUriForOneTimeCode` are internal Wasp APIs that may change in future versions. This guide relies on them because there is currently no public API for implementing fully custom OAuth flows.
+:::
+
+### 5. Create the login page
```
sodic, line 220: It feels like this section should also reference _something_ (maybe the official login docs), just to make it obvious where we're pulling this code from.

web/docs/guides/integrations/custom-oauth.md
```diff
@@ -0,0 +1,244 @@
...
+    </div>
+  );
+};
+```
+
+## Configuring Your OAuth Provider
+
+The Arctic library supports many providers. Check the [Arctic documentation](https://v1.arcticjs.dev/) for the full list and their specific setup requirements.
+
+Each provider follows the same pattern - just replace the Spotify-specific code with the appropriate provider from Arctic.
```
sodic: Hm, I would not know what to do after reading this:
- What does it even mean to "configure" an auth provider?
- Where do I do this?

It seems this section should have an example.

sodic (approved):

web/docs/guides/integrations/custom-oauth.md
infomiho: This is how you do OAuth integrations 😄 you set up a callback URL in your app and set it in the provider settings. I've added this to the content.

web/docs/guides/integrations/custom-oauth.md
infomiho: I've added more context + split it into two files so it's easier to see what's "pure Spotify logic" and what's Wasp auth logic.

web/docs/guides/integrations/custom-oauth.md
infomiho: I've added links to our docs 👍

web/docs/guides/integrations/custom-oauth.md
infomiho: Added some example code 👍

PR #3603, opened 2026-01-13
Add deployment guides

cprecioso:
This PR is only a starting-off point, and needs some manual involvement.

The current content is an LLM-generated conversion of the guides from wasp-lang/learning-materials, to the new Guides section in our website.

What to do next:
- Check that no content was skipped or missed from the original guide in wasp-lang/learning-materials.
- Check the content and format to make sure it is correct, consistent, and looks good.
- Check that it works with the latest version of Wasp.
- Add a property to the frontmatter indicating which versions you have verified it with:
    ```md
    ---
    last_checked_with_versions:
      Wasp: 0.21.0
      Tailwind: 4
      Shadcn: 2025-12-31
    ---
    ```
- Find any references to the old guides throughout the existing docs, and update them to point to the new guides.

You can work directly in this PR, and request a review when you're done.

### DIFF
--- a/web/docs/deployment/deployment-methods/self-hosted.md
+++ b/web/docs/deployment/deployment-methods/self-hosted.md
@@ -4,7 +4,6 @@ title: Self-Hosted
 
 import { ImgWithCaption } from '@site/blog/components/ImgWithCaption'
 import { GuideLink } from './GuideLink.tsx'
-import useBaseUrl from '@docusaurus/useBaseUrl';
 
 If you have your server or rent out a server, you can self-host your Wasp apps. Self-hosting your apps gives you full control over your apps and their data. It can be more cost-effective than a cloud provider since you can deploy multiple apps on a single server. However, you'll need to manage the server yourself, which can be time-consuming and require some technical knowledge.
 
@@ -70,7 +69,7 @@ On your server:
 
 Check out one of our step-by-step guides for more details:
 
-<GuideLink linkToGuide="/docs/guides/deployment/vps" title="Deploying Wasp with Docker on your server" description="Uses Ubuntu, Git, Caddy, Docker" />
+<GuideLink linkToGuide="../../guides/deployment/vps" title="Deploying Wasp with Docker on your server" description="Uses Ubuntu, Git, Caddy, Docker" />
 
 ### Coolify
 
@@ -93,7 +92,7 @@ Check out one of our step-by-step guides for more details:
 
 Check out one of our step-by-step guides for more details:
 
-<GuideLink linkToGuide="/docs/guides/deployment/coolify" title="Deploying Wasp with Coolify on your server" description="Uses Coolify, Github Actions, Github Container Registry" />
+<GuideLink linkToGuide="../../guides/deployment/coolify" title="Deploying Wasp with Coolify on your server" description="Uses Coolify, Github Actions, Github Container Registry" />
 
 ### CapRover
 
@@ -116,7 +115,7 @@ Check out one of our step-by-step guides for more details:
 
 Check out one of our step-by-step guides for more details:
 
-<GuideLink linkToGuide="/docs/guides/deployment/caprover" title="Deploying Wasp with Caprover on your server" description="Uses Caprover, Github Actions, Github Container Registry" />
+<GuideLink linkToGuide="../../guides/deployment/caprover" title="Deploying Wasp with Caprover on your server" description="Uses Caprover, Github Actions, Github Container Registry" />
 
 ## Database setup
 
--- a/web/docs/deployment/deployment-methods/wasp-deploy/ci-cd.md
+++ b/web/docs/deployment/deployment-methods/wasp-deploy/ci-cd.md
@@ -57,7 +57,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v6
         with:
-          node-version: '22'
+          node-version: "{minimumNodeJsVersion}"
 
       - name: Install Wasp
         # We pin the Wasp CLI version to avoid issues when a new Wasp version is released.
@@ -102,7 +102,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v6
         with:
-          node-version: '22'
+          node-version: "{minimumNodeJsVersion}"
 
       - name: Install Wasp
         # We pin the Wasp CLI version to avoid issues when a new Wasp version is released.
--- a/web/docs/guides/deployment/caprover.md
+++ b/web/docs/guides/deployment/caprover.md
@@ -2,6 +2,7 @@
 comments: true
 last_checked_with_versions:
   Wasp: 0.21.0
+  Caprover: 2026-01-30
 ---
 
 import { SecretGeneratorBlock } from "../../project/SecretGeneratorBlock";
--- a/web/docs/guides/deployment/vps.md
+++ b/web/docs/guides/deployment/vps.md
@@ -2,6 +2,8 @@
 comments: true
 last_checked_with_versions:
   Wasp: 0.21.0
+  Caddy: 2026-01-30
+  Ubuntu: 2026-01-30
 ---
 
 import { SecretGeneratorBlock } from "../../project/SecretGeneratorBlock";
--- a/web/src/remark/search-and-replace.ts
+++ b/web/src/remark/search-and-replace.ts
@@ -26,15 +26,10 @@ const plugin: Plugin<[], Root> = () => (tree) => {
   the text is searched and replaced.
 */
 function createReplacer(definitions: Record<string, string>) {
-  const compiled = Object.entries(definitions).map(([key, value]) => ({
-    regex: new RegExp(`\\{${key}\\}`, "g"),
-    value,
-  }));
-
   return {
     searchAndReplace(text: string): string {
-      for (const { regex, value } of compiled) {
-        text = text.replace(regex, value);
+      for (const [key, value] of Object.entries(definitions)) {
+        text = text.replaceAll(`{${key}}`, value);
       }
       return text;
     },
--- a/web/docs/deployment/deployment-methods/paas.md
+++ b/web/docs/deployment/deployment-methods/paas.md
@@ -541,7 +541,7 @@ Here’s an example configuration file to help you get started. This example wor
           id: setup-node
           uses: actions/setup-node@v5
           with:
-            node-version: '22'
+            node-version: "{minimumNodeJsVersion}"
 
         - name: Install Wasp
           run: npm i -g @wasp.sh/wasp-cli@{latestWaspVersion} # Change to your Wasp version
@@ -641,7 +641,7 @@ Here’s an example configuration file to help you get started. This example wor
           id: setup-node
           uses: actions/setup-node@v5
           with:
-            node-version: '22'
+            node-version: "{minimumNodeJsVersion}"
 
         - name: Install Wasp
           run: npm i -g @wasp.sh/wasp-cli@{latestWaspVersion} # Change to your Wasp version
--- a/web/docs/guides/deployment/coolify.md
+++ b/web/docs/guides/deployment/coolify.md
@@ -2,6 +2,7 @@
 comments: true
 last_checked_with_versions:
   Wasp: 0.21.0
+  Coolify: 2026-01-30
 ---
 
 import { SecretGeneratorBlock } from "../../project/SecretGeneratorBlock";


web/src/remark/search-and-replace.ts
```diff
@@ -10,6 +10,14 @@ const replacements = [
     search: /{latestWaspVersion}/g,
     replace: `^${latestWaspVersion}`,
   },
+  {
+    search: /{pinnedLatestWaspVersion}/g,
+    replace: `${latestWaspVersion}`,
+  },
+  {
+    search: /{minimumNodeJsVersion}/g,
```
infomiho: I used this to keep the Node.js version in the deployment guides in one place. I tried using it for the `quick-start.md` but the `{minimumNodeJsVersion}` syntax was tripping MDX up. Do we switch to something like `[minimumNodeJsVersion]`? I didn't want to do that refactor in this PR.

web/docs/deployment/deployment-methods/wasp-deploy/ci-cd.md
cprecioso, line 31: Looks like something that could be moved to the new guides in the future too

web/src/remark/search-and-replace.ts
cprecioso: Would this be fine in MDX?

```mdx
Hello you need Node.js version {"{minimumNodeJsVersion}"} to use Wasp.
```

.nvmrc
```diff
@@ -4,5 +4,6 @@
 # also change it in:
 #   - The compiler (Wasp.Node.Version).
 #   - The docs: /web/docs/introduction/getting-started.md -> The "Requirements" section.
+#   - The docs plugin: /web/src/remark/search-and-replace.ts
```
cprecioso, line 7: I think every place this links to also links back to here, maybe we can do the same.

web/docs/guides/deployment/caprover.md
```diff
@@ -0,0 +1,243 @@
...
+  push:
+    branches:
+      - "main"
+
+concurrency:
+  group: deployment
+  cancel-in-progress: true
+
+env:
+  WASP_VERSION: "{pinnedLatestWaspVersion}"
```
cprecioso, line 111: I'm unsure of making these parameters auto-updatable. Since it seems to defeat the whole point of "last tested with version blah", it could lead to confusion that some things look up-to-date and others not. Same with `minimumNodeJsVersion`.

On the other hand I do understand that it'd be quite annoying to have this not work because of this missing change. So I really don't know what to do.

RAW I guess

web/docs/guides/deployment/caprover.md
```diff
@@ -0,0 +1,243 @@
...
+### Step 1: Set Up Your Domain
+
+Point your DNS A records to your server IP:
+
+- `@` (root) → server IP (for `myapp.com` - client)
+- `api` → server IP (for `api.myapp.com` - server)
+
+:::tip
+If you followed Caprover's install instructions with `*.apps` subdomain setup, you can use `https://<app-name>.apps.mydomain.com` for quick testing.
+:::
```
cprecioso, line 39: If they do this, `myapp.apps.mycaprover.com` would work but a double subdomain like `api.myapp.apps.mycaprover.com` wouldn't work, right?

web/docs/guides/deployment/caprover.md
```diff
@@ -0,0 +1,243 @@
+---
+comments: true
+---
```
cprecioso, line 6: No `last_checked_with_versions`?

web/docs/guides/deployment/vps.md
```diff
@@ -0,0 +1,329 @@
...
+- Basic familiarity with SSH and Linux commands
+
+### Architecture Overview
+
+Our deployment setup includes:
+
+- **Ubuntu LTS** as the operating system
+- **Caddy** as a reverse proxy for HTTPS and domain handling
+- **Docker** for running the server and database
+- Serving the client with a static file server
```
cprecioso, line 30: I'd say more explicitly that we assume a Debia/Ubuntu server, and that you should adapt the steps for your distribution if different.

web/docs/guides/deployment/vps.md
```diff
@@ -0,0 +1,329 @@
...
+
+First, update your package list:
+
+```bash
+apt update
+```
+
+If Apache is installed, you may need to [uninstall it](https://askubuntu.com/a/387793) first. Check with `which apache2`.
+
+Install Caddy following the [official Ubuntu instructions](https://caddyserver.com/docs/install#debian-ubuntu-raspbian).
```
cprecioso, line 52: Optional, improvement

Since we're already using Docker to serve the app, we can just tell them to use Caddy as a Docker too, right? Seems like a good opportunity to provide them with a ready-made Docker compose even.

web/docs/deployment/deployment-methods/wasp-deploy/ci-cd.md
infomiho: Yes, this is an ideal candidate for a guide: specific technology used and has the potential to be outdated 😄

web/docs/guides/deployment/caprover.md
infomiho: I guess updating the Node.js version will trigger updating the guides for them to be "true". The guide will not work with the latest Wasp version if the Node.js is too low. So yeah, it's a tricky decision, I'm not sure what's a better option. Maybe we go with this since it's more ambitious and see if it comes back to haunt us?

web/docs/guides/deployment/caprover.md
infomiho: Nope, it wouldn't work. I tested the guide with `my-server.apps.*` and `my-client.apps.*`

web/docs/guides/deployment/caprover.md
infomiho: I was not aware I should add this, lemme check the materials to educate myself 😄

web/docs/guides/deployment/vps.md
infomiho: This is a specific guide so it uses a specific distro, of course they can do their own specific setup. So, unlike in the docs, where I'd add this kind of note, in the guide I don't feel compelled to do it. Do you think we should add it anyways?

web/docs/guides/deployment/vps.md
infomiho: This is _a_ way to do it, I didn't want to over-do it. I think there is a thing about ports and ufw I didn't also account for: https://www.reddit.com/r/selfhosted/comments/15f7ju5/docker_and_ufw_firewall/

Do we want to have _the best VPS setup_ or _a VPS setup_? I didn't want to invest time in creating a new setup right now. Maybe we can create an issue for this to tackle it in the future?

web/src/remark/search-and-replace.ts
infomiho: I'll try it out, thanks for the tip 👍 

The way we wrote the search & replace - it only looks in `code` blocks MDX nodes, I'll have to figure out which node type is this 😄

web/docs/guides/deployment/caprover.md
cprecioso: Then let's make it clearer here because it seems like we're saying they can have an `api.` sub-sub-subdomain.

web/docs/guides/deployment/caprover.md
cprecioso: Check the PR body bro 😁

web/docs/guides/deployment/vps.md
cprecioso: I do

web/docs/guides/deployment/vps.md
cprecioso: > Do we want to have the best VPS setup or a VPS setup? 

I want to have the one that gets users faster to a working production app while having the least failure modes 😁, that's why I'd say that if we're using Docker, use it all the way.
But I don't really mind right now, we'll do a pass with the Deployment epic anyway.

web/docs/guides/deployment/caprover.md
infomiho: image

It looked like this when I looked at it, I swear 😂

web/src/remark/search-and-replace.ts
infomiho: This led me down the path of needing to traverse `estree` AST to be able to replace this:
```ts
[
  {
    type: 'ExpressionStatement',
    expression: Node {
      type: 'Literal',
      start: 2308,
      end: 2332,
      value: '{minimumNodeJsVersion}',
      raw: '"{minimumNodeJsVersion}"',
      loc: [Object],
      range: [Array]
    },
    start: 2308,
    end: 2332,
    loc: { start: [Object], end: [Object] },
    range: [ 2308, 2332 ]
  }
]
```

This felt like too much, so I reverted the change.

web/docs/deployment/deployment-methods/wasp-deploy/ci-cd.md
infomiho: On second thought, this is part of our 1st party deployment command. Because of this, I think it should stay in the docs.

web/docs/guides/deployment/caprover.md
infomiho: Updated it to include last tested Wasp version, I think that's enough of a reference if these are recent instructions.

web/docs/guides/deployment/vps.md
infomiho: Now that I added it, it feels like a no-op:

image

If they adapt the steps - they are no longer following the same guide. I'd rather leave the note out.

web/docs/guides/deployment/caprover.md
cprecioso: why not also say which version of caprover? even if it's a date

cprecioso: also, weird that the PR body is not showing 🤔

web/docs/guides/deployment/vps.md
cprecioso: okay, i trust you then 😊

web/docs/deployment/deployment-methods/self-hosted.md
```diff
@@ -69,7 +70,7 @@ On your server:
 
 Check out one of our step-by-step guides for more details:
 
-<GuideLink linkToGuide="https://gist.github.com/infomiho/80f3f50346566e39db56c5e57fefa1fe" title="Deploying Wasp with Docker on your server" description="Uses Ubuntu, Git, Caddy, Docker" />
+<GuideLink linkToGuide="/docs/guides/deployment/vps" title="Deploying Wasp with Docker on your server" description="Uses Ubuntu, Git, Caddy, Docker" />
```
cprecioso: Nice!

cprecioso:
```suggestion
<GuideLink linkToGuide="../../guides/deployment/vps.md" title="Deploying Wasp with Docker on your server" description="Uses Ubuntu, Git, Caddy, Docker" />
```

So that versioned docs are guides are linked together.

web/src/remark/search-and-replace.ts
```diff
@@ -5,31 +5,38 @@ import docsVersions from "../../versions.json";
...
 
+/**
+  Accepts a record of key-value pairs. Replaces `{key}` with `value` when
+  the text is searched and replaced.
+*/
+function createReplacer(definitions: Record<string, string>) {
+  const compiled = Object.entries(definitions).map(([key, value]) => ({
+    regex: new RegExp(`\\{${key}\\}`, "g"),
+    value,
+  }));
```
cprecioso: it really feels like we don't need a regex here, calling `text.replaceAll("{minimumNodeJsVersion}", "22.12")` would be enough

web/docs/deployment/deployment-methods/self-hosted.md
```diff
@@ -4,6 +4,7 @@ title: Self-Hosted
 
 import { ImgWithCaption } from '@site/blog/components/ImgWithCaption'
 import { GuideLink } from './GuideLink.tsx'
+import useBaseUrl from '@docusaurus/useBaseUrl';
```
cprecioso: What are we using this for?

web/docs/guides/deployment/caprover.md
infomiho: That was a joke, I added a white rectangle 😄 I didn't read it properly! 

> why not also say which version of caprover? even if it's a date

I think a date is more suitable, I didn't want to use versions since I don't think anyone knows which version of Caprover or Coolify they are using.

web/docs/deployment/deployment-methods/self-hosted.md
infomiho: Leftover, sorry about that ✋

Issue #3473, opened 2025-12-04
Implement wider support for Vite


infomiho:
Based on the RFC produced #3472 implement changes so Vite is more exposed to the user. 

Internal Notion document with the RFC: https://www.notion.so/wasp-lang/RFC-Vite-unleashed-2b818a74854c80dabad9c91567c0cfd0?source=copy_link

PR #3604, opened 2026-01-13
Move Vite config into `wasp` plugin

infomiho:
## Description

Introduces the concept of `wasp` plugin which contains all the custom Vite config Wasp client app requires. 

This PR is just moving things around: from `vite.config.ts` config file to `wasp.ts` plugin. Doesn't change the config.

This PR is simplifying the way we define Vite plugins. There is only one plugin now that we care about so I got rid of the fancy plugin registration system.

Part of #3473

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/data/Generator/templates/react-app/vite/wasp.ts
+++ b/waspc/data/Generator/templates/react-app/vite/wasp.ts
@@ -1,6 +1,6 @@
 {{={= =}=}}
 import path from "node:path"
-import { type Plugin, mergeConfig } from "vite";
+import { type PluginOption, mergeConfig } from "vite";
 import { defaultExclude } from "vitest/config"
 import react, { type Options as ReactOptions } from "@vitejs/plugin-react";
 import { validateEnv } from "./plugins/validateEnv";
@@ -10,13 +10,13 @@ export interface WaspPluginOptions {
   reactOptions?: ReactOptions;
 }
 
-export function wasp(options?: WaspPluginOptions): Plugin[] {
+export function wasp(options?: WaspPluginOptions): PluginOption {
   return [
     validateEnv(),
-    ...react(options?.reactOptions),
+    react(options?.reactOptions),
     detectServerImports(),
     {
-      name: "wasp-config",
+      name: "wasp:config",
       config(config) {
         return mergeConfig({
           base: "{= baseDir =}",
--- a/waspc/data/Generator/templates/react-app/vite/plugins/detectServerImports.ts
+++ b/waspc/data/Generator/templates/react-app/vite/plugins/detectServerImports.ts
@@ -5,7 +5,7 @@ import { resolveProjectPath } from 'wasp/dev'
 
 export function detectServerImports(): Plugin {
   return {
-    name: 'wasp-detect-server-imports',
+    name: 'wasp:detect-server-imports',
     enforce: 'pre',
     resolveId(source, importer) {
       if (!importer) {
--- a/waspc/data/Generator/templates/react-app/vite/plugins/validateEnv.ts
+++ b/waspc/data/Generator/templates/react-app/vite/plugins/validateEnv.ts
@@ -12,7 +12,7 @@ const redColorFormatString = getColorizedConsoleFormatString('red');
 export function validateEnv(): Plugin {
   let validationResult: ReturnType<typeof getValidatedEnvOrError> | null = null
   return {
-    name: 'wasp-validate-env',
+    name: 'wasp:validate-env',
     configResolved: (config) => {
       const env = loadEnv(config.mode, process.cwd(), config.envPrefix)
       validationResult = getValidatedEnvOrError(env, clientEnvSchema)
--- a/waspc/e2e-tests/snapshots/kitchen-sink-golden/wasp-app/.wasp/out/.waspchecksums
+++ b/waspc/e2e-tests/snapshots/kitchen-sink-golden/wasp-app/.wasp/out/.waspchecksums
@@ -2440,20 +2440,20 @@
             "file",
             "web-app/vite/plugins/detectServerImports.ts"
         ],
-        "e9735d79c5d17f6f0ed45d978d705f7b815a00ca0a96dedde08d67b505db2225"
+        "4d4451b2d692a7e4d623ba91016cab1391294d204763813bf08192cb866052db"
     ],
     [
         [
             "file",
             "web-app/vite/plugins/validateEnv.ts"
         ],
-        "956bb762df6ac7a2bb493dfa335c5fbd9c53be082e709eaa8a0873a75f143784"
+        "e4d0bb7a2fb9d3415e0d1dc5c137fec5e9a4073a2eda33daf50396f5d477490e"
     ],
     [
         [
             "file",
             "web-app/vite/wasp.ts"
         ],
-        "ef44ec1a0c2f269bad2c08c81c66f392ac4faa51cdebcbc80a5bb7ef97555bf9"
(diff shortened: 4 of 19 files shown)


waspc/src/Wasp/Generator/WebAppGenerator/Vite/VitePlugin.hs
```diff
@@ -1,57 +1,77 @@
...
+            </> generatedCodeDirInDotWaspDir
+            </> C.webAppRootDirInProjectRootDir
+            </> C.webAppSrcDirInWebAppRootDir
+            </> [relfile|test/vitest/setup.ts|]
+      ]
+    relPathFromWebAppRootDirToWaspProjectDir =
+      fromJust $
+        SP.parseRelDirP $
+          FP.Extra.reversePosixPath $
+            SP.fromRelDir (dotWaspDirInWaspProjectDir </> generatedCodeDirInDotWaspDir </> C.webAppRootDirInProjectRootDir)
```
infomiho, line 69: Template data is moved from `vite.config.ts` generator - nothing changed.

waspc/data/Generator/templates/react-app/vite.config.ts
```diff
@@ -18,53 +14,7 @@ const _waspUserProvidedConfig = {};
...
-    globals: true,
-    environment: "jsdom",
-    // Since Vitest is running from the root of the project, we need
-    // to specify the path to the setup file relative to the root.
-    setupFiles: {=& vitest.setupFilesArray =},
-    exclude: [
-      ...defaultExclude,
-      "{= vitest.excludeWaspArtefactsPattern =}",
-    ]
-  },
```
infomiho, line 67: All of this config moved to the `wasp` plugin.

waspc/data/Generator/templates/react-app/vite/wasp.ts
```diff
@@ -0,0 +1,68 @@
...
+import { detectServerImports } from "./plugins/detectServerImports";
+
+export interface WaspPluginOptions {
+  reactOptions?: ReactOptions;
+}
+
+export function wasp(options?: WaspPluginOptions): Plugin[] {
+  return [
+    validateEnv(),
+    ...react(options?.reactOptions),
```
infomiho: `react` plugin is actually a list of plugins, so we need to deconstruct it like this.

waspc/data/Generator/templates/react-app/vite/wasp.ts
```diff
@@ -0,0 +1,68 @@
...
+import path from "node:path"
+import { type Plugin, mergeConfig } from "vite";
+import { defaultExclude } from "vitest/config"
+import react, { type Options as ReactOptions } from "@vitejs/plugin-react";
+import { validateEnv } from "./plugins/validateEnv";
+import { detectServerImports } from "./plugins/detectServerImports";
+
+export interface WaspPluginOptions {
+  reactOptions?: ReactOptions;
+}
```
infomiho, line 11: This config is not exposed to the user _yet_, but it will be once the `wasp` plugin moves to user-land.

waspc/src/Wasp/Generator/WebAppGenerator/Vite/VitePlugin.hs
```diff
@@ -1,57 +1,77 @@
...
+genVitePlugins :: AppSpec -> Generator [FileDraft]
+genVitePlugins spec =
+  sequence
+    [ genWaspPlugin spec,
+      genDetectServerImportsPlugin,
+      genValidateEnvPlugin
+    ]
 
-data WebAppVitePluginsDir
+vitePluginsGlob :: String
```
infomiho, line 36: We introduced a glob for the `tsconfig.vite.json` instead of need to list all of the plugins one by one.

waspc/data/Generator/templates/react-app/vite/wasp.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export function wasp(options?: WaspPluginOptions): Plugin[] {
+  return [
+    validateEnv(),
+    ...react(options?.reactOptions),
+    detectServerImports(),
+    {
+      name: "wasp-config",
+      config(config) {
+        return mergeConfig({
```
cprecioso, line 21: You don't need to do a `mergeConfig`, by default vite will merge whatever you return. [(link)](https://vite.dev/guide/api-plugin#config:~:text=can%20return%20a%20partial%20config%20object%20that%20will%20be%20deeply%20merged%20into%20existing%20config)

If you are explicitly using `mergeConfig` because you need our config to take precedence, add a note saying so, because it's easy to miss and someone (me) might want to "fix it" in the future.

waspc/data/Generator/templates/react-app/vite/wasp.ts
```diff
@@ -0,0 +1,68 @@
...
+  reactOptions?: ReactOptions;
+}
+
+export function wasp(options?: WaspPluginOptions): Plugin[] {
+  return [
+    validateEnv(),
+    ...react(options?.reactOptions),
+    detectServerImports(),
+    {
+      name: "wasp-config",
```
cprecioso: The convention for multi-plugins seems to be `$main-name:$sub-name`.
So I'd name this `wasp:config`, and the other two `wasp:detect-server-imports` and `wasp:validate-env`.
No real impact so RAW.

waspc/data/Generator/templates/react-app/vite/wasp.ts
cprecioso: You can use `PluginOption` (it's Vite's type for `config.plugins`) so that you don't need to depend on the React plugin being an array; Vite will flatten everything as needed.

```suggestion
export function wasp(options?: WaspPluginOptions): PluginOption {
  return [
    validateEnv(),
    react(options?.reactOptions),
```

waspc/data/Generator/templates/react-app/vite/wasp.ts
infomiho: I did not know that I can skip it 😄  But yeah, I want our options to "dominate" until we have a reason to do it differently (our users asking for it).

waspc/data/Generator/templates/react-app/vite/wasp.ts
infomiho: Good to know, I'll change it 👍

waspc/data/Generator/templates/react-app/vite/wasp.ts
infomiho: Talked about this offline, `mergeConfig` doesn't do anything for us here. If we need to force some options, we need to mutate the config [directly](https://vite.dev/guide/api-plugin#config). This something we can sort out as separate effort since this is not related to this effort since it's an existing concern.

infomiho: I have to correct myself, I assumed wrong how `mergeConfig` works. `mergeConfig` is defined as:
```ts
function mergeConfig(
  defaults: Record<string, any>,
  overrides: Record<string, any>,
  isRoot = true,
): Record<string, any>
```
which means that the first options are overridden by the second options. In our case, our options were first which meant that we overrode them with the existing user-provided config. This allowed users to override our options e.g. `server.open` to `false`. This is the intended behavior. 

After I removed the `mergeOptions`, we no longer overrode our options with the existing config. And when we returned our options object, Vite used it to override the existing options which meant that e.g. user couldn't change the `server.open` to `false` because we set it to `true` in our options. 

For now, I'll keep the `mergeConfig` in this PR and I'll experiment if we should remove it in the last PR in this Vite PR chain.

Issue #2007, opened 2024-05-05
Add support for lazy loading Wasp routes


infomiho:
Quote from [here](https://hackernoon.com/boosting-react-app-performance-a-guide-to-lazy-loading-and-suspense):
> Lazy loading is a strategy that defers the loading of non-essential resources until they are needed, reducing the initial loading time and improving the overall performance of the application. In React, we can achieve lazy loading using the lazy function and the Suspense component, introduced in React 16.6.

Our current React Router version (v5) supports using `React.lazy`: https://ui.dev/react-router-v5-code-splitting so it shouldn't be a big deal to implement this.

We should decide if we want to make this the default behaviour (some frameworks do this) or make it opt in.

Martinsos:
Another related Discord thread:
https://discord.com/channels/686873244791210014/1300740060114587680/1300740060114587680

PR #3608, opened 2026-01-14
Move client entry points to virtual modules

infomiho:
## Description

We are still using `web-app` dir and not in the user-land, but this is the last bit of preparation to move to running Vite in user-land. Once the `wasp` plugin is used in user's `vite.config.ts` and `npx vite` is started from user's project dir, the client will just work because all of the entry files (`index.html` and `index.tsx` are virtual and don't need to be in the user's project).

Part of #3473

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html/build.ts
+++ b/waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html/build.ts
@@ -2,10 +2,10 @@ import type { Plugin } from "vite";
 import { getIndexHtmlContent } from "../../virtual-files/index.js";
 import { makeVirtualFilesResolver, type VirtualFiles } from "../../virtual-files/resolver.js";
 
-const indexHtmlFileName = "index.html";
+const INDEX_HTML_FILE_NAME = "index.html";
 
 const resolveVirtualFiles = makeVirtualFilesResolver([
-  { id: indexHtmlFileName, load: getIndexHtmlContent },
+  { id: INDEX_HTML_FILE_NAME, load: getIndexHtmlContent },
 ]);
 
 export function waspHtmlBuild(): Plugin {
@@ -13,6 +13,7 @@ export function waspHtmlBuild(): Plugin {
 
   return {
     name: "wasp:html-build",
+    apply: "build",
     config() {
       return {
         build: {
@@ -20,7 +21,7 @@ export function waspHtmlBuild(): Plugin {
             // Vite tries to find the entry file on disk (which doesn't exist)
             // so the build fails. We tell Vite/Rollup to use `index.html` even
             // though it doesn't exist on the disk.
-            input: indexHtmlFileName,
+            input: INDEX_HTML_FILE_NAME,
           },
         },
       };
--- a/waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html/dev.ts
+++ b/waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html/dev.ts
@@ -1,11 +1,15 @@
-import type { Plugin, Connect } from "vite";
+import type { Plugin } from "vite";
 import { getIndexHtmlContent } from "../../virtual-files/index.js";
 
 export function waspHtmlDev(): Plugin {
   return {
     name: "wasp:html-dev",
+    apply: "serve",
     configureServer(server) {
       return () => {
+        // Post middleware: runs after Vite's built-in SPA fallback
+        // middleware which resolves routes to `/index.html` which
+        // we pick up here and return the virtual `index.html` content.
         server.middlewares.use(async (req, res, next) => {
           if (req.url === "/" || req.url === `/index.html`) {
             try {
--- a/examples/kitchen-sink/package-lock.json
+++ b/examples/kitchen-sink/package-lock.json
@@ -159,16 +159,6 @@
         "node": ">=22.12.0"
       }
     },
-    ".wasp/out/server/node_modules/@types/node": {
-      "version": "22.19.3",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-22.19.3.tgz",
-      "integrity": "sha512-1N9SBnWYOJTrNZCdh/yJE+t910Y128BoyY+zBLWhL3r0TYzlTmFdXrPwHL9DyFZmlEXNQQolTZh3KHV31QDhyA==",
-      "dev": true,
-      "license": "MIT",
-      "dependencies": {
-        "undici-types": "~6.21.0"
-      }
-    },
     ".wasp/out/web-app": {
       "name": "@wasp.sh/generated-webapp",
       "version": "0.0.0",
@@ -2978,19 +2968,13 @@
       "license": "MIT"
     },
     "node_modules/@types/node": {
-      "version": "25.0.3",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-25.0.3.tgz",
-      "integrity": "sha512-W609buLVRVmeW693xKfzHeIV6nJGGz98uCPfeXI1ELMLXVeKYZ9m15fAMSaUPBHYLGFsVRcMmSCksQOrZV9BYA==",
+      "version": "22.19.7",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-22.19.7.tgz",
+      "integrity": "sha512-MciR4AKGHWl7xwxkBa6xUGxQJ4VBOmPTF7sL+iGzuahOFaO0jHCsuEfS80pan1ef4gWId1oWOweIhrDEYLuaOw==",
       "license": "MIT",
       "dependencies": {
-        "undici-types": "~7.16.0"
+        "undici-types": "~6.21.0"
       }
-    },
-    "node_modules/@types/node/node_modules/undici-types": {
-      "version": "7.16.0",
-      "resolved": "https://registry.npmjs.org/undici-types/-/undici-types-7.16.0.tgz",
-      "integrity": "sha512-Zz+aZWSj8LE6zoxD+xrjh4VfkIG8Ya6LvYkZqtUQGJPZjYl53ypCaUwWqo7eI0x66KBGeRo+mlBEkMSeSZ38Nw==",
-      "license": "MIT"
     },
     "node_modules/@types/qs": {
       "version": "6.14.0",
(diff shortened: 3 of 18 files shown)


waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/virtual-files/index.virtual.ts
```diff
@@ -0,0 +1,31 @@
+{{={= =}=}}
+export function getIndexTsxContent(): string {
+  return `import React from "react";
+  import ReactDOM from "react-dom/client";
+  import { getWaspApp } from "wasp/client/app";
+  {=& routesMapping.importStatement =}
```
infomiho: I was wondering if it makes sense to inline the routes mapping here and not import it from another virtual module? Thoughts?

infomiho: Keeping it like this enables users to override `index.tsx` and just import the routes vs. them needing to recreate the routes mapping object.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html.ts
```diff
@@ -0,0 +1,87 @@
...
+      if (id === indexHtmlFileName) {
+        return indexHtmlPath;
+      }
+    },
+    load(id) {
+      // Provide content for virtual index.html
+      if (id === indexHtmlPath) {
+        return getIndexHtmlContent();
+      }
+    },
```
infomiho: I've grouped all the code related to `index.html` in this plugin. This is a virtual module implementation and it can go into the `wasp-virtual-modules` plugin, but I did this instead to keep the `index.html` logic together.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html.ts
```diff
@@ -0,0 +1,87 @@
...
+    configureServer(server) {
+      server.middlewares.use(spaFallbackMiddleware());
+
+      return () => {
+        server.middlewares.use(async (req, res, next) => {
+          if (req.url === "/" || req.url === `/${indexHtmlFileName}`) {
```
infomiho: Rewrite requests for `index.html` to use our "virtual' content.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html.ts
```diff
@@ -0,0 +1,87 @@
...
+import type { Plugin, Connect } from "vite";
+import path from "node:path";
+import { getIndexHtmlContent } from "./virtual-files/indexHtml.virtual.js";
+
+const indexHtmlFileName = "index.html";
+let indexHtmlPath: string;
+
+export function waspHtml(): Plugin {
+  return {
+    name: "wasp-html",
```
infomiho: Should I split this into `wasp-html-dev` and `wasp-html-build`?

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/virtualModules.ts
```diff
@@ -0,0 +1,44 @@
...
+  return {
+    name: "wasp-virtual-modules",
+    enforce: "pre",
+    configResolved(config) {
+      projectRoot = config.root;
+      // Using absolute paths gives proper context for resolving relative imports.
+      indexTsxPath = path.resolve(projectRoot, "{= indexVirtualFileName =}");
+      routesPath = path.resolve(projectRoot, "{= routesVirtualFileName =}");
+    },
+    resolveId(id) {
```
infomiho: Is there a way to maybe abstract away this two step process `resolveId` + `load`? I did it in a very basic way.

waspc/src/Wasp/Generator/SdkGenerator/Client/VitePlugin/HtmlPluginG.hs
```diff
@@ -0,0 +1,48 @@
...
+genVirtualIndexHtml :: AppSpec -> Generator FileDraft
+genVirtualIndexHtml spec =
+  return $
+    C.mkTmplFdWithData tmplPath tmplData
+  where
+    tmplPath = C.vitePluginsDirInSdkTemplatesDir </> virtualFilesDirInVitePluginsDir </> [relfile|indexHtml.virtual.ts|]
+    tmplData =
+      object
+        [ "title" .= (AS.App.title (snd $ getApp spec) :: String),
+          "head" .= (maybe "" (intercalate "\n") (AS.App.head $ snd $ getApp spec) :: String),
```
infomiho, line 50: Old `index.html` template data.

waspc/src/Wasp/Generator/JsImport.hs
```diff
@@ -49,3 +53,20 @@ jsImportToImportJson maybeJsImport = maybe notDefinedValue mkTmplData maybeJsImp
               "importStatement" .= jsImportStmt,
               "importIdentifier" .= jsImportIdentifier
             ]
+
+extImportToRelativeSrcImportFromViteExecution :: EI.ExtImport -> JsImport
```
infomiho, line 57: I'm open to suggestions how to make this a better name. The idea is that we are using it to generate relative imports from the user's `src` dir **from** where Vite is executed. Now that's `web-app` and we are using `../../../` as our relative path but in later PRs that will become just `.`.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/virtual-files/index.virtual.ts
cprecioso, line 1: I really dislike this string-inside-code-inside-templated-file solution.
I'd try to have these as naked templated files and do either of these:

(a)
```ts
const indexHtmlFile = new URL("./virtual-files/index.html", import.meta.url).href
resolveId(source) {
  if (source === "index.html") return indexHtmlFile
}
```


(b)
```ts
const indexHtmlFile = new URL("./virtual-files/index.html", import.meta.url).href
load(id) {
  if (id === "index.html") return fs.readFile(indexHtmlFile)
}
```

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html.ts
```diff
@@ -0,0 +1,87 @@
...
+    config() {
+      return {
+        appType: "custom",
+        build: {
+          rollupOptions: {
+            input: indexHtmlFileName,
```
cprecioso: Can you double-check if this is needed? If so, write a comment with the explanation pls

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/virtual-files/index.virtual.ts
cprecioso: I like that, can you write it in a comment so that we don't forget?

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html.ts
```diff
@@ -0,0 +1,87 @@
+{{={= =}=}}
+import type { Plugin, Connect } from "vite";
+import path from "node:path";
+import { getIndexHtmlContent } from "./virtual-files/indexHtml.virtual.js";
+
+const indexHtmlFileName = "index.html";
+let indexHtmlPath: string;
```
cprecioso: Otherwise this will throw TS errors, right?

```suggestion
let indexHtmlPath!: string;
```

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export function waspHtml(): Plugin {
```
cprecioso: I'd add a note that the `resolveId/load` thing is for build, and `configureServer` is for dev. You might want to split it into a `wasp:html:dev` and `wasp:html:build` plugins even.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html.ts
```diff
@@ -0,0 +1,87 @@
...
+        return getIndexHtmlContent();
+      }
+    },
+    configureServer(server) {
+      server.middlewares.use(spaFallbackMiddleware());
+
+      return () => {
+        server.middlewares.use(async (req, res, next) => {
+          if (req.url === "/" || req.url === `/${indexHtmlFileName}`) {
+            try {
```
cprecioso: Is the `try/catch` needed? I think since some time now connect will catch errors in middleware.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html.ts
```diff
@@ -0,0 +1,87 @@
...
+        const url = req.url || "/";
+        const { pathname } = new URL(url, "http://localhost");
+        // If it's a route (no extension, not a special Vite path), rewrite to /index.html
+        if (!pathname.includes(".") && !pathname.startsWith("/@")) {
+          req.url = `/${indexHtmlFileName}`;
+        }
+      }
+      next();
+    };
+  }
```
cprecioso: Can we add a note why this is needed?
Maybe instead of a fallback we generate the `index.html` for the defined pages in `main.wasp` and let it 404 for the rest.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/virtualModules.ts
cprecioso: You can centralize them all in an array or object and operate over that, makes it nicer to communicate intent. There's not a very nice way to mixin behaviour otherwise.

```ts
const virtualFiles = { "index.tsx": getIndexTsxContent, /* etc */ }

resolveId(source) {
  if (source in virtualFiles) return source
}

load(source) {
  if (source in virtualFiles) return virtualFiles[source]()
}
```

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/virtualModules.ts
```diff
@@ -0,0 +1,42 @@
...
+  let routesPath: string;
+
+  return {
+    name: "wasp-virtual-modules",
+    enforce: "pre",
+    configResolved(config) {
+      projectRoot = config.root;
+      // Using absolute paths gives proper context for resolving relative imports.
+      indexTsxPath = path.resolve(projectRoot, "{= indexVirtualFileName =}");
+      routesPath = path.resolve(projectRoot, "{= routesVirtualFileName =}");
```
cprecioso: I don't think this is needed, we don't even need to tell Vite to look for the modules on those paths.

Inspired by the convention in 1st party Vite plugins (check fast refresh in the React plugin), I'd have something like this:

`index.html`:
```html
<script type="module" src="/@wasp/client-entry"></script>
```

And resolve for that custom ID. Same for the routes path.

They could be actual exports of the Wasp SDK package even (it might be hard tho):
```
wasp/client/entry
wasp/client/routes
```

waspc/src/Wasp/Generator/SdkGenerator/Client/VitePlugin/VirtualModulesPlugin/VirtualRoutesG.hs
```diff
@@ -0,0 +1,85 @@
...
+          "isAuthEnabled" .= isAuthEnabled spec
+        ]
+
+-- NOTE: This should be prevented by Analyzer, so use error since it should not be possible
+determineRouteTargetComponent :: AppSpec -> (String, AS.Route.Route) -> String
+determineRouteTargetComponent spec (_, route) =
+  maybe
+    targetPageName
+    determineRouteTargetComponent'
+    (AS.Page.authRequired $ snd targetPage)
```
cprecioso: Seems like this is more understandable.

```hs
  determineRouteTargetComponent' $
    fromMaybe false $ 
      (AS.Page.authRequired $ snd targetPage)
```

I'd even bump the if to the main body then

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/virtual-files/routes.virtual.ts
```diff
@@ -0,0 +1,17 @@
...
+  import { createAuthRequiredPage } from "wasp/client/app"
+  {=/ isAuthEnabled =}
+
+  {=# pagesToImport =}
+  {=& importStatement =}
+  {=/ pagesToImport =}
+
+  export const routesMapping = {
+    {=# routes =}
+    {= name =}: {= targetComponent =},
```
cprecioso: Should these be `async () => (await import("file")).component`?
Otherwise we're loading all the pages always.

cprecioso (changes requested):
Good job wrangling Vite into shape! Mustn't have been easy. 🙏

I commented about changes I'd like to see:
- Strong "request changes" on how we template the virtual files
- Medium "request changes" for virtual files' IDs
- Medium "let's investigate" on removing the SPA fallback in the HTML server, and generating entry points for the pages we set.
- Soft for everything else 
- Do we allow people to override the entry points yet? If so, let's also take a first stab at documentation.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html.ts
infomiho: I guess it's meant to be: https://github.com/wasp-lang/wasp/pull/3608#discussion_r2698677247

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/virtual-files/routes.virtual.ts
infomiho: I'd do this as a separate effort, there is an issue for this: https://github.com/wasp-lang/wasp/issues/2007

waspc/src/Wasp/Generator/SdkGenerator/Client/VitePlugin/VirtualModulesPlugin/VirtualRoutesG.hs
infomiho: This was old code, but I'll give it go 👍

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/virtualModules.ts
infomiho: I'll try to implement it like that 👍 thanks for the suggestion

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html.ts
infomiho: This would mean that Vite is becoming our routing layer in dev? Did I misunderstand that? I'd like to keep Vite behavior as close to what it does today, just serve `index.html` for all the requests and let React Router handle the routing and 404s.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/virtual-files/index.virtual.ts
infomiho: Having the files in the SDK package cause `tsc` to check them and throw errors. They shouldn't exist as part of the SDK and shouldn't be checked by `tsc` since they are really `web-app` files. 

I'm not sure how to pull this off. I'll experiment with `exclude` in `tsconfig.json` and maybe `.vritual` file extension so Typescript ignores it.

cprecioso: `// @ts-ignore`?

infomiho: Then we'll be shipping it in user-land with that flag. But that maybe okay 🤷  I'll report back once it try it out.

cprecioso: Another option is to do the templating purely on JS side with JSON values received from Haskell, check `waspc/data/Generator/templates/server/src/views/wrong-port.ts`

infomiho: Doing `@ts-ignore` doesn't work, `tsc` still tries to resolve imports:
```
[  Wasp  ] > wasp@1.0.0 build
[  Wasp  ] > tsc && node ./scripts/copy-assets.js
[  Wasp  ]
[  Wasp  ] client/vite/virtual-files/files/index.virtual.tsx(4,28): error TS2307: Cannot find module 'wasp/client/app' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/index.virtual.tsx(5,31): error TS5097: An import path can only end with a '.tsx' extension when 'allowImportingTsExtensions' is enabled.
[  Wasp  ] client/vite/virtual-files/files/index.virtual.tsx(7,32): error TS2307: Cannot find module '../../../src/App' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/index.virtual.tsx(9,36): error TS2307: Cannot find module '../../../src/clientSetup' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/routes.virtual.tsx(4,26): error TS2307: Cannot find module '../../../src/pages/HomePage' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/routes.virtual.tsx(5,30): error TS2307: Cannot find module '../../../src/pages/CatchAllPage' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/routes.virtual.tsx(6,24): error TS2307: Cannot find module '../../../src/features/auth/pages/Signup' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/routes.virtual.tsx(7,23): error TS2307: Cannot find module '../../../src/features/auth/pages/Login' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/routes.virtual.tsx(8,52): error TS2307: Cannot find module '../../../src/features/auth/pages/PasswordReset' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/routes.virtual.tsx(9,60): error TS2307: Cannot find module '../../../src/features/auth/pages/EmailVerification' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/routes.virtual.tsx(10,66): error TS2307: Cannot find module '../../../src/features/auth/pages/RequestPasswordReset' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/routes.virtual.tsx(11,29): error TS2307: Cannot find module '../../../src/features/auth/pages/ProfilePage' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/routes.virtual.tsx(12,34): error TS2307: Cannot find module '../../../src/features/auth/pages/ManualSignupPage' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/routes.virtual.tsx(13,34): error TS2307: Cannot find module '../../../src/features/auth/pages/CustomSignupPage' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/routes.virtual.tsx(14,27): error TS2307: Cannot find module '../../../src/features/operations/pages/TasksPage' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/routes.virtual.tsx(15,44): error TS2307: Cannot find module '../../../src/features/operations/pages/TaskDetailPage' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/routes.virtual.tsx(16,35): error TS2307: Cannot find module '../../../src/features/operations/pages/SerializationPage' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/routes.virtual.tsx(17,26): error TS2307: Cannot find module '../../../src/features/jobs/pages/JobsPage' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/routes.virtual.tsx(18,26): error TS2307: Cannot find module '../../../src/features/apis/pages/ApisPage' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/routes.virtual.tsx(19,38): error TS2307: Cannot find module '../../../src/features/crud/pages/ListPage' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/routes.virtual.tsx(20,42): error TS2307: Cannot find module '../../../src/features/crud/pages/DetailPage' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/routes.virtual.tsx(21,52): error TS2307: Cannot find module '../../../src/features/streaming/pages/StreamingTestPage' or its corresponding type declarations.
[  Wasp  ] client/vite/virtual-files/files/routes.virtual.tsx(22,26): error TS2307: Cannot find module '../../../src/features/chat/pages/ChatPage' or its corresponding type declarations.
[  Wasp !] npm error Lifecycle script `build` failed with error:
[  Wasp !] npm error code 2
[  Wasp !] npm error path /Users/ilakovac/dev/wasp/examples/kitchen-sink/.wasp/out/sdk/wasp
[  Wasp !] npm error workspace wasp@1.0.0
[  Wasp !] npm error location /Users/ilakovac/dev/wasp/examples/kitchen-sink/.wasp/out/sdk/wasp
[  Wasp !] npm error command failed
[  Wasp !] npm error command sh -c tsc && node ./scripts/copy-assets.js

- SDK build failed with exit code: 2


❌ --- [Error] Compilation of wasp project failed: --------------------------------

1 errors found
```

If I exclude the files in the SDK's `tsconfig.json`, the TSC doesn't complain but I have to add the files to the `copy-assets.js` script to copy the source files to `dist`.

**Templating in JS with JSON from Haskell:** I think because we have a list of imports for the routes and list of head tags for `index.html` it will be kinda inefficient. We would be serializing it in Haskell so we can render it in Mustache, so we can pass into a function which will then generate the string in runtime. I dislike the added complexity.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/virtualModules.ts
infomiho: Using `/@wasp/client-entry` results in Vite not recognizing the files as TSX, so I went with `/@wasp/client-entry.tsx`.

Not doing the `path.resolve` results in:
```
[plugin:vite:import-analysis] Failed to resolve import "../../../src/App" from "../../../../../../../../../@wasp/client-entry.tsx". Does the file exist?

/@wasp/client-entry.tsx:7:31

4  |  import { getWaspApp } from "wasp/client/app";
5  |  import { routesMapping } from "/@wasp/routes.tsx";
6  |  import { App as App_ext } from "../../../src/App";
   |                                  ^
7  |  import { setup as setup_ext } from "../../../src/clientSetup";
8  |  await setup_ext();
```

infomiho: The two recommended approach are to use a filesystem based ID like the absolute path I used already or to use absolute imports `/src/App.tsx` (which doesn't work out of the box, I'd probably need to handle them in `resolveId` as well).

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html.ts
infomiho: Removing it results in:
```
vite v7.3.1 building client environment for production...
✓ 0 modules transformed.
✗ Build failed in 5ms
error during build:
Could not resolve entry module "index.html".
    at getRollupError (file:///Users/ilakovac/dev/wasp/examples/kitchen-sink/node_modules/rollup/dist/es/shared/parseAst.js:401:41)
    at error (file:///Users/ilakovac/dev/wasp/examples/kitchen-sink/node_modules/rollup/dist/es/shared/parseAst.js:397:42)
    at ModuleLoader.loadEntryModule (file:///Users/ilakovac/dev/wasp/examples/kitchen-sink/node_modules/rollup/dist/es/shared/node-entry.js:21731:20)
    at async Promise.all (index 0)
```

Wrote a comment that explains why it's needed 👍

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html.ts
infomiho: Refactored this bit of code.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html.ts
infomiho: > I think since some time now connect will catch errors in middleware.

Can you give me a source for this, I couldn't fine it?

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html.ts
infomiho: This middleware was indeed not necessary after I reverted the `appType: "custom"` config since we are still just a `"spa"` app.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html.ts
cprecioso: [https://expressjs.com/2024/10/15/v5-release.html#:\~:text=We%20don%E2%80%99t%20need%20to%20catch%20the%20error%20in%20line%20anymore%20if%20we%20want%20to%20rely%20on%20error%2Dhandling%20middleware%20instead%20because%20the%20router%20will%20now%20catch%20the%20rejected%20promise](https://expressjs.com/2024/10/15/v5-release.html#:\~:text=We%20don%E2%80%99t%20need%20to%20catch%20the%20error%20in%20line%20anymore%20if%20we%20want%20to%20rely%20on%20error%2Dhandling%20middleware%20instead%20because%20the%20router%20will%20now%20catch%20the%20rejected%20promise)

infomiho:
@cprecioso ready for another look :)

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html/build.ts
```diff
@@ -0,0 +1,44 @@
...
+import path from "node:path";
+import { getIndexHtmlContent } from "../../virtual-files/index.js";
+
+const indexHtmlFileName = "index.html";
+
+const htmlVirtualModule = {
+  id: indexHtmlFileName,
+  absPath: "",
+  getContent: () => getIndexHtmlContent(),
+};
```
cprecioso: can't we merge these two things?

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html/build.ts
```diff
@@ -0,0 +1,44 @@
+import type { Plugin } from "vite";
+import path from "node:path";
+import { getIndexHtmlContent } from "../../virtual-files/index.js";
+
+const indexHtmlFileName = "index.html";
+
+const htmlVirtualModule = {
+  id: indexHtmlFileName,
+  absPath: "",
```
cprecioso: is it needed if it's empty?

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html/build.ts
```diff
@@ -0,0 +1,44 @@
...
+  getContent: () => getIndexHtmlContent(),
+};
+
+export function waspHtmlBuild(): Plugin {
+  return {
```
cprecioso, line 14: One last thing in this pair of plugins is to give them the `apply` property to ensure they only apply to `build` or `serve`: https://vite.dev/guide/api-plugin#conditional-application

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html/build.ts
cprecioso: I'm not happy (structure-wise) with these variables: 
- going into what seems like a simple config object (and thus readonly), and modifying a property
- `absPath` can be an empty string which is misleading, it's better to have a type level nullable
- I don't mind much about the `getContent` indirection but personally I feel it's not needed.
- `absPath` should not be global but a per-plugin instance (since that is the scope where the `config` hook is valid)

So I'd have it more like this

**EDIT: Below I proposed another more generic solution that can be used here too**

```ts
import assert from "node:assert/strict";

const INDEX_HTML_FILENAME = "index.html";

function plugin() {
  let indexAbsPath: string | null = null;

  return {
    config() { indexAbsPath = path.resolve(...) },
    load(id) {
      assert(indexAbsPath);
      if (id === indexAbsPaht) { return getIndexHtmlContent(); }
    }
}
```

Or, if you don't feel that having type-level guarantees is needed:

```ts
const INDEX_HTML_FILENAME = "index.html";

function plugin() {
  let indexAbsPath!: string;

  return {
    config() { indexAbsPath = path.resolve(...) },
    load(id) {
      if (id === indexAbsPaht) { return getIndexHtmlContent(); }
    }
}
```

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/virtualModules.ts
```diff
@@ -0,0 +1,50 @@
...
+  routesEntryPoint: {
+    id: "{= routesEntryPointPath =}",
+    absPath: "",
+    getContent: () =>  getRoutesTsxContent(),
+  },
+};
+
+export function waspVirtualModules(): Plugin {
+  return {
+    name: "wasp-virtual-modules",
```
cprecioso:
```suggestion
    name: "wasp:virtual-modules",
```

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/virtualModules.ts
```diff
@@ -0,0 +1,50 @@
...
+    id: "{= clientEntryPointPath =}",
+    absPath: "",
+    getContent: () => getIndexTsxContent(),
+  },
+  routesEntryPoint: {
+    id: "{= routesEntryPointPath =}",
+    absPath: "",
+    getContent: () =>  getRoutesTsxContent(),
+  },
+};
```
cprecioso: Having these as an object makes it seem like if I wanted to add another entry file, it'd be enough to just add a property to this object, which is not the case since we're specifically referring to the objects we need.

In the interest of extracting common functionality, maybe we can have it like this (then we could also use it for `index.html`):


```ts
// plugins/util/ghost-files.ts

import * as path from "node:path";

export type GhostFiles = {
  ids: ReadonlyMap<string, string>;
  loaders: ReadonlyMap<string, () => Promise<string>>;
};

export const makeGhostFilesResolver =
  (files: { id: string; load: () => Promise<string> }[]) =>
  (rootPath: string): GhostFiles => {
    const filesWithAbsPath = files.map((d) => ({
      ...d,
      absPath: path.join(rootPath, d.id),
    }));

    const ids = new Map(filesWithAbsPath.map((d) => [d.id, d.absPath]));

    const loaders = new Map(
      filesWithAbsPath.map((d) => [d.absPath, d.load]),
    );

    return { ids, loaders };
  };
```

and usage like

```ts
const resolveGhostFiles = makeGhostFilesResolver([
  { id: "main.tsx", load: () => getMainTsx() },
  { id: "routes.tsx", load: () => getRoutesTsx() },
]);

export function plugin() {
  let ghostFiles!: GhostFiles

  return {
    configResolved: config => { ghostFiles = resolveGhostFiles(config.root) },
    resolveId: id => ghostFiles.ids.get(id),
    load: path => ghostFiles.loaders.get(path),
  }
}
```

(I decided on the name "Ghost" files, as they are not fully virtual files, i.e. they behave as if they were actually in the location we say they are -- feel free to use Virtual if you want but in my head it's slightly different meaning)

And as always, feel free to modify as you like.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/virtualModules.ts
cprecioso: Resolving in the interest of the other discussion

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/virtualModules.ts
```diff
@@ -0,0 +1,50 @@
...
+        return virtualModules.clientEntryPoint.absPath;
+      }
+      if (id === virtualModules.routesEntryPoint.id) {
+        return virtualModules.routesEntryPoint.absPath;
+      }
+    },
+    load(id) {
+      if (id === virtualModules.clientEntryPoint.absPath) {
+        const content = virtualModules.clientEntryPoint.getContent();
+        return { code: content, map: null };
```
cprecioso: You can return the straight-up string, no need for the object if you're just setting the `code` property

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/index.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export function getIndexHtmlContent(): string {
+  return getFileContentFromRelativePath("./files/index.html");
+}
+
+function getFileContentFromRelativePath(relativePath: string): string {
+  return fs.readFileSync(
+    new URL(relativePath, import.meta.url),
+    "utf-8"
+  );
```
cprecioso, line 19: This works, but in case you want to improve on it, if we ever bundle the SDK, bundlers usually try to inline both `import.meta.url`s and `fs.readFileSync`s. This indirection would not allow them to do so.

RAW

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/virtualModules.ts
```diff
@@ -1,40 +1,48 @@
...
+    id: "{= routesEntryPointPath =}",
+    absPath: "",
+    getContent: () =>  getRoutesTsxContent(),
+  },
+};
 
+export function waspVirtualModules(): Plugin {
   return {
     name: "wasp-virtual-modules",
     enforce: "pre",
```
cprecioso, line 19: This is nice, if/when we allow people to override these files, it should be as simple as changing it to `enforce: post` (IIUC)

cprecioso (commented):
As before, nice job wrangling Vite. I left a couple more suggestions on the Vite plugin side. I'd prefer for you to take it but I can drop it. Will approve then.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/virtualModules.ts
infomiho: Great suggestion, I was struggling how to shape this. I've went with a more "pedestrian" name of "virtual files" instead of "ghost files".

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html/build.ts
infomiho: Addressed this in your other comment 👍

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/virtualModules.ts
infomiho: Tried changing to `post` and adding `index.html` and a `src/main.tsx` but it didn't work of the box 😢 but we'll figure it out!

infomiho:
@cprecioso ready for another view, great stuff with the resolver idea 👍

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html/build.ts
```diff
@@ -0,0 +1,37 @@
+import type { Plugin } from "vite";
+import { getIndexHtmlContent } from "../../virtual-files/index.js";
+import { makeVirtualFilesResolver, type VirtualFiles } from "../../virtual-files/resolver.js";
+
+const indexHtmlFileName = "index.html";
```
cprecioso: nitpick, CAMEL_CASE

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html/build.ts
cprecioso: I see the 🚀 but I don't see the change, did you decide to do it?

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html/dev.ts
```diff
@@ -0,0 +1,31 @@
+import type { Plugin, Connect } from "vite";
+import { getIndexHtmlContent } from "../../virtual-files/index.js";
+
+export function waspHtmlDev(): Plugin {
+  return {
+    name: "wasp:html-dev",
+    configureServer(server) {
+      return () => {
```
cprecioso, line 9: I see we're using a "post" middleware here (https://vite.dev/guide/api-plugin#configureserver)

Is that on purpose? It seems like a "pre" should be a better fit. If the "post" is needed, can you add that as a comment (since the pattern of returning a function is not very understandable).

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html/build.ts
infomiho: Weird, I did do that change with `build` and `serve`. Hm, I'll reapply the change.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html/dev.ts
infomiho: I did it this way so that the Vite's built-in SPA fallback middleware runs first (reducing all requests to `/index.html`) and then we pick up the `/index.html` and provide the content. Does that make sense?

infomiho: I'll add a comment above the middleware 👍

PR #3610, opened 2026-01-14
Move `web-app` deps to SDK

infomiho:
## Description

Moves `web-app` deps to the SDK (slowly moving towards removing the `web-app`).

Part of #3473

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html/build.ts
+++ b/waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/html/build.ts
@@ -1,16 +1,16 @@
 import type { Plugin } from "vite";
-import path from "node:path";
 import { getIndexHtmlContent } from "../../virtual-files/index.js";
+import { makeVirtualFilesResolver, type VirtualFiles } from "../../virtual-files/resolver.js";
 
 const indexHtmlFileName = "index.html";
 
-const htmlVirtualModule = {
-  id: indexHtmlFileName,
-  absPath: "",
-  getContent: () => getIndexHtmlContent(),
-};
+const resolveVirtualFiles = makeVirtualFilesResolver([
+  { id: indexHtmlFileName, load: getIndexHtmlContent },
+]);
 
 export function waspHtmlBuild(): Plugin {
+  let virtualFiles!: VirtualFiles;
+
   return {
     name: "wasp:html-build",
     config() {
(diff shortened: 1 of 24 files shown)


waspc/src/Wasp/Generator/SdkGenerator.hs
```diff
@@ -208,6 +212,9 @@ npmDepsForSdk spec =
             ("express", expressVersionStr),
             ("mitt", "3.0.0"),
             ("react", show reactVersion),
+            ("react-dom", show reactDomVersion),
+            -- TODO: not sure if we go with peer or dev for React Query?
+            ("@tanstack/react-query", reactQueryVersion),
```
infomiho, line 216: @cprecioso could you help me with this one?

I know we moved it to peer deps for a reason: https://github.com/wasp-lang/wasp/pull/3158

cprecioso: Mmmm definitely not dev since it's used at runtime.
I'd say to move it to regular deps now actually, since the SDK is not expecting any parent package to provide this dep for them.

cprecioso: (if it gives you problems in regular deps, peer deps should keep on working fine)

cprecioso: I don't know why I said "the SDK is not expecting any parent package to provide this dep" since we force people to set React on their package 😅 I think I got confused.

This should be a peer dep, since we want to share the same one as the top package.

However:
- regular dep will probably work fine since we ensure it's the same version in both places, so npm will dedupe.
- and it's a first step towards removing all top-level deps in user's package, if we want to do that

So nbd if we keep it as regular dep, but I think it's more _correct_ to do peer dep.

infomiho: SDK is now a package in the workspace and it uses React Query directly and we don't require React Query in the user's `package.json`. Does that still make it a peer dep? I mean, it is, but it also a real dep, no?

> since we force people to set React on their package

Did you mean React or React Query?

cprecioso: > Does that still make it a peer dep? I mean, it is, but it also a real dep, no?

Yep, both are true, I think that's why we could go with either. Let's just choose one, we'll adapt if needed.

> Did you mean React or React Query?

I did mean React, but it's the same situation for RQ.

infomiho: Ok, let's keep it as a dep for now and we'll adjust along the way.

PR #3611, opened 2026-01-14
Move Vite entry point to user's project dir and remove `web-app`

infomiho:
## Description

We moved the Vite entry point to the project user dir, making the Vite integration "user-land". This meant we no longer needed the `web-app` shell which we then removed.

It's expected for the deployment to fail in the CI, we fix the deployment in the next PR: https://github.com/wasp-lang/wasp/pull/3623

Part of #3473

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/detectServerImports.ts
+++ b/waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/detectServerImports.ts
@@ -2,14 +2,14 @@
 import { type Plugin } from 'vite'
 import path from 'path'
 
-let parsePathToUserCode: ParsePathToUserCodeFn;
+let parsePathToUserCode!: ParsePathToUserCodeFn;
 
 export function detectServerImports(): Plugin {
   return {
     name: 'wasp:detect-server-imports',
     enforce: 'pre',
     configResolved(config) {
-      parsePathToUserCode = createPathToUserCodeParser(config.root);
+      parsePathToUserCode = createPathToUserCodeParser(config.root)
     },
     resolveId(source, importer) {
       if (!importer) {
--- a/waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/typescriptCheck.ts
+++ b/waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/typescriptCheck.ts
@@ -1,13 +1,32 @@
 import { type Plugin } from 'vite'
-import { execSync } from 'child_process'
+import { spawn } from 'node:child_process'
 
 export function typescriptCheck(): Plugin {
   return {
     name: 'wasp:typescript-check',
     apply: 'build',
-    buildStart() {
-      // Inherit stdio to display TypeScript errors in the terminal.
-      execSync('tsc --build --noEmit', { stdio: 'inherit' })
+    async buildStart() {
+      await runTsc()
     },
   }
+}
+
+function runTsc(): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const child = spawn(
+      'tsc',
+      ['--build', '--noEmit'],
+      {
+        stdio: 'inherit',
+        shell: process.platform === 'win32',
+      }
+    )
+
+    child.once('error', reject)
+    child.once('close', (code) =>
+      code === 0
+        ? resolve()
+        : reject(new Error(`TypeScript check failed (exit ${code})`))
+    )
+  })
 }
--- a/waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/routes.tsx
+++ b/waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/routes.tsx
@@ -4,7 +4,7 @@
 import { createAuthRequiredPage } from "wasp/client/app"
 {=/ isAuthEnabled =}
 
-// This file is used from user-land and the import paths below are relative to the
+// These files are used from user-land and the import paths below are relative to the
 // user's project dir, and not the SDK:
 {=# pagesToImport =}
 {=& importStatement =}
--- a/waspc/e2e-tests/SnapshotTest/snapshots/kitchen-sink-golden/wasp-app/.wasp/out/.waspchecksums
+++ b/waspc/e2e-tests/SnapshotTest/snapshots/kitchen-sink-golden/wasp-app/.wasp/out/.waspchecksums
@@ -690,7 +690,7 @@
             "file",
             "sdk/wasp/client/vite/plugins/detectServerImports.ts"
         ],
-        "8343c29095aa14554d0349f9089351de9bfc1377435258c62df065480caa6d96"
+        "4bebf5a19faa908bfe87960584d09107b687cbdfb90a4bd7fa65cdb9c9235870"
     ],
     [
         [
@@ -711,7 +711,7 @@
             "file",
             "sdk/wasp/client/vite/plugins/typescriptCheck.ts"
         ],
-        "4f0bfe611bd641b78fa362b7c53453272aa53db63df7958eadd5e5d59fb5ba20"
+        "737990d1ed27a0481690366da5a0c8cd7bd795a08e0984d35eac8f6a132fbd3c"
     ],
     [
         [
@@ -753,7 +753,7 @@
             "file",
             "sdk/wasp/client/vite/virtual-files/files/routes.tsx"
         ],
-        "876b428447aa38991eb87bca54fdb29292f7c1e54cff5c383ac016956d285c26"
+        "06a37b6c3c21000add6ceae5228249c836473582870465d56735237c708d93d5"
     ],
     [
         [
--- a/waspc/e2e-tests/SnapshotTest/snapshots/kitchen-sink-golden/wasp-app/.wasp/out/sdk/wasp/client/vite/plugins/detectServerImports.ts
+++ b/waspc/e2e-tests/SnapshotTest/snapshots/kitchen-sink-golden/wasp-app/.wasp/out/sdk/wasp/client/vite/plugins/detectServerImports.ts
@@ -1,14 +1,14 @@
 import { type Plugin } from 'vite'
 import path from 'path'
 
-let parsePathToUserCode: ParsePathToUserCodeFn;
+let parsePathToUserCode!: ParsePathToUserCodeFn;
 
 export function detectServerImports(): Plugin {
   return {
     name: 'wasp:detect-server-imports',
     enforce: 'pre',
     configResolved(config) {
-      parsePathToUserCode = createPathToUserCodeParser(config.root);
+      parsePathToUserCode = createPathToUserCodeParser(config.root)
     },
     resolveId(source, importer) {
       if (!importer) {
--- a/waspc/e2e-tests/SnapshotTest/snapshots/kitchen-sink-golden/wasp-app/.wasp/out/sdk/wasp/client/vite/plugins/typescriptCheck.ts
+++ b/waspc/e2e-tests/SnapshotTest/snapshots/kitchen-sink-golden/wasp-app/.wasp/out/sdk/wasp/client/vite/plugins/typescriptCheck.ts
@@ -1,13 +1,32 @@
 import { type Plugin } from 'vite'
-import { execSync } from 'child_process'
+import { spawn } from 'node:child_process'
 
 export function typescriptCheck(): Plugin {
   return {
     name: 'wasp:typescript-check',
     apply: 'build',
-    buildStart() {
-      // Inherit stdio to display TypeScript errors in the terminal.
-      execSync('tsc --build --noEmit', { stdio: 'inherit' })
+    async buildStart() {
+      await runTsc()
     },
   }
+}
+
+function runTsc(): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const child = spawn(
+      'tsc',
+      ['--build', '--noEmit'],
+      {
+        stdio: 'inherit',
+        shell: process.platform === 'win32',
+      }
+    )
+
+    child.once('error', reject)
+    child.once('close', (code) =>
+      code === 0
+        ? resolve()
+        : reject(new Error(`TypeScript check failed (exit ${code})`))
+    )
+  })
 }
--- a/waspc/e2e-tests/SnapshotTest/snapshots/kitchen-sink-golden/wasp-app/.wasp/out/sdk/wasp/client/vite/virtual-files/files/routes.tsx
+++ b/waspc/e2e-tests/SnapshotTest/snapshots/kitchen-sink-golden/wasp-app/.wasp/out/sdk/wasp/client/vite/virtual-files/files/routes.tsx
@@ -1,7 +1,7 @@
 // @ts-nocheck
 import { createAuthRequiredPage } from "wasp/client/app"
 
-// This file is used from user-land and the import paths below are relative to the
+// These files are used from user-land and the import paths below are relative to the
 // user's project dir, and not the SDK:
 import { HomePage } from './src/pages/HomePage'
 import { CatchAllPage } from './src/pages/CatchAllPage'
--- a/waspc/e2e-tests/SnapshotTest/snapshots/wasp-build-golden/wasp-app/.wasp/out/.waspchecksums
+++ b/waspc/e2e-tests/SnapshotTest/snapshots/wasp-build-golden/wasp-app/.wasp/out/.waspchecksums
@@ -270,7 +270,7 @@
             "file",
             "sdk/wasp/client/vite/plugins/detectServerImports.ts"
         ],
-        "8343c29095aa14554d0349f9089351de9bfc1377435258c62df065480caa6d96"
+        "4bebf5a19faa908bfe87960584d09107b687cbdfb90a4bd7fa65cdb9c9235870"
     ],
     [
         [
@@ -291,7 +291,7 @@
             "file",
             "sdk/wasp/client/vite/plugins/typescriptCheck.ts"
         ],
-        "4f0bfe611bd641b78fa362b7c53453272aa53db63df7958eadd5e5d59fb5ba20"
+        "737990d1ed27a0481690366da5a0c8cd7bd795a08e0984d35eac8f6a132fbd3c"
     ],
     [
         [
@@ -333,7 +333,7 @@
             "file",
             "sdk/wasp/client/vite/virtual-files/files/routes.tsx"
         ],
-        "e100525c32a033a26126a9b502daf98e0d76dbcccb371e3aa54fddaca0470d14"
+        "0ead25c5f37895afa471357c298fab3a0593fe4e12b1b7e9b404f7f7fe375d1a"
     ],
     [
         [
--- a/waspc/e2e-tests/SnapshotTest/snapshots/wasp-build-golden/wasp-app/.wasp/out/sdk/wasp/client/vite/plugins/detectServerImports.ts
+++ b/waspc/e2e-tests/SnapshotTest/snapshots/wasp-build-golden/wasp-app/.wasp/out/sdk/wasp/client/vite/plugins/detectServerImports.ts
@@ -1,14 +1,14 @@
 import { type Plugin } from 'vite'
 import path from 'path'
 
-let parsePathToUserCode: ParsePathToUserCodeFn;
+let parsePathToUserCode!: ParsePathToUserCodeFn;
 
 export function detectServerImports(): Plugin {
   return {
     name: 'wasp:detect-server-imports',
     enforce: 'pre',
     configResolved(config) {
-      parsePathToUserCode = createPathToUserCodeParser(config.root);
+      parsePathToUserCode = createPathToUserCodeParser(config.root)
     },
     resolveId(source, importer) {
       if (!importer) {
--- a/waspc/e2e-tests/SnapshotTest/snapshots/wasp-build-golden/wasp-app/.wasp/out/sdk/wasp/client/vite/plugins/typescriptCheck.ts
+++ b/waspc/e2e-tests/SnapshotTest/snapshots/wasp-build-golden/wasp-app/.wasp/out/sdk/wasp/client/vite/plugins/typescriptCheck.ts
@@ -1,13 +1,32 @@
 import { type Plugin } from 'vite'
-import { execSync } from 'child_process'
+import { spawn } from 'node:child_process'
 
 export function typescriptCheck(): Plugin {
   return {
     name: 'wasp:typescript-check',
     apply: 'build',
-    buildStart() {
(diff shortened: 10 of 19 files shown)


examples/websockets-realtime-voting/src/Main.css
```diff
@@ -1,4 +1,5 @@
 @import "tailwindcss" source(".");
+@source "../.flowbite-react/class-list.json";
```
infomiho, line 2: Flowbite added this automatically 🤷

wasp-app-runner/src/versions.ts
```diff
@@ -9,7 +9,13 @@ export interface VersionSettings {
 export const getSettingsForVersion = (
   version: WaspVersion,
 ): VersionSettings => {
-  if (semver.lt(version, "0.21.0")) {
+  // TODO: use `semver.lt(version, "0.22.0")` after we bump
```
infomiho: I'll address this at the end, after I know which version this will be used in. Promise not to forget 🤞

waspc/cli/src/Wasp/Cli/Command/BuildStart/Client.hs
```diff
@@ -17,24 +15,23 @@ buildClient :: BuildStartConfig -> ExceptJob
 buildClient config =
   runNodeCommandAsJobWithExtraEnv
     envVars
-    webAppDir
-    "npm"
-    ["run", "build"]
+    projectDir
+    -- TODO: Missing `tsc --build` from the original `npm run build`.
```
infomiho: Not sure how we should deal with this? We had `web-app`'s `package.json` file with the `build` script which contained both commands: `tsc --build && npx vite build`. 

We could require now in the user's `pacakge.json` something like `build:client` and `start:client` scripts? But this is weird because user's might run those script instead of `wasp start`.

waspc/src/Wasp/Generator/WebAppGenerator/Test.hs
```diff
@@ -0,0 +1,13 @@
...
+  ( testWebApp,
+  )
+where
+
+import StrongPath (Abs, Dir, Path')
+import qualified Wasp.Job as J
+import Wasp.Job.Process (runNodeCommandAsJob)
+import Wasp.Project.Common (WaspProjectDir)
+
+testWebApp :: [String] -> Path' Abs (Dir WaspProjectDir) -> J.Job
```
infomiho, line 11: This is migrated from the `WebAppGenerator` for the `wasp test client` command. We'll be able to move Vitest to user-land now that Vite is user-land. But I'll do that in some future PR.

waspc/src/Wasp/Project/WebApp/Common.hs
```diff
@@ -0,0 +1,29 @@
+module Wasp.Project.WebApp.Common
```
infomiho: Moved stuff from `Generator.WebAppGenerator.Common` to `Project.WebApp.Common`. Felt okay, it's kinda web app project information and not necessarily tied to a web app generator anymore.

waspc/src/Wasp/Generator/SdkGenerator/Client/VitePluginG.hs
```diff
@@ -46,18 +45,11 @@ genWaspPlugin spec = return $ C.mkTmplFdWithData tmplPath tmplData
           "defaultClientPort" .= WebApp.defaultClientPort,
           "vitest"
             .= object
-              [ "setupFilesArray" .= makeJsArrayFromHaskellList vitestSetupFiles,
+              [ -- Loading the setup fn from the SDK
+                "setupFilesArray" .= makeJsArrayFromHaskellList ["wasp/client/test/setup"],
```
infomiho: This bit will also go after we move Vitest to user-land, but for now it's loaded from the SDK. It used to be a file in `web-app` dir.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/wasp.ts
```diff
@@ -44,22 +46,26 @@ export function wasp(options?: WaspPluginOptions): Plugin[] {
...
+                //
                 // Vite doesn't look for `.prisma/client` imports in the `node_modules`
                 // folder. We point it to the correct place here.
                 // TODO: Check if we can remove when updating Prisma (#2504)
-                find: /^\.prisma\/(.+)$/,
-                replacement: path.join(
-                  resolveProjectPath("./"),
-                  "node_modules/.prisma/$1"
-                ),
+                find: /^\.prisma\/client\/(.+)$/,
```
infomiho, line 53: Saw this error, so I adjusted the code:
```
[commonjs--resolver] Could not load node_modules/.prisma/client/index-browser:
ENOENT: no such file or directory, open 'node_modules/.prisma/client/index-browser'
```

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/wasp.ts
```diff
@@ -35,7 +33,11 @@ export function wasp(options?: WaspPluginOptions): Plugin[] {
           },
           envPrefix: "REACT_APP_",
           build: {
-            outDir: "build",
+            // TODO: decide if we want to keep the old build dir or introduce
```
infomiho: Resolved this TODO in the next [PR](https://github.com/wasp-lang/wasp/pull/3623/changes#diff-e8b56293bc36393de972160932c3e2d2bbdf3271a002ffb662d0c2c42a945942L36).

examples/kitchen-sink/public/manifest.json
cprecioso, line 1: How come this file is shown as moved?

waspc/e2e-tests/snapshots/wasp-build-golden/wasp-app/.wasp/out/sdk/wasp/client/vite/virtual-files/files/routes.tsx
```diff
@@ -1,6 +1,6 @@
 // @ts-nocheck
```
cprecioso: We could add a comment before the imports in the template, because this tripped me up until I realized:

```ts
// The relative path in imports in this file will work as if they were in the users project dir.
```

or something like that

waspc/e2e-tests/snapshots/wasp-build-golden/wasp-app/.wasp/out/sdk/wasp/client/vite/plugins/detectServerImports.ts
```diff
@@ -43,10 +42,7 @@ function parsePathToUserCode(
...
 }
 
-// We can't pass the "waspProjectDir" path from Haskell because we need the absolute path:
-// e.g. /Users/{username}/dev/wasp/waspc/examples/todoApp
-// which contains machine specific info like the username which is different in the CI and locally.
-// This breaks our e2e tests in the CI because the path is different.
+
 function getWaspProjectDirAbsPathWhileInWebAppDir(): string {
-  return path.resolve(resolveProjectPath('./'))
+  return path.resolve('.')
```
cprecioso: mmm... I understand the helper is not needed because now `cwd` is the users project, but maybe we keep it anyway? I'm a bit afraid of the implicitness of this.

waspc/data/Generator/templates/react-app/public/manifest.json
cprecioso, line 1: Should we tell people to re-create this file in their projects?

waspc/data/Generator/templates/react-app/src/vite-env.d.ts
cprecioso, line 1: How do users get the vite env now? (i think this was done in a previous PR but can't exactly remember)

waspc/data/Generator/templates/react-app/netlify.toml
cprecioso, line 1: Are we losing functionality by removing this here?

waspc/data/Generator/templates/react-app/npmrc
cprecioso, line 1: This looks sensible. Do we miss that?

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/wasp.ts
cprecioso: what is the resolution? you kept the old dir, right?

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/wasp.ts
cprecioso: damn we really have to upgrade our prisma

waspc/cli/src/Wasp/Cli/Command/BuildStart/Client.hs
cprecioso: It seems simple to me, so I might be missing something: add `tsc --build` also as a command call here too, no?

waspc/src/Wasp/Generator/SdkGenerator/Client/VitePluginG.hs
```diff
@@ -46,18 +45,10 @@ genWaspPlugin spec = return $ C.mkTmplFdWithData tmplPath tmplData
           "defaultClientPort" .= WebApp.defaultClientPort,
           "vitest"
             .= object
-              [ "setupFilesArray" .= makeJsArrayFromHaskellList vitestSetupFiles,
+              [ "setupFilesArray" .= makeJsArrayFromHaskellList ["wasp/client/test/setup"],
```
cprecioso, line 56: Feels weird to have it hardcoded, but I see that it was like that before so RAW

waspc/src/Wasp/Project/WebApp/Common.hs
cprecioso: I don't agree. These file seems tied to the fact that we're a React app on the npm ecosystem. I'm going by the rule of thumb (i think you told me of it) that this should only fit outside of Generator if it would still make sense if we ever changed wasp to output e.g. python or desktop apps.

waspc/src/Wasp/Project/Analyze.hs
```diff
@@ -101,14 +98,12 @@ constructAppSpec waspDir compileOptions externalConfigs parsedPrismaSchema decls
             AS.prismaSchema = parsedPrismaSchema,
             AS.waspProjectDir = waspDir,
             AS.externalCodeFiles = externalCodeFiles,
-            AS.externalPublicFiles = externalPublicFiles,
```
cprecioso, line 104: Happy to get rid of these!

waspc/src/Wasp/Generator.hs
```diff
@@ -52,8 +51,7 @@ writeWebAppCode spec dstDir sendMessage = do
 
 genApp :: AppSpec -> Generator [FileDraft]
 genApp spec =
-  genWebApp spec
```
cprecioso, line 61: Damn. Feels weird 😁

waspc/src/Wasp/Generator/NpmDependencies.hs
```diff
@@ -80,19 +79,17 @@ ensureNoConflictWithUserDeps waspDeps userDeps
...
     conflicts = getNpmDepsConflicts waspDeps userDeps
 
-buildWaspFrameworkNpmDeps :: AppSpec -> NpmDepsFromWasp -> NpmDepsFromWasp -> Either String NpmDepsForFramework
-buildWaspFrameworkNpmDeps spec fromServer fromWebApp
+buildWaspFrameworkNpmDeps :: AppSpec -> NpmDepsFromWasp -> Either String NpmDepsForFramework
+buildWaspFrameworkNpmDeps spec fromServer
   | hasConflicts = Left "Could not construct npm dependencies due to a previously reported conflict."
   | otherwise =
       Right $
         NpmDepsForFramework
```
cprecioso: Do we need the `NpmDepsForFramework` struct anymore then?

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/wasp.ts
infomiho: Yes, we kept the `web-app` dir _only for the purposes of deployment_.

examples/kitchen-sink/public/manifest.json
infomiho: There is no more `web-app/public` to keep this file. Users will need to create this file in their `public` dir from now on to keep things as they were.

waspc/e2e-tests/snapshots/wasp-build-golden/wasp-app/.wasp/out/sdk/wasp/client/vite/plugins/detectServerImports.ts
infomiho: What is scaring you about this, can you maybe put it into words so I understand better? :)

waspc/data/Generator/templates/react-app/public/manifest.json
infomiho: Yep, this will be the part of the migration guide for sure 👍

waspc/data/Generator/templates/react-app/src/vite-env.d.ts
infomiho: `./src/vite-env.d.ts` comes with every starter:
```ts

// This is needed to properly support Vitest testing with jest-dom matchers.
// Types for jest-dom are not recognized automatically and Typescript complains
// about missing types e.g. when using `toBeInTheDocument` and other matchers.
// Reference: https://github.com/testing-library/jest-dom/issues/546#issuecomment-1889884843
import "@testing-library/jest-dom";
```

waspc/data/Generator/templates/react-app/netlify.toml
infomiho: Yes, we are losing some Netlify specific config. I wanted to get rid of this file for some time already, but this is a good break point for it. I'll need to update the deployment docs to tell users to add a `netlify.toml` in the `build` dir https://wasp.sh/docs/deployment/deployment-methods/paas#netlify

cprecioso: it seems nicely scoped to do that in a preceding PR to this one

waspc/cli/src/Wasp/Cli/Command/BuildStart/Client.hs
cprecioso: mmm just got the idea, can we have tsc checking _inside_ of vite? this way we also don't need to remember to call `tsc --build` e.g. in the deployment scripts

waspc/data/Generator/templates/react-app/npmrc
infomiho: I think it was a nice to have convenience - but I'm not sure how much it did for users since we already validate Node.js with the compiler. If you ran `wasp build` and then `npm run build` chances are you didn't switch Node.js versions.

If we feel like this is something we want to keep, we can include it in the starter templates by default in the root dir. I'm not convinced we should keep it 🤷

waspc/cli/src/Wasp/Cli/Command/BuildStart/Client.hs
infomiho: Is that even legal? 😂 I'd love to hear more!

wasp-app-runner/src/versions.ts
infomiho: I'm happy we got rid of this 😄

waspc/e2e-tests/snapshots/wasp-build-golden/wasp-app/.wasp/out/sdk/wasp/client/vite/plugins/detectServerImports.ts
cprecioso: `path.resolve()` resolves from `cwd` if not abs path is given. So with this change we're adding an implicit dependency on where this is being called from.

I realize that the whole system is dependent on that, but I like that that only happens on the top-level (where Vite is being called from), but the rest of our stuff takes on from that top-level dependency and doesn't use it anymore later (reading from `configResolved` or `resolveProjectPath`). Those are known quantities that we can look for and update if our setup ever changes.

So Id' like to make the implicitness explicit (with resolveProjectPath), or defer it to a higher level (reading from configResolved)

waspc/cli/src/Wasp/Cli/Command/BuildStart/Client.hs
cprecioso:
```ts
const myVitePlugin = { apply: "build", buildStart: () => exec("tsc --build") }
```
😁

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/wasp.ts
cprecioso: future work, we actually might not need this anymore since we're calling from the user dir. no need to explore now tho.

waspc/data/Generator/templates/react-app/netlify.toml
infomiho: I've made a PR to `main` here: https://github.com/wasp-lang/wasp/pull/3665

waspc/src/Wasp/Generator/NpmDependencies.hs
infomiho: Yep, no need for the indirection, I've dropped it 👍

waspc/e2e-tests/snapshots/wasp-build-golden/wasp-app/.wasp/out/sdk/wasp/client/vite/plugins/detectServerImports.ts
infomiho: Gotcha, I've reworked the plugin to use the `config.root` 😄

waspc/src/Wasp/Project/WebApp/Common.hs
infomiho: Yep, good call. I've needed to reintroduce the `WebAppGenerator` for deployment anyways, so I moved these modules back to `WebAppGenerator` scope.

infomiho:

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/detectServerImports.ts
```diff
@@ -1,12 +1,16 @@
 {{={= =}=}}
 import { type Plugin } from 'vite'
 import path from 'path'
-import { resolveProjectPath } from 'wasp/dev'
+
+let parsePathToUserCode: ParsePathToUserCodeFn;
```
cprecioso: Just one more thing, as with the other plugins, this is semantically in the plugin instance scope, not the global scope. So I'd put this `let` into the `detectServerImports` function.

You can also do the same `!` trick we did in other plugins

```suggestion
let parsePathToUserCode!: ParsePathToUserCodeFn;
```

Then you can drop the `if (!parsePathToUserCode` check.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/detectServerImports.ts
```diff
@@ -32,22 +36,16 @@ function isServerImport(moduleName: string): boolean {
...
-// This breaks our e2e tests in the CI because the path is different.
-function getWaspProjectDirAbsPathWhileInWebAppDir(): string {
-  return path.resolve(resolveProjectPath('./'))
+function createPathToUserCodeParser(waspProjectDirPath: string): ParsePathToUserCodeFn {
+  return (importerPath: string): RelativePathToUserCode | null => {
+    const importerPathRelativeToWaspProjectDir = path.relative(
+      waspProjectDirPath,
+      importerPath
+    )
+    return importerPathRelativeToWaspProjectDir.startsWith('{= srcDirInWaspProjectDir =}')
```
cprecioso, line 47: Given we just made it relative, will it start with `{= srcDirInWaspProjectDir =}` or with `"."`?

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/typescriptCheck.ts
```diff
@@ -0,0 +1,13 @@
+import { type Plugin } from 'vite'
+import { execSync } from 'child_process'
+
+export function typescriptCheck(): Plugin {
+  return {
+    name: 'wasp:typescript-check',
+    apply: 'build',
+    buildStart() {
+      // Inherit stdio to display TypeScript errors in the terminal.
+      execSync('tsc --build --noEmit', { stdio: 'inherit' })
```
cprecioso: suggestion, let's make it async, this way we don't block vite from doing more work in the background to prepare the build

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/routes.tsx
```diff
@@ -4,6 +4,8 @@
 import { createAuthRequiredPage } from "wasp/client/app"
 {=/ isAuthEnabled =}
 
+// This file is used from user-land and the import paths below are relative to the
+// user's project dir, and not the SDK:
```
cprecioso: nitpick: this seems like it would go _inside_ the `{=# pagesToImport =}` block

cprecioso (approved):
Some final nitpicks but good otherwise

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/detectServerImports.ts
infomiho: I will still start with `src` since all of user code lives in `src`. We made it relative to the project root e.g. `~/dev/my-app` so if the import is coming from `~/dev/my-app/src/Todo.tsx` this will produce `src/Todo.tsx` 😄

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/routes.tsx
infomiho: If I did that it would repeat the comment for each import, I've reworded the comment to be in plural so the placement now makes more sense.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/detectServerImports.ts
infomiho: We don't have that check now, the check is for the path itself:
```ts
const pathToUserCode = parsePathToUserCode(importer)
if (!pathToUserCode) {
  return
}
```

I implemented the change to be consistent with other Vite plugins 👍

cprecioso: @infomiho you forgot moving it down to _inside_ the plugin

PR #3612, opened 2026-01-14
Human readable info on `wasp info`

cprecioso:
## Description

With #3539, I didn't realize I was outputting raw JSON on the `wasp info` command (since it outputs directly what it reads from the file). I fixed it and edited the message a bit.

**Wasp 0.20.0** (old, human)
```
$ wasp info
Project information
Name: project
Database system: SQLite
Last compile: Generated on 2026-01-14 16:25:14.73091 UTC by waspc version "0.20.0" .
Project dir size: 292 MB
```

**Wasp 0.20.1** (new, json)
```
$ wasp info
Project information
Name: project
Database system: SQLite
Last compile: {"buildType":"Development","generatedAt":"2026-01-14T16:28:06.428171Z","waspVersion":"0.20.1"}
Project dir size: 292 MB
```

**This PR** (new, human)
```
$ wasp info
Project information
Name: project
Database system: SQLite
Last compile: Development build, generated at 2026-01-14 16:28:46.951633 UTC, by Wasp 0.21.0.
Project dir size: 222 MB
```

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:


waspc/cli/src/Wasp/Cli/Command/Info.hs
```diff
@@ -5,28 +5,35 @@ where
...
 import qualified Wasp.Util.Terminal as Term
 
 info :: Command ()
 info = do
   InWaspProject waspDir <- require
 
-  compileInfo <- liftIO $ readWaspCompileInfo waspDir
   projectSize <- liftIO $ readDirectorySizeMB waspDir
 
+  let generatedCodeDir =
```
Martinsos: It was nicer when all these details were in a helper function, as it was before. How is it that you gave up on that one and decided to inline it all here? If there is a reason you don't like it there in Common, I wonder if we could have a helper function at least here in this file.
Clean code would say that you went too much into detail here, going level of abstraction deeper than we want to see in this function, and I would agree, it makes it much harder to read it quickly.

waspc/src/Wasp/Generator/WaspInfo.hs
```diff
@@ -65,3 +67,13 @@ safeRead projectRootDir = do
   doesFileExist waspInfoFile >>= \case
     False -> return Nothing
     True -> decodeFileStrict $ toFilePath waspInfoFile
+
+showWaspInfo :: WaspInfo -> String
```
Martinsos: This could potentially be moved to Command/Info.hs, as this output is basically what that commands wants it to be. It is not terrible that it is here though, just I would be tempted to push it there I think. I see both sides though.

raw

waspc/src/Wasp/Generator/WaspInfo.hs
```diff
@@ -65,3 +67,13 @@ safeRead projectRootDir = do
...
+showWaspInfo waspInfo =
+  intercalate
+    ", "
+    [ unwords [show $ buildType waspInfo, "build"],
+      unwords ["generated at", show $ generatedAt waspInfo],
+      unwords ["by Wasp", waspVersion waspInfo]
+    ]
+    ++ "."
```
Martinsos: I have to admit I find this hard to parse: well, harded than I imagine it should be. `unwords` throw me off because they are used just for connecting two strings each time.

Would be nice if I could read this whil eundrestanding what it is going to look like.

e.g.
```hs
showWaspInfo waspInfo =
  [trimming|
    ${buildType'} build, generated at ${generatedAt'}, by Wasp ${waspVersion'}
  |]
  where
    buildType' = T.pack $ show $ buildType waspInfo
    generatedAt' = T.pack $ show $ generatedAt waspInfo
    waspVersion' = T.pack $ waspVersion waspInfo
```

Note: I didn't use `'` as suffix because I had to, I just found them convenient, and it is kind of an idiom in Haskell.

This whole suggestion: you don't have to do it, it is small piece of code, and it's not a big deal. But I personally do find this approach more appealing because now, I can at glance know what the text is going to look like.

Btw if you are wondering why do we have to turn it into Text -> that is just how this `trimming` imported function works. I haven't found alternative that accepts String. Haskellers claim String is a mistake and you should use Text for anything remotely serious hehe, so probalby that is why (in practice String is ok for small stuff and as long as texts are not long, but if they are or you care more about encodings or sometihng, then yeah you should use Text. Never had a single problem with String so far though).

RAW

Martinsos (approved):
@cprecioso Nice catch!
I approved it, but, check out the comments. I am sure I don't need to check afterwards though, so that is why I approved immedaitely.

Oh I do have one question: would it make any sense to have a cli e2e test that tests the output of `wasp info`? I am thinking onw that maybe it might. First of all, it would have caught this mistake earlier. Second, I guess we do kind of make this part of public API the moment we have it in command hah, so why not test it. Third, I think it should be very simple to write?

waspc/src/Wasp/Generator/WaspInfo.hs
cprecioso: I had actually done it with the quasiquote in the beginning, but I was afraid it would get flagged up in review so I did it like this. I will revert to the quasiquote.

cprecioso:
> have a cli e2e test that tests the output of wasp info?

yeah, as i understood that is coming in https://github.com/wasp-lang/wasp/pull/3465, right?

Martinsos:
> > have a cli e2e test that tests the output of wasp info?
> 
> yeah, as i understood that is coming in #3465, right?

Nice didn't know that, great!

waspc/src/Wasp/Generator/WaspInfo.hs
```diff
@@ -51,17 +53,23 @@ persist projectRootDir currentBuildType = do
...
   where
     isCompatible (WaspInfo {waspVersion = storedVersion, buildType = storedBuildType}) =
       (storedVersion == currentVersion) && (storedBuildType == currentBuildType)
 
     currentVersion = showVersion Paths_waspc.version
 
-safeRead :: Path' Abs (Dir ProjectRootDir) -> IO (Maybe WaspInfo)
-safeRead projectRootDir = do
-  let waspInfoFile = projectRootDir </> waspInfoInProjectRootDir
+type ReadResult = Either ReadError WaspInfo
```
Martinsos, line 63: I would probably skip this type alias, it obscures things a bit too much. It is easier to have Either in that return type and see what is happening. Doesn't matter for this PR, dont' bother with it, I am just commenting for the future.

waspc/cli/src/Wasp/Cli/Command/Common.hs
```diff
@@ -7,32 +7,42 @@ where
...
-  ifM
-    (IOUtil.doesFileExist dotWaspInfoFile)
-    (IOUtil.readFile dotWaspInfoFile)
-    (return "No compile information found")
+  either showError showWaspInfo
+    <$> WI.safeRead generatedCodeDir
   where
-    dotWaspInfoFile =
+    showError WI.NotFound = "No compile information found"
+    showError WI.IncompatibleFormat = "Incompatible compile information"
```
Martinsos, line 30: One might argue that it woudl be jnicer if this function, readCompileInfo, returned actual CompileInfo type, and then the user of this function can decide what to print in case of these errors. But ok, this also works, it's not a biggie, its hard to say what exactly is better in this situation.

Writing tihs just as a thought for future, don't do anything here.

Martinsos (commented):
Left two observations for the future, you don't have to do anything.

waspc/cli/src/Wasp/Cli/Command/Common.hs
Martinsos: Aha but I see now you also put the whole showWaspInfo here, ok got it.

PR #3505, opened 2025-12-16
Setting up SDK workspace infrastructure

FranjoMindek:
## Description

Part of issue: https://github.com/wasp-lang/wasp/issues/2247
More specifically, part 1 of: https://github.com/wasp-lang/wasp/issues/2247#issuecomment-3647334765

Confirmed working: 
- `examples/kitchen-sink`
	- `wasp compile` ✅
	- `wasp start` ✅
	- `wasp build` ✅
	- `wasp build start` ✅

## Type of change

- [ ] **🔧 Just code/docs improvement** 
- [ ] **🐞 Bug fix** 
- [ ] **🚀 New/improved feature** 
- [ ] **💥 Breaking change** 

## Checklist

- [ ] I tested my change in a Wasp app to verify that it works as intended.

- 🧪 Tests and apps:

  - [ ] I added **unit tests** for my change. 
  - [ ] _(if you fixed a bug)_ I added a **regression test** for the bug I fixed. 
  - [ ] _(if you added/updated a feature)_ I added/updated **e2e tests** in `examples/kitchen-sink/e2e-tests`.
  - [ ] _(if you added/updated a feature)_ I updated the **starter templates** in `waspc/data/Cli/templates`, as needed.
  - [ ] _(if you added/updated a feature)_ I updated the **example apps** in `examples/`, as needed.
    - [ ] _(if you updated `examples/tutorials`)_ I updated the tutorial in the docs (and vice versa).

- 📜 Documentation:

  - [ ] _(if you added/updated a feature)_ I **added/updated the documentation** in `web/docs/`.

- 🆕 Changelog: _(if change is more than just code/docs improvement)_
  - [ ] I updated `waspc/ChangeLog.md` with a **user-friendly** description of the change.
  - [ ] _(if you did a breaking change)_ I added a step to the current **migration guide** in `web/docs/migration-guides/`.
  - [ ] I **bumped the `version`** in `waspc/waspc.cabal` to reflect the changes I introduced.

PR #3618, opened 2026-01-15
Implement tsconfig project references infrastructure for SDK

### LINKED ISSUE #3517: Implement tsconfig project references infrastructure for SDK
Part of https://github.com/wasp-lang/wasp/issues/2247
This issue addresses points `1.` and `2.1.` from the following comment: https://github.com/wasp-lang/wasp/issues/2247#issuecomment-3647334765

We should:
1. Create SDK tsconfig project references infrastructure.
(one root project, sub-projects, build works, runtime works)
2. Refactors SDK pathing to support tsconfig project references.
(ideally arbitrary number of subprojects, i.e. easily extensible)
3. Create a minimal _tracer bullet_ implementation. 
(e.g. move `HttpError` to `core` subproject).

The following PR can be taken as inspiration, as it implements points `1.` and `3.`: https://github.com/wasp-lang/wasp/pull/3505

Estimate: 1 day

FranjoMindek:
## Description

Built on top of https://github.com/wasp-lang/wasp/pull/3505
Solves https://github.com/wasp-lang/wasp/issues/3517

Implements infrastructure for multiple tsconfig project SDK.
The new structure is:
```
└── sdk/wasp               # `root` project
    ├── tsconfig.json      
    ├── core               # `core` project
    │   └── tsconfig.json  
    ├── user-core          # `user-core` project
    │   └── tsconfig.json  
    ├── src                # part of `user-core` project
    └── dist               # notice that `dist` follows the source folder structure
        ├── core
        ├── src
        └── user-core
```
The `root` project hold scripts and references all other projects. 
It is the starting point of the SDK.

The `core` project will hold all code that does not depend on user's code (`src`) in any way.

The `user-core` project is sdk code that does depend on the user's code.
We can discuss how to actually name this project, changing the name is easy.

Even though the we the structure suggest that we have 4 separate projects (`root`, `core`, `user-core`, and `src`), `src` is actually part of the `user-core` project (for now). 

This PR features the following changes:
- Setting up the multiple tsconfig projects setup (sdk/wasp/tsconfig, sdk/wasp/core/tsconfig.json, sdk/wasp/user-core/tsconfig.json).
- Refactored `strong-path` infrastructure for multiple tsconfig project:
  - `SdkGenerator/Common.hs` - Shared utilities across all SDK projects
  - `SdkGenerator/Root/Common.hs` - Helpers for the root tsconfig project
  - `SdkGenerator/Core/Common.hs` - Helpers for the core tsconfig project
  - `SdkGenerator/UserCore/Common.hs` - Helpers for the user-core tsconfig project
  NOTE: I plan to move files used by the specific tsconfig projects into their respective folders in future PRs. E.g. `HttpError.ts` generator will be in `SdkGenerator/Core` folder.
  Edit: starting PR here https://github.com/wasp-lang/wasp/pull/3653
  
- Removed old server/client path helpers (they were used inconsistently, and now with multiple projects its more tedious to use them).
- Refactoring all `SdkGenerator/**/*.hs` files to use the new infrastructure.
- Moving most of `sdk/wasp` files to `sdk/wasp/user-core`, except for `HttpError.ts` which was moved to `sdk/wasp/core`.
  - This move automatically "formatted imports" on my IDE, so some files have extensions "optimized". I reverted all pure sorting changes to reduce the amount of changes.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:


waspc/data/Generator/templates/sdk/wasp/package.json
```diff
@@ -5,148 +5,144 @@
...
+    "./auth/pages/createAuthRequiredPage": "./dist/user-core/auth/pages/createAuthRequiredPage.jsx",
+    {=!  Used by our framework code (Websockets), undocumented (but accessible) for users. =}
+    "./api/events": "./dist/user-core/api/events.js",
+    {=!  Used by users, documented. =}
+    "./operations": "./dist/user-core/operations/index.js",
+    {=!  If we import a symbol like "import something form 'wasp/something'", we must =}
+    {=!  expose it here (which leaks it to our users). We could avoid this by =}
+    {=!  using relative imports inside SDK code (instead of library imports), =}
+    {=!  but I didn't have time to implement it. =}
+    "./src/*": "./dist/src/*.js",
```
FranjoMindek, line 51: Notice that `./src/*` is the only one that did not change the export to `./dist/user-core`.

This is because `src` did not change location, even though it is not a standalone tsconfig project (for now).

We have the following structure:

Root tsconfig project -> depends on all
`sdk/wasp` 
Core tsconfig project
`sdk/wasp/core`
User core tsconfig project -> depends on core
`sdk/wasp/user-core`
`sdk/wasp/src`

FranjoMindek (commented):
Comments migrated from PR #3606

waspc/data/Generator/templates/sdk/wasp/package.json
```diff
@@ -5,148 +5,144 @@
...
-      {=!  Used by our code, undocumented (but accessible) for users. =}
-      "./server/jobs/*": "./dist/server/jobs/*.js",
-      {=!  Used by our code, undocumented (but accessible) for users. =}
-      {=!  Todo(filip): This export becomes problematic once we start supporting different executors =}
-      "./server/jobs/core/pgBoss": "./dist/server/jobs/core/pgBoss/index.js",
+    {=! todo(filip): Check all exports when done with SDK generation =}
+    {=!  Some of the statements in the comments might become incorrect. =}
+    {=!  "our code" means: "web-app", "server" or "SDK", or "some combination of the three". =}
+    {=!  Used by our code, undocumented (but accessible) for users. =}
+    "./core/storage": "./dist/user-core/core/storage.js",
```
FranjoMindek, line 19: How to handle exports once we actually split all of the files?

We will probably have 3 situations:
1. Old `sdk/wasp/` file is moved fully to `core/`
	We can either export through `./dist/core/` or we can force the exports through `./dist/user-core/` by re-exporting.
	
2. Old `sdk/wasp/` file is moved fully to `user-core/`
	We will probably export through `./dist/user-core/`

3. Old `sdk/wasp/` file is split. It is partially moved to `core/` and partially moved to `user-core/`.
	We will probably re-export `core/` stuff in `user-core/` and point to `./dist/user-core`.
	Like with `HttpError` above.

Filip mentioned: Maybe a separate tsconfig project/dir for the public API?

waspc/data/Generator/templates/sdk/wasp/user-core/server/index.ts
```diff
@@ -7,7 +7,7 @@ export { default as prisma, type PrismaClient } from './dbClient.js'
 // PUBLIC API
 export { type ServerSetupFn } from './types/index.js'
 // PUBLIC API
-export { HttpError } from './HttpError.js'
+export { HttpError } from '../../core/server/HttpError.js'
```
FranjoMindek, line 10: Problem: `package.json` does not support the array options for `exports` field.
E.g., `"./server": ["./dist/core/server/index.js", "./dist/user-core/server/index.js"],` is not possible.

This means that when we split `sdk/wasp` into multiple tsconfig projects, e.g. `core` and `user-core` tsconfig projects, we will need to pick our new SDK public API exporting strategy. 

For now I:
1. Re-export `core` exports in appropriate `user-core` files.
   Example is `HttpError` above.
2. Change our `package.json` to export `./dist/user-core/<FILE>` instead of `./dist/<FILE>`.

To make relative import/export paths work even in built files (`'../../core/server/HttpError.js'` above), we need to recreate the same structure they had in source files.

That's why we have:
```
sdk/wasp/
├── tsconfig.json
├── package.json
├── ...
├── core/
│   ├── tsconfig.json
│   └── ...
├── user-core/
│   ├── tsconfig.json
│   └── ...
└── dist/
    ├── core/
    │   ├── .tsbuildinfo
    │   └── ...
    └── user-core/
        ├── .tsbuildinfo
        └── ...
```

waspc/data/Generator/templates/sdk/wasp/user-core/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  // TODO: Eventually split 'user-core' and 'src'
+  // into two different tsconfig projects.
+  "extends": "../tsconfig.sdk.json",
+  "compilerOptions": {
+    "rootDir": "..",
+    "outDir": "../dist/"
+  },
+  "include": [".", "../src"],
```
FranjoMindek, line 9: This is so that we can include both `sdk/wasp/src` and `sdk/wasp/user-core` as part of the `user-core` tsconfig project.

`src` is currently NOT a standalone tsconfig project.

waspc/data/Generator/templates/sdk/wasp/tsconfig.sdk.json
```diff
@@ -0,0 +1,89 @@
+{
```
FranjoMindek, line 1: Old SDK tsconfig without any "path" fields (`rootDir`, `outDir`...).

waspc/src/Wasp/Generator/SdkGenerator/UserCore/Auth/AuthFormsG.hs
```diff
@@ -17,60 +17,57 @@ import qualified Wasp.Generator.AuthProviders as AuthProviders
 import qualified Wasp.Generator.AuthProviders.OAuth as OAuth
 import Wasp.Generator.FileDraft (FileDraft)
 import Wasp.Generator.Monad (Generator)
-import Wasp.Generator.SdkGenerator.Common as C
+import Wasp.Generator.SdkGenerator.UserCore.Common
+  ( UserCoreTemplatesDir,
+    mkTmplFd,
+    mkTmplFdWithData,
+  )
 -- todo(filip) -- Should I put this under something like Wasp.Generator.Auth (doesn't exist) or Wasp.Generator.Common?
```
FranjoMindek: I'm also debating whether to put this path at start (as it might be an useful information) or at end (if we do not really need that detail).

Old code had both examples.
For now I always place them at bottom.

waspc/src/Wasp/Generator/SdkGenerator/AuthG.hs
```diff
@@ -52,34 +52,35 @@ genAuth spec =
         <++> genEmailAuth auth
         -- server stuff
         <++> sequence
-          [ genFileCopy [relfile|core/auth.ts|],
-            genFileCopy [relfile|auth/validation.ts|],
-            genFileCopy [relfile|auth/password.ts|],
-            genFileCopy [relfile|auth/jwt.ts|],
+          [ return $ mkTmplFd [relfile|core/auth.ts|],
```
FranjoMindek: This is the only one without `auth/` prefix!

waspc/src/Wasp/Generator/SdkGenerator/EmailSender/Providers.hs
```diff
@@ -3,32 +3,26 @@ module Wasp.Generator.SdkGenerator.EmailSender.Providers
...
-
-providersDirInSdkTemplatesDir :: Path' (Rel C.SdkTemplatesDir) (Dir ProvidersDir)
-providersDirInSdkTemplatesDir = [reldir|server/email/core/providers|]
-
 smtp :: EmailSenderProvider
 smtp =
   EmailSenderProvider
     { npmDependency = Just nodeMailerDependency,
-      setupFnFile = [relfile|smtp.ts|]
+      setupFnFile = [relfile|server/email/core/providers/smtp.ts|]
```
FranjoMindek: Instead of exporting both `providersDirInSdkTemplatesDir` and `setupFnFile` then combining them together (`</>`) somewhere else, I just inline it immediately.

waspc/src/Wasp/Generator/SdkGenerator/Server/JobGenerator.hs
```diff
@@ -119,19 +118,13 @@ getImportJsonForJobDefinition jobName =
...
-    genAllJobExecutors = concatMap genJobExecutor jobExecutors
-
-    -- Per each defined job executor, we generate the needed files.
-    genJobExecutor :: JobExecutor -> [FileDraft]
-    genJobExecutor PgBoss =
-      [ C.mkTmplFd [relfile|server/jobs/core/pgBoss/pgBoss.ts|],
-        C.mkTmplFd [relfile|server/jobs/core/pgBoss/pgBossJob.ts|],
-        C.mkTmplFd [relfile|server/jobs/core/pgBoss/types.ts|],
-        C.mkTmplFd [relfile|server/jobs/core/pgBoss/index.ts|]
+  _anyJob ->
```
FranjoMindek, line 124: I've simplified this to copy `PgBoss` files directly.

We had a generic gen job executor function, but it only created `PgBoss` stuff.
Let's just be honest and create them directly.

waspc/data/Generator/templates/sdk/wasp/package.json
```diff
@@ -5,148 +5,144 @@
...
+    it redirects to, we need to also let TypeScript know about it, and that can be done with
+    `typesVersions` field below.
+  =}
+  "typesVersions": {
+    "*": {
+      "client/api": ["api/index.ts"]
+    }
+  },
+  "license": "ISC",
+  {=& depsChunk =},
```
FranjoMindek, line 146: I could not find `include` field in `package.json` docs.

I removed it.

waspc/data/Generator/templates/sdk/wasp/user-core/auth/forms/internal/email/ForgotPasswordForm.tsx
```diff
@@ -1,6 +1,6 @@
 import { useContext } from 'react'
 import { useForm } from 'react-hook-form'
-import { requestPasswordReset } from '../../../email/actions/passwordReset.js'
+import { requestPasswordReset } from '../../../email/actions/passwordReset'
```
FranjoMindek: When I moved the whole folder, VSC `organized all imports` automatically for the moved files.
I tried to minimize the large changes (sorting) and left ones which I deem more beneficial:
- removal of `.jsx` or `.js` suffixes
- unused imports 

The rest were reverted to best of my abilities.

We can discuss if we want to revert all of them.

Should be okay because of: https://github.com/wasp-lang/wasp/issues/2492 https://github.com/wasp-lang/wasp/pull/2493

waspc/src/Wasp/Generator/SdkGenerator/Common.hs
```diff
@@ -1,74 +1,46 @@
...
-  where
-    relDstPath = castRel relSrcPath
-
-mkTmplFd :: Path' (Rel SdkTemplatesDir) File' -> FileDraft
-mkTmplFd path = mkTmplFdWithDst path (SP.castRel path)
-
-sdkRootDirInGeneratedCodeDir :: Path' (Rel ProjectRootDir) (Dir SdkRootDir)
-sdkRootDirInGeneratedCodeDir = [reldir|sdk/wasp|]
+sdkRootDirInProjectRootDir :: Path' (Rel ProjectRootDir) (Dir SdkRootDir)
+sdkRootDirInProjectRootDir = [reldir|sdk/wasp|]
```
FranjoMindek: I see this is renamed now.

Previously this file had both `sdkRootDirInProjectRootDir` and `sdkRootDirInGeneratedCodeDir`.

With 
```hs
sdkRootDirInGeneratedCodeDir = [reldir|sdk/wasp|]
sdkRootDirInProjectRootDir = [reldir|../.out|] </> sdkRootDirInGeneratedCodeDir
```

Once we merged the `.out` and `.build`, `sdkRootDirInProjectRootDir` was no longer neceesary.

No the question is which name is more correct.
In my code before merging the unification above, I used the `sdkRootDirInProjectRootDir` function always, since it was "more correct one". Always pointing to the right `sdk/wasp`. But I see we removed it from the code.

Which name do we prefer? 
I'd rather say in line with `sdkRootDirInProjectRootDir` while we still have `ProjectRootDir`.

waspc/data/Generator/templates/sdk/wasp/package.json
sodic: I still like the idea of a separate project for a public API. It will be very easy to keep track of it.

waspc/data/Generator/templates/sdk/wasp/package.json
```diff
@@ -5,148 +5,144 @@
...
-      "test": "echo \"Error: no test specified\" && exit 1",
-      "types": "tsc --declaration --emitDeclarationOnly --stripInternal --declarationDir dist"
-    },
+    "build": "tsc -b && node ./copy-assets.js",
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "types": "tsc --declaration --emitDeclarationOnly --stripInternal --declarationDir dist"
+  },
   {=! IMPORTANT: If you modify this list, remember to update the spreadsheet at =}
   {=! https://docs.google.com/spreadsheets/d/1QW-_16KRGTOaKXx9NYUtjk6m2TQ0nUMOA74hBthTH3g/edit =}
   "exports": {
```
sodic, line 14: Where is the export for the moved HttpError?

waspc/src/Wasp/Generator/SdkGenerator/Common.hs
```diff
@@ -1,74 +1,46 @@
...
 relDirToRelFileP :: Path Posix (Rel d) Dir' -> Path Posix (Rel d) File'
-relDirToRelFileP path = fromJust $ SP.parseRelFileP $ removeTrailingSlash $ SP.fromRelDirP path
+relDirToRelFileP path = fromJust $ parseRelFileP $ removeTrailingSlash $ fromRelDirP path
   where
     removeTrailingSlash = reverse . dropWhile (== '/') . reverse
 
 makeSdkImportPath :: Path Posix (Rel SdkRootDir) File' -> Path Posix (Rel s) File'
 makeSdkImportPath path = [reldirP|wasp|] </> path
 
-clientTemplatesDirInSdkTemplatesDir :: Path' (Rel SdkTemplatesDir) (Dir ClientTemplatesDir)
```
FranjoMindek, line 72: Wasn't really used that much.

waspc/src/Wasp/Generator/SdkGenerator/Client/OperationsGenerator.hs
```diff
@@ -14,27 +14,23 @@ import Wasp.Generator.Common (makeJsArrayFromHaskellList)
...
+  )
 import qualified Wasp.Generator.ServerGenerator as ServerGenerator
 import qualified Wasp.Generator.ServerGenerator.OperationsRoutesG as ServerOperationsRoutesG
 import Wasp.JsImport (JsImportName (JsImportField), JsImportPath (ModuleImportPath), getJsImportStmtAndIdentifier, makeJsImport)
 import Wasp.Util ((<++>))
 
 data ClientOpsTemplatesDir
 
-clientOpsDirInSdkTemplatesDir :: Path' (Rel SdkTemplatesDir) (Dir ClientOpsTemplatesDir)
-clientOpsDirInSdkTemplatesDir = clientTemplatesDirInSdkTemplatesDir </> [reldir|operations|]
```
FranjoMindek: Is moved down.

waspc/src/Wasp/Generator/SdkGenerator/Server/JobGenerator.hs
```diff
@@ -125,13 +127,6 @@ genJobExecutors spec = case getJobs spec of
         genServerJobFileCopy [relfile|core/pgBoss/index.ts|]
       ]
 
-serverJobsDirInSdkTemplatesDir :: Path' (Rel SdkTemplatesDir) Dir'
```
FranjoMindek: Is moved down.

waspc/src/Wasp/Generator/SdkGenerator/Server/OperationsGenerator.hs
```diff
@@ -21,16 +21,17 @@ import Wasp.AppSpec.Valid (isAuthEnabled)
...
 import Wasp.Generator.SdkGenerator.JsImport (extOperationImportToImportJson)
+import Wasp.Generator.SdkGenerator.UserCore.Common
+  ( UserCoreTemplatesDir,
+    mkTmplFdWithData,
+  )
 import Wasp.Util (toUpperFirst)
 
-data ServerOpsTemplatesDir
-
-serverOpsDirInSdkTemplatesDir :: Path' (Rel SdkTemplatesDir) (Dir ServerOpsTemplatesDir)
```
FranjoMindek: Is moved down.

waspc/src/Wasp/Generator/SdkGenerator/Root/Common.hs
```diff
@@ -0,0 +1,50 @@
...
+mkTmplFdWithDestAndData destFile tmplFile =
+  createTemplateFileDraft
+    (sdkRootDirInGeneratedCodeDir </> rootTemplatesDirInSdkRootDir </> destFile)
+    (sdkTemplatesDirInTemplatesDir </> rootTemplatesDirInSdkTemplatesDir </> tmplFile)
+
+rootTemplatesDirInSdkTemplatesDir :: Path' (Rel SdkTemplatesDir) (Dir RootTemplatesDir)
+rootTemplatesDirInSdkTemplatesDir = [reldir|.|]
+
+rootTemplatesDirInSdkRootDir :: Path' (Rel SdkRootDir) (Dir RootTemplatesDir)
+rootTemplatesDirInSdkRootDir = castRel rootTemplatesDirInSdkTemplatesDir
```
sodic: There are several mistakes in this file and this is the root cause - these two directories are not the same directory but they're marked as such with `RootTemplatesDir`. Fixing this will expose all the other mistakes.

waspc/src/Wasp/Generator/SdkGenerator/Root/Common.hs
```diff
@@ -0,0 +1,50 @@
...
+
+-- | Directory of the SDK root tsconfig project.
+-- References all other SDK tsconfig projects.
+data RootTemplatesDir
+
+mkTmplFd ::
+  Path' (Rel RootTemplatesDir) File' ->
+  FileDraft
+mkTmplFd tmplFile =
+  mkTmplFdWithDestAndData tmplFile tmplFile Nothing
```
sodic: For parameter names, let's try to follow the conventins in other Common.hs files (or change them for the better in all places in the refactoring PR).

waspc/src/Wasp/Generator/SdkGenerator/Core/Common.hs
```diff
@@ -0,0 +1,50 @@
...
+mkTmplFdWithDestAndData destFile tmplFile =
+  createTemplateFileDraft
+    (sdkRootDirInGeneratedCodeDir </> coreTemplatesDirInSdkRootDir </> destFile)
+    (sdkTemplatesDirInTemplatesDir </> coreTemplatesDirInSdkTemplatesDir </> tmplFile)
+
+coreTemplatesDirInSdkTemplatesDir :: Path' (Rel SdkTemplatesDir) (Dir CoreTemplatesDir)
+coreTemplatesDirInSdkTemplatesDir = [reldir|core|]
+
+coreTemplatesDirInSdkRootDir :: Path' (Rel SdkRootDir) (Dir CoreTemplatesDir)
+coreTemplatesDirInSdkRootDir = castRel coreTemplatesDirInSdkTemplatesDir
```
sodic: Similar mistake as in other files. These two are not the same directory.

waspc/src/Wasp/Generator/SdkGenerator/Server/JobGenerator.hs
```diff
@@ -59,10 +61,10 @@ genIndexTs jobs =
 
 genJob :: (String, Job) -> Generator FileDraft
 genJob (jobName, job) =
-  return $ mkTmplFdWithDstAndData tmplFile destFile (Just tmplData)
+  return $ mkTmplFdWithDestAndData destFile tmplFile (Just tmplData)
```
sodic: It seems like this function also changes the convention (`Dst` to `Dest`, flips argument order).

Let's try to keep things consistent with the existing system. I'm no fan of these names myself, but consistency is more important.

waspc/src/Wasp/Generator/SdkGenerator/UserCore/Common.hs
```diff
@@ -0,0 +1,50 @@
...
+  FileDraft
+mkTmplFdWithData tmplFile tmplData =
+  mkTmplFdWithDestAndData tmplFile tmplFile (Just tmplData)
+
+mkTmplFdWithDestAndData ::
+  Path' (Rel UserCoreTemplatesDir) File' ->
+  Path' (Rel UserCoreTemplatesDir) File' ->
+  Maybe Aeson.Value ->
+  FileDraft
+mkTmplFdWithDestAndData destFile tmplFile =
```
sodic: It seems this function flips the source and the destination. I've commented on it somewhere else - let's try to keep things as consistent as possible.

This is inconsistent both with:
- The other functions in this file which all take the template as the first argument.
- Other similar functions across the codebase.

I think this applies to other `Common` modules too.

waspc/src/Wasp/Generator/SdkGenerator/Common.hs
sodic: When removing stuff like this, I suggest pointing to the places where it was used. It saves the reviewer from searching themselves.

That said, we should bring these guys back.
They weren't used that much, that's true. But they should have been. The solution is therefore the exact opposite of removing them - We should be using them as much as possible!

Ideally, if we one day decide _"Hey, let's use `web-app` instead of `client`,"_ that change should be as simple as possible. And removing these constants is a step in the opposite direction, we're going all in into duplication instead of DRY code.

waspc/src/Wasp/Generator/SdkGenerator/Common.hs
```diff
@@ -1,74 +1,46 @@
...
-mkTmplFd :: Path' (Rel SdkTemplatesDir) File' -> FileDraft
-mkTmplFd path = mkTmplFdWithDst path (SP.castRel path)
-
 sdkRootDirInGeneratedCodeDir :: Path' (Rel ProjectRootDir) (Dir SdkRootDir)
 sdkRootDirInGeneratedCodeDir = [reldir|sdk/wasp|]
 
 sdkTemplatesDirInTemplatesDir :: Path' (Rel TemplatesDir) (Dir SdkTemplatesDir)
 sdkTemplatesDirInTemplatesDir = [reldir|sdk/wasp|]
 
+-- | External @src@ directory refers to the user's @src@ directory.
```
sodic: Good addition, but incomplete.

We should clearly explain we're talking about user code inside the SDK.
An even better option is renaming this in the refactoring PR into something like `generatedUserSrcInSdkRootDir`.

waspc/src/Wasp/Generator/SdkGenerator.hs
```diff
@@ -52,17 +52,23 @@ import Wasp.Generator.SdkGenerator.Client.AuthG (genNewClientAuth)
 import Wasp.Generator.SdkGenerator.Client.CrudG (genNewClientCrudApi)
 import qualified Wasp.Generator.SdkGenerator.Client.OperationsGenerator as ClientOpsGen
 import Wasp.Generator.SdkGenerator.Client.RouterGenerator (genNewClientRouterApi)
-import qualified Wasp.Generator.SdkGenerator.Common as C
+import Wasp.Generator.SdkGenerator.Common
+  ( extSrcDirInSdkRootDir,
+    sdkRootDirInGeneratedCodeDir,
+  )
+import qualified Wasp.Generator.SdkGenerator.Core.Common as Core
```
sodic: I'd probalby do `CoreC` to signal common. Applies to others too.

waspc/src/Wasp/Generator/SdkGenerator.hs
```diff
@@ -90,31 +92,31 @@ buildSdk projectRootDir = do
...
+      return $ UserCore.mkTmplFd [relfile|vite-env.d.ts|],
+      return $ UserCore.mkTmplFd [relfile|prisma-runtime-library.d.ts|],
+      return $ UserCore.mkTmplFd [relfile|api/index.ts|],
+      return $ UserCore.mkTmplFd [relfile|api/events.ts|],
+      return $ UserCore.mkTmplFd [relfile|core/storage.ts|],
+      return $ UserCore.mkTmplFd [relfile|server/index.ts|],
+      return $ UserCore.mkTmplFd [relfile|client/test/vitest/helpers.tsx|],
+      return $ UserCore.mkTmplFd [relfile|client/test/index.ts|],
+      return $ UserCore.mkTmplFd [relfile|client/hooks.ts|],
+      return $ UserCore.mkTmplFd [relfile|client/index.ts|],
```
sodic: Hm, a lot of dupplication here. Perhaps we could extract `genFileCopy`. We'd need two kinds, but it could still be nicer I think.


You could create these three functions in global scope, it might give us an easier time when renaming later.

sodic (commented):
Excellent!

Thanks again for splitting, it was a joy to review!

Left some comments, so let's comment on them :)

waspc/data/Generator/templates/sdk/wasp/package.json
FranjoMindek: I do agree with that one, but I will skip it for now.
I will do it as a separate PR once we sort `core` and `user-core`.

waspc/data/Generator/templates/sdk/wasp/package.json
FranjoMindek: For now, I export it through `/sdk/wasp/user-core/server/index.ts`, as it always was.
In future we will have a separate project for the API.

```ts
export { HttpError } from '../../core/server/HttpError.js'
```

waspc/src/Wasp/Generator/SdkGenerator/Common.hs
FranjoMindek: I've reverted this so it exists again.
For now we don't use those.

We want to reduce this duplication the most we can, but for that we need to solve the circular dependency problem.
So for now I am okay with thing not being used, and one we solve that problem, we can generalize everything.

waspc/src/Wasp/Generator/SdkGenerator/Root/Common.hs
FranjoMindek: I think I fixed all of them.
Should've paid more attention to what claude code did do.
The runtime worked (compiles and builds wasp projects properly).

waspc/src/Wasp/Generator/SdkGenerator/Root/Common.hs
FranjoMindek: Changed back.

waspc/src/Wasp/Generator/SdkGenerator/Server/JobGenerator.hs
FranjoMindek: Reverted `Dest` -> `Dst`

waspc/src/Wasp/Generator/SdkGenerator/UserCore/Common.hs
FranjoMindek: I've kept this one, as I think this is a mistake we have.
I will also refactor server/webapp generators later to swap the dst/src order.

waspc/src/Wasp/Generator/SdkGenerator/Common.hs
FranjoMindek: Now I am against that refactor, because I would want to rename "extSrc" in all of files, not just SDK gen. So for now I would keep it.

> We should clearly explain we're talking about user code inside the SDK.

I thought the function type and name explain that clearly?

waspc/src/Wasp/Generator/SdkGenerator.hs
FranjoMindek: Most of these will be dispersed later, so I would do nothing for now.

waspc/src/Wasp/Generator/SdkGenerator/Root/Common.hs
sodic: There's still some naming work here.

waspc/src/Wasp/Generator/SdkGenerator/Common.hs
sodic: We agreed to move this comment to the type definition. There's already a comment there, but it's not clear enough.

waspc/src/Wasp/Generator/SdkGenerator/Root/Common.hs
FranjoMindek: Done according to agreement

waspc/src/Wasp/Generator/SdkGenerator/UserCore/Common.hs
FranjoMindek: We reverted this one

waspc/src/Wasp/Generator/SdkGenerator.hs
sodic: `RootC` is still caleed `Root`.

waspc/src/Wasp/Generator/ExternalCodeGenerator/Common.hs
```diff
@@ -4,10 +4,11 @@ module Wasp.Generator.ExternalCodeGenerator.Common
...
--- TODO: consider refactoring the usage of GeneratedExternalCodeDir since
+-- TODO: consider refactoring the usage of 'GeneratedExternalCodeDir' since
 -- generated code might end up in multiple places (e.g. src/ but also public/).
 -- Name should probably be narrowed down to something that represents only the
--- src/ directory. Maybe GeneratedExtSrcDir?
+-- src/ directory. Maybe 'GeneratedExtSrcDir'?
 
--- | Path to the directory where we generate the external user code.
+-- | Directory where we generate the external code.
+-- External code refer to the code files in user's @src/@ directory.
```
sodic: There's a gramatical error here (refer). Also, I'd still interpret the comment as "This type refers to the code files...".

Can we do:
> The term 'External code' refers to the code users write in their `src` directory
> This type represents the folder containing user-written code inside our generated SDK code.

waspc/src/Wasp/Generator/SdkGenerator/Auth/AuthFormsG.hs
```diff
@@ -17,8 +17,8 @@ import qualified Wasp.Generator.AuthProviders as AuthProviders
 import qualified Wasp.Generator.AuthProviders.OAuth as OAuth
 import Wasp.Generator.FileDraft (FileDraft)
 import Wasp.Generator.Monad (Generator)
-import Wasp.Generator.SdkGenerator.Common
-  ( SdkTemplatesDir,
+import Wasp.Generator.SdkGenerator.UserCore.Common
+  ( SdkTemplatesUserCoreDir,
     mkTmplFd,
     mkTmplFdWithData,
```
sodic: For this two functions, a qualified import makes more sense because they share their name with other functions that could possibly end up in the same file. Qualifying the import will also help Claude (and you) when refactoring later.

Here's how I decide it when to use qualified imports: https://github.com/wasp-lang/wasp/pull/3501#discussion_r2669077795

waspc/src/Wasp/Generator/SdkGenerator/Auth/EmailAuthG.hs
```diff
@@ -18,12 +18,12 @@ import Wasp.Generator.AuthProviders.Email
 import qualified Wasp.Generator.DbGenerator.Auth as DbAuth
 import Wasp.Generator.FileDraft (FileDraft)
 import Wasp.Generator.Monad (Generator)
-import Wasp.Generator.SdkGenerator.Common
-  ( SdkTemplatesDir,
+import Wasp.Generator.SdkGenerator.JsImport (extImportToImportJson)
+import Wasp.Generator.SdkGenerator.UserCore.Common
+  ( SdkTemplatesUserCoreDir,
     mkTmplFd,
     mkTmplFdWithData,
```
sodic: Same story with qualified imports. I'll stop commenting on it to avoid noise.

waspc/src/Wasp/Generator/SdkGenerator/Client/AuthG.hs
```diff
@@ -91,9 +91,9 @@ genAuthGitHub auth =
     then sequence [genFileCopyInClientAuth [relfile|github.ts|]]
     else return []
 
-clientAuthDirInSdkTemplatesDir :: Path' (Rel SdkTemplatesDir) Dir'
-clientAuthDirInSdkTemplatesDir = [reldir|client/auth|]
+clientAuthDirInSdkTemplatesUserCoreDir :: Path' (Rel SdkTemplatesUserCoreDir) Dir'
```
sodic, line 100: This thing has a similar naming problem as mentioned in the other PR - one meaning injects itself in the middle of another:
- "SDK" and "User core" belong to one classfiication.
- "Templates" and "Generated" belong to another one.

So, `SdkUserCoreTemplatesDir` is more appropriate than `SdkTemplatesUserCoreDir`. As usual, common prefixes are a good heuristic.

waspc/src/Wasp/Generator/SdkGenerator/Client/CrudG.hs
```diff
@@ -51,16 +51,16 @@ genCrudOperations spec cruds = return $ map genCrudOperation cruds
...
+        (clientCrudDirInSdkTemplatesUserCoreDir </> [relfile|_crud.ts|])
+        (castRel (clientCrudDirInSdkTemplatesUserCoreDir </> fromJust (parseRelFile (name ++ ".ts"))))
         (Just tmplData)
       where
         tmplData = getCrudOperationJson name crud idField
         idField = getIdFieldFromCrudEntity spec crud
 
-clientCrudDirInSdkTemplatesDir :: Path' (Rel SdkTemplatesDir) Dir'
-clientCrudDirInSdkTemplatesDir = [reldir|client/crud|]
+clientCrudDirInSdkTemplatesUserCoreDir :: Path' (Rel SdkTemplatesUserCoreDir) Dir'
```
sodic: Same naming problem as mentioned above. I won't repeat it anymore to avoid noise, but it applies to many variables and types.

waspc/src/Wasp/Generator/SdkGenerator/UserCore/Common.hs
```diff
@@ -0,0 +1,61 @@
...
+  FileDraft
+mkTmplFdWithData srcFilePath tmplData =
+  mkTmplFdWithDstAndData
+    srcFilePath
+    (castRel srcFilePath)
+    (Just tmplData)
+
+mkTmplFdWithDstAndData ::
+  Path' (Rel SdkTemplatesUserCoreDir) File' ->
+  Path' (Rel SdkUserCoreDir) File' ->
```
sodic, line 54: Maybe `SdkUserCoreGeneratedDir`? Not sure, I know this is a mess so feel free to skip and RAW without explanation.
You're currently more familiar with the conventions and the implications of each change, so just making suggestions.

sodic (commented):
Ok, great work. I suggested some improvements, and I think I maybe even noticed a mistake or two. Check the comments .

waspc/src/Wasp/Generator/SdkGenerator/Auth/AuthFormsG.hs
FranjoMindek: > other functions that could possibly end up in the same file

IIRC the idea is that we make the "core", "user-core", and "root" project split also be reflected in the file structure.
So that should never really happen?

E.g., all "core" generators will only ever be in `SdkGenerator/Core.hs` file and `SdkGenerator/Core/` dir, and no other generator should be there.

Even during refactoring, this will never happen except in `SdkGenerator.hs` itself.
And we use qualified imports there.

Do you still think we should do it?

waspc/src/Wasp/Generator/SdkGenerator/Client/AuthG.hs
FranjoMindek: I wanted to follow the naming schema we had on our "root" dirs.
Specifically we have `SdkRootDir` and `SdkTemplatesDir`.
So I took them "as prefix" for later names.

If we are talking about something from sdk root dir its  `SdkRoot...Dir` and if its from sdk templates root dir its `SdkTemplates...Dir`.


This is so when you search for `SdkTemplates.*Dir` you get all dirs located in the sdk templates.
Or you search for `SdkTemplatesCore.*Dir` to get core dirs specifically.
It made sense to me.

So I actually still prefer my solution.

But this is no longer a problem since we removed these helpers.

waspc/src/Wasp/Generator/SdkGenerator/Auth/AuthFormsG.hs
sodic: If it's never gonna happen then I guess it's not necessary.

waspc/src/Wasp/Generator/SdkGenerator/Client/AuthG.hs
sodic: I'm not sure I understand.

> Specifically we have SdkRootDir and SdkTemplatesDir.
> So I took them "as prefix" for later names.

So does that mean that something would be called `SdkRootUserCoreDir`? If so I think that's a problem too, for two reasons:
- The one I described above.
- The name root being used as a prefix of a folder. The convention is that variables for folder names don't include details of those folder's paths. That's handled by Strong Path, so:
  - If "root" is there because we want to say "This folder is in Sdk's root dir," that's not something we should be doing in the variable's name.
  - If "root' is there because we're using the word "root" to mean "generated," than that's a problem because of naming.

Assuming "Domain" means "template or generated," and module means "server/user core/etc" The names I think make sense are:
- `DomainStuffModuleStuff` (e.g., `TemplatesSdkUserCoreDir`, `GeneratedSdkUserCoreDir`)
- `ModuleStuffDomainStuff` (e.g., `SdkUserCoreTemplatesDir`, `SdkUserCoreGeneratedDir`)

And we should choose the one more consistent with the rest of the project.

FranjoMindek: > So does that mean that something would be called SdkRootUserCoreDir? 

No, it's enough to know one is `Sdk...` and another is `SdkTemplates...`.
I've used names:
```hs
-- | Directory of the SDK core tsconfig project in generated code.
-- It contains all logic not dependent on the user's project.
data SdkCoreDir

-- | Directory of the SDK core tsconfig project in templates.
-- It contains all logic not dependent on the user's project.
data SdkTemplatesCoreDir
```

I've done it this way because we never used `Generated` name in our paths in the SDK generator. Files/dirs with `Templates` were in templates, and those without it were generated.

What makes sense to me is to go from more broad to more granular concepts.
The question is whether `Templates`, or `Sdk` should go first is valid:
- `Templates` is a broader subject
- `Sdk` being first fits more nicely into conceptual models we have. Mostly because non-templates stuff is almost always in the background (hidden inside of `mk` functions), so the primary split we need is between the projects (web-app, sdk, server).

`TemplatesSdkUserCoreDir` is nice because it follows the granularity rule, but the problem is that we've never done this pattern. That is the reason I did not do it this way. If we are fine with that we can go this way.

For `SdkUserCoreTemplatesDir` I don't like the fact that `Templates/Generated` split is so late in the name.

So I would prefer the domain before module approach from your viewpoint.

PR #3620, opened 2026-01-15
Add npm support in `wasp uninstall`

cprecioso:
## Description

- Closes https://github.com/wasp-lang/wasp/issues/3621

Updates the `wasp uninstall` command to work with the new npm installation method:

- `wasp uninstall` only tries to delete files and folders that exist (because the wasp executable won't exist in the expected location on npm)
- Allows the Wasp CLI to know if it was installed through an npm package or the legacy installer
- If `wasp uninstall` is called in an npm installation, it only says it has "removed data" and guides to uninstall through npm.

image

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:


waspc/cli/src/Wasp/Cli/Common.hs
```diff
@@ -18,3 +21,19 @@ waspWarns what = putStrLn $ Term.applyStyles [Term.Magenta] what
 
 waspScreams :: String -> IO ()
 waspScreams what = putStrLn $ Term.applyStyles [Term.Red] what
+
+data CliPackagingMode
+  = Installer
+  | NpmPackage
```
infomiho: I feel like `Installer` and `NpmPackage` are not in the same value space i.e. `Installer` does not feel like a packaging mode. 

I'd either rename this into `CliInstallMethod` and make it `Curl` and `Npm` or adjust the `CliPackagingMode` to be `NpmPackage` and maybe `Binary` or `GithubRelease`.

waspc/cli/src/Wasp/Cli/Common.hs
```diff
@@ -18,3 +21,19 @@ waspWarns what = putStrLn $ Term.applyStyles [Term.Magenta] what
...
+
+-- Keep the environment variable name and value in sync with
+-- scripts/make-npm-packages/templates/main-package/bin.js
+
+packagingModeEnvVar :: String
+packagingModeEnvVar = "WASP_CLI_PACKAGE_MODE"
```
infomiho:
```suggestion
packagingModeEnvVar = "WASP_CLI_PACKAGING_MODE"
```

I'd stay consistent with other names (if you decide to keep the packaging mode naming).

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
```diff
@@ -3,58 +3,99 @@ module Wasp.Cli.Command.Uninstall
...
+      SP.castDir $ getWaspCacheDir userCacheDir
+    ]
+
+getWaspFiles :: IO [Path' Abs File']
+getWaspFiles = do
+  homeDir <- getHomeDir
+
+  return
+    [ homeDir </> waspExecutableInHomeDir
+    ]
```
infomiho:
```suggestion
  return [homeDir </> waspExecutableInHomeDir]
```

Nipick

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
```diff
@@ -3,58 +3,99 @@ module Wasp.Cli.Command.Uninstall
...
-        <> " deleted all the docker volumes it might have created."
-        <> (" You can easily list them by doing `docker volume ls | grep " <> waspDevDbDockerVolumePrefix <> "`.")
+  cliSendMessageC dockerVolumeMsg
+
+npmUninstall :: Command ()
+npmUninstall = do
+  cliSendMessageC $ Msg.Start "Removing Wasp data..."
+  liftIO removeWaspFiles
+  cliSendMessageC $ Msg.Success "Removed Wasp data."
+  cliSendMessageC $ Msg.Info "To uninstall the Wasp CLI, please run 'npm uninstall -g @wasp/cli'."
```
infomiho: 1. 
	```suggestion
	  cliSendMessageC $ Msg.Info "To uninstall the Wasp CLI, please run 'npm uninstall -g @wasp.sh/wasp-cli'."
	```

2. What would happen if people uninstalled Wasp via `npm` instead of `wasp uninstall`? Will there be some leftover files?

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
cprecioso: oh i actually left it like this on purpose for consistency with the `getWaspDirectories` fn, but fixed it anyway

waspc/cli/src/Wasp/Cli/Command/Uninstall.hs
cprecioso: > `@wasp/cli`

damn that's a dumb mistake 😂

> What would happen if people uninstalled Wasp via npm instead of wasp uninstall? Will there be some leftover files?

Yeah, `npm` doesn't know that we write files to `.cache` (and there's no way to tell it)

waspc/cli/src/Wasp/Cli/Common.hs
cprecioso: I decided for

```hs
data CliInstallMethod
  = BinaryInstaller
  | NpmPackage
```

cprecioso:
@infomiho ready

waspc/cli/src/Wasp/Cli/Common.hs
infomiho: That's better, another nitpick incoming, _both_ are binary installers. But okay, I can with that 👍

infomiho (approved):
One nitpick, but looking good

waspc/cli/src/Wasp/Cli/Common.hs
cprecioso: 😅

PR #3630, opened 2026-01-16
Add npm installation instructions to docs and CLI

cprecioso:
- Resolves https://github.com/wasp-lang/wasp/issues/3448

## Description

As a final part of pushing npm as a default method for the future, adapts all references to the installer:

- In docs and examples, we prefer the new npm method, except:
	- Migration guide: We call it out explicitly and let the user choose
	- Installation guide: Give it as an option

- In the CLI, we detect the current installation method (as setup in https://github.com/wasp-lang/wasp/pull/3620), and show the instructions to keep using that one.

- Not migrated:
	- Versioned docs and old migration guides
	- Old blog posts

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/tests/AppSpec/ValidTest.hs
+++ b/waspc/tests/AppSpec/ValidTest.hs
@@ -99,9 +99,10 @@ spec_AppSpecValid = do
                              [ "Your Wasp version does not match the app's requirements.",
                                "You are running Wasp " ++ show WV.waspVersion ++ ".",
                                "This app requires Wasp ^" ++ show incompatibleWaspVersion ++ ".",
-                               "To get instuctions on how to install the correct Wasp version, run:",
-                               "  wasp version",
-                               "and follow the instructions there."
+                               "To install a specific version of Wasp, do:",
+                               "  curl -sSL https://get.wasp.sh/installer.sh | sh -s -- -v x.y.z",
+                               "where x.y.z is your desired version.",
+                               "Check https://github.com/wasp-lang/wasp/releases for the list of valid versions."
                              ]
                        ]
 


waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -116,10 +116,9 @@ validateWaspVersion specWaspVersionStr = eitherUnitToErrorList $ do
           [ "Your Wasp version does not match the app's requirements.",
             "You are running Wasp " ++ show actualVersion ++ ".",
             "This app requires Wasp " ++ show expectedVersionRange ++ ".",
-            "To install specific version of Wasp, do:",
-            "  curl -sSL https://get.wasp.sh/installer.sh | sh -s -- -v x.y.z",
-            "where x.y.z is your desired version.",
-            "Check https://github.com/wasp-lang/wasp/releases for the list of valid versions."
+            "To get instuctions on how to install the correct Wasp version, run:",
+            "  wasp version",
+            "and follow the instructions there."
```
cprecioso: Since we depend on `IO` to get the current packaging method, and we're very far from it here, we just point users to use `wasp version`, which will do the correct thing.

.github/workflows/release-examples-deploy.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
     steps:
       - name: Install Wasp
         run: |
-          curl -sSfL https://get.wasp.sh/installer.sh | sh -s -- -v "${{  inputs.version }}"
+          npm i -g @wasp.sh/wasp-cli@"${{  inputs.version }}"
```
infomiho, line 54:
```
$ npm show @wasp.sh/wasp-cli versions
[ '0.20.0', '0.20.1' ]
```

Given that some of our users might want to switch while still using old versions, do we want to maybe publish some older versions to `npm` as well? I think it wouldn't take a lot of work, but nice DX for our users to have them covered for _some_ of the older versions.

mage/Dockerfile
```diff
@@ -39,8 +39,7 @@ RUN mkdir -p /app/.wasp/build/server/node_modules
 # TODO: Use pm2?
 # TODO: Use non-root user (node).
 FROM base AS server-production
-RUN curl -sSL https://get.wasp.sh/installer.sh | sh -s -- -v 0.20.0
-ENV PATH "$PATH:/root/.local/bin"
```
infomiho, line 43: Nice simplification

waspc/cli/exe/Main.hs
```diff
@@ -209,15 +210,17 @@ printUsage =
 
 printVersion :: IO ()
 printVersion = do
+  installCommand <- getInstallationCommand <$> getPackagingMode
```
infomiho: Nitpick: `installCommand` is not really a value but a fn, so the name doesn't feel right. Maybe get the packaging mode here and then call the `getInstallationCommand` with all the params below or something. I don't have a great idea, but I just dislike the fact I thought this is a `String` but it turns out later is a function.

waspc/cli/src/Wasp/Cli/Common.hs
```diff
@@ -37,3 +38,13 @@ getPackagingMode =
...
+
+getInstallationCommand :: CliPackagingMode -> Maybe String -> String
+getInstallationCommand NpmPackage (Just version) =
+  "npm i -g @wasp.sh/wasp-cli@" ++ version
+getInstallationCommand NpmPackage Nothing =
+  "npm i -g @wasp.sh/wasp-cli@latest"
+getInstallationCommand Installer (Just version) =
+  "curl -sSL https://get.wasp.sh/installer.sh | sh -s -- -v " ++ version
+getInstallationCommand Installer Nothing =
+  "curl -sSL https://get.wasp.sh/installer.sh | sh -s"
```
infomiho: It would be nice if we could DRY out the `npm i -g @wasp.sh/wasp-cli` and `curl -sSL https://get.wasp.sh/installer.sh | sh -s` 

We should have a place where we store the npm package name `@wasp.sh/wasp-cli`. We use it in the uninstall command and here as well.

waspc/src/Wasp/AppSpec/Valid.hs
infomiho: What is the best DX for our users if we didn't have to worry about `IO`? Can we get there with some reasonable amount of effort around `IO`?

If we believe it's better DX to display the command without the `wasp version` indirection, I think we should give it a try.

web/docs/migration-guides/migrate-from-0-20-to-0-21.md
```diff
@@ -4,18 +4,55 @@ title: Migration from 0.20.X to 0.21.X
...
 
+</TabItem>
+<TabItem value="npm" label="npm">
+
+```sh
+npm i -g @wasp.sh/wasp-cli@0.21.0
+```
+
+</TabItem>
+</Tabs>
```
infomiho, line 41: This could be a nice component we extract and use in each migration guide.

web/docs/introduction/quick-start.md
```diff
@@ -14,7 +14,7 @@ Let's create and run our first Wasp app in 3 short steps:
 1. **To install Wasp on Linux / OSX / WSL (Windows), open your terminal and run:**
 
    ```shell
-   curl -sSL https://get.wasp.sh/installer.sh | sh
+   npm i -g @wasp.sh/wasp-cli@latest
    ```
 
    ℹ️ Wasp requires Node.js and will warn you if it is missing: check below for [more details](#requirements).
```
infomiho: Image

This kinda looks weird, if you don't have Node.js, you can even run `npm`.

web/docs/introduction/quick-start.md
```diff
@@ -106,14 +106,23 @@ If you need it, we recommend using [nvm](https://github.com/nvm-sh/nvm) for mana
...
+    {label: 'Linux / macOS (installer)', value: 'installer'},
     {label: 'Windows', value: 'win'},
     {label: 'From source', value: 'source'}
   ]}
 >
-  <TabItem value="linux/osx">
+  <TabItem value="npm">
+    Open your terminal and run:
+
+    ```shell
```
infomiho, line 116: Image

Does this note not apply for both `npm` and installer install methods?

cprecioso: Nope, we never taught the installer to install Apple Silicon builds, but I did add the feature in npm packages.

.github/workflows/release-examples-deploy.yaml
cprecioso: Mmm I can bring it up in planning since this could take me some time to package manually.

cprecioso: https://github.com/wasp-lang/wasp/issues/3636

waspc/src/Wasp/AppSpec/Valid.hs
cprecioso: I was thinking we might want to `unsafePerformIO`-it, since we're kind of treating it as a compiled-in variable.

web/docs/migration-guides/migrate-from-0-20-to-0-21.md
cprecioso: I thought about it, I think we can extract it when we actually duplicate it for the next migration guide.

waspc/cli/src/Wasp/Cli/Common.hs
cprecioso: i tried this but it feels less useful:

image

I will extract the package name only, but not DRY'ing the command

waspc/src/Wasp/AppSpec/Valid.hs
infomiho: I think we have a precedent around doing IO on CLI startup to resolve some value. Martin used it for resolving the `npm` path on Windows. 

cc: @Martinsos for input

cprecioso: Context for Martin, we're detecting at runtime whether we are being called from the npm package or the regular script-installed binary. The npm package passes `WASP_CLI_INSTALL_MODE=npm` as an env var to the CLI so that we can adapt the guidance to install/uninstall (like in `wasp version` or `wasp uninstall`). But given it's an env var, we need to use `IO`.

This specific thread is about a place inside of `AppSpec` where we're checking the `main.wasp#version` and crafting an error message with installer guidance, and we're very far away of `IO` here. My initial solution was to just tell people to run `wasp version` for correct guidance, but we'd like to avoid that and just tell people how to do it directly.

Martinsos: Yeah all makes sense!
What Miho said, since this is a global value that we don't expect to change during the execution of the binary (and even if it does change for real, we don't mind it remaining static on our side during the execution), we can use the unsafePerformIO to basically read the env var just once and then use that value from there on.
Check how I did it in the PR where I fix the exeuction of npm/npx/node, the point is basically to define a top level binding `waspCliInstallMode = ` that does unsafePerformIO to read that env var, and taht is it. OH and use the NO_INLINE pragma yeah, to ensure it does not get inlined.

Ok btw, you could actually just use unsafeperformIO in place, that is true, but this way it is nicer becuase you ensure it happens only once, the IO operation, and behaviour is consistent, ... . If not obvious, I can elaborate more on why I like the approach above better, but yeah even just calling unsafePerformIO directly when we need the valuie would work.

cprecioso: okay, incorporated this

cprecioso:
@infomiho ready

waspc/src/Wasp/Util/InstallMethod.hs
```diff
@@ -0,0 +1,42 @@
...
+  -- since we're treating it as a static value it's fine to treat this as a compiled-in
+  -- value.
+  case unsafePerformIO $ lookupEnv installMethodEnvVarName of
+    Just "npm" -> NpmPackage
+    _ -> BinaryInstaller
+
+npmPackageName :: String
+npmPackageName = "@wasp.sh/wasp-cli"
+
+getInstallationCommand :: CliInstallMethod -> Maybe String -> String
```
infomiho: Given that you are always doing `getInstallationCommand waspCliInstallMethod <something>`, I'd consider the following:

1. Create one general function that takes both the install method and the version
2. Create one specific function that takes only the version and uses the `waspCliInstallMethod` value.

Maybe: `getInstallationCommand <version>` and `getInstallationCommandFromInstallMethod <method> <version>` or something similar. Where `getInstallationCommand = getInstallationCommandFromInstallMethod waspCliInstallMethod`.

You'll use the first one only, but we still have the general one for future use.

infomiho (approved):
One idea how to simplify usage of the `getInstallationCommand` but LGTM

PR #3639, opened 2026-01-20
Npm package: Do not publish to `latest` tag

cprecioso:
### Context

**Our release process:** When we're doing a release, we create both RCs and final releases. Regardless of the kind of release we're doing, a GH Release is created as draft and then we manually mark it as Pre-Release or Release.

**npm tags:** in the npm registry, you can give aliases ("tags") to a specific package version e.g. `rc` tag &rarr; `0.21.0-rc1` version, or `nightly` &rarr; `0.0.0-dev+8a3c4d`. These tags can be updated to point to another version, making it a bit like "channels" for you to publish your package to.

   - `npm i wasp@0.21.0` &rarr; installs the specific version
   - `npm i wasp@rc` &rarr; installs the version that `rc` tag points to at that moment
   - `npm i wasp` &rarr; when no version or tag is defined, npm by default installs the `latest` tag
     - by default, `npm publish` also modifies the `latest` tag to point to your newly-published version, so the general case of just publishing regular versions is seamless

### Problem

The automation for publishing when we create a tag is not giving the `npm publish` command a tag, so even RCs are being released to the `latest` tag. Instead, we should make it behave a bit more like our current GH releases workflow, where the release is not widely made available, but requires a manual action to set as a release version.

### Solution

We make it so that the automation will publish everything to the `rc` channel. This one doesn't have any special meaning, just the fact that it won't present itself as `latest` (i.e. default). Then, when we have tested the release and want to make it widely available, we run an `npm` command to set it as `latest`, same as with the GH releases.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/.github/workflows/release-create.yaml
+++ b/.github/workflows/release-create.yaml
@@ -64,5 +64,7 @@ jobs:
         run: |
           set -x # Log each command for debugging
           for pkg in ./packages/*; do
+            # The tag name has no special meaning, just avoids us
+            # broadly publishing an RC to the "latest" tag.
             npm publish --tag rc "$pkg"
           done


cprecioso:
> Is it possible to differentiate RC releases from other releases that happen on PRs.

This is already the case. PR packages are published to a whole different registry, pkg.pr.new. That is taken care of in another automation. This is the automation that handles publishing when a new version tag is created.

> If we can differentiate RC releases

By default almost, every release we do here is either an RC, or an RC that will convert to a regular release.

FranjoMindek (approved):
Oh okay, then LGTM!

Let's just document this:

> Then, when we have tested the release and want to make it widely available, we run an npm command to set it as latest, same as with the GH releases.

cprecioso:
> Let's just document this:

Already in the PR :D

PR #3648, opened 2026-01-22
Migrate Wasp libs scripts to JS

infomiho:
## Description

Instead of needing to maintain both Linux/Unix and Windows scripts, we now have only one JS script.

I've also migrated the `get-wasp-version` command from `run` into a separate script for the same reason.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:


waspc/tools/get-waspc-version.ts
```diff
@@ -0,0 +1,17 @@
...
+// Gets the waspc version from the Cabal package file.
+
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const waspcDir = fileURLToPath(new URL("..", import.meta.url));
+const cabalFile = readFileSync(join(waspcDir, "waspc.cabal"), "utf-8");
+
+const match = cabalFile.match(/^version:\s*(.+)$/m);
```
infomiho, line 16: Before we used `cabal repl` to get the Wasp version. It was slow but it worked. On Windows it threw this error: 
```
Frame    Code address
  * 0x33c12fd5e0    0x7ff674531062 C:\ghcup\ghc\9.6.7\bin\ghc-9.6.7.exe+0x4cd1062

  Error: [Cabal-7125]
  repl failed for waspc-0.21.0. The build process terminated with exit code 11
```

So I went for a simpler solution since I don't have the tools to go deep on Windows to debug.

waspc/tools/build_libs.ts
```diff
@@ -0,0 +1,105 @@
...
+  console.log(`Installing ${libName} lib (${libDir})`);
+
+  if (libVersion !== waspcVersion) {
+    console.error(
+      `ERROR: ${libName} lib version (${libVersion}) != current Wasp version (${waspcVersion}).`,
+    );
+    console.error(
+      `       Update the lib version in package.json to ${waspcVersion}.`,
+    );
+    throw new Error();
```
infomiho, line 52: Instead of doing `process.exit()`.

waspc/tools/build_libs.ts
```diff
@@ -0,0 +1,105 @@
...
+
+function runCmd(
+  /** @type {string} */ cmd,
+  /** @type {string[]} */ args,
+  { cwd = waspcDir } = {},
+) {
+  return execFileSync(cmd, args, {
+    cwd,
+    encoding: "utf-8",
+    shell: true,
```
infomiho, line 76: Required on Windows to run `npm` and `node`.

waspc/run
```diff
@@ -22,7 +22,10 @@ DEFAULT_COLOR="\033[39m"
 WASP_PACKAGES_COMPILE="for pkg in ${SCRIPT_DIR}/data/packages/*/package.json; do \
   (cd \$(dirname \$pkg) && npm install && npm run build); \
 done"
-WASP_LIBS_COMPILE="${SCRIPT_DIR}/tools/install_libs_to_data_dir.sh"
+# NOTE: --experimental-strip-types allows Node to run TypeScript directly.
+# This flag became the default in Node v22.18+, but we need it explicitly
+# for our minimum supported version (v22.12.0).
+WASP_LIBS_COMPILE="node --experimental-strip-types ${SCRIPT_DIR}/tools/install_libs_to_data_dir.ts"
```
sodic: Just to check: does the flag work if you run this on Node 24? And if not, is that a problem?

waspc/tools/get-waspc-version.ts
sodic: Hm, not sure about this one. I'm fine with a hack if we know what's happening, and it seems we don't yet.

> So I went for a simpler solution since I don't have the tools to go deep on Windows to debug.

But this is also a factor. So I agree, we can stick with this. But if we're going to start taking windows development seriously, we need to find a way to do it.

Please just put everything from this comment into the code (or link to it from the code).

waspc/tools/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "waspc-tools",
+  "type": "module",
+  "devDependencies": {
+    "@types/node": "^22.19.2"
```
sodic, line 5: Can you use the one from `wasp/.nvmrc`? We recently tried to connect all Node versions and reduce duplicatoin.

If that's not possible:
- The comment in `nvmrc` should be updated.
- There should also be an appropriate comment here (check some other places where we list the Node version, you'll see what I mean).

waspc/tools/build_libs.ts
sodic: Please elaborate why in the code comment :)

waspc/tools/build_libs.ts
sodic, line 1: The script is all good, nice work.

But I do think we should be able to just place the files into the data dir directly. You noticed that you wouldn't know how to ignore all the extra stuff, but I think that's taken care of.

Here's how it works now: https://github.com/wasp-lang/wasp/pull/3483/files
We don't have an "ignore list," we have an "allow list" that's coming directly from our cabal file. So:
- If something is listed in waspc.cabal, it ends up in a binary.
- If it's not, it doesn't.

I think that allows us to not move anything around, just build the libs directly in the data directory. Check with @cprecioso please.

waspc/run
infomiho: The CI is passing with latest Node.js version, so I'd say there is no problem 👍

waspc/tools/package.json
infomiho: I've used the same version as the `make-npm-packages` package uses. I'll update the version in that package as well.

waspc/tools/get-waspc-version.ts
infomiho: This is not a hack, just a different implementation that doesn't rely of `cabal repl`. The issue with Windows is that we don't support it, but we have it in the CI so it's half-serious support for now. 

I'll add a comment on top to investigate if we can use `cabal repl` on Windows in the future.

waspc/tools/build_libs.ts
infomiho: It's not visible in this change, but the comment was already above this line. Strange 😄

waspc/tools/package.json
infomiho: In the `nvmrc` we list `v22.12.0` but we don't really honor it in any of the `@types/node` deps.

We use different `@types/node` version in:
- `make-npm-package` (`^22.19.2`)
- `auth` lib (`^22.17.1`)
- `deploy` package (`^22.17.0`)
- `prisma` package (`^22.17.0`)
- `studio` package (`^22.17.0`)
- `ts-inspect` package (`^22.17.0`)
- `wasp-config` package (`^22.17.0`)

Also, we specify `@types/node` as `^22.0.0` in our generated code.

This all equals to _latest minor release_ of `@type/node@22`.

What do we want to do here?

waspc/tools/build_libs.ts
infomiho: I've moved the libs to live in `data` dir, I wasn't aware our setup already worked great for this 👍

waspc/tools/package.json
infomiho: Encoded this in an issue: https://github.com/wasp-lang/wasp/issues/3672

infomiho:
Created the issue here: https://github.com/wasp-lang/wasp/issues/3672

PR #3643, opened 2026-01-21
Add user-defined dependency overrides via wasp.overriddenDeps

cprecioso:
- Resolves #3120
- Part of #3120:
  1. #3655
    - Why: Since we're going to allow overriding with configs at the `package.json` level, we'll need to bring dependency conflicts checking to the PackageJsonValidator.
  2. #3656 (this one)
    - Why: Now that we check for conflicts before going into the generator proper, we can remove that code from it.
  3. #3643
    - The meat of the change

---

## Description

This PR implements support for user-defined npm dependency overrides, allowing developers to use different versions of Wasp-managed dependencies when needed. Users declare overrides via a `wasp.overriddenDeps` object in their package.json, explicitly acknowledging the version Wasp requires to ensure conscious adoption.

This is done through the following changes:
- **Generator.Valid.PackageJson**
  - Adds `package.json#wasp` validation, warning if the override doesn't match the current Wasp requirement
  - In `package.json#*dependencies`, it relaxes constraints if the dependency is present in the override list
- **Generator.NpmDependencies**
  - When writing webapp's and server's package.json, it first merges the hardcoded dependencies with the ones coming from the user's manifest.

<details>
<summary><b>Error message gallery</b></summary>

<table>
<tr>
 <th>Wrong version
 <td>Screenshot 2026-01-26 at 16 01 18
<tr>
 <th>Wrong version but override active<br><em>(The double message is from an old version, it only shows once now)</em>
 <td>Screenshot 2026-01-26 at 16 01 29
<tr>
 <th>Wrong version but wrong override
 <td>Screenshot 2026-01-26 at 16 01 44
<tr>
 <th>Correct version but wrong override
 <td>Screenshot 2026-01-26 at 16 02 10
<tr>
 <th>Unnecessary override
 <td>Screenshot 2026-01-26 at 16 02 56
</table>

</details>

## Type of change

- [x] **🚀 New/improved feature** (non-breaking change which adds functionality)

## Checklist

- [x] I added **unit tests** for my change.
- [x] I **added/updated the documentation** in `web/docs/`.

The implementation is fully tested with 512 unit tests passing, including comprehensive test cases for override validation, conflict detection, and dependency merging.

🧤 Generated with Claude Code

Issue #3120, opened 2025-09-05
Allow the user to override internal dependencies


cprecioso:
Once all the packages we have for a project are in sync, we can find a way to let users override our internal dependencies:

- Users can upgrade dependencies further than what we’ve explicitly support.

  For example, we might only allow `vite@^4` because that’s what we test with; but a user might want to try out `vite@^5` if the breaking changes do not affect our usage of it.

- This is done through an explicit workaround, signalling that they’re out of our happy path and intended support.

Posible solution: Allow usage of the `package.json#override`.

- https://github.com/wasp-lang/wasp/issues/2830

sodic:
Is this bloced by #3233?

cprecioso:
no

cprecioso:
Due to someone asking for React 19, I already tried the `package.json#overrides` route, and it doesn't seem to work for top-level dependencies, as you can't define conflicting ranges in the top-level:

```json
{
  "dependencies": {
    "react": "^18"
  },
  "overrides": {
    "react": "^19"
  }
}
```

&uarr; These two directly conflict, and in `npm`'s perspective they're both under your control, so it tells you to make up your mind. There's no disable button for that check.

cprecioso:
Another option could be to disable our version check with a specific key:

```json
{
  "dependencies": {
    "react": "^19"
  },
  "wasp": {
    "unsupportedOverrides": {
      "react": "^18"
    }
  }
}
```

The idea would be that `unsupportedOverrides` forces you to write the actual version Wasp requires, so that if we update it internally it will tell you that this is not the internal version anymore, and you are explicitly asked to reevaluate if you still need the override.

PR #3655, opened 2026-01-23
Check for possible dependency conflicts in Validator

cprecioso:
- Part of #3120:
  1. #3655 (this one)
    - Why: Since we're going to allow overriding with configs at the `package.json` level, we'll need to bring dependency conflicts checking to the PackageJsonValidator.
  2. #3656
    - Why: Now that we check for conflicts before going into the generator proper, we can remove that code from it.
  3. #3643
    - The meat of the change

---

## Description

This PR improves the package.json dependency validator to detect version conflicts between user-specified dependencies and all Wasp-managed dependencies.

Previously, the validator only checked user dependencies against a small explicit list of required/optional packages. Now it validates against all dependencies used across the entire Wasp ecosystem:
- User requirements (required runtime deps, dev deps, optional deps)
- WebApp generator dependencies
- Server generator dependencies

This prevents subtle version conflicts where a user might install a package version that conflicts with what Wasp internally uses, which could cause hard-to-debug runtime issues.

Key changes:
- Added `toPair` helper to `Dependency` module
- Created new `Common.hs` module with shared dependency specifications and `getAllWaspDependencies` function
- Refactored validator to check optional deps against the full dependency map
- Threaded `AppSpec` through the validator chain to access generator dependency info

## Type of change

- [x] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:


waspc/tests/Generator/Valid/PackageJsonTest.hs
```diff
@@ -66,6 +67,24 @@ spec_PackageJson = do
         <-- (pkgJson `withDep` (depType, ("required-pkg", "2.0.0")))
         ~~> (([fieldName, "required-pkg"] ==) . V.fieldPath)
 
+  describe "makeForbiddenDepValidator" $ do
```
cprecioso, line 70: This function had no unit tests so I added them

waspc/src/Wasp/Generator/Valid/PackageJson/Common.hs
```diff
@@ -0,0 +1,89 @@
...
+
+type DependencySpecification = (P.PackageName, P.PackageVersion)
+
+-- | List of dependencies that are required to be present in the user's
+-- package.json#dependencies.
+requiredUserRuntimeDeps :: [DependencySpecification]
+requiredUserRuntimeDeps =
+  [ -- Installing the wrong version of "react-router-dom" can make users believe that they
+    -- can use features that are not available in the version that Wasp supports.
+    ("react-router-dom", show DepVersions.reactRouterVersion),
```
infomiho: Do we need to update this to `react-router` now? https://github.com/wasp-lang/wasp/pull/3658

Issue #3431, opened 2025-11-25
Specify custom server URL via an explict option


infomiho:
We allow users to specify the server URL that the client will connect to using an environment variable:
```
REACT_APP_API_URL=https://my-server-url.com wasp deploy <provider> deploy
```

This feels like a workaround and not a real 1st part option of the Wasp Deploy feature. It would be better if we exposed an option:
```
wasp deploy <provider> deploy --custom-server-url https://my-server-url.com
```

This also makes it more future-proof as the environment variable might change from one Wasp version to another.

PR #3674, opened 2026-01-28
Add --custom-server-url CLI option for deploy

cprecioso:
- Fixes https://github.com/wasp-lang/wasp/issues/3431

## Description

Add a `--custom-server-url` CLI option to the `wasp deploy` command for both Fly.io and Railway providers. This allows users to explicitly specify the server URL that the client connects to, replacing the workaround of using the `REACT_APP_API_URL` environment variable.

Users can now use:
```bash
wasp deploy fly deploy --custom-server-url https://my-server.com
wasp deploy railway deploy myproject --custom-server-url https://my-server.com
```

The implementation follows a clean architecture where each provider's deploy function resolves the URL (custom option takes priority, otherwise defaults are used), and passes it to the shared `buildClient` function.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

- 🆕 Changelog:

### DIFF
--- a/web/docs/deployment/deployment-methods/wasp-deploy/fly.md
+++ b/web/docs/deployment/deployment-methods/wasp-deploy/fly.md
@@ -6,6 +6,7 @@ title-llm: Automated Deployment to Fly.io with Wasp CLI
 import { Required } from '@site/src/components/Tag';
 import LaunchCommandEnvVars from './\_launch-command-env-vars.md'
 import CiCdMention from './\_ci-cd-mention.md'
+import CustomServerUrlOption from './\_custom-server-url-option.md'
 
 [Fly.io](https://fly.io/) is a platform for running containerized apps and microservices on servers around the world. It makes deploying and managing your apps straightforward with minimal setup.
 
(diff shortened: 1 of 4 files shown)


web/docs/deployment/deployment-methods/wasp-deploy/fly.md
```diff
@@ -261,6 +261,14 @@ If you've added any [client-side environment variables](../../../project/env-var
 REACT_APP_ANOTHER_VAR=somevalue wasp deploy fly launch my-wasp-app dfw
 ```
 
+#### Custom Server URL
```
infomiho: Since we are repeating this section verbatim two times - let's extract it into a partial. Both here and in Railway docs.

web/docs/migration-guides/migrate-from-0-20-to-0-21.md
```diff
@@ -304,7 +313,26 @@ We upgraded our testing support from Vitest v1 to Vitest v4. Most of the breakin
...
 3. [Migration guide from Vitest v3 to v4](https://vitest.dev/guide/migration.html#vitest-4)
 
-### 8. Add `netlify.toml` if deploying to Netlify
+### 8. Update custom server URL usage in deployment
+
+**If you weren't using `REACT_APP_API_URL` environment variable during deployment, you can skip this step.**
+
+If you were using the `REACT_APP_API_URL` environment variable to specify a custom server URL during deployment:
+
+```sh
```
infomiho, line 325: Instead of doing Before/After in comments, let's use tabs like we do in other places.

waspc/data/packages/deploy/src/providers/fly/index.ts
```diff
@@ -70,6 +70,12 @@ class FlyCommand extends Command {
       [],
     );
   }
+  addCustomServerUrlOption(): this {
```
infomiho, line 73: Nipick: if would be awesome if we could somehow reuse this option definition across Fly and Railway. If there is no precedent for this - let's not do it here for the first time.

cprecioso: There is indeed no precedent for this. Resolving

PR #3675, opened 2026-01-28
Small refactors for `wasp news`

sodic:
## Description

Covering some comments:
  - https://github.com/wasp-lang/wasp/pull/3534#discussion_r2640607966
  - https://github.com/wasp-lang/wasp/pull/3534#discussion_r2645382371
  - https://github.com/wasp-lang/wasp/pull/3534#discussion_r2639836966

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:


waspc/cli/src/Wasp/Cli/Command/News/Display.hs
sodic, line 1: What went smoothly:
  - Add trimming (https://github.com/wasp-lang/wasp/pull/3534#discussion_r2640607966 and https://github.com/wasp-lang/wasp/pull/3534#discussion_r2645382371
  
What didn't go smoothly:
- Extract a separator function (https://github.com/wasp-lang/wasp/pull/3534#discussion_r2645386465 and https://github.com/wasp-lang/wasp/pull/3534#discussion_r2640629058). Problems:
  - We want different styling for left text, right text, and the separators. ANSI characters make the length calculation tricky.
  - It went against @Martinsos above suggestion (interpolate `title`, `dots` and `publishedAt` inside trimming). 
  - Separator shave a mimium and a maximum, adding to the complexity and reducing the functions generality
  - Separators have spaces on both sides, which is too edge-casey for a generic function..
  - All in all, this would end up being a very specific function pretending to be something more general, and it would have to do some very strange stuff (because of ANSI characters).
- Use pretty printer (https://github.com/wasp-lang/wasp/pull/3534#discussion_r2640763543). I tried it but gave up, it was too ugly

waspc/src/Wasp/Util.hs
```diff
@@ -196,20 +196,22 @@ trim :: String -> String
 trim = reverse . dropWhile isSpace . reverse . dropWhile isSpace
 
 wrapString :: Int -> String -> String
-wrapString maxLength = intercalate "\n" . map unwords . groupWordsIntoLines . words
+wrapString lineLength = intercalate "\n" . map unwords . groupWordsIntoLines . words
```
sodic, line 199: I spent half a day thinking about how to write this function nicely and couldn't come with anything I was happy with.

The current version implements suggestions from this discussion: https://github.com/wasp-lang/wasp/pull/3534#discussion_r2639836966. I still don't like it (e.g., there's duplication in calculating the `remainingLength` and matching empty ists) but can't justify working on it anymore.
For posterity, here's the original version from the hackathon: https://github.com/wasp-lang/wasp/blob/6b39fac09fbad584b5410df3a5b9f5c211a7bada/waspc/cli/src/Wasp/Cli/Command/News.hs#L134
It's better in some ways but worse in others. I prefer this one so I kept it.

You can try your hand at writing something nicer than that or this, but I warn you - it's a time sink. Feel free to merge whatever you like, I don't mind :)

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
Martinsos: Yeah that trimming part is much nicer now!

As for the fnuction that produces separtor: you could have just written a function that takes two strings and separates them with bunch of `.` and one space before and after those and that is it, given the desired length, a simple local function you then call above. Yeah you don't then have `${title} ${dots} ${publishedAt}` in the `trimming`, butok, maybe in that case we go back to `${titleAndMetadata}`, or to unlines approach, if we value this more.
Or you could have made that function so it doesn't combine the two things with the separator, instead it returns the separator (given the two strings and desired length) and you can use then use it on your own in the `trimming`.

I think any of these work, the main thing I wanted was to make it more readable if easy to do (and it was, taking into account initial code you wrote), and to skip having random `- 2` in there, which is obvious as to why it is bad. And I don't you actually did a good job with those `-2` -> you put it in variable, but how is one to figure out what variable is about? I certianly wouldn't figrue that out.

waspc/cli/src/Wasp/Cli/Command/News/Display.hs
```diff
@@ -1,33 +1,42 @@
...
-    levelLine = showNewsLevel newsEntry.level
-
-    bodyBlock = Term.applyStyles [Term.Grey] $ indent indentSize $ wrapString (maxColumns - indentSize) newsEntry.body
-
-    dotCount = max minDotsCount (maxColumns - length newsEntry.title - length date - 2)
     date = formatTime defaultTimeLocale "%Y-%m-%d" newsEntry.publishedAt
 
-    maxColumns = 80
+    dotCount = max minDotsCount $ maxColumns - length newsEntry.title - length date - spacesInSeparatorCount
+    spacesInSeparatorCount = 2
```
Martinsos: How am I to figure out this is referring to the two spaces around `${space}` in `${title} ${dots} ${publishedAt}`? I think that is very hard to see / connect, so this doesn't really help. You should find a way to solve this, either put a detailed comment on it, or connect it logically, abandon trimming if you want, there are a couple of ways to go, all are good with me as long as there is a connection.

PR #3677, opened 2026-01-28
Parallelize e2e snapshot tests creation

FranjoMindek:
## Description

Currently snapshot tests function in the following way:
1. We create snapshot files by executing `SnapshotTest`'s bash command.
2. We send the created snapshot files to the `Tasty`'s `goldenVsFileDiff` to create a `TestTree`.
3. `Tasty` executes the `TestTree` in parallel (just file comparison, very fast).

The slowest and most time consuming part of this process is 1., which is done sequentially.
This PR parallelizes it.

So we do 1. in parallel and save that data to `SnapshotTestData`.
Then we take that data and create the `TestTree` sequentially.
Finally, we as usually execute the `TestTree` in parallel (default behavior).

Please look at total time (last value) for "duration":
Old time:
image

New time:
image

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change**

### DIFF
--- a/waspc/e2e-tests/Main.hs
+++ b/waspc/e2e-tests/Main.hs
@@ -1,4 +1,5 @@
-import SnapshotTest (runSnapshotTests)
+import Control.Concurrent.Async (mapConcurrently)
+import SnapshotTest (testTreeFromSnapshotTest)
 import System.Info (os)
 import Test (testTreeFromTest)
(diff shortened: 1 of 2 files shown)


FranjoMindek:
Note for future.
I couldn't move snapshot tests to be parallelized by `Tasty` (moving all command execution stuff and IO stuff into `Spec`). At least I didn't find a solution for that.

PR #3678, opened 2026-01-28
Parallelize e2e tests creation

FranjoMindek:
## Description

Currently tests function in the following way:
1. We create a test spec by executing all `TestCase` bash commands of a `Test` (very slow). 
2. The spec is converted into a `TestTree`.
3. `Tasty` executes the `TestTree` in parallel (very fast, just pattern matches exit status to pass or fail).

The slowest and most time consuming part of this process is 1., which is done sequentially.

This PR parallelizes it by moving all `TestCase` shell command executions to happen during `Tasty` test execution, instead of executing them beforehand and letting `Tasty` just compare the `ExitCode`.

Please look at total time (last value) for "duration":
Old time:
image

New time:
image

For me personally it varied from ~1:25 - ~1:45

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change**

### DIFF
--- a/waspc/waspc.cabal
+++ b/waspc/waspc.cabal
@@ -815,7 +815,6 @@ test-suite waspc-e2e-tests
   build-depends:
     , waspc
     , aeson
-    , async
     , directory
     , base
     , filepath


waspc/e2e-tests/Test.hs
```diff
@@ -30,18 +31,49 @@ data TestCase = TestCase
...
-runTest test = do
+-- | Prepares a list of 'Test's in parallel (executing shell commands and capturing exit codes),
+--  and then creates test trees that check the results.
+runTests :: [Test] -> IO [TestTree]
+runTests tests = do
   testOutputsDir <- getTestOutputsDir
   callCommand $ "mkdir -p " ++ fromAbsDir testOutputsDir
 
-  testSpec test.name (mapM_ (createAndExecuteTestCase test.name) test.testCases)
+  mapConcurrently prepareTest tests >>= mapM createTestTree
```
cprecioso: Maybe this is clearer?


```suggestion
  mapConcurrently (prepareTest >=> createTestTree) tests
```

FranjoMindek: Yes, didn't know about `>=>` operator.

PR #3679, opened 2026-01-29
Tests for `isOlderThanNHours`

sodic:
## Description

Adds tests for `isOlderThanNHours` as agreed here: https://github.com/wasp-lang/wasp/pull/3536#discussion_r2655079970

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/tests/UtilTest.hs
+++ b/waspc/tests/UtilTest.hs
@@ -230,13 +230,13 @@ spec_isOlderThanNHours = do
       _3h0m1s_ago <- secondsFromNow (negate $ 3 * 3600 + 1)
       isOlderThanNHours 3 _3h0m1s_ago `shouldReturn` True
 
-    it "returns False when time is newer than N hours" $ do
+    it "returns False when time is not older than N hours" $ do
       _2h59m59s_ago <- secondsFromNow (negate $ 2 * 3600 + 59 * 60 + 59)
       isOlderThanNHours 3 _2h59m59s_ago `shouldReturn` False
 
-    it "returns False when time is in the future" $ do
-      oneSecondFromNow <- secondsFromNow 1
-      isOlderThanNHours 1 oneSecondFromNow `shouldReturn` False
+    it "returns False when time was exactly N hours ago" $ do
+      threeHoursAgo <- secondsFromNow (negate $ 2 * 3600 + 59 * 60 + 59)
+      isOlderThanNHours 3 threeHoursAgo `shouldReturn` False
 
     it "returns True for past time when N is 0" $ do
       oneSecondAgo <- secondsFromNow (negate 1)


waspc/tests/UtilTest.hs
```diff
@@ -221,3 +222,28 @@ spec_checkIfEnvValueIsTruthy = do
...
+
+spec_isOlderThanNHours :: Spec
+spec_isOlderThanNHours = do
+  describe "isOlderThanNHours" $ do
+    it "returns True when time is older than N hours" $ do
+      _3h0m1s_ago <- secondsFromNow (negate $ 3 * 3600 + 1)
+      isOlderThanNHours 3 _3h0m1s_ago `shouldReturn` True
+
+    it "returns False when time is newer than N hours" $ do
+      _2h59m59s_ago <- secondsFromNow (negate $ 2 * 3600 + 59 * 60 + 59)
```
sodic, line 234: I didn't use words because the name would be ridiculously long.

waspc/tests/UtilTest.hs
```diff
@@ -221,3 +222,28 @@ spec_checkIfEnvValueIsTruthy = do
           ]
     checkIfEnvValueIsTruthy . fst <$> testCases
       `shouldBe` snd <$> testCases
+
+spec_isOlderThanNHours :: Spec
+spec_isOlderThanNHours = do
```
sodic, line 227: I didn't use property tests because they seemed like an overkill. Here's how that might have looked like:
```hs
spec_isOlderThanNHours :: Spec
spec_isOlderThanNHours = do
  describe "isOlderThanNHours" $ do
    prop "returns True when time is older than N hours" $
      forAll (choose (0, 1000)) $ \(nHours :: Int) ->
        forAll (choose (1, 10000)) $ \(extraSeconds :: Int) ->
          ioProperty $ do
            time <- secondsFromNow (negate $ fromIntegral nHours * 3600 + fromIntegral extraSeconds)
            isOlderThanNHours (fromIntegral nHours) time

    prop "returns False when time is newer than N hours" $
      forAll (choose (1, 1000)) $ \(nHours :: Int) ->
        forAll (choose (1, 10000)) $ \(marginSeconds :: Int) ->
          ioProperty $ do
            time <- secondsFromNow (negate $ fromIntegral nHours * 3600 - fromIntegral marginSeconds)
            not <$> isOlderThanNHours (fromIntegral nHours) time

    prop "returns False when time is in the future" $
      forAll (choose (0, 1000)) $ \(nHours :: Int) ->
        forAll (choose (1, 10000)) $ \(futureSeconds :: Int) ->
          ioProperty $ do
            time <- secondsFromNow (fromIntegral futureSeconds)
            not <$> isOlderThanNHours (fromIntegral nHours) time
```

waspc/tests/UtilTest.hs
```diff
@@ -221,3 +222,28 @@ spec_checkIfEnvValueIsTruthy = do
...
+  describe "isOlderThanNHours" $ do
+    it "returns True when time is older than N hours" $ do
+      _3h0m1s_ago <- secondsFromNow (negate $ 3 * 3600 + 1)
```
FranjoMindek, line 230: For me is weird how we didn't abstract `3600` or (`60`) concept somewhere in our time modules.

Like seconds to minutes and minutes to hours, the opposite order (hours to minutes...) or anything in that manner.

Nothing really related to this test PR, but we could've used it here too instead of magic numbers.

waspc/tests/UtilTest.hs
FranjoMindek, line 234: Wouldn't actual lower border be 3h? Do we want to use that instead?
It uses `>` not `>=`?

waspc/tests/UtilTest.hs
```diff
@@ -221,3 +222,28 @@ spec_checkIfEnvValueIsTruthy = do
...
+      isOlderThanNHours 3 _3h0m1s_ago `shouldReturn` True
```
FranjoMindek, line 231: Why `3` hours?
Wouldn't `1` be the simplest we can go :D?

waspc/tests/UtilTest.hs
```diff
@@ -221,3 +222,28 @@ spec_checkIfEnvValueIsTruthy = do
...
+
+    it "returns False when time is newer than N hours" $ do
+      _2h59m59s_ago <- secondsFromNow (negate $ 2 * 3600 + 59 * 60 + 59)
+      isOlderThanNHours 3 _2h59m59s_ago `shouldReturn` False
+
+    it "returns False when time is in the future" $ do
```
FranjoMindek: This is the same as `returns False when time is newer than N hours` just not an edge case? Why then don't we have a general test for `True and past` then?

waspc/tests/UtilTest.hs
FranjoMindek: Yes, I think we just need to test the edge cases and maybe general cases. That is all. It's just numbers and math, so it can't be unpredictable.

waspc/tests/UtilTest.hs
sodic: I don't know. I just chose it randomly, 1 is the lowest possible number the function can take so it seemed "not arbitrary enough," if that makes sense.

waspc/tests/UtilTest.hs
sodic: Not quite:
 1. Time is more than N hours in the past (first test)
 2. Time is less than N hours in the past (second test)
 3. Time is in the future (third test)

There are two inflection points important for the function: the current time, and N hours before the current time. Those two points divide the input domain into three distinct areas (listed above). We want to test all three areas.

When N is 0, the two inflection points become the same one, so we only need to test two areas - the future and the past.

waspc/tests/UtilTest.hs
sodic: Hm, not necessarily. There are numbers that are so well known that extracting them to a constant reduces readbility.
  - For example, `linuxProcessLimit` obviously should be a constant and not a magic number.
  - On the other hand, `roundtripDistance = tripDistance * 2` is better than `roundtripDistance = tripDistance * numTripsInRoundtrip` :)
  
These things are right on the line I think - I wouldn't mind seeing extracted constants, but they're so universal that it doesn't make a difference.

waspc/tests/UtilTest.hs
sodic: I've added an extra case that tests exactly 3 hours.

PR #3680, opened 2026-01-29
Support for .env.client file

infomiho:
## Description

Adds support for `.env.client` file and disables the default Vite `.env` loading behavior. 

Part of #3473 

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:


waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/detectServerImports.ts
```diff
@@ -2,9 +2,8 @@
 import { type Plugin } from 'vite'
 import path from 'path'
 
-let parsePathToUserCode!: ParsePathToUserCodeFn;
```
infomiho, line 5: @cprecioso requested this change - I forgot to do it in a previous PR

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/waspConfig.ts
```diff
@@ -0,0 +1,56 @@
+{{={= =}=}}
+import { type PluginOption, mergeConfig } from "vite";
+import { defaultExclude } from "vitest/config"
+
+export function waspConfig(): PluginOption {
+  return {
+    name: "wasp:config",
+    enforce: 'pre',
```
infomiho, line 8: Moved the `wasp:config` plugin to a separate file and added `enforce: 'pre'` so it would run before the `wasp:env-file` plugin.

waspc/src/Wasp/Generator/SdkGenerator/Client/VitePluginG.hs
```diff
@@ -22,29 +22,34 @@ import Wasp.Project.Common
...
       genDetectServerImportsPlugin,
-      genValidateEnvPlugin,
-      genTypescriptCheckPlugin
+      genFileCopy [relfile|validateEnv.ts|],
+      genFileCopy [relfile|typescriptCheck.ts|]
     ]
     <++> getVirtualModulesPlugin spec
     <++> genHtmlPlugin spec
+  where
+    genFileCopy = return . C.mkTmplFd . (C.vitePluginsDirInSdkTemplatesDir </>)
```
infomiho, line 42: Introduced this helper to clean up the code a bit.

waspc/src/Wasp/Generator/SdkGenerator.hs
```diff
@@ -214,6 +214,8 @@ npmDepsForSdk spec =
           [ ("@prisma/client", show prismaVersion),
             ("prisma", show prismaVersion),
             ("axios", show axiosVersion),
+            ("dotenv", "^17.2.3"),
```
infomiho: We also have `dotenv` in the server deps. It's a different version. Do we want to extract this version into deps versions and use the same version in the SDK and server?

infomiho:
Comment by Carlos:

> if you're making more files participate in the build process, you should [call this.addWatchFile() on it](https://rollupjs.org/plugin-development/#this-addwatchfile) so that the build is reloaded if the user edits it

waspc/src/Wasp/Generator/SdkGenerator.hs
cprecioso: For sure

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/validateEnv.ts
```diff
@@ -1,11 +1,12 @@
...
+import { loadWaspEnvClient } from './envFile.js'
 import {
   getValidatedEnvOrError,
   formatZodEnvErrors,
-} from 'wasp/env/validation'
-import { clientEnvSchema } from 'wasp/client/env/schema'
-import { getColorizedConsoleFormatString } from 'wasp/universal/ansiColors'
+} from '../../../env/validation.js'
+import { clientEnvSchema } from '../../env/schema.js'
+import { getColorizedConsoleFormatString } from '../../../universal/ansiColors.js'
```
cprecioso, line 9: We're compiling these with `tsc`, right? shouldn't it take care of the extensions?

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/envFile.ts
```diff
@@ -0,0 +1,87 @@
...
+      return {
+        // Disable Vite's default .env loading.
+        envDir: false,
+        define: prefixedVars,
+      }
+    },
+    buildStart() {
+      if (existsSync(envFilePath)) {
+        this.addWatchFile(envFilePath)
+      }
```
cprecioso: i wouldn't gate this behind `existsSync`, if the file didn't exist but it starts existing, we want to rebuild too!

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/envFile.ts
```diff
@@ -0,0 +1,87 @@
...
+    },
+    handleHotUpdate(ctx) {
+      if (ctx.file === envFilePath) {
+        // We need to do a full server reload to ensure the new env vars are picked up.
+        ctx.server.restart()
```
cprecioso: TIL!

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/envFile.ts
```diff
@@ -0,0 +1,87 @@
...
+    name: 'wasp:env-file',
+    enforce: 'pre',
+    config(config) {
+      const rootDir = config.root || process.cwd()
+      const envVars = loadWaspEnvClient(rootDir, config.envPrefix!)
+      envFilePath = resolve(rootDir, envFileName)
+
+      const prefixedVars = Object.entries(envVars)
+        .reduce((acc, [key, value]) => {
+          acc[`import.meta.env.${key}`] = JSON.stringify(value)
```
cprecioso, line 66: what if an env has non-valid-identifier character (like `-`)? we should handle that

i think this works:


```suggestion
          acc[`import.meta.env[${ JSON.stringify(key) }]`] = JSON.stringify(value)
```

even if it's defined with the `obj["property"]` syntax, it does replace the `obj.property` too [(playground link)](https://esbuild.github.io/try/#dAAwLjI3LjIAewogIGRlZmluZTogewogICAgJ2ltcG9ydC5tZXRhLmVudlsiSE9MQSJdJzogJyJ0ZXh0IicsCiAgICAnaW1wb3J0Lm1ldGEuZW52WyJBRElPUy1ZRVMiXSc6ICcidGV4dCInLAogIH0sCn0AY29uc29sZS5sb2coaW1wb3J0Lm1ldGEuZW52LkhPTEEp)

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/envFile.ts
```diff
@@ -0,0 +1,87 @@
...
+import { type Plugin, type UserConfig } from 'vite'
+import { resolve } from 'node:path'
+import { existsSync, readFileSync } from 'node:fs'
+import { parse as parseDotenv } from 'dotenv'
+import { expand, type DotenvPopulateInput } from 'dotenv-expand'
+
+const envFileName = '{= clientEnvFileName =}'
+
+// Based on: https://github.com/vitejs/vite/blob/8bb32036792a6f522f5c947112f3d688add755a0/packages/vite/src/node/env.ts
+export function loadWaspEnvClient(rootDir: string, envPrefix: NonNullable<UserConfig['envPrefix']>): Record<string, string> {
```
cprecioso: I'd convert the fs functions in here to be async, we probably want to be maximally parallel and not block the bundler to do other internal initialization routines in the background.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/envFile.ts
```diff
@@ -0,0 +1,87 @@
...
+      }
+    }
+
+    // Check if there are actual env variables starting with the prefix
+    // these are typically provided inline and should be prioritized.
+    for (const key in process.env) {
+      if (envPrefixNormalized.some(prefix => key.startsWith(prefix))) {
+        env[key] = process.env[key] as string
+      }
+    }
```
cprecioso, line 44: suggestion: i have some trouble understanding what exactly these checks are protecting us from. some examples would help greatly.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/envFile.ts
infomiho: Our system doesn't cover the cases of file appearing after the dev server is started and it doesn't cover the case of file being deleted. I've tested vanilla Vite and they cover both add, remove and modify actions. Our system only covers modify. 

I'll research a bit more what can be done. I've removed the exists check as it didn't do anything.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/envFile.ts
infomiho: I've followed the internal implementation of Vite (and their comment is copied verbatim, I think I'll adjust it a bit).

https://github.com/vitejs/vite/blob/8bb32036792a6f522f5c947112f3d688add755a0/packages/vite/src/node/env.ts#L83

The "why" of this comment is covered by this Vite PR: https://github.com/vitejs/vite/commit/e2ea6afd47057bac798abb2ae85cc0f529e9ab4d

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/envFile.ts
infomiho: I was following what Vite code did (they didn't preprocess the key like that) + Wasp compiler will throw on invalid env keys: 

```
wasp-cli: user error (Failed to parse dot env file: /Users/ilakovac/dev/wasp/examples/kitchen-sink/.env.client:2:16:
  |
2 | REACT_APP_ADIOS-YES="test"
  |                ^
unexpected '-'
expecting '=', '_', digit, or letter
)
```

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/validateEnv.ts
infomiho: Without the extension, Vite fails with:
```ts
[ Client!] failed to load config from /Users/ilakovac/dev/wasp/examples/kitchen-sink/vite.config.ts
[ Client!] error when starting dev server:
[ Client!] Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/Users/ilakovac/dev/wasp/examples/kitchen-sink/.wasp/out/sdk/wasp/dist/client/vite/plugins/waspConfig' imported from /Users/ilakovac/dev/wasp/examples/kitchen-sink/.wasp/out/sdk/wasp/dist/client/vite/plugins/wasp.js
[ Client!]     at finalizeResolution (node:internal/modules/esm/resolve:274:11)
[ Client!]     at moduleResolve (node:internal/modules/esm/resolve:859:10)
[ Client!]     at defaultResolve (node:internal/modules/esm/resolve:983:11)
[ Client!]     at #cachedDefaultResolve (node:internal/modules/esm/loader:731:20)
[ Client!]     at ModuleLoader.resolve (node:internal/modules/esm/loader:708:38)
[ Client!]     at ModuleLoader.getModuleJobForImport (node:internal/modules/esm/loader:310:38)
[ Client!]     at ModuleJob._link (node:internal/modules/esm/module_job:182:49)
```

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/envFile.ts
infomiho: Alrighite, we can directly add event handlers to the server's watcher:
```ts
configureServer(server) {
  const reloadServerOnEnvFileEvent = (path: string) => {
    if (path === envFilePath) {
      server.restart()
    }
  }

  server.watcher.on('add', reloadServerOnEnvFileEvent)
  server.watcher.on('change', reloadServerOnEnvFileEvent)
  server.watcher.on('unlink', reloadServerOnEnvFileEvent)
},
```

waspc/src/Wasp/Generator/SdkGenerator.hs
```diff
@@ -214,7 +215,7 @@ npmDepsForSdk spec =
           [ ("@prisma/client", show prismaVersion),
             ("prisma", show prismaVersion),
             ("axios", show axiosVersion),
-            ("dotenv", "^17.2.3"),
```
cprecioso, line 217: Can't we upgrade the old version instead of downgrading the new one?

infomiho: I decided to keep us on version 16 since version 17 outputs logs by default: https://github.com/motdotla/dotenv/issues/876

I'd say let's figure out if we want to stay on `dotenv` in the future - maybe we'll be using native `.env` file loading in the server and ~~let's see what happens with Vite 8~~.

infomiho: Vite 8 still uses `dotenv@17` but they only use `parse` so they don't have the same problem as we do. Using `dotenv@17` forces us to add extra env variables to suppress the default logging. I disliked that and I don't think I should do the bump in this PR because it's not just a bump.

infomiho: Created an issue to tackle this later: https://github.com/wasp-lang/wasp/issues/3718

PR #3682, opened 2026-01-29
Implement exact and tilde version support for AppSpec Wasp version

FranjoMindek:
## Description

Add support for `X.Y.Z` and `~X.Y.Z` version in Wasp's `AppSpec`.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/src/Wasp/AppSpec/Valid.hs
+++ b/waspc/src/Wasp/AppSpec/Valid.hs
@@ -93,12 +93,16 @@ validateWaspVersion specWaspVersionString = eitherUnitToErrorList $ do
   where
     parseAndValidateWaspVersion :: String -> Either ValidationError SV.Range
     parseAndValidateWaspVersion waspVersionString =
+      -- TODO: It would be nice to create a general verison range parse
+      -- and use that instead.
       case WV.parseWaspVersion waspVersionString of
         Right waspVersionRange
           -- Ensure the input matches exactly what we parsed (no trailing garbage,
           -- no partial versions like "0.5" which would parse as "0.5.0").
           | show waspVersionRange == waspVersionString -> Right waspVersionRange
-        _ -> Left $ GenericValidationError "Wasp version should be in the format ^major.minor.patch, ~major.minor.patch, or major.minor.patch"
+        _ ->
+          -- TODO: Show the 'Parsec' error message to the user.
+          Left $ GenericValidationError "Wasp version should be in the format ^major.minor.patch, ~major.minor.patch, or major.minor.patch"
 
     incompatibleVersionError :: SV.Version -> SV.Range -> ValidationError
     incompatibleVersionError actualVersion expectedVersionRange =
--- a/waspc/src/Wasp/SemanticVersion/Version.hs
+++ b/waspc/src/Wasp/SemanticVersion/Version.hs
@@ -53,6 +53,8 @@ nextBreakingChangeVersion = \case
   (Version 0 x _) -> Version 0 (succ x) 0
   (Version x _ _) -> Version (succ x) 0 0
 
+-- FIXME: this is incorrect, ~ behaves different depending if
+-- the minor is defined, but we don't support that in our current implementation.
 nextApproximatelyEquivalentVersion :: Version -> Version
 nextApproximatelyEquivalentVersion = \case
   (Version 0 x y) -> Version 0 x (succ y)
--- a/waspc/tests/SemanticVersionTest.hs
+++ b/waspc/tests/SemanticVersionTest.hs
@@ -117,6 +117,9 @@ spec_SemanticVersion = do
     Range [backwardsCompatibleWith [v|0.2.3|]] ~> [vi| [0.2.3, 0.3.0) |]
     Range [backwardsCompatibleWith [v|1.2.3|]] ~> [vi| [1.2.3, 2.0.0) |]
     Range [approximatelyEquivalentWith [v|1|]] ~> [vi| [1, 2) |]
+    -- TODO: This should pass once we refactor the 'SemanticVersion' module.
+    -- See: https://github.com/wasp-lang/wasp/issues/3699
+    -- Range [approximatelyEquivalentWith [v|1.0.0|]] ~> [vi| [1, 1.1.0) |]
     Range [approximatelyEquivalentWith [v|1.2|]] ~> [vi| [1.2, 1.3) |]
     Range [approximatelyEquivalentWith [v|1.2.3|]] ~> [vi| [1.2.3, 1.3.0) |]
     Range [approximatelyEquivalentWith [v|0.2.3|]] ~> [vi| [0.2.3, 0.2.4) |]


waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -91,24 +91,32 @@ validateWaspVersion :: String -> [ValidationError]
...
+        Nothing -> Left badWaspVersionFormatError
+
+    parsePrefixedVersion :: String -> String -> (SV.Version -> SV.ComparatorSet) -> Maybe SV.Range
+    parsePrefixedVersion waspVersionRangeString prefix versionToComparatorSet = do
+      versionStr <- stripPrefix prefix waspVersionRangeString
+      case parseVersion versionStr of
+        -- Ensure the version string is exactly "major.minor.patch" format
+        -- by checking that the parsed version gives us back the input.
+        -- This rejects partial versions like "0.5" and trailing garbage like "0.5.0;2".
+        Right version | show version == versionStr -> Just $ SV.Range [versionToComparatorSet version]
```
FranjoMindek: Version parser succeeds even if version string is followed by garbage.

E.g. `0.5;2` is parsed as `0.5.0`.
Therefore to avoid accepting such scenarios for now I required the parsed version string to be equal to the raw input string.

waspc/tests/SemanticVersion/VersionTest.hs
```diff
@@ -18,6 +18,7 @@ spec_SemanticVersion_Version = do
     parseVersion "1.2.3foobar" `shouldBe` Right (Version 1 2 3)
     parseVersion "1.2.3-alpha" `shouldBe` Right (Version 1 2 3)
     parseVersion "1.2.3 some other stuff" `shouldBe` Right (Version 1 2 3)
+    parseVersion "0.5;2" `shouldBe` Right (Version 0 5 0)
```
FranjoMindek, line 21: This is the reason we need to guard in `parsePrefixedVersion`.

waspc/data/Generator/libs/auth/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@wasp.sh/lib-auth",
-  "version": "0.21.0",
```
FranjoMindek, line 3: This is a new process!

waspc/src/Wasp/SemanticVersion/Version.hs
```diff
@@ -51,3 +52,9 @@ nextBreakingChangeVersion = \case
   (Version 0 0 x) -> Version 0 0 (succ x)
   (Version 0 x _) -> Version 0 (succ x) 0
   (Version x _ _) -> Version (succ x) 0 0
+
+getTildeVersionUpperBound :: Version -> Version
```
FranjoMindek: In general, I have a problem with naming stuff in this PR. This is because we use a semantic name for our caret operator (`^`), which is "backwards compatible with".

The problem is that there is no real semantic meaning to `~`, maybe approximately equivalent to?
I'm not sure what to call this function.

Same problem appears at `TildeCompatibleWith`.
I'm not sure how to name it.

waspc/data/Generator/libs/auth/package.json
Martinsos: What does this mean?

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -91,24 +91,32 @@ validateWaspVersion :: String -> [ValidationError]
 validateWaspVersion specWaspVersionStr = eitherUnitToErrorList $ do
   specWaspVersionRange <- parseWaspVersionRange specWaspVersionStr
   unless (SV.isVersionInRange WV.waspVersion specWaspVersionRange) $
-    Left $
-      incompatibleVersionError WV.waspVersion specWaspVersionRange
+    Left (incompatibleVersionError WV.waspVersion specWaspVersionRange)
   where
-    -- TODO: Use version range parser from SemanticVersion when it is fully implemented.
```
Martinsos, line 97: Nice, TODO got resolved :D. So it actually does happen :D.

waspc/src/Wasp/AppSpec/Valid.hs
Martinsos: Yeah, that is actually fine for version parser -> you want parsers to succeed regardless of what follows, because then you can use it in place where something follows the version, combine it with other parsers, ... .
And if you do need nothing to follow, you can always do that, while you couldn't do the opposite if parser was consuming and failing on waht comes after the valid string. So that is general approach to writing parser combinators.

As for situation here: ok, I guess it makes sense to be strict, why not.

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -91,24 +91,32 @@ validateWaspVersion :: String -> [ValidationError]
...
+    parsePrefixedVersion waspVersionRangeString prefix versionToComparatorSet = do
+      versionStr <- stripPrefix prefix waspVersionRangeString
+      case parseVersion versionStr of
+        -- Ensure the version string is exactly "major.minor.patch" format
+        -- by checking that the parsed version gives us back the input.
+        -- This rejects partial versions like "0.5" and trailing garbage like "0.5.0;2".
+        Right version | show version == versionStr -> Just $ SV.Range [versionToComparatorSet version]
+        _ -> Nothing
+
+    badWaspVersionFormatError :: ValidationError
```
Martinsos: Sucks a bit that this is somewhat far away from the lines above where we dfine that we actually care only about "", "~" and "^". WOuld be better if those were veyr close, so I can't miss updating this comment if I update that logic. Or that they are somehow connected logically, not hardcoded.

waspc/src/Wasp/AppSpec/Valid.hs
Martinsos: I think this works correctly, but it looks quite fragile, specifically the part where you specify certain prefixes and strip them.

Ideally, you would instead:
1. Call general semantic version parser to parse the version, then inspect the result to confirm it is conforming to your restrictions (which are "", "~" or "^") and that is it. Inspection would ideally be easy to do -> if you find yourself havign to dig to deep into internals of how semantic version object is constructed, then I would consider adding some helper functions in the semantic version module itself that can answer these questions easily.
2. Call multiple parsers, and if one of them passes, you are good. That is how it is usually done in parser combinators. So you would have three parsers, one for exact version, one for tilde version, one for caret version. They are very easy to implement, you just write a new Parsec parser that first consumes a single char (or nothing), and then calls the version parser. You would call these one after another, there are combinator functions in Parsec that do this for you, I think there is `choice`, and you might have to also use `try`, but in any case, quite simple, I believe LLM can write that correctly.

I kind of prefer the first approach, but maybe I would change my mind to the second one if I went about implementing, hard to say. But I start from the first one in your case and switch to the second one if that doesn't feel right.

Additional question: Before we had only `^` if I am correct. I believe the reason was that I didn't want to bother implementing other stuff, but I do remember also that there was additional reason -> I think we didn't need more really? I am just wondering, are we relying anywhere in our system on the fact that it was ^ and nothing else? Have you researched that? I would put a bit of research time into that, also ask Claude Code to look for it, and yourself do a grep for `^` maybe hm.
And then the question is also: how do we decide what to support now? Ok, now you support exact version, tilde and caret. Why not more, why only that? WHy not any kind of semantic version? I am not saying we should, but if not, why not? Does constraining the version give us some guarantees that are important to us, for the rest of the tooling or something?

waspc/src/Wasp/SemanticVersion/Version.hs
Martinsos: Well caret operator does have a semantic name yeah, so I think it is good we used it.

The problem I guess is that tilde doesn't. How would one describe it? Allow changes only in the patch version, right? You could call it "patchUpdatesOnly" or something.
Ok I did quick googling https://stackoverflow.com/questions/22343224/whats-the-difference-between-tilde-and-caret-in-package-json -> "approximately equivalent" seems to be the term used. So I would go with that probably.

Or, if you want to go with "tilde" instead, then ok go with that, but don't call it "TildeCompatibleWith" hah, instead you would call the Comparator just `Tilde`! Right? WHy stick half of some other name on it?

Martinsos: Btw I am not sure if this function should be here, doesn't feel like it should. The one above, `nextBreakingChangeVersion`, that one is purely property of the version -> it has breaking, minor and patch parts of it, so we are looking into those properties with this function.

But, tilde is an operator on top of it kind of, I would say it goes beyond the concerns of mere version, so I would consider  pushing this function into the SemanticVersion module.
Btw why not implement function that returns both upper and lower bound for tilde, like we do for the other comparators? Aha I see you do have that function, but you use this here to implemnt it basically. Yeah I would move this to SemanticVersion module.

Martinsos: Ok one more thing, I am just reading semver spec, it says:

```
Allows patch-level changes if a minor version is specified on the comparator. Allows minor-level changes if not.
```

This is the relevant part in the spec: https://github.com/npm/node-semver?tab=readme-ov-file#tilde-ranges-123-12-1 .

This means it is not the same if we have `~1` or `~1.0` if I am correct. Yeah, check the exapmles there. I guess that is why the name `approximatelly equivalent` makes sense hah. But yeah it is tricky because it changes behaviour based on minor version being specified or not. I think we are not handling this correctly right now then, right?

DId you check this spec while implementing this? It is linked at the top of the module, it says we are following it.
Make sure to also add their stuff as test cases, that will be a good start to implementing it correclty.

waspc/data/Generator/libs/auth/package.json
FranjoMindek: Just that now we also have to update lib versions when we update `waspc.cabal`.
Perfect in time for `./run version-bump` 😉

FranjoMindek:
We've decided to close this PR because implementing the `~` operator is impossible with current `SemanticVersion` implementation.
https://github.com/wasp-lang/wasp/issues/3699

We will open a new PR with exact version only.

PR #3684, opened 2026-01-30
Add libs tests in CI

### LINKED ISSUE #3068: Run libs tests in the CI
We want to run the tests that are written for all libraries in the CI.

infomiho:
## Description

- Runs Wasp libs tests in the CI. 
- Reuses code from the `build:libs` script.

Closes #3068 

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/examples/ask-the-documents/package-lock.json
+++ b/examples/ask-the-documents/package-lock.json
@@ -6143,7 +6143,7 @@
     "node_modules/@wasp.sh/lib-auth": {
       "version": "0.21.0",
       "resolved": "file:.wasp/out/libs/auth/wasp.sh-lib-auth-0.21.0.tgz",
-      "integrity": "sha512-0LXoJPLnb/UNJW80Wjh4BEG8/oIu8lVKIMmuArSeI+qt/lIkX190asobhau89ZuXvDdQrGnK79poPglG14iBjQ==",
+      "integrity": "sha512-wVcvXnkNUhVKa1JHfzgeibWaGXBv6jkzI0geFMp9J8QRMO6zkh3CWNdSw4vyV1e8ZjmmppnlhcFISE8nDDq9Jg==",
       "license": "MIT",
       "dependencies": {
         "@node-rs/argon2": "^2.0.2",
--- a/examples/kitchen-sink/package-lock.json
+++ b/examples/kitchen-sink/package-lock.json
@@ -3157,7 +3157,7 @@
     "node_modules/@wasp.sh/lib-auth": {
       "version": "0.21.0",
       "resolved": "file:.wasp/out/libs/auth/wasp.sh-lib-auth-0.21.0.tgz",
-      "integrity": "sha512-0LXoJPLnb/UNJW80Wjh4BEG8/oIu8lVKIMmuArSeI+qt/lIkX190asobhau89ZuXvDdQrGnK79poPglG14iBjQ==",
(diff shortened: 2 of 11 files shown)


waspc/run
```diff
@@ -44,6 +44,8 @@ WASP_CLI_TESTS_CMD="cabal test wasp-cli-tests"
 
 WASPLS_TESTS_CMD="cabal test waspls-tests"
 
+LIBS_TESTS_CMD="node --experimental-strip-types ${SCRIPT_DIR}/tools/libs/test.ts"
```
FranjoMindek, line 47: :o typescript scripts, I hope soon we can remove the `--experimental-strip-types` flag.

waspc/tools/libs/lib-utils.ts
```diff
@@ -0,0 +1,32 @@
+/// <reference types="node" />
+
+import { execFileSync, type ExecFileSyncOptions } from "node:child_process";
+import { readdirSync, readFileSync } from "node:fs";
+import { join } from "node:path";
+
+export function discoverLibDirs(dataLibsDirPath: string): string[] {
+  return readdirSync(dataLibsDirPath, { withFileTypes: true })
+    .filter((dirent) => dirent.isDirectory())
```
FranjoMindek: `dirent` got me really confused what it is.
I then read the type and it was `DirectoryEntry`, ok okay so it's not a single word but two, but it also wasn't too helpful without reading it`s description. Not that we can fix that :D.

I would at least give proper casing `dirEnt`, because I didn't read it as two words.

waspc/tools/libs/test.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+}
+
+function testLib(libDir: string): void {
+  const { name: libName } = getLibPackageJson(libDir);
+
+  console.log(`Testing ${libName} lib (${libDir})`);
+
+  runCmd("npm", ["install"], { cwd: libDir, stdio: "inherit" });
+  runCmd("npm", ["run", "check"], { cwd: libDir, stdio: "inherit" });
```
FranjoMindek: Not really relevant to this PR (you can't fix it here), but it feels weird that tests are under `check`?

Mostly because there are [`npm-check`](https://www.npmjs.com/package/npm-check) and [`npm-check-updates`](https://www.npmjs.com/package/npm-check-updates) packages, which are fairly well-known in `npm` ecosystem. So one might confuse them with that. I thought it was one of those at first, before checking the `package.json`.

waspc/tools/libs/utils.ts
FranjoMindek, line 1: I would say `libs/utils.ts` is the same as `libs/lib-utils.ts`, but RAW.

waspc/run
infomiho: Since `22.18.0` it's enabled by default and we don't need the flag. We require minimum of `22.12`, so until we bump that, we need to keep the flag 😄

waspc/tools/libs/lib-utils.ts
infomiho: Good catch, I just went with the name Claude suggested when I refactored this bit. I'll improve the name 👍

waspc/tools/libs/test.ts
infomiho: Makes sense, it would probably make sense to group everything under `test` and name other scripts with the `test:` prefix.

waspc/tools/libs/utils.ts
infomiho: I wanted to avoid the generic `utils.ts` filename, but I guess,`lib-utils` is just pretending to be better 😂 I'll simplify the name 👍

PR #3697, opened 2026-02-02
fix: move version to first line in .nvmrc

### LINKED ISSUE #3696: .nvmrc contains comments which breaks fnm/nvm
The `.nvmrc` has comments before the version number, which breaks fnm/nvm:

```
Can't find an installed Node version matching # important: this should be set to...
```

The version needs to be on line 1 - comments can go after. See [nvm docs](https://github.com/nvm-sh/nvm#nvmrc).

Also: fnm doesn't support comments yet ([fnm#1209](https://github.com/Schniz/fnm/issues/1209)), but putting the version first works anyway since it only reads the first line.

Pipboyguy:
Fixes #3696

Comments before the version break fnm/nvm. The version needs to be on line 1.

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,3 +1,4 @@
+v22.12.0
 # IMPORTANT: This should be set to the lowest Node version Wasp claims to
 # support. The CI reads this file when running tests.
 # If you change Wasp's lowest official supported Node version here, you must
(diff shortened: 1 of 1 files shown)


cprecioso:
I removed the `fix:` part since this is not a bugfix impacting end users, but everything was fine, thanks for the fix!

Issue #3699, opened 2026-02-02
Fix the semantic version module in Wasp


FranjoMindek:
## **Describe the bug**

Current `waspc/src/Wasp/SemanticVersion.hs` implementation does not properly follow:
1. Semver `version` specification - https://semver.org/
2. Node's semver specification - https://github.com/npm/node-semver

This happend due to us combining two different concepts together:
1. Semver `version`
2. Node's semver `partial version` used in it's `range` implementation.

So to solve the issue, we should separate our `version` into those two different concepts.

### Semver version specification

We don't follow the `version` specification properly.
The specification says:

> A normal version number MUST take the form X.Y.Z where X, Y, and Z are non-negative integers, and MUST NOT contain leading zeroes. X is the major version, Y is the minor version, and Z is the patch version. Each element MUST increase numerically. For instance: 1.9.0 -> 1.10.0 -> 1.11.0.

1. Our `version` parser accepts versions with less than 3 values (e.g. X.Y, or X). This is not allowed per semver specification. 

2. Less concerning, but we allow leading zeroes in the parser.

If we look at it from perspective of our tests, we should fail the first 3 cases (but we pass them): 
```hs
it "parseVersion" $ do
  parseVersion "1" `shouldBe` Right (Version 1 0 0) -- PASS, but should FAIL
  parseVersion "1.2" `shouldBe` Right (Version 1 2 0)  -- PASS, but should FAIL
  parseVersion "1.2.03" `shouldBe` Right (Version 1 2 3)  -- PASS, but should FAIL
  parseVersion "1.2.3" `shouldBe` Right (Version 1 2 3) -- PASS
```

The new `version` implementation should fail them accordingly.

### Node's semver specification

We didn't fully implement the node semver specification. This was intentional as to save work in that point of time. But now when we actually want to fully implement it, we encountered a problem. We can't implement the full specification due to improper `range` implementation.

We implemented the node semver `range` on top of our semver `version`. This is incorrect because `range` internally uses `partial version`, which is a different concept from the semver `version`.

Our `version` implementation supports not defining all values (which is required for `partial version`), but we default the missing values to 0. This is not correct behavior, because some operators give special meaning to missing values, [like tilde (~)](https://github.com/npm/node-semver#tilde-ranges-123-12-1). So the missing values must stay "missing". 

To fix this, we should implement the `partial version` separately from `version`.

<details>
[...]

PR #3709, opened 2026-02-03
Fix semantic version module demo 1

FranjoMindek:
## Description

Fixes https://github.com/wasp-lang/wasp/issues/3699

TLDR: The issue is that we don't support optional minor and patch version. We can't just assume that missing values are "0", because some operators (~) have special behavior around missing values.

So the `Version` has to support not defining minor/patch values.

Approaches I've thought about:

### Just add `Maybe` approach (also the one I implemented)

```hs
data Version = Version
  { major :: !Natural,
    minor :: !(Maybe Natural),
    patch :: !(Maybe Natural)
  }
  deriving (TH.Lift)
```
Requires overriding default `Eq` and `Ord` implementation:
```hs
-- Also similar for 'Ord'
instance Eq Version where
  Version maj1 mnr1 ptc1 == Version maj2 mnr2 ptc2 =
    maj1 == maj2
      && fromMaybe 0 mnr1 == fromMaybe 0 mnr2
      && fromMaybe 0 ptc1 == fromMaybe 0 ptc2
```
Other than that there are no real changes to the old code, except the new syntax for `Version`, which is also the biggest downside:
```hs
prismaVersion :: SV.Version
prismaVersion = SV.Version 5 (Just 19) (Just 1)
```
To help with that we can use pattern synonyms:
```hs
{-# LANGUAGE PatternSynonyms #-}

-- | Pattern synonym for the common case where all three version components are present.
pattern FullVersion :: Natural -> Natural -> Natural -> Version
pattern FullVersion maj mnr ptc = Version maj (Just mnr) (Just ptc)
```
Which is then used as:
```hs
prismaVersion :: SV.Version
prismaVersion = SV.FullVersion 5 19 1
```
### Three constructors approach

```hs
data Version
  = MajorOnly !Natural
  | MajorMinor !Natural !Natural
  | MajorMinorPatch !Natural !Natural !Natural
  deriving (TH.Lift)
```
Seemed like it would bring too much complexity.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/src/Wasp/SemanticVersion.hs
+++ b/waspc/src/Wasp/SemanticVersion.hs
@@ -1,5 +1,8 @@
+{-# LANGUAGE PatternSynonyms #-}
+
 module Wasp.SemanticVersion
   ( Version (..),
+    pattern FullVersion,
     Range (..),
     ComparatorSet,
     isVersionInRange,
@@ -17,7 +20,7 @@ import Control.Monad (guard)
 import Data.List (intercalate, nub)
 import qualified Data.List.NonEmpty as NE
 import Data.Maybe (isJust)
-import Wasp.SemanticVersion.Version (Version (..), nextBreakingChangeVersion)
+import Wasp.SemanticVersion.Version (Version (..), nextBreakingChangeVersion, pattern FullVersion)
 import Wasp.SemanticVersion.VersionBound
   ( HasVersionBounds (versionBounds),
     VersionBound (Exclusive, Inclusive, Inf),
--- a/waspc/tests/SemanticVersionTest.hs
+++ b/waspc/tests/SemanticVersionTest.hs
@@ -10,21 +10,17 @@ spec_SemanticVersion :: Spec
 spec_SemanticVersion = do
   -- TODO: Add more tests to cover everything.
   describe "`show` produces valid semver strings" $ do
-    it "show Version" $ do
-      show (Version 0 0 0) `shouldBe` "0.0.0"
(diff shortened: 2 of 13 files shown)


waspc/tests/SemanticVersionTest.hs
```diff
@@ -10,21 +10,17 @@ spec_SemanticVersion :: Spec
 spec_SemanticVersion = do
   -- TODO: Add more tests to cover everything.
   describe "`show` produces valid semver strings" $ do
-    it "show Version" $ do
```
FranjoMindek, line 13: I removed the `it "show Version" $ do` block, because we already test `show Version` in `waspc/tests/SemanticVersion/VersionTest.hs`

waspc/src/Wasp/SemanticVersion.hs
```diff
@@ -1,5 +1,8 @@
+{-# LANGUAGE PatternSynonyms #-}
```
FranjoMindek, line 1: Note, if we use a qualified import we don't need to use this language extension. Which is what we already do in most cases (except here and in tests).

Importing everything also works without language extension:
`import Wasp.SemanticVersion.Version`

FranjoMindek:
We realized this is not the right approach.
See more info here: https://github.com/wasp-lang/wasp/issues/3699#issuecomment-3840006460

Or ask Franjo or Martin about more info.

Issue #3627, opened 2026-01-16
wasp-app-runner: Docker containers left running after hard process.exit()


cprecioso:
## Problem

The `wasp-app-runner` package frequently leaves Docker containers running after the process terminates, requiring manual cleanup with `docker rm -f <container-name>`.

## Root Cause

The codebase has multiple hard `process.exit()` calls throughout that prevent proper cleanup.

When `process.exit()` is called, the `ChildProcessManager` cleanup handlers may not have sufficient time to run, and importantly, **Docker container cleanup is not handled at all**. While containers are started with the `--rm` flag, this only works if the container exits gracefully.

PR #3712, opened 2026-02-03
Implement cache support for SDK assets copying

infomiho:
## Description

Copying of SDK assets (CSS, HTML files) happens on each SDK build which happens on each code change. This causes Vite to do a full-reload because now we are copying the virtual `index.html` in the SDK.

### Solution

If the source file is newer or the destination file doesn't exist - copy the source file.

### Context from Discord

Right now on _each_ user file change:
1. Wasp compiler regenerates files
    - Generator is "smart" by using checksums
2. Runs ⁨⁨⁨⁨⁨⁨⁨⁨⁨⁨⁨⁨`tsc`⁩⁩⁩⁩⁩⁩⁩⁩⁩⁩⁩⁩ to rebuild the SDK
    - smart with ⁨⁨⁨⁨⁨⁨⁨⁨⁨⁨`incremental`⁩⁩⁩⁩⁩⁩⁩⁩⁩⁩ compilation
3. Copies extra files for the SDK ⁨⁨⁨⁨⁨⁨⁨⁨⁨⁨⁨⁨`dist`⁩⁩⁩⁩⁩⁩⁩⁩⁩⁩⁩⁩ dir
    - just copies files each time

**This causes Vite dev server to detect these changes and reload the client app changes. The side-effect is sometimes full page reload which wipes the client state.**

:one: We prevented the ⁨⁨⁨⁨⁨⁨⁨⁨⁨⁨⁨⁨`tsc`⁩⁩⁩⁩⁩⁩⁩⁩⁩⁩⁩⁩ from doing too many changes by implementing ⁨⁨⁨⁨⁨⁨⁨⁨⁨⁨⁨⁨`incremental`⁩⁩⁩⁩⁩⁩⁩⁩⁩⁩⁩⁩ compilation: https://github.com/wasp-lang/wasp/pull/1931

:two: Since we introduced asset copying (for CSS, HTML files which ⁨⁨⁨⁨⁨⁨⁨⁨⁨⁨⁨⁨`tsc`⁩⁩⁩⁩⁩⁩⁩⁩⁩⁩⁩⁩ doesn't handle) in the SDK, this started happening again more aggressively. 

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/data/Generator/templates/sdk/wasp/scripts/copy-assets.js
+++ b/waspc/data/Generator/templates/sdk/wasp/scripts/copy-assets.js
@@ -10,14 +10,9 @@ await copyAssets([
 ])
 
 async function copyAssets(globs) {
-  try {
-    const sourceFiles = await findSourceFiles(globs)
-    const { copied, skipped } = await copyChangedFiles(sourceFiles)
-    console.log(`[copy-assets] ${copied} copied, ${skipped} skipped`)
-  } catch (error) {
-    console.error("[copy-assets] Fatal error:", error.message)
-    process.exit(1)
-  }
+  const sourceFiles = await Array.fromAsync(fs.glob(globs, { cwd: base }))
+  const { copied, skipped } = await copyChangedFiles(sourceFiles)
+  console.log(`[copy-assets] ${copied} copied, ${skipped} skipped`)
 }
 
 async function copyChangedFiles(sourceFiles) {
@@ -25,9 +20,11 @@ async function copyChangedFiles(sourceFiles) {
   let skipped = 0
 
   for (const file of sourceFiles) {
+    const sourcePath = getSourcePath(file)
+    const destPath = getDestinationPath(file)
     try {
-      if (await shouldCopyFile(file)) {
-        await copyFile(file)
+      if (await shouldCopyFile(sourcePath, destPath)) {
+        await copyFile(sourcePath, destPath)
         copied++
       } else {
         skipped++
@@ -45,11 +42,12 @@ async function copyChangedFiles(sourceFiles) {
  * File should be copied if:
  * - It doesn't exist in the destination directory.
  * - Its modification time is newer than the destination file.
+ *
+ * We want to avoid needless copying of files to avoid triggering
+ * Vite's file watcher which can cause full page reloads when not
+ * necessary.
  */
-async function shouldCopyFile(file) {
-  const sourcePath = getSourcePath(file)
-  const destPath = getDestinationPath(file)
-
+async function shouldCopyFile(sourcePath, destPath) {
   try {
     const [sourceStat, destStat] = await Promise.all([
       fs.stat(sourcePath),
@@ -62,10 +60,9 @@ async function shouldCopyFile(file) {
   }
 }
 
-async function copyFile(file) {
-  const dest = getDestinationPath(file)
-  await fs.mkdir(path.dirname(dest), { recursive: true })
-  await fs.copyFile(getSourcePath(file), dest)
+async function copyFile(sourcePath, destPath) {
+  await fs.mkdir(path.dirname(destPath), { recursive: true })
+  await fs.copyFile(sourcePath, destPath)
 }
 
 function getSourcePath(file) {
@@ -74,12 +71,4 @@ function getSourcePath(file) {
 
 function getDestinationPath(file) {
   return path.join(base, "dist", file)
-}
-
-async function findSourceFiles(globs) {
-  const files = []
-  for await (const file of fs.glob(globs, { cwd: base })) {
-    files.push(file)
-  }
-  return files
 }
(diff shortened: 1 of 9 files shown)


waspc/data/Generator/templates/sdk/wasp/scripts/copy-assets.js
```diff
@@ -1,18 +1,143 @@
...
+  }
+}
+
+async function copyFile(file) {
+  const dest = getDestinationPath(file)
+  await fs.mkdir(path.dirname(dest), { recursive: true })
+  await fs.copyFile(getSourcePath(file), dest)
+}
+
+async function removeOrphanedFiles(sourceFiles, cache) {
```
cprecioso: I think we don't need the "orphan files" system at all. On our local workstations, the file will still be there but won't get referenced, so no harm done. On our CI, the repo is checked out fresh every time, so there won't be any orphan files.

The situation where we delete or rename an asset file is exceedingly rare and has no ill side-effects, so I'd remove the handling for it.

waspc/data/Generator/templates/sdk/wasp/scripts/copy-assets.js
```diff
@@ -1,18 +1,85 @@
...
   "./client/vite/virtual-files/files/**/*.*",
-];
+])
 
-for await (const file of fs.glob(globs, { cwd: base })) {
-  const src = path.join(base, file);
-  const dest = path.join(base, "dist", file);
+async function copyAssets(globs) {
+  try {
+    const sourceFiles = await findSourceFiles(globs)
```
cprecioso:
```suggestion
    const files = await Array.fromAsync(fs.glob(globs, { cwd: base }))
```

and then you can remove `findSourceFiles`

waspc/data/Generator/templates/sdk/wasp/scripts/copy-assets.js
```diff
@@ -1,18 +1,85 @@
...
+}
+
+/**
+ * File should be copied if:
+ * - It doesn't exist in the destination directory.
+ * - Its modification time is newer than the destination file.
+ */
+async function shouldCopyFile(file) {
+  const sourcePath = getSourcePath(file)
+  const destPath = getDestinationPath(file)
```
cprecioso: optional, style: you can probably bump these two `const`s to the `copyChangedFiles` function and pass those down.

waspc/data/Generator/templates/sdk/wasp/scripts/copy-assets.js
```diff
@@ -1,18 +1,85 @@
...
+  await fs.copyFile(getSourcePath(file), dest)
+}
+
+function getSourcePath(file) {
+  return path.join(base, file)
+}
+
+function getDestinationPath(file) {
+  return path.join(base, "dist", file)
+}
```
cprecioso, line 74: minor, style: i know you like this kind of functions so RAW; but I'd much rather just calculate these paths inline, in the loop, and just pass those variables down to where necessary

waspc/data/Generator/templates/sdk/wasp/scripts/copy-assets.js
```diff
@@ -1,18 +1,85 @@
...
-  const src = path.join(base, file);
-  const dest = path.join(base, "dist", file);
+async function copyAssets(globs) {
+  try {
+    const sourceFiles = await findSourceFiles(globs)
+    const { copied, skipped } = await copyChangedFiles(sourceFiles)
+    console.log(`[copy-assets] ${copied} copied, ${skipped} skipped`)
+  } catch (error) {
+    console.error("[copy-assets] Fatal error:", error.message)
+    process.exit(1)
```
cprecioso: I'd like for us to **stop using `process.exit()` completely**, it [gives us problems](https://github.com/wasp-lang/wasp/issues/3627) because it doesn't exit cleanly. You don't need it here because the function will return after the `catch` anyway.
If you want to set the exit code, you can either let the error bubble up to the top level (we don't really need the "Fatal error" message and it hides the stack trace), or set `process.exitCode = 1`.

cprecioso (approved):
lgtm, some suggestions. please remove the `process.exit` before merging.

waspc/data/Generator/templates/sdk/wasp/scripts/copy-assets.js
infomiho: Did not know about `Array.fromAsync` 😄 This is some Haskell level helper right here!

waspc/data/Generator/templates/sdk/wasp/scripts/copy-assets.js
infomiho: I was just waiting for this comment 😄 I almost edited the code, but I wanted to see if I was predicting the future correctly 😂

waspc/data/Generator/templates/sdk/wasp/scripts/copy-assets.js
infomiho: I'll keep them after all - I like the clarity of defining how source and destination paths are calculated.

PR #3715, opened 2026-02-04
Franjo/semantic version refactor test

FranjoMindek:
## Description

There is distinction between:
- SemVer version: https://semver.org/
- node-semver "partial versions": https://github.com/npm/node-semver

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/waspc.cabal
+++ b/waspc/waspc.cabal
@@ -451,10 +451,10 @@ library
     Wasp.Psl.Util
     Wasp.Psl.Valid
     Wasp.SemanticVersion
-    Wasp.SemanticVersion.Parsers
-    Wasp.SemanticVersion.PartialVersion
     Wasp.SemanticVersion.Comparator
     Wasp.SemanticVersion.ComparatorSet
+    Wasp.SemanticVersion.Parsers
+    Wasp.SemanticVersion.PartialVersion
     Wasp.SemanticVersion.Range
     Wasp.SemanticVersion.Version
     Wasp.SemanticVersion.VersionBound


waspc/src/Wasp/SemanticVersion/Comparator.hs
```diff
@@ -0,0 +1,172 @@
...
+    VersionBound (..),
+  )
+
+-- | A comparator is composed of an operator and a partial version.
+-- It represents a single version constraint in a range.
+-- See: https://github.com/npm/node-semver#ranges
+data Comparator
+  = PrimitiveComparator PrimitiveOperator PartialVersion
+  | BackwardsCompatibleWith PartialVersion
+  | ApproximatelyEquvivalentTo PartialVersion
```
FranjoMindek: **Typo: `ApproximatelyEquvivalentTo` → `ApproximatelyEquivalentTo`**

_Review by Claude (Opus 4.6)_

"Equvivalent" is misspelled — it should be "Equivalent" (missing the `i` after `u`). This typo propagates through the entire codebase: the constructor here, the helper function `approximatelyEquvivalentTo` in `ComparatorSet.hs`, and all test references.

Since this is a public constructor (re-exported via `Wasp.SemanticVersion`), fixing it later would be a more disruptive rename. Better to fix it now while the PR is in flight.

```haskell
-- Current:
| ApproximatelyEquvivalentTo PartialVersion

-- Suggested:
| ApproximatelyEquivalentTo PartialVersion
```

waspc/src/Wasp/SemanticVersion/Comparator.hs
```diff
@@ -0,0 +1,172 @@
...
+-- See `hyphen` definition here: https://github.com/npm/node-semver#range-grammar
+hyphenRangeComparatorParser :: Parsec String () Comparator
+hyphenRangeComparatorParser = do
+  lowerVersion <- partialVersionParser
+  _ <- hyphenParser
+  upperVersion <- partialVersionParser
+  pure $ HyphenRange lowerVersion upperVersion
+  where
+    hyphenParser :: Parsec String () Char
+    hyphenParser = P.spaces *> P.char '-' <* P.spaces
```
FranjoMindek: **Bug: `hyphenParser` does not require spaces around the hyphen**

_Review by Claude (Opus 4.6)_

`P.spaces` matches **zero or more** spaces. This means `"1.2.3-4.5.6"` (no spaces around the hyphen) would be falsely parsed as `HyphenRange (Full 1 2 3) (Full 4 5 6)`.

Per the [node-semver grammar](https://github.com/npm/node-semver#range-grammar), hyphen ranges require the form `<partial> <space> - <space> <partial>`. The hyphen must be surrounded by at least one space on each side, otherwise it is ambiguous with pre-release identifiers (e.g. `1.2.3-alpha` or `1.2.3-4.5.6`).

This matters because:
1. Once pre-release support is added (there is a TODO for it), `1.2.3-4` should parse as version `1.2.3` with pre-release `4`, not as a hyphen range.
2. Even now, `parseRange "1.2.3-4.5.6"` would incorrectly succeed as a hyphen range instead of failing or being treated as a single version with trailing content.

Suggested fix — require at least one space on each side:
```haskell
hyphenParser :: Parsec String () Char
hyphenParser = P.many1 P.space *> P.char '-' <* P.many1 P.space
```

waspc/src/Wasp/SemanticVersion/Comparator.hs
```diff
@@ -0,0 +1,172 @@
...
+        toLessThanUpperBound :: PartialVersion -> VersionBound
+        toLessThanUpperBound Any = Exclusive $ Version 0 0 0
+        toLessThanUpperBound (Major mjr) = Exclusive $ Version mjr 0 0
+        toLessThanUpperBound (MajorMinor mjr mnr) = Exclusive $ Version mjr mnr 0
+        toLessThanUpperBound (Full mjr mnr ptc) = Exclusive $ Version mjr mnr ptc
+    LessThanOrEqual -> (Inf, toXRangeUpperBound pv)
+    GreaterThan -> (toGreaterThanLowerBound pv, Inf)
+      where
+        toGreaterThanLowerBound :: PartialVersion -> VersionBound
+        toGreaterThanLowerBound Any = Inf
```
FranjoMindek: **Edge case: `>*` (GreaterThan Any) incorrectly matches all versions**

_Review by Claude (Opus 4.6)_

`toGreaterThanLowerBound Any = Inf` produces the interval `(Inf, Inf)` which represents "all versions" in the `VersionBound` system. However, per node-semver, `>*` should match **no** versions (there is no version greater than the universal set).

For comparison:
- `<*` → `(Inf, Exclusive 0.0.0)` → nothing ✓
- `<=*` → `(Inf, Inf)` → everything ✓
- `>=*` → `(Inclusive 0.0.0, Inf)` → everything ✓
- `=*` → `(Inclusive 0.0.0, Inf)` → everything ✓
- `>*` → `(Inf, Inf)` → everything ✗ (should be nothing)

This is admittedly an obscure edge case — nobody writes `>*` in practice — but it is a correctness issue. The cleanest fix might be to produce an empty interval where lower > upper, e.g.:
```haskell
toGreaterThanLowerBound Any = Exclusive (Version 0 0 0)
-- Combined with upper bound Inf... that's still wrong.
```

Actually the simplest approach: special-case `>*` at the top level:
```haskell
versionBounds (PrimitiveComparator GreaterThan Any) = 
  (Inclusive (Version 0 0 0), Exclusive (Version 0 0 0))  -- empty interval
```

Or just document it as a known limitation since `>*` is nonsensical.

waspc/src/Wasp/SemanticVersion/ComparatorSet.hs
```diff
@@ -0,0 +1,108 @@
...
+-- | We define concatenation of two comparator sets as a union of their comparators.
+instance Semigroup ComparatorSet where
+  (ComparatorSet leftComps) <> (ComparatorSet rightComps) = ComparatorSet $ NE.nub $ leftComps <> rightComps
+
+instance HasVersionBounds ComparatorSet where
+  versionBounds (ComparatorSet comps) = foldr1 intervalIntersection $ versionBounds <$> comps
+
+-- Helper methods for constructing a 'ComparatorSet'.
+-- While 'Comparator' works with 'PartialVersion' internally, we only use it through 'Version' in our code.
+-- To create 'PartialVersion' comparator sets, pelease use the 'ComparatorSet' constructor directly.
```
FranjoMindek: **Typo: `pelease` → `please`**

_Review by Claude (Opus 4.6)_

Minor typo in the comment.

waspc/src/Wasp/SemanticVersion.hs
```diff
@@ -1,146 +1,16 @@
...
-    ComparatorSet,
-    isVersionInRange,
-    lt,
-    lte,
-    gt,
-    gte,
-    eq,
-    backwardsCompatibleWith,
-    doesVersionRangeAllowMajorChanges,
+  ( module Wasp.SemanticVersion.Comparator,
```
FranjoMindek, line 2: **Blanket re-exports expose internal implementation details**

_Review by Claude (Opus 4.6)_

Using `module Wasp.SemanticVersion.Comparator` etc. re-exports **everything** from each submodule, including parser internals like `simpleComparatorParser`, `hyphenRangeComparatorParser`, `partialVersionParser`, `naturalNumberParser`, `comparatorSetParser`, and `PrimitiveOperator` constructors.

The old module was selective about its public API (just `Version`, `Range`, `ComparatorSet`, `isVersionInRange`, helper constructors like `lt`/`gt`/`eq`, etc.).

Consumers who do `import qualified Wasp.SemanticVersion as SV` now have access to all these internals. While the qualified import mitigates namespace pollution, it still makes it unclear what the intended public API is.

Consider being explicit about what gets re-exported:
```haskell
module Wasp.SemanticVersion
  ( -- * Types
    Version (..),
    PartialVersion (..),
    Comparator (..),
    PrimitiveOperator (..),
    ComparatorSet (..),
    Range (..),
    -- * Range operations
    isVersionInRange,
    doesVersionRangeAllowMajorChanges,
    -- * ComparatorSet helpers
    lt, lte, gt, gte, eq,
    backwardsCompatibleWith,
    -- * Parsing
    parseRange,
    parseVersion,
    parsePartialVersion,
    -- * Version bounds
    HasVersionBounds (..),
    VersionBound (..),
    ...
  )
```

This makes the public API explicit and keeps parser internals as implementation details of their respective modules.

waspc/tests/SemanticVersion/VersionTest.hs
```diff
@@ -1,30 +1,78 @@
 module SemanticVersion.VersionTest where
 
-import Data.Either (isLeft)
+import Data.Either (isLeft, isRight)
```
FranjoMindek, line 3: **Two typos in this file**

_Review by Claude (Opus 4.6)_

1. Line 3: `versiosn` → `versions` (in the comment)
2. Line 18 (inside the `describe` block): `prases` → `parses`

waspc/src/Wasp/SemanticVersion/Version.hs
```diff
@@ -2,46 +2,53 @@
...
 instance Show Version where
   show (Version mjr mnr ptc) = printf "%d.%d.%d" mjr mnr ptc
 
 parseVersion :: String -> Either ParseError Version
 parseVersion = parse versionParser ""
 
+-- | Follows SemVer specification.
+-- See: https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions
+-- TODO: Add pre-release (-) and build (+) support.
 versionParser :: Parsec String () Version
```
FranjoMindek, line 39: **Behavior change: `versionParser` no longer accepts partial versions**

_Review by Claude (Opus 4.6)_

The old `versionParser` accepted partial versions:
- `"1"` → `Version 1 0 0`
- `"1.2"` → `Version 1 2 0`

The new parser requires all three components, which is correct per the SemVer spec (a valid SemVer version always has `MAJOR.MINOR.PATCH`).

This is the right change semantically, but it is a **behavior change** for callers. In particular, `Wasp.Node.Internal.parseVersionFromCommandOutput` uses `SV.versionParser`. If any code path ever produced a version string with only 1 or 2 components, it would now fail.

In practice this is likely fine since `node --version` returns full `vX.Y.Z` strings, but worth verifying that no other callsite depends on the old lenient behavior. Also, the `v` quasi-quoter now rejects `[v|1|]` and `[v|0.2|]` which were previously valid — if any compile-time usage exists elsewhere in the codebase with partial versions, the build would break.

(The old tests that verified these have been correctly updated, so this is just a heads-up to check for other usages.)

waspc/src/Wasp/SemanticVersion/Parsers.hs
```diff
@@ -0,0 +1,18 @@
...
+  )
+where
+
+import Control.Applicative ((<|>))
+import Numeric.Natural (Natural)
+import Text.Parsec (Parsec, digit, many, notFollowedBy, oneOf, string, try)
+
+naturalNumberParser :: Parsec String () Natural
+naturalNumberParser = do
+  read <$> (zeroParser <|> noLeadingZeroNaturalNumberParser)
```
FranjoMindek, line 12: **Nit: redundant `do`**

_Review by Claude (Opus 4.6)_

The `do` on line 11 is unnecessary since the body is a single expression:

```haskell
-- Current:
naturalNumberParser = do
  read <$> (zeroParser <|> noLeadingZeroNaturalNumberParser)

-- Could be:
naturalNumberParser =
  read <$> (zeroParser <|> noLeadingZeroNaturalNumberParser)
```

Very minor, not a blocker at all.

PR #3721, opened 2026-02-04
Check invalid declaration names in ts spec

### LINKED ISSUE #2916: Improve Invalid Route Name Handling
**Describe the bug**
Naming a route with a space does not throw a Wasp error. Rather, it causes a syntax error in the `.wasp/out/sdk/wasp/client/router/index.ts` file: `error TS1005: ',' expected.` It's not hard to figure out why once you visit the file:
```
  Home Page: {
    to: "/",
    build: (
      options?:
      OptionalRouteOptions
    ) => interpolatePath(
        
        "/",
        undefined,
        options?.search,
        options?.hash
      ),
  },
```

Wasp should throw an error when attempting to add a route with a space rather than let it continue and fail silently.

**To Reproduce**

1. Create a route with a space in its name.
2. Run `wasp start` and wait for the SDK to attempt to build.
3. See TS error.

**Expected behavior**
A Wasp-generated error explaining there cannot be spaces in a route name.

sodic:
## Description

Fixes #2916 and then some.

Here's how it looks like:

image

## Type of change

- [x] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/e2e-tests/SnapshotTest.hs
+++ b/waspc/e2e-tests/SnapshotTest.hs
@@ -9,6 +9,7 @@ module SnapshotTest
   )
 where
 
+import Control.Exception (bracket)
 import Control.Monad (filterM)
 import qualified Data.Aeson as Aeson
 import qualified Data.Aeson.Encode.Pretty as AesonPretty
@@ -30,8 +31,9 @@ import StrongPath (Abs, Dir, File, Path', parseRelDir, (</>))
 import qualified StrongPath as SP
 import System.Directory (doesFileExist)
 import System.Directory.Recursive (getDirFiltered)
+import System.Exit (ExitCode (..))
 import System.FilePath (equalFilePath, isExtensionOf, makeRelative, takeFileName)
-import System.Process (callCommand)
+import System.Process (CreateProcess (..), callCommand, createProcess, interruptProcessGroupOf, shell, waitForProcess)
 import Test.Tasty (TestTree, testGroup)
 import Test.Tasty.Golden (goldenVsFileDiff)
 
@@ -104,7 +106,7 @@ executeSnapshotTestCommand :: SnapshotTest -> Path' Abs (Dir SnapshotDir) -> IO 
 executeSnapshotTestCommand snapshotTest snapshotDir = do
   putStrLn $ "Executing snapshot test: " ++ snapshotTest.name
   putStrLn $ "Running the following command: " ++ snapshotTestCommand
-  callCommand $ "cd " ++ SP.fromAbsDir snapshotDir ~&& snapshotTestCommand
+  callCommandInProcessGroup $ "cd " ++ SP.fromAbsDir snapshotDir ~&& snapshotTestCommand
   where
     snapshotTestCommand :: ShellCommand
     snapshotTestCommand = foldr1 (~&&) $ buildShellCommand snapshotTestContext snapshotTest.shellCommandBuilder
@@ -224,3 +226,18 @@ type FileName = String
 
 createFilenameFilter :: [FileName -> Bool] -> FilePath -> Bool
 createFilenameFilter predicates filePath = all ($ takeFileName filePath) predicates
+
+-- | Interruptible version of callCommand that terminates the entire process tree on async exception.
+-- Uses process groups so that when a thread is cancelled (e.g., when another concurrent test fails),
+-- all child processes are also terminated rather than continuing to run.
+callCommandInProcessGroup :: String -> IO ()
+callCommandInProcessGroup cmd =
+  bracket
+    (createProcess (shell cmd) {create_group = True})
+    (\(_, _, _, ph) -> interruptProcessGroupOf ph)
+    ( \(_, _, _, ph) -> do
+        exitCode <- waitForProcess ph
+        case exitCode of
+          ExitSuccess -> return ()
+          ExitFailure code -> fail $ "Command failed with exit code " ++ show code ++ ": " ++ cmd
+    )


waspc/tests/AppSpec/ValidTest.hs
```diff
@@ -453,6 +454,47 @@ spec_AppSpecValid = do
...
+        testInvalidDeclName makeBasicPageDecl "Import"
+        testInvalidDeclName makeBasicJobDecl "1st job"
+        testInvalidDeclName makeBasicQueryDecl "my-query"
+        testInvalidDeclName makeBasicActionDecl "my action"
+
+      describe "returns no error for valid identifiers" $ do
+        testValidDeclName makeBasicPageDecl "Import"
+        testValidDeclName makeBasicJobDecl "_1stJob"
+        testValidDeclName makeBasicQueryDecl "my_query"
+        testValidDeclName makeBasicActionDecl "myAction"
```
sodic, line 481: These are the tests related to the new logic. The other tests are for old code and I added them in passing.

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -335,6 +337,42 @@ validateDeclarationNames spec =
...
+        - This is where we keep similar app spec validations.
+        - It reuses the actual lexer instead of duplicating its rules in
+          TypeScript (and potential future spec variants)
+      -}
+      let invalidDeclNames = filter (not . isValidWaspIdentifier) $ map getDeclName $ AS.decls spec
+          declNameRules =
+            [ "must start with a letter or an underscore",
+              "must contain only letters, numbers, or underscores",
+              "must not be a Wasp keyword"
+            ]
```
sodic, line 366: There's some apparent duplication here and in `ProjectDescription#parseWaspProjectNameIntoAppName`.

I tried to find a way to remove it but it was too fiddly, the main reason being that the `parseWaspProjectNameIntoAppName` additionally accepts dashes. Weaving that extra bit of information into the shared message while satisfying both formats was more trouble than it was worth, so I left it as is.

waspc/tests/AppSpec/ValidTest.hs
```diff
@@ -453,6 +454,47 @@ spec_AppSpecValid = do
           `shouldBe` [ Valid.GenericValidationError
                          "You must have at least one route in your app. You can add it using the 'route' declaration."
                      ]
+
+    describe "invalid declaration names validation" $ do
+      let testInvalidDeclName makeDecl invalidName = it invalidName $ do
+            let decl = makeDecl invalidName
+            let errors = ASV.validateAppSpec (basicAppSpec {AS.decls = [basicAppDecl, decl, basicRouteDecl]})
+            case errors of
+              [err] -> invalidName `isInfixOf` show err `shouldBe` True
```
sodic, line 463: I didn't want to match the entire string because it seemed to brittle.

waspc/src/Wasp/AppSpec/Valid.hs
FranjoMindek: I would say those two are 2 different cases then, and should be treated so.

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -335,6 +337,42 @@ validateDeclarationNames spec =
...
+       in case invalidDeclNames of
+            [] -> []
+            _ ->
+              [ GenericValidationError $
+                  intercalate "\n" $
+                    ("Please rename: " ++ intercalate ", " invalidDeclNames ++ ". Each declaration name:")
```
FranjoMindek: Only thing I would so is to make `parseWaspProjectNameIntoAppName` error message and this one be more similar. To be consistent with out messaging.

So pick one from `it can` and `must`, etc.

waspc/ChangeLog.md
```diff
@@ -29,6 +29,7 @@ Remember to check out the [migration guide](https://wasp.sh/docs/migration-guide
 - Fixed an issue where if Wasp was installed from npm, `wasp new` would not work correctly. ([#3585](https://github.com/wasp-lang/wasp/pull/3585))
 - Fixed an issue where the `Link` component might not update with the values of `search` and `hash` (by @Pipboyguy). ([#3633](https://github.com/wasp-lang/wasp/pull/3633))
 - We now support Railway CLI v4.18.1 and greater on `wasp deploy railway`. ([#3667](https://github.com/wasp-lang/wasp/pull/3667))
+- Wasp TS spec now validates declaration names ([#3721](https://github.com/wasp-lang/wasp/pull/3721))
```
FranjoMindek:
```suggestion
- Wasp TS spec now validates declaration names. ([#3721](https://github.com/wasp-lang/wasp/pull/3721))
```

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -297,7 +298,8 @@ validateUniqueDeclarationNames spec =
 validateDeclarationNames :: AppSpec -> [ValidationError]
 validateDeclarationNames spec =
   concat
-    [ capitalizedOperationsErrorMessage,
+    [ invalidDeclNamesErrorMessage,
```
FranjoMindek: Okay I got a bit confused with the names down in the tests because of this `invalidDeclNamesErrorMessage` name.

We have following defined lines in our tests:
- the tests section - `invalid declaration names validation`
- helper functions - `testInvalidDeclName`, `testValidDeclName`


Makes it seem like tests test only this specific scenario, but it is really not. It's testing the whole `validateDeclarationNames`.  Let's clear this up a bit.

`invalid declaration names validation` -> maybe `declaration names validation` or `validateDeclarationNames` if it seems like we validating that specific function?

`invalidDeclNamesErrorMessage` seems fine. I don't know if we can change it really.

PR #3722, opened 2026-02-04
Fix e2e snapshot tests parallelization interruptions

FranjoMindek:
## Problem

Currently, when a snapshot test process errors and exists early, that error propagates and also interrupts the main process (`cabal test e2e-tests`), but other snapshot test shell command processes continue working orphapned.

image
image

So the workflow is:

In the failing test:
1. wasp fails → shell exits → `callCommand` throws exception (it internally checks exit code)
2. Exception propagates up to `mapConcurrently`

In sibling tests:
1. `mapConcurrently` throws `AsyncCancelled` to them
2. `callCommand` is blocked in `waitForProcess` internally
3. `waitForProcess` is interrupted by `AsyncCancelled`
4. No cleanup
5. `AsyncCancelled` propagates, threads die
6. Shell + wasp processes are still running (nobody sent them a signal)

Then:
1. `mapConcurrently` propagates the original exception from the failing test
2. It reaches `main`
3. Haskell process exits
4. Shell + wasp processes are now orphans (re-parented to PID 1)
5. They keep running and writing to the terminal

## Solution

So what we want to do is for all processes started by a snapshot test process to die together with it.

My solution is snapshot test process run the commands as separate process groups, that cleanup on interruptions:
```hs
-- | Interruptible version of callCommand that terminates the entire process tree on async exception.
-- Uses process groups so that when a thread is cancelled (e.g., when another concurrent test fails),
-- all child processes are also terminated rather than continuing to run.
callCommandInProcessGroup :: String -> IO ()
callCommandInProcessGroup cmd =
  bracket
    (createProcess (shell cmd) {create_group = True})
    (\(_, _, _, ph) -> interruptProcessGroupOf ph)
    ( \(_, _, _, ph) -> do
        exitCode <- waitForProcess ph
        case exitCode of
          ExitSuccess -> return ()
          ExitFailure code -> fail $ "Command failed with exit code " ++ show code ++ ": " ++ cmd
    )
```

So the workflow is:

In the failing test:
1. wasp fails → shell exits → `waitForProcess` returns `ExitFailure`
2. We throw via `fail` - "Command failed..."
3. `bracket` cleanup runs (but process already exited, so it's a no-op)
4. Exception propagates up to `mapConcurrently`

In sibling tests:
1. `mapConcurrently` throws `AsyncCancelled` to them
2. `waitForProcess` is interrupted
3. `bracket` cleanup runs → `interruptProcessGroupOf` kills the running processes
4. `bracket` re-throws `AsyncCancelled`
5. Threads die

Then:
1. `mapConcurrently` propagates the original exception from the failing test
2. It reaches `main`
3. Haskell process exits

## Type of change

- [x] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/e2e-tests/SnapshotTest.hs
+++ b/waspc/e2e-tests/SnapshotTest.hs
@@ -33,7 +33,7 @@ import System.Directory (doesFileExist)
 import System.Directory.Recursive (getDirFiltered)
 import System.Exit (ExitCode (..))
 import System.FilePath (equalFilePath, isExtensionOf, makeRelative, takeFileName)
-import System.Process (CreateProcess (..), createProcess, interruptProcessGroupOf, shell, waitForProcess)
+import System.Process (CreateProcess (..), callCommand, createProcess, interruptProcessGroupOf, shell, waitForProcess)
 import Test.Tasty (TestTree, testGroup)
 import Test.Tasty.Golden (goldenVsFileDiff)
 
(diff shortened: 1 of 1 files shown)


waspc/e2e-tests/SnapshotTest.hs
FranjoMindek, line 100: I really don'  think we need to change these to `callCommandInProcessGroup`.
They are short lasting, and will most likely finish before any interruptions can happen.

It's mostly the snapshot test bash commands that can take a while to exit.

Feel free to comment here if you want me to change these too.

waspc/e2e-tests/SnapshotTest.hs
```diff
@@ -224,3 +226,18 @@ type FileName = String
 
 createFilenameFilter :: [FileName -> Bool] -> FilePath -> Bool
 createFilenameFilter predicates filePath = all ($ takeFileName filePath) predicates
+
+-- | Interruptible version of callCommand that terminates the entire process tree on async exception.
+-- Uses process groups so that when a thread is cancelled (e.g., when another concurrent test fails),
+-- all child processes are also terminated rather than continuing to run.
+callCommandInProcessGroup :: String -> IO ()
+callCommandInProcessGroup cmd =
+  bracket
```
cprecioso, line 235: TIL!

waspc/e2e-tests/SnapshotTest.hs
cprecioso: I'll follow your lead

cprecioso (approved):
Great, thanks!

PR #3723, opened 2026-02-04
Docs: Rename versions

### LINKED ISSUE #3716: Docs: Tweak our docs versioning convention for AIs
> Prompted by @Pipboyguy at https://github.com/wasp-lang/wasp/pull/3708

We version our docs everywhere by using full versions: `0.20.0`, `0.19.0`, `0.18.0`, etc.
However, LLMs might see that the installed version is a hotfix e.g. `0.20.1`, and trip up trying to search for that exact version, not knowing that the `0.20.0` docs are good for any `0.20.*` release.

I think we should:
- [X] Add a note to `/llms.txt` saying that the `0.X.0` docs are good for any `0.X.*` version (easy)
- [x] Tweak our versioning convention so that it turns into `0.X` instead of `0.X.0`

---

For reference, these version numbers show up (user-facing):
- In the docs navbar, in the version selector
- In a versioned doc, in the URL
- In https://wasp.sh/llms.txt in the list of links to each version
- In the name of each versioned file (e.g. https://wasp.sh/llms-0.19.0.txt)

cprecioso:
Renames versioned documentation folders from `version-0.X.0` to `version-0.X` format and adds Cloudflare Pages redirects to ensure old URLs continue to work.

Resolves #3716

## Type of change

- [ ] **🐞 Bug fix**
- [ ] **💥 Breaking change**

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -102,6 +102,8 @@ When releasing new version of Wasp, what we do is run `npm run docusaurus docs:v
 
 The version should not be a full version (e.g. `0.21.0`), but valid for all the non-breaking changes in that range (e.g. `0.21`).
 
+Note: This means users can't access docs for a specific patch version (e.g., `0.21.2`) - they always see the docs for the latest patch in that minor version (e.g., `0.21.4`). This could lead to users reading about features not yet available in their exact version. We consider this an acceptable trade-off for now, since upgrading within the same minor version has no breaking changes. We may revisit this approach in the future if needed.
+
 This command does everything for us, and since we use Docusaurus' default settings for versions,
 there is nothing else we need to do, it will be picked up as the latest version by default.
 


web/versions.json
```diff
@@ -1,12 +1,12 @@
...
+  "0.20",
+  "0.19",
+  "0.18",
+  "0.17",
+  "0.16",
+  "0.15",
+  "0.14",
+  "0.13",
+  "0.12",
   "0.11.8"
```
cprecioso, line 11: I left this one alone as it seems to be specific.

web/scripts/generate-llm-files.ts
```diff
@@ -135,7 +135,7 @@ async function generateFiles() {
 
 function buildDocsMapsByVersionSection(versions: string[]): string {
   const latestVersion = versions[0];
-  let section = `## Documentation Maps by Version\n*IMPORTANT:* You should run \`wasp version\` to get the installed Wasp CLI version before choosing the correct link. The 0.X.0 documentation is good for any 0.X.* version.\n`;
```
cprecioso, line 138: This note was introduced as a bandaid to fix the issue but it's not needed anymore

web/static/_redirects
```diff
@@ -20,3 +20,14 @@
 /docs/migrate-from-0-13-to-0-14    /docs/migration-guides/migrate-from-0-13-to-0-14    301
 
 /docs/deployment/deployment-methods/cli /docs/deployment/deployment-methods/wasp-deploy/overview 301
+
+# Version path redirects (0.X.0 -> 0.X)
```
Martinsos: Maybe mention that we addedt his because we previously had these old ones -> wouldn't want somebody to keep adding these (be it human or AI) for every new docs version because they think it is a thing we do :D (which one could easil conclude looking at this without this additioanl historical context).

Btw nice job thinking of this.

web/README.md
```diff
@@ -100,6 +100,8 @@ Since we don't want our users to read `docs/next` ("current" docs), we don't pub
 
 When releasing new version of Wasp, what we do is run `npm run docusaurus docs:version {version}` to create new versioned docs from the current docs. We do this on every new Wasp release.
 
+The version should not be a full version (e.g. `0.21.0`), but valid for all the non-breaking changes in that range (e.g. `0.21`).
```
Martinsos, line 103: And this does mean that users can't get exact docs for let's say 0.21.2, but instead they are reading those for 0.21.4 (latest with that major version). So they might by accident use a feature that is not supported in 0.21.2.
But I think this is trade off we are ok with? Especially since they can easily upgrade it since there is no breaking change, right?

I mean I guess in the ideal world, we would have docs for every single minor version also. SHould we even consider that at this point though?

One thing I could see us doing right now is labeling docs with minor version also, let's say 0.21.2. Maybe that is better, at least we are precise? And having alias 0.21 that points to it? WOuld that be a better system?

Hm I wonder what others do.

Or we can decide to do something about it later when we are smarter.
In which case, I would at least capture here, in this README.md, a bit of info about this trade off -> yes, we know this means they can't get to the specific minor version docs, and they are also not sure for which latest minor version the docs are, but it is good enough for now. Because I would be asking myself these questions if I was reading this line, so let's answer them immmediately?

RAW

PR #3771, opened 2026-02-13
Update Vite config plugin to force certain options

### LINKED ISSUE #3635: Force certain Vite options
The way we currently merge user options and Wasp options for Vite makes is possible for users to override certain options that Wasp needs to control. Examples are:
- `envPrefix` needs to be of certain value because rest of the Wasp framework sets these env variables
- `build.outDir` needs to be of certain value because our deployment package expects the client to be built in a certain folder
- `server.port` needs to be of certain value because our server expect to see the client on a certain port 
- etc.

Based on https://github.com/wasp-lang/wasp/pull/3604#discussion_r2704596020

infomiho:
## Description

This PR introduces validation for 3 options we don't want users to change in their `vite.config.ts`:

**1. These 2 options I'm sure we should force:**
- `base`
  - user can change this via `client.baseDir` Wasp file option (this changes both Vite options and React Router options so we need to control this option) 
- `build.outDir`
  - our `deploy` package expects the value to be `.wasp/out/web-app/build` so users can't change it

**2. These options might be okay for user to change:**
- `envPrefix`
  - if they change the `envPrefix` to something else, our docs become irrelevant e.g. everywhere where we say "define REACT_APP_API_URL as x" becomes incorrect. 
  - Users can change it, but then they are on their own. 
  - Since this is related to security (setting the prefix ensures only prefixed options are included in the browser build) I'm more inclined to forbid changing it. 

Closes #3635 

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:


waspc/e2e-tests/Tests/ViteOptionsValidationTest.hs
```diff
@@ -0,0 +1,52 @@
...
+viteOptionsValidationTest =
+  Test
+    "vite-options-validation"
+    [ TestCase
+        "fail-overriding-base-option"
+        ( sequence
+            [ createTestWaspProject Minimal,
+              inTestWaspProjectDir
+                [ writeViteConfigWithForcedOption,
+                  expectCommandFailure <$> waspCliStart
```
infomiho: Even though I expected the command to fail, the e2e tests fail on this test. Am I doing something wrong?

cprecioso: I commented `expectCommandFailure` helper to @FranjoMindek and he told me it was not possible because of some reason I forget. Maybe he knows?

web/docs/project/custom-vite-config.md
```diff
@@ -46,6 +46,17 @@ Here's the minimal required configuration:
 The `wasp()` plugin must be the **first** plugin in the `plugins` array. Any other plugins (like Tailwind CSS) should be added after it.
 :::
 
+## Enforced Options
+
+The `wasp()` plugin enforces certain Vite config values that Wasp needs to function correctly. If you set any of these in your `vite.config.ts`, Wasp will throw an error asking you to remove them.
+
+| Option | Value | Why it's needed |
```
cprecioso: It looks like we're asking _the user_ to add this options to the config. I'd rephrase a bit to note that they don't need to add these.


```suggestion
| Option | Internal value | Why you can't customize it |
```

waspc/ChangeLog.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.21.2
```
cprecioso, line 3: So, _strictly speaking_ this is a breaking change, since there was something users could explicitly do before (set a different `port`), that they can't do now. But yeah, it would suck to do a breaking change just for this dumb thing.

Another option is to let people set the `port`, so there wouldn't be any breaking change in documented functionality.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/waspConfig.ts
```diff
@@ -1,25 +1,46 @@
...
 export function waspConfig(): PluginOption {
   return {
     name: "wasp:config",
     enforce: 'pre',
     config(config) {
-      return mergeConfig({
-        base: "{= baseDir =}",
+      throwIfOverridingForcedOptions(config);
+
+      return {
```
cprecioso, line 32: let's add a note that when we `return` from this function, that object gets `mergeConfig`'d to the user config.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/waspConfig.ts
```diff
@@ -42,15 +63,44 @@ export function waspConfig(): PluginOption {
...
         },
-      }, config);
+      };
     }
   };
 }
+
+function useUserValue<T>(userValue: T | undefined, defaultValue: T): T {
+  return userValue ?? defaultValue;
+}
```
cprecioso, line 82: style: I dislike this helper, it's just an alias for `??`, which is plenty descriptive for me. The function is not enforcing that the `userValue` does in fact come from the user, so it doesn't feel needed at all. RAW

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/waspConfig.ts
```diff
@@ -42,15 +63,44 @@ export function waspConfig(): PluginOption {
           ],
         },
         test: {
-          globals: true,
-          environment: "jsdom",
+          globals: useUserValue((config as any).test?.globals, true),
+          environment: useUserValue((config as any).test?.environment, "jsdom"),
```
cprecioso: Let's import the types for `vitest` so that you don't have to do `config as any`, and the whole `test` object is typed.

```ts
/// <reference types="vitest/config" />
```

(actually i think an import does the trick too:
```ts
import type "vitest/config";
```
)

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/waspConfig.ts
```diff
@@ -42,15 +63,44 @@ export function waspConfig(): PluginOption {
...
+        `but Wasp requires ${JSON.stringify(forcedValue)}`
+      );
+    }
+  }
+  if (conflicts.length > 0) {
+    throw new Error(
+      `Your vite.config.ts sets options that Wasp controls:\n` +
+      conflicts.join('\n') +
+      `\n\nRemove these from your Vite config. ` +
+      `The wasp() plugin sets them automatically.`
```
cprecioso: nitick: the mix of <code>\`</code>, `\n`, and `+` makes these lines a bit inconsistent.

cprecioso: nitpick: the `wasp()` plugin is an implementation detail, and I wouldn't like people to see this and decide to try and remove it 😂


```suggestion
      `Wasp sets them automatically.`
```

waspc/e2e-tests/Tests/ViteOptionsValidationTest.hs
FranjoMindek: It was a different reason in my case.

`TestCase` accepts shell command builder in `TestContext`.
And e.g. `waspCliStart` is a shell command builder in `WaspProjectContext`.

So even if I prepend the `"! "`,  to `waspCliStart`, I still can't put `! waspCliStart` directly into `TestCase` array, it must be inside of `inTestWaspProjectDir` array. 

I wanted to test the case when I run `wasp start` outside of a wasp project.
To escape this constraint I defined a new command without context which I can run in invalid setting.

FranjoMindek: Maybe that can be fixed with Haskell, but I didn't think of solution at that time.

FranjoMindek: Aha, I don't think we ever made the long lasting commands work.
E.g. `wasp start` `wasp db start` `wasp db studio`.
I thought they had `FIXME` above but I was wrong.
Only their tests do:

```hs
-- FIXME: @waspCliStart@ - figure out long lasting processes
waspStartTest :: Test
waspStartTest =
  Test
```

In short, we only ever use `wasp compile` in our tests currently.

FranjoMindek: We didn't work it out because it was complex to in bash + we are going to refactor bash code anyway.

infomiho: Ok, I could then write it as `wasp compile` + `npx vite start` because it will crash at that command. 

I guess long term, there could be some sort of _fail after X seconds_ timeout? I think I could write that in bash hmmm I'll try that and report back.

waspc/ChangeLog.md
infomiho: I've went with allowing users to set the `port`, it's the most harmless of the forced options.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/waspConfig.ts
infomiho: I wanted to communicate the extra information with this helper even though is quite basic. Instead of needing to write a comment.

waspc/e2e-tests/Tests/ViteOptionsValidationTest.hs
FranjoMindek: I tried to do that one, but never managed to make it work okay.

infomiho:

waspc/ChangeLog.md
```diff
@@ -5,6 +5,7 @@
 ### 🔧 Small improvements
 
 - `REACT_APP_API_URL` is now **required** when building the client for production (it no longer defaults to any value). In development mode, it still defaults to the local server URL. This prevents silently broken production deployments where the client can't reach the server. ([#3740](https://github.com/wasp-lang/wasp/pull/3740))
+- - The `wasp()` Vite plugin now enforces certain Vite config options that Wasp requires (e.g. `base`, `envPrefix`, `build.outDir`) and throws an error if the user tries to override them. ([#3771](https://github.com/wasp-lang/wasp/pull/3771))
```
cprecioso:
```suggestion
- The `wasp()` Vite plugin now enforces certain Vite config options that Wasp requires (e.g. `base`, `envPrefix`, `build.outDir`) and throws an error if the user tries to override them. ([#3771](https://github.com/wasp-lang/wasp/pull/3771))
```

PR #3804, opened 2026-02-17
chore: migrate from oslo to jose for JWT and cookie handling (#2558)

### LINKED ISSUE #2558: Migrate away from Oslo
The [oslo](https://www.npmjs.com/package/oslo) package is now deprecated in favor of the https://oslojs.dev/ project.

Some things will be drop in replacements, but for some e.g. `oslo/cookies` we'll have to investigate if we want to use a different library or write the implementation ourselves.

jourlez:
## Changes

Migrates the `@wasp.sh/lib-auth` package from the deprecated `oslo` package to `jose` for JWT operations and an inline implementation for cookie parsing.

Closes #2558

### What changed

**JWT signing/verification** (`src/node/jwt.ts`):
- Replaced `oslo/jwt` imports (`createJWT`, `validateJWT`, `JWTAlgorithm`) with `jose` (`SignJWT`, `jwtVerify`)
- The `createJWTHelpers` factory function API is **unchanged** — downstream templates continue to work without modification
- Error messages are mapped to match expected patterns (`Expired JWT`, `Invalid signature`, `Unexpected JWT error`)
- jose produces **identical JWT tokens** to oslo for the same inputs (verified by tests)

**TimeSpan class** (`src/node/jwt.ts`):
- Implemented a local `TimeSpan` class replacing `oslo`'s export
- Supports all units used in templates: `s`, `m`, `h`, `d`, `w`
- Constructor: `new TimeSpan(value, unit)` with a `.seconds` readonly property

**Cookie parsing** (`src/node/index.ts`):
- Replaced `oslo/cookie` `parseCookies` re-export with inline implementation
- Returns `Map<string, string>` (same as oslo's API)
- Handles URL-encoded values with fallback for invalid encodings

### Test results

- All 18 tests pass (6 JWT + 12 password)
- 100% code coverage
- Type exports check passes (attw)
- Build succeeds (tsdown)

### One test expectation change

`jwt.test.ts`: Updated expected error for tampered payload from `/Unexpected/` to `/Invalid signature/`. This is because jose correctly identifies a tampered payload as a signature verification failure (the HMAC over `header.payload` doesn't match), which is more accurate than oslo's generic error.

### Dependencies

- Removed: `oslo` ^1.1.2
- Added: `jose` ^6.1.3
- `@node-rs/argon2` unchanged (password.ts was already oslo-free)


cprecioso:
Hey, @jourlez, I noticed that you have sent a lot of PRs that seem vibe-coded, untested, and not reviewed by a human. I understand that you are trying to help, but this is not the right way to contribute to Wasp, and it is bordering on spammy.
I will close all of these PRs, but if you're up to see it through and pass it through review, feel free to reopen one (only one of them!), and we can work together on it; and if it's a good experience we can see about the other ones. You can also always drop by [our Discord](https://discord.gg/rzdnErX) and ask for help or guidance.

PR #3563, opened 2025-12-30
`wasp-app-runner`: Change parallelism between client and server

cprecioso:
## Description

- Resolves #3498 

From the description of #3498:

> Right now wasp-app-runner does (pseudo code):
> 
> ```js
> parallel([
>   series([ clientBuild, clientStart ]),
>   series([ serverBuild, serverStart ])
> ])
> ```
> This leads to the client usually being ready much sooner than the server, and the Vite message with the URLs being scrolled out by the server build logs.
> 
> I think it'd be much better to do:
> ```js
> series([
>   parallel([ clientBuild, serverBuild ]),
>   parallel([ clientStart, serverStart ])
> ])
> ```
> 
> So both sides start more or less at the same time.

In the end what this PR does is, more or less:

```js
await Promise.all([clientBuild(), serverBuild()])

const longRunningClient = clientStart()

// we need to wait for this first because client must strictly be up before server
await httpReady(longRunningClient)

const longRunningServer = serverStart()

await Promise.race([longRunningClient, longRunningServer])
```

<details>
<summary>Checklist not applicable, click here to show anyway</summary>

## Type of change

- [ ] **🔧 Just code/docs improvement** 
- [ ] **🐞 Bug fix** 
- [ ] **🚀 New/improved feature** 
- [ ] **💥 Breaking change** 

## Checklist

- [ ] I tested my change in a Wasp app to verify that it works as intended.

- 🧪 Tests and apps:

  - [ ] I added **unit tests** for my change. 
  - [ ] _(if you fixed a bug)_ I added a **regression test** for the bug I fixed. 
  - [ ] _(if you added/updated a feature)_ I added/updated **e2e tests** in `examples/kitchen-sink/e2e-tests`.
  - [ ] _(if you added/updated a feature)_ I updated the **starter templates** in `waspc/data/Cli/templates`, as needed.
  - [ ] _(if you added/updated a feature)_ I updated the **example apps** in `examples/`, as needed.
    - [ ] _(if you updated `examples/tutorials`)_ I updated the tutorial in the docs (and vice versa).

- 📜 Documentation:

  - [ ] _(if you added/updated a feature)_ I **added/updated the documentation** in `web/docs/`.

- 🆕 Changelog: _(if change is more than just code/docs improvement)_
  - [ ] I updated `waspc/ChangeLog.md` with a **user-friendly** description of the change.
  - [ ] _(if you did a breaking change)_ I added a step to the current **migration guide** in `web/docs/migration-guides/`.
[...]

PR #3829, opened 2026-02-24
Refactor wasp-app-runner to eliminate process.exit() calls

cprecioso:
## Description

Refactors `wasp-app-runner` to eliminate all `process.exit()` calls, replacing them with structured error handling that ensures Docker containers and child processes are properly cleaned up before exit.

**Motivation:** `process.exit()` bypasses normal cleanup paths, which can leave Docker containers and child processes running after the runner exits. By switching to thrown errors and `Disposable`/`using` patterns, we get deterministic cleanup through normal control flow.

**Key changes:**

- **`CLIError` + `logger.fatal()`**: A new `fatal()` method on the logger throws a `CLIError` (typed as `never`), which propagates up to `main()` where it sets `process.exitCode = 1` instead of calling `process.exit()`.

- **`AbortController`-based shutdown** (`shutdown.ts`): A centralized `shutdownSignal` replaces the old `ChildProcessManager` class. SIGINT/SIGTERM abort the controller, which propagates to all child processes.

- **`Process` class using `execa`**: Replaces the old `spawnWithLog`/`spawnAndCollectOutput` free functions with a `Process` class that supports `.print()`, `.collect()`, `.wait()`, and `.kill()`. Uses `detached` mode + process-group kill for reliable cleanup (handles Wasp not always forwarding signals to its children). We're using `execa` because it has facilities for easily doing what we want to do (process line-by-line, automatically kill children once the script is done, and such).

- **`startServer()` helper + `Disposable` pattern**: A new `Server` type wraps a `Process` and implements `Disposable`, so `using` declarations automatically kill containers/processes when their scope exits. Used for Postgres, SMTP, and Wasp dev/build servers.

- **`waitUntilHttp()` readiness checks**: `waspStart` and `waspBuildStart` now use `startServer()` with HTTP readiness polling on ports 3000/3001, so the runner knows when the app is actually ready.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:


wasp-app-runner/src/http.ts
```diff
@@ -1,7 +1,7 @@
 import pRetry from "p-retry";
 import { createLogger } from "./logging.js";
 
-export async function waitUntilAppReady({
+export async function waitUntilHttp({
```
infomiho: I think we had the discussion between these two names in the original PR: https://github.com/wasp-lang/wasp/pull/3563#discussion_r2652850343

wasp-app-runner/src/util/async-iterable.ts
```diff
@@ -0,0 +1,8 @@
+export async function forEach<T>(
```
infomiho, line 1: It's dealing with `async` iterables not any iterables, can we maybe reflect that in the fn and file names?

wasp-app-runner/src/logging.ts
```diff
@@ -34,5 +45,8 @@ export function createLogger(processName: string) {
     debug(message: string): void {
       log("debug", message);
     },
+    fatal(message: string): never {
```
infomiho, line 48: Cool

wasp-app-runner/src/process.ts
```diff
@@ -1,134 +1,121 @@
-import { ChildProcess, spawn } from "child_process";
-import readline from "readline";
-import { createLogger } from "./logging.js";
+import { $ } from "execa";
+import { text } from "node:stream/consumers";
```
infomiho: Maybe import this as `streamConsumers as *` to it's more clear what `text` does?

wasp-app-runner/src/process.ts
```diff
@@ -1,134 +1,121 @@
...
-      return;
-    }
-    this.logger.warn(`Received ${reason}. Cleaning up...`);
-    this.children.forEach((child) => {
-      if (!child.killed) {
-        child.kill();
-      }
-    });
-    process.exit();
+  print(): this {
```
infomiho: Why is `print` chainable and some other methods are not?

Given that chaining is not _the API_ we chose for process, I'd maybe indicate with the fn name that it's a "composer" or "builder" fn e.g. `withPrint()` or something like that.

wasp-app-runner/src/process.ts
```diff
@@ -1,134 +1,121 @@
...
+
+    // And finally we let execa attempt to kill the process for bookkeeping.
+    this.#ignoreKillError(() => this.#proc.kill("SIGINT"));
+  }
 
-  rl.on("line", callback);
+  #ignoreKillError(fn: () => void) {
+    try {
+      fn();
+    } catch (error: any) {
```
infomiho: Can we do better than `any`?

wasp-app-runner/src/server-starter.ts
```diff
@@ -0,0 +1,47 @@
+import { Logger } from "./logging.js";
```
infomiho, line 1: I got that this is a generic `server` process helper, but the `server` name is kinda overloaded in this package, it confused me. Can we name it more precisely? Can we move into some sub-dir that would give additional context?

wasp-app-runner/src/shutdown-controller.ts
```diff
@@ -0,0 +1,24 @@
+import { createLogger } from "./logging.js";
```
infomiho, line 1: Same with this file `shutdown.ts`, the logger is named `shutdown-controller`, so maybe that's a more precise name? Sub-dir with all of these process helpers looks more and more appealing.

wasp-app-runner/src/smtp.ts
```diff
@@ -1,24 +1,21 @@
...
+        "run",
+        "--rm",
+        "-p",
+        "1080:1080",
+        "-p",
+        "1025:1025",
+        "marlonb/mailcrab:latest",
+      ],
+    },
+    () => Promise.resolve(),
```
infomiho: Can this be the default value maybe?

wasp-app-runner/src/waspCli.ts
```diff
@@ -117,26 +136,24 @@ export async function getWaspVersion({
...
+  const logger = createLogger("wasp-version");
+  const { stdout, exitCode } = await new Process({
+    logger,
     cmd: waspCliCmd.cmd,
     args: [...waspCliCmd.args, "version"],
     cwd: pathToApp,
-  });
-  const stdoutDataWithoutAnsiChars = stripVTControlCharacters(stdoutData);
+  }).collect();
+  const stdoutClean = stripVTControlCharacters(stdout);
```
infomiho: Why not keep `stdoutDataWithoutAnsiChars`, I feel like it tells the story better 🤷

infomiho (commented):
Nice improvement, left some comments

infomiho:
# Some extra review comments from Claude

These are take it or leave it kind of comments, feel free to ignore them if nonsensical. 

## Code Review: `cprecioso/remove-process-exit`

### Issues

**1. `process.ts`: `env` doesn't inherit `process.env`**
`process.ts` (new) - The `Process` constructor passes `options.env` directly to execa. Unlike the old `spawnWithLog` which did `{ ...process.env, ...extraEnv }`, execa's `$()` with an explicit `env` replaces `process.env` entirely. This means `waspMigrateDb`, `waspStart`, etc. will lose `PATH` and everything else from the environment.

Either merge with `process.env` in the constructor, or use execa's `extendEnv: true` option (which is the default for `execa()` but should be verified for `$()`).

```ts
this.#proc = $(options.cmd, options.args, {
  cwd: options.cwd,
  env: options.env,
  extendEnv: true, // explicit is better here
  // ...
});
```

**2. `process.ts`: `print()` has fire-and-forget promises**
`process.ts` (new, `print()` method) - `iterable.forEach(...)` returns a promise that is never awaited or caught. If the iteration throws, it becomes an unhandled rejection.

```ts
print(): this {
  iterable.forEach(this.stdoutLines, (line) => this.#logger.info(line));
  iterable.forEach(this.stderrLines, (line) => this.#logger.error(line));
  return this;
}
```

Consider adding `.catch()` handlers or storing the promises to await later.

**3. `process.ts`: `collect()` and `print()` both consume streams**
If someone calls `.print()` then `.collect()`, the streams will have been consumed by `print()` already. There's nothing preventing this misuse. Consider making them mutually exclusive or documenting the constraint.

**4. `shutdown.ts`: `shutdownPromise` is exported but never used**
`shutdown.ts:11` - `shutdownPromise` is defined and exported but no file in this PR imports it. Remove it or use it.

wasp-app-runner/src/http.ts
cprecioso: I think it's telling that I saw this again four months later and felt the name was inappropriate 😂

`waitUntilHttp({ port: 3000 })` feels like an extremely good signature to me though, do you still feel it isn't?

wasp-app-runner/src/process.ts
cprecioso: mmm no, not really, the checks for each field access would be more effort than is worth. node doesn't use a specific Error class, so the properties are not in any type.

wasp-app-runner/src/smtp.ts
cprecioso: the type wizardry needed for having a default value feels more complex than just having this thing

cprecioso:
@infomiho ready. ai review was not helpful at all.

wasp-app-runner/src/http.ts
infomiho: I still feel the same 😄 "wait until HTTP" what? It's more like "wait until HTTP is available on port" or "wait until app is ready" (which is more abstract I admit, but that's why I also liked it.

wasp-app-runner/src/waspCli.ts
```diff
@@ -155,32 +170,27 @@ export async function waspInfo({
...
-    name: "wasp-info",
+  const { stdout, exitCode } = await new Process({
+    logger,
     cmd: waspCliCmd.cmd,
     args: [...waspCliCmd.args, "info"],
     cwd: pathToApp,
-  });
-  const stdoutDataWithoutAnsiChars = stripVTControlCharacters(stdoutData);
+  }).collect();
+  const stdoutClean = stripVTControlCharacters(stdout);
```
infomiho: What about this instance of `stdoutClean`, I've seen a rocket on the other comment.

infomiho (approved):
Responded to comment, added one more missed rename comment. I'll approve it, merge once cleaned up 😄

wasp-app-runner/src/waspCli.ts
cprecioso: can't a guy miss a spot? 😁

PR #3835, opened 2026-02-25
`web`: Add remove-version script and instructions to regenerate docs

cprecioso:
## Description

Normally we do a change to the docs on `release` it's because we want that change to take effect on the currently live docs, without waiting for next version.

However, there's a common gotcha that we have to edit both the _current_ and _versioned_ docs. This makes the workflow a bit easier by allowing us to just edit the _current_ docs and regenerate the _versioned_ docs from them.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -111,7 +111,7 @@ there is nothing else we need to do, it will be picked up as the latest version 
 
 If you are writing/updating docs on `main` for the new release of Wasp, you should edit "current" docs (docs/).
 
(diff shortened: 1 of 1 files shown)


web/scripts/remove-version.mjs
```diff
@@ -0,0 +1,38 @@
+// @ts-check
```
infomiho, line 1: Maybe you can use Typescript like we do with some of the other scripts e.g. https://github.com/wasp-lang/wasp/blob/main/waspc/tools/get-waspc-version.ts

cprecioso: I did it initially, but the TypeScript config for Docusaurus does not like ESM imports with defaults, so it was throwing up errors for me (also in `.mts`), that's why I settled on `.mjs`.

PR #3854, opened 2026-03-02
Fix crossposting canonical URL for non-blog sections

vincanger:
## Summary
- Fixes `buildCanonicalUrl` in the crossposting script to derive the URL section (`blog`, `resources`, etc.) from the file's parent directory instead of hardcoding `/blog/`
- Articles in `resources/` now correctly get `wasp.sh/resources/...` canonical URLs

## Test plan
- [x] Dry run on `resources/2026-02-24-best-frameworks-web-dev-2026.mdx` produces `https://wasp.sh/resources/2026/02/24/best-frameworks-web-dev-2026`


FranjoMindek (approved):
A bit flaky that we depend on file structure but not that really important now.

PR #3855, opened 2026-03-03
SSR: Shell prerendering

### LINKED ISSUE #3887: [Deep Freeze] Initial Static SSR support
## Goal

Implement initial static SSR support using React Router's prerendering capabilities. This is the foundational spike work that proves the approach is viable.

Tracks PR #3855.

## What this does

- Integrate React Router's prerendering into Wasp's Vite build output
- At build time, render marked pages to static HTML files using React's `prerender` API
- Ensure hydration works correctly — the CSR app loads on top of the prerendered HTML without visual glitches or re-renders
- Set up the Vite dev server to serve prerendered pages during development

cprecioso:
- Resolves #3887

## Description

This PR puts into place the **Static SSR machinery** and uses it to **pre-render the `index.html`** — the same file that was previously added as a static file — with the same output as before.

> [!NOTE]
> Making React in charge of prerendering the HTML layout file is important so, in the future, while prerendering actual pages, it can automatically insert content in the `<head>` (`<meta>`, `<script>`, `<style>`, `<link>`, etc) as it finds it throughout the component tree (that is a React 19 feature).

Key changes:
- **Layout component** replaces `index.html`: A React component (`layout.tsx`) now renders the full `<html>` document, with Mustache-injected `title` and `head`.
- **SSR entry**: New virtual file (`ssr-entry.tsx`) uses `prerenderToNodeStream` + `createStaticRouter` to prerender pages at build/dev time.
- **Client entry rewritten for hydration**: Uses `hydrateRoot` for SSR-prerendered pages and `createRoot` for the SPA fallback, with a new `useIsClient` hook for `clientOnly` fallback handling.
- **Extracted `getRouteObjects()`** from `router.tsx` so both client and SSR entries share route construction.
- **Removed `waspHtmlDev` / `waspHtmlBuild` plugins** — the SSR plugin handles all HTML serving and building.
- SSR entry renders Layout + RouterProvider only (no QueryClientProvider/WebSocketProvider) — these are context-only providers that don't affect DOM, so hydration still works with them on the client side.

## Type of change

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

- 🆕 Changelog:

### DIFF
--- a/examples/tutorials/TodoApp/package-lock.json
+++ b/examples/tutorials/TodoApp/package-lock.json
@@ -2962,14 +2962,14 @@
       "license": "MIT"
     },
     "node_modules/axios": {
-      "version": "1.13.3",
-      "resolved": "https://registry.npmjs.org/axios/-/axios-1.13.3.tgz",
-      "integrity": "sha512-ERT8kdX7DZjtUm7IitEyV7InTHAF42iJuMArIiDIV5YtPanJkgw4hw5Dyg9fh0mihdWNn1GKaeIWErfe56UQ1g==",
+      "version": "1.14.0",
+      "resolved": "https://registry.npmjs.org/axios/-/axios-1.14.0.tgz",
+      "integrity": "sha512-3Y8yrqLSwjuzpXuZ0oIYZ/XGgLwUIBU3uLvbcpb0pidD9ctpShJd43KSlEEkVQg6DS0G9NKyzOvBfUtDKEyHvQ==",
       "license": "MIT",
       "dependencies": {
-        "follow-redirects": "^1.15.6",
-        "form-data": "^4.0.4",
-        "proxy-from-env": "^1.1.0"
+        "follow-redirects": "^1.15.11",
+        "form-data": "^4.0.5",
+        "proxy-from-env": "^2.1.0"
       }
     },
     "node_modules/balanced-match": {


examples/kitchen-sink/main.wasp
```diff
@@ -4,7 +4,7 @@ app KitchenSink {
   },
   title: "Wasp Kitchen Sink",
   head: [
-    "<link rel='manifest' href='/manifest.json'>",
+    "<link rel='manifest' href='/manifest.json' />",
```
infomiho, line 7: Is this because we are now injecting these into React and no longer HTML?

web/docs/migration-guides/migrate-from-0-21-to-0-22.md
```diff
@@ -34,3 +34,8 @@ app MyApp {
 **If you don't have `app.client. envValidationSchema` or `app.server. envValidationSchema` defined in your Wasp file, you can skip this step.**
 
 Review your schemas for compatibility with Zod v4. Most schemas will work without changes, but some deprecated APIs have been removed. Refer to the [Zod v4 migration guide](https://zod.dev/v4/changelog) for details.
+### 3. Update your `head` tags to valid JSX
```
infomiho: I think we need to add some context above and explain _what changed_ e.g. "we now render the client with ... to enable SSR support later on ..."

This gives context for this breaking change and it makes more sense.

waspc/ChangeLog.md
```diff
@@ -7,6 +7,7 @@
 Remember to check out the [migration guide](https://wasp.sh/docs/migration-guides/migrate-from-0-21-to-0-22) for step-by-step documentation on how to upgrade.
 
 - Upgraded the Zod version used for env validation to Zod v4. If you use custom env validation schemas, you may need to update them to be compatible with the latest Zod API. ([#3879](https://github.com/wasp-lang/wasp/pull/3879))
+- Tags specified in `app.head` must now be valid React JSX. ([#3855](https://github.com/wasp-lang/wasp/pull/3855))
```
infomiho: I'd maybe expand on `Tags` bit here, it feels too concise to me 😄

waspc/src/Wasp/Generator/SdkGenerator/Client/VitePluginG.hs
```diff
@@ -45,6 +44,17 @@ genViteIndex = return $ C.mkTmplFd tmplPath
   where
     tmplPath = C.viteDirInSdkTemplatesDir </> [relfile|index.ts|]
 
+genWaspPlugin :: AppSpec -> Generator FileDraft
+genWaspPlugin _spec = return $ C.mkTmplFdWithData tmplPath tmplData
```
infomiho: Why provide it a `_spec` if it won't be used?

waspc/src/Wasp/Generator/SdkGenerator/Client/VitePlugin/VirtualModulesPluginG.hs
```diff
@@ -51,26 +50,39 @@ getVirtualModulesTs =
...
 
-genVirtualIndexTsx :: AppSpec -> Generator FileDraft
-genVirtualIndexTsx spec =
+routeObjectsImportJson :: Value
+routeObjectsImportJson =
+  object
+    [ "isDefined" .= True,
+      "importStatement" .= ("import { routeObjects } from \"" ++ routesEntryPointPath ++ "\""),
+      "importIdentifier" .= ("routeObjects" :: String)
+    ]
```
infomiho: Can we construct the `JsImport` object and then convert it to JSON using existing helpers?

waspc/data/Generator/templates/sdk/wasp/client/webSocket/WebSocketProvider.tsx
```diff
@@ -20,7 +20,7 @@ export const socket: Socket<ServerToClientEvents, ClientToServerEvents> = io(
   config.apiUrl,
   {
     transports: ['websocket'],
-    autoConnect: {= autoConnect =},
+    autoConnect: import.meta.env.SSR ? false : {= autoConnect =},
```
infomiho: Nitpick: This will produce code like `import.meta.env.SSR ? false : false` if `autoConnect` is false then this is just `false` otherwise it's `!import.meta.env.SSR`.

waspc/data/Generator/templates/sdk/wasp/api/index.ts
```diff
@@ -80,15 +80,17 @@ api.interceptors.response.use(undefined, (error) => {
 // Ref: https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event
 // "Note: This won't work on the same page that is making the changes — it is really a way
 // for other pages on the domain using the storage to sync any changes that are made."
-window.addEventListener('storage', (event) => {
-  if (event.key === storage.getPrefixedKey(WASP_APP_AUTH_SESSION_ID_NAME)) {
-    if (!!event.newValue) {
-      apiEventsEmitter.emit('sessionId.set')
-    } else {
-      apiEventsEmitter.emit('sessionId.clear')
+if (typeof window !== 'undefined') {
```
infomiho, line 80: Nitpick: do we want to name this e.g. `isSSRContext` or similar so it's more obvious why we disabled some code path?

waspc/data/Generator/templates/sdk/wasp/client/app/layout.tsx
```diff
@@ -0,0 +1,82 @@
+{{={= =}=}}
+import { StrictMode, type ReactNode, useSyncExternalStore } from "react";
+
+export function Layout({
```
infomiho, line 5: Could break down the logic here for me?

I see that `isClient` means we render children of the layout. Does `isClient` mean rendered in browser? What is the second option, rendered on build time/server? Why don't we then render children?

There is this `clientEntrySrc` we include if it's provided, when do we provide it, when rendering in browser or no?

As you can see, I got a bit lost in the logic. Maybe we can improve the readability and clarity of this file so it's clear e.g. in the browser we do _this_, but on the server we do _that_.

waspc/data/Generator/templates/sdk/wasp/client/app/layout.tsx
```diff
@@ -0,0 +1,82 @@
...
+              />
+            ) : null
+          }
+        </body>
+      </html>
+    </StrictMode>
+  );
+}
+
+function useIsClient() {
```
infomiho: This maybe deserves its own `useClient.ts` file given it's a unit of logic which include some stub functions that the users e.g. `layout.tsx` doesn't need to know.

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/routes.tsx
```diff
@@ -10,8 +13,35 @@ import { createAuthRequiredPage } from "wasp/client/app"
...
   {=/ routes =}
 } as const;
+
+{=# rootComponent.isDefined =}
+{=& rootComponent.importStatement =}
+{=/ rootComponent.isDefined =}
+
+{=# setupFn.isDefined =}
+{=& setupFn.importStatement =}
+{=/ setupFn.isDefined =}
```
infomiho, line 15: I'd put these import statements on top (unless they are dynamic imports?)

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/routes.tsx
```diff
@@ -10,8 +13,35 @@ import { createAuthRequiredPage } from "wasp/client/app"
...
+
+{=# setupFn.isDefined =}
+await {= setupFn.importIdentifier =}()
+{=/ setupFn.isDefined =}
+
+initializeQueryClient()
```
infomiho, line 53: This feels a bit off, executing top level setup in a routes file. Isn't there a more _root_ place we can do it?

waspc/data/Generator/templates/sdk/wasp/core/storage.ts
```diff
@@ -1,50 +1,72 @@
...
+  get(key: string): unknown;
+  remove(key: string): void;
+  clear(): void;
+};
+
+export const storage = (
+  typeof window === "undefined" || !window.localStorage
+    ? createMemoryDataStore
+    : createLocalStorageDataStore
+)("wasp");
```
infomiho: Nitpick: I find inline expressions to resolve to a function which is then invoked a bit unreadable. How about:

```suggestion
const createStorage = typeof window === "undefined" || !window.localStorage
    ? createMemoryDataStore
    : createLocalStorageDataStore
export const storage = createStorage("wasp");
```

waspc/data/Generator/libs/vite-ssr/src/plugins/common/constants.ts
```diff
@@ -0,0 +1,10 @@
+import pkg from "@wasp.sh/lib-vite-ssr/package.json" with { type: "json" };
+
+export const PACKAGE_NAME = pkg.name as `$PACKAGE_NAME`;
```
infomiho: What is going on here? Never seen this kind of type assertion.

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/client-entry.tsx
```diff
@@ -0,0 +1,42 @@
...
+// React Router will put hydration data on this property of the `window` object.
+// https://reactrouter.com/7.13.1/start/data/custom#4-hydrate-in-the-browser
+const hydrationData = (window as any).__staticRouterHydrationData as HydrationState | undefined;
+
+const router = createBrowserRouter({= routeObjects.importIdentifier =}, {
+  basename: "{= baseDir =}",
+  hydrationData,
+})
+
+function App({ isFallbackPage }: { isFallbackPage: boolean }) {
```
infomiho, line 20: Could you break down the connection between fallback page and hydration data existing? :)

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/ssr-entry.tsx
```diff
@@ -0,0 +1,65 @@
+{{={= =}=}}
+// @ts-nocheck
```
infomiho: Do we do `@ts-nocheck` in other virutal files, I don't remember. Do we have to do it? It's okay if we do, just checking.

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/ssr-entry.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+const FALLBACK_FILE = "{= ssrFallbackFile =}";
+
+export default ((async (route, ctx) => {
+  const isFallbackPage = route === FALLBACK_FILE;
+
+  if (isFallbackPage) {
+     return await prerenderApp({ isFallbackPage: true, children: null }, ctx);
+  } else {
+
```
infomiho:
```suggestion
```

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/ssr-entry.tsx
```diff
@@ -0,0 +1,65 @@
...
+  RouterProvider,
+} from "react-router";
+import { Layout } from "wasp/client/app/layout";
+import { WaspApp } from "wasp/client/app";
+
+{=& routeObjects.importStatement =}
+
+const FALLBACK_FILE = "{= ssrFallbackFile =}";
+
+export default ((async (route, ctx) => {
```
infomiho: Nitpick: Can we give this export a name?

waspc/data/Generator/libs/vite-ssr/src/plugins/common/routes.ts
```diff
@@ -0,0 +1,44 @@
+import * as posixPath from "node:path/posix";
+
+/**
```
infomiho: Nitpick: comment feel redundant

waspc/data/Generator/libs/vite-ssr/src/plugins/common/routes.ts
```diff
@@ -0,0 +1,44 @@
...
+  path: string;
+
+  /** Vite's Module ID for the route */
+  id: string;
+}
+
+export class Routes {
+  public readonly byId: ReadonlyMap<string, Readonly<Route>>;
+  public readonly byPath: ReadonlyMap<string, Readonly<Route>>;
+  public readonly fallback: Readonly<Route>;
```
infomiho: Nitpick: you could define the `fallback` value here as well?

waspc/data/Generator/libs/vite-ssr/src/plugins/common/routes.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+export interface Route {
+  /** The pathname visible in the browser's address bar. */
+  path: string;
+
+  /** Vite's Module ID for the route */
+  id: string;
+}
+
+export class Routes {
```
infomiho: I'm missing some _what is the purpose of this class_ comment, maybe I'll get once I view all the plugin files.

waspc/data/Generator/libs/vite-ssr/src/plugins/dev/css.ts
```diff
@@ -0,0 +1,49 @@
...
+  id: string;
+}
+
+declare module "vite/types/customEvent.d.ts" {
+  interface CustomEventMap {
+    [REGISTER_CSS_EVENT_NAME]: RegisterCssData;
+  }
+}
+
+const registerCssCode = (id: string) =>
```
infomiho: Nitpick: naming, it's not registering the CSS code, but it's creating the CSS registration code.

waspc/data/Generator/libs/vite-ssr/src/plugins/dev/css.ts
```diff
@@ -0,0 +1,49 @@
...
+  `
+    if (import.meta.hot) {
+      const register = () => import.meta.hot.send(${JSON.stringify(REGISTER_CSS_EVENT_NAME)}, { id: ${JSON.stringify(id)} });
+      register();
+      import.meta.hot.accept();
+      import.meta.hot.prune(register);
+    }
+  `;
+
+export const addRegisterCss = (id: string, code: string) =>
```
infomiho: This does some sort append of new CSS registration code to existing code? Does the end result look like:
```ts
if (import.meta.hot) {
 // ...
}
if (import.meta.hot) {
 // ...
}
if (import.meta.hot) {
 // ...
}
if (import.meta.hot) {
 // ...
}
```

Could we maybe concat the IDs and then produce one single code that registers all CSS files in a loop?

waspc/data/Generator/libs/vite-ssr/src/plugins/dev/css.ts
```diff
@@ -0,0 +1,49 @@
...
+  [code, registerCssCode(id)].join(";\n");
+
+export const collectRegisteredCss = async <T>(
+  hot: NormalizedHotChannel,
+  fn: () => Promise<T>,
```
infomiho: I understand this function listens on `REGISTER_CSS_EVENT_NAME` to collect all the CSS files into a `Set`, but I don't get the point of `fn`, can we name it with more intention revealing name? 😄

waspc/data/Generator/libs/vite-ssr/src/plugins/dev/index.ts
```diff
@@ -0,0 +1,98 @@
...
+
+      handler(code, id, opts) {
+        if (opts?.ssr) {
+          return addRegisterCss(id, code);
+        }
+      },
+    },
+
+    configureServer:
+      (server) =>
```
infomiho:
```suggestion
    configureServer(server {
```

Then we `return () => {}`. This is a readability nitpick and it brings the `configureServer` in line with `handler` fn style above.

waspc/data/Generator/libs/vite-ssr/src/plugins/dev/index.ts
```diff
@@ -0,0 +1,98 @@
...
+        server.middlewares.use(async ({ url, originalUrl }, res, next) => {
+          /*
+            This middleware will run after Vite's own middlewares, so they have
+            already handled requests for static assets or virtual helpers.
+
+            Only the requests for the SPA handler should reach this middleware,
+            and we can identify them by checking if `req.url === "/index.html"`.
+
+            `req.originalUrl` has the original URL before Vite's middlewares
+            potentially rewrite it, so we can prerender the correct route.
```
infomiho: I got a bit confused by the wording, the meaning is `originalUrl` is kept intact so we can be sure about the URL?

waspc/data/Generator/libs/vite-ssr/src/plugins/dev/index.ts
```diff
@@ -0,0 +1,98 @@
...
+              const { default: prerenderApp }: { default: PrerenderFn } =
+                await ssrEnv.runner.import(ssrEntrySrc);
+
+              return await prerenderApp(route, {
+                clientEntrySrc,
+                transformIndexHtml: (html) =>
+                  server.transformIndexHtml(originalUrl, html),
+              });
+            },
+          );
```
infomiho: Uf, I'd split the two concerns:
- collecting registered CSS files
- rendering the app

It feels weird to have the _rendering step_ as part of a callback for collecting CSS file.

Can we make them two distinct steps e.g.:
```ts
const cssFiles = await collectRegisteredCss(...)
const html = await ssr(...)
```

Or go even further so it's a wrapper e.g.:

```ts
const html = await appendStylesToHead(ssr(...))
```

I'd like to see some more high-level these kind of functions so it's easier to follow 😄

waspc/data/Generator/libs/vite-ssr/src/plugins/dev/index.ts
```diff
@@ -0,0 +1,98 @@
...
+
+          if (!html) {
+            return next();
+          }
```
infomiho, line 76: When can this happen, what is the UX for the user? I assume the `index.html` loads and then the SPA boots?

waspc/data/Generator/libs/vite-ssr/src/plugins/common/routes.ts
infomiho: I liked the `ssrPaths` variable name I saw where you init this class. So maybe this class should be `SsrRoutes` so it's clear - the routes to be SSRed.

waspc/data/Generator/libs/vite-ssr/src/plugins/set-environments.ts
```diff
@@ -0,0 +1,12 @@
+import { ENVIRONMENT_NAMES, PACKAGE_NAME } from "./common/constants";
+
+export const ssrSetEnvironments = () => ({
+  name: `${PACKAGE_NAME}:set-environments`,
+
+  config: () => ({
+    environments: {
+      [ENVIRONMENT_NAMES.SSR]: {},
+      [ENVIRONMENT_NAMES.CLIENT]: {},
+    },
```
infomiho, line 13: Some _why_ comment could be useful for future readers.

waspc/data/Generator/libs/vite-ssr/package.json
```diff
@@ -0,0 +1,37 @@
...
+    }
+  },
+  "files": [
+    "dist/**/*"
+  ],
+  "scripts": {
+    "build": "tsdown",
+    "dev": "tsdown --watch",
+    "prepare": "npm run build",
+    "test": "exit 0"
```
infomiho, line 24: Do we want to add some unit tests for _some of the logic_? I think it should be a thing with all of our libs. They should be proper JS packages covered with unit tests.

What to unit test? I'm not sure, maybe some of the pure logic e.g. routes manipulation or collecting the CSS file. If there aren't any good candidates, let's not do it, but let's quickly analyze and then decide.

waspc/data/Generator/libs/vite-ssr/README.md
```diff
@@ -0,0 +1,93 @@
+# @wasp.sh/vite-ssr
+
+A Vite plugin that adds SSR prerendering to React Router apps. Configured routes are prerendered to static HTML at build time, while all other routes are served a client-only SPA fallback.
+
+It is framework- and router-agnostic but designed with React 19 and React Router in mind.
```
infomiho:
```suggestion
It is framework and router-agnostic but designed with React 19 and React Router in mind.
```

waspc/data/Generator/libs/vite-ssr/README.md
```diff
@@ -0,0 +1,93 @@
...
+At runtime, the plugin serves these prerendered HTML files linking to the **Client entry** for their respective routes. For any routes not prerendered, it serves a minimal SPA shell that mounts the React app client-side.
+
+You pass it a list of **SSR paths** to prerender, and a **fallback path** used internally to generate the SPA fallback HTML. Visitors hitting an SSR path get the prerendered HTML, while all other routes serve the SPA fallback.
+
+## Setup
+
+### 1. Install
+
+```bash
+npm install @wasp.sh/vite-ssr
```
infomiho, line 26: I guess this bit isn't true yet, we didn't publish it?

waspc/data/Generator/libs/vite-ssr/README.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+**Options:**
+
+| Option            | Description                                            |
+| ----------------- | ------------------------------------------------------ |
+| `clientEntrySrc`  | Absolute path to the client entry file                 |
+| `ssrEntrySrc`     | Absolute path to the SSR entry file                    |
+| `ssrPaths`        | Array of route paths to prerender                      |
+| `ssrFallbackFile` | Path used internally to generate the SPA fallback HTML |
```
infomiho: Could you help me understand this a bit better, why do we need a special path that we use internally. Is it a _special value_ that we detect and then server `index.html`? When does this path occur? Sorry if I should have got it from the rest of the code, but I'd appreciate some high level explanation, thanks!

waspc/data/Generator/libs/vite-ssr/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+```ts
+// ssr-entry.tsx
+import type { PrerenderFn } from "@wasp.sh/vite-ssr/types";
+
+const prerender: PrerenderFn = async (route, ctx) => {
+  // Render your app to HTML using prerenderToNodeStream, renderToString, etc.
+  // Use ctx.transformIndexHtml to apply Vite's HTML transforms.
+  // Use ctx.clientEntrySrc to inject a <script> tag for the client bundle.
+  // Return a Response with the HTML body.
```
infomiho: Nipick: maybe offer example code like you did below 🤷  nbd

waspc/data/Generator/libs/vite-ssr/tsdown.config.ts
```diff
@@ -0,0 +1,14 @@
...
+  outDir: "dist",
+  clean: true,
+
+  platform: "node",
+  target: "node22",
+  format: "esm",
+
+  sourcemap: true,
+  dts: { sourcemap: true },
+});
```
infomiho, line 14: Do you think it's worth it for us to somehow standardize the libs setup/building process? e.g. through a common helper lib, through some documentation. This is out of scope for this PR, but let's maybe create an issue for this if you agree.

waspc/data/Generator/libs/vite-ssr/src/plugins/preview.ts
```diff
@@ -0,0 +1,54 @@
...
+
+          `req.originalUrl` has the original URL before Vite's middlewares
+          potentially rewrite it, so we can prerender the correct route.
+        */
+
+        if (!url || url !== "/index.html" || !originalUrl) {
+          return next();
+        }
+
+        const route = routes.byPath.get(originalUrl) ?? routes.fallback;
```
infomiho: Nitpick: This is repeat in both `dev` and in `preview`, maybe extract it into something like:
```ts
if (!isValidUrl(url, originalUrl)) {
  next()
}

const route = getRoute(originalUrl)
```

waspc/data/Generator/libs/vite-ssr/src/plugins/preview.ts
```diff
@@ -0,0 +1,54 @@
...
+};
+
+async function pathExists(...params: Parameters<typeof fs.access>) {
+  try {
+    await fs.access(...params);
+    return true;
+  } catch {
+    return false;
+  }
+}
```
infomiho: Nitpick: maybe extract this and `removeLeadingSlash` into `util/paths.ts`

waspc/data/Generator/libs/vite-ssr/src/plugins/build.ts
```diff
@@ -0,0 +1,118 @@
...
+export const ssrBuild = (
+  routes: Routes,
+  { ssrEntrySrc, clientEntrySrc }: Options,
+): Plugin => {
+  let prerenderApp: PrerenderFn | null = null;
+
+  return {
+    name: `${PACKAGE_NAME}:build`,
+    apply: "build",
+    sharedDuringBuild: true,
```
infomiho, line 18: TIL https://vite.dev/guide/api-environment-plugins#shared-plugins-during-build

waspc/data/Generator/libs/vite-ssr/src/plugins/build.ts
```diff
@@ -0,0 +1,118 @@
...
+          [ENVIRONMENT_NAMES.SSR]: {
+            build: {
+              ssr: true,
+              rollupOptions: { input: ssrEntrySrc },
+            },
+          },
+          [ENVIRONMENT_NAMES.CLIENT]: {
+            build: {
+              rollupOptions: {
+                input: Array.from(routes.byId.keys()),
```
infomiho: Nitpick: for clarity, a fn on the `Routes` class e.g. `getRouteIds()`? Or what ever the best name is for what we are doing here (getting route IDs which are used a input files for Rollup).

waspc/data/Generator/libs/vite-ssr/src/plugins/build.ts
```diff
@@ -0,0 +1,118 @@
...
+            );
+
+            const entryChunk = ssrOutput.output[0];
+            assert(
+              entryChunk,
+              "Expected ssr build output to have at least one chunk",
+            );
+            assert(
+              entryChunk.exports.includes("default"),
+              "Expected ssr build output chunk to export a default export",
```
infomiho:
```suggestion
              "Expected ssr build output chunk to have a default export",
```

Nitpick

waspc/data/Generator/libs/vite-ssr/src/plugins/build.ts
```diff
@@ -0,0 +1,118 @@
...
+            await builder.build(clientEnv);
+          },
+        },
+      };
+    },
+
+    resolveId: {
+      filter: { id: /\.html$/ },
+      handler(id, _, { ssr }) {
+        if (ssr) return;
```
infomiho:
```suggestion
        if (ssr) {
          return;
        }
```

Nitpick: I find this style less error prone, and it aligns it with the if statement below. I'd do it for all if statements in this file.

waspc/data/Generator/libs/vite-ssr/src/plugins/build.ts
```diff
@@ -0,0 +1,118 @@
...
+      async handler(id, { ssr = false } = {}) {
+        if (ssr) return;
+
+        assert(
+          prerenderApp,
+          "Expected ssr build to have completed before resolving ids in client build",
+        );
+
+        const route = routes.byId.get(id);
+        assert(route, `Unexpected id ${id} not found in ssrRoutes`);
```
infomiho:
```suggestion
        assert(route, `Expected id ${id} to exist in ssrRoutes`);
```
To be consistent with other assert messages.

waspc/data/Generator/templates/sdk/wasp/client/webSocket/WebSocketProvider.tsx
cprecioso: I know, I actually first had it as `!import.meta.env.SSR && {= autoConnect =}`, but then I changed it to the current thing as it felt more explicative

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/routes.tsx
cprecioso: no, this is the top-most place that is common to both SSR and client. in reality we shouldn't be doing side-effects at a distance at all, IMO this would be a top-level await in the query client file itself. but this is for another moment 😅

waspc/data/Generator/libs/vite-ssr/src/plugins/common/constants.ts
cprecioso: mmm the `$` makes it look like it is magic but it doesn't really have any meaning. this is unnecessary in the current incarnation of the code so I will remove it.

waspc/data/Generator/libs/vite-ssr/src/plugins/dev/css.ts
cprecioso: > Could we maybe concat the IDs and then produce one single code that registers all CSS files in a loop?

From what I tried, we can't be smarter about this unless we go full-complexity (walking the whole import tree, which is code I'd like to avoid; or creating a runtime-side registry for CSS URLs which feels overkill). 

> Does the end result look like [...]?

Yes, that's kind of the end result. This is only SSR code, this code never reaches the client. If we want to collect the IDs we still would need to communicate each of them with the `import.meta.hot`, and Vite is doing basically a regex with it and enabling specific stuff per-chunk, so there's not a straightforward way of extracting it.

waspc/data/Generator/libs/vite-ssr/src/plugins/dev/index.ts
cprecioso: > I'd like to see some more high-level these kind of functions

this was my attempt at a higher-level code 😅

> It feels weird to have the rendering step as part of a callback for collecting CSS file.

this is because the rendering step is the one that decides which CSS to collect, they can't be separated (unless we want to do the exact same work twice, once collecting the CSS and once without).

would it feel better if I called the function `withRegisteredCssCollection(blah)`, or something similar?

waspc/data/Generator/libs/vite-ssr/src/plugins/dev/index.ts
cprecioso: this would just load Vite's 404 handler. this is already the handler that should be loading the `index.html`, so if it doesn't (because the prerenderer decided not to give an html response), Vite will scream about it

waspc/data/Generator/libs/vite-ssr/README.md
cprecioso: nope, this is correct. in english when doing "framework-agnostic and router-agnostic" you can leave the dash hanging to convey that you're waiting for the suffix repetition.

waspc/data/Generator/libs/vite-ssr/README.md
cprecioso: not yet hehe

waspc/data/Generator/templates/sdk/wasp/api/index.ts
cprecioso: no, we want to keep the check inline so the bundler can detect it and remove the whole block if `false`. i can add a comment.

waspc/data/Generator/libs/vite-ssr/src/plugins/common/routes.ts
cprecioso: i don't understand this comment

waspc/data/Generator/libs/vite-ssr/package.json
cprecioso: I did think about it but decided there's nothing useful to test here. :/

waspc/data/Generator/libs/vite-ssr/tsdown.config.ts
cprecioso: we can make an issue yeah

waspc/data/Generator/libs/vite-ssr/src/plugins/preview.ts
cprecioso: the logic is somewhat different for `route`, and I think the `if`s are very informative in the body of the function, so I will keep stuff where it is if that's alright.

waspc/data/Generator/libs/vite-ssr/src/plugins/common/routes.ts
infomiho: Instead of assigning it in constructor, assign it right here :)

waspc/data/Generator/libs/vite-ssr/README.md
infomiho: Do we want to remove in the meantime?

waspc/data/Generator/libs/vite-ssr/tsdown.config.ts
infomiho: Let's do it 👍

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/client-entry.tsx
infomiho: @cprecioso what about this comment

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/ssr-entry.tsx
infomiho: @cprecioso I got a rocket, but I was wondering about the question still :)

waspc/data/Generator/libs/vite-ssr/package.json
infomiho: Alright, sometimes there isn't anything to unit test 🤷 think about it for the second time and then RAW

waspc/data/Generator/libs/vite-ssr/README.md
infomiho: @cprecioso I got a rocket, but could you maybe write a bit here as well to help me understand :)

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/routes.tsx
infomiho: > IMO this would be a top-level await in the query client file itself. but this is for another moment

Do we want to create an issue for this?

waspc/data/Generator/libs/vite-ssr/src/plugins/common/routes.ts
infomiho: My bad, that's not possible 😂  RAW

waspc/data/Generator/libs/vite-ssr/src/plugins/dev/css.ts
infomiho: Makes sense! 

Nitpick: I'd still maybe improve the naming a bit `append/add/CssRegistrationCode` 😄

waspc/data/Generator/libs/vite-ssr/src/plugins/dev/css.ts
```diff
@@ -0,0 +1,49 @@
...
+  const listener = ({ id }: RegisterCssData) => cssFiles.add(id);
+  hot.on(REGISTER_CSS_EVENT_NAME, listener);
+
+  try {
+    const result = await fnRegistersCss();
+    return { result, cssFiles };
+  } finally {
+    hot.off(REGISTER_CSS_EVENT_NAME, listener);
+  }
+};
```
infomiho, line 61: I think I understand the function now a lot better. So it's a function that renders the app _while capturing CSS files_. After it renders the app once and has the captured CSS files, those captured files are appended to the rendered HTML. Now I understand why they are intertwined like they are.

I'd recommend moving the appending to head to this fn, Claude and I made this
```ts
export async function prerenderWithCss(
  hotChannel: NormalizedHotChannel,
  render: () => Promise<string>,
): Promise<string | null> {
  const registeredCss = new Set<string>();
  const onCssRegistered = ({ id }: RegisterCssData) => registeredCss.add(id);

  hotChannel.on(REGISTER_CSS_EVENT_NAME, onCssRegistered);
  try {
    const html = await render();

    if (!html) return null;
    if (registeredCss.size === 0) return html;

    const linkTags = [...registeredCss]
      .map((href) => `<link rel="stylesheet" href="${href}">`)
      .join("\n");

    return appendToHead(html, linkTags);
  } finally {
    hotChannel.off(REGISTER_CSS_EVENT_NAME, onCssRegistered);
  }
}
```

The the usage looks like this:
```ts
const html = await prerenderWithCss(ssrEnv.hot, async () => {
  const { default: prerenderApp } = await ssrEnv.runner.import(ssrEntrySrc);
  return prerenderApp(route, {
    clientEntrySrc,
    transformIndexHtml: (html) => server.transformIndexHtml(originalUrl, html),
  });
});

if (!html) return next();
```

I kinda like this, because it gives us the final `html` after you called the fn. It's  `const html = addCssToMyStuff(renderFn)` which is easy for me to understand.

waspc/data/Generator/libs/vite-ssr/src/plugins/dev/index.ts
```diff
@@ -0,0 +1,101 @@
...
+      filter: { id: /\.css$/ },
+
+      // We're asking to go _after_ Vite's own CSS handling, which converts CSS
+      // imports into JS modules that inject styles at runtime. We want to run
+      // after that so we can register the final CSS file paths.
+      order: "post",
+
+      handler(code, id, opts) {
+        if (opts?.ssr) {
+          return addRegisterCss(id, code);
```
infomiho: Nitpick: it would be awesome if this fn was somehow connected to the `withCollectingRegisteredCss` fn e.g.
```ts
const cssRegistry = createCssRegistry()
// (here you do)
cssRegistry.injectRegistrationCode(id, code)
// (below you do)
cssRegistry.prerenderWithCss(...)
```

web/docs/migration-guides/migrate-from-0-21-to-0-22.md
```diff
@@ -34,3 +34,11 @@ app MyApp {
...
 
 Review your schemas for compatibility with Zod v4. Most schemas will work without changes, but some deprecated APIs have been removed. Refer to the [Zod v4 migration guide](https://zod.dev/v4/changelog) for details.
+
+### 3. Update your `app.head` tags to be valid JSX
+
+**If you don't have `app.head` defined in your Wasp file, you can skip this step.**
+
+We now use React to output the base `index.html` for your client app, as we set the groundwork for SSR support in the future. This means that the contents of `app.head` are now rendered as React JSX instead of raw HTML.
+
+Make sure all tags in `app.head` are valid React JSX — every tag must either be self-closing (e.g. `<meta ... />`) or have a matching closing tag (e.g. `<script>...</script>`). In JSX, even void HTML elements (like `<link>`, `<meta>`, and `<base>`) need a trailing `/>`.
```
infomiho:
```suggestion
To make all tags in `app.head` are valid React JSX, make sure that every tag is either a self-closing tag (e.g. `<meta ... />`) or has a matching closing tag (e.g. `<script>...</script>`). In JSX, even elements like `<link>`, `<meta>`, and `<base>` need a trailing `/>`.
```

waspc/data/Generator/templates/sdk/wasp/client/app/hooks/useIsClient.ts
```diff
@@ -0,0 +1,22 @@
+import { useSyncExternalStore } from "react"
```
infomiho, line 1: Nipitck: `util.ts` is not very informative name, `useClient.ts` could be in line with how people name hooks files in the ecosystem?

waspc/data/Generator/templates/sdk/wasp/client/app/layout.tsx
```diff
@@ -0,0 +1,92 @@
...
+          <title>{= title =}</title>
+        </head>
+        <body>
+          <noscript>You need to enable JavaScript to run this app.</noscript>
+
+          {
+            // We don't really need to wrap the app in a div nor name it "root",
+            // but we keep it for backwards compatibility with older Wasp
+            // versions.
+          }
```
infomiho, line 61: Nitpick: we could even remove his comment, it's documenting something we _didn't do, but could_

waspc/data/Generator/libs/vite-ssr/src/plugins/build.ts
```diff
@@ -0,0 +1,124 @@
...
+          return;
+        }
+
+        assert(
+          prerenderApp,
+          "Expected ssr build to have completed before resolving ids in client build",
+        );
+
+        const route = routes.byId.get(id);
+        assert(route, `Epected id ${id} to exist in ssrRoutes`);
```
infomiho:
```suggestion
        assert(route, `Expected id ${id} to exist in ssrRoutes`);
```

waspc/data/Generator/libs/vite-ssr/README.md
```diff
@@ -0,0 +1,207 @@
...
+Let's do an analogy. In your body, all the cells have the same DNA, that gets expressed differently. When you need a new red blood cell, you might have an already specialized cell available in your body's storage; or otherwise you might have to send a stem cell that has the ability to turn into anything. But regardless, when either cell reaches the bloodstream, they both have to be able to act like a red blood cell!
+
+In the same way, with prerendering, we can create different kinds of outputs from the same source code. When a page is requested, you might have done the work ahead of time and send a prepopulated HTML. Or you might send a blank HTML that has just enough code so that it can turn into any page of the app. Regardless, whenever either of the HTML pages reach the client, it's showtime, and they must "turn on" and start presenting the page the user requested.
+
+#### Application
+
+The Vite SSR plugin can output two different kinds of pages:
+
+- A **prerendered** page that is specialized to the requested route, and then hydrated in the client
+- A **fallback** page that is mostly blank wiht just the "shell", and then fully rendered on the client
```
infomiho:
```suggestion
- A **fallback** page that is mostly blank with just the "shell", and then fully rendered on the client
```

I'd recommend asking Claude to check for typos 👍

waspc/data/Generator/libs/vite-ssr/README.md
```diff
@@ -0,0 +1,207 @@
...
+if (window.__staticRouterHydrationData) {
+  // Prerendered page — hydrate
+  hydrateRoot(document, <App />);
+} else {
+  // SPA fallback — full client render
+  createRoot(document).render(<App />);
+}
+```
+
+## FAQ
```
infomiho, line 110: Great explanations 👍

waspc/src/Wasp/JsImport.hs
```diff
@@ -73,10 +74,14 @@ getJsImportStmtAndIdentifier (JsImport importPath importName maybeImportAlias) =
   (importStatement, importIdentifier)
   where
     importStatement = "import " ++ importClause ++ " from '" ++ pathString ++ "'"
+
     (importIdentifier, importClause) = getJsImportIdentifierAndClause importName maybeImportAlias
+
     pathString = case importPath of
       RelativeImportPath relPath -> normalizePath $ SP.fromRelFileP relPath
       ModuleImportPath modulePath -> SP.fromRelFileP modulePath
+      ExternalImportName moduleName -> moduleName
```
infomiho: Nitpick: `RawImportName`?

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/client-entry.tsx
cprecioso: i think this is now explained IMO?

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/ssr-entry.tsx
cprecioso: you got a rocket because it's not needed in the end, just a vestige of development :D

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/routes.tsx
cprecioso: https://github.com/wasp-lang/wasp/issues/3986

waspc/data/Generator/libs/vite-ssr/tsdown.config.ts
cprecioso: wasp-lang/wasp#3988

waspc/data/Generator/libs/vite-ssr/README.md
cprecioso: So, we usually have two kinds of outputs for web frameworks
- **SPA**, which means everything goes to a single file (usually `/index.html` or `/200.html`
- **MPA**, which means every request will go their own file (`/about.html`, `/products/lemon-pie.html` etc)

We're trying to tack on an opt-in MPA on top of an existing SPA here, so we need to do a multi-tiered approach.

1. We first do the MPA routing of looking if there's an HTML file that satisfies the request
2. If not, we serve the SPA file

The problem is that we can't name the SPA file `/index.html` because we might have prerendered **the actual** `/` route (which has to get saved at `/index.html`). So we ask the user of the plugin what they want to name this SPA fallback file, which I usually placehold it as `_fallback.html` (but based on research from Franjo, might better be called `200.html`).

---

Then, when the SSR entry gets asked to render a path, it can check if it's the same string as the `ssrFallbackFile`, and special-case a render that will work as an SPA by not pre-rendering any specific content.

waspc/data/Generator/libs/vite-ssr/src/plugins/dev/index.ts
cprecioso: I decided against this, there's no real runtime connection (i.e. no shared object) between the two things, so it would be lying a bit. I think all of the logic being in the `css.ts` is enough.

waspc/data/Generator/templates/sdk/wasp/client/app/layout.tsx
cprecioso: I'd keep it, I know that if I didn't know the context, I would immediately go on to try to remove the unnecessary wrapper.

waspc/data/Generator/libs/vite-ssr/README.md
infomiho: Thanks 👍 all clear!

PR #3856, opened 2026-03-03
Make Version be a valid SemVer version

FranjoMindek:
## Summary
- Makes our `Version` implementation follow the SemVer specification: https://semver.org/
	- The main difference is that SemVer version requires all 3 version components to be present (`major.minor.patch`), so partial versions are not allowed (e.g. `1.2`).
- Temporarily that means that our `Range` implementation does not allow partial versions (even though it should). We will be solving this in the next PR: https://github.com/wasp-lang/wasp/pull/3857

### DIFF
--- a/waspc/src/Wasp/SemanticVersion/Parsers.hs
+++ b/waspc/src/Wasp/SemanticVersion/Parsers.hs
@@ -5,14 +5,14 @@ where
 
 import Control.Applicative ((<|>))
 import Numeric.Natural (Natural)
-import Text.Parsec (Parsec, digit, many, notFollowedBy, oneOf, string, try)
+import qualified Text.Parsec as P
 
-naturalNumberParser :: Parsec String () Natural
+naturalNumberParser :: P.Parsec String () Natural
 naturalNumberParser =
   read <$> (zeroParser <|> noLeadingZeroNaturalNumberParser)
   where
-    zeroParser :: Parsec String () String
-    zeroParser = try (string "0" <* notFollowedBy digit)
+    zeroParser :: P.Parsec String () String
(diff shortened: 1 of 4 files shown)


waspc/src/Wasp/SemanticVersion/Parsers.hs
```diff
@@ -0,0 +1,18 @@
+module Wasp.SemanticVersion.Parsers
+  ( naturalNumberParser,
+  )
+where
+
+import Control.Applicative ((<|>))
+import Numeric.Natural (Natural)
+import Text.Parsec (Parsec, digit, many, notFollowedBy, oneOf, string, try)
+
+naturalNumberParser :: Parsec String () Natural
```
FranjoMindek: This is extracted to a separate module because this parser will be re-used in the future PRs.

waspc/src/Wasp/SemanticVersion/Version.hs
```diff
@@ -2,46 +2,54 @@
...
+-- TODO: Add pre-release (-) and build (+) support.
 versionParser :: Parsec String () Version
 versionParser = do
-  naturalP `sepBy` char '.' >>= \case
-    [a] -> return $ Version a 0 0
-    [a, b] -> return $ Version a b 0
-    [a, b, c] -> return $ Version a b c
-    _invalidFormat -> fail "Invalid version format"
+  (mjr, mnr, ptc) <- versionCoreParser
+  pure (Version mjr mnr ptc)
```
FranjoMindek, line 43: `Wasp.SemanticVersion` already used `pure` instead of `return`.
That's why I also use it here.

FranjoMindek:
Moved to https://github.com/wasp-lang/wasp/pull/3915

PR #3857, opened 2026-03-03
Add full Comparator `node-semver` implementation

FranjoMindek:
## Summary
- This PR does the heavy lifting of the `node-semver` logic implementation. It properly implements the full `comparator` logic which internally uses `partial version`s: https://github.com/npm/node-semver#advanced-range-syntax
	- **BUT the added files are not yet wired together into the old code**, so this PR basically adds "dead code". The next PR will refactor the `SemanticVersioning` module to use the new comparator here: https://github.com/wasp-lang/wasp/pull/3858

Our previous `comparator` solution:
- didn't support all operators
- used weird half SemVer version half partial version implementation internally

This PR `comparator` solution:
- supports all operators
- uses properly  implemented `partial version`
- also provides a parser

### DIFF
--- a/waspc/tests/SemanticVersion/ComparatorTest.hs
+++ b/waspc/tests/SemanticVersion/ComparatorTest.hs
@@ -111,6 +111,8 @@ spec_SemanticVersion_Comparator = do
       isLeft (strictParseHyphenRangeComp "1.2") `shouldBe` True
       isLeft (strictParseHyphenRangeComp "1.2 - ") `shouldBe` True
       isLeft (strictParseHyphenRangeComp "1.2 - a") `shouldBe` True
+      isLeft (strictParseHyphenRangeComp "1.2 -  3.4") `shouldBe` True
+      isLeft (strictParseHyphenRangeComp "1.2  - 3.4") `shouldBe` True
       isLeft (strictParseHyphenRangeComp "1.2-3.4") `shouldBe` True
 
   describe "versionBounds" $ do


waspc/src/Wasp/SemanticVersion/Comparator.hs
```diff
@@ -0,0 +1,169 @@
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
+
+{-# HLINT ignore "Use <$>" #-}
```
FranjoMindek, line 3: Otherwise it complains on the `hyphenRangeComparatorParser`.
IMHO current approach is really clean, and I don't want to use `<$>`.

waspc/src/Wasp/SemanticVersion/Comparator.hs
```diff
@@ -0,0 +1,169 @@
...
+    noVersionsInterval,
+  )
+
+-- | A comparator is composed of an operator and a partial version.
+-- It represents a single version constraint in a range.
+-- See: https://github.com/npm/node-semver#ranges
+data Comparator
+  = PrimitiveComparator PrimitiveOperator PartialVersion
+  | BackwardsCompatibleWith PartialVersion
+  | ApproximatelyEquivalentTo PartialVersion
```
FranjoMindek: Previously we only had `PrimitiveComparator` + `BackwardsCompatibleWith`.
The `ApproximatelyEquivalentTo`, `XRange` and `HyphenRange` are new.

waspc/src/Wasp/SemanticVersion/Comparator.hs
```diff
@@ -0,0 +1,169 @@
...
+
+-- | We rely on this 'show' implementation to produce valid `node-semver` comparator.
+instance Show Comparator where
+  show (PrimitiveComparator op pv) = show op ++ show pv
+  show (BackwardsCompatibleWith pv) = "^" ++ show pv
+  show (ApproximatelyEquivalentTo pv) = "~" ++ show pv
+  show (XRange pv) = show pv
+  show (HyphenRange pv1 pv2) = show pv1 ++ " - " ++ show pv2
+
+instance HasVersionBounds Comparator where
```
FranjoMindek, line 66: This one might be a bit confusing, I recommend looking here to get the picture of what we are describing:

https://github.com/wasp-lang/wasp/pull/3857/changes#diff-5daec1d19f4855877af0896c4c9380de120e783e8301c862ed827d75ff761d2eR115-R174

waspc/src/Wasp/SemanticVersion/PartialVersion.hs
```diff
@@ -0,0 +1,74 @@
...
+import Text.Parsec (ParseError, Parsec, char, oneOf, optionMaybe, parse, try)
+import Text.Printf (printf)
+import Wasp.SemanticVersion.Parsers (naturalNumberParser)
+import Wasp.SemanticVersion.Version (Version (..))
+import Wasp.Util.TH (quasiQuoterFromParser)
+
+-- | Version representation used in `node-semver` range expressions.
+-- Unlike 'Version' which always has all 3 components,
+-- 'PartialVersion' can represent partial/wildcard versions as they appear in range.
+data PartialVersion
```
FranjoMindek, line 27: This one much better describes the state than the "maybe" approach.

E.g. 
```hs
type PartialVersion' = Maybe (Natural, Maybe (Natural, Maybe Natural))
```

infomiho:
@FranjoMindek I'll review this later, but first one question, this looks a lot like something we'd publish as a separate `semver` lib (due to its completeness). This got me to thinking, is there an existing lib, and there is https://hackage.haskell.org/package/semver

Can we just use the existing lib for our needs and not maintain the whole semver impl? If the lib is not sufficient, does that mean we should publish our implementation as a better lib for others to use?

FranjoMindek:
The linked library only does the SemVer version.
This is the easy part.

The above is Haskell implementation of `node-semver` package which does more than just SemVer version. It is about version ranges you see in the node ecosystem, e.g. `>2.1.3 || 3 || ^4.3.0 >4.1`.

The `node-semver` ranges internally use different logic than SemVer.
E.g. `1` is an allowed range in `node-semver` but it is not a valid SemVer version.

waspc/src/Wasp/SemanticVersion/Comparator.hs
```diff
@@ -0,0 +1,163 @@
...
+
+-- | A comparator is composed of an operator and a partial version.
+-- It represents a single version constraint in a range.
+-- See: https://github.com/npm/node-semver#ranges
+data Comparator
+  = PrimitiveComparator PrimitiveOperator PartialVersion
+  | BackwardsCompatibleWith PartialVersion
+  | ApproximatelyEquivalentTo PartialVersion
+  | XRange PartialVersion
+  | HyphenRange PartialVersion PartialVersion
```
infomiho: I'd love to see some docs above each type of comparator e.g.

```suggestion
  = 
  -- <exampleOperator1> 1.2.3, <exampleOperator2> 1.2.3
  PrimitiveComparator PrimitiveOperator PartialVersion
  | 
  -- ^1.2.3
  BackwardsCompatibleWith PartialVersion
  |
  -- ~1.2.3
  ApproximatelyEquivalentTo PartialVersion
  |
  -- 1.2.x
  XRange PartialVersion
  |
  -- 1.2.3 - 2.0.0
  HyphenRange PartialVersion PartialVersion
```

Something to make it clear what do these mean so it's easier to follow along.

waspc/src/Wasp/SemanticVersion/Comparator.hs
```diff
@@ -0,0 +1,163 @@
...
+toXRangeLowerBound (MajorMinor mjr mnr) = Inclusive $ Version mjr mnr 0
+toXRangeLowerBound (Full mjr mnr ptc) = Inclusive $ Version mjr mnr ptc
+
+data PrimitiveOperator
+  = Equal
+  | LessThan
+  | LessThanOrEqual
+  | GreaterThan
+  | GreaterThanOrEqual
+  deriving (Eq)
```
infomiho: I'd pull this up next to `Comparator` data type

waspc/src/Wasp/SemanticVersion/Comparator.hs
```diff
@@ -0,0 +1,163 @@
...
+-- See `hyphen` definition here: https://github.com/npm/node-semver#range-grammar
+hyphenRangeComparatorParser :: P.Parsec String () Comparator
+hyphenRangeComparatorParser = do
+  lowerVersion <- partialVersionParser
+  _ <- hyphenParser
+  upperVersion <- partialVersionParser
+  pure $ HyphenRange lowerVersion upperVersion
+  where
+    hyphenParser :: P.Parsec String () Char
+    hyphenParser = P.space *> P.char '-' <* P.space
```
infomiho, line 133: Does this handle `1.2.1   -    2.0` (multiple spaces before and after `-`)?

waspc/src/Wasp/SemanticVersion/PartialVersion.hs
```diff
@@ -0,0 +1,77 @@
...
+
+-- | We rely on this 'show' implementation to produce valid `node-semver` partial version.
+instance Show PartialVersion where
+  show (Full mjr mnr ptc) = printf "%d.%d.%d" mjr mnr ptc
+  show (MajorMinor mjr mnr) = printf "%d.%d" mjr mnr
+  show (Major mjr) = printf "%d" mjr
+  show Any = "*"
+
+fromVersion :: Version -> PartialVersion
+fromVersion (Version mjr mnr ptc) = Full mjr mnr ptc
```
infomiho: Another suggestion by our buddy Claude. It's used in the `Comparator` helper fns.

Consider extracting a `toMinVersion` helper that names the "fill zeros" pattern.

```suggestion
fromVersion :: Version -> PartialVersion
fromVersion (Version mjr mnr ptc) = Full mjr mnr ptc

toMinVersion :: PartialVersion -> Version
toMinVersion Any = Version 0 0 0
toMinVersion (Major mjr) = Version mjr 0 0
toMinVersion (MajorMinor mjr mnr) = Version mjr mnr 0
toMinVersion (Full mjr mnr ptc) = Version mjr mnr ptc
```

waspc/src/Wasp/SemanticVersion/Comparator.hs
```diff
@@ -0,0 +1,163 @@
...
+toXRangeUpperBound Any = Inf
+toXRangeUpperBound (Major mjr) = Exclusive (Version (mjr + 1) 0 0)
+toXRangeUpperBound (MajorMinor mjr mnr) = Exclusive (Version mjr (mnr + 1) 0)
+toXRangeUpperBound (Full mjr mnr ptc) = Inclusive (Version mjr mnr ptc)
+
+toXRangeLowerBound :: PartialVersion -> VersionBound
+toXRangeLowerBound Any = Inclusive $ Version 0 0 0
+toXRangeLowerBound (Major mjr) = Inclusive $ Version mjr 0 0
+toXRangeLowerBound (MajorMinor mjr mnr) = Inclusive $ Version mjr mnr 0
+toXRangeLowerBound (Full mjr mnr ptc) = Inclusive $ Version mjr mnr ptc
```
infomiho: Here's a suggestion from our buddy Claude on how to refactor it. I had trouble with parsing the instance definition due to nested expressions and the `where` block. This makes sense, one top level `versionBounds` + top level helper methods for all the ranges. 

```suggestion
instance HasVersionBounds Comparator where
  versionBounds (PrimitiveComparator op pv) = primitiveBounds op pv
  versionBounds (XRange pv) = (lowerBound pv, upperBound pv)
  versionBounds (HyphenRange lower upper) = (lowerBound lower, upperBound upper)
  versionBounds (BackwardsCompatibleWith pv) = (lowerBound pv, caretUpperBound pv)
  versionBounds (ApproximatelyEquivalentTo pv) = (lowerBound pv, tildeUpperBound pv)

-- | Lower bound that fills unspecified components with zeros.
lowerBound :: PartialVersion -> VersionBound
lowerBound = Inclusive . toMinVersion

-- | Upper bound for X-range semantics: the next version above the specified range.
-- For a full version, the upper bound is inclusive (exact match).
upperBound :: PartialVersion -> VersionBound
upperBound Any = Inf
upperBound (Major mjr) = Exclusive (Version (mjr + 1) 0 0)
upperBound (MajorMinor mjr mnr) = Exclusive (Version mjr (mnr + 1) 0)
upperBound (Full mjr mnr ptc) = Inclusive (Version mjr mnr ptc)

-- | Version bounds for primitive operators (=, <, <=, >, >=).
primitiveBounds :: PrimitiveOperator -> PartialVersion -> VersionInterval
primitiveBounds Equal pv = (lowerBound pv, upperBound pv)
primitiveBounds GreaterThanOrEqual pv = (lowerBound pv, Inf)
primitiveBounds LessThanOrEqual pv = (Inclusive (Version 0 0 0), upperBound pv)
primitiveBounds LessThan Any = noVersionsInterval
primitiveBounds LessThan pv = (Inclusive (Version 0 0 0), Exclusive (toMinVersion pv))
primitiveBounds GreaterThan Any = noVersionsInterval
primitiveBounds GreaterThan (Full mjr mnr ptc) = (Exclusive (Version mjr mnr ptc), Inf)
primitiveBounds GreaterThan (Major mjr) = (Inclusive (Version (mjr + 1) 0 0), Inf)
primitiveBounds GreaterThan (MajorMinor mjr mnr) = (Inclusive (Version mjr (mnr + 1) 0), Inf)

-- | Caret (^) upper bound: allows changes that don't modify the leftmost non-zero digit.
caretUpperBound :: PartialVersion -> VersionBound
caretUpperBound Any = Inf
caretUpperBound (Major 0) = Exclusive (Version 1 0 0)
caretUpperBound (Major mjr) = Exclusive (Version (mjr + 1) 0 0)
caretUpperBound (MajorMinor 0 0) = Exclusive (Version 0 1 0)
caretUpperBound (MajorMinor 0 mnr) = Exclusive (Version 0 (mnr + 1) 0)
caretUpperBound (MajorMinor mjr _) = Exclusive (Version (mjr + 1) 0 0)
caretUpperBound (Full 0 0 ptc) = Exclusive (Version 0 0 (ptc + 1))
caretUpperBound (Full 0 mnr _) = Exclusive (Version 0 (mnr + 1) 0)
caretUpperBound (Full mjr _ _) = Exclusive (Version (mjr + 1) 0 0)

-- | Tilde (~) upper bound: allows patch-level changes if minor is specified.
tildeUpperBound :: PartialVersion -> VersionBound
tildeUpperBound Any = Inf
tildeUpperBound (Major mjr) = Exclusive (Version (mjr + 1) 0 0)
tildeUpperBound (MajorMinor mjr mnr) = Exclusive (Version mjr (mnr + 1) 0)
tildeUpperBound (Full mjr mnr _) = Exclusive (Version mjr (mnr + 1) 0)
```

waspc/src/Wasp/SemanticVersion/PartialVersion.hs
```diff
@@ -0,0 +1,77 @@
...
+import Wasp.SemanticVersion.Parsers (naturalNumberParser)
+import Wasp.SemanticVersion.Version (Version (..))
+import Wasp.Util.TH (quasiQuoterFromParser)
+
+-- | Version representation used in `node-semver` range expressions.
+-- Unlike 'Version' which always has all 3 version components,
+-- 'PartialVersion' can represent partial/wildcard versions as they appear in a range.
+data PartialVersion
+  = -- | Major, minor and patch version (e.g. 1.2.3).
+    Full !Natural !Natural !Natural
```
infomiho: Maybe go all-in on the naming and call this one `MajorMinorPatch`?

waspc/src/Wasp/SemanticVersion/PartialVersion.hs
```diff
@@ -0,0 +1,77 @@
...
+    [Nothing] -> pure Any -- "*" / "x" / "X"
+    [Just mjr] -> pure (Major mjr) -- "1"
+    [Just mjr, Nothing] -> pure (Major mjr) -- "1.x"
+    [Just mjr, Nothing, Nothing] -> pure (Major mjr) -- "1.x.x"
+    [Just mjr, Just mnr] -> pure (MajorMinor mjr mnr) -- "1.2"
+    [Just mjr, Just mnr, Nothing] -> pure (MajorMinor mjr mnr) -- "1.2.x"
+    [Just mjr, Just mnr, Just ptc] -> pure (Full mjr mnr ptc) -- "1.2.3"
+    [Nothing, _, _] -> fail "wildcard must be the only component"
+    [_, Nothing, Just _] -> fail "patch cannot be specified if minor is wildcard"
+    _ -> fail "invalid version form"
```
infomiho, line 71: Can we do something about the wildcard parsing to be more explicit? Now we use `Nothing` to signal it, but we have a nice type system which we might model this with e.g. `data VersionComponent = Wildcard | Number Natural`. Then the pattern matching would read more naturally:

```haskell
case versionComponents of
    [Wildcard] -> pure Any
    [Number mjr] -> pure (Major mjr)
    [Number mjr, Wildcard] -> pure (Major mjr)
    [Number mjr, Wildcard, Wildcard] -> pure (Major mjr)
    [Number mjr, Number mnr] -> pure (MajorMinor mjr mnr)
    [Number mjr, Number mnr, Wildcard] -> pure (MajorMinor mjr mnr)
    [Number mjr, Number mnr, Number ptc] -> pure (Full mjr mnr ptc)
    [Wildcard, _, _] -> fail "wildcard must be the only component"
    [_, Wildcard, Number _] -> fail "patch cannot be specified if minor is wildcard"
    _ -> fail "invalid version form"
```

waspc/tests/SemanticVersion/ComparatorTest.hs
```diff
@@ -0,0 +1,180 @@
...
+      strictParseSimpleComp "1.2" `shouldBe` Right (XRange (MajorMinor 1 2))
+      strictParseSimpleComp "1.x" `shouldBe` Right (XRange (Major 1))
+      strictParseSimpleComp "1" `shouldBe` Right (XRange (Major 1))
+      strictParseSimpleComp "X" `shouldBe` Right (XRange Any)
+      strictParseSimpleComp "x" `shouldBe` Right (XRange Any)
+      strictParseSimpleComp "*" `shouldBe` Right (XRange Any)
+
+    it "parses simple comparators with trailing content" $ do
+      parseSimpleComp "* 1.2.3" `shouldBe` Right (XRange Any)
+      parseSimpleComp "<1.2.3 || 5" `shouldBe` Right (PrimitiveComparator LessThan (Full 1 2 3))
```
infomiho: We don't support the `||` operator? Just checking if I understood!

FranjoMindek: Thats the job for `ComparatorSet` (next PR) which can hold multiple `Comparator` and joins them via the `||` operator. `Comparator` itself doesn't know about `||`, so it should ignore it.

waspc/src/Wasp/SemanticVersion/Comparator.hs
FranjoMindek: I actually had those before :D, but removed them somewhere along the way.
I guess I should return them.

waspc/src/Wasp/SemanticVersion/Comparator.hs
FranjoMindek: Also debated this myself, wasn't sure about it.
Made it so its `Comparator` -> `PrimitiveComparator` -> their `Show` -> comparators `HasVersionBounds`

waspc/src/Wasp/SemanticVersion/Comparator.hs
FranjoMindek: It shouldn't, going by grammar. I've added tests which check this:
```hs
isLeft (strictParseHyphenRangeComp "1.2 -  3.4") `shouldBe` True
isLeft (strictParseHyphenRangeComp "1.2  - 3.4") `shouldBe` True
```

waspc/src/Wasp/SemanticVersion/PartialVersion.hs
FranjoMindek: `toXRangeLowerBound` kinda did that homework, I'll check the refactor suggestion below first.

Edit: will discuss this in comment below.

waspc/src/Wasp/SemanticVersion/Comparator.hs
FranjoMindek: This: 
- Moves `where` helpers to outside.
  - I purposely didn't want that. Those functions are used at once place and I wanted them colocated (less jumping around the code).
- Renames `toXRangeUpperBound` -> `upperBound`. Same for lower bound.
  - Removing context purposely just to add it as a comment? Makes no sense. 
- Extracts `toMinVersion` from `toXRangeLowerBound`, just so it can be **partialy reused** for `LessThan` primitive operator.
  - I explicitly didn't want to do this. `LessThan` is a special case, more so it can be seen because it just partially matches the `toXRangeLowerBound` bounds. Makes no sense to make abstractions where they don't exist, but are accidents.

The only subjective one (for me) is the first one.
The choice if functions should be module level or scoped to their use case.

We can discuss that one.
So for now I won't do this suggestion.

infomiho: It was very hard to parse the instance for me and it's a focused module, so having things on top level is not that big of a deal. Nesting is the bigger evil here if you ask me, so I'd avoid it if possible. For the rest, I didn't dig deeper into Claude's suggestion but the one I'd like the most is this simplification:
```hs
instance HasVersionBounds Comparator where
  versionBounds (PrimitiveComparator op pv) = ...
  versionBounds (XRange pv) = ...
  versionBounds (HyphenRange lower upper) = ...
  versionBounds (BackwardsCompatibleWith pv) = ...
  versionBounds (ApproximatelyEquivalentTo pv) = ...
```

waspc/src/Wasp/SemanticVersion/PartialVersion.hs
infomiho: I'd argue that `toXRangeLowerBound` is too specific of a name 🤷 it's not specific to X range?

waspc/src/Wasp/SemanticVersion/Comparator.hs
```diff
@@ -0,0 +1,168 @@
...
+  ( HasVersionBounds (..),
+    VersionBound (..),
+    noVersionsInterval,
+  )
+
+-- | A comparator is composed of an operator and a partial version.
+-- It represents a single version constraint in a range.
+-- See: https://github.com/npm/node-semver#ranges
+data Comparator
+  = -- | =1.2.3, >1.2.3, <1.2.3, >=1.2.3, <=1.2.3
```
infomiho, line 28:
```suggestion
  = -- | 1.2.3, >1.2.3, <1.2.3, >=1.2.3, <=1.2.3
```

Maybe keep the canonical form here as well 🤷

waspc/src/Wasp/SemanticVersion/PartialVersion.hs
```diff
@@ -0,0 +1,79 @@
...
+    [_, Wildcard, Number _] -> fail "patch cannot be specified if minor is wildcard"
+    _ -> fail "invalid version form"
+  where
+    versionComponentParser :: P.Parsec String () VersionComponent
+    versionComponentParser = (Wildcard <$ wildcardParser) <|> (Number <$> naturalNumberParser)
+
+    wildcardParser :: P.Parsec String () ()
+    wildcardParser = void (P.oneOf "xX*")
+
+data VersionComponent = Wildcard | Number Natural
```
infomiho, line 79: Missing newline at the end of the file.

FranjoMindek:
Moved here: https://github.com/wasp-lang/wasp/pull/3916

PR #3858, opened 2026-03-03
Wire new `node-semver` implementations, make `SemanticVersion` a re-export hub

FranjoMindek:
## Summary

- Extracts the `ComparatorSet` and `Range` into separate modules.
- Makes the `ComparatorSet` use the new `Comparator` module added in the previous PR.
- Adds both `ComparatorSet` and `Range` parsers.
- Makes the `Wasp.SemanticVersion` a re-export hub for all submodules. So consumers which used `import qualified Wasp.SemanticVersion as SV` don't have to change anything.

Solves:https://github.com/wasp-lang/wasp/issues/3699

### DIFF
--- a/waspc/src/Wasp/SemanticVersion/Comparator.hs
+++ b/waspc/src/Wasp/SemanticVersion/Comparator.hs
@@ -25,11 +25,24 @@ import Wasp.SemanticVersion.VersionBound
 -- It represents a single version constraint in a range.
 -- See: https://github.com/npm/node-semver#ranges
 data Comparator
-  = PrimitiveComparator PrimitiveOperator PartialVersion
-  | BackwardsCompatibleWith PartialVersion
-  | ApproximatelyEquivalentTo PartialVersion
-  | XRange PartialVersion
-  | HyphenRange PartialVersion PartialVersion
+  = -- | =1.2.3, >1.2.3, <1.2.3, >=1.2.3, <=1.2.3
+    PrimitiveComparator PrimitiveOperator PartialVersion
+  | -- | ^1.2.3
+    BackwardsCompatibleWith PartialVersion
+  | -- | ~1.2.3
+    ApproximatelyEquivalentTo PartialVersion
+  | -- | X, 1.X, 1.2.X (can use x/X/*)
+    XRange PartialVersion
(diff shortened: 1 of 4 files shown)


waspc/src/Wasp/SemanticVersion/ComparatorSet.hs
```diff
@@ -0,0 +1,108 @@
...
+gte :: Version -> ComparatorSet
+gte = mkPrimCompSet GreaterThanOrEqual
+
+eq :: Version -> ComparatorSet
+eq = mkPrimCompSet Equal
+
+backwardsCompatibleWith :: Version -> ComparatorSet
+backwardsCompatibleWith = ComparatorSet . pure . BackwardsCompatibleWith . fromVersion
+
+approximatelyEquvivalentTo :: Version -> ComparatorSet
```
FranjoMindek, line 70: The only new thing here is the helpers for new comparators:
- `approximatelyEquvivalentTo`
- `xRange`
- `hyphenRange`

And the parser.

waspc/src/Wasp/SemanticVersion/Range.hs
```diff
@@ -0,0 +1,74 @@
...
+  where
+    doesVersionRangeAllowOnlyMinorChanges versionRange = isJust $ do
+      let versionInterval = versionBounds versionRange
+      let lowerBound = fst versionInterval
+      lowerBoundVersion <- versionFromBound lowerBound
+      let noMajorChangesInterval =
+            (lowerBound, Exclusive $ nextBreakingChangeVersion lowerBoundVersion)
+      guard $ versionInterval `isSubintervalOf` noMajorChangesInterval
+
+parseRange :: String -> Either ParseError Range
```
FranjoMindek: This parser is the only new thing.

waspc/src/Wasp/Node/Internal.hs
FranjoMindek, line 17: Why not use the new range parser (or partial version parser) here?
Well `node` outputs the SemVer version. So `versionParser` is enough.

FranjoMindek:
Moved here https://github.com/wasp-lang/wasp/pull/3917

Issue #3768, opened 2026-02-13
Documentation plan for TS spec transition


sodic:
We must decide out how to handle docs when we switch to the TS spec.

Options:
- Clean cut: new version docs show only TS spec, older versioned docs keep the DSL examples.
- Show DSL and TS spec side by side during the coexistence period if there is one (check #3764).

We should also consider an API reference for the TS spec (#1401).

cprecioso:
- API reference implemented at https://github.com/wasp-lang/wasp/pull/4169
- Docs for TS Spec itself + migration guides at https://github.com/wasp-lang/wasp/pull/4157

cprecioso:
Done in #4157

sodic:
I think this was closed by accident. This issue is about migrating all our docs to the TS spec. I don't believe the mentioned PR does that.

@cprecioso Please double check

infomiho:
I believe it is closed by #4191 

If you open https://wasp-docs-on-main.pages.dev/ you won't find any examples in latest docs written in DSL.

Issue #3642, opened 2026-01-21
Add CLI e2e tests for TS spec


sodic:
If we're serious about swtiching to the TS spec (and we are), we should cover it with CLI e2e tests. All current tests deal only with our DSL.

Previous attempt: https://github.com/wasp-lang/wasp/pull/2428

infomiho:
Moved with recent PRs #4154 and #4175

Issue #2390, opened 2024-11-28
Rework the TS setup command for `wasp` TS config


sodic:
When we released the Wasp TS config as an experimental feature, we added a `wasp ts-setup` command to prepare the project for using the TS config.

Currently (as of 0.15.2), the command doesn't do much.

To properly set up a project that uses the TS config, it's necessary to go through the steps described in [the docs](https://wasp-lang.dev/docs/general/wasp-ts-config), and `wasp ts-setup` currently only covers two of them.

All the steps are:
- Rename and create new `tsconfig.json` files.
- Potentially add `type: module` to package.json.
- Rename `main.wasp` to something else.
- Run `wasp ts-setup`
  - It adds the `wasp-config` package to your package.json's `devDependencies`.
  - It then runs `npm install`
- Create an empty `main.wasp.ts` file and start writing inside.

What we should do:
- Let new users choose what they want to use (Wasp or TS) when creating a new project.
- Let old users migrate more easily, by improving the `wasp ts-setup` command to do everything or by detecting that the user wants to use the TS SDK by looking for `main.wasp.ts` (which is tricky due to the way `wasp start` is set up).

sodic:
Let's say:
- Most optimistic is 25 hours (5h research, 15h implementation, 5h fixes).
- Most pesimistic is 75 hours (15h reserach, 45h implementation, 15h fixes).

We want to make the optimistic one happen so cut off time if necessary.

sodic:
I spent a day researching this.

The strategy we're going with is:
  - Ship `wasp-config` (the spec package) together with the CLI and "install it" like we install the SDK but earlier (as the first step of Wasp start). We're not shipping it through NPM because we don't want to force users to update the config package every time they update the CLI - those two are very coupled.
  - Include the genreated code (e.g., entity names) through interface module augmentation. This is the most popular modern solution (sources: SST, Next, TanStack).
  - Save the augmented interfaces to either the hidden `.wasp` dir (like Next and Nuxt) or expose them to the user (like TanStack). 

Besides this strategy, I considered making an NPM package wrapper around code generated in `node_modules`, like Prisma did before version 7.
Prisma is moving away from this and we decided not to implement it because:
  - The TS language server doesn't watch `node_modules` -> stale types -> frequent LSP restarts.
  - NPM assumes full control over `node_modules`, we want to avoid stepping on its toes.
[...]

PR #3861, opened 2026-03-04
Install `wasp-config` from  `.wasp`

sodic:
## Description

This PR is a part of #2390.

It exchanges the ugly, non-portable global installation path for `wasp-config` for a slighly-less-ugly portable local installation one.

### How it works
Before, the `ts-setup` command:
1. Added a `file:.../.../../wasp-config` path to `package.json`.
2. Called `npm install`.

Now, the `ts-setup` command:
1. Copies `wasp-config` to `.wasp`.
2. Calls `npm install`.

There's no need to add `wasp-config`'s relative location to `pacakge.json` since `wasp new` already put it there (https://github.com/wasp-lang/wasp/pull/3862). 

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

  - No tests yet. They'll be a part of #3642.

- 📜 Documentation:
  - Docs updates are a part of #3768

If any of those doesn't happen, we have a contingency plan: https://github.com/wasp-lang/wasp/issues/3864

### DIFF
--- a/waspc/ChangeLog.md
+++ b/waspc/ChangeLog.md
@@ -2,14 +2,23 @@
 
 ## 0.22.0
 
+### ⚠️ Breaking Changes
+
+Remember to check out the [migration guide](https://wasp.sh/docs/migration-guides/migrate-from-0-21-to-0-22) for step-by-step documentation on how to upgrade.
+
+- Upgraded the Zod version used for env validation to Zod v4. If you use custom env validation schemas, you may need to update them to be compatible with the latest Zod API. ([#3879](https://github.com/wasp-lang/wasp/pull/3879))
+
+### 🎉 New Features
+
+- The Wasp TS config file (`*.wasp.ts`) now supports async logic in the default export, enabling use cases like file-based routing and dynamic configuration. ([#3900](https://github.com/wasp-lang/wasp/pull/3900))
+
+### 🐞 Bug fixes
+
+- Projects created with `wasp new` were missing their `.gitignore` file. ([#3870](https://github.com/wasp-lang/wasp/pull/3870))
+
 ### 🔧 Small improvements
 
 - The `wasp-config` package is now installed from a local copy inside the project (`.wasp/wasp-config/`) instead of directly from the global Wasp data directory. This makes projects more self-contained and avoids issues with npm resolving paths outside the project. ([#3861](https://github.com/wasp-lang/wasp/pull/3861))
-
-## 0.21.2
-
-### 🔧 Small improvements
-
 - `REACT_APP_API_URL` is now **required** when building the client for production (it no longer defaults to any value). In development mode, it still defaults to the local server URL. This prevents silently broken production deployments where the client can't reach the server. ([#3740](https://github.com/wasp-lang/wasp/pull/3740))
 - The `wasp()` Vite plugin now enforces certain Vite config options that Wasp requires (e.g. `base`, `envPrefix`, `build.outDir`) and throws an error if the user tries to override them. ([#3771](https://github.com/wasp-lang/wasp/pull/3771))
 - Upgraded our internal `nodemailer` dependency to v8.0.1 to address [a possible security issue](https://github.com/advisories/GHSA-rcmh-qjqh-p98v). ([#3756](https://github.com/wasp-lang/wasp/pull/3756))


waspc/cli/src/Wasp/Cli/Command/CreateNewProject/StarterTemplates/Templating.hs
```diff
@@ -49,9 +52,9 @@ replaceTemplatePlaceholdersInPackageJsonFile appName projectName projectDir =
 
 replaceTemplatePlaceholdersInFileOnDisk :: NewProjectAppName -> NewProjectName -> Path' Abs (File f) -> IO ()
 replaceTemplatePlaceholdersInFileOnDisk appName projectName file = do
-  waspConfigPackagePath <- getPackageInstallationPath WaspConfigPackage
+  let waspConfigPackageSpecifier = getPackageJsonSpecifierForPackage WaspConfigPackage
   let waspTemplateReplacements =
-        [ ("__waspConfigPath__", waspConfigPackagePath),
+        [ ("__waspConfigPackageSpecifier__", waspConfigPackageSpecifier),
```
sodic, line 57: The name "specifier" comes from NPM.

waspc/cli/src/Wasp/Cli/Command/TsConfigSetup.hs
```diff
@@ -1,38 +1,20 @@
...
-installWaspConfigPackage :: Chan J.JobMessage -> Path' Abs (Dir a) -> IO (Either String ())
-installWaspConfigPackage chan projectDir = do
-  installationPath <- getPackageInstallationPath WaspConfigPackage
-  (_, exitCode) <-
-    concurrently
-      (readJobMessagesAndPrintThemPrefixed chan)
-      (runNodeCommandAsJob projectDir "npm" ["install", "--save-dev", "file:" ++ installationPath] J.Wasp chan)
-  return $ case exitCode of
-    ExitSuccess -> Right ()
-    ExitFailure _ -> Left "Failed to install wasp-config package"
```
sodic, line 38: This part is now handled through starter templates and `wasp new`.

waspc/src/Wasp/NodePackageFFI.hs
```diff
@@ -87,11 +90,27 @@ getPackageProcessOptions package args = do
...
-  waspDataDir <- Data.getAbsDataDirPath
-  let absPackagePath = waspDataDir </> packagesDirInDataDir </> installablePackageDirInPackagesDir package
-  return $ fromAbsDir absPackagePath
+getPackageJsonSpecifierForPackage :: InstallablePackage -> String
+getPackageJsonSpecifierForPackage package =
+  "file:" ++ fromRelDir (getPackageInstallationPathInProject package)
+
+getInstallablePackageName :: InstallablePackage -> String
+getInstallablePackageName = \case
+  -- NOTE: These names must match the 'name' fields in packages' package.json files.
```
sodic, line 99: This is the only bit of duplication I couldn't remove.

waspc/src/Wasp/NodePackageFFI.hs
```diff
@@ -87,11 +90,27 @@ getPackageProcessOptions package args = do
...
+  WaspConfigPackage -> "wasp-config"
+
+ensurePackageIsAtInstallationPathInProject :: Path' Abs (Dir WaspProjectDir) -> InstallablePackage -> IO ()
+ensurePackageIsAtInstallationPathInProject projectDir package = do
+  let dstPackageDirInProject = projectDir </> getPackageInstallationPathInProject package
+  unlessM (IOUtil.doesDirectoryExist dstPackageDirInProject) $ do
```
sodic: If I leave this check, the project will break if the user updates Wasp and doesn't do `wasp clean`. Should I remove the check and always copy `wasp-config`?

waspc/data/Cli/starters/ts-minimal/package.json
```diff
@@ -13,6 +13,6 @@
     "prisma": "5.19.1",
     "typescript": "5.8.2",
     "vite": "^7.0.6",
-    "wasp-config": "file:.wasp/wasp-config"
+    "wasp-config": "__waspConfigPackageSpecifier__"
```
infomiho, line 16: Is this a new starter? Will all other starters need to include this? I guess Open SaaS won't be able to use the placeholder, do we want to account for that scenarios i.e. add some validation?

waspc/src/Wasp/NodePackageFFI.hs
infomiho: - I guess we could go and be more precise, checking the installed version of `wasp-config`?
- Or we install it every time.

I dislike the fact that user uses Wasp normally, then updates the CLI version and then everything breaks. _Unless_ they do the clean, which feels like not the usual flow.

waspc/src/Wasp/NodePackageFFI.hs
```diff
@@ -60,8 +63,8 @@ runnablePackageDirInPackagesDir = \case
   WaspStudioPackage -> [reldir|studio|]
 
 installablePackageDirInPackagesDir :: InstallablePackage -> Path' (Rel PackagesDir) (Dir PackageDir)
-installablePackageDirInPackagesDir = \case
-  WaspConfigPackage -> [reldir|wasp-config|]
+installablePackageDirInPackagesDir package = case package of
+  WaspConfigPackage -> fromJust $ parseRelDir $ getInstallablePackageName package
```
infomiho: The `case` here (and in `getPackageInstallationPathInProject`) is redundant; both single-branch cases just delegate to `getInstallablePackageName`. They can drop the `case` and call it directly.

On the broader type class question: not worth it while `InstallablePackage` is a closed, internal type with one variant. You'd trade [GHC exhaustiveness warnings](https://blog.janestreet.com/what-do-haskellers-have-against-exhaustiveness/) (free with the sum type) for [open-world flexibility](https://wiki.haskell.org/OOP_vs_type_classes) you don't need.

waspc/data/Cli/starters/ts-minimal/package.json
sodic: We'll have validation, yes. There's a task for it :)

As for what is this new starter, check #3862.

waspc/src/Wasp/NodePackageFFI.hs
sodic: > The case here (and in getPackageInstallationPathInProject) is redundant; both single-branch cases just delegate to getInstallablePackageName. They can drop the case and call it directly.

Yes, but I wanted to communicate the intentionality: `installablePackageDirInPackagesDir` doesn't need to be identical to the package's name, there's no semantic tie between those two. But for this particular pacakge, they are equal.

I think I'll still drop the case though, not that important.

waspc/src/Wasp/NodePackageFFI.hs
sodic: I'll change the code to install it every time for now. We do something smarter in the next PR.

PR #3872, opened 2026-03-05
Move Vitest to user-land

### LINKED ISSUE #3735: Move Vitest to user-land
Since Vite setup is migrated to user-land, Vitest migration to user-land is now possible. The main motivation to move Vitest to user-land is that users will be able to use it for **server code tests as well.**

I've done it already once in Vite POCs (https://github.com/wasp-lang/wasp/pull/3584) and it required:
- removing all testing deps from the SDK
- removing all Vitest setup from `wasp:config` plugin
- remove the testing helpers from the SDK and documenting them in the docs

infomiho:
## Description

- Moves Vitest and all testing dependencies out of the Wasp SDK into user-land.
- Previously, Wasp bundled Vitest, auto-injected test configuration, and provided `renderInContext`/`mockServer` helpers, but this locked users into a specific Vitest version and installed testing deps for everyone.
- Now, users install Vitest themselves, configure it in `vite.config.ts`.
- The `wasp/client/test` export is slimmed down to just `WaspTestWrapper`, with other helpers documented as copy-paste in the Vitest guide.
- Removes the `wasp test client` CLI command.

Fixes #3735

## Type of change

- [ ] **🐞 Bug fix**
- [x] **💥 Breaking change**

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

- 🆕 Changelog:


waspc/data/Generator/templates/sdk/wasp/server/env.ts
```diff
@@ -170,10 +170,12 @@ const serverEnvSchema = z.discriminatedUnion('NODE_ENV', [
 
 const defaultNodeEnvValue = serverDevSchema.shape.NODE_ENV.value;
 const { NODE_ENV: inputNodeEnvValue, ...restEnv } = process.env;
+// Vitest sets NODE_ENV to 'test', which isn't a valid discriminator value. Treat it as 'development'.
+const nodeEnvValue = inputNodeEnvValue === 'test' ? 'development' : inputNodeEnvValue;
```
infomiho, line 174: Do we do this or ask users to set `NODE_ENV` to `development` in tests?

waspc/data/Generator/templates/sdk/wasp/client/test/WaspTestWrapper.tsx
```diff
@@ -0,0 +1,12 @@
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
+import { BrowserRouter } from 'react-router'
+import type { ReactNode } from 'react'
+
+// PUBLIC API
+export function WaspTestWrapper({ children }: { children: ReactNode }) {
+  return (
+    <QueryClientProvider client={new QueryClient({ defaultOptions: { queries: { retry: false } } })}>
```
infomiho, line 8: Not sure if we need `retry: false`, it was an addition by Claude, we didn't have it before.

waspc/data/Generator/templates/sdk/wasp/client/vite/plugins/detectServerImports.ts
```diff
@@ -4,13 +4,19 @@ import path from 'path'
...
+  let isTest = false
   return {
     name: 'wasp:detect-server-imports',
     enforce: 'pre',
     configResolved(config) {
       parsePathToUserCode = createPathToUserCodeParser(config.root)
+      isTest = config.mode === 'test'
     },
     resolveId(source, importer) {
+      if (isTest) {
```
infomiho, line 16: Running server tests with Vitest triggers this check which stops users from test server code, so I disabled it for tests.

waspc/data/Cli/starters/basic/vite.config.ts
```diff
@@ -7,4 +8,13 @@ export default defineConfig({
   server: {
     open: true,
   },
+  test: {
```
infomiho, line 11: Do we want to include unit tests set up with the `basic` template? There is still `minimal` without them.

web/docs/guides/libraries/vitest.md
```diff
@@ -0,0 +1,207 @@
...
+  },
+  // highlight-start
+  test: {
+    globals: true,
+    environment: 'jsdom',
+    setupFiles: ['./src/test/setup.ts'],
+    exclude: ['**/node_modules/**', '**/.wasp/**'],
+    env: {
+      DATABASE_URL: 'postgresql://localhost/test',
+    },
```
infomiho, line 52: Do we want to do this or use maybe `vi.stubEnv` in the `src/test/setup.ts`? I'm not really an expert on what is the best way to set env vars for tests. I'm not sure if we can use `.env.test` since we have our own special env vars handling for `.env.client`.

cprecioso:
@infomiho didn't do the actual review yet, but I already looked over the changes and I don't agree too much 😅

Specifically, I think we're putting back too much work on the hands of our users, for things that we already know all about:
- I'd keep the `renderInContext` and `mockServer` functions. In the regular app, we own the render entry and the URL routes, so it feels weird that now we leave it up to the user to have to mock their internals instead of giving them an high-level helper for it.
- Similarly, I'd keep `wasp test client` as a command that just (1) checks vitest is installed and (2) calls `npx vitest`, for the same reason that we don't let the user directly do `npx vite` for the most part. This one I feel less strongly about.
- And if we could keep a default `setup.ts` for tests and automatically add it, I think that'd also be good.

I think this PR should focus on **adding more choice** instead of **adding more boilerplate**. e.g.: unlink Wasp from any specific Vitest, and making sure that Vitest can be called without our helpers, but keep breaking changes mostly to 0.

infomiho:
> I'd keep the renderInContext and mockServer functions. In the regular app, we own the render entry and the URL routes, so it feels weird that now we leave it up to the user to have to mock their internals instead of giving them an high-level helper for it.

Providing those higher-level helpers requires us to shipping specific testing libraries with specific versions. I wanted to avoid maintaining those helpers as they'll need to be updated over time. Testing being a user-land concern frees users from using _our versions_ and frees us from playing catch up. 

> Similarly, I'd keep wasp test client as a command that just (1) checks vitest is installed and (2) calls npx vitest, 

I'd avoid having `wasp test client` specifically since Vitest can be used for both client and server tests. Having `wasp test` makes more sense to me, but again, this would mean we provide the whole Vitest machinery. I can't see how we can _half support_ Vitest and provide nice helpers. I guess we either move Vitest out of Wasp, or don't. I'm more for moving it out of Wasp for reasons mentioned above :)

> for the same reason that we don't let the user directly do npx vite for the most part. This one I feel less strongly about.

Users build the client for production by calling `npx vite` themselves in the root of the project, check: https://wasp.sh/docs/deployment/deployment-methods/paas#deploying-the-client

> And if we could keep a default setup.ts for tests and automatically add it, I think that'd also be good.

This is all provided that people will want to _our test setup_. I believe there are many ways to set up testing in a JS app (of course 90% of people use Vitest) but I wanted to give the users a choice. 

### Ways to avoid the boilerplate

1. Write the helpers in a better way? I'm sure if I can, because then we have to depend on certain libs being available.
2. Provide the boilerplate in our starters like Open SaaS and `basic` starter?

cprecioso:
Regarding depending on specific libraries being installed, can't we have the helpers `peerDepend` on the dependencies?

infomiho:
> can't we have the helpers peerDepend on the dependencies?

Wouldn't that install them always since for Typescript to pass, it needs those deps? Maybe I'm misunderstanding something here 😄 I don't want to install 5, 6 testing deps if I'm not using tests at all.

Issue #3625, opened 2026-01-16
Run more extensive tests on npm-packaged binaries


cprecioso:
## Context

Currently, our CI tests that the npm package binaries download, install, and run correctly, but we only verify by running `wasp version`. The actual CLI e2e tests run on unpackaged binaries, not on the final packaged products.

This gap means packaging issues can slip through to release, as we recently experienced (#3585, [Discord](https://discord.com/channels/686873244791210014/1460004988842414233)). While the packaging code doesn't change often, when it does go wrong, users get a broken release from the start; and fixing it requires an urgent hotfix.

## Proposal

Extend testing of npm-packaged binaries beyond just `wasp version`. Two approaches to consider:

### Option A: Ad-hoc tests for packaged binaries

Write specific tests that exercise key functionality on the packaged binaries (e.g., `wasp new`, `wasp build`, basic project operations).

**Estimate:** ~1 week. Some infrastructure work needed, but manageable if we run them as CI jobs. Running them on Tasty would probably be more work and we might be better off then doing the second option.

### Option B: Run full CLI e2e test suite on packaged binaries

Run the existing CLI e2e tests against the packaged/release candidate binaries instead of (or in addition to) unpackaged ones.

**Estimate:** TBD during refinement. Likely way more work than Option A.

## Trade-offs

- **CI time impact:** The npm packager job depends on some of our slowest jobs and normally waits ~20 mins before it can start. Adding extensive tests here would increase overall CI times.
- **Frequency:** These tests would primarily run on release candidates, not on every commit.
- **Maintenance:** Option A requires maintaining separate test code; Option B reuses existing tests but needs more infrastructure.

cprecioso:
I am doing https://github.com/wasp-lang/wasp/pull/3873 as an approach which instead of running tests on the npm-packaged Wasp, just ensures that the binary works and the data files are identical to what the in-tree Wasp has.

PR #3870, opened 2026-03-05
Fix .gitignore stripped by npm during skeleton packaging

cprecioso:
## Description

The skeleton template's `.gitignore` file was being stripped by npm during `npm publish`, preventing new Wasp projects from getting the `.gitignore` file. This happened because npm treats `.gitignore` files as packaging directives rather than inert content.

The fix follows the existing pattern used elsewhere in the codebase for dotfiles (e.g., `dockerignore`, `server/gitignore`, `server/npmrc`): rename the file to drop the leading dot (`gitignore`), then restore it during project creation.

## Type of change

- [x] **🐞 Bug fix** 

## Checklist

- [x] I tested my change in a Wasp app to verify that it works as intended.

- 🧪 Tests and apps:

  - [ ] I added **unit tests** for my change. 
  - [ ] _(if you fixed a bug)_ I added a **regression test** for the bug I fixed. 
  - [ ] _(if you added/updated a feature)_ I added/updated **e2e tests** in `examples/kitchen-sink/e2e-tests`.
  - [ ] _(if you added/updated a feature)_ I updated the **starter templates** in `waspc/data/Cli/templates`, as needed.
  - [ ] _(if you added/updated a feature)_ I updated the **example apps** in `examples/`, as needed.
    - [ ] _(if you updated `examples/tutorials`)_ I updated the tutorial in the docs (and vice versa).

- 📜 Documentation:

  - [ ] _(if you added/updated a feature)_ I **added/updated the documentation** in `web/docs/`.

- 🆕 Changelog: _(if change is more than just code/docs improvement)_
  - [x] I updated `waspc/ChangeLog.md` with a **user-friendly** description of the change.
  - [ ] _(if you did a breaking change)_ I added a step to the current **migration guide** in `web/docs/migration-guides/`.
  - [ ] I **bumped the `version`** in `waspc/waspc.cabal` to reflect the changes I introduced.

Fixes #3850

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3873, opened 2026-03-06
Add data files verification to build workflow

cprecioso:
- Resolves https://github.com/wasp-lang/wasp/issues/3625

## Description

We've found differences between the in-tree Wasp version and the npm-packaged one (https://github.com/wasp-lang/wasp/pull/3870). But we already test the in-tree wasp binaries in the build workflows. Running these same tests on the npm-packaged binaries would be wasteful, and would make a big part of our CI wait on one of our longest jobs instead of parallelizing.

Instead, I opted for a simpler solution where npm-package tests get tested in two ways:
- (Already existed) Installing the npm package and calling `wasp version` -> Checks that the binary is included and works.
- (New) Compares that the list of data files of installed npm packages are the same as they should be according to the `waspc.cabal` file.

With these two tests, we can keep testing with the in-tree Wasp, because we're just ensuring that the binary works and that it will find the same files in the same places.

## Changes

Adds a step in the Wasp build workflow that outputs a sorted list of all data files with per-file SHA-256 hashes by running `cabal sdist --list-only` and computing a hash for each file. This list is passed to the npm package test workflow, which verifies that each platform-specific npm package's `./data` folder contains exactly the expected files with matching content hashes.

This ensures consistency between what Cabal considers part of the source distribution and what actually ends up in the npm packages, catching any mismatches in file inclusion early.

## Type of change

- [ ] **🐞 Bug fix**
- [ ] **💥 Breaking change**

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

- 🆕 Changelog:


waspc/src/Wasp/Project/ExternalConfig/ViteConfig.hs
```diff
@@ -0,0 +1,67 @@
+{-# LANGUAGE QuasiQuotes #-}
```
infomiho, line 1: I think you might need to rebase, these are some changes on `main`

.github/workflows/ci-npm-package-test.yaml
```diff
@@ -85,3 +90,99 @@ jobs:
...
+          os=$(echo "$NPM_TARGET" | jq -r '.os')
+          cpu=$(echo "$NPM_TARGET" | jq -r '.cpu')
+          libc=$(echo "$NPM_TARGET" | jq -r '.libc // "unknown"')
+
+          pkg_name="@wasp.sh/wasp-cli-${os}-${cpu}-${libc}"
+          pkg_dir="./packages/wasp-sh-wasp-cli-${os}-${cpu}-${libc}"
+
+          # Pack the sub-package
+          mkdir packed-tarballs
+          tarball=$(npm pack --pack-destination ./packed-tarballs "$pkg_dir")
```
infomiho: We are recreating the process of packing the tarball, can we somehow use the _already packed_ tarballs which will be published to perform these tests? That would make it less repetitive and IMHO more relevant.

cprecioso: You're right. I set this workflow after the publishing workflow so that it uses the `pkg.pr.new` packages

cprecioso:
@infomiho ready for re-review

PR #3877, opened 2026-03-09
Fix ansi formatting in validation

FranjoMindek:
## Description

Fixes multiline env validation errors only coloring the first line, caused
by Wasp reading child process output line-by-line, which breaks ANSI codes spanning multiple lines.

But before that, the old code used the `getColorizedConsoleFormatString` function incorrectly, it just preprended colored `%s` to existing string:
image
image

Changes:
- Replace `getColorizedConsoleFormatString` (format-string approach) with
`colorize(colorKey, text)` that wraps each line with its own ANSI open/close
codes
- Skip ANSI colorization in browser contexts (`globalThis.window`) where it
doesn't work

image
image

## Type of change

- [x] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

### DIFF
--- a/waspc/data/Generator/templates/sdk/wasp/env/validation.ts
+++ b/waspc/data/Generator/templates/sdk/wasp/env/validation.ts
@@ -11,8 +11,7 @@ export function ensureEnvSchema<Schema extends z.ZodType>(
   if (result.success) {
     return result.data;
   } else {
-    const errorMessage = formatZodEnvError(result.error);
-    console.error(colorize("red", errorMessage));
+    console.error(colorize("red", formatZodEnvError(result.error)));
     throw new Error("Error parsing environment variables");
   }
 }


waspc/data/Generator/templates/sdk/wasp/env/validation.ts
```diff
@@ -13,8 +11,13 @@ export function ensureEnvSchema<Schema extends z.ZodTypeAny>(
   if (result.success) {
     return result.data;
   } else {
-    console.error(`${redColorFormatString}${formatZodEnvErrors(result.error.issues)}`);
-    throw new Error('Error parsing environment variables');
+    const errorMessage = formatZodEnvErrors(result.error.issues);
+    if (globalThis.window) {
+      console.error(errorMessage);
```
FranjoMindek: Browser message is already red because of  `console.error`.

But yes, browser console does not support ANSI formatting so we don't use it.

FranjoMindek: Other alternative is to hide this logic inside of `colorize` but that seemed weird. I.e. selectively apply ANSI coloring if applicable environment.

waspc/data/Generator/templates/sdk/wasp/universal/ansiColors.ts
```diff
@@ -1,14 +1,30 @@
...
+ * Each line is individually wrapped because Wasp reads child process
+ * output line-by-line and re-prints it with a prefix (e.g. `[ Server ]`).
+ * A single color code spanning multiple lines would only color the first line.
+ *
+ * @example
+ * ```typescript
+ * console.log(colorize('red', 'This is red text'));
+ * ```
+ *
+ * @internal This is a private API for: SDK, client.
```
FranjoMindek, line 15: `@internal` is keyword that `tsc` can use:

Image

https://www.typescriptlang.org/docs/handbook/compiler-options.html

cprecioso: But we don't have `stripInternal` in the SDK :D

waspc/data/Generator/templates/sdk/wasp/env/validation.ts
cprecioso: I think it's fine to hide the logic inside colorize. It'd feel weird to have to do this check everywhere you want to color stuff. You can check process.stderr.hasColors() to know whether to output colors or not.

And, if you want to drop in this rabbit hole, you could detect when we're in the browser and use CSS formatting in that env: https://developer.chrome.com/docs/devtools/console/format-style#style-specifier

Actually I just saw ANSI is now also supported in the Chrome console so you might not need to gate anything at all: https://developer.chrome.com/docs/devtools/console/format-style#style-ansi (do test it out in Safari and Firefox too)

waspc/data/Generator/templates/sdk/wasp/universal/ansiColors.ts
cprecioso: Remember to add to the Wasp API sheet https://docs.google.com/spreadsheets/d/1QW-_16KRGTOaKXx9NYUtjk6m2TQ0nUMOA74hBthTH3g/edit?pli=1&gid=2143046114#gid=2143046114

FranjoMindek: I know, just thought it would be good to start using it :).

waspc/data/Generator/templates/sdk/wasp/env/validation.ts
FranjoMindek: Not in firefox, can confirm. Used that for dev.

Will move the logic inside, makes sense.

FranjoMindek: > check process.stderr.hasColors() to know whether to output colors or not

Oh this is a deep deep rabbit hole.
First I have to detect if `process` even exists (it does not in browser).
So for browser this does not help.
If we want to color it in browsers which support it (chrome) we must do separate check there.

Second problem is we don't know if the caller is using `.log` or `.error` so stream specific stuff shouldn't be used.

I'll try to get something good enough without going too hard into it.

FranjoMindek: Funnily, our server has neither `isTTY` or `hasColors()`

cprecioso: ooh feel free to skip if it's too deep of a rabbit hole then :D

FranjoMindek:
Working on chrome now:
image

waspc/data/Generator/templates/sdk/wasp/env/validation.ts
FranjoMindek: It's due to Wasp.
I did it in simple way.
Browser yes if chrome.
Otherwise yes if node unless `NO_COLOR` is set.

PR #3891, opened 2026-03-09
Make react-router routes lazy-loaded

### LINKED ISSUE #3892: [Deep Freeze] Pages are loaded async (Optional)
## Problem

Currently, the output bundle contains all of the code for all pages in a single file, which grows large quickly. This means that even when a page is prerendered and the user sees HTML instantly, the full JS bundle still has to load for hydration — negating some of the performance benefit of SSR.

## Solution

Use [React Router's facilities for async loading](https://reactrouter.com/start/data/route-object#lazy) so that each page gets its own JS bundle. When a user visits a prerendered page, only that page's JS loads for hydration, not the entire app.

Vite handles the code splitting automatically once routes are marked as `lazy`.

## Relationship to SSR

This is complementary to static prerendering but independent:
- **SSR** gives fast HTML (user sees content immediately)
- **Code splitting** reduces how much JS loads for hydration (page becomes interactive faster)

Both together give the best result, but either can be done independently. This issue is optional for the initial Phase 1 release.

cprecioso:
- Fixes #3892

## Description

Converts the react-router route generation from eager imports to lazy-loaded routes using react-router's `lazy` property. Each page component now becomes its own code-split chunk, loaded on-demand as users navigate instead of being bundled upfront.

This is controlled by a new `lazy: <boolean>` field on the `route` declaration. It is optional and `true` by default.

I documented this in a new advanced routing page in the docs, and added some advanced features from RR we already support, so it's not so empty.

---

The result is reducing the initial load time of Wasp applications, especially those with many pages.

<details>
<summary>Example from OpenSaaS.sh</summary>

(this is not an ideal app as it doesn't have many pages, so the savings are modest)

<table>
<thead>
<tr>
 <th>
 <th scope=col>Before
 <th scope=col>After
</thead>
<tbody>
<tr>
 <th scope=row>Lighthouse
 <td>image
 <td>image
<tr>
 <th scope=row rowspan=2>Network tab
 <td>

- 8 / 37 requests
- 551 kB / 2,713 kB transferred
- 1,514 kB / 3,722 kB resources
- Finish: 439 ms
- DOMContentLoaded: 284 ms
- Load: 288 ms

 <td>

- 12 / 41 requests
- 271 kB / 2,491 kB transferred
- 856 kB / 3,077 kB resources
- Finish: 159 ms
- DOMContentLoaded: 64 ms
- Load: 67 ms

<tr>
	<td colspan=2>

_(In the `X / Y` data points, `X` is everything except images, and `Y` is everything including images.)_

</tbody>
</table>

</details>

<details>
<summary>Example from community member app (56 pages)</summary>

<table>
<tr>
 <td>
 <th scope=col>Before
 <th scope=col>After
<tr>
 <th scope=row>Lighthouse
 <td>image
 <td>image
<tr>
 <th scope=row>Lighthouse Treemap
 <td>image
 <td>image
</table>

</details>

## Type of change

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:


waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/routes.tsx
```diff
@@ -4,14 +4,21 @@
...
-{=# pagesToImport =}
-{=& importStatement =}
-{=/ pagesToImport =}
-
 export const routesMapping = {
   {=# routes =}
-  {= name =}: {= targetComponent =},
+  {= name =}: async () => {
+    {=# isDefaultExport =}
+    const { default: Component } = await import('{=& importPath =}')
```
infomiho: It would be awesome if we could upgrade our `JsImport` to support `show` as async import. So you could reuse maybe the `extToJson...` helpers?

examples/kitchen-sink/main.wasp
```diff
@@ -349,6 +349,18 @@ page ChatPage {
...
 }
 
+// --------- Lazy Loading --------- //
+
+route EagerRoute { path: "/lazy/no", to: EagerPage, lazy: false }
+page EagerPage {
+  component: import { EagerPage } from "@src/features/lazy-loading/pages/EagerPage"
+}
+
+route LazyRoute { path: "/lazy/yes", to: LazyPage, lazy: true }
```
infomiho: We are not testing the default value here, maybe drop `lazy: true` so it will break if our default changes?

waspc/data/Generator/templates/sdk/wasp/client/app/components/WaspApp.tsx
```diff
@@ -11,7 +11,7 @@ import { WebSocketProvider } from '../../webSocket/WebSocketProvider'
 
 export type WaspAppProps = {
   rootElement?: React.ReactNode;
-  routesMapping: Record<string, React.ComponentType>;
+  routesMapping: Record<string, { lazy: () => Promise<{ Component: React.ComponentType }> } | { Component: React.ComponentType }>;
```
infomiho: Is it worth creating a `RouteMapping` type and re-use in all the places?

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/routes.tsx
infomiho: @cprecioso what about this comment?

web/docs/advanced/routing.md
```diff
@@ -0,0 +1,89 @@
...
+  // Visiting /files/docs/report.txt → filePath = "docs/report.txt"
+  return <div>File: {filePath}</div>
+}
+```
+
+Read more in the [React Router docs on splats](https://reactrouter.com/7.12.0/start/data/routing#splats).
+
+## Lazy-Loaded Routes
+
+By default, Wasp lazy-loads all page routes using React Router's [`lazy`](https://reactrouter.com/how-to/code-splitting) property. This means each page's code is only downloaded when the user navigates to it, resulting in smaller initial bundle sizes -- especially useful for apps with many routes.
```
infomiho:
```suggestion
By default, Wasp lazy-loads all page routes using React Router's [`lazy`](https://reactrouter.com/how-to/code-splitting) property. This means each page's code is only downloaded when the user navigates to it, resulting in smaller initial bundle sizes - especially useful for apps with many routes.
```

Not sure if AI or not, but m-dashes look always like AI to me 😂

waspc/src/Wasp/Generator/JsImport.hs
```diff
@@ -54,6 +58,19 @@ jsImportToImportJson maybeJsImport = maybe notDefinedValue mkTmplData maybeJsImp
               "importIdentifier" .= jsImportIdentifier
             ]
 
+jsImportToDynamicImportJson :: JsImport -> Aeson.Value
```
infomiho: I don't see you using this helper? That's what I had in mind for cleaning up 👍

waspc/src/Wasp/Generator/SdkGenerator/Client/VitePlugin/VirtualModulesPlugin/VirtualRoutesG.hs
```diff
@@ -29,51 +34,69 @@ genVirtualRoutesTsx spec =
...
   where
     tmplPath = C.viteDirInSdkTemplatesDir </> virtualFilesFilesDirInViteDir </> [relfile|routes.tsx|]
+    routes = AS.getRoutes spec
+    eagerRoutes = filter (not . isRouteLazy . snd) routes
     tmplData =
       object
-        [ "routes" .= map (createRouteTemplateData spec) (AS.getRoutes spec),
-          "pagesToImport" .= map createPageTemplateData (AS.getPages spec),
+        [ "routes" .= map (createRouteTemplateData spec) routes,
+          "eagerImports" .= map (createEagerImportData spec) eagerRoutes,
```
infomiho: I found this split a bit arbitrary and like the data is not neatly organised. Maybe that's the way we have to do it. But I thought maybe we could think about the template data for routes as one `routes` concept which has:
```json
{
  "routes": [
    {
      "isLazy": true,
      "importStatement": ...
    },
    {
      "isLazy": false,
      ...
    }
  ]
}
```

Then in the template you iterate twice, outputting top level imports for `isLazy: false` and then below output async expression for `isLazy: true`.

I understand this is a bit hand wavy, but I just feel like this code can be cleaned up and with a bit more care it could be easier to follow along. This file suffers from being rewritten partially many times (I did myself recently, I'm guilty) but never fully rewritten with new fresh eyes.

waspc/e2e-tests/test-outputs/snapshots/wasp-build-golden/wasp-app/.wasp/out/web-app/build/index.html
```diff
@@ -11,7 +11,6 @@
 
         <title>wasp-app</title>
       <script type="module" crossorigin src="/assets/index.js"></script>
-      <link rel="stylesheet" crossorigin href="/assets/index.css">
```
cprecioso, line 14: Because the pages do not lay on the entry point anymore, their CSS files do not get eagerly imported either.

This affects the performance for the loading of code and assets. The waterfall of asset loading for an example page would change as follows:

shapes at 26-03-20 15 08 50

This does have the unfortunate implication of adding one step before loading the target files.

However, there is an extenuating circumstance: the CSS file that was loaded directly from the `index.html` didn't have much effect as there is no substantive content to be styled until the `<MainPage>` component is loaded.

And, it has an upside too: the tradeoff of loading files that only have the specific page requested is apparently already worth it for comparatively small pages such as OpenSaaS.sh [(see PR description on top)](https://github.com/wasp-lang/wasp/pull/3891#issue-4046492108)

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/routes.tsx
cprecioso: addressed in VirtualRoutesG.hs i think

waspc/src/Wasp/Generator/SdkGenerator/Client/VitePlugin/VirtualModulesPlugin/VirtualRoutesG.hs
cprecioso: Check the newest version, I think this is the best I got to between me and my friend Claude.

Issue #3624, opened 2026-01-16
Reconsider our policy about which external Wasp deps get included in user-level package.json


Martinsos:
As I was looking at what wasp deps we include in package.json, I was a bit confused about why React is included but Express not. I then started a discussion, and since there might be something to it, I am not capturing it here as an issue so we can consider it.

Here we (@infomiho) explain our thinking so far: https://github.com/wasp-lang/wasp/issues/2408 .

Here is a recent discussion on Discord: 

https://discord.com/channels/686873244791210014/1351908377533354025/1461302913778323582

PR #3900, opened 2026-03-10
Support async default export in *.wasp.ts files

### LINKED ISSUE #2573: Wasp TS  spec allow async logic
Let's say I want to do stuff in async manner in order to produce the final app spec -> how can I do it? @fecony said that right now he can't use async. Which makes sense since the default export we don't expect to be a Promise. Does that make sense or am I talking nonsense? Anyway, we should look into this. @sodic FYI.

sodic:
## Description

Fixes #2573 

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

  - [x] _(if you added/updated a feature)_ I added/updated **e2e tests** in `examples/kitchen-sink/e2e-tests` (waspello, not kitchen sink).

- 📜 Documentation:

### DIFF
--- a/examples/waspello/main.wasp.ts
+++ b/examples/waspello/main.wasp.ts
@@ -1,7 +1,10 @@
+import { readFile } from "fs/promises";
 import { ActionConfig, App, ExtImport } from "wasp-config";
 
+const title = (await readFile("appTitle.txt", "utf-8")).trim();
+
 const app = new App("waspello", {
-  title: "Waspello",
+  title,
   wasp: { version: "^0.21.2" },
 });
 
--- a/examples/waspello/package.json
+++ b/examples/waspello/package.json
@@ -22,6 +22,7 @@
   "devDependencies": {
     "@playwright/test": "1.55.1",
     "@tailwindcss/vite": "^4.1.18",
+    "@types/node": "^22.0.0",
     "@types/react": "^19.2.7",
     "@types/react-dom": "^19.2.3",
     "@wasp.sh/wasp-app-runner": "^0.0.13",
--- a/examples/waspello/appTitle.txt
+++ b/examples/waspello/appTitle.txt
@@ -1,0 +1,1 @@
+Waspello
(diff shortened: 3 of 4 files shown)


examples/waspello/package.json
```diff
@@ -22,6 +22,7 @@
   "devDependencies": {
     "@playwright/test": "1.55.1",
     "@tailwindcss/vite": "^4.1.18",
+    "@types/node": "^22.0.0",
```
sodic, line 25: How come we don't include this in the templates?

examples/waspello/main.wasp.ts
```diff
@@ -1,7 +1,10 @@
+import { readFile } from "fs/promises";
 import { ActionConfig, App, ExtImport } from "wasp-config";
 
+const title = (await readFile("appTitle.txt", "utf-8")).trim();
```
FranjoMindek, line 4: I would love to do this as part of `kitchen-sink` too, since it is a kitchen sink.

sodic: It can't since kitchen sink isn't in the TS spec yet.

examples/waspello/package.json
FranjoMindek: Not sure, I think Miho and Martin had discussion on what deps we include in user projects and what not. E.g. React so they don't confuse the verison which we use...

Personally, I don't know the decisions.

waspc/data/packages/wasp-config/src/appAnalyzer.ts
```diff
@@ -24,7 +24,7 @@ export async function analyzeApp(
 }
 
 async function getApp(mainWaspJs: string): Promise<Result<App, string>> {
-  const usersDefaultExport: unknown = (await import(mainWaspJs)).default;
+  const usersDefaultExport: unknown = await (await import(mainWaspJs)).default;
```
FranjoMindek, line 27: 👍🏻

examples/waspello/package.json
infomiho: Captured in this issue https://github.com/wasp-lang/wasp/issues/3624

PR #3902, opened 2026-03-10
Add --run and --exit flags to wasp-app-runner

cprecioso:
## Description

Add two new command-line flags to wasp-app-runner for automated testing and CI workflows:

- **`--run <cmd>`**: Waits for both client (port 3000) and server (port 3001) to be ready, then executes the given command with `$client_url` and `$server_url` environment variables set.
- **`--exit`**: Exits the process after both servers are ready (and after the `--run` command completes, if provided).

This allows us to do commands such as

```sh
$ run-wasp-app build --run 'npx lighthouse $client_url' --exit
```

## Type of change

- [ ] **🐞 Bug fix**
- [ ] **💥 Breaking change**

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

- 🆕 Changelog:


wasp-app-runner/src/build/index.ts
```diff
@@ -49,5 +55,23 @@ export async function startAppInBuildMode({
...
 
-  await wasp.proc.wait();
+  if (run) {
+    await new Process({
+      logger: createLogger("run"),
+      cmd: process.env.SHELL ?? "sh",
+      args: ["-c", run],
+      env: {
+        client_url: "http://localhost:3000",
+        server_url: "http://localhost:3001",
```
infomiho: Why the hard-coded URLs?

wasp-app-runner/src/build/index.ts
```diff
@@ -12,12 +14,16 @@ export async function startAppInBuildMode({
...
+  run,
+  exit,
 }: {
   waspCliCmd: WaspCliCmd;
   pathToApp: PathToApp;
   appName: AppName;
   dbType: DbType;
   dbImage: DockerImageName;
+  run?: string;
+  exit: boolean;
```
infomiho, line 25: `--run-cmd`?

`--exit-immediately` or `--no-wait`?

wasp-app-runner/src/dev/index.ts
```diff
@@ -34,5 +40,23 @@ export async function startAppInDevMode({
...
+    })
+      .print()
+      .wait();
+  }
+
+  if (exit) {
+    return;
+  } else {
+    await wasp.proc.wait();
+  }
```
infomiho: Feels ripe for DRYing up

infomiho (commented):
Left some comments

wasp-app-runner/src/build/index.ts
cprecioso: there's nowhere we can get them from, AFAIK?

wasp-app-runner/src/build/index.ts
cprecioso: What do you think is not being communicated with the current names?

infomiho: Both are pretty generic names, offering some alternatives that might be more specific. They are short and nice, but that's not usually _the metric_ we look for in names AFAIK.

wasp-app-runner/src/build/index.ts
infomiho: Then I misunderstood the command, you gave this example `run-wasp-app build --run 'npx lighthouse $client_url' --exit`. (Dramatic pause...) Now I get it, the point is to start the client and the server and they are by default available on those URLs. Makes sense.

We don't allow any customization that would render those values invalid? Do we want to encode the fact that these are the default URLs/ports somehow? So it's clear that these are not some stub URLs (that's what I assumed which confused me).

Issue #2481, opened 2025-02-03
Allow splitting `main.wasp.ts` into multiple files


sodic:
It's possible this works already but requires some massaging:
- https://dev.to/medardm/splitting-wasp-ts-config-2iei
- https://gist.github.com/medardm/263b2c933186eb1a496fa743c756edc7
- https://discord.com/channels/686873244791210014/1216264399552122922/1329931986286022697

In any case, we must come up with an official API and document it.

sontl:
Yeah, I came here to suggest the same enhancement but found this. Really looking forward to this feature in the next release!

developerzeke:
My ideas on this from the discord: https://discord.com/channels/686873244791210014/1378788346389729542/1379130408616071248

sodic:
Idea from the recent RFC:
```ts
// src/features/cards/wasp.ts
import { Feature } from "wasp-config";

export const cards = new Feature();

cards.query("getListsAndCards", { ... });
cards.action("createList", { ... });
cards.action("updateList", { ... });

// main.wasp.ts
import { cards } from "./features/cards/wasp";
import { payments } from "./features/payments";

const app = new App("waspello", { ... });
app.use(cards);
app.use(payments);

export default app;
```
We should dicusss this at length and see how it plays out with FSM led by @infomiho.

infomiho:
Notes based on two small prototypes I did today (`tsc` and `rolldown` based).

The pipeline looks like:

```txt
main.wasp.ts
  -> construct reachable TS graph
  -> ergonomic imports transform on reachable .wasp.ts files
  -> output runnable JS
  -> import JS and analyze App
  -> write decls.json
```

For both prototypes, I decided:

- `main.wasp.ts` remains the only root entrypoint.
- Only reachable `*.wasp.ts` files get Wasp semantics.
- Plain `.ts` / `.tsx` files stay normal helper files.
- Plain `.ts` / `.tsx` files cannot value-import `@src/*`.
-`*.wasp.ts` files that are not imported are ignored.

### `tsc` prototype

`tsc` outputs in `.wasp` dir the following file tree:

```txt
.wasp/
  decls.json
  ts-spec/
    main.wasp.js
    src/
      features/
        auth/
          auth.wasp.js
        tasks/
          schema.js
          tasks.wasp.js
      shared/
        appMeta.js
        routeNames.js
```

Pros:

- Reuses the existing TypeScript compiler path.
- Keeps TypeScript semantic checking in the same step as emit.

[...]

PR #3911, opened 2026-03-11
Add validation for other tsconfigs

### LINKED ISSUE #2426: Ensure config files in a Wasp project are set up correctly
We already have `tsconfig.json` and `package.json` validation, but there's more work to be done.

When the users are using the TS config, Wasp should check multiple `tsconfig.json` files:
- There must be a main `tsconfig.json` file referencing `tsconfig.app.json` and `tsconfig.node.json`.
- The `tsconfig.app.json` (TS config for src) file must pass validation (we already do this)
- The `tsconfig.node.json` file must pass validation.

When the users are using the Wasp DSL, Wasp should validate the `tsconfig.json` file (TS config for src). I believe we already do this.

Related: #2390.

### LINKED ISSUE #2694: Mention which `tsconfig.*.json` has errors when outputing `tsconfig.*.json` errors
When using the Typescript spec, we have three different `tsconfig.json` files:
- `tsconfig.json`
- `tsconfig.src.json`
- `tsconfig.wasp.json`

Image

We should mention the file name in the errors and not just say TS config.

sodic:
## Description

Fixes #2426 
Fixes #2694

We validate all `tsconfig.json` files depending on the project variant, and we output the correct name for each one. I had to change up how we organize our code so check the comments scattered across the PR.

In short:
- We now do all external config valiadations before analyzing the Wasp file and creating an `AppSpec`. You can easily see this change in `Analyze.hs`
- The modules inside `Wasp/ExternalConfig` are now completely independent of Wasp. They model general config files (e.g., `tsconfig.json`, `package.json`, etc.). These files only do general validation that's true regardless of Wasp - like whether the `compilerOptions` is an object. Most of this is just parsing and unmarshalling the file.
- The modules inside `Wasp/Project/ExternalConfig`  also deal with external configs, but within the context of a Wasp project. This is where all Wasp-specific validation now lives - like whether `outDir` is `.wasp/out/user`.

### Ts spec projects
```
❯ ../run wasp-cli info

❌ --- [Error] Analyzing wasp project failed: -------------------------------------

3 errors found:
- In "tsconfig.src.json" → "compilerOptions.jsx":
    Expected "preserve" but got "peserve".

- In "tsconfig.wasp.json" → "compilerOptions.module":
    Expected "NodeNext" but got "NdeNext".

- In "tsconfig.json" → "references":
    Missing required reference { "path": "tsconfig.src.json" }.
```
### Dsl projects
```
❯ ../run wasp-cli info

❌ --- [Error] Analyzing wasp project failed: -------------------------------------

1 errors found:
- In "tsconfig.json" → "include":
    Expected ["src"] but got ["\128076"].
```

## Type of change

- [x] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:


waspc/src/Wasp/Validator.hs
```diff
@@ -90,6 +90,12 @@ eqJust expected Nothing =
 ifJust :: Validator a -> Validator (Maybe a)
 ifJust = maybe success
 
+-- | Requires the value to be Just, then runs the inner validator on the
+-- unwrapped value. If the value is Nothing, the validation fails.
+required :: Validator a -> Validator (Maybe a)
+required _ Nothing = failure "Missing required value."
+required innerValidator (Just value) = innerValidator value
```
sodic, line 97: @cprecioso Ideas for a better name welcome :)

infomiho:
Heads up: the ergonomic import-form stack changes the required `tsconfig.wasp.json` shape for Wasp TS specs.

Stack base PR: https://github.com/wasp-lang/wasp/pull/4112

Once that stack lands, `Wasp.Project.ExternalConfig.WaspTsConfig.validateWaspTsConfig` should validate the new import-form config shape instead of the current old shape. In particular, the required config should to whatever [`tsconfig.wasp.json`](https://github.com/wasp-lang/wasp/pull/4118/changes#diff-9032ee2937b4997e49bd4b09333adc16286a7a7057c9bfb495a89fa8e78497c1) looks like in that PR.

waspc/src/Wasp/Project/ExternalConfig.hs
```diff
@@ -1,55 +1,46 @@
...
+import Wasp.Project.ExternalConfig.PackageJson (parseAndValidateUserPackageJson)
+import Wasp.Project.ExternalConfig.RootTsConfig (parseAndValidateRootTsConfig)
+import Wasp.Project.ExternalConfig.SrcTsConfig (parseAndValidateSrcTsConfig)
 import Wasp.Project.ExternalConfig.ViteConfig (validateViteConfig)
+import Wasp.Project.ExternalConfig.WaspTsConfig (parseAndValidateWaspTsConfig)
 
 data ExternalConfigs = ExternalConfigs
-  { _packageJson :: PackageJson,
-    _srcTsConfig :: TsConfig
+  { _packageJson :: PackageJson
```
sodic, line 22: This is the only one we actually need, the `_srcTsConfig` was included only to be validated later.

waspc/src/Wasp/Project/ExternalConfig/ViteConfig.hs
```diff
@@ -9,16 +9,17 @@ import qualified Data.Text as T
...
 import System.Directory (doesFileExist)
+import Validation (Validation (..))
 import Wasp.Project.Common
   ( CompileError,
     WaspProjectDir,
   )
 import qualified Wasp.Util.IO as IOUtil
 
-validateViteConfig :: Path' Abs (Dir WaspProjectDir) -> IO (Either CompileError ())
+validateViteConfig :: Path' Abs (Dir WaspProjectDir) -> IO (Validation [CompileError] ())
```
sodic, line 19: Had to convert this to Validation. Check the note in Project/ExternalConfig.hs to understand why.

waspc/src/Wasp/Project/ExternalConfig/TsConfig.hs
sodic, line 1: This module is now a general validator for TSConfig files in the Wasp project. Unlike `ExternalConfig/TsConfig.hs` which is about tsconfig.json files in general, this module is specifically concerned with `tsconfig.json` files inside Wasp projects.

waspc/src/Wasp/ExternalConfig/TsConfig.hs
sodic, line 1: This is our general model of TsConfig files (unrelated to Wasp projects). It could theoretically be considered a TS config parsing library - there's nothing Wasp-specific inside. It's not a publishable library though, as we're adding fields  only when we begin caring about them. :)

This means that `CompilerOptions` had to become optional (because according to `tsconfig` specification, it is). But we also updated the validation to check for it. So:
- It used to fail during parsing.
- Now it fails during validation of the `tsconfig.src.json`, which is more appropriate.

waspc/src/Wasp/Project/ExternalConfig/SrcTsConfig.hs
sodic, line 1: The validation logic in this file is identical to what we had before in `waspc/src/Wasp/Project/ExternalConfig/TsConfig.hs`. It's just moved.

The only difference is ` V.required` for `compilerOptions`. This is because missing `compilerOptions` no longer fail on the parsing stage.

waspc/src/Wasp/Project/Common.hs
```diff
@@ -20,9 +20,12 @@ module Wasp.Project.Common
     nodeModulesDirInWaspProjectDir,
     srcDirInWaspProjectDir,
     prismaSchemaFileInWaspProjectDir,
-    getSrcTsConfigInWaspProjectDir,
-    srcTsConfigInWaspLangProject,
-    srcTsConfigInWaspTsProject,
+    getTsConfigPathsForWaspProject,
+    tsConfigPahtsInWaspLangProjects,
+    tsConfigPathsInWaspTsProjects,
```
sodic: Wasp lang ts config paths are exported because of the LSP.
Wasp ts ts config paths are exported because they're referenced in root ts config validation.

waspc/tests/Project/ExternalConfig/RootTsConfigTest.hs
sodic, line 1: All these validator tests are basically duplication of the rules in the validator. But since it's so user-facing, I'd keep them.

waspc/src/Wasp/Project/Common.hs
```diff
@@ -96,16 +113,26 @@ dotWaspInfoFileInGeneratedAppDir = [relfile|.waspinfo|]
...
+
+tsConfigPathsInWaspTsProjects :: TsConfigPaths
+tsConfigPathsInWaspTsProjects =
+  TsConfigPaths
+    { srcTsConfig = [relfile|tsconfig.src.json|],
+      rootTsConfig = Just [relfile|tsconfig.json|],
+      waspTsConfig = Just [relfile|tsconfig.wasp.json|]
+    }
+
+tsConfigPahtsInWaspLangProjects :: TsConfigPaths
```
infomiho:
```suggestion
tsConfigPathsInWaspLangProjects :: TsConfigPaths
```

waspc/src/Wasp/Project/ExternalConfig/TsConfig.hs
```diff
@@ -1,75 +1,37 @@
...
-  where
-    findTsConfigOrError = maybeToEither fileNotFoundMessage <$> findFileInWaspProjectDir waspDir srcTsConfigPath
-    fileNotFoundMessage = "Couldn't find the tsconfig.json file in the " ++ toFilePath waspDir ++ " directory"
-
-validateSrcTsConfig :: T.TsConfig -> [CompileError]
-validateSrcTsConfig config =
-  show <$> V.execValidator tsConfigValidator config
+  Path' (Rel WaspProjectDir) (File f) ->
+  IO (Validation [CompileError] T.TsConfig)
+parseAndValidateTsConfigFile validator waspDir someTsConfigInProjectDir =
```
infomiho:
```suggestion
parseAndValidateTsConfigFile validator waspDir tsConfigFile =
```
I found the `some` prefix weird. Of course it's _some_ ts config. I don't think we need to encode "there might be others" in the name, as far as this function is concerned, it's a ts config file and it can do its job.

waspc/src/Wasp/Project/ExternalConfig/WaspTsConfig.hs
```diff
@@ -0,0 +1,40 @@
...
+parseAndValidateWaspTsConfig ::
+  Path' Abs (Dir WaspProjectDir) ->
+  Path' (Rel WaspProjectDir) (File WaspTsConfigFile) ->
+  IO (Validation [CompileError] T.TsConfig)
+parseAndValidateWaspTsConfig = parseAndValidateTsConfigFile waspTsConfigValidator
+
+waspTsConfigValidator :: V.Validator T.TsConfig
+waspTsConfigValidator =
+  V.all
+    [ V.inField ("include", T.include) $ V.eqJust ["main.wasp.ts"],
```
infomiho, line 25: What are our thoughts on `findWaspFile` allowing any `*.wasp.ts` file and we hard-coding `main.wasp.ts` in validation? I think it's okay if we support this for now. I guess we'd need to expand it to `*.wasp.ts` once we add support for multiple files.

infomiho (approved):
Looking good, one typo and some small comments

waspc/src/Wasp/Project/ExternalConfig/WaspTsConfig.hs
sodic: If we can do `*.wasp.ts`, then sure. I believe you'll hit this in https://github.com/wasp-lang/wasp/issues/2481 :)

I also assume we'll somehow make `findWaspFile` more precise.

waspc/src/Wasp/Project/ExternalConfig/TsConfig.hs
sodic: Ok, fair enough.

PR #3913, opened 2026-03-12
SSR: Enable index prerender

cprecioso:
This is not supposed to be merged.

Just a test to try out `/` prerendering based on the system put in place in https://github.com/wasp-lang/wasp/pull/3855


cprecioso:
Not needed anymore

PR #3915, opened 2026-03-13
Make Version be a valid SemVer version

FranjoMindek:
## Summary
- Makes our `Version` implementation follow the SemVer specification: https://semver.org/
	- The main difference is that SemVer version requires all 3 version components to be present (`major.minor.patch`), so partial versions are not allowed (e.g. `1.2`).
- Temporarily that means that our `Range` implementation does not allow partial versions (even though it should). We will be solving this in the next PRs.


waspc/src/Wasp/SemanticVersion/VersionBound.hs
```diff
@@ -36,6 +38,18 @@ versionFromBound (Exclusive v) = Just v
 -- we know only its lower and upper bound.
 type VersionInterval = (VersionBound, VersionBound)
 
+-- | Version interval which satisfies any possible SemVer version.
```
FranjoMindek, line 41: Will be used in later PRs.

waspc/src/Wasp/SemanticVersion/Parsers.hs
```diff
@@ -0,0 +1,18 @@
+module Wasp.SemanticVersion.Parsers
```
FranjoMindek, line 1: This is extracted to a separate module because this parser will be re-used in the future PRs.

waspc/src/Wasp/SemanticVersion/Version.hs
```diff
@@ -2,49 +2,61 @@
...
-parseVersion = parse versionParser ""
+v :: TH.QuasiQuoter
+v = quasiQuoterFromParser strictParseVersion
+
+strictParseVersion :: String -> Either P.ParseError Version
+strictParseVersion = P.parse (versionParser <* P.eof) ""
 
-versionParser :: Parsec String () Version
+parseVersion :: String -> Either P.ParseError Version
+parseVersion = P.parse versionParser ""
```
cprecioso: The current `parseVersion` is a footgun -- if you want to parse a version you should use the whole string (`strictParseVersion`), or embed the parser into another one (`versionParser`). There's not a situation where you want to call the parser but not consume the whole string, seems like a mistake.

I would remove this "non-strict" version, and rename the "strict" one just `parseVersion`.

waspc/src/Wasp/SemanticVersion/Parsers.hs
```diff
@@ -0,0 +1,18 @@
...
+  ( naturalNumberParser,
+  )
+where
+
+import Control.Applicative ((<|>))
+import Numeric.Natural (Natural)
+import qualified Text.Parsec as P
+
+naturalNumberParser :: P.Parsec String () Natural
```
cprecioso, line 12: I'm not sure this is the correct name for this parser. These are not general rules for natural number parsers, just for the ones in semver versions. I would prefix it with "version" or something

cprecioso: Tbh I am not sure if it's very important that we fail on 0-prefixed numbers, I think we might get by using the Parsec-builtin natural number parser.

waspc/src/Wasp/SemanticVersion/VersionBound.hs
```diff
@@ -14,13 +14,15 @@ module Wasp.SemanticVersion.VersionBound
...
 where
 
 import qualified Language.Haskell.TH.Quote as TH
 import qualified Language.Haskell.TH.Syntax as TH
 import qualified Text.Parsec as P
-import Wasp.SemanticVersion.Version (Version, versionParser)
+import Wasp.SemanticVersion.Version (Version (..), versionParser)
 import Wasp.Util.TH (quasiQuoterFromParser)
 
 data VersionBound = Inclusive !Version | Exclusive !Version | Inf
```
cprecioso, line 28: (i realize it's not code in this PR, but) what does `Inf` mean?

cprecioso: Oooh "Infinite" I guess? If you want to get into this -- I'd rename it "Unspecified", I don't think Infinite conveys what it is. But feel free to ignore if you deem out of scope.

FranjoMindek: But it should be infinite? e.g. `<-∞, 5]` would use the `Inf` for the left bound?

waspc/src/Wasp/SemanticVersion/Parsers.hs
FranjoMindek: This is because I followed the SemVer's BNF grammar: 
https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions
See `<numeric identifier>`.


We also have the following paragraph.
> A normal version number MUST take the form X.Y.Z where X, Y, and Z are non-negative integers, and MUST NOT contain leading zeroes.

waspc/src/Wasp/SemanticVersion/Version.hs
FranjoMindek: > The current parseVersion is a footgun -- if you want to parse a version you should use the whole string (strictParseVersion), or embed the parser into another one (versionParser).

Yes. That's exactly the reason we have both of them.
Makes sense then not to export `parseVersion` as that parser is usually embedded into other parsers.

waspc/src/Wasp/SemanticVersion/Parsers.hs
cprecioso: No no, I get that semver requires it! But I was thinking we don't need to be a strict semver parser, and the world won't end if we parse a leading zero. 😁
But since the code is already done, I basically don't mind it either way.

waspc/src/Wasp/SemanticVersion/VersionBound.hs
cprecioso: Yep yep, I understand, I just thought it would be more explicative (for the current context) to name it "unspecified" or something similar. But anyway, I really really don't mind much.

> `<-∞, 5]`

but then where's my `NegInf`!!1!!1! 😡👊

anyway, leave it as is 😁

FranjoMindek: Ask @Martinsos 😉.

I think he simplified it that an inifnity on left bound is considered negative.
But yes.

PR #3916, opened 2026-03-13
Re-implememnt Comparator and Comparator set logic

FranjoMindek:
## Summary

This PR does the heavy lifting of the `node-semver` logic implementation. 

It implements the "middle layer" between the `Semantic Version` and `Range`
Implementation of the middle layer differs from the `node-semver` (it also did before) purposely. Mostly because we modeled a much richer `node-semver` domain, instead of hiding the logic in the parsing layer. 

**The added files are not yet wired together into the old code**, so this PR basically adds "dead code". The next PR will refactor the `SemanticVersioning` module to use the new logic.


waspc/src/Wasp/SemanticVersion/PartialVersion.hs
```diff
@@ -0,0 +1,79 @@
...
+  deriving (Eq, TH.Lift)
+
+-- | We rely on this 'show' implementation to produce valid `node-semver` partial version.
+instance Show PartialVersion where
+  show (MajorMinorPatch mjr mnr ptc) = printf "%d.%d.%d" mjr mnr ptc
+  show (MajorMinor mjr mnr) = printf "%d.%d" mjr mnr
+  show (Major mjr) = printf "%d" mjr
+  show Any = "*"
+
+fromVersion :: Version -> PartialVersion
```
FranjoMindek, line 45: This!

Issue #2941, opened 2025-07-10
Support tilde (~) ranges in our semantic versioning


sodic:
Add support for tilde ranges in our `SementicVersion` module

cprecioso:
@sodic Is it really necessary to emulate the behaviour of missing numbers?

I don't really see a use for having a `Version (Natural, Maybe Natural, Maybe Natural)`, we're probably always going to specify the full triple `Version (Natural, Natural, Natural)` in the compiler.

In the same vein, we for example don't implement prerelease identifiers (`-beta.8`) or build numbers (`+8fc62b`), because we're most probably not needing to use them.

sodic:
> we're probably always going to specify the full triple Version (Natural, Natural, Natural) in the compiler.

Yes, definitely. But we also want to use this module to work with user-defined versions (e.g., for #1859).

PR #3917, opened 2026-03-13
Wire new `node-semver` implementations, make `SemanticVersion` a re-export hub

FranjoMindek:
## Summary
- Extracts the `Range` logic into a separate submodule.
- Makes the `Wasp.SemanticVersion` a re-export of the `SemanticVersion` "public API". Consumer which used `import qualified Wasp.SemanticVersion as SV` don't have to change anything.
- Makes the `Wasp.SemanticVersion` use the new code added in previous PRs instead.

Solves: partially https://github.com/wasp-lang/wasp/issues/2941 (the rest is in next PR), https://github.com/wasp-lang/wasp/issues/3699


FranjoMindek:
Will merge this PR with it's parent PR, since this one changes the public API.

PR #3919, opened 2026-03-13
Migrate to the new `SemanticVersion` code

FranjoMindek:
## Summary
- Rewrites `Wasp.SemanticVersion` from a monolithic module into a re-export hub for the new `SemanticVersion` API built in previous PRs.
- Deletes old `SemanticVersionTest` (coverage moved to per-module tests)
- Migrates all consumers (`DepVersions`, `SdkGenerator`, `ServerGenerator`, `EmailSender`, `JobGenerator`, `WebSocket`, `PackageJson`, `AppSpec`, `Node`) to the new API

Solves: [wasp-lang/wasp#2941](https://github.com/wasp-lang/wasp/issues/2941), [wasp-lang/wasp#3699](https://github.com/wasp-lang/wasp/issues/3699)

### DIFF
--- a/waspc/ChangeLog.md
+++ b/waspc/ChangeLog.md
@@ -11,6 +11,7 @@ Remember to check out the [migration guide](https://wasp.sh/docs/migration-guide
 ### 🎉 New Features
 
 - The Wasp TS config file (`*.wasp.ts`) now supports async logic in the default export, enabling use cases like file-based routing and dynamic configuration. ([#3900](https://github.com/wasp-lang/wasp/pull/3900))
+- The `wasp.version` field now accepts any valid npm-compatible version range (e.g. `>=0.15.0 <0.22.0`, `~0.21.0`, `0.21.x`) instead of only `^x.y.z`. ([#3921](https://github.com/wasp-lang/wasp/pull/3921))
 
 ### 🐞 Bug fixes
 
--- a/waspc/src/Wasp/AppSpec/Valid.hs
+++ b/waspc/src/Wasp/AppSpec/Valid.hs
@@ -12,10 +12,10 @@ module Wasp.AppSpec.Valid
 where
 
 import Control.Monad (unless)
+import Data.Bifunctor (first)
 import Data.List (find, group, groupBy, intercalate, sort, sortBy)
 import Data.Maybe (fromJust, fromMaybe, isJust, isNothing)
-import Text.Read (readMaybe)
-import Text.Regex.TDFA ((=~))
+import qualified Text.Parsec as P
 import Wasp.Analyzer.Parser (isValidWaspIdentifier)
 import Wasp.AppSpec (AppSpec)
 import qualified Wasp.AppSpec as AS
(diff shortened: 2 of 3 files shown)


waspc/src/Wasp/SemanticVersion/Range.hs
```diff
@@ -0,0 +1,139 @@
...
+      let versionInterval = versionBounds versionRange
+      let lowerBound = fst versionInterval
+      lowerBoundVersion <- versionFromBound lowerBound
+      let noMajorChangesInterval =
+            (lowerBound, Exclusive $ nextBreakingChangeVersion lowerBoundVersion)
+      guard $ versionInterval `isSubintervalOf` noMajorChangesInterval
+
+-- Helper methods for constructing a 'Range'.
+
+backwardsCompatibleWith :: Version -> Range
```
FranjoMindek, line 88: There are moved from `ComparatorSet` level to `Range` level.
I've made it so that `ComparatorSet` is an internal implementation detail, the public API of `SemanticVersion` does not expot it.

We should either work with `Version` or `Range` and nothing else.

waspc/src/Wasp/SemanticVersion.hs
```diff
@@ -1,146 +1,145 @@
...
-doesVersionRangeAllowMajorChanges = not . doesVersionRangeAllowOnlyMinorChanges
-  where
-    doesVersionRangeAllowOnlyMinorChanges versionRange = isJust $ do
-      let versionInterval = versionBounds versionRange
-      let lowerBound = fst versionInterval
-      lowerBoundVersion <- versionFromBound lowerBound
-      let noMajorChangesInterval =
-            (lowerBound, Exclusive $ nextBreakingChangeVersion lowerBoundVersion)
-      guard $ versionInterval `isSubintervalOf` noMajorChangesInterval
+{--
```
FranjoMindek, line 47: TODO: still cleanup this comment.

waspc/src/Wasp/Generator/WebSocket.hs
```diff
@@ -16,10 +16,10 @@ areWebSocketsUsed :: AppSpec -> Bool
 areWebSocketsUsed spec = isJust $ AS.App.webSocket $ snd $ getApp spec
 
 socketIoVersionRange :: SV.Range
-socketIoVersionRange = SV.Range [SV.backwardsCompatibleWith (SV.Version 4 6 1)]
+socketIoVersionRange = [SV.r|^4.6.1|]
 
 socketIoComponentEmitterVersionRange :: SV.Range
-socketIoComponentEmitterVersionRange = SV.Range [SV.backwardsCompatibleWith (SV.Version 4 0 0)]
+socketIoComponentEmitterVersionRange = [SV.r|^4.0.0|]
```
cprecioso, line 22: noice

cprecioso: Would a `IsString` instance be useful too?
https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Data-String.html#t:IsString

FranjoMindek: I would say no.
1 reason as to why: the parsing can fail. Not all strings are valid ranges. And we would find out our mistakes only in the runtime. 

QuasiQuoters actually work compile time. So I prefer them.

Issue #3306, opened 2025-11-05
Investigate if we can remove `dev-tool.project`


sodic:
We have a `dev-tool.project` file for managing developer tools (Ormolu, Stan, Hlint).

HLS brings all of these tools with it, so ideally we'd just use its versions and ensure that everyone uses the same version of HLS.

The only problem is CI and bulk use (e.g., `./run ormolu:format`). We're not sure whether we can easily triggerr the same tool versions HLS uses outside of the editor. If not, we'll probably have to keep this file.

Even if we keep the file, we should reconsider what we need (perhaps nothing except Ormolu).

Context: https://github.com/wasp-lang/wasp/pull/3287#discussion_r2490737674

cprecioso:
Related: https://www.notion.so/RFC-Declarative-toolchain-management-with-mise-24118a74854c80cc9703cf456dec3bd7

cprecioso:
Related issue: https://github.com/wasp-lang/wasp/issues/3927

PR #3940, opened 2026-03-19
Use Mise to manage all dev and CI toolchains

cprecioso:
## Description

Migrate all of our development and CI toolchains to [mise](https://mise.jdx.dev/), so that local development and CI use the exact same tool versions from a single source of truth.

Previously, tool versions were scattered across several files and mechanisms:

- GHC / HLS / Cabal via `ghcup` (and `./run ghcup-set`)
- Node via `.nvmrc`
- Ormolu / cabal-gild installed ad-hoc through Cabal (`dev-tool.project`)
- Fly / Railway CLIs pinned inside individual workflows

Now everything is declared in a single root `mise.toml`, and both contributors and CI install tools the same way.

- Resolves https://github.com/wasp-lang/wasp/issues/3927

## Changes

The meat of the PR is in `mise.toml`, which describe the tools we require (GHC/HLS/Cabal, Node, formatters), the system package manager requirements for them, and a few settings.

This file contains every tool needed to develop on Wasp. A fresh computer can install Mise, run `mise bootstrap`, and have everything it needs in a single step.

Similarly, I created a `prepare-env` action that only needs to do `mise bootstrap` to completely set up a runner, the rest are one Windows workaround and setting up caches. It takes about the same time that the previous `setup-node` + `setup-haskell` actions took (1-2 min).

The rest of the PR is adapting the workflows to use this new action instead of each one declaring its stuff.

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:


.github/workflows/ci-deploy-test.yaml
```diff
@@ -58,12 +49,14 @@ jobs:
...
           # NOTE:
+          # - We tell users to install the latest version of Fly CLI, so we use
+          # it here too.
           # - The `yes` command is necessary because the `fly launch` command
           # prompts for confirmation.
           # - We use a Bash array for `$ENV_VAR_ARGUMENTS` to ensure proper
           # word splitting (i.e., force Bash to interpret the flags separately
           # instead of passing it as a single string value).
-          yes | wasp-cli deploy fly launch "$APP_PREFIX" "$FLY_REGION" --org "$FLY_ORG" "${ENV_VAR_ARGUMENTS[@]}"
+          yes | mise x flyctl@latest -- wasp-cli deploy fly launch "$APP_PREFIX" "$FLY_REGION" --org "$FLY_ORG" "${ENV_VAR_ARGUMENTS[@]}"
```
cprecioso: We use `mise x` (like `npx`) to ensure the latest version instead of the `mise.toml` one.

.github/workflows/ci-npm-package-test.yaml
```diff
@@ -19,8 +19,7 @@ jobs:
 
           - # Linux with Musl
             os: ubuntu-latest
-            container: node:24-alpine3.23
-            install-deps: apk add curl
```
cprecioso, line 23: Alpine ships with `wget` installed, which is enough to pick up Mise, and then Mise installs every other dependency.

.github/workflows/ci-waspc-build.yaml
```diff
@@ -71,35 +62,20 @@ jobs:
...
               apt-get update -y
-              # GHCup dependencies (https://www.haskell.org/ghcup/install/#version-2004-2010)
-              apt-get install -y build-essential curl libffi-dev libffi7 libgmp-dev libgmp10 libncurses-dev libncurses5 libtinfo5
-              # Cabal dependencies
-              apt-get install -y zlib1g-dev
+              apt-get install -y curl
 
           - name: linux-x86_64-static
             runner: ubuntu-latest
-            # actions/setup-node does not work in alpine.
```
cprecioso, line 83: Mise can install Node on Alpine so this note is not needed anymore and we can use the regular Alpine image.

.github/workflows/release-create.yaml
```diff
@@ -62,15 +68,6 @@ jobs:
...
           persist-credentials: false
 
-      - uses: actions/setup-node@v6
-        with:
-          node-version-file: .nvmrc
-          # Disable caching to avoid a potentially poisoned cache introducing a
-          # vulnerability not present in the stored artifacts.
-          # https://docs.zizmor.sh/audits/#cache-poisoning
-          cache: false
-          package-manager-cache: false
```
cprecioso, line 72: This was not really needed before either, the Runner by definition has a Node.js that can do `npx npm`. We don't care about that version because we're not building or running Wasp code, just publishing it.

waspc/cabal.project
```diff
@@ -1,3 +1,4 @@
+-- Keep in sync with /mise.toml
 with-compiler: ghc-9.6.7
```
cprecioso, line 2: This is not needed anymore actually, if you have mise installed it should never have the wrong version. We can keep it for possible warnings if we like though.

waspc/dev-tool.project
```diff
@@ -6,8 +7,6 @@ packages: .
 jobs: $ncpus
 
 constraints:
-  cabal-gild ==1.5.0.1,
-  ormolu ==0.7.2.0,
   stan ==0.2.1.0,
   hlint ==3.6.1
```
cprecioso, line 11: These are the only tools that resisted mise-ification, since they don't publish built artifacts. They are not run on CI and only occasionally (never?) by us locally. HLS already bundles them too; so we can just decide to remove these tools (and then also this whole file!).

In the meantime, they will work as they have been working for now, building from source with GHC when we call them. No change and no mise involved.

waspc/run
```diff
@@ -93,9 +93,6 @@ TEST_CMD="cabal test && \
 
 # Utility
 GHCID_CMD="ghcid --command=cabal repl"
-# HLS version below should be one that works with the GHC version from the
-# `with-compiler` field in `cabal.project`
-GHCUP_SET="ghcup set hls 2.10.0.0"
```
cprecioso, line 96: No more imperative version switching!

web/src/remark/search-and-replace.ts
```diff
@@ -16,7 +16,7 @@ const plugin: Plugin<[], Root> = () => (tree, file) => {
     latestWaspVersion: `^${waspVersion}`,
     pinnedLatestWaspVersion: waspVersion,
     // NOTE: Don't change Wasp's lowest supported Node version without updating it
-    // in all required places. Check /.nvmrc for the full list.
+    // in all required places. Check /mise.toml for the full list.
     minimumNodeJsVersion: "24.14.1",
```
cprecioso, line 20: In a future PR we might want to read this dynamically off `mise.toml`

.nvmrc
cprecioso, line 1: Bye nvm I never liked you

mise.lock
cprecioso, line 1: An actual lockfile for our dev environment versions! Nothing special to say here, just nice to have. The `mise` CLI maintains it automatically on installs an updates.

mise.toml
```diff
@@ -0,0 +1,76 @@
+min_version = "2026.6.11"
+
+[tools]
```
cprecioso, line 3: Mise supports different places where to get the binaries from. Tools in `mise.toml` are called `backend:tool`. e.g. `aqua:ghcup` gets `ghcup` from the `aqua` registry; and `github:tfausak/cabal-gild` gets it from that repo's GitHub Releases.

Tools in the [Mise registry](https://mise.jdx.dev/registry.html#tools) don't need to be prefixed with their backend.

mise.toml
```diff
@@ -0,0 +1,76 @@
...
+"flyctl" = "0.3.164"
+"github:tfausak/cabal-gild" = "1.5.0.1"
+"cabal" = "3.12.1.0"
+"ormolu" = "0.7.2.0"
+"railway" = "4.11.1"
+
+# The `mise-ghcup` plugin declares `aqua:ghcup` as a dependency specifically
+# instead of just `ghcup` to avoid conflicts, so we need to declare it like that
+# here as well.
+"aqua:ghcup" = "0.2.6.1"
```
cprecioso, line 4: Aqua is the default backend of Mise, and [their registry](https://github.com/aquaproj/aqua-registry/) is just a bunch of YAML files with URLs and checksums to the official binaries. I contributed the Windows definitions for GHCup there, because it was only pointing to the Mac and Linux ones.

mise.toml
```diff
@@ -0,0 +1,76 @@
...
+"mise-ghcup:hls" = "2.12.0.0"
+
+# Keep in sync with:
+# - The compiler (Wasp.Node.Version).
+# - The docs: /docs/introduction/quick-start.md ("Requirements" section).
+# - The docs plugin: /web/src/remark/search-and-replace.ts
+"node" = "24.14.1"
+
+[plugins]
+"vfox:mise-ghcup" = "https://github.com/cprecioso/mise-ghcup.git"
```
cprecioso: I wrote my own Mise plugin for getting Haskell tools from GHCup:
https://github.com/cprecioso/mise-ghcup
It's some simple Lua code calling GHCup with the correct parameters, we can bring it into the Wasp org on GitHub.

Mise supports ASDF plugins, and there's an existing one for GHCup at https://github.com/mise-plugins/mise-ghcup. However, that one doesn't support Windows, and all ASDF plugins are deprecated on Mise, so it was better to create a new Mise-specific plugin that does what we need.

mise.toml
```diff
@@ -0,0 +1,76 @@
...
+
+[bootstrap.packages]
```
cprecioso, line 34: `brew` packages for Mac are supported, but none is needed (I tested in a fresh VM, and apart from `git` to clone the repo and `mise` to run `mise install`, I didn't need anything else)

cprecioso (commented):
Some explanatory comments on why some stuff changed, and some Mise concepts that don't fit as comments on the files.

waspc/dev-tool.project
cprecioso: It seems we wanted to remove them anyway?
- https://github.com/wasp-lang/wasp/issues/3306

.github/actions/setup-env/action.yaml
```diff
@@ -0,0 +1,85 @@
+name: Setup environment and dev tools
+description: |
+  This action sets up our Mise-defined dev tools and environment for use in
+  actions. It also sets up caches for package manager dependencies.
+
+inputs:
+  cabal-project-dir:
+    description: |
+      The working directory for the action.
+      The default is /waspc.
```
Martinsos: Is this old? Feels a bit detached from what the action is now. I remember this making sense for the haskell setup action, but for this one?
Maybe it does, but we should rename the name of the variable, and/or its description?

.github/actions/setup-env/action.yaml
```diff
@@ -0,0 +1,85 @@
...
+      with:
+        bootstrap: true
+        bootstrap_args: --update --yes
+
+    - # `cabal.project` sets `with-compiler: x.y.z`, so Cabal resolves the bare
+      # program name `ghc-x.y.z` on PATH. On Windows, that hits Mise's shim,
+      # which Cabal can't determine the version of, and the build fails. We
+      # override `with-compiler` with the absolute path to the real `ghc.exe` so
+      # Cabal probes the actual binary instead of the shim. Using forward
+      # slashes keeps Cabal's config parser happy.
```
Martinsos: Hm I didn't fully get this one.

Ok cabal is looking for ghc-x.y.z because of `with-compiler` in `cabal.project`.

On Windows that doesn't work because it hits Mise's shim and Cabal can't determine the version of it -> what does that mean? It also hits Mise's shim on Linux, right? So that is not the unique part I guess, but the fact that in case of Windows, it can't figure the version of Mise's shim? What means it can't figure out its version, version of what? Version of shim? If it is looking for ghc-x.y.z and hits a shim, isn't the shim the right version, I mean it is in the name, how did it hit it otherwise?

.github/actions/setup-tools/action.yaml
```diff
@@ -0,0 +1,85 @@
...
+      # override `with-compiler` with the absolute path to the real `ghc.exe` so
+      # Cabal probes the actual binary instead of the shim. Using forward
+      # slashes keeps Cabal's config parser happy.
+      name: Pin Cabal to the real GHC binary (Windows)
+      if: runner.os == 'Windows'
+      shell: pwsh
+      working-directory: ${{ inputs.cabal-project-dir }}
+      run: |
+        $ghc = (mise which ghc) -replace '\\', '/'
+        "with-compiler: $ghc" | Set-Content -Path cabal.project.local -Encoding ascii
```
Martinsos: Aha, so you actually produce cabal.project.local that points with-compiler directly to the GHC that mise installed. I guess that does fix it on CI. But what if I want to do local development on Windows? I will still be having same issues, right?

Would be great if instead we fixed this more "upstream" to say it that way, so we don't have to do this "patch". But I don't know enough to say more, maybe once you shine more light on my comment above.

.github/actions/setup-env/action.yaml
```diff
@@ -0,0 +1,85 @@
...
+      working-directory: ${{ inputs.cabal-project-dir }}
+      run: |
+        $ghc = (mise which ghc) -replace '\\', '/'
+        "with-compiler: $ghc" | Set-Content -Path cabal.project.local -Encoding ascii
+
+    - name: Cache npm dependencies
+      uses: actions/setup-node@v6
+      with:
+        # We don't specify a Node version here, so it won't install any new one
+        # and just use the one from Mise.
```
Martinsos: Why do we even need it? Oh, we are using it just for caching stuff? So we install node through mise, but use this action to cache npm correctly on github? Maybe you can make that even more explicit. This comment could say that: We only use part of action that does caching.

.github/actions/setup-env/action.yaml
```diff
@@ -0,0 +1,85 @@
...
+        cache: npm
+        package-manager-cache: true
+
+    - name: Probe environment
+      id: probe
+      shell: bash
+      run: |
+        set -euxo pipefail
```
Martinsos: AI loves using this hehe, I never used it before.

.github/actions/setup-env/action.yaml
```diff
@@ -0,0 +1,85 @@
...
+        # apart.
+        ldd_version="$(ldd --version 2>&1 || true)"
+        case "$ldd_version" in
+          *musl*) libc=musl ;;
+          *) libc=default ;;
+        esac
+        echo "LIBC=$libc" >> "$GITHUB_OUTPUT"
+
+        # The first run of `cabal path` has some extra messages telling you that
+        # it's creating the store, so we run the command once to get rid of it.
```
Martinsos: Maybe would be more robust to get the last line of output from these commands below? This feels potentially flaky in the future, what if cabal path decides to print something else extra in the commands below, obviously it is already doing that.

.github/actions/setup-env/action.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+    - name: Cache npm dependencies
+      uses: actions/setup-node@v6
+      with:
+        # We don't specify a Node version here, so it won't install any new one
+        # and just use the one from Mise.
+        cache: npm
+        package-manager-cache: true
+
+    - name: Probe environment
```
Martinsos: Ok this was a bit mistery to me, why are we probing environemnt?
I read through action to understand what is happening: you are obtaining versions of some tools. GHC, Cabal. Ok, also libc.
Why not get these from mise itself, I am sure it can return them? Except for libc I guess?

.github/actions/setup-tools/action.yaml
```diff
@@ -0,0 +1,85 @@
...
+        # it's creating the store, so we run the command once to get rid of it.
+        cabal path 2>&1 >/dev/null
+        echo "CABAL_STORE=$(cabal path --store-dir)" >> $GITHUB_OUTPUT
+        echo "CABAL_INDEX=$(cabal path --remote-repo-cache)" >> $GITHUB_OUTPUT
+
+    # Based on the official recipe for Cabal caching:
+    # https://github.com/actions/cache/blob/v5.0.4/examples.md#haskell---cabal
+    - name: Cache Cabal dependencies
+      uses: actions/cache@v5
+      with:
```
Martinsos, line 26: Previously we cached only one of these three paths, and had big comment explaining why. What changed, why are we now caching all of them?

.github/actions/setup-env/action.yaml
```diff
@@ -0,0 +1,85 @@
...
+        path: |
+          ${{ steps.probe.outputs.CABAL_STORE }}
+          ${{ steps.probe.outputs.CABAL_INDEX }}
+          ${{ inputs.cabal-project-dir }}/dist-newstyle
+        key: |
+          cabal-v2-${{ inputs.cabal-project-dir }}-${{ runner.os }}-${{ runner.arch }}-${{ steps.probe.outputs.LIBC }}-${{ steps.probe.outputs.GHC_VERSION }}-${{ hashFiles(format('{0}/*.cabal', inputs.cabal-project-dir), format('{0}/*.project', inputs.cabal-project-dir), format('{0}/*.project.freeze', inputs.cabal-project-dir)) }}
+        restore-keys: |
+          cabal-v2-${{ inputs.cabal-project-dir }}-${{ runner.os }}-${{ runner.arch }}-${{ steps.probe.outputs.LIBC }}-${{ steps.probe.outputs.GHC_VERSION }}-
+
+    - name: Update Cabal package index
```
Martinsos: Is this new, cabal update? We weren't doing this before? If so, why now?

.github/actions/setup-tools/action.yaml
```diff
@@ -0,0 +1,85 @@
+name: Setup environment and dev tools
+description: |
+  This action sets up our Mise-defined dev tools and environment for use in
```
Martinsos, line 3: Ok so this action is now general, "tools for the whole repo".
Which I guess makes sense, since mise is like that, tools for whole repo, it is not just waspc or web.
And is that reasonable? I guess it is, we will be keeping all things in this repo on the same tool versions, right? For now at least.

Ok so before we have specific setup-haskell and setup-node actions, now we have this one. I guess this means we are sometimes setting up redundant stuff -> CI for web for example doesn't need cabal. But we don't mind? How are we thinking about this.

It's just that before it was clearer to me, I am setting up what I need for certain action. Now it is this general wide set up hm.

.github/actions/setup-env/action.yaml
```diff
@@ -0,0 +1,85 @@
...
+    # https://github.com/actions/cache/blob/v5.0.4/examples.md#haskell---cabal
+    - name: Cache Cabal dependencies
+      uses: actions/cache@v5
+      with:
+        path: |
+          ${{ steps.probe.outputs.CABAL_STORE }}
+          ${{ steps.probe.outputs.CABAL_INDEX }}
+          ${{ inputs.cabal-project-dir }}/dist-newstyle
+        key: |
+          cabal-v2-${{ inputs.cabal-project-dir }}-${{ runner.os }}-${{ runner.arch }}-${{ steps.probe.outputs.LIBC }}-${{ steps.probe.outputs.GHC_VERSION }}-${{ hashFiles(format('{0}/*.cabal', inputs.cabal-project-dir), format('{0}/*.project', inputs.cabal-project-dir), format('{0}/*.project.freeze', inputs.cabal-project-dir)) }}
```
Martinsos: Why v2 here?

.github/workflows/ci-deploy-test.yaml
```diff
@@ -32,23 +32,14 @@ jobs:
...
-          node-version-file: .nvmrc
+      - uses: ./.github/actions/setup-env
 
       - uses: ./.github/actions/install-wasp-cli-artifact
 
       - name: Install app dependencies
         working-directory: ${{ env.APP_TO_DEPLOY }}
         run: wasp-cli install
 
-      - name: Install Fly CLI
```
Martinsos, line 47: This reasoning here does feel a bit specific to this very github action -> comment explains it. Hm. Should fly really be installed by Mise? Because it is not a development tool needed by us really, it is a dep of `wasp` CLI. I would maybe rather we expeience the same experience as our users have, and they won't have mise managing it for them. I am mixing a couple of concepts here but I still think they might drive to the same point, that maybe fly shouldn't be manage by mise.

.github/workflows/ci-deploy-test.yaml
Martinsos: Ok this solves somewhat my comment from above! But I still wonder if we should manage Fly at all in our mise.toml.

.github/workflows/ci-deploy-test.yaml
```diff
@@ -95,12 +88,14 @@ jobs:
     if: always()
 
     steps:
-      - uses: superfly/flyctl-actions/setup-flyctl@v1
+      # NOTE: We use our pinned Fly version because we don't want the changes in
```
Martinsos: Hm. "just incidentally using the same tool Wasp uses" -> well not incidentally, we are using Fly because Wasp uses Fly and we are testing how Wasp app gets to deployed to fly, right? We are in deploy-test.yaml.
I guess it would be more correct to say that here, these commands written down, are written in mind with version of fly that supports them, so supports `destroy` command. So if something changed and fly stopped working like this, these commands would fail.
I guess that kind of makes sense. But if we just didn't manage Fly in mise and used the latest one, ok maaaaaybe this would fail, but we would get error message and could fix it right? Also doesn't sound very likely.
But if we keep this one pinned -> well what if that pin becomes quite old and then it fails because of that, cli is too old? I don't know hm, it doesn't sound fully convincing.

.github/workflows/ci-deploy-test.yaml
```diff
@@ -119,31 +114,26 @@ jobs:
...
-      - name: Install Railway CLI
-        run: npm install -g @railway/cli@4.11.1
-
       - name: Deploy app to Railway
         working-directory: ${{ env.APP_TO_DEPLOY }}
         run: |
           echo "Deploying with prefix: $APP_PREFIX"
           mapfile -d $'\0' -t ENV_VAR_ARGUMENTS < <(node ../../scripts/deploy-test/prepare-deploy-secrets.mjs .env.server.example)
           # NOTE:
+          # - We tell users to install the latest version of Railway CLI, so we
```
Martinsos: All the same stuff for Railway as we said for Fly.

.github/workflows/ci-formatting.yaml
```diff
@@ -16,11 +16,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
-
-      - uses: actions/setup-node@v6
-        with:
-          cache: "npm"
-          node-version-file: .nvmrc
+      - uses: ./.github/actions/setup-env
```
Martinsos: Hm yeah doesn't it feel a bit silly here that we set up the whole env when we really are doing just formatting?

.github/workflows/ci-npm-package-test.yaml
```diff
@@ -19,8 +19,7 @@ jobs:
...
+            container: alpine:3.23
```
Martinsos, line 22: Why did you drop node in front of alpine? Because now mise will set up node?

.github/workflows/ci-waspc-build.yaml
```diff
@@ -4,15 +4,6 @@ name: "Build - Wasp CLI binaries"
 # We can call it manually (workflow_dispatch) or from other workflows (workflow_call).
 on:
   workflow_call:
-    inputs:
-      node-version:
-        # NOTE: This should be equal to Wasp's lowest supported Node version.
```
Martinsos, line 9: I checked comments on node version in mise.toml, but I don't think I saw it mentioning that we want to keep this equal to the Wasp's lower supported Node version, as it is mentioned here. Did that get lost on purpose?

.github/workflows/ci-waspc-test.yaml
```diff
@@ -25,14 +24,6 @@ jobs:
...
           # We are using a fixed Ubuntu version instead of 'ubuntu-latest' to
           # prevent the version from changing on its own.
-          #
-          # We also set the Ubuntu version to the oldest supported LTS version
-          # rather than the latest LTS version. Doing so avoids GLIBC version
-          # mismatch errors (and possibly other similar issues).
-          #
-          # This might become redundant if we start building with Alpine/Musl
-          # and link the binaries statically:
-          # https://github.com/wasp-lang/wasp/issues/650#issuecomment-1180488040
```
Martinsos, line 35: This got solved?

.github/workflows/ci-waspc-test.yaml
```diff
@@ -9,7 +9,6 @@ env:
 
 defaults:
   run:
-    shell: bash
```
Martinsos, line 12: This was redundant?

.github/workflows/ci-waspc-test.yaml
```diff
@@ -104,4 +84,7 @@ jobs:
       - name: Run `waspc` e2e tests
         env:
           WASP_E2E_TESTS_SKIP_DOCKER: ${{ runner.os == 'macOS' && '1' || '' }}
-        run: ./run test:waspc:e2e
+          NODE_TOOL: ${{ matrix.node-version == 'default' && 'node' || format('node@{0}', matrix.node-version) }}
```
Martinsos: Aha this is how we now pick correct node hm. Quite convluted here. What if I add another action that needs to run something with mise and node?
Btw is node-version only used here? Maybe call it mise-node-name directly? But yeah sucks to have that there in the matrix, mise is an implementation detail that happens later. Hm.
I just found it quite convluted reading this.

.github/workflows/release-create.yaml
```diff
@@ -23,6 +23,12 @@ jobs:
     with:
       override-npm-package-version: ${{ github.ref_name }}
 
+  # These following jobs are the only ones where we don't want to use our
```
Martinsos: I would skip the "only one" part because you don't ahve to assume what is happening with other workfklows, you can just say that in these jobs we don't want to use cache and that is good enough (and more robust of a statement).
nitpick.

waspc/cabal.project
Martinsos: Hm I wonder myself what we should do. It feels like a riht place to specify. But then I don't like duplication. Would you rather drop it? Or maybe this lets you use it without mise, like you know, this is the source of truth? Probably should drop it? HMmmm I would probalby dig a bit deeper into what this field actually signifies here, then decide.

waspc/dev-tool.project
Martinsos: Yeah I would say let's just drop them. They are now part of hls and while that does not make it the same, as tehy are not run programmatically so easily, I don't tihnk we ever seriously run tham programmatically, so mabye best drop it for now and we can try to reitrnoduce them later.

Btw I am sure these can also be misifed? I guess one could provide instructions how to build them locally?

waspc/README.md
```diff
@@ -50,25 +50,11 @@ Running `./run` without any arguments will print help/usage, which is a good way
...
-
-> [!NOTE]
-> On Mac, we recommend using the official [ghcup](https://www.haskell.org/ghcup/) installer over Homebrew, as it works out of the box.
-
-#### Javascript / TypeScript toolchain
-
-You need to have the correct version of `node` (and `npm`) installed and set.
-
-Check [.nvmrc](.nvmrc) file to learn which version of `node` that should be.
+Install it by following the [official instructions](https://mise.jdx.dev/getting-started.html), then run `mise install` from the repo root to install the required tools.
```
Martinsos: I would rename this to "Dev tooling", and then explain we use Mise, give short explanation what it is, and how we use it, go a bit more verbose. I checked their docs and all and they are friendly, but it was not immediately clear to me what i am dealing wiht. So explaining a bit that we use it to manage the whole toolchain, that you can use it with `mise x` per demand but should in general set it up with `activate`, and maybe a bit more context, would be great. Imagine new member of the team, didn't attend workshop, reading this section, maybe junior.

web/docs/introduction/quick-start.md
```diff
@@ -79,7 +79,7 @@ Check [More Details](#more-details) section below if anything went wrong with th
 You must have Node.js (and NPM) installed on your machine and available in `PATH`.
 <!--
   NOTE: Don't change Wasp's lowest supported Node version without updating it in
-  all required places. Check /.nvmrc for the full list.
+  all required places. Check /mise.toml for the full list.
```
Martinsos, line 82: I wonder why we had "check .nvmrc for the full list" -> why full list? I mean it was never a list there, was it? We kept it though -> do we have a list now? List of what, all places of where it is used? I doubt it, I would always just grep for that specific node version hm.

web/src/remark/search-and-replace.ts
Martinsos: Yeah taht would be nice wouldn't it. But there are quite a few places and I doubt all fot hem would allow reading it.

.nvmrc
```diff
@@ -1,10 +0,0 @@
-v24.14.1
-# IMPORTANT: This should be set to the lowest Node version Wasp claims to
-# support. The CI reads this file when running tests.
-# If you change Wasp's lowest official supported Node version here, you must
-# also change it in:
-#   - The compiler (Wasp.Node.Version).
```
Martinsos, line 6: Ok I take it back we actually had a list! Hah. I would in general be against this though, this will get unmaintained sooner or later, and gives false sense of "ok all is listed here", better let people/AI grep.

mise.toml
```diff
@@ -0,0 +1,76 @@
+min_version = "2026.6.11"
+
+[tools]
+"flyctl" = "0.3.164"
+"github:tfausak/cabal-gild" = "1.5.0.1"
+"cabal" = "3.12.1.0"
+"ormolu" = "0.7.2.0"
+"railway" = "4.11.1"
```
Martinsos: Maybe we could sort/groups these by some logic, fly and railway together, cabal ormolu and cabal-gild together, and similar?
We don't have to, but might be nice, I probably would.

mise.toml
```diff
@@ -0,0 +1,76 @@
...
+
+# The `mise-ghcup` plugin declares `aqua:ghcup` as a dependency specifically
+# instead of just `ghcup` to avoid conflicts, so we need to declare it like that
+# here as well.
```
Martinsos: Hm I don't really get this statement. Ok obviously I don't know what a plugin is or how it declares deps, but how does this help it avoid conflicts, what kind of conflicts, why does it need to avoid them? "so we need to delcare it like that here as well" -> like what?

mise.toml
```diff
@@ -0,0 +1,76 @@
...
+"aqua:ghcup" = "0.2.6.1"
+
+# Keep in sync with:
+# - /waspc/cabal.project
+# - /waspc/dev-tool.project
+"mise-ghcup:ghc" = "9.6.7"
+
+# Make sure that this version of HLS has a pre-built binary for the GHC version
+# we use above.
```
Martinsos: So when you say it like this, "Make sure that ...", I read it as "line below is making sure that ...". But that is not so, instead you want to communicate that we should ensure the version specified in the line below satisfies some conditions. I woudl then phrase that as "NOTE: We assume that the version of mise-ghcup:hls specified below has a pre-built binary for the GHC version we use above.".

Martinsos: Btw how can we know that? Let's say I want to upgrade versions, how will I check it does have a pre-built binary?

mise.toml
```diff
@@ -0,0 +1,76 @@
...
+"mise-ghcup:hls" = "2.12.0.0"
+
+# Keep in sync with:
+# - The compiler (Wasp.Node.Version).
+# - The docs: /docs/introduction/quick-start.md ("Requirements" section).
+# - The docs plugin: /web/src/remark/search-and-replace.ts
```
Martinsos, line 23: I would drop the list and just warn thtere are multiple places in the codebase that duplicate this and to make sure to also update them. Because list risks becoming incorrect and I don't want human or AI relying on it.

mise.toml
Martinsos: Sure, let's pull it in into wasp-lang org! Sounds good.

Martinsos: Why is this vfox: needed as prefix? And what is a plugin?

mise.toml
Martinsos, line 34: It kind of sucks we have to list and later maintain this.
This is were Nix is nicer, because this is handled for us.
But I guess it is small enough scope and tools change slow enough that it won't be an issue for us.

mise.toml
```diff
@@ -0,0 +1,76 @@
...
+"apk:zlib-static" = "latest"
+# Prisma dependencies
+"apk:openssl" = "latest"
+
+[settings]
+lockfile = true
+experimental = true # Needed for backend plugins (mise-ghcup) to work.
+
+[settings.node]
+# The Node backend has a knack for compiling from source on Alpine even when
```
Martinsos, line 75: And why do we prefer prebuilt binary?

Martinsos (commented):
Nice stuff! I forgot to put RAW on comments, but a lot of them are RAW. Some are not though. I think you will figure it out easily from the comments themselves.

.github/actions/setup-env/action.yaml
cprecioso: Yep, it was a weird mix of concerns. I've instead separated the Cabal cache parts into its own action, so everything feels better contained.

.github/actions/setup-env/action.yaml
cprecioso: To be honest I am not exactly sure why this happens either. The comment is from a bunch of iterations so I lost the plot a bit in this comment haha.

My hypothesis is that on Windows it is somehow checking the EXE file metadata of the shim instead of calling it. The shim does _not_ have the correct EXE file metadata (as it's just a simlink to mise.exe), so Cabal refuses to run.

I've updated the comment to be less specific to the problem, since I don't want to devote too much more time since the workaround works fine (I even installed Windows on an old Intel MacBook to see if I could replicate and I wasn't able to). The problem is that Cabal is somehow not detecting the shim's version correctly, so we'll point it to the correct binary directly.

.github/actions/setup-tools/action.yaml
cprecioso: > But what if I want to do local development on Windows?

This issue is bypassed locally since Mise will work with hooks and not shims.

.github/actions/setup-env/action.yaml
cprecioso: This wasn't AI, it was CI (carlos intelligence). But I love it yeah haha

.github/actions/setup-env/action.yaml
cprecioso: I realized I could just use `-v0` to make cabal not print any messages, so I can drop the workaround

.github/actions/setup-env/action.yaml
cprecioso: We were actually getting them from mise, they had `mise tool ghc --requested` to get the version 😁
But at any rate, I refactored this part and now it knows nothing about Mise https://github.com/wasp-lang/wasp/pull/3940#discussion_r3475182980

.github/actions/setup-tools/action.yaml
cprecioso: Yep, I was testing if this was still needed and never got it back. The `dist-newstyle` reasons are still applicable, so I dropped it from the cache and re-added the commment. The `~/.cabal/packages` one we can now cache, so I added it back and didn't add the comment (main problem was it was located in a different path for every system, but we now do a "probe" before so we can know where it is).

.github/actions/setup-env/action.yaml
cprecioso: The `setup-haskell` action did it internally.

.github/actions/setup-tools/action.yaml
cprecioso: Yep, I decided to just make everything come through here. We lose some time on e.g. Haskell when only trying to run a formatter, but everything is cached together now so we win time+space there. Plus it reduces complexity, we know that all the tools are always there, installed in the same way.
I checked CI times and there isn't any big effect on install times (positive nor negative).

.github/actions/setup-env/action.yaml
cprecioso: The MUSL detection is different in this version of the cache. It wasn't caching properly before, and mixing LIBC versions (which is why it sometimes grew too large).
We want to stop using it, so we add this prefix to the cache to completely unlink it and let it naturally fall off when we stop using ti.

.github/workflows/ci-deploy-test.yaml
cprecioso: I will not do any changes on fly or railway in this PR and we can decide at some later point

.github/workflows/ci-formatting.yaml
cprecioso: Resolving in favor of https://github.com/wasp-lang/wasp/pull/3940#discussion_r3475312712

.github/workflows/ci-npm-package-test.yaml
cprecioso: Yep

.github/workflows/ci-waspc-build.yaml
cprecioso: It was never on the .nvmrc so it got lost there. I dropped this argument so dropped the comment without thinking. I added it in mise.toml.

.github/workflows/ci-waspc-test.yaml
cprecioso: This comment originally came from `ci-waspc-build`, but it was not applicable here as we're not building publishable artifacts.

.github/workflows/ci-waspc-test.yaml
cprecioso: Kind of unneeded, because we were then undoing this on windows or linux. We can just use the default shell for each OS and evertyhing works fine without overrides.

.github/workflows/ci-waspc-test.yaml
cprecioso: Simplified it quite a bit

waspc/cabal.project
cprecioso: > Or maybe this lets you use it without mise, like you know, this is the source of truth? 

This was never the source of truth for what GHCup installed, just makes Cabal warn if we don't have the correct version. We can drop safely, we assume that our devs have the correct versions of stuff installed (be it through mise or because you manually installed it).

waspc/dev-tool.project
cprecioso: > Btw I am sure these can also be misifed? 

They could, yeah, but we'd have to write a plugin ourselves to do a kind of 'cabalx' (like 'npx' but for cabal). More hassle than I want right now.

> I guess one could provide instructions how to build them locally?

Right now `./run` is kind of those instructions.

I think I will not touch these for now and then we can see in #3306 what we want to do, instead of mushing that into this PR.

mise.toml
cprecioso: Mise has many plugins by default. They teach mise where to pull the binaries from and how. e.g.:

| Tool | Plugin | Behavior |
|------|--------|----------|
| `npm:@railway/cli` | `npm` | Downloads [the npm package](https://www.npmjs.com/package/@railway/cli) and links its `package.json#bin` script |
| `github:tfausak/cabal-gild` | `github` | Gets [the github releases of the repo](https://github.com/tfausak/cabal-gild/releases) and uses heuristics to find the correct release for the current platform, and a binary inside of it |
| `aqua:node` | `aqua` | Checks [the aqua registry yaml file](https://github.com/aquaproj/aqua-registry/blob/main/pkgs/nodejs/node/registry.yaml) and interprets it to find the correct download |

In this case, we want mise to use `ghcup` to install and download GHC and HLS, but by default it doesn't know how, so we created a plugin (https://github.com/wasp-lang/mise-ghcup) that teaches it to run GHCup and make it output a GHC/HLS installation where it expects it.

The `vfox` prefix is because mise supports [vfox-style plugins](https://mise.jdx.dev/dev-tools/backends/vfox.html) (coded in Lua, supported on Windows), and [asdf-style plugins](https://mise.jdx.dev/dev-tools/backends/asdf.html) (coded in Bash, don't support Windows), and we're using the vfox-style, so we tell Mise what style it should use.

mise.toml
cprecioso: It's much much faster, and we don't need build tools installed for it.

mise.toml
cprecioso: Yep, this does suck a bit. But I think it's good that now it's here in a known place instead of being repeated in a handful of workflows.

.github/actions/setup-env/action.yaml
cprecioso: Good news, this workaround is not needed anymore as of [mise 2026.06.11](https://github.com/jdx/mise/releases/tag/v2026.6.11#:~:text=%40jdx\).-,shim%20\(Windows\),-%3A%20Default%20exe). Apparently the exact problem was that mise on windows generated both a`ghc-x.y.z.exe` shim and a `ghc-x.y.z` shim, and that was tripping up Cabal (and apparently other tools), so they removed the extension-less one and now there's no need for the workaround.

mise.toml
cprecioso: We've changed the requirement twice already since Filip added this list and it's been of great help actually. I would keep it.

.github/actions/setup-tools/action.yaml
cprecioso: Resolving this thread in favor of https://github.com/wasp-lang/wasp/pull/3940#discussion_r3480175025

waspc/cabal.project
Martinsos: Yeah it feels correct having cabal at least being able to produce warning like that! But ok, we are saying that is now specific in mise and that is it, makes sense. Not worth the duplication.
raw

waspc/dev-tool.project
Martinsos: Ok! Although you probably have enough context now to remove them finally and who knows when will somebody else finally have it heh, so it is tempting to do it now :D. Maybe you can do 3306 soon? or just cut then now and if we will want them we will complain? But all ok, whatever you do is fine with me.

RAW

mise.toml
Martinsos: Awesome thanks! Aha, so this is a plugin for installing stuff via ghcup basically. Ok cool I like it!

RAW

.github/actions/setup-env/action.yaml
Martinsos: Awesome! Resolve when you update the code (if you hven't yet).

.github/actions/setup-env/action.yaml
Martinsos: Oh ok! And what is our reasoning! Ok I guess we are pre-warming the cabal cache hm yeah, because initial cabal-update does take long time.

raw

.github/actions/setup-tools/action.yaml
Martinsos: What does get lost is explicit dependencies of what each CI action really needs, I guess. So for example removing a specific dep from this action might be tricky, because we are not sure what will be the effect of it, which CI actions expect it.

One idea of possibly having the best of both worlds is to keep it centralized like this, but make it declarative, so that action exposes "knobs" for turning on stuff, and each CI action turns on only what it needs. Which could work if Mise also exposes that level of granularity, I guess it does. Then we do have a cenrtalzied config, but consumer can be specific about they want and get only that.

That is some extra work though and I don't know how elegant it is to do. If it is ugly, it is likely not worth it.

RAW

.github/actions/setup-env/action.yaml
Martinsos: Makes sense, but it sounds like it has to do something with v2 of cabal :D. Even a small comment explaining what v2 stands for might help, or naming it cacheV2 or something.

RAW

.github/workflows/ci-waspc-test.yaml
Martinsos: Hm but are we doing with this comment then? Is there another palce where it should be mentioned?
It is saying we use old Ubuntu version to ensure compatibility with GLIBC. We are still doing this, right? So I guess it should be mentioned somewhere? Or it is already mentione don the rigth place and this was just copy that can now go away? If so, RAW.

mise.toml
Martinsos: Ok, empirical data! Still has the risk of us or AI trusting it too much. Maybe slightly adjusting it to say these are the places we are aware of, but there could be more? I guess the word is "non-exclusive list"?

But I don't get why one couldn't just grep for the node version and get the same result, why we need a list hm. Ok I guess a bit less results but much more robust.

Funnily I think I remember agreeing with Filip to but the list in the past here, maybe even suggesting it, but maybe I am hallucinating.

RAW

Martinsos:
Did a quick round of review again, bust just focusing on comments, not code. Will do a quick check through code once you proclaim it ready.

.github/workflows/ci-waspc-test.yaml
cprecioso: > Or it is already mentione don the rigth place and this was just copy that can now go away? If so, RAW.

yep, it's in `ci-waspc-buil.yaml`!

.github/actions/setup-tools/action.yaml
cprecioso: Mmmm I have an idea how we can do it (we'd have to separate caches too) but this PR has been going on for too long so I'll prefer to do it in a new one later.

waspc/cabal.project
cprecioso: Will drop this in a future PR

waspc/dev-tool.project
cprecioso: I will do it right away after this one, no worries

mise.toml
cprecioso: It was a bit of a brainfart, i dropped it

.github/actions/setup-cabal-cache/action.yaml
```diff
@@ -0,0 +1,91 @@
...
+          # of the build, and easily reusable.
+          |
+          ${{ steps.probe-ghc-cabal.outputs.CABAL_STORE }}
+          ${{ steps.probe-ghc-cabal.outputs.CABAL_INDEX }}
+        key:
+          # The `v2` prefix is used so we don't mix with an older cache that
+          # didn't have a good glibc/musl detection.
+          >
+          ${{
+            format(
```
Martinsos, line 62: Nice use of format!

.github/actions/setup-tools/action.yaml
```diff
@@ -0,0 +1,27 @@
+name: Setup dev tools
```
Martinsos, line 1: Late in the game but maybe better name is `setup-dev-env` / "Setup dev env"? You will know the best, just seemed more precise after reading the comment below, maybe not important.

RAW

.github/actions/setup-tools/action.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+    - name: Cache npm dependencies
+      uses: actions/setup-node@v6
+      with:
+        # We don't specify a node-version here because we already have it
+        # installed. We will only use this step to setup the npm cache.
+        cache: npm
+        package-manager-cache: true
+
+    - name: Cache Cabal dependencies
```
Martinsos, line 24: Ok nice that this is extracted now!

.github/actions/setup-tools/action.yaml
```diff
@@ -0,0 +1,27 @@
...
+      uses: jdx/mise-action@v4
+      with:
+        bootstrap: true
+        bootstrap_args: --update --yes
+
+    - name: Cache npm dependencies
+      uses: actions/setup-node@v6
+      with:
+        # We don't specify a node-version here because we already have it
+        # installed. We will only use this step to setup the npm cache.
```
Martinsos:
```suggestion
        # installed. We use this step to setup the npm cache.
```

NITPIIICKKKK

waspc/README.md
Martinsos, line 1: Awesome, this will help people like me :D. Maaaaybe I wouldn't mind one or two more sentences on how exactly to do the `activate` thing, but I guess I can figure it out in the docs. Also, there is a recommended way almost everyboyd will do which is `activate` right? The rest is more situational and we probably won't use it ften. So I would make it very clear what is recommended workflow for this project and give more attention to it, the rest you can literally just list and tell me to go look in the docs.
Yeah actually this would be good.

mise.toml
```diff
@@ -0,0 +1,77 @@
...
+# https://haskell-language-server.readthedocs.io/en/latest/support/ghc-version-support.html
+# to see which combination of GHC and HLS versions are compatible.
+"mise-ghcup:hls" = "2.12.0.0"
+
+# This should be equal to Wasp's lowest supported Node version.
+# Keep in sync with:
+# - The compiler (Wasp.Node.Version).
+# - The docs: /docs/introduction/quick-start.md ("Requirements" section).
+# - The docs plugin: /web/src/remark/search-and-replace.ts
+# (Also just grep for the version in case we forgot somewhere.)
```
Martinsos, line 24: Sweet, no mistakes from AI reading this!

Martinsos (approved):
LGTM @cprecioso ! This will be nice.
I approved, there are still a few comments but do with those as you wish and merge. Yay!

PR #3982, opened 2026-03-26
Add prerender field to Route and wire to Vite SSR plugin

### LINKED ISSUE #3895: [Deep Freeze] Connect prerendering field
## Goal

Wire the agreed-upon prerendering field into the Wasp compiler and build pipeline.

## Scope

- [ ] Parse the new prerender field from the Wasp DSL (or TS Spec, whichever is on `main`)
- [ ] Validate: emit a compiler error if the field is set on a page with URL parameters
- [ ] Pass the list of pages marked for prerendering to the build pipeline

cprecioso:
## Description

Adds an optional `prerender` boolean field to the Route declaration type (`false` by default). When a user sets `prerender: true` on a route, that route's path is collected and passed to the Vite SSR plugin as a prerendering path (populating the previously empty `ssrPaths` array).

- Creates prerendering test pages in `kitchen-sink`, checking that the page is correctly prerendered and no hydration mismatches occur.
- Adds prerendering of `/` to kitchen-sink, waspleau, and ask-the-documents (the ones that don't have an auth-required main page).
- Change the lazy routes to use `React.lazy` instead of React Router's native lazy handling. [It has a bug](https://github.com/remix-run/react-router/issues/14955) that caused React to throw away the prerendering.
- We tell the client pages if they were rendered as a fallback or not (so there are no hydration mismatches). This is done by adding a Wasp-controlled way of passing data from the prerendering to the rendering stage, similar to React Router's way.
- Simplified the SSR entry by just deferring to the Layout component
- Fixed a hydration mismatch in our WebSocket support

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

    - [x] _(if you updated `examples/tutorials`)_ I updated the tutorial in the docs (and vice versa).

- 📜 Documentation:


waspc/data/Generator/templates/sdk/wasp/client/app/hooks/useIsClient.ts
cprecioso, line 1: Much simpler and easy to understand. Plus it `useSyncExternalStore` forces synchronous rendering, which might make Suspense fallbacks appear. Using an effect is maximally concurrent and [the recommended pattern by React](https://react.dev/reference/react-dom/client/hydrateRoot#handling-different-client-and-server-content).

waspc/data/Generator/templates/sdk/wasp/client/app/router.tsx
```diff
@@ -48,17 +47,3 @@ export function getRouteObjects({
...
-}: {
-  routesMapping: RouteMapping,
-  rootElement: ReactNode,
-}) {
-  const routeObjects = getRouteObjects({ routesMapping, rootElement })
-  const browserRouter = createBrowserRouter(routeObjects, {
-    basename: '{= baseDir =}',
-  })
-  return <RouterProvider router={browserRouter} />;
-}
```
cprecioso, line 64: We weren't using this function anywhere, better to keep the different kinds of router creation in the SSR and Client entries.

examples/kitchen-sink/src/components/FeatureContainer.tsx
```diff
@@ -152,6 +152,16 @@ const features: Feature[] = [
...
+  {
+    to: "/prerender",
+    title: "Prerendering",
+    isPublic: true,
+  },
+  {
+    to: "/hydration-mismatch",
+    title: "Hydration Mismatch",
+    isPublic: true,
+  },
```
cprecioso, line 164: I created an issue to work on improving organization in this sidebar, as it risks becoming unwieldly
https://github.com/wasp-lang/wasp/issues/4043

examples/kitchen-sink/src/App.tsx
```diff
@@ -181,9 +181,7 @@ function StatusInfo({
         </span>
       </div>
 
-      {date && (
-        <div className="text-gray-500">Loaded {date.toLocaleString()}</div>
-      )}
+      {date && <div className="text-gray-500">Loaded {date.toISOString()}</div>}
```
cprecioso, line 184: This is needed because the locale might be different between the server and the client, so it could produce a hydration mismatch.

waspc/data/Generator/templates/sdk/wasp/client/app/hooks/useIsClient.ts
```diff
@@ -1,22 +1,16 @@
...
-  // like a hack, but it conforms to the semantics of `useSyncExternalStore`,
-  // with *the environment* being the "external store" in this case. We just
-  // don't have any real subscription logic, since the value is static.
-  return useSyncExternalStore(emptySubscribe, getClientValue, getServerValue)
-}
+  const [isClient, setIsClient] = useState(false);
+
+  useEffect(() => {
+    setIsClient(true);
+  }, []);
```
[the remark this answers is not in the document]
cprecioso: Explained here https://github.com/wasp-lang/wasp/pull/3982#discussion_r3052634246

waspc/src/Wasp/Generator/SdkGenerator/Client/VitePluginG.hs
```diff
@@ -44,16 +46,20 @@ genViteIndex = return $ C.mkTmplFd tmplPath
...
+genWaspPlugin :: AppSpec -> Generator FileDraft
+genWaspPlugin spec = return $ C.mkTmplFdWithData tmplPath tmplData
   where
     tmplPath = C.vitePluginsDirInSdkTemplatesDir </> [relfile|wasp.ts|]
     tmplData =
       object
         [ "clientEntryPointPath" .= clientEntryPointPath,
           "ssrEntryPointPath" .= ssrEntryPointPath,
-          "spaFallbackFile" .= spaFallbackFile
+          "spaFallbackFile" .= spaFallbackFile,
```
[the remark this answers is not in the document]
cprecioso: Will do in a following PR

examples/kitchen-sink/e2e-tests/tests/prerender.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    await page.goto("/prerender");
+
+    await expect(page.getByTestId("render-location")).toHaveText("server");
+  });
+
+  test("prerender page hydrates on the client", async ({ page }) => {
+    await page.goto("/prerender");
+
+    await expect(page.getByTestId("render-location")).toHaveText("client");
+  });
```
FranjoMindek, line 19: Note for myself: not really a dangerous race condition.
The `await expect(page.getByTestId("render-location")).toHaveText("client");` line has a timeout, so it should be fine.

examples/kitchen-sink/e2e-tests/tests/prerender.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    expect(hydrationWarning.hasHappened()).toBe(true);
+  });
+});
+
+function checkHydrationWarning(page: Page) {
+  let hasHappened = false;
+
+  function listener(msg: Error) {
+    hasHappened ||=
```
FranjoMindek, line 50: First time I see a valid `||=` use case. Nice.

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/ssr-entry.tsx
```diff
@@ -1,64 +1,56 @@
...
-    return await appToHtml(
-      { isFallbackPage: false, children: <RouterProvider router={router} /> },
-      ctx,
-    );
-  }
-}
+  const router = createStaticRouter(dataRoutes, context);
 
-export default prerenderApp;
+  const WASP_SSR_DATA: WaspSSRData = { isFallbackPage }
```
FranjoMindek: Nitpickiest nitpick, move it below `function App`, above `html`.
But we can move `isFallbackPage` here, if it's not used before.

waspc/data/Generator/templates/sdk/wasp/client/webSocket/WebSocketProvider.tsx
```diff
@@ -49,9 +49,14 @@ export const WebSocketContext: Context<WebSocketContextValue> = createContext<We
 
 // PRIVATE API
 export function WebSocketProvider({ children }: { children: ReactNode }) {
-  const [isConnected, setIsConnected] = useState(socket.connected)
+  const [isConnected, setIsConnected] = useState(
+    // In the first render this should be false so we avoid hydration mismatches.
```
FranjoMindek: I would just word it differently, e.g. in sense of that our hooks need to be SSR-safe, rather than it avoids hydration mismatch. 

Mostly since I think more people will understand SSR-safe over avoid hydration mismatches.

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -434,6 +436,36 @@ validateAtLeastOneRoute spec =
   where
     routes = AS.getRoutes spec
 
+validatePrerenderRoutes :: AppSpec -> [ValidationError]
+validatePrerenderRoutes spec =
+  concatMap validateRoute prerenderRoutes
+  where
+    prerenderRoutes = filter ((== Just True) . Route.prerender . snd) (AS.getRoutes spec)
+
+    validateRoute (routeName, route) =
```
FranjoMindek:
```suggestion
  concatMap validatePrerenderRoute prerenderRoutes
  where
    prerenderRoutes = filter ((== Just True) . Route.prerender . snd) (AS.getRoutes spec)

    validatePrerenderRoute (routeName, route) =
```

waspc/src/Wasp/AppSpec/Valid.hs
```diff
@@ -434,6 +436,36 @@ validateAtLeastOneRoute spec =
...
+                ++ "' has prerender enabled but its page has authRequired set to true."
+                ++ " Prerendered routes cannot require authentication."
+            | pageRequiresAuth
+          ]
+        ]
+      where
+        (_, page) = AS.resolveRef spec (Route.to route)
+        pageRequiresAuth = Page.authRequired page == Just True
+
+    hasDynamicSegments path = any (`elem` path) [':', '*', '?']
```
FranjoMindek: Maybe move `hasDynamicSegments` to inner `where`?
I know you have other stuff in inner `where` because it has to access the `validateRoute` params.

But since it already exists, and we don't use `hasDynamicSegments` outside of `validateRoute` we can move it inside?

RAW

waspc/tests/AppSpec/ValidTest.hs
```diff
@@ -468,6 +468,68 @@ spec_AppSpecValid = do
...
+
+      it "returns an error for prerendered route with optional segment" $ do
+        let errors = ASV.validateAppSpec (makeSpec "/photo/:id/edit?" (Just True) Nothing)
+        length errors `shouldBe` 1
+        -- Has both : and ?, but we just check at least one error about dynamic segments
+        show (head errors) `shouldSatisfy` ("dynamic segments" `isInfixOf`)
+
+      it "returns an error for prerendered route pointing to authRequired page" $ do
+        let errors = ASV.validateAppSpec (makeSpec "/dashboard" (Just True) (Just True))
+        -- One error from validateAppAuthIsSetIfAnyPageRequiresAuth (app.auth not set)
```
FranjoMindek, line 522: Weird that there is no `basicAuthAppSpec`.

web/docs/advanced/prerendering.md
```diff
@@ -0,0 +1,147 @@
...
+---
+
+By default, Wasp apps are single-page applications: the browser downloads JavaScript, and React renders the page on the client. This means search engines, AI crawlers, and users on slow connections see a blank page until JavaScript loads and executes.
+
+Wasp can **prerender** specific routes at build time, producing static HTML files that are served immediately. The page then hydrates on the client for full interactivity.
+
+This gives you:
+
+- **Better SEO:** search engines index real HTML content instead of an empty shell.
+- **LLM and AI readability:** AI crawlers (ChatGPT, Perplexity, etc.) can read your content directly.
```
FranjoMindek: No `Claude` here, but we do mentioned it below 😢

web/docs/advanced/prerendering.md
```diff
@@ -0,0 +1,147 @@
...
+}
+```
+
+:::caution
+Wasp reports an error at compile time if you try to prerender a route with a dynamic path or an auth-required page.
+:::
+
+## Troubleshooting
+
+### Hydration mismatches
```
FranjoMindek, line 85: Let's also link to useful titles in React docs related to these problems. We should delegate it to React, if possible.

Like, we give short practical thing, and then have a read more here link to React docs title.

e.g.
https://react.dev/reference/react-dom/client/hydrateRoot#suppressing-unavoidable-hydration-mismatch-errors
https://react.dev/reference/react-dom/client/hydrateRoot#handling-different-client-and-server-content

web/docs/advanced/prerendering.md
```diff
@@ -0,0 +1,147 @@
...
+
+When React hydrates a prerendered page, it expects the prerendered HTML to match what the client renders. If they differ, React logs a warning and may discard the prerendered HTML, losing the performance benefits.
+
+#### Common causes
+
+- **Checking for `window` or `document`:** code like `typeof window !== 'undefined'` or `import.meta.env.SSR` returns different values on the prerender vs. the client, and might change everything that depends on it.
+- **Non-deterministic values during render:** functions like `Date.now()`, or `Math.random()` produces different results on each render.
+- **Browser-only APIs:** accessing `window.innerWidth`, `navigator.userAgent`, `localStorage`, or similar APIs during render will fail while prerendering. This also applies to some third-party libraries that access these APIs, or less obvious JS APIs like `Intl.DateTimeFormat` that can use different timezones and locales on the prerender vs. client.
```
FranjoMindek:
```suggestion
- **Browser-only APIs:** accessing `window.innerWidth`, `navigator.userAgent`, `localStorage`, or similar APIs during render will fail while prerendering. This also applies to some third-party libraries that access these APIs, or less obvious JS APIs like `Intl.DateTimeFormat`, which can use different timezones and locales on the prerender vs. client.
```

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/client-entry.tsx
```diff
@@ -1,23 +1,25 @@
...
+
 import { Layout } from "wasp/client/app/layout";
 import { WaspApp } from "wasp/client/app";
 
 {=& routeObjects.importStatement =}
 
-// React Router will put hydration data on this property of the `window` object.
-// https://reactrouter.com/7.13.1/start/data/custom#4-hydrate-in-the-browser
-const hydrationData = (window as any).__staticRouterHydrationData as HydrationState | undefined;
+// We embed this data at prerendering time
```
FranjoMindek:
```suggestion
// We embed this data at prerendering time.
```

To respect Martin

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/ssr-entry.tsx
```diff
@@ -1,64 +1,56 @@
...
-  { clientEntrySrc }: PrerenderContext,
-) {
-  const app = (
-    <Layout isFallbackPage={isFallbackPage} clientEntrySrc={clientEntrySrc}>
-      <WaspApp>
-        {children}
-      </WaspApp>
-    </Layout>
-  );
+  function App() {
```
FranjoMindek, line 35: So we can always pass children to `Layout` since layout will handle if it renders them or not.

But still, don't we do unnecessary work when we in fact have fallback page?
This is not really just a simplification but "regression"? Or am I missing something?

waspc/data/Generator/templates/sdk/wasp/vite-env.d.ts
```diff
@@ -1 +1,10 @@
 /// <reference types="vite/client" />
+
+interface WaspSSRData {
```
FranjoMindek, line 3: I would love if we could do:

```ts
/// <reference types="wasp" />
```

FranjoMindek (changes requested):
Nicely done.
I only have one question regarding the `ssr-entry` changes. The rest seems fine, and I just have some additional suggestions.

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/ssr-entry.tsx
cprecioso: No, not really. React will never call the component's function until it is part of a rendered tree. There's no overhead in doing `<WaspApp>` if it is never rendered.
https://codepen.io/cprecioso/pen/pvEQYZz?editors=0011

waspc/src/Wasp/AppSpec/Valid.hs
cprecioso: I remodeled it a bit so I avoid the nested `where`

waspc/data/Generator/templates/sdk/wasp/vite-env.d.ts
cprecioso: I'll deem this out of scope for now 😅

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/ssr-entry.tsx
FranjoMindek: No I meant like the `!isFallbackPage` path of `prerenderApp`.

Doing `createStaticHandler`, `new Request`, `await query`... to `router` even if it's a fallback page.

Not really a huge thing as `prerenderApp` is not really run that often (I assume, not sure 100% how dev works, but for prod it should be only on build?).

Resolve it as you will, you will have more context.
I just wanted to ask if this is important in any way.

FranjoMindek (approved):
I like the changes.
Let's go!

waspc/data/Generator/templates/sdk/wasp/client/vite/virtual-files/files/ssr-entry.tsx
cprecioso: Ahh I understand now. Yeah, we could avoid do it. I had it like that in an earlier version of the PR. I'd like to keep it, to ensure we don't miss anything if we add something later. And the code is clearer if there's only one path to follow. But anyway, given that we don't have any loaders, all this amounts to a no-op almost.

PR #3998, opened 2026-03-31
Replace axios with native fetch in SDK api module

### LINKED ISSUE #2935: Remove or upgrade to latest axios
We should upgrade our `axios` dependency, as it seems to have severe security vulnerabilities. (https://github.com/advisories/GHSA-wf5p-g6vw-rhxx)

cprecioso:
## Description

Remove the `axios` dependency from the SDK and replace it with a native `fetch`-based `api` function. The new `api(input, init?)` function takes the same parameters as the `Request` constructor, prepends the API base URL for relative paths, adds Bearer auth headers, handles 401 session invalidation (with the same race-condition protection), and throws `WaspHttpError` on non-2xx responses — merging the old `handleApiError` logic directly into the function. All consumer files are updated to use the new calling convention (`api(path, { method, headers, body })` + `response.json()` instead of `api.post()`/`api.get()` + `response.data`).

- Fixes #2935

## Type of change

- [ ] **🐞 Bug fix**
- [x] **💥 Breaking change**

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:

🤖 Generated with [Claude Code](https://claude.com/claude-code)


waspc/data/Generator/templates/sdk/wasp/api/index.ts
```diff
@@ -1,14 +1,7 @@
...
-
 import { config } from 'wasp/client'
 import { storage } from 'wasp/core/storage'
 import { apiEventsEmitter } from './events.js'
 
-// PUBLIC API
-export const api: AxiosInstance = axios.create({
-  baseURL: config.apiUrl,
-})
-
```
cprecioso, line 11: Note to self: update public API table

waspc/data/Generator/templates/sdk/wasp/api/index.ts
```diff
@@ -130,11 +119,18 @@ class WaspHttpError extends Error {
...
   const prefix = 'Bearer '
   if (header && header.startsWith(prefix)) {
     return header.substring(prefix.length)
   } else {
     return null
   }
 }
+
+function toAbsoluteUrl(input: RequestInfo | URL): RequestInfo | URL {
+  if (typeof input === 'string' && input.startsWith('/')) {
```
infomiho, line 132: Axios worked with `foo/bar` (without a leading slash), so I think we need to either point out that we don't handle this or maybe handle this case.

Based on [Axios logic](https://github.com/axios/axios/blob/e2bed7f84d86e87d0641bc5910e96607092d9934/lib/core/buildFullPath.js):

```ts
function toAbsoluteUrl(input: RequestInfo | URL): RequestInfo | URL {
    if (typeof input !== 'string') return input

    // Absolute URLs (http://, https://, //) pass through
    if (/^([a-z][a-z\d+\-.]*:)?\/\//i.test(input)) return input

    // Relative URLs get baseURL prepended
    // Strip trailing slash from base, leading slashes from input, join with /
    return config.apiUrl.replace(/\/+$/, '') + '/' + input.replace(/^\/+/, '')
  }
  ```

File: waspc/data/Generator/templates/sdk/wasp/api/index.ts [hunk 3]
```diff
@@ -36,44 +29,63 @@ export function removeLocalUserData(): void {
...
+
+  if (!response.ok) {
+    let message = response.statusText
+    let data: unknown
+    try {
+      const responseJson = await response.json()
+      message = responseJson?.message ?? message
+      data = responseJson
+    } catch {
+      // Response body is not JSON, use statusText as message
```
infomiho, line 82: To match Axios, we could use the raw body as `data`. This is Claude suggestion.

waspc/data/Generator/templates/sdk/wasp/api/index.ts
```diff
@@ -95,30 +107,7 @@ if (typeof window !== 'undefined') {
 }
 
 // PRIVATE API (sdk)
-/**
- * Takes an error returned by the app's API (as returned by axios), and transforms into a more
- * standard format to be further used by the client. It is also assumed that given API
- * error has been formatted as implemented by HttpError on the server.
- */
-export function handleApiError<T extends AxiosError<{ message?: string, data?: unknown }>>(error: T): T | WaspHttpError {
```
infomiho, line 103: I guess it's not a breaking change since we never advertised this, and we'll be throwing the `WaspHttpError` directly which is nice.

waspc/data/Generator/templates/sdk/wasp/api/index.ts
```diff
@@ -95,30 +107,7 @@ if (typeof window !== 'undefined') {
...
-    const responseStatusCode = error.response.status
-    return new WaspHttpError(responseStatusCode, responseJson?.message ?? error.message, responseJson)
-  } else {
-    // If any other error, we just propagate it.
-    return error
-  }
-}
-
-class WaspHttpError extends Error {
+export class WaspHttpError extends Error {
```
infomiho, line 110: We'd probably want to consider this now public API, since you'd need it for `instanceof`? Maybe note this in the migration guide as well.

infomiho (commented):
I really like the direction, dropping `axios`, going for native `fetch`!

I dislike the DX degradation for e.g. `api.post(route, data)` which now needs a headers and `JSON.stringify` for payload. 

Could we do something like using `ky` or `ofetch` or even go for a Axios-like wrapper of our own?

To paint the picture, here's some code, take it or leave it, let's discuss!

## Alternative: minimal axios-like API on top of fetch

Keep `api.get()` / `api.post()` / `response.data` working so users don't need to change their calling code. Only error handling breaks (`AxiosError` → `WaspHttpError`).

Streaming uses `api(url, init)` which returns a raw `Response`.

<details>
<summary><code>api/index.ts</code></summary>

```ts
import { config } from 'wasp/client'
import { storage } from 'wasp/core/storage'
import { apiEventsEmitter } from './events.js'

const WASP_APP_AUTH_SESSION_ID_NAME = 'sessionId'

// ... setSessionId, getSessionId, clearSessionId, removeLocalUserData unchanged ...

// ── Types ──────────────────────────────────────────────────────

export interface ApiResponse<T = any> {
  data: T
  status: number
  statusText: string
  headers: Headers
}

interface RequestOptions {
  headers?: Record<string, string>
  signal?: AbortSignal
}

interface Api {
  /** Raw fetch access (for streaming, custom needs) */
  (input: RequestInfo | URL, init?: RequestInit): Promise<Response>

  get<T = any>(url: string, options?: RequestOptions): Promise<ApiResponse<T>>
  post<T = any>(url: string, body?: unknown, options?: RequestOptions): Promise<ApiResponse<T>>
  put<T = any>(url: string, body?: unknown, options?: RequestOptions): Promise<ApiResponse<T>>
  delete<T = any>(url: string, body?: unknown, options?: RequestOptions): Promise<ApiResponse<T>>
}

// ── Implementation ─────────────────────────────────────────────

async function rawFetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
  const request = new Request(toAbsoluteUrl(input), init)

  const sessionId = getSessionId()
  if (sessionId !== null) {
    request.headers.set('Authorization', `Bearer ${sessionId}`)
  }

  const response = await fetch(request)

  if (response.status === 401) {
    const failingSessionId = getSessionIdFromAuthorizationHeader(
      request.headers.get('Authorization')
    )
    const currentSessionId = getSessionId()
    if (failingSessionId === currentSessionId) {
      clearSessionId()
    }
  }

  return response
}

async function jsonRequest<T>(
  method: string,
  url: string,
  body?: unknown,
  options?: RequestOptions,
): Promise<ApiResponse<T>> {
  const headers: Record<string, string> = { ...options?.headers }
  let serializedBody: string | undefined

  if (body !== undefined) {
    headers['Content-Type'] = 'application/json'
    serializedBody = JSON.stringify(body)
  }

  const response = await rawFetch(url, {
    method,
    headers,
    body: serializedBody,
    signal: options?.signal,
  })

  let data: T
  try {
    data = await response.json()
  } catch {
    data = undefined as T
  }

  if (!response.ok) {
    throw new WaspHttpError(
      response.status,
      (data as any)?.message ?? response.statusText,
      data,
    )
  }

  return { data, status: response.status, statusText: response.statusText, headers: response.headers }
}

// PUBLIC API
export const api: Api = Object.assign(rawFetch, {
  get:    <T>(url: string, opts?: RequestOptions) => jsonRequest<T>('GET', url, undefined, opts),
  post:   <T>(url: string, body?: unknown, opts?: RequestOptions) => jsonRequest<T>('POST', url, body, opts),
  put:    <T>(url: string, body?: unknown, opts?: RequestOptions) => jsonRequest<T>('PUT', url, body, opts),
  delete: <T>(url: string, body?: unknown, opts?: RequestOptions) => jsonRequest<T>('DELETE', url, body, opts),
})

// ... WaspHttpError, getSessionIdFromAuthorizationHeader, toAbsoluteUrl unchanged ...
```

</details>

### What breaks

| Pattern | Breaks? |
|---|---|
| `api.get(url)` / `api.post(url, body)` | No |
| `response.data` | No |
| `response.status` | No |
| `catch (e) { (e as AxiosError).response?.status }` | **Yes** → `WaspHttpError.statusCode` |
| `import { AxiosError } from 'axios'` | **Yes** → remove |
| `api.post(url, body, { responseType: 'stream' })` | **Yes** → use `api(url, init)` |
| `api.get(url, { params: ... })` | **Yes** → build query string manually |

cprecioso:
Closed in favor of #4004

PR #4001, opened 2026-03-31
Tie up loose SemVer ends

FranjoMindek:
## Description

Fixes some leftover stuff me and Filip discussed.

- Renamed `fromVersion` to `versionToPartialVersion`.
- `Range` is now a non-empty list instead of a list.
- Removed `Comparator` as concept, since it wasn't a proper `Comparator`. Same for `ComparatorSet` (since it had non-comparator items). Now instead we have a `RangeExpression` which can be either a `HyphenRange` or set of `SimpleRangeExpression`s, closely following the grammar (whose our models are based upon).
   As such `Comparator.hs` and `ComparatorSet.hs` are merged into `RangeExpression.hs`.

So the current naming is:

`Range` - non empty list of `RangeExpression` 

`RangeExpression` - either: 
 - non empty list of `SimpleRangeExpression`
 - `HyphenRange`

`SimpleRangeExpression` - either:
 - `Primitive` (>, <, =, >=, <=)
 - `Caret Range` (^)
 - `Tilde Range` (~)

Ideas for the future:
Currently we depend on `VersionBounds` to translate our "range grammar" models into actual version ranges. While this works (we get the same end result) maybe we can changes to be more in-line with `node-semver`. Instead of `VersionBounds` I would make a type class which transforms grammar models into comparators and then changes the versioning logic to work with comparators as base units (the same way `node-semver` does).

So external modules would still work with `Range` (the rich model) but under the hood we would do the `Comparator` operations. We keep the nice API, while hiding the `Comparator` stuff behind. And if users want to use `Comparator`s directly, they would be able to (now they can't).

## Type of change

- [ ] **🐞 Bug fix** 
- [ ] **💥 Breaking change** 

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:


waspc/tests/SemanticVersion/RangeTest.hs
```diff
@@ -125,14 +135,6 @@ spec_SemanticVersion_Range = do
...
             `shouldBe` map snd versionsWithResults
 
-    it "no version is in empty range" $
-      testRange
-        mempty
-        [ ([v|0.5.5|], False),
-          ([v|1.0.0|], False),
-          ([v|1.2.3|], False),
-          ([v|2.0.0|], False)
-        ]
```
sodic, line 135: I'm guessing this is because Range is no longer a monoid and does not have an identity?

waspc/src/Wasp/SemanticVersion/RangeExpression.hs
```diff
@@ -28,34 +24,24 @@ import Wasp.SemanticVersion.VersionBound
...
 
--- | A comparator set is either a sequence of simple range expressions or a hyphen range.
---
--- The name comparator set comes from the fact that all items of comparator set
--- can be desguared into comparators.
-data ComparatorSet
-  = SimpleComparatorSet (NE.NonEmpty SimpleRangeExpression)
+-- | A range expression is either a set of simple range expressions or a hyphen range.
+data RangeExpression
+  = Simple (NE.NonEmpty SimpleRangeExpression)
```
sodic: Simple -> SimpleRange

waspc/src/Wasp/SemanticVersion/RangeExpression.hs
```diff
@@ -28,34 +24,24 @@ import Wasp.SemanticVersion.VersionBound
...
--- X-Range is already supported on all operators through the 'PartialVersion' implementation.
+-- NOTE: X-Range is already supported on all operators through the 'PartialVersion' implementation.
 data SimpleRangeExpression
   = -- | 1.2.3 (=1.2.3), >1.2.3, <1.2.3, >=1.2.3, <=1.2.3
-    Primitive Comparator
+    Primitive PrimitiveOperator PartialVersion
   | -- | ~1.2.3
     TildeRange PartialVersion
   | -- | ^1.2.3
     CaretRange PartialVersion
```
sodic: These guys have `Range` in their name but are not types of `Range`. That's a little weird.

waspc/src/Wasp/SemanticVersion/RangeExpression.hs
```diff
@@ -64,22 +50,18 @@ data PrimitiveOperator
...
+-- | We rely on this 'show' implementation to produce a valid `node-semver` output.
+instance Show RangeExpression where
+  show (Simple simpleRangeExpressions) = unwords $ show <$> NE.toList simpleRangeExpressions
   show (HyphenRange lower upper) = show lower ++ " - " ++ show upper
 
--- | We rely on this 'show' implementation to produce valid `node-semver` simple comparator.
+-- | We rely on this 'show' implementation to produce a valid `node-semver` output.
 instance Show SimpleRangeExpression where
-  show (Primitive comparator) = show comparator
+  show (Primitive primOp pv) = show primOp ++ show pv
```
sodic: I'd lose the acronyms (I don't know what `pv` or `primOp` means). Some of it is old code, so you decide what to do.

waspc/src/Wasp/SemanticVersion/RangeExpression.hs
```diff
@@ -89,25 +71,20 @@ instance Show PrimitiveOperator where
...
--- Hyphen Ranges can't be combined with other comparator sets.
-instance Semigroup ComparatorSet where
-  (SimpleComparatorSet left) <> (SimpleComparatorSet right) = SimpleComparatorSet $ NE.nub $ left <> right
-  (HyphenRange _ _) <> _ = error "Cannot combine Hyphen Range with other comparator sets"
-  _ <> (HyphenRange _ _) = error "Cannot combine Hyphen Range with other comparator sets"
+-- | We define concatenation of two range expressions as their intersection.
+-- 'HyphenRange' can't be combined with other range expressions.
+instance Semigroup RangeExpression where
+  (Simple left) <> (Simple right) = Simple $ NE.nub $ left <> right
+  (HyphenRange _ _) <> _ = error "Cannot intersect Hyphen Range with other range expressions"
```
sodic: I don't think intersect is the right word here because simple rangers are not intersected. Combine was the correct term, no?

waspc/src/Wasp/SemanticVersion/RangeExpression.hs
```diff
@@ -219,3 +181,15 @@ comparatorParser = primitiveComparatorParser
           Equal <$ P.char '=',
           Equal <$ P.string ""
         ]
+
+-- See `hyphen` definition here: https://github.com/npm/node-semver#range-grammar
+hyphenRangeParser :: P.Parsec String () RangeExpression
```
sodic: `hyphenRangeParser` but it returns a range expression, a little strange.

We have a lot of mixing up of range and range expression. I'll leave it to you to make sense of it (maybe it can stay as is, your choice).

waspc/src/Wasp/SemanticVersion/Range.hs
```diff
@@ -82,61 +76,53 @@ doesVersionRangeAllowMajorChanges = not . doesVersionRangeAllowOnlyMinorChanges
...
-gte = mkComparatorRange GreaterThanOrEqual
+gte = mkPrimitiveRange GreaterThanOrEqual
 
 eq :: Version -> Range
-eq = mkComparatorRange Equal
+eq = mkPrimitiveRange Equal
 
-mkComparatorRange :: PrimitiveOperator -> Version -> Range
-mkComparatorRange op = Range . pure . SimpleComparatorSet . NE.fromList . pure . Primitive . Comparator op . fromVersion
+mkPrimitiveRange :: PrimitiveOperator -> Version -> Range
```
sodic, line 108: Where is this word `primitive` coming from, NPM semver?

sodic (approved):
I have some comments but, in accordance with our culture of shipping, am approving the PR.

You can make all the remaining choices, no need to contact me.

waspc/src/Wasp/SemanticVersion/RangeExpression.hs
sodic: Ok, so it turns out it is an intersection.

waspc/src/Wasp/SemanticVersion/RangeExpression.hs
```diff
@@ -153,25 +132,28 @@ toXRangeLowerBound (Major mjr) = Inclusive $ Version mjr 0 0
...
+
+    -- `node-semver` parses empty input as the equals any range (*).
+    emptyRangeParser :: P.Parsec String () RangeExpression
+    emptyRangeParser = (Simple . pure $ Primitive Equal Any) <$ P.eof
 
--- | Parses a simple range expression.
 -- See `simple` definition here: https://github.com/npm/node-semver#range-grammar
-simpleRangeExpressionParser :: P.Parsec String () SimpleRangeExpression
-simpleRangeExpressionParser =
+simpleRangeParser :: P.Parsec String () SimpleRangeExpression
```
sodic: SimpleRangeExpressionParser

sodic (approved):

PR #4004, opened 2026-04-01
Replace axios with ky in SDK client API module

### LINKED ISSUE #2935: Remove or upgrade to latest axios

cprecioso:
## Description

Replaces the Axios in the SDK's `api` module with a [ky](https://github.com/sindresorhus/ky) instance. This gives us a cleaner, more ergonomic HTTP client with built-in JSON handling, method shortcuts, hooks for interceptors, and automatic error throwing -- all built on top of the modern `fetch` instead of the older XHR.

The `api` export from `wasp/client/api` is now a `ky.extend()` instance configured with:
- `prefixUrl` for the API base URL (replaces the manual `toAbsoluteUrl` helper)
- `beforeRequest` hook for adding the Authorization header
- `afterResponse` hook for 401 session invalidation (with the existing race-condition prevention)
- Native ky `HTTPError` for non-2xx responses (replaces the custom `WaspHttpError` class)
This all mirrors what we had for Axios.

All internal callers are updated to use ky's ergonomic API: `api.post('/path', { json: data }).json(ResponseSchema)`.

- Fixes #2935

## Type of change

- [ ] **🐞 Bug fix**
- [x] **💥 Breaking change**

## Checklist

- 🧪 Tests and apps:

- 📜 Documentation:


waspc/data/Generator/templates/sdk/wasp/api/index.ts
```diff
@@ -36,43 +30,63 @@ export function removeLocalUserData(): void {
...
+        if (isHTTPError(error)) {
+          ;(error as any).statusCode = error.response.status
+          const body = error.data as Record<string, unknown> | undefined
+          if (body && typeof body.message === 'string') {
+            error.message = body.message
+          }
+        }
+        return error
+      },
+    ],
```
cprecioso: This logic was previously part of `handleApiError`, which we were calling at a `catch` in every call site.

waspc/src/Wasp/Generator/DepVersions.hs
```diff
@@ -58,9 +57,6 @@ expressVersionRange = [SV.r|~5.1.0|]
 expressTypesVersionRange :: SV.Range
 expressTypesVersionRange = [SV.r|^5.0.0|]
 
-axiosVersionRange :: SV.Range
-axiosVersionRange = [SV.r|^1.4.0|]
```
cprecioso, line 62: I didn't add the `ky` version here since we're not really sharing it anywhere

waspc/src/Wasp/Generator/SdkGenerator.hs
```diff
@@ -198,10 +197,10 @@ npmDepsForSdk spec =
         Npm.Dependency.fromList
           [ ("@prisma/client", show prismaVersionRange),
             ("prisma", show prismaVersionRange),
-            ("axios", show axiosVersionRange),
             ("dotenv", show dotenvVersionRange),
             ("dotenv-expand", "^12.0.3"),
             ("express", show expressVersionRange),
+            ("ky", "^2.0.0-0"),
```
cprecioso: Flagging: this is a preview version, but it has many QoL improvements we'd like to have (e.g. automatic Standard Schema support, and in previous versions they forced every path _not_ to start with `/`). We can ship with the release version, as it seems to work fine enough, or wait until it is released for real.

waspc/data/Generator/templates/sdk/wasp/auth/responseSchemas.ts
cprecioso, line 1: Given that `ky` types `.json()` responses as `unknown`, we can take advantage of the built-in schema support for good runtime type checking

examples/kitchen-sink/src/features/apis/pages/ApisPage.tsx
```diff
@@ -53,14 +53,12 @@ function useCustomApi<Data = unknown>(
...
+  error: HTTPError | null;
 } {
-  const { isLoading, data, error } = useQuery<Data, AxiosError>(
+  const { isLoading, data, error } = useQuery<Data, HTTPError>(
     [endpoint],
-    () => api.get(endpoint).then((response) => response.data),
-    {
-      retry: false,
-    },
+    async () => (await api.get(endpoint).json()) as Data,
```
sodic: Hm, how come this assertion is necessary?

examples/kitchen-sink/src/features/jobs/pages/JobsPage.tsx
```diff
@@ -161,7 +160,7 @@ function StatusBadge({ state }: { state: UppercaseTextRequestState }) {
 }
 
 function getMutationErrorMessage(error: unknown): string | undefined {
-  if (error instanceof AxiosError) {
+  if (error instanceof Error) {
```
sodic, line 163: This is much broader than before, and I think it changes the behavior. Is it on purpose?

If so, please leave a PR comment explaining the change in the future, since this PR reads like it's only supposed to be a refactor.

waspc/data/Generator/templates/sdk/wasp/auth/email/actions/login.ts
```diff
@@ -1,13 +1,12 @@
...
-    try {
-        const response = await api.post('{= loginPath =}', data);
-        await initSession(response.data.sessionId);
-    } catch (e) {
-        throw handleApiError(e);
-    }
+    const { sessionId } = await api.post('{= loginPath =}', {
+        json: data,
+    }).json(SessionResponseSchema);
+    await initSession(sessionId);
```
sodic: This (and kicking out `handleApiError`) is a behavioral change as well, right? Did we want to change it, and what changes and how?

waspc/data/Generator/templates/sdk/wasp/auth/useAuth.ts
```diff
@@ -20,13 +20,9 @@ function createUserGetter(): Query<void, AuthUser | null> {
   const getMeRelativePath = 'auth/me'
   const getMeRoute = { method: HttpMethod.Get, path: `/${getMeRelativePath}` }
   const getMe: QueryFunction<void, AuthUser | null> = async () =>  {
-    try {
-      const response = await api.get(getMeRoute.path)
-      const userData = deserialize<AuthUserData | null>(response.data)
-      return makeAuthUserIfPossible(userData)
-    } catch (error) {
-      throw handleApiError(error)
```
sodic, line 28: Mentioned this before, but... are we sure we didn't lose some functionality by kicking this out?

waspc/data/Generator/templates/sdk/wasp/auth/useAuth.ts
```diff
@@ -20,13 +20,9 @@ function createUserGetter(): Query<void, AuthUser | null> {
...
-    }
+    const json = await api.get(getMeRoute.path).json()
+    const userData = deserialize<AuthUserData | null>(json as any)
```
sodic: Is the assertion necessary? If so, why - I'd expect `deserialize` to accept `unknown` and `json` to be `unknown` just like `reponse.data`. Did I get something wrong?

web/docs/advanced/apis.md
```diff
@@ -102,15 +102,15 @@ For example, if your app is running at `https://example.com` then from the above
...
 
 ```tsx title="src/pages/SomePage.tsx" auto-js with-hole
 import React, { useEffect } from "react";
 import { api } from "wasp/client/api";
 
 async function fetchCustomRoute() {
-  const res = await api.get("/foo/bar");
-  console.log(res.data);
+  const data = await api.get("/foo/bar").json();
+  console.log(data);
```
sodic, line 113: I think this exampel shows a different thing than before: text vs data. Is it on purpose? If so, we should rename the variable to something like `jsonData`.

web/docs/migration-guides/migrate-from-0-21-to-0-22.md
```diff
@@ -101,7 +105,61 @@ app MyApp {
...
+The `api` object was previously an Axios instance. It is now a [ky](https://github.com/sindresorhus/ky) instance with a pre-configured base URL and authentication. Update your code as follows:
+
+<Tabs>
+  <TabItem value="before" label="Before">
+    ```ts
+    import { api } from 'wasp/client/api'
+
+    // Making requests
+    const response = await api.get('/foo/bar')
+    const data = response.data
```
sodic: Is this also a different usage than what we had so far?

web/docs/migration-guides/migrate-from-0-21-to-0-22.md
```diff
@@ -101,7 +105,61 @@ app MyApp {
 
 Review your schemas for compatibility with Zod v4. Most schemas will work without changes, but some deprecated APIs have been removed. Refer to the [Zod v4 migration guide](https://zod.dev/v4/changelog) for details.
 
-### 5. Update Alpine packages in your Dockerfile
+### 5. Update client code that uses `api` from `wasp/client/api`
```
sodic: What about the streaming responses API? I didn't check all the details, but I have a feeling that Wasp's public API relied on Axios more than what we've covered here.

waspc/data/Generator/templates/sdk/wasp/client/operations/internal/index.ts
```diff
@@ -7,13 +7,11 @@ export type OperationRoute = { method: HttpMethod.Post, path: string }
...
-  try {
-    const serializedArgs = serialize(args)
-    const response = await api.post(operationRoute.path, serializedArgs)
-    return deserialize(response.data)
-  } catch (error) {
-    throw handleApiError(error)
-  }
+  const serializedArgs = serialize(args)
+  const json = await api.post(operationRoute.path, {
+    json: serializedArgs,
```
sodic: How come this changed from `payload` to `{ json: payload }`. That's unexpected unless Axios wrapped the entire thing into `json` under the hood, which sounds suspicious.

waspc/data/Generator/templates/sdk/wasp/client/operations/internal/index.ts
```diff
@@ -7,13 +7,11 @@ export type OperationRoute = { method: HttpMethod.Post, path: string }
...
+  }).json()
+  return deserialize(json as any)
```
sodic: Hm, another suspicious `any` assertion. Doesn't `deserialize` work with unknown?

waspc/data/Generator/templates/sdk/wasp/api/index.ts
sodic: Oh, I see.

I think I expressed my concern about losing this elsewhere. Is the new behavior equivalent to the old one? If not, do we like the changes and how do they look?

Also, we kicked out `WaspHttpError`, how come? As far as I undertand, that's unrelated to `axios` vs `ky`.

We also lost all comments explaining why and what happens in this function, was that intentional?

sodic (commented):
Took a look and left some comments.

The PR looks a little suspicious to me, specifically:
1. It seems we're changing more stuff than strictly necessary for the migration (some of it is refactoring, other parts are even behavioral changes). Not sure if that's intentional or not.
2. Parts of the code (or comments) are left out. I'm also not sure whether this was intentional, whether it affects the behavior, and how much.
3. I'd expect more public api changes to follow from removing axios. I know at least one is missing from the migration guide (the streaming thing), but I expect even more. I have a feeling we're leaking more to the outside than we're letting on. Maybe I'm wrong, but please double check.

examples/kitchen-sink/src/features/apis/pages/ApisPage.tsx
cprecioso: It's not, this is a leftover from a previous version

examples/kitchen-sink/src/features/jobs/pages/JobsPage.tsx
cprecioso: > since this PR reads like it's only supposed to be a refactor

mmm it's marked as a breaking change, it's not a simple refactor, we're changing APIs! 😅

cprecioso: indeed on this specific thing, it does strictly change behaviour, but it makes sense to print the error message wherever there's an error we can access, right? there's nothing specific to WaspHttpError in this function

waspc/data/Generator/templates/sdk/wasp/auth/email/actions/login.ts
cprecioso: I moved the error handler to inside the `ky` instance. It is not a behavioral change. Every callsite of the Axios API in the SDK was calling the `handleApiError` in their `catch`, so I just foldedt it into the common denominator.

waspc/data/Generator/templates/sdk/wasp/auth/useAuth.ts
cprecioso: Resolving, already handled in previous thread.

waspc/data/Generator/templates/sdk/wasp/auth/useAuth.ts
cprecioso: `deserialize` doesn't accept `unknown`

image

web/docs/advanced/apis.md
cprecioso: The two snippets do the same thing, as Axios decodes JSON by default, Ky does not.

web/docs/migration-guides/migrate-from-0-21-to-0-22.md
cprecioso: no, same answer as before

web/docs/migration-guides/migrate-from-0-21-to-0-22.md
cprecioso: We had no API for streaming responses, because Axios had no first-class support for it (being XHR-based). Check the modifications to `web/docs/advanced/apis.md` in this PR. We were telling people to use native `fetch` or write a wrapper around some messy parts of Axios. Now they don't need to do either, as `ky` just has this by default. And I updated the docs.

waspc/data/Generator/templates/sdk/wasp/client/operations/internal/index.ts
cprecioso: > That's unexpected unless Axios wrapped the entire thing into json under the hood, which sounds suspicious.

Axios wrapped the entire thing into json under the hood. Now you're starting to get why I didn't like Axios 😂

waspc/data/Generator/templates/sdk/wasp/client/operations/internal/index.ts
cprecioso: Resolving, same comment as before

sodic (approved):
Approve, please do what we discussed :)

cprecioso:
In the end I restored the WaspHttpError behavior and catchers, as I want to close this before I go and don't have good time before I leave. Will merge once CI passes.

